PR #1138, opened 2021-08-13
skip connecting to invalid identities in the identity table and log error instead of fatal

vishalchangrani:
Found one more place where it was `log.fatal` on a bad identity.


zhangchiqing (approved):

yhassanzadeh13 (approved):

network/p2p/peerManager.go
```diff
@@ -118,15 +118,8 @@ func (pm *PeerManager) updatePeers() {
...
-		return
-	}
-
-	if IsUnconvertibleIdentitiesError(err) {
-		// log conversion error as fatal since it indicates a bad identity table
-		pm.logger.Fatal().Err(err).Msg("failed to connect to peers")
-		return
+	if err != nil {
+		// one of more identities in the identity table could not be connected to
+		pm.logger.Error().Err(err).Msg("failed to connect to one or more peers")
```
huitseeker, line 123: It seems this can only be an `NewUnconvertableIdentitiesError` in our implementation. Could we unpack this case and thereby maybe get a more usable error message?

huitseeker (commented):
Nice catch!

network/p2p/peerManager.go
vishalchangrani: so `UnconvertableIdentitiesError` is [logged as multiple ](https://github.com/onflow/flow-go/blob/187f08c4724d84f6d77e5de49868fe22842b6b18/network/p2p/libp2pConnector.go#L40-#L46)individual errors (multi-error) which will list out the Identity and the error reason.
Will that suffice?

huitseeker: yep, all good!

huitseeker (approved):
thanks!

huitseeker:
@vishalchangrani you could use bors to merge this

vishalchangrani:
bors merge

PR #1139, opened 2021-08-14
GCP Bucket block data uploader

m4ksio:
I don't find leaking of `TrieUpdate` elegant, but I don't really see other easy way to achieve this exact data structure without recalculating it.
Maybe there should be two separate components gathering data and matching it before uploading, but it seems even more far fetched than current solution


engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			u.log.Warn().Err(err).Str("object_name", objectName).Msg("error while closing GCP object")
+		}
+	}()
+
+	return WriteComputationResultsTo(computationResult, writer)
+}
+
+func GCPBlockDataObjectName(computationResult *execution.ComputationResult) string {
+	return fmt.Sprintf("%s.cbor", computationResult.ExecutableBlock.ID().String())
```
huitseeker, line 104: IIUC, this will silently overwrite the object.
1. do we want to document this in a comment?
2. do we want to consider object versioning in the case of multiple writers?
3. do we want to have GCPBucketUploader accept a constructor string prefix (that would allow e.g. a hot-swappable secondary) ?

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -0,0 +1,116 @@
...
+	<-async.Done()
+
+	require.Equal(t, int64(0), metrics.Counter.Load())
+}
+
+func Test_GCPBucketUploader(t *testing.T) {
+
+	t.Skip("requires GCP Bucket setup")
+
+	bucketName := os.Getenv("FLOW_TEST_GCP_BUCKET")
```
huitseeker, line 63: What are the expectations on running this test? In particular, do I need a signed URL to run this?

engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (a *AsyncUploader) Upload(computationResult *execution.ComputationResult) error {
+	a.unit.Launch(func() {
+		a.metrics.ExecutionBlockDataUploadStarted()
+		defer a.metrics.ExecutionBlockDataUploadFinished()
+
+		err := a.uploader.Upload(computationResult)
+		if err != nil {
+			a.log.Error().Err(err).Msg("error while uploading block data")
```
huitseeker, line 52: Maybe a dumb question, but considering that if the `Upload` call returns an `error`, it's probably the percolation of a `cbor` error, would we want to make sure we can log the embedded error in detail?

huitseeker (commented):
Thanks! This looks really good, but there's a bunch of places on which I'd love clarification on the details.

module/metrics/execution.go
```diff
@@ -636,6 +646,14 @@ func (ec *ExecutionCollector) ExecutionCollectionRequestRetried() {
...
 }
 
+func (ec *ExecutionCollector) ExecutionBlockDataUploadStarted() {
+	ec.blockDataUploadsInProgress.Inc()
+}
+
+func (ec *ExecutionCollector) ExecutionBlockDataUploadFinished() {
+	ec.blockDataUploadsInProgress.Dec()
+}
+
```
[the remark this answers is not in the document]
m4ksio: That would be highly correlated with block size (or rather changes generated by the block). But why not

engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -0,0 +1,128 @@
...
+		dir: dir,
+	}
+}
+
+type FileUploader struct {
+	dir string
+}
+
+func (f *FileUploader) Upload(computationResult *execution.ComputationResult) error {
+	file, err := os.Create(path.Join(f.dir, fmt.Sprintf("%s.cbor", computationResult.ExecutableBlock.ID())))
```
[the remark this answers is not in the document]
m4ksio: We should probably rather expose mapping ComputationResult/BlockID -> filename as a function, but that can wait once we have some new format in mind

engine/execution/computation/computer/uploader/model.go
```diff
@@ -0,0 +1,50 @@
...
+		TxResults:   txResults,
+		Events:      events,
+		TrieUpdates: computationResult.TrieUpdates,
+	}
+}
+
+func WriteComputationResultsTo(computationResult *execution.ComputationResult, writer io.Writer) error {
+	blockData := ComputationResultToBlockData(computationResult)
+
+	encoder := cbor.NewEncoder(writer)
```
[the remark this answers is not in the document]
m4ksio: Doesn't those option refer to private repo? This way we would need DPS to have access to the same

engine/execution/computation/computer/uploader/model.go
```diff
@@ -0,0 +1,50 @@
...
+	txResults := make([]*flow.TransactionResult, len(computationResult.TransactionResults))
+	for i := 0; i < len(computationResult.TransactionResults); i++ {
+		txResults[i] = &computationResult.TransactionResults[i]
+	}
+
+	events := make([]*flow.Event, 0)
+	for _, eventsList := range computationResult.Events {
+		for i := 0; i < len(eventsList); i++ {
+			events = append(events, &eventsList[i])
+		}
```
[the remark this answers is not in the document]
m4ksio: Yep, go gotcha, when iteration variable is reused and it has the same address all the time.

engine/execution/computation/computer/uploader/model.go
```diff
@@ -0,0 +1,50 @@
...
+
+type BlockData struct {
+	Block       *flow.Block
+	Collections []*entity.CompleteCollection
+	TxResults   []*flow.TransactionResult
+	Events      []*flow.Event
+	TrieUpdates []*ledger.TrieUpdate
+}
+
+func ComputationResultToBlockData(computationResult *execution.ComputationResult) *BlockData {
```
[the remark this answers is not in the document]
m4ksio: We only read the object, so I don't think we need any guarantees at this point. Assumption is that object is not modified while its being serialized.
The target format is something we agreed with DPS team

engine/execution/computation/computer/uploader/uploader.go
m4ksio: While all those comments are valid question, I think we should address them in next planned version.

engine/execution/computation/computer/uploader/uploader_test.go
m4ksio: You need to have configured access to a named GCP bucket

engine/execution/computation/computer/uploader/uploader.go
m4ksio: we log error which should log entire structure, won't it?

m4ksio:
bors try

m4ksio:
bors try

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
huitseeker: So, I'm not suggesting you use the options as defined in the network package. I'm suggesting you use the choice of standard native cbor options in the same way that the network package does. In other words, that's by value, not by reference.

In particular [`CoreDetEncOptions`](https://github.com/fxamacker/cbor/blob/6cf31e9ea023629801ee28fc133a743f0be20467/encode.go#L358) seems important.

huitseeker (approved):
Thanks a bunch!

vishalchangrani:
is there a reason why we did this in-code versus having a cron job sitting on the execution node uploading files periodically?

m4ksio:
> is there a reason why we did this in-code versus having a cron job sitting on the execution node uploading files periodically?

A few reasons actually:
1. Latency. This way we upload stuff as soon as its produced
2. This data is not really easily available on disk for cron job to pick up. EN has it in memory.
3. Its different format than WAL. Its per block data, easy to digest by DPS, WAL is per update but its unknown if a segment has complete set of changes for block, whether one update spans multiple blocks etc.

engine/execution/computation/computer/uploader/uploader.go
smnzhu: I think object versioning would be setup on the GCP bucket itself, so it wouldnt involve any changes here. GCS would automatically update the version when a new file of the same name is uploaded.

cmd/execution/main.go
```diff
@@ -161,6 +176,36 @@ func main() {
...
+
+				blockDataUploader = asyncUploader
+
+				return asyncUploader, nil
+			}
+
+			// Since we don't have conditional component creation, we just use Noop one.
+			// It's functions will be once per startup/shutdown - non-measurable performance penalty
+			// blockDataUploader will stay nil and disable calling uploader at all
+			return &module.NoopReadDoneAware{}, nil
```
smnzhu, line 206: An alternative to this is to just omit this particular call to `.Component` altogether if `enableblockDataUpload` is false.

smnzhu: This would require you to store the NodeBuilder in a variable first.

cmd/execution/main.go
```diff
@@ -161,6 +176,36 @@ func main() {
...
+					return nil, fmt.Errorf("cannot create GCP Bucket uploader: %w", err)
+				}
+
+				asyncUploader := uploader.NewAsyncUploader(
+					gcpBucketUploader,
+					logger,
+					collector,
+				)
+
+				blockDataUploader = asyncUploader
```
smnzhu, line 198: Do we want to use the async uploader by default? It never returns any error, which means that the computation manager will not know about any errors that occur.

In general, this means that if even a single upload fails, we will never try again and DPS would get stuck at that point because that the result for that block would never be uploaded right?

Do we want to build in some retry mechanism for the uploader so it can hopefully succeed to upload blocks in the face of transient failures?

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
m4ksio: > is this assumption actually guaranteed everywhere in the code?
No. Go in general lacks immutability of objects.

> And should we document this, so that people know that trying to perform the serialization while the object is being modified may have unexpected results?
Bigger part of our codebase (or any Go codebase really) would need such warning. This function does not modify the object, so if called is expecting modification it should create copy first.

cmd/execution/main.go
m4ksio: Not really, as flags are parsed only after NodeBuilder is finished configuring and run

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
m4ksio: No.
`event` in most inner loop would be of type `Event` and we want `*Event`. 
If you do `events = append(events, &event)` you would append the same element multiple times, since `event` is reused in the loop and only initialized once

cmd/execution/main.go
m4ksio: I'm adding some retry logic, however we cannot rely on this for too much - don't want to have queue backs up and destabilise EN operation.
I think this is the very first version, in subsequent we would want ability for multiple EN to write to same bucket.

cmd/execution/main.go
smnzhu: That's not true, they are already parsed by the time `Initialize()` is finished running.

What I'm suggesting is, you could do

```
builder := cmd.FlowNode(...).ExtraFlags(...).ValidateFlags(...).Initialize(...)
if enableUpload {
    builder.Component(...)
}
// continue building the remaining modules
builder.Module(...).
```

See [this](https://github.com/onflow/flow-go/blob/5f3460d1814c27907e21c5a85ff5a44b83a3046e/cmd/access/main.go#L382-L429) for an example

cmd/execution/main.go
smnzhu: Sure, I think it would make sense to retry fewer times, and emit some metric if in the end all retries failed.

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
smnzhu: So in all cases where this function is being used, there would be no modification happening during or after the call, correct? In particular, there is no chance that `cr` will be further modified after this call:

https://github.com/onflow/flow-go/pull/1139/files#diff-9e005ebdb50b934004efede53802cbdda764e9c4090b5605dcf38d16a516b417R78

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -0,0 +1,120 @@
...
+	uploader, err := NewGCPBucketUploader(context.Background(), bucketName, zerolog.Nop())
+	require.NoError(t, err)
+
+	cr := generateComputationResult(t)
+
+	buffer := &bytes.Buffer{}
+	err = WriteComputationResultsTo(cr, buffer)
+	require.NoError(t, err)
+
+	err = uploader.Upload(cr)
```
smnzhu, line 78: Here

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
m4ksio: Fixed

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
m4ksio: No. We intent that serialization happens on an object which won't be modified. In EN this happens after the computation has happen

cmd/execution/main.go
m4ksio: This is not a a best piece of code architecture-wise, as it break builder pattern with back modification of components.
In this PR this would also mean we change configuration of component that has already been initialized. 

Ideally we would separate init process into two objects - configuration builder and than use it to construct Node object, allowing component to have their own list of dependencies for system to initialize, but that's outside of scope of this PR

engine/execution/computation/computer/uploader/model.go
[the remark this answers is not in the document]
smnzhu: Ah yes, so this instead

```go
for _, eventsList := range computationResult.Events {
    for _, event := range eventsList {
      e := &event
	    events = append(events, e)
    }
}
```

cmd/execution/main.go
```diff
@@ -161,6 +178,38 @@ func main() {
 			pendingBlocks = buffer.NewPendingBlocks() // for following main chain consensus
 			return nil
 		}).
+		Component("Block data uploader", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+			if enableBlockDataUpload {
+
+				logger := node.Logger.With().Str("component_name", "block_data_uploader").Logger()
```
vishalchangrani, line 184:
```suggestion
				logger := node.Logger.With().Str("component_name", "block_data_uploader").Str("gcp_bucket", gcpBucketName).Logger()
```

cmd/execution/main.go
```diff
@@ -161,6 +178,38 @@ func main() {
...
+				gcpBucketUploader, err := uploader.NewGCPBucketUploader(
+					context.Background(),
+					gcpBucketName,
+					logger,
+				)
+				if err != nil {
+					return nil, fmt.Errorf("cannot create GCP Bucket uploader: %w", err)
+				}
+
+				asyncUploader := uploader.NewAsyncUploader(
```
vishalchangrani, line 194:
```suggestion
				blockDataUploader = uploader.NewAsyncUploader(
```

vishalchangrani: nit: and remove the temp variable `asyncUpdloader`

engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -0,0 +1,151 @@
...
+		err := retry.Do(a.unit.Ctx(), cappedFibRetry, func(ctx context.Context) error {
+			err := a.uploader.Upload(computationResult)
+			if err != nil {
+				a.log.Warn().Err(err).Msg("error while uploading block data, retrying")
+			}
+			return retry.RetryableError(err)
+		})
+
+		if err != nil {
+			a.log.Error().Err(err).Msg("failed to upload block data")
```
smnzhu, line 73:
```suggestion
			a.log.Error().Err(err).Str("block_id", computationResult.ExecutableBlock.Block.ID()).Msg("failed to upload block data")
```

Just so we know which one failed if we need to manually fix it later

engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		a.metrics.ExecutionBlockDataUploadFinished(time.Since(start))
```
smnzhu, line 76: nit: If `err != nil` above, do we want to immediately `return` after logging the error or do we we still want to emit this "UploadFinished" metric?

If we do, would it make sense to have different metric to distinguish between "UploadFinished" and "UploadFailed"?

smnzhu (approved):
Thanks! Looks good for the most part, just left two more comments.

cmd/execution/main.go
m4ksio: Bucket name doesn't change, I don't think will get anything extra from logging this but bloating logs

cmd/execution/main.go
vishalchangrani: but the component wouldn't be initialized if `enableUpload` is `false`. I think skipping the component completely would be inline with what is being done with other optional components as well.

m4ksio: Agreed, but doing this properly would require refactoring. I find this more coherent and performance penalty is non-existing

engine/execution/computation/computer/uploader/uploader.go
m4ksio: We always want to emit it, otherwise there will be "pending" uploads, as we monitor queue size. 
Failing uploads aren't something we anticipate really, so no point in having a metrics, failing upload is rather a showstopper.

vishalchangrani (approved):

PR #1140, opened 2021-08-14
[Consensus Voting] Initial Interfaces

zhangchiqing:
<del>NOTE: Please review https://github.com/onflow/flow-go/pull/1142 first</del>

This PR attempts to merge the existing changes on the consensus voting feature branch to master.

The feature branch has accumulated some changes that have been reviewed. Since they are interfaces and module implementations that are not being used by existing hotstuff, merging them to master will not change any node behavior. But good to allow CI and tests to verify the correctness.

Once the interfaces are merged, the plan moving forward is to make future consensus voting related PRs to master if possible. And module will not be used until a "swap" PR that swaps the new hotstuff with the old hotstuff gets merged to master. 

And that "swap" PR will be a feature branch that includes minimal code and tests needed for the swap.


consensus/hotstuff/randombeacon_signer.go
```diff
@@ -0,0 +1,57 @@
+package hotstuff
+
+import "github.com/onflow/flow-go/crypto"
+
+type RandomBeaconSigner interface {
```
durkmurder, line 5: I am lost again, can you explain why do we have `RandomBeaconSigner` interface which can verify, aggregate and sign and we also have `RandomBeaconAggregator` and `RandomBeaconReconstructor` which are very similar and know only how to aggregate, appreciate use cases as well.

consensus/hotstuff/randombeacon_signer.go
```diff
@@ -0,0 +1,57 @@
...
+	// return other error if there is exception
+	Verify(signerIndex int, share crypto.Signature) error
+
+	// TrustedAdd adds a verified share to the internal signature shares store
+	// The operation is sequential.
+	// It assumes the signature share has been verified and is valid.
+	// return (true, nil) if the signature has been added
+	// return (false, nil) if the signature is a duplication
+	// return (false, error) if there is any exception
+	// TODO: should we let it also return EnoughShares?, because otherwise
```
durkmurder, line 19: Generally speaking I would return if it has accumulated enough weight in `TrustedAdd` this makes easier to use it.

consensus/hotstuff/signature.go
```diff
@@ -0,0 +1,60 @@
+package hotstuff
+
+import (
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// RandomBeaconReconstructor collects verified signature shares, and reconstructs the
+// group signature with enough shares.
+type RandomBeaconReconstructor interface {
```
durkmurder, line 10: Very confusing, what is the difference between `RandomBeaconReconstructor` and `RandomBeaconAggregator` ?

consensus/hotstuff/signature.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// SignatureAggregator aggregates the verified signatures.
+// It can either be used to aggregate staking signatures or aggregate random beacon signatures,
+// but not a mix of staking signature and random beacon signature.
+type SignatureAggregator interface {
+	// TrustedAdd adds an already verified signature.
+	// return (true, nil) means the signature has been added
+	// return (false, nil) means the signature is a duplication
+	TrustedAdd(signerID flow.Identifier, sig crypto.Signature) (bool, error)
```
durkmurder, line 37: I think we are missing a getter here to report how much weight we have or an extra return value for `TrustedAdd` to report weight

consensus/hotstuff/signature/randombeacon_aggregator.go
```diff
@@ -0,0 +1,5 @@
+package signature
+
+// RandomBeaconAggregtaorImpl aggregates the random beacon signatures
+type RandomBeaconAggregtaorImpl struct {
```
durkmurder, line 4:
```suggestion
type RandomBeaconAggregatorImpl struct {
```

consensus/hotstuff/voteaggregator/interfaces.go
```diff
@@ -0,0 +1,31 @@
...
+// VoteCollectors is an interface which allows VoteAggregator to interact with collectors structured by
+// view and blockID.
+// Implementations of this interface are responsible for state transitions of `VoteCollector`s and pruning of
+// stale and outdated collectors by view.
+type VoteCollectors interface {
+	// GetOrCreateCollector is used for getting hotstuff.VoteCollector, calling this function for first time
+	// will create a new collector.
+	// collector is indexed by blockID and view
+	// It returns the vote collector state machine, true and nil if found,
+	// It returns (nil, false, nil) if not found
```
durkmurder, line 18: There is a typo here, it should return `(collector, true, nil)` or `(collector, false, nil)` in case of success. Nil collector can be only in case of error

consensus/hotstuff/votecollector/combined_aggregator.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// combinedAggregator aggregates signatures from different signers for the same message.
+// The instance must be initialized with the message, the required weight to be
+// sufficient for aggregation, as well as the weight for each signer.
+type combinedAggregator struct {
```
durkmurder, line 13: What interface we are implementing here? I thought it should be `SignatureAggregator` but it has too many functions for it.

durkmurder (changes requested):
I believe we have some duplicated interfaces and implementations, need to fix it before merging to master.

consensus/hotstuff/signature.go
```diff
@@ -0,0 +1,60 @@
...
+type RandomBeaconReconstructor interface {
+	// TrustedAdd a beacon share without verifying it.
+	TrustedAdd(signerID flow.Identifier, sig crypto.Signature) (bool, error)
+	HasSufficientShares() bool
+	// Reconstruct the beacon signature if there are enough shares. It errors if not enough shares were collected.
+	Reconstruct() (crypto.Signature, error)
+}
+
+// RandomBeaconAggregator aggregates the random beacon signatures
+type RandomBeaconAggregator interface {
```
zhangchiqing, line 19: will remove

consensus/hotstuff/votecollector/consensus_verifying_vote_collector.go
```diff
@@ -0,0 +1,16 @@
+package votecollector
+
+import (
+	"sync/atomic"
+
+	"github.com/onflow/flow-go/consensus/hotstuff"
+)
+
+type ConsensusVerifyingVoteCollector struct {
```
durkmurder, line 9: This seems obsolete we should move members [here](https://github.com/onflow/flow-go/pull/1140/files#diff-9e06d16b426797fa5261aa8342f0e5259eadd121ae65e4bca23774728cbbf852R13)

consensus/hotstuff/votecollector/interfaces.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// SigType is the aggregable signature type.
+type SigType int
+
+// There are two signatures are aggregable, one is the normal staking signature,
```
durkmurder, line 11: This whole file doesn't make sense I think since we have other interfaces for aggregators

consensus/hotstuff/votecollector/random_beacon_reconstructor.go
```diff
@@ -0,0 +1,34 @@
+package votecollector
+
+import "github.com/onflow/flow-go/crypto"
+
+type randomBeaconReconstructor struct {
```
durkmurder, line 5: Obsolete

consensus/hotstuff/votecollector/combined_aggregator.go
durkmurder: Seems obsolete

consensus/hotstuff/vote_aggregator_v2.go
```diff
@@ -0,0 +1,55 @@
...
+// TODO: rename to remove V2 when replacing V1
+type VoteAggregatorV2 interface {
+	// AddVote verifies and aggregates a vote.
+	// The voting block could either be known or unknown.
+	// If the voting block is unknown, the vote won't be processed until AddBlock is called with the block.
+	// This method can be called concurrently, votes will be queued and processed asynchronously.
+	// It returns:
+	//  - nil if the vote has been added to the queue for async processing
+	//  - error if there is exception adding the vote to the queue
+
```
durkmurder, line 37:
```suggestion
```

consensus/hotstuff/vote_aggregator_v2.go
```diff
@@ -0,0 +1,55 @@
...
+	AddVote(vote *model.Vote) error
+
+	// AddBlock notifies the VoteAggregator about a known block so that it can start processing
+	// pending votes whose block was unknown.
+	// It also verifies the proposer vote of a block, and return whether the proposer signature is valid.
+	// It returns:
+	// 	- nil if the block is valid
```
durkmurder, line 44:
```suggestion
	// - nil if the block is valid
```

consensus/hotstuff/vote_collector.go
```diff
@@ -0,0 +1,92 @@
...
+	// Status returns the status of the vote collector
+	Status() VoteCollectorStatus
+}
+
+// VerifyingVoteCollector is a VoteCollector and also implement the same interface as BlockSigner, so that
+// when the voter ask VoteAggregator(via BlockSigner interface)
+// to sign the block, and VoteAggregator will read the VerifyingVoteCollector from the vote collectors
+// map and produce the vote.
+// Note CachingVoteCollector can't create vote, only VerifyingVoteCollector can
+type VerifyingVoteCollector interface {
```
durkmurder, line 89: Not sure we need to expose `VerifyingVoteCollector` at this level

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
```diff
@@ -0,0 +1,177 @@
...
+	unit                *engine.Unit
+	log                 zerolog.Logger
+	notifier            hotstuff.Consumer
+	committee           hotstuff.Committee
+	voteValidator       hotstuff.Validator
+	signer              hotstuff.SignerVerifier
+	highestPrunedView   counters.StrictMonotonousCounter
+	collectors          hotstuff.VoteCollectors
+	queuedVotesNotifier engine.Notifier
+	queuedVotes         *fifoqueue.FifoQueue // keeps track of votes whose blocks can not be found
```
durkmurder, line 29:
```suggestion
	queuedVotes         *fifoqueue.FifoQueue
```

durkmurder (approved):
Looks good at this point. Thanks for all the discussions and revisions. Good job!

consensus/hotstuff/vote_collector.go
zhangchiqing: The StateMachine will see this interface instead of a concrete struct, such that the collection cluster could provide a different implementation

huitseeker (commented):
I wonder if there isn't a missing API on the consensus vote / random beacon signer, that would allow one to query the identity of the nodes which have so far provided a (valid) vote to each collector.

I think we need this API, because we need to prevent cross-signature-format double voting.

In other words, right now, we (have in mind to) prevent:
- node N sends a staking key signature, then node N sends another staking key signature (staking double vote)
- node N sends a beacon signature, then node N sends another beacon signature (beacon double vote)

We also need to think about:
- node N sends a staking key signature, then node N sends a beacon signature (staking-beacon double vote, since beacon keys are counted towards HS)
- node N sends a beacon signature, then node N sends a staking key signature (beacon-staking double vote, since beacon keys are counted towards HS)

So when I accept a "valid" vote from node N, and I'm about to add it to collector c (depending on the type of this valid vote), I need to know if the *other* collector has already received a vote from N.

tarakby (commented):
Another round of review mostly focused on the API descriptions. 
I'm not done yet but I still send these comments for now 😬

zhangchiqing:
Close for now. 

This branch will be used as master branch for the consensus voting v2 feature

PR #1142, opened 2021-08-14
[Consensus] Enable linting on consensus

zhangchiqing:
Just noticed the linting was disabled on consensus/hotstuff package. This PR fixes the linting errors after enabling it again

PR #1144, opened 2021-08-15
[Consensus] Fix linting

zhangchiqing:
Review https://github.com/onflow/flow-go/pull/1142 first

Branch `feature/consensus-voting` doesn't compile, because the linting was not turned on. https://github.com/onflow/flow-go/pull/1142 fixes the linting, so please review that first.

This PR merges 1142 to the feature branch, so I can fix the linting on feature branch while waiting for 1142 to be reviewed.

I think I've fixed everything except errors in vote aggregator v2. @durkmurder could you help fixing the rest?


durkmurder:
The problem with linting is that all of remaining errors are due to unused variables/functions. We can't simply fix it, need to write code, disable linting or comment it out. I have disabled linting in my [PR](https://github.com/onflow/flow-go/pull/1146).

PR #1146, opened 2021-08-16
[Consensus] Disabled linting for unused symbols in new implementation

durkmurder:
### Context 

The problem with linting is that all of remaining errors are due to unused variables/functions. We can't simply fix it, need to write code, disable linting or comment it out.


zhangchiqing:
Yeah, makes sense

zhangchiqing (approved):

PR #1148, opened 2021-08-16
Fix leader selection error

zhangchiqing:


consensus/hotstuff/committees/leader/cluster.go
```diff
@@ -23,6 +23,7 @@ func SelectionForCluster(cluster protocol.Cluster, epoch protocol.Epoch) (*Leade
 	identities := cluster.Members()
 	seed, err := epoch.Seed(indices.ProtocolCollectorClusterLeaderSelection(cluster.Index())...)
 	if err != nil {
+		return nil, fmt.Errorf("could not get leader selection seed for cluster (index: %v) at epoch: %v: %w", cluster.Index(), counter, err)
```
huitseeker, line 26: Does it make sense to continue at all here?

huitseeker (commented):
Nice catch!

jordanschalm (approved):

consensus/hotstuff/committees/leader/cluster.go
jordanschalm: I don't think so. An error here indicates a critical problem with the `epoch` object we receive as argument

zhangchiqing:
bors merge

zhangchiqing:
bors merge

PR #1151, opened 2021-08-16
increasing timeout of transit pull command to accommodate the download of the root.checkpoint

vishalchangrani:
T-systems ran into an issue in mainnet-8 spork where the transit script key timing out when trying to download root.checkpoint. `root.checkpoint` can be upwards of 18GB and may take more than 30 seconds to download.
Hence bumping up the timeout value for `transit pull`


zhangchiqing (approved):

NhlanhlaHasane (approved):

huitseeker (approved):
Amazing that they were the only ones to hit this!

PR #1155, opened 2021-08-17
add generic retry func

kc1116:
This PR refactors a generic try func into the utils dir as to stay DRY, and adds exponential backoff retry to DKG private message sending.

### DIFF
--- a/engine/consensus/dkg/messaging_engine.go
+++ b/engine/consensus/dkg/messaging_engine.go
@@ -128,7 +128,6 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 			}
 
 			retry.BackoffExponential(f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
-
 		case <-e.unit.Quit():
 			return
 		}
--- a/engine/consensus/dkg/messaging_engine.go
+++ b/engine/consensus/dkg/messaging_engine.go
@@ -1,6 +1,7 @@
 package dkg
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/onflow/flow-go/utils/retry"
@@ -118,7 +119,7 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 	for {
 		select {
 		case msg := <-e.tunnel.MsgChOut:
-			f := func() error {
+			f := func(ctx context.Context) error {
 				err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
 				if err != nil {
 					return fmt.Errorf("error sending dkg message: %v", err)
@@ -127,7 +128,7 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 				return nil
 			}
 
-			retry.BackoffExponential(f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
+			retry.BackoffExponential(context.TODO(),"MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
 		case <-e.unit.Quit():
 			return
 		}
--- a/engine/consensus/dkg/messaging_engine.go
+++ b/engine/consensus/dkg/messaging_engine.go
@@ -128,7 +128,7 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 				return nil
 			}
 
-			retry.BackoffExponential(context.TODO(),"MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
+			retry.BackoffExponential(e.unit.Ctx(),"MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
 		case <-e.unit.Quit():
 			return
 		}
--- a/engine/consensus/dkg/messaging_engine.go
+++ b/engine/consensus/dkg/messaging_engine.go
@@ -128,7 +128,7 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 				return nil
 			}
 
-			retry.BackoffExponential(e.unit.Ctx(),"MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
+			retry.BackoffExponential(e.unit.Ctx(), "MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
 		case <-e.unit.Quit():
 			return
 		}
--- a/engine/consensus/dkg/messaging_engine.go
+++ b/engine/consensus/dkg/messaging_engine.go
@@ -3,6 +3,8 @@ package dkg
 import (
 	"fmt"
 
+	"github.com/onflow/flow-go/utils/retry"
+
 	"github.com/rs/zerolog"
 
 	"github.com/onflow/flow-go/engine"
@@ -12,6 +14,13 @@ import (
 	"github.com/onflow/flow-go/module/dkg"
 	"github.com/onflow/flow-go/network"
 )
+
+// RETRY_MAX is the maximum number of times the engine will attempt to broadcast
+// a message
+const RETRY_MAX = 5
+
+// RETRY_MILLISECONDS is the number of milliseconds to wait between retries
+const RETRY_MILLISECONDS = 1000
 
 // MessagingEngine is a network engine that enables DKG nodes to exchange
 // private messages over the network.
@@ -109,10 +118,17 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
 	for {
 		select {
 		case msg := <-e.tunnel.MsgChOut:
-			err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
-			if err != nil {
-				e.log.Err(err).Msg("error sending dkg message")
+			f := func() error {
+				err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
+				if err != nil {
+					return fmt.Errorf("error sending dkg message: %v", err)
+				}
+
+				return nil
 			}
+
+			retry.BackoffExponential(f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
+
 		case <-e.unit.Quit():
 			return
 		}
--- a/module/epochs/base_client.go
+++ b/module/epochs/base_client.go
@@ -98,21 +98,21 @@ func (c *BaseClient) WaitForSealed(ctx context.Context, txID sdk.Identifier, sta
 		return nil
 	}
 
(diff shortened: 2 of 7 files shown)


engine/consensus/dkg/messaging_engine.go
```diff
@@ -13,6 +15,13 @@ import (
 	"github.com/onflow/flow-go/network"
 )
 
+// RETRY_MAX is the maximum number of times the engine will attempt to broadcast
+// a message
+const RETRY_MAX = 5
+
+// RETRY_MILLISECONDS is the number of milliseconds to wait between retries
```
tarakby, line 22: Maybe this is more accurate since the wait between retries is not constant
```suggestion
// RETRY_MILLISECONDS is the number of milliseconds to wait between the two first tries
```

utils/retry/retry.go
```diff
@@ -0,0 +1,24 @@
...
+
+// BackoffExponential makes maxRetry attempts to executed function f. After the nth attempt,
+// we wait retryMilliseconds*2^n ms before retrying. Returns an error after
+// maxRetry unsuccessful attempts.
+func BackoffExponential(f func() error, maxRetry int, retryMilliseconds int, logger zerolog.Logger) bool {
+	for attempt := 1; attempt <= maxRetry; attempt++ {
+		err := f()
+		if err != nil {
+			logger.Warn().Err(err).Msgf("attempt %d/%d failed", attempt, maxRetry)
+			wait := time.Duration(retryMilliseconds<<(attempt-1)) * time.Millisecond
```
tarakby, line 17: 👌

tarakby: Same outcome, different code
```
wait := time.Duration(retryMilliseconds) * time.Millisecond
for attempt := 1; attempt <= maxRetry; attempt++ { 
   .. 
   time.Sleep(wait)
   wait <<= 1
}

**Review by tarakby — APPROVED** [review]
Nice 👌

**Comment by tarakby** [comment]
I forgot to add this comment:
I suggest to use two different `RETRY_MAX` constants for `Broadcast` and `forwardOutgoingMessages`. The reason is:
- Based on the logs, I am noticing frequent failures of broadcasting (reaching attempt 3 and 4 sometimes).
- the consequence of failing to broadcast in DKG is to be disqualified (end up without keys), while failing to send a private message is recoverable. 

I suggest to keep `5` for `forwardOutgoingMessages` and `8` for `Broadcast`. 

cc @jordanschalm

File: utils/retry/retry.go [hunk 3]
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// BackoffExponential makes maxRetry attempts to executed function f. After the nth attempt,
+// we wait retryMilliseconds*2^n ms before retrying. Returns an error after
+// maxRetry unsuccessful attempts.
+func BackoffExponential(f func() error, maxRetry int, retryMilliseconds int, logger zerolog.Logger) bool {
```
m4ksio, line 12: We would need ability to cancel this retries, ie, in case of node shutdown.
That would probably require this to be turned into a component with `ReadyDoneAware` quantities.

kc1116: We can add that in a separate ticket and PR , this retry for now solves our issue for retrying messages in a tight loop.

kc1116: I am in the process if adding a func RetryWithContext . This is more code, but cleaner.

utils/retry/retry.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+		}
+		return true
+	}
+	return false
+}
+
+// WithTimeout attempts to execute function f with a retryMilliseconds sleep in between attempts
+// until the function f either succeeds, context is done or timeout has been reached.
+func WithTimeout(ctx context.Context, retryFuncName string, f func(context.Context) error, timeoutDuration time.Duration, retryMilliseconds int, logger zerolog.Logger) bool {
```
jordanschalm, line 35:
```suggestion
func WithTimeout(ctx context.Context, retryFuncName string, f func(context.Context) error, timeoutDuration time.Duration, retryDuration time.Duration, logger zerolog.Logger) bool {
```
Might as well just accept a duration type directly

utils/retry/retry.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// BackoffExponential makes maxRetry attempts to executed function f. After the nth attempt,
+// we wait retryMilliseconds*2^n ms before retrying. Returns an error after
+// maxRetry unsuccessful attempts. It will short circuit if ctx is Done.
+func BackoffExponential(ctx context.Context, retryFuncName string, f func(context.Context) error, maxRetry int, retryMilliseconds int, logger zerolog.Logger) bool {
```
jordanschalm, line 13:
```suggestion
func BackoffExponential(ctx context.Context, retryFuncName string, f func(context.Context) error, maxRetry int, retryDuration time.Duration, logger zerolog.Logger) bool {
```
same here

utils/retry/retry.go
```diff
@@ -0,0 +1,60 @@
...
+	wait := time.Duration(retryMilliseconds) * time.Millisecond
+	for attempt := 1; attempt <= maxRetry; attempt++ {
+		err := f(ctx)
+		if err != nil {
+			logger.Warn().Err(err).Str("retrying_func", retryFuncName).Msgf("attempt %d/%d failed", attempt, maxRetry)
```
jordanschalm, line 18: It would be nice to include the time elapsed and retry attempt number as structured fields, similar to https://github.com/onflow/flow-go/blob/6cb1baeb5f0a6177de7a41a2501d46edb3c7ca8a/module/epochs/base_client.go#L84

utils/retry/retry.go
```diff
@@ -0,0 +1,60 @@
...
+
+	ctxWithTimeOut, cancel := context.WithTimeout(ctx, timeoutDuration)
+
+	defer cancel()
+
+	var err error
+	for {
+		select {
+		case <-ticker:
+			err = f(ctx)
```
jordanschalm, line 47:
```suggestion
			err = f(ctxWithTimeOut)
```

Since we are attaching deadline information to the context, we should pass it down the stack, so that well-behaved functions can respond to the context being cancelled as soon as possible. 

We know that not all functions we retry here will be able to respect the context cancellation immediately because, for example, sending a private DKG message using the networking layer does not accept a context at all. But we should still it along so that functions that can handle the context cancellation are able to do so (eg. sending dkg broadcast transactions).

utils/retry/retry.go
```diff
@@ -0,0 +1,60 @@
...
+	attempt := 0
+
+	ctxWithTimeOut, cancel := context.WithTimeout(ctx, timeoutDuration)
+
+	defer cancel()
+
+	var err error
+	for {
+		select {
+		case <-ticker:
```
jordanschalm, line 46: I would suggest using `time.After` here as well, so the waiting mechanics are consistent with `BackoffExponential`. `Ticker` and `After` behave a bit differently (I think both are OK here, but we should use the same one for both methods).

jordanschalm (changes requested):
A few more suggestions. 

Blocking because of the changed behaviour of `WaitForSealed`

m4ksio:
Also, I found there are some ready made libraries which provide this, like 
https://github.com/go-retry/retry 
or
https://github.com/sethvargo/go-retry
Maybe it would be much easier to just use one of those?

engine/consensus/dkg/messaging_engine.go
```diff
@@ -109,10 +119,16 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
...
+			f := func(ctx context.Context) error {
+				err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
+				if err != nil {
+					return fmt.Errorf("error sending dkg message: %w", err)
+				}
+
+				return nil
 			}
+
+			retry.BackoffExponential(context.TODO(),"MessagingEngine.forwardOutgoingMessages", f, RETRY_MAX, RETRY_MILLISECONDS, e.log)
```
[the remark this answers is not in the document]
kc1116: Nice I did not know that.

module/epochs/base_client.go
```diff
@@ -78,44 +84,35 @@ func (c *BaseClient) SendTransaction(ctx context.Context, tx *sdk.Transaction) (
...
+	// if the transaction has expired we skip waiting for seal
+	if result.Status == sdk.TransactionStatusExpired {
+		return fmt.Errorf("transaction has expired")
 	}
+
+	if result.Status == sdk.TransactionStatusSealed {
+		return nil
+	}
+
+	return fmt.Errorf("unexpected transaction status %s", result.Status.String())
```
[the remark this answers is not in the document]
kc1116: Just noticed that good catch, I think we can achieve the desired by adding the blocking for loop back. What do you think ?

utils/retry/retry.go
```diff
@@ -0,0 +1,62 @@
...
+	attempt := 0
+
+	ctxWithTimeOut, cancel := context.WithTimeout(ctx, timeoutDuration)
+
+	defer cancel()
+
+	var err error
+	for {
+		select {
+		case <-time.After(wait):
```
jordanschalm, line 49: In this function we wait before the first attempt to call `f` -- in `BackoffExponential` we call `f` immediately. I think we should be consistent between these two methods, and that calling `f` immediately is the best approach for both.

utils/retry/retry.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// BackoffExponential makes maxRetry attempts to executed function f. After the nth attempt,
+// we wait retryMilliseconds*2^n ms before retrying. Returns an error after
+// maxRetry unsuccessful attempts. It will short circuit if ctx is Done.
+func BackoffExponential(ctx context.Context, retryFuncName string, f func(context.Context) error, maxRetry int, retryMilliseconds time.Duration, logger zerolog.Logger) bool {
+	wait := retryMilliseconds * time.Millisecond
+	started := time.Now()
+	for attempt := 1; attempt <= maxRetry; attempt++ {
+		log := logger.With().Int("attempt", attempt).Float64("time_elapsed_s", time.Since(started).Seconds()).Logger()
```
jordanschalm, line 17: could we put time elapsed into the other retry method as well?

utils/retry/retry.go
```diff
@@ -0,0 +1,62 @@
+package retry
+
+import (
+	"context"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// BackoffExponential makes maxRetry attempts to executed function f. After the nth attempt,
```
jordanschalm, line 10: could you update this godoc to correspond to the new method. For example, we now return a bool rather than an error, let's describe what that return value means and under what conditions we return. Likewise for `WithTimeout`

m4ksio:
Just an idea - I used github.com/sethvargo/go-retry in my PR - you can see it in action here https://github.com/onflow/flow-go/pull/1139/files#diff-91e67474ef67f8eaacb1167b906e15ce66594cdb24cf8e6c5994bd5125ab4e70R54-R70

PR #1156, opened 2021-08-17
Fix checking storage over randomly ordered list

janezpodhostnik:
When looping through touched addresses for the purposes of checking storage, the code returns early. The list of touche addresses is acquired from a map of addresses, and thus randomly sorted.

This caused that when we were checking storage and there was a error before the end of the loop, some random addresses were not touched.

This could potentially fail verification at the same step.


fvm/state/state.go
```diff
@@ -197,12 +198,18 @@ func (s *State) MergeState(other *State) error {
...
+
+	// sort list wile constructing it
 	for k := range s.updatedAddresses {
-		addresses = append(addresses, k)
+		i := sort.Search(len(addresses), func(i int) bool { return addresses[i].Hex() >= k.Hex() })
+		addresses = append(addresses, flow.EmptyAddress)
+		copy(addresses[i+1:], addresses[i:])
+		addresses[i] = k
 	}
+
```
huitseeker, line 212:
```
type byHex []flow.Address
func (al byHex) Len() int { return len(a) }
func (al byHex) Swap(i, j int) { al[i], al[j] = al[j], al[i] }
func (al byHex) Less(i, j int) bool { return al[i].Hex() <= al[j].Hex() }
...
addresses := make([]flow.Address, 0, len(s.updatedAddresses))
for _, address := range s.updatedAddresses {
	addresses = append(addresses, address)
}

sort.Sort(byHex(addresses))
```

... is another option.

fvm/state/state.go
```diff
@@ -197,12 +198,18 @@ func (s *State) MergeState(other *State) error {
...
 
-// UpdatedAddresses returns a list of addresses that were updated (at least 1 register update)
+// UpdatedAddresses returns a sorted list of addresses that were updated (at least 1 register update)
 func (s *State) UpdatedAddresses() []flow.Address {
 	addresses := make([]flow.Address, 0, len(s.updatedAddresses))
+
+	// sort list wile constructing it
 	for k := range s.updatedAddresses {
-		addresses = append(addresses, k)
+		i := sort.Search(len(addresses), func(i int) bool { return addresses[i].Hex() >= k.Hex() })
```
m4ksio, line 207: I'm afraid this will be used a lot (for every transaction?) and using `.Hex()` isn't best performance-wide. `bytes.Compare` instead?

huitseeker (commented):
Concept ack, though I'm not the most familiar with that area.

m4ksio (approved):

fvm/state/state.go
ramtinms: I was going to suggest similar thing, using bytes.Compare over slices or conversion to uint64 and compare considering the number of bytes of addresses

ramtinms (approved):

fvm/state/state.go
janezpodhostnik: I would rather do byte comparison... converting to uint introduces an element that could fail.

m4ksio: > converting to uint introduces an element that could fail.

Not really - one of the Go quirks I learned recently, it will silently truncate the value https://golang.org/ref/spec#Conversions

fvm/state/state.go
```diff
@@ -197,12 +198,24 @@ func (s *State) MergeState(other *State) error {
...
+// UpdatedAddresses returns a sorted list of addresses that were updated (at least 1 register update)
 func (s *State) UpdatedAddresses() []flow.Address {
-	addresses := make([]flow.Address, 0, len(s.updatedAddresses))
+	addresses := make(sortedAddresses, len(s.updatedAddresses))
+
+	i := 0
 	for k := range s.updatedAddresses {
-		addresses = append(addresses, k)
+		addresses[i] = k
+		i++
```
ramtinms, line 214: what is i ?

fvm/state/state.go
```diff
@@ -198,18 +198,24 @@ func (s *State) MergeState(other *State) error {
 	return s.checkMaxInteraction()
 }
 
+type sortedAddresses []flow.Address
+
+func (a sortedAddresses) Len() int           { return len(a) }
+func (a sortedAddresses) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a sortedAddresses) Less(i, j int) bool { return bytes.Compare(a[i].Bytes(), a[j].Bytes()) >= 0 }
```
m4ksio, line 205: One more tip is to use `a[i][:]` which should give us byte slice faster than calling separate function.

fvm/state/state.go
```diff
@@ -197,12 +198,24 @@ func (s *State) MergeState(other *State) error {
...
+	addresses := make(sortedAddresses, len(s.updatedAddresses))
+
+	i := 0
 	for k := range s.updatedAddresses {
-		addresses = append(addresses, k)
+		addresses[i] = k
+		i++
 	}
+
+	sort.Sort(addresses)
```
zhangchiqing, line 217: Is there any performance concern with the Sort?

janezpodhostnik (commented):
I'm having a lot of issues getting this through CI

fvm/state/state.go
janezpodhostnik: The list is usually small and the comparison is with byte slices so the performance cost should be small. however, the list needs to be sorted...

zhangchiqing (approved):

huitseeker:
bors merge

janezpodhostnik:
bors merge

janezpodhostnik:
bors merge

PR #1119, opened 2021-08-11
[crypto] update the map-to-curve algorithm used in BLS

tarakby:
- Update the hash to curve used in BLS signature. This doesn't update the algorithm (still simplified SWU) but updates an algorithm parameter to match the [IRTF draft](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/) for hashing to curve (section H.2).
- The new implementation is taken from Relic.
- Relic version is updated to include the new map to curve function (the commits since the last version were audited).
- The new map to curve function is 1.66x slower. Updating the parameter in the older implementation might be a better option to avoid the performance regression (https://github.com/dapperlabs/flow-go/issues/5748)
- The old implementation is left in the code for the future performance investigation, but is disabled. 
- Simplify the tests involving the map to curve (cross BLST and hash-to-curve test vector).

PR #1158, opened 2021-08-17
[crypto] Unplug ValidSignatureBytesBLST

huitseeker:
The blst keygen is unstable on 1.15 (in ways that are hard to reproduce), leading to CI failures. https://github.com/onflow/flow-go/pull/1119 disabled the use of the blst keygen in generating valid slices of scalar bytes for cross-testing bls.go with BLST, this completes the job in de-activating the use of the keygen for creating siganture material.
We leave the auxiliary functions in place for future upgrades (the blst keygen is stable on go 1.16 and some architectures on go 1.15)


tarakby (approved):

Kay-Zee (approved):

huitseeker:
bors merge

PR #1160, opened 2021-08-17
new tag and job for blst tests

tarakby:
- revert all cross-blst tests
- add a new tag for cross-blst tests 
- add a new ci job for cross-blst tests using Go v1.16


.github/workflows/ci.yml
```diff
@@ -96,3 +96,29 @@ jobs:
         timeout_minutes: 15
         max_attempts: 2
         command: make ci-integration
+
+  blst-test:
```
huitseeker, line 100: Looks great! Could we introduce a periodic build as a [scheduled event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) to get signal?

tarakby: From what I read about the schedules, it seems to be possible to schedule the entire workflow, but not a specific job. Did you mean to create a new workflow for the blst tests?

huitseeker: I'm actually fine with running this job step as an out-of-matrix job for now, we can actually move on to dealing with it as a separate workflow if we get further issues from CI.

huitseeker (approved):
Thanks a bunch!

.github/workflows/ci.yml
tarakby: Do you mean to leave the job as it is in the PR for now? 
I'm also looking at how to make the job not required for merging PRs but I don't seem to find the GH settings. Any idea how to set this up?

huitseeker: > Do you mean to leave the job as it is in the PR for now?
Yes

> I'm also looking at how to make the job not required for merging PRs but I don't seem to find the GH settings. Any idea how to set this up?


It's a positive setting in status checks for the repo:
https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging

zhangchiqing (approved):

janezpodhostnik (approved):
LGTM

.github/workflows/ci.yml
tarakby: @Kay-Zee I think I would need your help to set this new job to not required to merge PRs 🙏

PR #1161, opened 2021-08-17
[Consensus Voting] Handle Votes with invalid view

zhangchiqing:
This PR handles a case for consensus voting v2 where vote aggregator processes an invalid vote with a wrong view.

A malicious node might be able to attack block rate by sending invalid votes to cause leader to drop valid votes.

How the attack might work before this fix:
Let's say there is a block A at view 20, and a node's finalized view is 5. so all vote collectors below view 5 have been pruned.
This node is the leader for view 10 and 20. A malicious node sends an invalid vote for blockA with view 10. The leader receives the vote, created a vote collector that is indexed by view 10 only. Later, this leader receives valid votes, and the block. These valid votes, and block are all stored in the vote collector that is indexed by the wrong view (10). 
Later on, block 15 is finalized, and we will prune all vote collectors including this vote collector that was indexed by view 10, which is incorrect. 
Since this vote collector has been pruned, which shouldn't, the valid votes along with the block are all dropped, the leader might not be able to collect enough votes for building QC.

The fix changes from indexing vote collector by block ID only to indexing by both blockID and view, such that votes with the same blockID but different views will be stored in different vote collectors.

### DIFF
--- a/consensus/hotstuff/voteaggregator/vote_collectors.go
+++ b/consensus/hotstuff/voteaggregator/vote_collectors.go
@@ -2,6 +2,7 @@
 package voteaggregator
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/onflow/flow-go/consensus/hotstuff"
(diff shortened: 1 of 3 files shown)


consensus/hotstuff/votecollector/statemachine.go
```diff
@@ -28,6 +28,8 @@ type StateMachine struct {
 	createVerifyingCollector NewVerifyingCollectorFactoryMethod
 }
 
+var _ hotstuff.VoteCollector = &StateMachine{}
```
zhangchiqing, line 31: ensure StateMachine has implemented VoteCollector interface

consensus/hotstuff/voteaggregator/vote_collectors.go
```diff
@@ -22,15 +35,14 @@ func (v *voteCollectors) GetOrCreateCollector(view uint64, blockID flow.Identifi
...
 // PruneUpToView prunes all collectors which target same view
 func (v *voteCollectors) PruneUpToView(view uint64) error {
-	v.lock.Lock()
-	defer v.lock.Unlock()
-	for blockID := range v.blockIDsByView[view] {
-		delete(v.collectors, blockID)
-	}
-	delete(v.blockIDsByView, view)
-	return nil
+	panic("implement me")
```
durkmurder, line 38: All good except this one, pruning logic is good, we shouldn't remove it.

durkmurder (approved):
Looks good, left one comment.

consensus/hotstuff/voteaggregator/vote_collectors.go
zhangchiqing: Wouldn't we [implement a similar logic like this](https://github.com/onflow/flow-go/blob/feature/consensus-voting/module/forest/leveled_forrest.go#L67-L92)?

durkmurder: Wasn't expecting that we will skip views, but sure, more general approach works as well

durkmurder:
This needs to be revisited after new details from @huitseeker. https://github.com/dapperlabs/flow-go/issues/5740#issuecomment-904670491

PR #1162, opened 2021-08-18
[DKG] improve logging

tarakby:
Improve logging for DKG broadcasted messages.


module/dkg/broker.go
```diff
@@ -98,7 +98,9 @@ func (b *Broker) Broadcast(data []byte) {
 	if b.broadcasts > 0 {
 		// The Warn log is used by the integration tests to check if this method
 		// is called more than once within one epoch.
-		b.log.Warn().Msgf("DKG broadcast number %d", b.broadcasts+1)
+		b.log.Warn().Msgf("DKG broadcast number %d with header %d", b.broadcasts+1, data[0])
+	} else {
+		b.log.Info().Msgf("DKG message broadcast with header %d", data[0])
```
huitseeker, line 103: Aren't we rather indicating that we're entering a DKG message broadcast _phase_?

jordanschalm (approved):

module/dkg/broker.go
tarakby: There are already logs for finishing a DKG phase and entering a new one. Here I'm adding a more granular log for every broadcast. It helps me understand what is happening in unhappy paths.

huitseeker: OK, maybe "DKG initial message broadcast" makes sense in the message then?

huitseeker (approved):
LGTM

module/dkg/broker.go
tarakby: The log here is made for every call to broadcast, and not necessarily for the initial broadcast. This includes the initial message, the complaint(s) and their answer(s).

PR #1163, opened 2021-08-18
[network] Check the OriginID of a libp2p message corresponds to its authenticated source

### LINKED ISSUE #1115: [Network] Verify originID against libp2p ID
Today, the network layer does not verify the [`OriginID`](https://github.com/onflow/flow-go/blob/master/network/message/message.pb.go#L29) of a message, and it passes it directly to the engines.

There are many reasons that this is bad (opens the door to impersonation attacks), and now with unstaked nodes it is even more important to address this.

* The verification for staked nodes can be done by checking the `OriginID` against the [`From`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.pb.go#L144) field of the [raw message we get from libp2p](https://github.com/onflow/flow-go/blob/master/network/p2p/readSubscription.go#L58). Given a Flow ID, we should be able to deduce the libp2p peer ID from the protocol state.
* For unstaked nodes, we will do the validation by comparing the given `OriginID` against the result of whatever deterministic libp2p ID -> Flow ID mapping we end up deciding on. 
* For staked AN's, we will validate messages from the unstaked network inside the libp2p topic validator, but using the same method as above.

As a result, we may need to do some refactoring  so that the network layer of the code is aware of whether it is a staked or unstaked network, so it can choose which verification logic to use from above. Alternatively, we can implement some message validators that do this validation.

We can write a test to ensure that this is all working:
* Start up two nodes, and from one of them send a message with an incorrect origin ID
* Check that this message is discarded by the receiving node.

huitseeker:
Contributes to #1115.
Follow-up of #1129.

Edit: there was no problem with the integration tests, 🤦 

However, TestMiddlewareTestSuit/TestUnsubscribe is still flaky based on the delay in the last step (sending the message once a mesh has formed). But it is problematic on master, see [[1]](https://github.com/onflow/flow-go/runs/3357056741) [[2]](https://github.com/onflow/flow-go/runs/3356508175), so this is not relevant to this PR per se.

### DIFF
--- a/network/p2p/middleware.go
+++ b/network/p2p/middleware.go
@@ -317,7 +317,7 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
 	log.Info().Msg("incoming stream received")
 
 	//create a new readConnection with the context of the middleware
-	conn := newReadConnection(m.ctx, s, m.processMessage, log, m.metrics, LargeMsgMaxUnicastMsgSize)
+	conn := newReadConnection(m.ctx, s, m.processAuthenticatedMessage, log, m.metrics, LargeMsgMaxUnicastMsgSize)
 
 	// kick off the receive loop to continuously receive messages
 	m.wg.Add(1)
@@ -362,6 +362,9 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
 }
 
 // processAuthenticatedMessage processes a message and a source (indicated by its PublicKey) and eventually passes it to the overlay
+// In particular, it checks the claim of protocol authorship situated in the message against `originKey`
+// The assumption is that the message has been authenticated at the network level (libp2p) to origin at the network public key `originKey`
+// this requirement is fulfilled by e.g. the output of readConnection and readSubscription
 func (m *Middleware) processAuthenticatedMessage(msg *message.Message, originKey crypto.PublicKey) {
 	identities, err := m.ov.Identity()
 	if err != nil {
--- a/network/p2p/middleware.go
+++ b/network/p2p/middleware.go
@@ -13,6 +13,7 @@ import (
 	libp2pnetwork "github.com/libp2p/go-libp2p-core/network"
 	"github.com/rs/zerolog"
 
+	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/engine"
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/module"
@@ -334,7 +335,7 @@ func (m *Middleware) Subscribe(channel network.Channel) error {
 	}
 
 	// create a new readSubscription with the context of the middleware
-	rs := newReadSubscription(m.ctx, s, m.processMessage, m.log, m.metrics)
+	rs := newReadSubscription(m.ctx, s, m.processAuthenticatedMessage, m.log, m.metrics)
 	m.wg.Add(1)
 
 	// kick off the receive loop to continuously receive messages
@@ -358,6 +359,32 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
 	m.peerManagerUpdate()
 
 	return nil
+}
+
+// processAuthenticatedMessage processes a message and a source (indicated by its PublicKey) and eventually passes it to the overlay
+func (m *Middleware) processAuthenticatedMessage(msg *message.Message, originKey crypto.PublicKey) {
+	identities, err := m.ov.Identity()
+	if err != nil {
+		m.log.Error().Err(err).Msg("failed to retrieve identities list while delivering a message")
+		return
+	}
+
+	// check the origin of the message corresponds to the one claimed in the OriginID
+	originID := flow.HashToID(msg.OriginID)
+
+	originIdentity, found := identities[originID]
+	if !found {
+		m.log.Warn().Msgf("received message with claimed originID %x, which is not known by this node, and was dropped", originID)
+		return
+	} else if originIdentity.NetworkPubKey == nil {
+		m.log.Warn().Msgf("received message with claimed originID %x, wich has no network identifiers at this node, and was dropped", originID)
+		return
+	} else if !originIdentity.NetworkPubKey.Equals(originKey) {
+		m.log.Warn().Msgf("received message claiming to be from nodeID %v with key %x was actually signed by %x and dropped", originID, originIdentity.NetworkPubKey, originKey)
+		return
+	}
+
+	m.processMessage(msg)
 }
 
 // processMessage processes a message and eventually passes it to the overlay
--- a/network/p2p/readSubscription.go
+++ b/network/p2p/readSubscription.go
@@ -2,12 +2,17 @@ package p2p
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 
+	lcrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
+
 	"github.com/rs/zerolog"
 
+	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/module"
 	"github.com/onflow/flow-go/network/message"
 	_ "github.com/onflow/flow-go/utils/binstat"
@@ -20,13 +25,13 @@ type readSubscription struct {
 	log      zerolog.Logger
 	sub      *pubsub.Subscription
 	metrics  module.NetworkMetrics
-	callback func(msg *message.Message)
+	callback func(msg *message.Message, pubKey crypto.PublicKey)
 }
 
 // newReadSubscription reads the messages coming in on the subscription
 func newReadSubscription(ctx context.Context,
 	sub *pubsub.Subscription,
-	callback func(msg *message.Message),
+	callback func(msg *message.Message, pubKey crypto.PublicKey),
 	log zerolog.Logger,
 	metrics module.NetworkMetrics) *readSubscription {
 
@@ -77,6 +82,19 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
 			return
 		}
 
+		// if pubsub.WithMessageSigning(true) and pubsub.WithStrictSignatureVerification(true),
+		// the emitter is authenticated
+		emitterKey, err := messagePubKey(rawMsg)
+		if err != nil {
+			r.log.Err(err).Msg("failed to extract libp2p public key of message")
+			return
+		}
+		flowKey, err := FlowPublicKeyFromLibP2P(emitterKey)
+		if err != nil {
+			r.log.Err(err).Msg("failed to extract flow public key of libp2p key")
+			return
+		}
+
 		var msg message.Message
 		// convert the incoming raw message payload to Message type
 		//bs := binstat.EnterTimeVal(binstat.BinNet+":wire>1protobuf2message", int64(len(rawMsg.Data)))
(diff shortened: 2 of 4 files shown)


network/p2p/middleware.go
```diff
@@ -360,8 +361,28 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
...
+	// check the origin of the message corresponds to the one claimed in the OriginID
+	originID := flow.HashToID(msg.OriginID)
+
+	originIdentity, found := identities[originID]
+	if !found || !originIdentity.NetworkPubKey.Equals(originKey) {
+		m.log.Warn().Msgf("message claiming to be from nodeID %v with key %x was actually signed by %x and dropped", originID, originIdentity.NetworkPubKey, originKey)
+		return
+	}
+
+	m.processUnauthenticatedMessage(msg)
```
zhangchiqing, line 381: The `msg` is passed to a method that processes "Unauthenticated" message, as if the message was "unauthenticated" message, but actually the message's origin has been verified already.

I found this is a bit confusing

zhangchiqing (approved):
Looks good, just a comment

network/p2p/readSubscription.go
```diff
@@ -87,10 +92,56 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
 			return
 		}
 
+		// if pubsub.WithMessageSigning(true) and pubsub.WithStrictSignatureVerification(true),
+		// the emitter is authenticated
+		emitterKey, err := messagePubKey(rawMsg)
+		if err != nil {
+			r.log.Err(err).Msg("failed to extract libp2p public key of message")
```
vishalchangrani, line 99:
```suggestion
			r.log.Err(err).Msg("failed to extract libp2p public key of message")
			return
```

network/p2p/readSubscription.go
```diff
@@ -87,10 +92,56 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
...
+		}
+
+		flowKey, err := FlowPublicKeyFromLibP2P(emitterKey)
+		if err != nil {
+			r.log.Err(err).Msg("failed to extract flow public key of libp2p key")
+			return
+		}
+
```
vishalchangrani, line 107: I would also move the [message unmarshalling](https://github.com/onflow/flow-go/pull/1163/files#diff-b25dcc55780477ea9266ec07bf26a1b6c2f10dc6143aed7f60da81b0d861cccbR85-R93) which is a more heavier operation here after the key extraction since why bother to unmarshal if we can't extract the key.

network/p2p/middleware.go
smnzhu: I agree, perhaps we could keep the naming of `processMessage` for the old method, and name the new method `authenticateAndProcessMessage` or something.

network/p2p/readSubscription.go
```diff
@@ -87,10 +92,56 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
...
-		r.callback(&msg)
+		r.callback(&msg, flowKey)
+	}
+}
+
+// This reproduces the exact logic of the private https://github.com/libp2p/go-libp2p-pubsub/blob/ba28f8ecfc551d4d916beb748d3384951bce3ed0/sign.go#L77
+func messagePubKey(m *pubsub.Message) (lcrypto.PubKey, error) {
+	var pubk lcrypto.PubKey
+
+	pid, err := peer.IDFromBytes(m.From)
```
vishalchangrani, line 120:
```suggestion
        // m.From is the original sender of the message (versus `m.ReceivedFrom` which is the last hop which sent us this message)
	pid, err := peer.IDFromBytes(m.From)
```

vishalchangrani: just for clarification

vishalchangrani (approved):
left some comments, otherwise LGTM!

network/p2p/middleware.go
```diff
@@ -360,8 +361,28 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
...
+		m.log.Error().Err(err).Msg("failed to retrieve identities list while delivering a message")
+		return
+	}
+
+	// check the origin of the message corresponds to the one claimed in the OriginID
+	originID := flow.HashToID(msg.OriginID)
+
+	originIdentity, found := identities[originID]
+	if !found || !originIdentity.NetworkPubKey.Equals(originKey) {
+		m.log.Warn().Msgf("message claiming to be from nodeID %v with key %x was actually signed by %x and dropped", originID, originIdentity.NetworkPubKey, originKey)
```
smnzhu, line 377: Note, we may want to distinguish the error messages for the two cases above.

In particular, if `!found` is true, in the worst case you may get a segfault by trying to do `originIdentity.NetworkPubKey` on this line.

In the best case, `originIdentity.NetworkPubKey` is probably null and the error log is confusing for the reader

network/p2p/readSubscription.go
```diff
@@ -87,10 +92,56 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
...
 		// log metrics
 		r.metrics.NetworkMessageReceived(msg.Size(), msg.ChannelID, msg.Type)
 
 		// call the callback
-		r.callback(&msg)
+		r.callback(&msg, flowKey)
+	}
+}
+
+// This reproduces the exact logic of the private https://github.com/libp2p/go-libp2p-pubsub/blob/ba28f8ecfc551d4d916beb748d3384951bce3ed0/sign.go#L77
```
smnzhu, line 116: Can we also include a brief summary of what it actually does?

network/test/middleware_test.go
```diff
@@ -268,6 +268,45 @@ func (m *MiddlewareTestSuite) TestEcho() {
...
+	}
+
+	// set up waiting for m.size pubsub tags indicating a mesh has formed
+	for i := 0; i < m.size; i++ {
+		select {
+		case <-m.obs:
+		case <-time.After(2 * time.Second):
+			assert.FailNow(m.T(), "could not receive pubsub tag indicating mesh formed")
+		}
+	}
```
smnzhu, line 293: Could you explain when a pubsub tag is emmitted? Thanks

smnzhu: Andy why getting `m.size` of them indicates that a mesh has formed?

network/test/middleware_test.go
```diff
@@ -268,6 +268,45 @@ func (m *MiddlewareTestSuite) TestEcho() {
...
+
+	var spoofedID flow.Identifier
+	copy(spoofedID[:16], firstNode[:16])
+	copy(spoofedID[16:], lastNode[16:])
```
smnzhu, line 297: Are you just trying to create an invalid identifier here, or is there some reason you specifically take the first and last half of the identifiers of the first and last nodes?

If just trying to create an invalid identifier, you can just use `unittest.IdentifierFixture`

Also, I think we should test two separate cases here:
* the spoofed ID is actually a valid ID, but not the one that belongs to the sender
* the spoofed ID is an invalid ID that doesn't belong to anyone

network/test/middleware_test.go
```diff
@@ -268,6 +268,45 @@ func (m *MiddlewareTestSuite) TestEcho() {
...
+
+	message1 := createMessage(spoofedID, lastNode, "hello1")
+
+	err := m.mws[first].Publish(message1, testChannel)
+	assert.NoError(m.T(), err)
+
+	// assert that the spoofed message is not received by the target node
+	assert.Never(m.T(), func() bool {
+		return !m.ov[last].AssertNumberOfCalls(m.T(), "Receive", 0)
+	}, 2*time.Second, time.Millisecond)
```
smnzhu, line 307: nit: probably don't need to check every Millisecond? every 100 ms probably good enough.

network/p2p/readSubscription.go
huitseeker: Given that we have `pubsub.WithMessageSigning(true)` and `pubsub.WithStrictSignatureVerification(true)`, I would note such a message would not arrive here, but in theory, agreed.

network/p2p/middleware.go
huitseeker: I don't think the segfault is possible given `||` shortcuts: https://play.golang.org/p/g7CGIVG7dQ2
Moreover, given that we have `pubsub.WithMessageSigning(true)` and `pubsub.WithStrictSignatureVerification(true)`, I would note a messsage for which `found != true` would likely not arrive here but in theory, agreed.

network/test/middleware_test.go
huitseeker: Libp2p emits `Protect` calls with tag arguments marked by their pubsub topic when establishing each peering in the mesh:
https://github.com/libp2p/go-libp2p-pubsub/blob/master/tag_tracer.go

You would usually wait for [`libp2p.GossipSubD`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L33) such links, but here there's only two nodes.

huitseeker:
bors merge

huitseeker:
bors merge

network/p2p/middleware.go
```diff
@@ -360,6 +361,32 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
 	return nil
 }
 
+// processAuthenticatedMessage processes a message and a source (indicated by its PublicKey) and eventually passes it to the overlay
+func (m *Middleware) processAuthenticatedMessage(msg *message.Message, originKey crypto.PublicKey) {
```
zhangchiqing, line 365: I found still a bit confusing. 

It sounds like the message has been authenticated, but actually we are going to check the authentication, right?

If so, would `checkOriginAndProcess` be better?

huitseeker: > It sounds like the message has been authenticated,

It has, the message has a clear origin, which is the signer of its envelope, which signature has been checked.

> we are going to check the authentication

No, we are going to check the protocol-level authorship claim against the prior authentication.

huitseeker:
bors retry

network/p2p/middleware.go
zhangchiqing: Thanks, I see. would be great to add these knowledges into comments

network/p2p/middleware.go
smnzhu: I meant that even in the log message itself, you are calling `originIdentity.NetworkPubkey`

network/test/middleware_test.go
smnzhu: I'd like to understand this a bit more.

From what I understand, for each connection *both* peers would get a `Protect` call for each peering, right?

So if there are 5 peers in total, then we should expect to see 8 tags before we can be confident that a mesh has formed, correct? (5 peers, 5 - 1 = 4 edges required for a connected graph, 4 * 2 = 8 tags).

Here, you are only checking for 5 tags. 

Maybe I'm not understanding how this works?

smnzhu: Also, are we sure that `Protect` is only called once for each peer when establishing a peering? In otherwise, is counting the number of tags enough, or would it make sense to actually keep track of the peer IDs in case we get multiple tags for the same peer?

huitseeker: So [here's where the calls to `Protect` are intercepted](https://github.com/onflow/flow-go/blob/73d6864593dfc899cd36ef178fcce7f0fad511d6/network/test/testUtil.go#L45-L90). 

[This](https://github.com/libp2p/go-libp2p-pubsub/blob/65f48fb2c8b37563525e3d44042ea2418650f854/trace.go#L468-L493) is where they originate at the libp2p level, in reaction to a join.

I don't think you're incorrect in your analysis. However:

- when I said "You would usually wait for libp2p.GossipSubD such links" I misremembered, we actually wait for  `pubsub.GossipSubD*count` [in tests with count = 10](https://github.com/onflow/flow-go/blob/master/network/test/meshengine_test.go#L169)
- the way I got to that number is by running the test many times at a log level that allowed me to witness the number of tags emitted, 
- indeed, the value there is meant as a fast & loose heuristic to make sure we wait until after the `flow/networking` level that triggers the `UpdatePeers` and then the `libp2p` mesh update downstream of that, NOT as a countdown to full mesh,
- it was nowhere near the theoretical number we would obtain: if, when a node joins a topic, we get 2m graft messages when there are k nodes already on it (m = k if k < D, m = D otherwise, since we don't count pruning), then for 10 nodes we'd expect 90 messages, and we only witness around 65-ish,
- we don't need to finish mesh formation to ensure message delivery, as you noticed we only need a connected graph.

I'm happy to update to a value you'd find more suitable, but for count = 2 (in the test we're commenting on), I note we have, practically and theoretically 2 graft messages, every time.

> are we sure that Protect is only called once for each peer when establishing a peering? 

Yep, the Protect calls otherwise made by our engines are all tagless.

smnzhu: Thanks for the explanation :)

smnzhu (approved):

PR #1164, opened 2021-08-18
[Consensus Voting] Sig aggregator implementation

durkmurder:
[flow-go/issues/5744](https://github.com/dapperlabs/flow-go/issues/5744)

### Context

This PR implements simple signature aggregator. This aggregator will be used in `CombinedAggregator` to track staking and threshold signatures. Since we have decided that we will use `CombinedAggregator` all weight tracking as well as concurrency safety guarantees will be implemented there. 

Next step will be to integrate this aggregator into `CombinedAggregator`.


consensus/hotstuff/signature/signature_aggregator.go
```diff
@@ -1,33 +1,57 @@
...
-func (s *SignatureAggregatorImpl) TrustedAdd(signerID flow.Identifier, sig crypto.Signature) (totalWeight uint64, exception error) {
-	panic("to be implemented")
+func (s *SignatureAggregatorImpl) TrustedAdd(signerID flow.Identifier, weight uint64, sig crypto.Signature) (totalWeight uint64, exception error) {
+	if hasSigned := s.sigCollector.HasSigned(signerID); hasSigned {
+		return 0, fmt.Errorf("signature for (%x) was already collected", signerID)
+	}
+
+	s.sigCollector.Add(signerID, sig)
+	s.totalWeight += weight
+	return s.totalWeight, nil
```
zhangchiqing, line 44: The error is for exceptions, but this case is not an exception. 

Please refer to [this comment](https://github.com/onflow/flow-go/pull/1164/files#diff-faa591ccd4ac38232acffa4b2fcf61532c45e459d3a99b2e203a2489269f7a4cR36)

It's better for `SigCollector.Add` to return an `added bool`, depending on the result we can add weight or not.
```suggestion
	added := s.sigCollector.Add(signerID, sig)
	if added {
		s.totalWeight += weight
	}
	return s.totalWeight, nil
```

consensus/hotstuff/signature/signature_aggregator.go
```diff
@@ -1,33 +1,57 @@
...
+}
+
 // TrustedAdd adds an already verified signature, and look up the weight for the given signer,
 // and add it to the total weight, and returns the total weight that have been collected.
 // return (1000, nil) means the signature has been added, and 1000 weight has been collected in total.
 // return (1000, nil) means the signature is a duplication and 1000 weight has been collected in total.
-func (s *SignatureAggregatorImpl) TrustedAdd(signerID flow.Identifier, sig crypto.Signature) (totalWeight uint64, exception error) {
-	panic("to be implemented")
+func (s *SignatureAggregatorImpl) TrustedAdd(signerID flow.Identifier, weight uint64, sig crypto.Signature) (totalWeight uint64, exception error) {
+	if hasSigned := s.sigCollector.HasSigned(signerID); hasSigned {
```
jordanschalm:
```suggestion
	if s.sigCollector.HasSigned(signerID) {
```
Minor thing, but I feel like the named variable doesn't really add more context, since it has the same name as the method.

consensus/hotstuff/signature/signature_aggregator.go
```diff
@@ -1,33 +1,57 @@
...
+		return 0, fmt.Errorf("signature for (%x) was already collected", signerID)
```
jordanschalm:
```suggestion
		return 0, fmt.Errorf("signature for signer (signer_id=%x) was already collected", signerID)
```
To make it obvious that the ID is referring to a node ID, rather than a block etc.

consensus/hotstuff/signature/signature_aggregator_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	mock2 "github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/module/mock"
```
jordanschalm, line 11:
```suggestion
	testifymock "github.com/stretchr/testify/mock"
	"github.com/stretchr/testify/require"

	"github.com/onflow/flow-go/crypto"
	modulemock "github.com/onflow/flow-go/module/mock"
```
Clearer aliases than the default for the conflicting import names

consensus/hotstuff/signature/signature_aggregator.go
```diff
@@ -1,33 +1,57 @@
...
 // TotalWeight returns the total weight presented by the collected sig shares.
 func (s *SignatureAggregatorImpl) TotalWeight() uint64 {
-	panic("to be implemented")
+	return s.totalWeight
 }
 
 // Aggregate assumes enough shares have been collected, it aggregates the signatures
 // and return the aggregated signature.
-// if called concurrently, only one thread will be running the aggregation.
 func (s *SignatureAggregatorImpl) Aggregate() ([]byte, error) {
```
zhangchiqing, line 54: I think we should return `flow.AggregatedSignature, error` instead of `[]byte, error`, because the [signerIDs will be needed.](https://github.com/onflow/flow-go/blob/feature/consensus-voting/consensus/hotstuff/verification/combined_signer.go#L176)

jordanschalm (approved):
LGTM

consensus/hotstuff/signature/signature_aggregator.go
zhangchiqing: I would suggest to let SignatureAggregator track signers, because `SignatureAggregator` handles duplication, and easier to ensure that the aggregated sigs and signerIDs are consistent.

consensus/hotstuff/signature/signature_aggregator.go
durkmurder: Don't even know why I did this :)

consensus/hotstuff/signature/signature_aggregator.go
durkmurder: removed the error completely

consensus/hotstuff/signature/signature_aggregator.go
durkmurder: it's really easy to track `signerIDs` if we do this:
```
added, _ := aggregator.TrustedAdd(...)
if added {
     store in signerIDs set
}
```

vs 

```
signers1, _ := stakingAggregator.Aggregate()
signers2, _ := threholsAggregator.Aggregate()

build unique set of signer1 and signer2

```

durkmurder:
Closing for know as we aren't sure how exactly BLS signature should be aggregated. To be continued by @tarakby

PR #1165, opened 2021-08-18
[network] Validate Unstaked ID translation

huitseeker:
This shows we need to use Secp256k1 on the bootstrap key generation for unstaked Access nodes (and positive keys at that), which I'll add in a further PR. 

Unit-tests and validates the ID translation for unstaked network.

### DIFF
--- a/network/p2p/unstaked_translator.go
+++ b/network/p2p/unstaked_translator.go
@@ -1,6 +1,8 @@
 package p2p
 
 import (
+	"fmt"
+
 	crypto_pb "github.com/libp2p/go-libp2p-core/crypto/pb"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/multiformats/go-multihash"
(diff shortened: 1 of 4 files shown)


network/p2p/unstaked_translator.go
```diff
@@ -27,16 +29,16 @@ func (t *UnstakedNetworkIDTranslator) GetPeerID(flowID flow.Identifier) (peer.ID
...
 	pk, err := peerID.ExtractPublicKey()
 	if err != nil {
-		// return error
+		return flow.ZeroID, fmt.Errorf("Cannot generate an unstaked FlowID for peerID %v: corresponding libp2p key is not extractible from PeerID", peerID)
 	}
 
-	if pk.Type() != crypto_pb.KeyType_ECDSA {
-		// fail
+	if pk.Type() != crypto_pb.KeyType_Secp256k1 {
+		return flow.ZeroID, fmt.Errorf("Cannot generate an unstaked FlowID for peerID %v: corresponding libp2p key is not a Secp256K1 key", peerID)
```
m4ksio, line 36: I think we have mapping from key to its canonical string name

network/p2p/unstaked_translator_test.go
```diff
@@ -0,0 +1,98 @@
...
+		pk, err := pID.ExtractPublicKey()
+		require.NoError(t, err)
+
+		// for a secp256k1 key, this is compressed representation preceded by 00 bits
+		// indicating the multihash.IDENTITY
+		pkBytes, err := pk.Raw()
+		require.NoError(t, err)
+
+		// key is positive, roundtrip should be possible
+		if pkBytes[0] == 0x02 {
```
smnzhu, line 59: Can we enforce that this is actually run a certain number of times, instead of just running the loop 50 times in total and potentially (though hopefully not likely) never getting one with 0x02?

What I'm actually worried about is whether byte [0] is actually the correct byte to check here, or whether `pk.Raw()` actually returns some more bytes in the beginning.

For example, above you note

> for a secp256k1 key, this is compressed representation preceded by 00 bits indicating the multihash.IDENTITY

So would it be the case that `pkBytes[0] == 0x00` all the time? If that were true, this test would always pass because the if statement always fails.

This is something I was meaning to check and hence why I had it as a TODO above, but since the TODO is now being removed I'd like to make sure of it here.

huitseeker: > Can we enforce that this is actually run a certain number of times, instead of just running the loop 50 times in total and potentially (though hopefully not likely) never getting one with 0x02?

Sure! Though the probability is 50%. I'll add the change.

> So would it be the case that pkBytes[0] == 0x00 all the time? 

I wrote bits, not byte.

smnzhu: Thanks for checking :)

PR #1167, opened 2021-08-19
[FLIP] Context propagation and network layer API changes

smnzhu:
https://github.com/onflow/flow-go/blob/smnzhu/runnable-flip/flips/context-propagation.md

### DIFF
--- a/flips/context-propagation.md
+++ b/flips/context-propagation.md
@@ -1,11 +1,5 @@
 # Context Propagation and Network Layer Refactor (Core Protocol)
 
-| Status        | Proposed                                       |
-:-------------- |:---------------------------------------------- |
-| **FLIP #**    | [xxx](https://github.com/onflow/flow/pull/xxx) |
-| **Author(s)** | Simon Zhu (simon.zhu@dapperlabs.com)           |
-| **Sponsor**   | Simon Zhu (simon.zhu@dapperlabs.com)           |
-| **Updated**   | 2021-xx-xx                                     |
 
 
 ## Objective
--- a/flips/context-propagation.md
+++ b/flips/context-propagation.md
@@ -22,14 +22,15 @@ The [`ReadyDoneAware`](https://github.com/onflow/flow-go/blob/7763000ba5724bb03f
 When the network layer receives a message, it will pass the message to the [`Engine`](https://github.com/onflow/flow-go/blob/7763000ba5724bb03f522380e513b784b4597d46/network/engine.go) registered on
 the corresponding channel by [calling the engine's `Process` method](https://github.com/onflow/flow-go/blob/d31fd63eb651ed9faf0f677e9934baef6c4d9792/network/p2p/network.go#L406), passing it the Flow ID of the message sender.
 
-#### Potential problems 
-
+### Potential problems 
+#### Non-idempotency of `ReadyDoneAware`
 The current `ReadyDoneAware` interface is misleading, as by the name one might expect that it is only used to check the state of a component. However, in almost all current implementations the `Ready` method is used to both start the component *and* check when it has started up, and similarly for the `Done` method. 
 
-This introduces issues of concurrency safety, as most implementations do not properly handle the case where the `Ready` or `Done` methods are called multiple times. See [this example](https://github.com/onflow/flow-go/pull/1026).
+This introduces issues of concurrency safety, as most implementations do not properly handle the case where the `Ready` or `Done` methods are called more than once. See [this example](https://github.com/onflow/flow-go/pull/1026).
 
 [Clearer documentation](https://github.com/onflow/flow-go/pull/1032) and a new [`LifecycleManager`](https://github.com/onflow/flow-go/pull/1031) component were introduced as a step towards fixing this by providing concurrency-safety for components implementing `ReadyDoneAware`, but this still does not provide a clear separation between the ability to start / stop a component and the ability to check its state. A component usually only needs to be started once, whereas multiple other components may wish to check its state.
 
+#### Tight coupling of network layer to Flow concepts
 The current network layer API was designed with the assumption that all messages sent and received either target or originate from staked Flow nodes. This is why an engine's [`Process`](https://github.com/onflow/flow-go/blob/master/network/engine.go#L28) method accepts a Flow ID identifying the message sender, and outgoing messages [must specify Flow ID(s)](https://github.com/onflow/flow-go/blob/master/network/conduit.go#L62) as targets.
 
 This assumption is no longer true today. The access node, for example, may communicate with multiple (unstaked) consensus followers. It's perceivable that in the future there will be even more cases where communication with unstaked parties may happen (for example, execution nodes talking to DPS).
@@ -39,6 +40,8 @@ Furthermore, many engines actually don't need to know the `OriginID` of a messag
 Finally, as of today the network layer does not actually verify the [`OriginID`](https://github.com/onflow/flow-go/blob/2641774451dcc1e147843a74265afb58466b50ab/network/message/message.pb.go#L29) field of messages received from the network. This opens the door to impersonation attacks, as a node can send a message pretending to be any other node simply by putting the other node's `OriginID` field inside the message they send.
 
 ## Design Proposal
+
+### Separate lifecycle management from state observation
 
 Moving forward, we should split the `ReadyDoneAware` interface into two separate interfaces:
 - A `Runnable` interface which can be used to start and stop a component:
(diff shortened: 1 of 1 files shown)


smnzhu:
This touches on two somewhat distinct issues, but they are related at a high level so I've included them together for now to give the bigger picture.

They can be separated later on into two FLIPs.

flips/context-propagation.md
```diff
@@ -0,0 +1,208 @@
...
+  - If context propagation is done properly, there is no need to worry about any cleanup code in the `Done` method.
+- This allows us to separate the capability to check a component's state from the capability to start / stop it. We may want to give multiple other components the capability to check its state, without giving them the capability to start or stop it. Here is an [example](https://github.com/onflow/flow-go/blob/b50f0ffe054103a82e4aa9e0c9e4610c2cbf2cc9/engine/common/splitter/network/network.go#L112) of where this would be useful.
+- This provides a clearer way of defining ownership of components, and hence may potentially eliminate the need to deal with concurrency-safety altogether. Whoever creates a component should be responsible for starting it, and therefore they should be the only one with access to its `Runnable` interface. If each component only has a single parent that is capable of starting it, then we should never run into concurrency issues.
+- This enables us to define components which startup upon creation, and don't need to be explicitly `Run()` at a later point in time. Such components would only implement the `ReadyDoneAware` interface.
```
m4ksio, line 65: I probably miss something, but if component starts up immediately and has no startup/shutdown code, why even bother with `ReadyDoneAware`? It can just be simple Go object without special lifecycle right?
We use `.Module()` in `NodeBuilder`, which is also badly named, as this is just a callback, nothing modular there.

flips/context-propagation.md
```diff
@@ -0,0 +1,208 @@
...
+    // Done returns a channel that will close when component shutdown has completed.
+    Done() <-chan struct{}
+  }
+  ```
+
+A component will now be started by passing a `Context` to its `Run` method, and can be stopped by cancelling the `Context`. If a component needs to startup subcomponents, it can create child `Context`s from this `Context` and pass those to the subcomponents.
+#### Motivations
+- `Context`s are the standard way of doing go-routine lifecycle management in Go, and adhering to standards helps eliminate confusion and ambiguity for anyone interacting with the `flow-go` codebase. This is especially true now that we are beginning to provide API's and interfaces for third parties to interact with the codebase (e.g DPS).
+  - Even to someone unfamiliar with our codebase (but familiar with Go idioms), it is clear how a method signature like `Run(context.Context)` will behave. A method signature like `Ready()` is not so clear.
+  - If context propagation is done properly, there is no need to worry about any cleanup code in the `Done` method.
```
m4ksio, line 62: I understand the idea, but `ReadyDoneAware` was plugged because cancelling context doesn't provide a way for component to signalise cleanup. 
Its especially useful where components rely on being shutdown in order - for example. Ledger and WAL and Checkpointers.

m4ksio:
I have only reviewed proposed changes regarding `ReadyDoneAware`. While I agree we there are certain drawbacks I feel that proposed rework would increase implied complexity by introducing by pushing component readiness as a status all components need to be aware of.
There are places that would benefit from the proposed approach, but I think this can be fixed by adding extra checks/propagation on top of existing system and only apply them where needed.

flips/context-propagation.md
```diff
@@ -0,0 +1,208 @@
...
+
+When the network layer receives a message, it will pass the message to the [`Engine`](https://github.com/onflow/flow-go/blob/7763000ba5724bb03f522380e513b784b4597d46/network/engine.go) registered on
+the corresponding channel by [calling the engine's `Process` method](https://github.com/onflow/flow-go/blob/d31fd63eb651ed9faf0f677e9934baef6c4d9792/network/p2p/network.go#L406), passing it the Flow ID of the message sender.
+
+### Potential problems 
+#### Non-idempotency of `ReadyDoneAware`
+The current `ReadyDoneAware` interface is misleading, as by the name one might expect that it is only used to check the state of a component. However, in almost all current implementations the `Ready` method is used to both start the component *and* check when it has started up, and similarly for the `Done` method.
```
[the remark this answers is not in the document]
smnzhu: I think this poses a few difficulties.

First, if multiple components want to wait for the startup of another to complete (ie common dependency), then with this approach we would need to pass the channel returned from `Ready` to all of those components. This also means that we cannot construct those components until we've started the dependency, because we cannot get the ready channel until we call `Ready` to start it. Ideally, we don't want this restriction because it forces sequentialism of certain steps that could potentially run in parallel.

Now, what if these components also want to wait for the dependency's shutdown to complete later on? Then we have no way of doing this, because we only get a done channel when we call `Done` to shutdown the dependency. But these components have already been created by then, meaning they certainly don't have this done channel, and any mechanism we come up with of injecting it after wards is probably gonna be ugly.

flips/context-propagation.md
```diff
@@ -0,0 +1,208 @@
...
+
+When the network layer receives a message, it will pass the message to the [`Engine`](https://github.com/onflow/flow-go/blob/7763000ba5724bb03f522380e513b784b4597d46/network/engine.go) registered on
+the corresponding channel by [calling the engine's `Process` method](https://github.com/onflow/flow-go/blob/d31fd63eb651ed9faf0f677e9934baef6c4d9792/network/p2p/network.go#L406), passing it the Flow ID of the message sender.
+
+### Potential problems 
+#### Non-idempotency of `ReadyDoneAware`
+The current `ReadyDoneAware` interface is misleading, as by the name one might expect that it is only used to check the state of a component. However, in almost all current implementations the `Ready` method is used to both start the component *and* check when it has started up, and similarly for the `Done` method. 
+
+This introduces issues of concurrency safety, as most implementations do not properly handle the case where the `Ready` or `Done` methods are called more than once. See [this example](https://github.com/onflow/flow-go/pull/1026).
```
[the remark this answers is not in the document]
smnzhu: My point here is more about the semantics of `ReadyDoneAware`.

You are correct that the owner of a component should manager the lifecycle. The problem here is that we have not separated lifecycle management from state observation, because `ReadyDoneAware` currently implements both.

While only one owner should be able to start a component, multiple other components may want to wait for it to be ready. This is where it becomes a problem if calling `Ready` also starts the component each time.

flips/context-propagation.md
```diff
@@ -0,0 +1,208 @@
...
+## Proposal
+
+### Separate lifecycle management from state observation
+
+Moving forward, we should split the `ReadyDoneAware` interface into two separate interfaces:
+- A `Runnable` interface which can be used to start and stop a component:
+  ```golang
+  type Runnable interface {
+    // Run will start the component
+    Run(context.Context)
```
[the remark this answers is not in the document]
smnzhu: Sure, I'm not tied to the name `Runnable` haha 😛.

> Also, does this function returns immediately or after the component has finished initialisation?

Let me add a better comment to answer this

> I see no added benefit over using Ready() (with this or other name). Returned channel gives great flexibility, including options to build notification or multiple waits system,

The benefits are listed below. We are not getting rid of a returned channel! The channel will just not be returned from this particular method, but from the `Ready` method (below).

flips/context-propagation.md
smnzhu: So this would refer to a component which *does* have startup / shutdown code, but the startup code isn't explicitly triggered by a call to `Run` (or `Start`, which is probably a better name as you suggested).

So instead of this:
```golang

// create component
component := NewComponent()

// start component
component.Start(ctx)

// wait for it to be ready
<-component.Ready()

// stop component
cancel()

// wait for it to be done
<-component.Done()
```

We *could* have something like this

```golang

// create and start component
component := NewComponent(ctx)

// wait for it to be ready
<-component.Ready()

// stop component
cancel()

// wait for it to be done
<-component.Done()
```

flips/context-propagation.md
smnzhu: > I understand the idea, but ReadyDoneAware was plugged because cancelling context doesn't provide a way for component to signalise cleanup.

Yes, and this is precisely why I'm not suggesting we get rid of `ReadyDoneAware`. We should keep it, as a way for components to signal that they have finished startup for finished shutdown / cleanup, but it should *not* also be responsible for actually starting or stopping the component.

smnzhu:
> I have only reviewed proposed changes regarding `ReadyDoneAware`. While I agree we there are certain drawbacks I feel that proposed rework would increase implied complexity by introducing by pushing component readiness as a status all components need to be aware of.
> There are places that would benefit from the proposed approach, but I think this can be fixed by adding extra checks/propagation on top of existing system and only apply them where needed.

@m4ksio Thanks for the review :)

I've answered all of your comments above and I hope I've addressed them clearly, but happy to discuss more.

> I feel that proposed rework would increase implied complexity by introducing by pushing component readiness as a status all components need to be aware of.

This sentence is a little hard to understand for me, can you explain?

flips/context-propagation.md
```diff
@@ -0,0 +1,210 @@
...
+    return f
+  }
+
+  func (f *FooComponent) childRoutine(ctx context.Context) {
+    for {
+      select {
+      case <-ctx.Done():
+        return
+      default:
+        // do work...
```
yhassanzadeh13, line 193: Not sure if this is a reliable way of life-cycle management for Flow. I acknowledge the Golang best practices around contexts that you also elaborated. The problem is with this approach we keep delegating the life-cycle management of sub-components to themselves, which is in contrast to managing the life-cycle by the parent component. 

For example, assume the `// do work` part here is a long-running task. Canceling the context wouldn't kill that task until it returns and the subsequent iteration of `for` starts. However, the parent component called `ctx.Cancel()` assumes all sub-components have been terminated immediately. Right?

The problem is a faulty component that faced deadlock may never get terminated in this approach, and there is no way that the parent component is aware of it.

flips/context-propagation.md
```diff
@@ -0,0 +1,210 @@
...
+	Publish(event interface{}, targetIDs ...peer.ID) error
+	Unicast(event interface{}, targetID peer.ID) error
+	Multicast(event interface{}, num uint, targetIDs ...peer.ID) error
+}
+```
+
+Finally, we should either deprecate or modify the behavior of [`engine.Unit`](https://github.com/onflow/flow-go/blob/master/engine/unit.go).
+
+#### Motivations
+- The `OriginID` field really provides no use. The value of this field is set by the sender of the message, so we certainly cannot rely on its correctness. We can validate it by [checking it against the peer ID of the sender](https://github.com/onflow/flow-go/pull/1163), but this is actually unnecessary since the peer ID is enough to tell us who the sender is. We can extract the sender's network key from their peer ID, and if a downstream engine needs to know the Flow ID corresponding to this network key, we can determine this from the protocol state. If we cannot find a corresponding Flow ID, then the sender is not a staked node.
```
yhassanzadeh13, line 95: Wouldn't this be easier to keep the `OriginID` as part of payload, require every message being signed by its `OriginID` and then on the recipient side verify the signature on payload against the verification key of the `OriginID`? 

The advantage this signature checking is reducing the heavy lifting of proposed cross-referencing between peer ID and flow ID at the recipient size to just a signature verification (which we literally need to do). It also saves a lot from the application layer running the assumption that incoming messages are already checked against authenticity and integrity and reduces the surface of redundancy that decoupling the authenticity (at network layer) and integrity checking (at the application layer) in the proposed solution introduces. 

P.S: I think most of our message types are literally meant to have a valid signature. So, the only networking level compromise we make is to send an extra 32 bytes of `OriginID` while at return we have a more coherent authenticity and integrity checking for our messages.

yhassanzadeh13 (commented):
Thanks for the write-up. Added a few comments, would be happy to discuss in details.

flips/context-propagation.md
smnzhu: > The advantage of this signature checking is reducing the heavy lifting of proposed cross-referencing between peer ID and flow ID at the recipient side to just a signature verification (which we literally need to do). It also saves a lot from the application layer running the assumption that incoming messages are already checked against authenticity and integrity and reduces the surface of redundancy that decoupling the authenticity (at the network layer) and integrity checking (at the application layer) in the proposed solution introduces.

I don't think the cross-referencing between peer and flow ID is actually a lot of work, in fact with a hashmap it would just be a (approximately) constant-time lookup. LibP2P already does signature verification, so I'm not sure why we would want to do again at the application level as this seems redundant. I don't think the proposed approach introduces more redundancy, because authenticity and integrity checking are completely separate, so the two stages are doing two separate but necessary things. We would be introducing redundancy by checking signatures at the application layer, as this combines authenticity and integrity checking into a single step but authentication is already done by libp2p

flips/context-propagation.md
smnzhu: > For example, assume the // do work part here is a long-running task. Canceling the context wouldn't kill that task until it returns and the subsequent iteration of for starts. However, the parent component called ctx.Cancel() assumes all sub-components have been terminated immediately. Right?

No, it doesn't assume that all sub-components have been terminated immediately. That's what the `Done` method is for. You used the channel returned from `Done` to actually wait for shutdown to complete. the cancelling of the context just triggers the shutdown.

smnzhu:
> I can see that one of the reasons this FLIP proposed was to handle the non-idempotency of the ReadyDoneAware. However, it seems the problem is bleeding to a separate domain by introducing Runnable. I consider the solution does not fully comply with the problem defined by the FLIP. In my opinion, nothing can still prevent a Runnable component to be Run several times. By accepting distinct contexts, Run does not show an idempotent behavior.
>
> Hypothetically, even introducing the capability of binding the same Run function to several distinct contexts looks unsafer than the existing ReadyDoneAware non-idempotency vulnerability. As we still totally delegate this responsibility to the developer to treat Run as a non-idempotent component. This can cause more serious safety vulnerabilities.
>
> So, while checking the readiness of a component is idempotent in this solution, the way Run method is defined extends the surface of vulnerability.

@yhassanzadeh13 If we really want to make `Run` idempotent, we can easily do so using something like `sync.Once`. Then, both `Run` and readiness checking will idempotent.

However, as @m4ksio and @zhangchiqing have both noted, defensive programming isn't the standard Go approach and is not actually needed in most cases. 

The reason why non-idempotency was an issue in the first place is because there was no separation of the readiness checking capability and starting capability for a component. Readiness checking is the one that actually needs to be idempotent, because this capability must be available to multiple other components and readiness checking may be done multiple times. However, starting a component doesn't necessarily need to be idempotent, because a component only needs to be started once. Because these two currently aren't separated, then we have to make them either idempotent or non-idempotent together. By separating them, we can make the necessary one idempotent (readiness checking), while avoiding the complexity of introducing idempotency for the one that's not necessary (starting the component)

flips/context-propagation.md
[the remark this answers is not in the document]
zhangchiqing: I thought about another solution, that might be easier to implement.

Since the goal is to make `ReadyDoneAware` idempotent, we could achieve by making `unit.Ready` and `unit.Done` idempotent, and each engine could wraps its Ready work with `unit.Ready`.

How to make `unit.Ready` idempotent?
We could cache the ready channel. If Ready is called more than once, then the cached ready channel will be returned. This way the second call will return the same ready channel and won't trigger additional initialization work.

In fact, it doesn't make sense to me that a module would initialize more than once, and I don't think there is such case exist now. 

So what I'm proposing is to just adjust `unit.Ready`:
```golang
func (u *Unit) Ready(checks ...func()) <-chan struct{} {
        // concurrent call will be blocking by the lock, this ensures only the first process will 
        // call the `checks` function, the second process will wait, and get notified once the first 
        // process finish all the checks.
        u.Lock()
        defer u.Unlock()

        if u.ready != nil {
                return u.ready
        }
            
	ready := make(chan struct{})
	go func() {
		for _, check := range checks {
			check()
		}
		close(ready)
	}()

        c.ready = ready
	return ready
}

```

In your example, A could be constructed with B whose Ready has been called by its owner, and A can call B's Ready again, which will not trigger another initialization, but wait on the previous initialization call to finish

AlexHentschel:
### Summary of discussion
Alex & Simon; Sept 10, 2021

* Networking layer must provide strict admission control for staked nodes
  - when some node attempts to open connection, the networking layer must check
     that this is an authorized connection and refuse it otherwise
  - this must happen at the point of the handshake
  - only messages from authorized peers should be decoded (!)
     Otherwise a potential attack vector
* For gossip network: messages should be de-duplicated based on payload alone 
   - message origin should _not_ factor into whether a message is considered a duplicate or not
   - for example, consider a block proposal from Alice. If Bob re-broadcasts the proposal, we need to consider it as a duplicate. Otherwise,  1/3 of byzantine nodes (about 40 consensus nodes) could 40x message processing overhead by just re-broadcasting all blocks. 
   - this also allows us to properly separate networks between staked nodes and unstaked nodes, without having to worry about repetition of the same message (e.g block) broadcasted by multiple access nodes  
* There should be a _strong_ separation between network of staked nodes and network for unstaked nodes
   * Minimum separation on an engine level: Engines should either deal with messages from staked or unstaked nodes; but not both. Processing both by the same engine is an invitation for liveness attacks, because then messages from the unstaked nodes can starve liveness-critical messages from staked nodes of processing resources.
   * better: complete process separation (different OS processes for processing messages from staked vs unstaked nodes)
  - ideally (eventual goal): hardware separation 
  
  In all cases, we only require one identifier:
     flowID (for staked nodes) or network-derived peerID (unstaked nodes)
* For messages from staked nodes, we would like an Inspector object: 
   * networking layer for staked network must guarantees message integrity and authenticity 
   * if the recipient finds that something is wrong with the message, we can _always_ slash the message origin 
   * proof that origin sent this message, would be generated on-demand 
   * could be an extension of context:
      - provide proof that peer really sent the message
      - we can also include the `peer.ID` of the sender in this inspector object
      - it makes for requests from staked nodes that they can be canceled

For API safety, it would probably make sense to have different APIs for the staked and unstaked network. Thereby, the API reflects that messages from both networks should not be mixed.

smnzhu:
Once we have message queues implemented on all engines according to https://github.com/onflow/flow/pull/343, we should be able to combine the context passed in to `engine.Process` with the context passed into any threads processing the queue with something like https://github.com/teivah/onecontext

smnzhu:
Split networking FLIP to https://github.com/onflow/flow-go/pull/1306

flips/context-propagation.md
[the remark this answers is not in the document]
smnzhu: I think this is one valid solution, and I agree it may be easier to implement given the current state of the codebase. However, I would argue that if I were to design the API from scratch today, I would design it the way described in this FLIP, for all the reasons described in the "Motivations" section. Using contexts to start / stop components is so much more idiomatic in Go, and makes the APIs much more self explanatory. It also makes the separation of concerns more explicit.

huitseeker:
This should probably be revised to reflect on just the API changes we want to do on top of #1275.

AlexHentschel (approved):
This is a great proposal. Thanks for the detailed explanations and sample code. 
I don't have concerns; only a few questions.  

In addition to the specific questions, which I have added as comments, I have the following high-level question:

#### Would it make sense to make `Startable.Start` method idempotent (or error on repeated starts)?

Generally, we create an instance of a particular component and start it once (not sure if there is even a component we start repeatedly). If a component crashes, we create a new component, which then again is also started only once. 

I feel there are potentially a lot of things that could go wrong, if we would repeatedly call `Start` on a component. Eliminating this case (e.g. returning an error) might make the implementation easier to understand.

flips/context-propagation.md
yhassanzadeh13: > LibP2P already does signature verification, so I'm not sure why we would want to do again at the application level as this seems redundant

This application layer signature verification is not redundant: 

When a peer _sends_ message `M` it signs it with its _networking_ key, which is verified at any intermediary pubsub node as well as the receiver. This [signature verification is done at libp2p level](https://github.com/libp2p/go-libp2p-pubsub/blob/65f48fb2c8b37563525e3d44042ea2418650f854/validation.go#L275-L283) against the [networking key](https://github.com/libp2p/go-libp2p-pubsub/blob/65f48fb2c8b37563525e3d44042ea2418650f854/sign.go#L85-L106) of the _sender_. The aim of this signature verification provided by libp2p is solely to ensure **integrity** on the **end-to-end delivery** of the message. 

The reason we can't only rely on the libp2p signature verification is it does not account for the authenticity of the message, i.e., Alice receiving a message from Bob can ensure the integrity of message _sent_ by Bob and not the _authenticity_ of the message as whether it has been _created_ by Bob. Moving down the BFT road and relying only on signature verification at networking layer, the messages fail to be attributable to their creator and lack [non-repudiation](https://www.cryptomathic.com/products/authentication-signing/digital-signatures-faqs/what-is-non-repudiation#:~:text=Non%2Drepudiation%20is%20the%20assurance,deny%20the%20validity%20of%20something.&text=In%20this%20context%2C%20non%2Drepudiation,the%20sending%20of%20a%20message.). Backing to our example, when Alice receives message `M` from Bob, she cannot verify whether Bob created that message or Eve, or anyone else. In other words, the _origin_ of the message remains in mist from the cryptographic point of view. 

That is why we have a protocol level key (i.e., staking key) as well, and at the protocol level the messages are supposed to be signed by the staking key of their corresponding _creator_. In this way, no matter who Alice receives message `M` from, she always can check it against the public staking key of the _creator_ to ensure it is _originated_ from that creator. This later on enables Alice to submit message `M` as an _evidence_ of a malicious behavior originated from Eve (i.e., slashing), and _anyone_ can verify the correctness of Alice's claim by (1) checking whether content of `M` represents a misbehavior, and (2) whether `M` includes a signature on its context that is verifiable against Eve's public staking key. This is just an example of a vital protocol feature of Flow that sticking solely on signature verification at libp2p level fails to provide. Another example is the general provider scenario that we aim that literally nodes can cache some entities they receive (e.g., `ChunkDataPack`) and provide them to other nodes requesting on the same topic. In this case, a verification node receiving a `ChunkDataPack` from another verification node needs to make sure that this `ChunkDataPack` has been created by a _staked_ execution node, and this assurance is provided by checking the content of `ChunkDataPack` for a signature on its payload that is valid against the public key of its origin ID, i.e., the execution node that created that `ChunkDataPack`. 

Now the reason including `OriginID` in the message `M` is crucial is that it enables any node receiving `M` to lookup public staking key of `M` from protocol state and check the signature of the message against `OriginID`'s public key. 
Including the `OriginID` together with signature in the messages makes them self-certifying, i.e., the message contains all forensic information needed to verify it later on by any party. Note that libp2p signature and protocol level signatures are different. The latter is part of the protocol-level message. 

> I don't think the cross-referencing between peer and flow ID is actually a lot of work, in fact with a hashmap it would just be a (approximately) constant-time lookup.

Looking up from hash table is a constant-time operation with respect to the number of entries in the hash table. However, it is not necessarily cheap, as it includes a hash operation which can be computation-heavy. I admit that we have hundreds of places that we do hashmap lookup, and I am not against adding such lookup at networking layer. But as we go BFT and especially on unsteaked network, can we need to be careful about adding extra operations at networking layer. Such operations can be exploited by adversarial nodes to perform DDoS attacks, i.e., flood of messages that cause additional hashmap lookups. To the best of my knowledge, our security approach at networking layer is mostly proactive than reactive especially when it comes to availability-based attacks such as flooding. So the more we can save networking layer from extra bookkeeping, the more secure networking layer we have against such DDoS attacks. Of course DDoS preventions and rate limiting are possible, but they have their own overheads as well.

yhassanzadeh13 (approved):
Thanks for the FLIP, the component life cycle management part looks good, however, I agree with [this comment](https://github.com/onflow/flow-go/pull/1167#pullrequestreview-762461257) of @AlexHentschel to be considered before merging. Also, I have [this comment](https://github.com/onflow/flow-go/pull/1167#discussion_r715231542) outstanding about `OriginID` but it seems that it is no longer part of this FLIP. So considering the life cycle management it is good to go.

flips/component-interface.md
```diff
@@ -0,0 +1,267 @@
...
+[Clearer documentation](https://github.com/onflow/flow-go/pull/1032) and a new [`LifecycleManager`](https://github.com/onflow/flow-go/pull/1031) component were introduced as a step towards fixing this by providing concurrency-safety for components implementing `ReadyDoneAware`, but this still does not provide a clear separation between the ability to start / stop a component and the ability to check its state. A component usually only needs to be started once, whereas multiple other components may wish to check its state.
+
+Also, there currently does not exist any mechanism to handle fatal errors within a component. We want to enable a way for such errors to be handled by the parent of the component (e.g restart the component, propagate the error further up the stack by throwing a fatal error itself) rather than crashing the entire program.
+
+## Proposal
+
+Moving forward, we will add two new interfaces in addition to the existing `ReadyDoneAware`:
+- A `Startable` interface which can be used to start and stop a component:
+  ```golang
+  type Startable interface {
```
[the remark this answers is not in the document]
smnzhu: That's the great thing about contexts: We don't need an explicit method to stop the component.

The component is stopped simply by cancelling the context.

flips/context-propagation.md
smnzhu: > The reason we can't only rely on the libp2p signature verification is it does not account for the authenticity of the message, i.e., Alice receiving a message from Bob can ensure the integrity of message sent by Bob and not the authenticity of the message as whether it has been created by Bob. Moving down the BFT road and relying only on signature verification at networking layer, the messages fail to be attributable to their creator and lack non-repudiation. Backing to our example, when Alice receives message M from Bob, she cannot verify whether Bob created that message or Eve, or anyone else. In other words, the origin of the message remains in mist from the cryptographic point of view.

Actually, the cool thing is that libp2p verifies both :)

[ReceivedFrom](https://github.com/libp2p/go-libp2p-pubsub/blob/0c7092d1f50091ae88407ba93103ac5868da3d0a/pubsub.go#L214) is for integrity, and [From](https://github.com/libp2p/go-libp2p-pubsub/blob/0c7092d1f50091ae88407ba93103ac5868da3d0a/pb/rpc.pb.go#L144) is for authenticity.

In your example, `ReceivedFrom` would be Bob, and `From` would be the origin of the message. The entire message envelope is signed by the origin and the signature is included along with the message, and so libp2p is able to verify both of these :)

smnzhu:
> This is a great proposal. Thanks for the detailed explanations and sample code.
> I don't have concerns; only a few questions.
> 
> In addition to the specific questions, which I have added as comments, I have the following high-level question:
> 
> #### Would it make sense to make `Startable.Start` method idempotent (or error on repeated starts)?
> Generally, we create an instance of a particular component and start it once (not sure if there is even a component we start repeatedly). If a component crashes, we create a new component, which then again is also started only once.
> 
> I feel there are potentially a lot of things that could go wrong, if we would repeatedly call `Start` on a component. Eliminating this case (e.g. returning an error) might make the implementation easier to understand.

This is a good point and one I've thought about quite a bit. 

> Thanks for the FLIP, the component life cycle management part looks good, however, I agree with [this comment](https://github.com/onflow/flow-go/pull/1167#pullrequestreview-762461257) of @AlexHentschel to be considered before merging. Also, I have [this comment](https://github.com/onflow/flow-go/pull/1167#discussion_r715231542) outstanding about `OriginID` but it seems that it is no longer part of this FLIP. So considering the life cycle management it is good to go.

@yhassanzadeh13 I've replied to your outstanding comment :)

flips/component-interface.md
```diff
@@ -0,0 +1,267 @@
...
+  func (e *ErrorManager) ThrowError(err error) {
+    e.errors <- err
+    runtime.Goexit()
+  }
+  ```
+
+  Error handling for components can be implemented like so:
+
+  ```golang
+  type ErrorHandler func(ctx context.Context, errors <-chan error, restart func())
```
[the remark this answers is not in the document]
smnzhu: Hey @AlexHentschel, thanks for the comment. Please see https://github.com/onflow/flow-go/pull/1275, which has a more up-to-date design for the error-handling part of this FLIP. I will actually be removing the error handling stuff from this FLIP, since it is already fully implemented in https://github.com/onflow/flow-go/pull/1275 and any further discussion can happen there.

Briefly, to answer your question, the idea is that if a component has experienced an irrecoverable error, it is not necessarily the case that it should *always* be restarted. For example, we may actually decide that the error cannot be fixed by a restart and should actually be propagated up the stack by throwing an irrecoverable error in the parent component as well. Or we may wish to do some additional cleanup / actions before restarting, such as firing an alert to pagerduty, logging the error, etc.

`ErrorHandler` gives the parent component the power to decide how to handle errors in the child component. The key here is that the *parent* is the one who should decide how to handle an irrecoverable error that occurs in the child component. An irrecoverable error is, by definition, an error that the child component doesn't know how to handle (if it did, the error would be recoverable), so all it can do is tell the parent component about the error and hope the parent knows how to handle it (whether that be via restarting, propagating the error further up the stack, etc).

flips/component-interface.md
```diff
@@ -0,0 +1,267 @@
...
+      select {
+      case <-ctx.Done():
+        return nil, nil, ctx.Err()
+      case <-component.Ready():
+      }
+
+      go errorHandler(runCtx, component.Errors(), func() {
+        restartChan <- struct{}{}
+        runtime.Goexit()
+      })
```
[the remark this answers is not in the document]
smnzhu: As you said, it doesn't actually matter much. The reason I started it after is because we want to make sure the component actually starts up properly, otherwise we don't even need to consider this part.

flips/context-propagation.md
huitseeker: "Verifies both" is a bit of a simplification.
- libp2p [as configured in the Flow codebase today](https://github.com/onflow/flow-go/blob/be9979097c576ab8e71358f4d8c5553a7fb64f2d/network/p2p/libp2pNode.go#L742-L744) requires and checks a message signature from the source of a message (the `From`) at every carrier of the message,
- libp2p also has transport encryption and authentication on every direct p2p link, using a Noise key exchange with an XX pattern, which guarantees that the `ReceivedFrom` field is the PeerID of the actor on the other end of the direct link.
- Checking that this `ReceivedFrom` is the Flow node we expect on the other end of the line, or _a fortiori_ a valid Flow node, is left as an exercise to the out-of-libp2p implementer. There are places where we do this upon link establishment (e.g. subscription validator, readConnection).

AlexHentschel (approved):
Thanks @smnzhu for the work on this FLIP and all contributors for the great discussion. 👏 

### Lets get this FLIP over the finish line
1. Are there any open change requests pending? (I have a tiny one: see below). 
2. Merge 


#### How do we handle repeated calls to `Startable.Start`?
* Making the `Startable.Start` method idempotent adds complexity, which is not necessary. Agree, lets not make it itempotent. 
* But what about erroring on repeated `Start` calls:
    > since each component should only have a single "parent" / "owner", the parent should be responsible for making sure that they only start the component once
   * I think we all agree that "a component should only be started once" is a useful convention.
   * If we were certain that our code is bug-free, we could stop the discussion here. 
   * Reality is that we are human and make mistakes. Therefore, we should weigh the complexity of enforcing this convention vs the potential problems it could cause if we did _not_ enforce the condition and started the component accidentally twice:
       * complexity of enforcing convention "Component should only be started once": I think this can be achieved by a _single_ atomic `started := atomic.NewBool(false)`:
          ```golang
          if !started.CAS(false, true) {
               return fmt.Errorf("repeated call to starup")
          }
          ```
          I would say: implementation complexity is negligible (including a test)
       * Benefit of checking: 
          * we have _many_ components and enforcing the convention would apply to _all_ components. 
          * Accidentally starting a component twice could lead to very subtle errors (especially, if the component was designed to be single-threaded). 
          * Since we probably want to restart _some_ components, we already have cases where `Start` is called repeatedly (per convention on different components). I think erroring if we started a component repeatedly would be a useful safety measure.

peterargue:
this looks great. maybe update the example with your ideas from https://github.com/onflow/flow-go/pull/1275 and https://github.com/onflow/flow-go/pull/1355. They address Alex's comments.

smnzhu:
> this looks great. maybe update the example with your ideas from #1275 and #1355. They address Alex's comments.

I'll do that, thanks :)

smnzhu:
@AlexHentschel Okay, I am happy to enforce the convention that `Start` should only be called once. If you are interested, I've implemented a version of `ComponentManager` in https://github.com/onflow/flow-go/pull/1355/files, please feel free to take a look.

There I am using `sync.Once`, but if we would like to explicitly throw an error we can certainly change it to do that as well.

smnzhu:
@AlexHentschel Okay, I am happy to enforce the convention that `Start` should only be called once. If you are interested, I've implemented a version of `ComponentManager` in https://github.com/onflow/flow-go/pull/1355/files, please feel free to take a look.

There I am using `sync.Once`, but if we would like to explicitly throw an error we can certainly change it to do that as well.

huitseeker (changes requested):
Requesting changes to keep track of this as needing updates from #1275 (an possibly #1355 too).

flips/component-interface.md
```diff
@@ -0,0 +1,364 @@
...
+// this is the context for the routine which manages the component
+var parentCtx context.Context
+
+ctx, cancel := context.WithCancel(parentCtx)
+signaler := irrecoverable.NewSignaler()
+signalerCtx := irrecoverable.WithSignaler(ctx, signaler)
+
+go func() {
+  select {
+  case err := <-signaler.Error():
```
huitseeker, line 102: We have a bunch of `Error()` APIs, especially in the context of `componentXXXError`, where they serve a completely different purpose. A comment here indicating:
- the signaler is creating a channel,
- where to find it after
Would help

flips/component-interface.md
```diff
@@ -0,0 +1,364 @@
...
+- This allows us to separate the capability to check a component's state from the capability to start / stop it. We may want to give multiple other components the capability to check its state, without giving them the capability to start or stop it. Here is an [example](https://github.com/onflow/flow-go/blob/b50f0ffe054103a82e4aa9e0c9e4610c2cbf2cc9/engine/common/splitter/network/network.go#L112) of where this would be useful.
+- This provides a clearer way of defining ownership of components, and hence may potentially eliminate the need to deal with concurrency-safety altogether. Whoever creates a component should be responsible for starting it, and therefore they should be the only one with access to its `Startable` interface. If each component only has a single parent that is capable of starting it, then we should never run into concurrency issues.
+
```
huitseeker, line 141: We may want to introduce those (valid a great points) with  a line or two to highlight the hierarchical supervision paradigm, even if the cancellation explanation hints at it.

Each `Component` is equipped with a fresh signaler to its parent at launch, and is thus supervised by his parent for any irrecoverable errors it may encounter (the call to `WithSignaler` replaces the signaler in a parent context). As a consequence, sub-components themselves started by a component have it as a supervisor, which handles their irrecoverable failures, and so on.

flips/component-interface.md
```diff
@@ -0,0 +1,364 @@
...
+      case <-restartChan:
+      }
+
+      return nil
+    }
+
+    for {
+      if err := start(); err != nil {
+        return err // failure to start
+      }
```
huitseeker, line 217: A line of comment for this special case would help.

flips/component-interface.md
```diff
@@ -0,0 +1,364 @@
...
+    OnStart(ComponentStartup) ComponentManagerBuilder
+
+    // AddWorker adds a worker routine for the ComponentManager
+    AddWorker(ComponentWorker) ComponentManagerBuilder
+
+    // AddComponent adds a new sub-component for the ComponentManager.
+    // This should be used for critical sub-components whose failure would be
+    // considered irrecoverable. For non-critical sub-components, consider using
+    // RunComponent instead.
+    AddComponent(Component) ComponentManagerBuilder
```
huitseeker, line 257: "`// See #1355`" ?

huitseeker (commented):
Thanks a bunch, this is definitely going in the right direction˜

smnzhu:
bors merge

PR #1168, opened 2021-08-19
[Consensus Voting] Event handler voting v2

zhangchiqing:
Close https://github.com/dapperlabs/flow-go/issues/5622


consensus/hotstuff/eventhandler/event_handler_v2.go
```diff
@@ -0,0 +1,395 @@
...
+	if !viewChanged {
+		// this is a sanity check
+		// when I've processed the block for the current view and I'm not the next leader,
+		// the pacemaker should trigger a view change
+		nextView := curView + 1
+		return fmt.Errorf("pacemaker should trigger a view change to (view %v), but didn't", nextView)
+	}
+
+	// current view has changed, go to new view
+	return e.startNewView()
```
AlexHentschel, line 368: ## ⚠️ missing logic for primary to collect votes (before view change)

we should only process to next view, if the PaceMaker tells us to

AlexHentschel (commented):
Overall looks good. 

I have tried to consolidate the code a bit in the event handler. Please see my PR https://github.com/onflow/flow-go/pull/1434 targeting your branch. This PR also fixes the two problems, which I flagged below.

consensus/hotstuff/eventhandler/event_handler_v2_test.go
```diff
@@ -0,0 +1,527 @@
...
+
+	// has block for newview
+	newviewblock := createBlockWithQC(es.newview.View, es.newview.View-1)
+	es.forks.blocks[newviewblock.BlockID] = newviewblock
+
+	// I'm the next leader
+	es.committee.leaders[es.newview.View+1] = struct{}{}
+
+	// no QC for the new view
+	err := es.eventhandler.OnQCConstructed(qc)
```
AlexHentschel, line 178: I think some tests from the past version collapse to the scenario in Consensus Voting V2: 
* previously, the event handler would process votes. In multiple of the following tests, the setup is similar
   * I have a block for the past view: https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/eventhandler/event_handler_test.go#L479
   * and we have the ability to produce a QC for the past view if we get another vote: https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/eventhandler/event_handler_test.go#L481
*  we now trigger a view change by receiving another vote for the past view https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/eventhandler/event_handler_test.go#L496
   1. we are still missing votes to propose right away (test `TestInNewView_NotLeader_HasBlock_NoVote_IsNextLeader_NoQC`)
   2. we could already have enough votes stored to propose right away (`TestInNewView_NotLeader_HasBlock_NoVote_IsNextLeader_QCBuilt_ViewChanged`)

I think it deserves some thought how to properly translate the tests to the new scenario. I am wondering whether these two tests might collapse into the same test case. 

In my mind, the most critical aspect to cover in this test is:
* Consider Consensus node `Alice` that is currently in view `V`, where `Alice` is leader for view `V +1`
* When Alice is in view `curView = V` and receives the block for `curView` (i.e. block `b1`), Alice remains in view `curView` to collect votes:
   * On the happy path, she gets enough votes and constructs a QC, i.e. she is in view `curView` and  her vote aggregator calls [`OnQCConstructed(qc)`](https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/eventhandler/event_handler_v2.go#L67) where `qc.View == curView`
   * What should happen is that by receiving the QC, Alice's  PaceMaker  transitions to the new view `curView <- V+1`,  where she will construct her block proposal using the QC just received. 
   * I believe the current implementation already does all this (see [here](https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/eventhandler/event_handler_v2.go#L407) and [here](https://github.com/onflow/flow-go/blob/33372766853dcf6c2e62dcb4aaff52a287285cd7/consensus/hotstuff/pacemaker/pacemaker.go#L80-L88)). Though, we don't test it. Instead, we test a scenario that cannot occur in practise, because by receiving block `b2`, the event handler would already transition to view `curView+1`

consensus/hotstuff/eventhandler/event_handler_v2_test.go
```diff
@@ -0,0 +1,527 @@
...
+}
+
+// a QC for future view triggered view change
+func (es *EventHandlerV2Suite) TestQCBuiltFutureViewChanged() {
+	// voting block exists
+	curView := es.paceMaker.CurView()
+
+	b1 := createBlockWithQC(curView, curView-1)
+	b2 := createBlockWithQC(curView+1, curView)
+	b3 := createBlockWithQC(curView+2, curView+1)
```
AlexHentschel, line 131: This test is great. Lets make sure we test with a correct scenario. The case you tested with cannot occur in reality, because having seen block b2 would already advance the view. Same for b3. 

The key point where the view does not advance is when the `qc.view` is not increasing. This would be a "fast-forward" attack by two byzantine nodes, that don't wait for the proposer of b1 to publish its block but just skip him. It is important that other consensus nodes don't just blindly follow and instead vote and the leader waits and attempts to construct a qc. 
```suggestion
// TestQCBuiltFutureViewChanged tests the handling of a "fast-forward" attack by two byzantine nodes: 
// The byzantine nodes don't wait for the previous proposer to publish its block but just skip him. 
// It is important that other consensus nodes don't just blindly follow and instead vote and the leader waits and attempts to construct a qc. 
func (es *EventHandlerV2Suite) TestQCBuiltFutureViewChanged() {
	// voting block exists
	curView := es.paceMaker.CurView()

	b1 := createBlockWithQC(curView, curView-1)
	b2 := createBlockWithQC(curView+1, curView-1)
	b3 := createBlockWithQC(curView+2, curView-1)
```

AlexHentschel (changes requested):
* The EventHandler logic requires some revisions (see comments)
* Regarding the `EventHandlerV2Suite`, I feel we already have some technical debt in our old tests, which we propagate here. (at least the documentation is a bit hard to follow, which makes it even harder to revise the tests). 
   * I would suggest to go over the old tests and try to understand what precisely they were attempting to test. Then we can decide how to carry over the test scenario to V2. 
 
  I have already started this, but it takes quite some time. Will add more comments regarding the test revisions.

AlexHentschel (commented):
Got through the PR. Otherwise looks good 👍

consensus/hotstuff/eventhandler/event_handler_v2.go
```diff
@@ -0,0 +1,416 @@
...
+		return nil
+	}
+
+	// we skip validation for our last own proposal
+	if proposal.Block.BlockID != e.ownProposal {
+
+		// validate the block. exit if the proposal is invalid
+		err := e.validator.ValidateProposal(proposal)
+		if model.IsInvalidBlockError(err) {
+			_ = e.voteAggregator.InvalidBlock(proposal)
```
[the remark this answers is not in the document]
zhangchiqing: I think the API spec is wrong. `InvalidBlock` [will not return the `mempool.DecreasingPruningHeightError`](https://github.com/onflow/flow-go/blob/18003f6c4782ac7196e508e83f88e4acf2f296f7/consensus/hotstuff/voteaggregator/vote_aggregator_v2.go#L211). 

Even if this returns this error, it means it is an invalid block below finalized height. I think the current way that event handler handles it is correct, which is just logging it as warning.

The invalid block detection has been handled by the vote aggregator and slashing will be handled by other modules, not event handler, right?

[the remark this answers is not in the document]
zhangchiqing: nvm, I think the case was previously the UnverifiableBlock case, so should be handled the same way

consensus/hotstuff/eventhandler/event_handler_v2_test.go
zhangchiqing: >In my mind, the most critical aspect to cover in this test is:


I think the test cases what you described are already covered: 
1. TestOnReceiveProposal_ForCurView_NoVote_IsNextLeader_NoQC
2. TestOnReceiveProposal_ForCurView_NoVote_IsNextLeader_QCBuilt_ViewChange

consensus/hotstuff/voteaggregator/vote_aggregator.go
```diff
@@ -360,7 +360,6 @@ func (va *VoteAggregator) tryBuildQC(blockID flow.Identifier) (*flow.QuorumCerti
 	}
 
 	va.createdQC[blockID] = qc
-	va.notifier.OnQcConstructedFromVotes(qc)
```
zhangchiqing, line 363: since the notification now moved from vote aggregator to [event handler](https://github.com/onflow/flow-go/pull/1168/files#diff-0cd5bf32d36129e3bf313db65087f712644e35809e4afc35ddd332cec784b723R77), there will be a bit delay from the moment, the qc is created to the moment its processed by event handler. The delay is from the qc go through the event loop.

In order to make both moments clear and be able to measure the delay, I [added an info log when the QC was created by the vote aggregator.](https://github.com/onflow/flow-go/pull/1168/files#diff-9fa312161217502a3ae5a872cbda8043a2afe6d36363e3f47193c0cb32f0800aR220-R225)

AlexHentschel: nice. Thanks.

AlexHentschel: Would be great to have the same log to `CombinedVoteProcessorV3`: https://github.com/onflow/flow-go/blob/d2d9003adf8611a2587c84e83ac3bbbb2aa026aa/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L231

AlexHentschel (approved):
very nice. Clean code with clear comments ✨

consensus/hotstuff/voter/voter_test.go
```diff
@@ -83,7 +84,7 @@ func testBelowVote(t *testing.T) {
 
 	_, err := voter.ProduceVoteIfVotable(block, curView)
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "not for current view")
+	require.Contains(t, err.Error(), "expecting block for current view")
```
durkmurder, line 87: we have changed what error is returned, let's make sure that we are getting an exception and not a sentinel error

consensus/hotstuff/voter/voter_test.go
```diff
@@ -95,7 +96,7 @@ func testAboveVote(t *testing.T) {
...
```
durkmurder, line 99: we have changed what error is returned, let's make sure that we are getting an exception and not a sentinel error

consensus/hotstuff/voter/voter_test.go
```diff
@@ -107,7 +108,7 @@ func testEqualLastVotedView(t *testing.T) {
 
 	_, err := voter.ProduceVoteIfVotable(block, curView)
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "not above the last voted view")
+	require.Contains(t, err.Error(), "must be larger than the last voted view")
```
durkmurder, line 111: we have changed what error is returned, let's make sure that we are getting an exception and not a sentinel error

consensus/hotstuff/voter/voter_test.go
```diff
@@ -119,7 +120,7 @@ func testBelowLastVotedView(t *testing.T) {
...
```
durkmurder, line 123: we have changed what error is returned, let's make sure that we are getting an exception and not a sentinel error

consensus/hotstuff/eventhandler/event_handler_v2.go
```diff
@@ -0,0 +1,427 @@
...
+	voteAggregator hotstuff.VoteAggregatorV2
+	voter          hotstuff.Voter
+	validator      hotstuff.Validator
+	notifier       hotstuff.Consumer
+	ownProposal    flow.Identifier
+}
+
+var _ hotstuff.EventHandlerV2 = &EventHandlerV2{}
+
+// New creates an EventHandlerV2 instance with initial components.
```
durkmurder, line 36:
```suggestion
// NewEventHandlerV2 creates an EventHandlerV2 instance with initial components.
```

durkmurder (approved):
Very good stuff. Nicely done. Left few comments tests related, otherwise ready to be :ship: .

zhangchiqing:
bors merge

zhangchiqing:
bors merge

zhangchiqing:
bors merge

consensus/hotstuff/eventhandler/event_handler_test.go
```diff
@@ -32,40 +33,40 @@ const (
 // TestPaceMaker is a real pacemaker module with logging for view changes
 type TestPaceMaker struct {
 	hotstuff.PaceMaker
-	t *testing.T
+	t require.TestingT
```
smnzhu, line 36: why was this updated?

consensus/hotstuff/eventhandler/event_handler_test.go
```diff
@@ -32,40 +33,40 @@ const (
...
 }
 
-func NewTestPaceMaker(t *testing.T, startView uint64, timeoutController *timeout.Controller, notifier hotstuff.Consumer) *TestPaceMaker {
+func NewTestPaceMaker(t require.TestingT, startView uint64, timeoutController *timeout.Controller, notifier hotstuff.Consumer) *TestPaceMaker {
 	p, err := pacemaker.New(startView, timeoutController, notifier)
 	if err != nil {
-		t.Fatal(err)
+		panic(err)
```
smnzhu, line 42: why this was changed?

consensus/hotstuff/eventhandler/event_handler_test.go
```diff
@@ -32,40 +33,40 @@ const (
...
 	}
 	return &TestPaceMaker{p, t}
 }
 
 func (p *TestPaceMaker) UpdateCurViewWithQC(qc *flow.QuorumCertificate) (*model.NewViewEvent, bool) {
 	oldView := p.CurView()
 	newView, changed := p.PaceMaker.UpdateCurViewWithQC(qc)
-	p.t.Logf("pacemaker.UpdateCurViewWithQC old view: %v, new view: %v\n", oldView, p.CurView())
+	log.Info().Msgf("pacemaker.UpdateCurViewWithQC old view: %v, new view: %v\n", oldView, p.CurView())
```
smnzhu, line 50: for tests, i think it makes sense to use `t.Logf`

consensus/hotstuff/eventhandler/event_handler_test.go
zhangchiqing: To be reused by rapid tests

consensus/hotstuff/eventhandler/event_handler_test.go
zhangchiqing: because t doesn't have the Fatal method

consensus/hotstuff/eventhandler/event_handler_test.go
zhangchiqing: yeah, but the `require.TestingT` doesn't have that method

zhangchiqing:
why can't I merge this PR?

zhangchiqing:
Thanks for fixing it @durkmurder

PR #689, opened 2021-05-06
Direct contract function call

janezpodhostnik:
ref: https://github.com/onflow/cadence/issues/888

depends on this cadence PR: https://github.com/onflow/cadence/pull/878

Changes deducting transaction fees from a meta transaction to a direct contract function call. This bypasses a lot of boilerplate that is run during meta-transaction invocation.

This does not directly implement  https://github.com/onflow/cadence/issues/859 but is the first step to it. To implement the final solution, the hostEnv reference needs to stay the same (cross transactions) in the FVM execution, or should be settable in the interpreter. Both of those solutions require a lot more work.

This PR has two side effects:
- fee deduction should be faster
- events (from fee deduction) should be properly triggered

PR #1169, opened 2021-08-19
FVM cleanup

janezpodhostnik:
Just some random cleanup I did going through the module.

The only functional change is in `fvm/errors/accounts.go` where the address was previously not set.


fvm/errors/codes.go
```diff
@@ -38,24 +38,24 @@ const (
...
-	ErrCodeAccountAuthorizationError   ErrorCode = 1055
-	ErrCodeOperationAuthorizationError ErrorCode = 1056
-	ErrCodeOperationNotSupportedError  ErrorCode = 1057
+	ErrCodeFVMInternalError          ErrorCode = 1050
+	ErrCodeValueError                ErrorCode = 1051
+	ErrCodeInvalidArgumentError      ErrorCode = 1052
+	ErrCodeInvalidAddressError       ErrorCode = 1053
+	ErrCodeInvalidLocationError      ErrorCode = 1054
+	ErrCodeAccountAuthorizationError ErrorCode = 1055
+	// ErrCodeOperationAuthorizationError ErrorCode = 1056
```
ramtinms, line 47: I think this should be in use, if its not probably I made a mistake somewhere, it should be used to account creation or other limited operations.

ramtinms (approved):

fvm/errors/codes.go
janezpodhostnik: I tried to find a references in past commits, but I cant find any. Do you have any more clues?

ramtinms: this part has to change https://github.com/onflow/flow-go/blob/master/fvm/errors/base.go#L165

fvm/errors/codes.go
```diff
@@ -38,24 +38,24 @@ const (
...
-	ErrCodeStorageCapacityExceeded            ErrorCode = 1103
-	ErrCodeGasLimitExceededError              ErrorCode = 1104
+	ErrCodeCadenceRunTimeError      ErrorCode = 1101
+	ErrCodeEncodingUnsupportedValue ErrorCode = 1102
+	ErrCodeStorageCapacityExceeded  ErrorCode = 1103
+	// ErrCodeGasLimitExceededError              ErrorCode = 1104
 	ErrCodeEventLimitExceededError            ErrorCode = 1105
 	ErrCodeLedgerIntractionLimitExceededError ErrorCode = 1106
-	ErrCodeStateKeySizeLimitError             ErrorCode = 1107
+	// ErrCodeStateKeySizeLimitError             ErrorCode = 1107
```
ramtinms, line 58: this should also be used here : https://github.com/onflow/flow-go/blob/master/fvm/errors/execution.go#L159

fvm/blueprints/account.go
```diff
@@ -1,43 +0,0 @@
-package blueprints
```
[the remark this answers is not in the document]
janezpodhostnik: This used to be needed to deduct transaction fees, but was deleted during a previous PR https://github.com/onflow/flow-go/pull/689/files#diff-6de3603dfa48d4d7f7a6727ac8275d2cc21704cf324722208e70aaba823cdfe6

However this code was also moved at the same time by a different PR, which resulted in it not being deleted.

fvm/errors/codes.go
```diff
@@ -27,9 +27,9 @@ const (
 const (
 	// tx validation errors 1000 - 1049
 	// ErrCodeTxValidationError         ErrorCode = 1000 - reserved
-	ErrCodeInvalidTxByteSizeError        ErrorCode = 1001
-	ErrCodeInvalidReferenceBlockError    ErrorCode = 1002
-	ErrCodeExpiredTransactionError       ErrorCode = 1003
+	ErrCodeInvalidTxByteSizeError     ErrorCode = 1001
+	ErrCodeInvalidReferenceBlockError ErrorCode = 1002
+	// ErrCodeExpiredTransactionError       ErrorCode = 1003
```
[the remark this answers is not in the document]
janezpodhostnik: Since these are constants that should be better yes! So we avoid someone reusing the same constant again. 

I will change this.

janezpodhostnik:
bors merge

PR #1170, opened 2021-08-19
[network] Check the originID in sendDirect too

### LINKED ISSUE #1115: [Network] Verify originID against libp2p ID

huitseeker:
Follow-up to #1163.
This closes #1115 (finally) by adding anti-spoofing to `SendDirect` connections as well.

### DIFF
--- a/network/p2p/middleware.go
+++ b/network/p2p/middleware.go
@@ -317,7 +317,7 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
 	log.Info().Msg("incoming stream received")
 
 	//create a new readConnection with the context of the middleware
-	conn := newReadConnection(m.ctx, s, m.processMessage, log, m.metrics, LargeMsgMaxUnicastMsgSize)
+	conn := newReadConnection(m.ctx, s, m.processAuthenticatedMessage, log, m.metrics, LargeMsgMaxUnicastMsgSize)
 
 	// kick off the receive loop to continuously receive messages
 	m.wg.Add(1)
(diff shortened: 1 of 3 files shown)


vishalchangrani:
at this point we can make the origin ID check a [message validator](https://github.com/onflow/flow-go/blob/master/network/validator.go). Can we done as a separate issue since the validator interface will have to change a bit to accept a public key in addition to the message.

vishalchangrani (approved):
left a comment but otherwise lgtm. thanks

smnzhu (approved):

PR #1172, opened 2021-08-19
add retry logic to broadcast and forward message

kc1116:
Add retry logic to DKG broker broadcast and submit result, and dkg messaging engine wait for sealed.


module/epochs/base_client.go
```diff
@@ -79,25 +86,28 @@ func (c *BaseClient) SendTransaction(ctx context.Context, tx *sdk.Transaction) (
 // WaitForSealed waits for a transaction to be sealed
 func (c *BaseClient) WaitForSealed(ctx context.Context, txID sdk.Identifier, started time.Time) error {
 
-	attempts := 1
-	for {
-		log := c.Log.With().Int("attempt", attempts).Float64("time_elapsed_s", time.Since(started).Seconds()).Logger()
+	constRetry, err := retry.NewConstant(waitForSealedRetryInterval)
+	if err != nil {
+		return fmt.Errorf("cannot create retry mechanism: %w", err)
```
jordanschalm, line 91: I would also `log.Fatal` here, like you're doing for the other cases.

module/dkg/broker.go
```diff
@@ -104,17 +107,22 @@ func (b *Broker) Broadcast(data []byte) {
...
-	if !success {
-		b.log.Error().Msg("failed to broadcast message")
+
+	expRetry, err := retry.NewExponential(RETRY_MILLISECONDS)
+	if err != nil {
+		b.log.Fatal().Err(err).Msg("create retry mechanism")
+	}
+	maxedExpRetry := retry.WithMaxRetries(RETRY_MAX, expRetry)
+
+	err = retry.Do(context.TODO(), maxedExpRetry, func(ctx context.Context) error {
```
jordanschalm, line 117: Given the API, I feel like the most reasonable thing to do here is to use `context.Background()` (likewise for the other instances of `context.TODO()`)

module/dkg/broker.go
```diff
@@ -20,7 +23,7 @@ import (
 const RETRY_MAX = 5
 
 // RETRY_MILLISECONDS is the number of milliseconds to wait between retries
-const RETRY_MILLISECONDS = 1000
+const RETRY_MILLISECONDS = 1000 * time.Millisecond
```
jordanschalm, line 26: If it's typed as a duration, we can omit the unit in the name
```suggestion
const RETRY_PERIOD = 1000 * time.Millisecond
```

jordanschalm (approved):
LGTM

jordanschalm:
Could you update retry logic [here](https://github.com/onflow/flow-go/blob/f0fc6f38057e60a4765cf384bc5c6d15c9f58924/module/epochs/qc_voter.go#L57) as well? Then we'll be consistent across all the the epoch-related components.

module/dkg/broker.go
kc1116: It's more for readability and aesthetics

jordanschalm: Sure, but it is no longer a variable representing a number of milliseconds. It's a variable representing a number of nanoseconds, with type information that lets it be treated abstractly as a duration without worrying about the underlying unit of measurement. If someone tried to use this value to represent a number of milliseconds, they would get the wrong value!

jordanschalm (approved):

engine/consensus/dkg/messaging_engine.go
```diff
@@ -13,6 +16,13 @@ import (
 	"github.com/onflow/flow-go/network"
 )
 
+// RETRY_MAX is the maximum number of times the engine will attempt to forward
+// a message
+const RETRY_MAX = 5
+
+// RETRY_MILLISECONDS is the number of milliseconds to wait between the two first tries
+const RETRY_MILLISECONDS = 1000
```
m4ksio, line 24: nitpicking, but usually in go we use duration like `1000*time.Millisecond`, improves readability

engine/consensus/dkg/messaging_engine.go
```diff
@@ -109,10 +119,20 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
...
+
+			maxedExpRetry := retry.WithMaxRetries(RETRY_MAX, expRetry)
+			err = retry.Do(e.unit.Ctx(), maxedExpRetry, func(ctx context.Context) error {
+				err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
+				if err != nil {
+					e.log.Err(err).Msg("error sending dkg message retrying")
+				}
+				return retry.RetryableError(err)
+			})
+			e.log.Error().Err(err).Msg("error sending dkg message")
```
m4ksio, line 135: Won't this always report error even if `err` is nil?

engine/consensus/dkg/messaging_engine.go
```diff
@@ -109,10 +119,20 @@ func (e *MessagingEngine) forwardOutgoingMessages() {
...
 		case msg := <-e.tunnel.MsgChOut:
-			err := e.conduit.Unicast(&msg.DKGMessage, msg.DestID)
+			expRetry, err := retry.NewExponential(RETRY_MILLISECONDS)
 			if err != nil {
-				e.log.Err(err).Msg("error sending dkg message")
+				e.log.Fatal().Err(err).Msg("failed to create retry mechanism")
 			}
+
+			maxedExpRetry := retry.WithMaxRetries(RETRY_MAX, expRetry)
+			err = retry.Do(e.unit.Ctx(), maxedExpRetry, func(ctx context.Context) error {
```
m4ksio, line 128: Is this OK that we block for the retries here? Both answers are valid, I don't have enough knowledge, just wanted to make sure we are OK with blocking until the message is propagated.
This will also block eventual quitting.

m4ksio (commented):
Looks good, just few questions

module/dkg/broker.go
```diff
@@ -175,16 +185,24 @@ func (b *Broker) Poll(referenceBlock flow.Identifier) error {
...
+	expRetry, err := retry.NewExponential(RETRY_MILLISECONDS)
+	if err != nil {
+		b.log.Fatal().Err(err).Msg("failed to create retry mechanism")
 	}
+	maxedExpRetry := retry.WithMaxRetries(RETRY_MAX, expRetry)
+
+	err = retry.Do(context.Background(), maxedExpRetry, func(ctx context.Context) error {
+		err := b.dkgContractClient.SubmitResult(pubKey, groupKeys)
+		if err != nil {
+			b.log.Error().Err(err).Msg("error submitting DKG result retrying")
```
huitseeker, line 197: nit
```suggestion
			b.log.Error().Err(err).Msg("error submitting DKG result, retrying")
```

module/dkg/broker.go
```diff
@@ -104,17 +106,25 @@ func (b *Broker) Broadcast(data []byte) {
...
+	expRetry, err := retry.NewExponential(RETRY_MILLISECONDS)
+	if err != nil {
+		b.log.Fatal().Err(err).Msg("create retry mechanism")
+	}
+	maxedExpRetry := retry.WithMaxRetries(RETRY_MAX, expRetry)
+
+	err = retry.Do(context.Background(), maxedExpRetry, func(ctx context.Context) error {
+		err := b.dkgContractClient.Broadcast(bcastMsg)
+		if err != nil {
+			b.log.Error().Err(err).Msg("error broadcasting DKG result retrying")
```
huitseeker, line 119: nit
```suggestion
			b.log.Error().Err(err).Msg("error broadcasting DKG result, retrying")
```

module/epochs/base_client.go
```diff
@@ -79,25 +86,28 @@ func (c *BaseClient) SendTransaction(ctx context.Context, tx *sdk.Transaction) (
...
+	attempts := 0
+	err = retry.Do(ctx, maxedConstRetry, func(ctx context.Context) error {
+		attempts++
+		log := c.Log.With().Int("attempt", attempts).Float64("time_elapsed_s", time.Since(started).Seconds()).Logger()
 
 		result, err := c.FlowClient.GetTransactionResult(ctx, txID)
 		if err != nil {
-			log.Error().Err(err).Msg("could not get transaction result")
-			continue
+			msg := "could not get transaction result retrying"
```
huitseeker, line 102: nit
```suggestion
			msg := "could not get transaction result, retrying"
```

engine/consensus/dkg/messaging_engine.go
kc1116: @jordanschalm ^

engine/consensus/dkg/messaging_engine.go
```diff
@@ -13,6 +16,13 @@ import (
 	"github.com/onflow/flow-go/network"
 )
 
+// RETRY_MAX is the maximum number of times the engine will attempt to forward
+// a message
+const RETRY_MAX = 5
```
tarakby, line 21: We could use Go naming conventions and make this variable private, like `retryMax`

engine/consensus/dkg/messaging_engine.go
tarakby: My comment is about the naming and the constant visibility: `retryMilliseconds` maybe?

module/dkg/broker.go
```diff
@@ -17,7 +19,7 @@ import (
 
 // RETRY_MAX is the maximum number of times the broker will attempt to broadcast
 // a message or publish a result
-const RETRY_MAX = 5
+const RETRY_MAX = 8
```
tarakby, line 22: maybe?
```suggestion
const retryMax = 8
```

module/dkg/broker.go
```diff
@@ -17,7 +19,7 @@ import (
...
 
 // RETRY_MILLISECONDS is the number of milliseconds to wait between retries
 const RETRY_MILLISECONDS = 1000
```
tarakby, line 25:
```suggestion
const retryMilliseconds = 1000
```

engine/consensus/dkg/messaging_engine.go
kc1116: Yes this is okay we want to make sure message is propagated to avoid unnecessary white board messages during failure.

m4ksio (approved):

engine/consensus/dkg/messaging_engine.go
jordanschalm: I agree we should use the `Duration` type, my point was _given_ we are using the more abstract `Duration` type (which abstracts away the scale and unit of measure) rather than a numeric type, we should not include a specific unit or scale of measure in the variable name.

jordanschalm (approved):

PR #1173, opened 2021-08-19
Fixing the flaky middleware unsubscribe test

vishalchangrani:
This passes on this branch locally while fails immediately on master:
`go  test -tags relic -run TestMiddlewareTestSuit  ./network/test -v  -race -count 10 -failfast`


network/test/middleware_test.go
```diff
@@ -491,17 +491,20 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+	m.ov[last].On("Receive", firstNode, mockery.Anything).Return(nil).Run(func(_ mockery.Arguments) {
+		msgRcvd <- struct{}{}
+	})
 
 	// first test that when both nodes are subscribed to the channel, the target node receives the message
 	err := m.mws[first].Publish(message1, testChannel)
 	assert.NoError(m.T(), err)
 
-	assert.Eventually(m.T(), func() bool {
-		return m.ov[last].AssertCalled(m.T(), "Receive", firstNode, mockery.Anything)
```
vishalchangrani, line 503: `AssertCalled` calls `Fail` if the assertion doesn't succeed. This caused the flakiness if the message was not delivered in 1 millisecond before the first call to `AssertCalled`. Subsequent calls may succeed but by then the test has already been marked as failed :(

vishalchangrani:
bors merge

network/test/middleware_test.go
huitseeker: 🤯

huitseeker (commented):
This makes sense!
OK, is this true of other assert calls?
I'm worried about these in our code base:
https://gist.github.com/3eecce8d13fc60e45180039855171324

vishalchangrani:
> This makes sense!
> OK, is this true of other assert calls?
> I'm worried about these in our code base:
> https://gist.github.com/3eecce8d13fc60e45180039855171324

I think so - the `AssertXXX` calls eventually call `testing.Errorf` which mark the test as failed.
Hence this pattern to calling `Eventually` { `AssertXXX`} will not work. I will fix those in a different PR.

huitseeker (approved):
Thanks a ton! I recorded the leftovers in https://github.com/onflow/flow-go/issues/1179

yhassanzadeh13 (approved):
👏

vishalchangrani:
bors merge

smnzhu:
Thanks @vishalchangrani !

PR #1174, opened 2021-08-20
Fix uploader module

m4ksio:
Turns out we cannot use single variable.
Fix propagating the module

### DIFF
--- a/engine/execution/computation/computer/uploader/uploader_test.go
+++ b/engine/execution/computation/computer/uploader/uploader_test.go
@@ -123,7 +123,7 @@ func Test_AsyncUploader(t *testing.T) {
 
 func Test_GCPBucketUploader(t *testing.T) {
 
-	t.Skip("requires GCP Bucket setup")
+	//t.Skip("requires GCP Bucket setup")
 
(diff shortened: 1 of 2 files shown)


vishalchangrani (approved):

huitseeker (approved):

huitseeker (changes requested):
You may want to fix the comment.

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -123,7 +123,7 @@ func Test_AsyncUploader(t *testing.T) {
 
 func Test_GCPBucketUploader(t *testing.T) {
 
-	t.Skip("requires GCP Bucket setup")
+	//t.Skip("requires GCP Bucket setup")
```
huitseeker, line 126: Are you sure?

m4ksio: Of course not.
We need CI for branches badly.
Thanks for catching this

huitseeker (approved):
Thanks!

PR #1133, opened 2021-08-13
Identity Provider

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/5735
closes https://github.com/dapperlabs/flow-go/issues/5739

### TODO:
Write tests for:
- [x] filtered provider
- [x] fixed provider
- [x] hierarchical translator
- [x] peerstore provider
- [x] protocol state provider
- [x] test `UpdateNodeAddresses`
- [x] Write comments
- [x] make it clear `UpdatableIDProvider` is test-only, or test it

### For a follow-up PR (?):
~~- create a default for repeated instances of [`id.NewFilteredIdentifierProvider(...`](https://github.com/onflow/flow-go/pull/1133/files#diff-8d9f5e7f5c389eaf47e11124c15e78a38eb3f941db0cc46529d95decbc7d4ef2R38-R45)~~ (made obsolete by the `SyncEngineIdentifierProvider` in #1186

PR #1180, opened 2021-08-20
[bootstrap] Key Generation for Unstaked Access Nodes

huitseeker:
This introduces a generator for the unstaked Access Nodes, which by convention only have positive secp256k1 keys.

The keys in question have a specific format (positive ECDSA Secp256k1 keys) because:
- they are meant to serve ephemeral nodes,
- allow a bijection between flow.NodeID, flow.NetworkPublicKey, libp2p.PeerID, libp2p.PublicKey
- hence allowing us to only retain the minimum amount of information for these nodes.
For this bijection, see #1165 and #1129.

This is to be used in a bootstrap of the unstaked Access Node post #1133. 

For now, having this in master allows the direct bootstrapping of network in tests by creating suitable libp2p peers.

### DIFF
--- a/cmd/bootstrap/utils/key_generation.go
+++ b/cmd/bootstrap/utils/key_generation.go
@@ -1,7 +1,12 @@
 package utils
 
 import (
+	"crypto/sha256"
 	"fmt"
+	gohash "hash"
+	"io"
+
+	"golang.org/x/crypto/hkdf"
 
 	sdkcrypto "github.com/onflow/flow-go-sdk/crypto"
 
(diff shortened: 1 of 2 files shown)


cmd/bootstrap/utils/key_generation.go
```diff
@@ -18,6 +23,75 @@ func GenerateMachineAccountKey(seed []byte) (crypto.PrivateKey, error) {
...
+			panic(err)
+		}
+		iteration_count++
+	}
+
+	if err != nil || iteration_count == max_iterations {
+		return nil, err
+	}
+	return foundKey, nil
+}
```
smnzhu, line 76:
```suggestion
func GenerateUnstakedNetworkingKey(seed []byte) (key crypto.PrivateKey, err error) {
	hkdf := hkdf.New(func() gohash.Hash { return sha256.New() }, seed, nil, []byte("unstaked network"))

	round_seed := make([]byte, len(seed))
	max_iterations := 20 // 1/(2^20) failure chance

	for i := 0; i < max_iterations; i++ {
		if _, err = io.ReadFull(hkdf, round_seed); err != nil {
			// the hkdf Reader should not fail
			panic(err)
		}
		if key, err = drawUnstakedKey(round_seed); err == nil {
			return
		}
	}

	return
}
```

Personally, I find this cleaner and easier to read.

smnzhu (approved):

cmd/bootstrap/utils/key_generation.go
huitseeker: Yup, thanks!

cmd/bootstrap/utils/key_generation.go
```diff
@@ -18,6 +23,63 @@ func GenerateMachineAccountKey(seed []byte) (crypto.PrivateKey, error) {
 	return keys[0], nil
 }
 
+// The unstaked nodes have special networking keys, in two aspects:
+// - they use crypto.ECDSASecp256k1 keys, not crypto.ECDSAP256 keys,
+// - they use only positive keys (in the sense that the elliptic curve point of their public key is positive)
```
tarakby, line 28: I didn't find a definition that uses the terms "positive" and "negative" points. It would be good to add a link to definitions using these terms since they are not self explanatory. 
I am more familiar with the "parity" term used in this context, or to not use any adjective and just give a definition (like `0x02` if `y mod p` is even, and 0x3 otherwise).

cmd/bootstrap/utils/key_generation.go
```diff
@@ -18,6 +23,63 @@ func GenerateMachineAccountKey(seed []byte) (crypto.PrivateKey, error) {
...
+		// negative key -> unsuitable
+		return nil, fmt.Errorf("Unsuitable negative key")
+	}
+	return key, nil
+}
+
+// GenerateUnstakedNetworkingKey draws ECDSASecp256k1 keys until finding a suitable one.
+// though this will return fast, this is not constant-time and will leak ~1 bit of information through its runtime
+func GenerateUnstakedNetworkingKey(seed []byte) (key crypto.PrivateKey, err error) {
+	hkdf := hkdf.New(func() gohash.Hash { return sha256.New() }, seed, nil, []byte("unstaked network"))
```
tarakby, line 51: 👌

tarakby (approved):
🏄

cmd/bootstrap/utils/key_generation.go
huitseeker: - I agree that connecting this bit with a notion of sign requires the reader to be familiar with curves defined in Weierstrass form with p>3 (and then the sign is deduced from the notion of point inverse and the square in the equation). We could indeed reformulate that part.
- I'm confused about where the notion of parity comes from here: it seems to me standards specify both little-endianness and the ỹ bit as an MSB ? (whereas a parity bit would be a LSB?)
e.g.:
- IEEE 1363 ( doi 10.1109/ieeestd.2000.92292 ) section A.9.6
- [SEC 1 v2](https://www.secg.org/sec1-v2.pdf) section 2.3.3
- [X9.62 section 4.2-4.3](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.202.2977&rep=rep1&type=pdf)

huitseeker:
bors merge

vishalchangrani (approved):

cmd/bootstrap/utils/key_generation.go
tarakby: - I think even considering prime fields with p>3, it's still not clear which point should be called positive between (x,y) and (x,p-y). In some other cases (like the Zcash definition for BLS12-381) the sign of a point is defined as the comparison of y with `(p-1)/2`. This is why I thought we had to specify what definition we are referring to in our case. 

Regarding the second point:
- IEEE 1363 defines ỹ as (y mod 2) with y seen as an integer less than p (because strictly speaking, y is a field element). It also uses the "rightmost" bit of the integer y. 
- SEC 1 uses the "rightmost" bit of y. Looking at the definition of octet strings, "rightmost" represents the LSB.
- X9.62 uses (y mod 2)

huitseeker: > I think even considering prime fields with p>3, it's still not clear which point should be called positive between (x,y) and (x,p-y)

OK, I get your point better, now. 

> SEC 1 uses the "rightmost" bit of y. Looking at the definition of octet strings, "rightmost" represents the LSB.

Yep, there's one step between the bit string and the octet sting I missed.
And of course, looking at a mod 2 over a field element, the notion of parity makes sense.

I'll issue a PR to reformulate comments.

tarakby: There is no rush to create a specific PR for this. Including the X9.62 link as you did was good.
We can just update the references to positive and negative keys in a future libp2p related PR.

PR #1188, opened 2021-08-23
[Consensus Voting] Vote signature verification 

durkmurder:
### Context

After discussion with @tarakby and @zhangchiqing we came to conclusion that signature verification has to be done by same module as aggregation since those operations are "close" from cryptography point of view. 

Also this idea aligns with [proposal](https://www.notion.so/dapperlabs/Architecture-Proposal-07c8a782e95c48cea67fa8caa1ea0e18) made by @AlexHentschel. 

### What was done

- This PR implements signature verification as it was designed originally and removes separate module that was introduced for that purpose. 
- Moved signature verification to to aggregation module. `RandomBeaconReconstructor` for beacon signature and `SignatureAggregator` for staking signature.


consensus/hotstuff/signature.go
```diff
@@ -8,6 +8,10 @@ import (
 // RandomBeaconReconstructor collects verified signature shares, and reconstructs the
 // group signature with enough shares.
 type RandomBeaconReconstructor interface {
+	// Verify returns true if and only if the signature is valid.
+	// It expects that correct type of signature is passed. Only SigTypeRandomBeacon is supported
+	Verify(signerID flow.Identifier, sig crypto.Signature) (bool, error)
```
zhangchiqing, line 13:
```suggestion
	Verify(signerID flow.Identifier, sig crypto.Signature) error
```

We could use different sentinal errors for different reasons of why the sig is invalid.

consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
```diff
@@ -33,17 +31,45 @@ func (c *ConsensusClusterVoteCollector) CreateVote(block *model.Block) (*model.V
...
 	}
 
-	sigType, err := c.validator.ValidateVote(vote)
+	sigType, err := c.parseSigType(vote)
 	// handle InvalidVoteError
+	if err != nil {
+		return fmt.Errorf("could parse vote signature type: %w", err)
+	}
+
+	valid, err := c.validateSignature(sigType, vote)
```
zhangchiqing, line 63: `sigType` is part of `vote.SigData`.

so you would actually do this:
```
  sigType, sig, err := c.parseSigType(v.SigData)
  if sigType == SigTypeStaking {
    c. combinedAggr.Verify(signerID, sig)
  } else if sigType == SigTypeRandomBeacon {
    c.reconstructor.Verify(signerID, sig)
  } else {
     return fmt.Errorf("unknown sig type")  
  }
```

consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
```diff
@@ -33,17 +31,45 @@ func (c *ConsensusClusterVoteCollector) CreateVote(block *model.Block) (*model.V
...
 	if err != nil {
 		return fmt.Errorf("could not verify vote signature: %w", err)
 	}
 
+	// TODO: handle invalid error with sentinel error or callback
+	if !valid {
+		return fmt.Errorf("invalid signature")
```
smnzhu, line 70: nit: `fmt.Errorf` typically is used if you need a format string for the error message. Since here the error message is a static string without any parameters, better to use `errors.New("invalid signature")`

consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
```diff
@@ -33,17 +31,45 @@ func (c *ConsensusClusterVoteCollector) CreateVote(block *model.Block) (*model.V
 	panic("implement me")
 }
 
+// parseSigType returns type of signature contained in model.Vote.
+func (c *ConsensusClusterVoteCollector) parseSigType(vote *model.Vote) (hotstuff.SigType, error) {
+	panic("implement me")
```
zhangchiqing, line 36: Added single sig [encoding & decoding interface here](https://github.com/onflow/flow-go/pull/1140/commits/06deac4f1900c87abf9f8abfc2488d71190f760c#diff-c567bc762113b75ac2bbb6f67b24c844001dd354f8e53b6adc732a6abe9e00c7R12)

This method could use simply use that function

durkmurder: yes, I agree, it mainly depends right now on https://github.com/onflow/flow-go/pull/1196

consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
durkmurder: there is a TODO to replace it with proper sentinel error, haven't just decided how specifically it will be done.

consensus/hotstuff/signature.go
durkmurder: I would leave it as it is, just to align with other `Verify` functions.

PR #1189, opened 2021-08-23
[FVM] let cadence runtime record generic traces

ramtinms:
This PR unlocks cadence runtime to capture generic traces instead of a fixed set which is captured alongside metrics.


fvm/scriptEnv.go
```diff
@@ -660,70 +660,43 @@ func (e *ScriptEnv) ImplementationDebugLog(message string) error {
 	return nil
 }
 
-func (e *ScriptEnv) ProgramParsed(location common.Location, duration time.Duration) {
+func (e *ScriptEnv) RecordTrace(operation string, location common.Location, duration time.Duration, logs []opentracing.LogRecord) {
 	if e.isTraceable() {
-		locStr := ""
+		if logs == nil {
+			logs = make([]opentracing.LogRecord, 0)
+		}
```
turbolent, line 667: Does this have to be an allocated slice? Can it just be `nil`, or does `RecordSpanFromParent` not like that? (Same for the other `RecordTrace` function below)

fvm/scriptEnv.go
```diff
@@ -660,70 +660,43 @@ func (e *ScriptEnv) ImplementationDebugLog(message string) error {
...
+			})
 		}
-		e.ctx.Tracer.RecordSpanFromParent(e.traceSpan, trace.FVMCadenceParseProgram, duration,
-			[]opentracing.LogRecord{
-				{Timestamp: time.Now(),
-					Fields: []traceLog.Field{traceLog.String("location", locStr)},
-				},
-			},
-		)
+		spanName := trace.SpanName(fmt.Sprintf("%s.%s", trace.FVMCadenceTrace, operation))
```
turbolent, line 673: It might be good to avoid string formatting here and require callers to provide a constant string instead

fvm/scriptEnv.go
```diff
@@ -660,70 +660,43 @@ func (e *ScriptEnv) ImplementationDebugLog(message string) error {
...
-		if location != nil {
-			locStr = location.String()
-		}
-		e.ctx.Tracer.RecordSpanFromParent(e.traceSpan, trace.FVMCadenceCheckProgram, duration,
-			[]opentracing.LogRecord{{Timestamp: time.Now(),
-				Fields: []traceLog.Field{traceLog.String("location", locStr)},
-			},
-			},
-		)
-	}
```
turbolent, line 692: Nice, thank you for cleaning this up! ✨ 👌

fvm/scriptEnv.go
ramtinms: actually it can be nil, mostly I added this to handle the logic in the if location != nil condition, but I could move it there, so we don't have do any allocation for empty locations

fvm/scriptEnv.go
ramtinms: hmm, I think the idea here is to prefix the traces with the component path (e.g. everything under env is prefixed with 'fvm.env.xxx' and I was thinking everything from cadence runtime be prefixed with `fvm.cadence.xxx`. 

I was thinking to move this logic into the SpanName as a method called child which yields a new SpanName and hide this magic there. what do you think ? 

```
type SpanName string

func (s SpanName) Child(subOp string) SpanName {
	return SpanName(fmt.Sprintf("%s.%s", s, subOp))
}
```

turbolent: Yes, that makes sense, but I was worried about the cost of the formatting on each trace. Currently the names are constants

ramtinms: I see, that's definitely a cost, though I think the cost might be worth it initially, I did change it to string concatenation which takes a few nano seconds instead. I thought even if we decide to do be passed as a constant string by the caller, caller probably has to build the string.

fvm/scriptEnv.go
ramtinms: move the logic inside the location logging part

module/trace/constants.go
```diff
@@ -200,11 +200,7 @@ const (
 	FVMEnvRemoveAccountContractCode SpanName = "fvm.env.removeAccountContractCode"
 	FVMEnvGetSigningAccounts        SpanName = "fvm.env.getSigningAccounts"
 
-	FVMCadenceParseProgram     SpanName = "fvm.cadence.parseProgram"
-	FVMCadenceCheckProgram     SpanName = "fvm.cadence.checkProgram"
-	FVMCadenceInterpretProgram SpanName = "fvm.cadence.interpretProgram"
-	FVMCadenceEncodeValue      SpanName = "fvm.cadence.encodeValue"
-	FVMCadenceDecodeValue      SpanName = "fvm.cadence.decodeValue"
```
janezpodhostnik, line 207: would it make sense to still keep these as `FVMCadenceDecodeValue  SpanName = "decodeValue"`?

janezpodhostnik (approved):
Nice. Code looks cleaner with this.

fvm/scriptEnv.go
```diff
@@ -660,70 +660,43 @@ func (e *ScriptEnv) ImplementationDebugLog(message string) error {
 	return nil
 }
 
-func (e *ScriptEnv) ProgramParsed(location common.Location, duration time.Duration) {
+func (e *ScriptEnv) RecordTrace(operation string, location common.Location, duration time.Duration, logs []opentracing.LogRecord) {
 	if e.isTraceable() {
```
SupunS, line 664: Not really related to the new changes; maybe inverting the condition would be a good idea to reduce nested levels?

fvm/transactionEnv.go
```diff
@@ -941,70 +941,44 @@ func (e *TransactionEnv) ImplementationDebugLog(message string) error {
 	return nil
 }
 
-func (e *TransactionEnv) ProgramParsed(location common.Location, duration time.Duration) {
+func (e *TransactionEnv) RecordTrace(operation string, location common.Location, duration time.Duration, logs []opentracing.LogRecord) {
 	if e.isTraceable() {
```
SupunS, line 945: Similarly here

SupunS (approved):
LGTM!

fvm/scriptEnv.go
ramtinms: Smart :D

fvm/transactionEnv.go
ramtinms: updated

fvm/scriptEnv.go
ramtinms: updated

PR #1193, opened 2021-08-24
[Consensus Voting] Refactor verify QC

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5784

Review https://github.com/onflow/flow-go/pull/1209 first

### DIFF
--- a/consensus/hotstuff/verification/combined_verifier.go
+++ b/consensus/hotstuff/verification/combined_verifier.go
@@ -132,7 +132,7 @@ func (c *CombinedVerifier) VerifyConsensusQC(signers flow.IdentityList, sigData 
 		return fmt.Errorf("could not get dkg: %w", err)
 	}
 
-	aggregatedSigData, err := c.packer.Unpack(signers, sigData)
+	aggregatedSigData, err := c.packer.Unpack(block.BlockID, signers, sigData)
 	if err != nil {
 		// TODO: Use sentinal error
 		return fmt.Error("could not unpack sig data for block (%v): %w", block.BlockID, sigData)
--- a/consensus/hotstuff/verification/combined_verifier.go
+++ b/consensus/hotstuff/verification/combined_verifier.go
@@ -21,6 +21,7 @@ type CombinedVerifier struct {
 	keysAggregator *stakingKeysAggregator
 	beacon         module.ThresholdVerifier
 	merger         module.Merger
+	packer         hotstuff.Packer
 }
 
 // NewCombinedVerifier creates a new combined verifier with the given dependencies.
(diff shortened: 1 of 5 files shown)


consensus/hotstuff/verification/combined_verifier.go
```diff
@@ -122,3 +123,65 @@ func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, b
 	}
 	return stakingValid, nil
 }
+
+// VerifyConsensusQC verifies the validity of a combined signature on a quorum certificate.
+func (c *CombinedVerifier) VerifyConsensusQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
```
durkmurder, line 128: Who will call this function? Event handler?

zhangchiqing: This function will be removed to replace the existing `VerifyQC` method. I didn't replace now because I don't want to break the existing integration tests. 

Will replace once we swap all the V2 modules and fix the integration tests

consensus/hotstuff/verification/combined_verifier.go
```diff
@@ -122,3 +123,65 @@ func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, b
...
+	}
+	if !beaconValid {
+		return fmt.Errorf("invalid random beacon signature: %w", signature.ErrInvalidFormat)
+	}
+	// verify the aggregated staking signature next (more costly)
+	// TODO: eventually VerifyMany will be a method of a stateful struct. The struct would
+	// hold the message, all the participants keys, the latest verification aggregated public key,
+	// as well as the latest list of signers (preferably a bit vector, using indices).
+	// VerifyMany would only take the signature and the new list of signers (a bit vector preferably)
+	// as inputs. A new struct needs to be used for each epoch since the list of participants is upadted.
```
tarakby, line 156: We can update this TODO as we're choosing a different path.

```suggestion
  // TODO: use new V2 module module/signature.PublicKeyAggregator
```

consensus/hotstuff/verification/combined_verifier.go
```diff
@@ -122,3 +123,65 @@ func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, b
 	}
 	return stakingValid, nil
 }
+
+// VerifyConsensusQC verifies the validity of a combined signature on a quorum certificate.
```
AlexHentschel, line 127: can we please document the expected error returns

AlexHentschel (commented):
I would suggest to pause work on this PR until we have the Vote Processing in place. For me, there are a lot of dependencies on unfinished code such as:
* I think we have proper BLS aggregation implemented by now (@tarakby is that true?).
   If that is the case, I feel we could completely remove the `module.AggregatingVerifier` and `module.ThresholdVerifier` and instead directly call into the crypto primitives
* Previously, validating QCs was only done by the event handler (or the consensus follower). Therefore, we didn't need concurrency safety. The situation has now changed:
   * we process votes concurrently and potentially construct QCs concurrently
   * I feel we should stick with the current safety precautions: we validate a QC after we produced it (to avoid publishing a block with an invalid QC and getting slashed). Now that we produce QCs concurrently, should we validate them concurrently too? 
* I feel the error conventions for the sub-components are still not finalized. It is not possible for me to check whether error handling covers all protocol paths without a proper error convention of the sub components.

zhangchiqing:
close in favor of https://github.com/onflow/flow-go/pull/1601

PR #1194, opened 2021-08-24
[Consensus Voting] Sign Vote and Verify Proposer's Vote

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5968

When verifying a vote, there are two cases:
1. A leader to verify the follower's votes. This is stateful.
2. A followers to verify the leader's vote in the proposal (proposer's vote). This is stateless.

[1 is implemented by that PR](https://github.com/onflow/flow-go/pull/1527)

And this PR is to implement a node signing a vote and 2 (follower verifying leader's vote) in the above case.

This PR adds new modules for the signers and verifiers, it didn't replace the existing modules which will be done in future PRs. And also marked some modules to be deleted/replaced later, as doing it now will break a lot of existing test cases, making a lot of noises that will be hard to review.


AlexHentschel (commented):
Overall looks good. 

Some comments:
1. Tarak prefers if we don't use the term `DKG key` anymore: DKG is the process by which we _generate the key_, but it is not a property of the key. I think he prefers Random Beacon Key. Would be great to clean this up (at least for all the files this PR touches).
2. To me, it looks like there is a lot of overlap between this PR and https://github.com/onflow/flow-go/pull/1208 (?) 
   * specifically, `consensus/hotstuff/verification/combined_signer_v2.go`  seems to be identical in both PRs? Am I missing something 
3. I would prefer if `EpochAwareSignerStore` was implemented a bit more defensively 
    * Here, we assume that if I don't find the key for an epoch, I don't have a DKG key: https://github.com/onflow/flow-go/blob/eef63c14afcdec4435631caa65b7eeffe639dd61/module/signature/signer_store.go#L46-L51
    * But we are emitting an internal notification when the epoch is completed. So the node can know very precisely whether I has a dkg key or not. 
    
   Maybe that is a topic which is beyond the scope of this PR ... just wanted to mention it.

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+
+	"github.com/onflow/flow-go/consensus/hotstuff"
+	"github.com/onflow/flow-go/consensus/hotstuff/model"
+	"github.com/onflow/flow-go/consensus/hotstuff/signature"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module"
+)
+
+// CombinedSignerV2 creates votes for the main consensus. Per protocol specification
```
tarakby, line 14: I didn't get which module is `CombinedSigner` in the architecture proposal https://www.notion.so/dapperlabs/Architecture-Proposal-07c8a782e95c48cea67fa8caa1ea0e18.

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+// random beacon key it is available.
+type CombinedSignerV2 struct {
+	staking              module.AggregatingSigner
+	thresholdSignerStore module.ThresholdSignerStore
+	signerID             flow.Identifier
+}
+
+// NewCombinedSignerV2 creates a new combined signer with the given dependencies:
+// - the hotstuff committee's state is used to retrieve public keys for signers;
+// - the staking signer is used to create and verify aggregatable signatures for the first signature part;
```
tarakby, line 32:
```suggestion
// - the staking signer is used to create and verify aggregatable signatures for Hotstuff
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+// - the thresholdVerifier is used to verify threshold signatures
+// - the thresholdSignerStore is used to get threshold-signers by epoch/view;
+// - the signer ID is used as the identity when creating signatures;
+func NewCombinedSignerV2(
+	committee hotstuff.Committee,
+	staking module.AggregatingSigner,
+	thresholdVerifier module.ThresholdVerifier,
```
tarakby, line 39: This seems to be used only for testing. I suggest to remove it from the constructor.

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/onflow/flow-go/consensus/hotstuff/model"
+	"github.com/onflow/flow-go/consensus/hotstuff/signature"
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/module"
+	modulemock "github.com/onflow/flow-go/module/mock"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+// 1. if DKG was completed, then sign with random beacon key
+func TestDKGComplete(t *testing.T) {
```
tarakby, line 21:
```suggestion
// 1. if the node has beacon keys, then sign with random beacon key
func TestWithBeaconKeys(t *testing.T) {
```

tarakby (commented):
I assume there will be subsequent updates to other modules that do not fit, though that can be done in other PRs (`AggregatingSigner` , `ThresholdSigner` that should only sign but do a bunch of other things)

engine/consensus/dkg/reactor_engine.go
```diff
@@ -173,15 +173,21 @@ func (e *ReactorEngine) EpochCommittedPhaseStarted(currentEpochCounter uint64, f
...
+
+	if !hasDKGKey {
+		e.log.Err(err).Msgf("checking DKG key consistency: no DKG private info for next epoch: %v", currentEpochCounter+1)
 		return
 	}
 
 	nextDKGPubKey, err := nextDKG.KeyShare(dkgPrivInfo.NodeID)
 	if err != nil {
 		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG public key for next epoch")
+		return
```
zhangchiqing, line 190: I added this, which seems that was a bug, no? @jordanschalm 

If so, better to patch to master first, as this PR only goes to master

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+		signers:     make(map[uint64]module.MsgSigner),
+	}
+}
+
+// GetSigner returns the threshold-signer for signing objects at a
+// given view. The view determines the epoch, which determines the DKG private
+// key underlying the signer.
+// It returns:
+//  - (signer, nil) if DKG was completed in the epoch of the view, signer is not nil
+//  - (nil, protocol.ErrEpochNotCommitted) if no epoch found for given view
```
zhangchiqing, line 31: Does the caller care about this case?

engine/consensus/dkg/reactor_engine.go
jordanschalm: Yes this should be a return after the log, thanks Leo. Made PR to master for this: https://github.com/onflow/flow-go/pull/1583

module/signature/random_beacon_signer_store.go
jordanschalm: The caller would likely treat this case the same as a general unexpected exception.

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -0,0 +1,124 @@
...
+	nodeID.NodeID = signerID
+	nodeID.StakingPubKey = stakingPriv.PublicKey()
+
+	me, err := local.New(nil, stakingPriv)
+	require.NoError(t, err)
+	staking := modulesig.NewSingleSigner(encoding.ConsensusVoteTag, me)
+	signer := NewCombinedSignerV2(staking, beaconSignerStore, signerID)
+
+	dkg := &mocks.DKG{}
+	dkg.On("KeyShare", signerID).Return(pk, nil)
```
durkmurder, line 108: hm, if there is no private key how can we obtain a PK?

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+		return false, fmt.Errorf("could not split signature: %w", modulesig.ErrInvalidFormat)
+	}
+
+	dkg, err := c.committee.DKG(block.BlockID)
+	if err != nil {
+		return false, fmt.Errorf("could not get dkg: %w", err)
+	}
+
+	// get the signer dkg key share
+	beaconPubKey, err := dkg.KeyShare(signer.NodeID)
```
durkmurder, line 69: if we have failed DKG and we don't have a private key, will we have a public key in such situation?
We can move this down the path, after checking if we have `beaconShare` this way we will localize beacon sig checks and staking sig checks

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/storage"
+)
+
+type EpochAwareRandomBeaconSignerStore struct {
```
durkmurder, line 12: a minimal godoc would be appreciated

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+//  - (nil, error) if there is any exception
+func (s *EpochAwareRandomBeaconSignerStore) GetSigner(view uint64) (module.MsgSigner, error) {
+	// fetching the epoch by view, if epoch is found, then DKG must have been completed
+	epoch, err := s.epochLookup.EpochForViewWithFallback(view)
+	if err != nil {
+		return nil, fmt.Errorf("could not get epoch by view %v: %w", view, err)
+	}
+
+	// when DKG has completed,
+	// 		1. if a node successfully generated the DKG key, the private key will be stored in database.
```
durkmurder, line 42:
```suggestion
	// 1. if a node successfully generated the DKG key, the private key will be stored in database.
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	//    2. if a node failed to generate the DKG key, we will save a record in database to indicate this
```
durkmurder, line 43:
```suggestion
	// 2. if a node failed to generate the DKG key, we will save a record in database to indicate this
```

module/signature/random_beacon_signer_store.go
durkmurder: and note about concurrent access is desirable

module/signature/signer_relic.go
```diff
@@ -0,0 +1,45 @@
...
+
+package signature
+
+import (
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/crypto/hash"
+	"github.com/onflow/flow-go/module"
+)
+
+type RandomBeaconSigner struct {
```
durkmurder, line 11: minimal godoc, does it implement some interface?

module/signature/single.go
```diff
@@ -0,0 +1,25 @@
...
+
+package signature
+
+import (
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/crypto/hash"
+	"github.com/onflow/flow-go/module"
+)
+
+type SingleSigner struct {
```
durkmurder, line 11: minimal godoc, does this implement `MsgSigner` interface?

durkmurder (approved):
Thanks for great stuff, didn't find nothing major, mostly stylistic comments.

consensus/hotstuff/verification/combined_verifier_v2.go
zhangchiqing: Good catch @durkmurder 

Also a question for @tarakby:

Say there are `A,B,C,D,E,F,G` nodes, nodeA completed the DKG, but only node A failed to generate the DKG private key, all other nodes successfully generated their DKG private key. Then does node C who succeeded, know that A failed? or node C will have a special public key for nodeA but no one knows its private key?

The reason I'm asking is that, if node C knows nodeA has succeeded DKG or not, then it can reject votes proposals or votes from node A that doesn't have random beacon key share.

consensus/hotstuff/verification/combined_signer_v2.go
zhangchiqing: Since leader can sign blocks in a stateless manner, the combined signer is for consensus nodes to produce both signatures.

In V2, it makes sense to be called "Combined" signer, because we combine both staking sig and random beacon sig. And there is a "Single" signer, which is for collection cluster, who only sign staking sig always.

In V3, consensus leader only sign one sig, either staking and random beacon, yes, it's a bit weird to continue calling it "Combined" signer, since we are no longer combining any sig. But should be rename "Combined" signer to "Single" signer. Because "Single" signer is for the collection cluster. We might need to find a better name for V3...

storage/dkg_keys.go
```diff
@@ -5,6 +5,16 @@ import (
...
-	RetrieveMyDKGPrivateInfo(epochCounter uint64) (*dkg.DKGParticipantPriv, error)
+
+	// insert a record in database to indicate that when DKG was completed, the node failed DKG, there is no DKG key generated.
+	InsertNoDKGPrivateInfo(epochCounter uint64) error
+
+	// receive the DKG key for the given Epoch, it returns:
+	// - (key, true, nil) when DKG was completed, and a DKG private key was saved
+	// - (nil, false, nil) when DKG was completed, but no DKG private key was saved
+	// - (nil, false, storage.ErrNotFound) no DKG private key is found in database, it's unknown whether a DKG key was generated
+	// - (nil, false, error) for other exceptions
```
tarakby, line 18:
```suggestion
	// look up the random beacon private key for the given Epoch, it returns:
	// - (key, true, nil) when DKG succeeded locally ( DKG completed locally, and a non-nil private key was saved)
	// - (nil, false, nil) when DKG completed locally, but no non-nil private key was saved
	// - (nil, false, storage.ErrNotFound) no private key is found in database
	// - (nil, false, error) for other exceptions
```

storage/dkg_keys.go
```diff
@@ -5,6 +5,16 @@ import (
 )
 
 type DKGKeys interface {
+	// insert the DKG private key to database when DKG was completed, and a DKG private key was successfully generated.
 	InsertMyDKGPrivateInfo(epochCounter uint64, key *dkg.DKGParticipantPriv) error
-	RetrieveMyDKGPrivateInfo(epochCounter uint64) (*dkg.DKGParticipantPriv, error)
+
+	// insert a record in database to indicate that when DKG was completed, the node failed DKG, there is no DKG key generated.
```
tarakby, line 11: I didn't get the meaning of the description, is it possible to rephrase it?
```suggestion
	// insert a record in database to indicate that when DKG was completed, the node failed DKG, there is no DKG key generated.
```

storage/dkg_keys.go
```diff
@@ -5,6 +5,16 @@ import (
 )
 
 type DKGKeys interface {
+	// insert the DKG private key to database when DKG was completed, and a DKG private key was successfully generated.
```
tarakby, line 8:
```suggestion
	// insert the random beacon private key to database when DKG didn't fail locally (the node completed DKG  and a private key was successfully generated)
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+
+func NewEpochAwareRandomBeaconSignerStore(epochLookup module.EpochLookup, keys storage.DKGKeys) *EpochAwareRandomBeaconSignerStore {
+	return &EpochAwareRandomBeaconSignerStore{
+		epochLookup: epochLookup,
+		keys:        keys,
+		signers:     make(map[uint64]module.MsgSigner),
+	}
+}
+
+// GetSigner returns the threshold-signer for signing objects at a
```
tarakby, line 26: Since we've decided to go for a stateless signature, it doesn't make sense to speak about threshold signing here. From the perspective of a non-leader node, it is just generating a simple signature using a private key.

```suggestion
// GetSigner returns the random beacon signer for signing objects at a
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+// given view. The view determines the epoch, which determines the DKG private
+// key underlying the signer.
+// It returns:
+//  - (signer, nil) if DKG was completed in the epoch of the view, signer is not nil
```
tarakby, line 30:
```suggestion
//  - (signer, nil) if DKG succeeded locally in the epoch of the view, signer is not nil
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+//  - (nil, protocol.ErrEpochNotCommitted) if no epoch found for given view
+//  - (nil, DKGIncompleteError) if DKG was not completed in the epoch of the view
```
tarakby, line 32:
```suggestion
//  - (nil, DKGFailError) if DKG failed locally in the epoch of the view
```

module/signature/random_beacon_signer_store.go
tarakby:
```suggestion
	// 		1. if a node succeeded locally, the private key will be stored in database.
```

module/signature/random_beacon_signer_store.go
tarakby:
```suggestion
	//    2. if a node failed locally, we will save a record in database to indicate this
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	// fetching the epoch by view, if epoch is found, then DKG must have been completed
+	epoch, err := s.epochLookup.EpochForViewWithFallback(view)
+	if err != nil {
+		return nil, fmt.Errorf("could not get epoch by view %v: %w", view, err)
+	}
+
+	// when DKG has completed,
+	// 		1. if a node successfully generated the DKG key, the private key will be stored in database.
+	//    2. if a node failed to generate the DKG key, we will save a record in database to indicate this
+	//       node has no DKG key for this epoch
```
tarakby, line 44:
```suggestion
	//       node has no private key for this epoch
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	// with the epoch, we can lookup my DKG key for the epoch. There are 3 cases:
```
tarakby, line 45:
```suggestion
	// with the epoch, we can lookup my random beacon private key for the epoch. There are 3 cases:
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	// 1. DKG has completed, and the private key is stored in database, and we can retrieve it (happy path)
+	// 2. DKG has completed, but we failed to generate our private key, and we marked in the database
```
tarakby, line 47: The reason for this is that DKG can generate a private key but we would judge not to use it. 
```suggestion
	// 2. DKG has completed, but we failed it, and we marked in the database
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	// 		that there is no DKG key for this epoch
```
tarakby, line 48:
```suggestion
	// 		that there is no private key for this epoch
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+	// 3. DKG has completed, but we
```
tarakby, line 49: I think the sentence isn't completed :) 
```suggestion
	// 3. DKG has completed, but we
```

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+		// A nil signer means that we don't have a Random Beacon key for this epoch.
+		if signer == nil {
+			return nil, fmt.Errorf("did not complete DKG for epoch %v, at view %v: %w",
+				epoch, view, module.DKGIncompleteError)
+		}
+		return signer, nil
+	}
+
+	privDKGData, hasDKGKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
+	// this is an edge case where the epoch has determined, but the result of whether we have the DKG
```
tarakby, line 61:
```suggestion
	// this is an edge case where the epoch has determined, but the result of whether we have the
```

module/signature/signer_store.go
```diff
@@ -40,11 +41,13 @@ func (s *EpochAwareSignerStore) GetThresholdSigner(view uint64) (module.Threshol
...
 	}
 
-	privDKGData, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
+	privDKGData, hasDKGKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
 	if errors.Is(err, storage.ErrNotFound) {
 		signer = NewThresholdProvider(encoding.RandomBeaconTag, nil)
 	} else if err != nil {
 		return nil, fmt.Errorf("could not retrieve DKG private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
+	} else if !hasDKGKey {
+		return nil, fmt.Errorf("no DKG private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
```
tarakby, line 50:
```suggestion
		return nil, fmt.Errorf("no random beacon private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
```

module/signature/signer_store.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/onflow/flow-go/storage"
 )
 
+// TODO: to be replaced with EpochAwareRandomBeaconSignerStore
```
tarakby, line 12: 👌🏼

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module"
+	modulesig "github.com/onflow/flow-go/module/signature"
+)
+
+// CombinedVerifier is a verifier capable of verifying two signatures for each
+// verifying operation. The first type is a signature from an aggregating signer,
+// which verifies either the single or the aggregated signature. The second type is
+// a signature from a threshold signer, which verifies either the signature share or
+// the reconstructed threshold signature.
```
tarakby, line 22:
```suggestion
// CombinedVerifier is a verifier capable of verifying two signatures, one for each
// scheme. The first type is a signature from a staking signer,
// which verifies either a single or an aggregated signature. The second type is
// a signature from a random beacon signer, which verifies either the signature share or
// the reconstructed threshold signature.
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	beacon         hash.Hasher
+	keysAggregator *stakingKeysAggregator
+	merger         module.Merger
+}
+
+// NewCombinedVerifier creates a new combined verifier with the given dependencies.
+// - the hotstuff committee's state is used to retrieve the public keys for the staking signature;
+// - the staking tag is used to create hasher to verify staking signatures;
+// - the beacon tag is used to create hasher to verify random beacon signatures;
+// - the merger is used to combined & split staking & random beacon signatures;
```
tarakby, line 35:
```suggestion
// - the merger is used to combine and split staking and random beacon signatures;
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+func NewCombinedVerifierV2(committee hotstuff.Committee, stakingTag string, beaconTag string, merger module.Merger) *CombinedVerifierV2 {
+	return &CombinedVerifierV2{
+		committee:      committee,
+		staking:        crypto.NewBLSKMAC(stakingTag),
+		beacon:         crypto.NewBLSKMAC(beaconTag),
```
tarakby, line 40: I suggest to add more clarity
```suggestion
		stakingHasher:        crypto.NewBLSKMAC(stakingTag),
		beaconHasher:         crypto.NewBLSKMAC(beaconTag),
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return true, nil
+}
+
+// VerifyQC verifies the validity of a combined signature on a quorum certificate.
+func (c *CombinedVerifierV2) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) (bool, error) {
+
+	dkg, err := c.committee.DKG(block.BlockID)
+	if err != nil {
+		return false, fmt.Errorf("could not get dkg: %w", err)
```
tarakby, line 105:
```suggestion
		return false, fmt.Errorf("could not get dkg data: %w", err)
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	// split the aggregated staking & beacon signatures
+	stakingAggSig, beaconThresSig, err := c.merger.Split(sigData)
+	if err != nil {
+		return false, fmt.Errorf("could not split signature: %w", modulesig.ErrInvalidFormat)
+	}
+
+	msg := MakeVoteMessage(block.View, block.BlockID)
+	// TODO: verify if batch verification is faster
+
+	// verify the beacon signature first
```
tarakby, line 118:
```suggestion
	// verify the beacon signature first since it is faster to verify (no public key aggregation needed)
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if !beaconValid {
+		return false, nil
+	}
+	// verify the aggregated staking signature next (more costly)
+	// TODO: eventually VerifyMany will be a method of a stateful struct. The struct would
+	// hold the message, all the participants keys, the latest verification aggregated public key,
+	// as well as the latest list of signers (preferably a bit vector, using indices).
+	// VerifyMany would only take the signature and the new list of signers (a bit vector preferably)
+	// as inputs. A new struct needs to be used for each epoch since the list of participants is upadted.
```
tarakby, line 131: These are my comments, it's safe to delete them now.
```suggestion
	
```

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+
+	aggregatedKey, err := c.keysAggregator.aggregatedStakingKey(signers)
```
tarakby, line 133: This is using a module that is going to be deleted in V2, the module that should be used to aggregate public keys is module/signature.PublicKeyAggregator. We can make this change in another PR.
```suggestion
       // TODO: update to use module/signature.PublicKeyAggregator
	aggregatedKey, err := c.keysAggregator.aggregatedStakingKey(signers)
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+		SigData:  sigData,
+	}
+
+	return vote, nil
+}
+
+// genSigData generates the signature data for our local node for the given block.
+// It returns:
+//  - (stakingSig, nil) if there is no DKG private key  The sig is 48 bytes long
+//  - (stakingSig+randomBeaconSig, nil) if there is DKG private key. The sig is 96 bytes long
```
tarakby, line 95:
```suggestion
//  - (stakingSig, nil) if there is no random beacon private key  The sig is 48 bytes long
//  - (stakingSig+randomBeaconSig, nil) if there is a random beacon private key. The sig is 96 bytes long
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+// participate in the DKG.
+// The default behaviour of a node is to contribute to both the random beacon and
+// hotstuff, as it will yield higher rewards. Therefore, we always sign with the
+// random beacon key if it is available.
+// TODO: to be replaced by CombinedSignerV3 for mature V2 solution.
+// The difference between V2 and V3 is that V2 will sign 2 sigs, whereas
+// V3 only sign 1 sig.
+type CombinedSignerV2 struct {
+	staking           module.MsgSigner
+	beaconSignerStore module.RandomBeaconSignerStore
```
tarakby, line 28: As mentioned earlier, all the signers in `RandomBeaconSignerStore` are using duplicated hashers. It is better to create only one hasher in `CombinedSignerV2` in make the store a random beacon private key store instead.

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,123 @@
...
+// hotstuff. The latter is a useful fallback for a node that failed to properly
+// participate in the DKG.
+// The default behaviour of a node is to contribute to both the random beacon and
+// hotstuff, as it will yield higher rewards. Therefore, we always sign with the
+// random beacon key if it is available.
+// TODO: to be replaced by CombinedSignerV3 for mature V2 solution.
+// The difference between V2 and V3 is that V2 will sign 2 sigs, whereas
+// V3 only sign 1 sig.
+type CombinedSignerV2 struct {
+	staking           module.MsgSigner
```
tarakby, line 27: `MsgSigner` is also a thin layer around `Local` to re use its `Sign` method. I would just use `Local.Sign` straight away and avoid the extra module, but I have no strong opinion for this.

tarakby (commented):
Thanks for diving into these changes! I have left some comments below:

consensus/hotstuff/verification/combined_verifier_v2.go
tarakby: > if we have failed DKG and we don't have a private key, will we have a public key in such situation? 

If DKG failed globally then there will be no keys at all, but the epoch transition would have failed earlier before we get to Hotstuff of the new epoch (code above). If we get to this code, then there will be always public keys, regardless of the local private key result. 

> Then does node C who succeeded, know that A failed DKG?
C would have its local subjective view of how DKG ended for everyone (A and others). Though its local view may not correspond to what the objective reality is. That's why we do not rely on it. 
There is one case where all nodes can be sure that A failed, though that is not implemented yet (it can be useful to make everyone skip A's beacon signatures when they know A failed)

module/signer.go
```diff
@@ -29,6 +32,33 @@ type ThresholdSigner interface {
...
+
+// RandomBeaconSignerStore returns the signer for the given view, and it internally caches
+// signer objects by epoch. And epoch is determined by view.
+type RandomBeaconSignerStore interface {
+	// It returns:
+	//  - (signer, nil) if the node has beacon keys in the epoch of the view
+	//  - (nil, DKGIncompleteError) if the node doesn't have beacon keys in the epoch of the view
+	//  - (nil, error) if there is any exception
+	GetSigner(view uint64) (MsgSigner, error)
+}
```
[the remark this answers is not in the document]
zhangchiqing: I will just be one MsgSigner per epoch, rather than one MsgSigner per view. 

If we return the private key, who should hold the hasher? 

I'm planning to make a Factory method, such that it ensures the same hasher will be used by both SignerStore and Verifier. 

Something like:
```
// for shared code
randomBeaconFactory = makeFactoryWithTag(RandomBeaconTag) // create hasher

// for signer
randomBeaconSignerStore = randomBeaconFactory.CreateSigerStore()
...

// for verifier
randomBeaconVerifier = randomBeaconFactory.CreateVerifier()
```

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,30 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
...
 	defer tx.Discard()
-	return k.retrieveTx(epochCounter)(tx)
+	priv, err := k.retrieveTx(epochCounter)(tx)
+	if err != nil {
+		return nil, false, err
+	}
+	if priv == nil {
+		return nil, false, nil
+	}
+	return priv, true, nil
```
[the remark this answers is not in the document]
zhangchiqing: error is used for exception. 

If there is no exception, then there are two cases: has or not have the private key. 

I don't like to ask caller to differentiate these by checking `nil`, because if they forgot, and called any methods directly on the nil private key, the complier won't complain and will panic in runtime. Whereas, an extra bool return value will force you to handle it, otherwise won't compile.

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if !beaconValid {
+		return false, fmt.Errorf("invalid beacon sig")
+	}
+	return true, nil
+}
+
+// VerifyQC verifies the validity of a combined signature on a quorum certificate.
+func (c *CombinedVerifierV2) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) (bool, error) {
```
AlexHentschel, line 101: I think this method is not compatible with a QC from version V2: it still uses the old convention. Can we please update it, or stub it out? Having an unclear mixture of updated and legacy code makes it super hard to review.

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// TODO: to be replaced by packer.Unpack method
+	// split the aggregated staking & beacon signatures
+	stakingAggSig, beaconThresSig, err := c.merger.Split(sigData)
+	if err != nil {
+		return false, fmt.Errorf("could not split signature: %w", modulesig.ErrInvalidFormat)
+	}
+
+	msg := MakeVoteMessage(block.View, block.BlockID)
+	// TODO: verify if batch verification is faster
```
AlexHentschel, line 116: this can be removed
```suggestion
```
(we already do batch verification of the aggregated staking sig)

AlexHentschel (commented):
Didn't get through the review. 

I have some high-level comments regarding the code structure:
* Overall, I feel we have an unclear mixture of code for signing:
   *  some code is in `consensus/hotstuff/signature`
   * other code is in `consensus/hotstuff/verification`, which I find very unintuitive 
* I think the `EpochAwareRandomBeaconSignerStore` should go into `consensus/hotstuff/signature`. We explicitly tied random beacon signing to hotStuff. Therefore, I think it is fair for the code to live there. The current location `module/signature/random_beacon_signer_store.go` has various problems
   * Components in the `module/signature/` are generally used throughout the code base in a concurrent setting. Therefore, implementations should be concurrency safe; but `EpochAwareRandomBeaconSignerStore` is not.
   * I am concerned about the performance of your implementation, because `EpochAwareRandomBeaconSignerStore` always performs data base lookups of the Epoch without any caching. Utilizing knowledge how the `EpochAwareRandomBeaconSignerStore` in HotStuff is used would allow for a much more efficient (and potentially simpler) implementation. Also for this reason it makes sense to put `EpochAwareRandomBeaconSignerStore` in the context of HotStuff. 
  
  for more details, please see [my comment below](https://github.com/onflow/flow-go/pull/1194#discussion_r745389305)

* Lastly, I like [Tarak's suggestion](https://github.com/onflow/flow-go/pull/1194#discussion_r745270065) to directly work with the `module.Local` and to maintain the hashers in the signer objects. Thereby, we don't have to worry about concurrency as the EventHandler (which triggers the signing) is single threaded.

module/signature/random_beacon_signer_store.go
```diff
@@ -0,0 +1,109 @@
...
+// given view. The view determines the epoch, which determines the DKG private
+// key underlying the signer.
+// It returns:
+//  - (signer, nil) if DKG was completed in the epoch of the view, signer is not nil
+//  - (nil, protocol.ErrEpochNotCommitted) if no epoch found for given view
+//  - (nil, DKGIncompleteError) if DKG was not completed in the epoch of the view
+//  - (nil, error) if there is any exception
+func (s *EpochAwareRandomBeaconSignerStore) GetSigner(view uint64) (module.MsgSigner, error) {
+	// fetching the epoch by view, if epoch is found, then DKG must have been completed
+	epoch, err := s.epochLookup.EpochForViewWithFallback(view)
```
AlexHentschel, line 36: I am worried about the performance here. To the best of my understanding, this _always_ hits the data base which will impact the hot path of consensus. 

We can utilize some knowledge about the signing process to optimize the implementation:
* a consensus replica only signs (at most) once for its current view. Hence, `view` increases strictly monotonously.  
* Chances are overwhelmingly high that I am _exactly_ in the same epoch as during the last call `GetSigner`, as Epoch changes are relatively rare. 

Therefore, it is totally sufficient to cache 
* the _signer_ from the last call, along with some epoch details:
   * first view of epoch and last view of epoch
   * thereby, I can easily determine whether I am still in the same epoch
   * if I get a view _before_ the current epoch, I can just error (unexpected critical)
   * If I am in a view _beyond_ the view of the current epoch, I load a new signer and cache it.

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -0,0 +1,142 @@
...
+// verifying operation. The first type is a signature from an aggregating signer,
+// which verifies either the single or the aggregated signature. The second type is
+// a signature from a threshold signer, which verifies either the signature share or
+// the reconstructed threshold signature.
+type CombinedVerifierV2 struct {
+	committee      hotstuff.Committee
+	staking        hash.Hasher
+	beacon         hash.Hasher
+	keysAggregator *stakingKeysAggregator
+	merger         module.Merger
```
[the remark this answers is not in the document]
durkmurder: I believe we can leave merger as it is, I have addressed it in my PR which is based on this one. https://github.com/onflow/flow-go/issues/1601

storage/badger/dkg_keys.go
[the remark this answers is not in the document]
AlexHentschel: @tarakby this pattern is similar to go's map lookup 
```golang
m := map[string]float64{"pi": 3.14}
v, found := m["pi"] // v == 3.14  found == true
```
it transforms the `nil` check into a boolean check ... semantically different ... helps people who don't like `nil`s 😉

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,30 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
 	}
 }
 
+// InsertMyDKGPrivateInfo insert the DKG private key to database when DKG was completed, and a DKG private key was successfully generated.
 func (k *DKGKeys) InsertMyDKGPrivateInfo(epochCounter uint64, info *dkg.DKGParticipantPriv) error {
```
AlexHentschel, line 70: I know you only added to this implementation. Nevertheless, it would be great if you could document possible error returns here as well as in the interface https://github.com/onflow/flow-go/blob/df5e576c4203cbcd559adc4b47a5466a8f78bbaa/storage/dkg_keys.go#L8-L9
that would be much appreciated 🙏 🙇

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,30 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
...
 	return operation.RetryOnConflictTx(k.db, transaction.Update, k.storeTx(epochCounter, info))
 }
 
-func (k *DKGKeys) RetrieveMyDKGPrivateInfo(epochCounter uint64) (*dkg.DKGParticipantPriv, error) {
+// InsertNoDKGPrivateInfo insert a record in database to indicate that when DKG was completed, the node failed DKG, there is no DKG key generated.
+func (k *DKGKeys) InsertNoDKGPrivateInfo(epochCounter uint64) error {
```
AlexHentschel, line 75: 🙏 please document possible error returns during normal operation here as well as in the interface 🙇   https://github.com/onflow/flow-go/blob/df5e576c4203cbcd559adc4b47a5466a8f78bbaa/storage/dkg_keys.go#L11-L12

storage/dkg_keys.go
```diff
@@ -5,6 +5,16 @@ import (
...
 	InsertMyDKGPrivateInfo(epochCounter uint64, key *dkg.DKGParticipantPriv) error
-	RetrieveMyDKGPrivateInfo(epochCounter uint64) (*dkg.DKGParticipantPriv, error)
+
+	// insert a record in database to indicate that when DKG was completed, the node failed DKG, there is no DKG key generated.
+	InsertNoDKGPrivateInfo(epochCounter uint64) error
```
AlexHentschel, line 12: 😢 missing possible error returns during normal operation here as well as in the interface.

consensus/hotstuff/verification/combined_verifier_v2.go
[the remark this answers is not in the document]
AlexHentschel: 🎉

module/signer.go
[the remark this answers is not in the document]
tarakby: You are right that one extra hasher per epoch is a negligible overhead. 
My suggestion is more related to the clarity of the design : a store for different data per epoch can be limited to what is really different every epoch. In this case it is just the private key.

module/signature/single.go
zhangchiqing: Good point, updated

module/signer.go
```diff
@@ -29,6 +32,33 @@ type ThresholdSigner interface {
...
+	// It returns:
+	//  - (signer, nil) if the node has beacon keys in the epoch of the view
+	//  - (nil, DKGIncompleteError) if the node doesn't have beacon keys in the epoch of the view
+	//  - (nil, error) if there is any exception
 	GetThresholdSigner(view uint64) (ThresholdSigner, error)
 }
+
+// MsgSigner signs a given message and produces a signature
+// TODO: could be renamed to Signer once the original Signer is replaced with MsgSigner
+type MsgSigner interface {
```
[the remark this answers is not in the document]
tarakby: The specific hasher we're using for BLS is concurrent safe [It's sad to see you leaving, but I'm sure ](https://github.com/onflow/flow-go/blob/master/crypto/hash/kmac.go#L147-L150). 
This doesn't apply to all hashers though. If we're gonna take advantage fo the hasher concurrency here, it would be good to mention it in a comment.

consensus/hotstuff/signature/randombeacon_signer_store.go
```diff
@@ -0,0 +1,101 @@
...
+	// 2. DKG has completed, but we failed it, and we marked in the database
+	// 		that there is no private key for this epoch (unhappy path)
+	// 3. DKG has completed, but for some reason we don't find the private key in the database (exception)
+	// 4. DKG was not completed (fatal error, we should not run into with EECC, becauwe we still stay at
+	// 		the current Epoch where DKG has completed.
+	key, ok := s.readKey(epoch)
+	if ok {
+		// A nil key means that we don't have a Random Beacon key for this epoch.
+		if key == nil {
+			return nil, fmt.Errorf("did not complete DKG for epoch %v, at view %v: %w",
```
tarakby, line 61:
```suggestion
			return nil, fmt.Errorf("DKG for epoch %v failed, at view %v: %w",
```

consensus/hotstuff/signature/randombeacon_signer_store.go
```diff
@@ -0,0 +1,101 @@
...
+				epoch, view, module.DKGFailError)
+		}
+		return key, nil
+	}
+
+	privBeaconKeyData, hasRandomBeaconKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve DKG private key for epoch counter %v, at view %v, err: %w",
```
tarakby, line 69:
```suggestion
		return nil, fmt.Errorf("could not retrieve random beacon private key for epoch counter %v, at view %v, err: %w",
```

consensus/hotstuff/signature/randombeacon_signer_store.go
```diff
@@ -0,0 +1,101 @@
...
+			epoch, view, err)
+	}
+
+	// if DKG was not completed, there will be no DKG private key, since this fact
+	// never change, we can cache a nil signer for this epoch, so that we this function
+	// is called again for the same epoch, we don't need to query database.
+	if !hasRandomBeaconKey {
+		s.writeKey(epoch, nil)
+		return nil, fmt.Errorf("didn't complete DKG for epoch %v, at view %v: %w",
```
tarakby, line 78:
```suggestion
		return nil, fmt.Errorf("DKG for epoch %v failed, at view %v: %w",
```

consensus/hotstuff/signature/randombeacon_signer_store.go
```diff
@@ -0,0 +1,101 @@
...
+			epoch, view, module.DKGFailError)
+	}
+
+	// DKG was completed and a random beacon key is available,
```
tarakby, line 82:
```suggestion
	// DKG succeeded and a random beacon key is available,
```

consensus/hotstuff/signature/randombeacon_signer_store.go
```diff
@@ -0,0 +1,101 @@
...
+		return key, nil
+	}
+
+	privBeaconKeyData, hasRandomBeaconKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve DKG private key for epoch counter %v, at view %v, err: %w",
+			epoch, view, err)
+	}
+
+	// if DKG was not completed, there will be no DKG private key, since this fact
```
tarakby, line 73:
```suggestion
	// if DKG failed, there will be no random beacon private key, since this fact
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,128 @@
...
+		SigData:  sigData,
+	}
+
+	return vote, nil
+}
+
+// genSigData generates the signature data for our local node for the given block.
+// It returns:
+//  - (stakingSig, nil) if there is no random beacon private key  The sig is 48 bytes long
+//  - (stakingSig+randomBeaconSig, nil) if there is a random beacon private key. The sig is 96 bytes long
```
tarakby, line 100: It is better not to reference signature constants that are not agnostic of the signature algorithm/curve/serialization used.
```suggestion
//  - (stakingSig, nil) if there is no random beacon private key.
//  - (stakingSig+randomBeaconSig, nil) if there is a random beacon private key.
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,128 @@
...
+	beaconKey, err := c.beaconKeyStore.ByView(block.View)
+	if err != nil {
+		if errors.Is(err, module.DKGFailError) {
+			return stakingSig, nil
+		}
+		return nil, fmt.Errorf("could not get threshold signer for view %d: %w", block.View, err)
+	}
+
+	// if the node is a Random Beacon participant and has completed its DKG, then using the random beacon key
+	// to sign the block
```
tarakby, line 121:
```suggestion
	// if the node is a Random Beacon participant and has succeeded DKG, then using the random beacon key
	// to sign the block
```

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not generate staking signature: %w", err)
+	}
+
+	beaconKey, err := c.beaconKeyStore.ByView(block.View)
+	if err != nil {
+		if errors.Is(err, module.DKGFailError) {
+			return stakingSig, nil
+		}
+		return nil, fmt.Errorf("could not get threshold signer for view %d: %w", block.View, err)
```
tarakby, line 117:
```suggestion
		return nil, fmt.Errorf("could not get random beacon private key for view %d: %w", block.View, err)
```

consensus/hotstuff/verification/combined_signer_v3.go
```diff
@@ -0,0 +1,127 @@
...
+// genSigData generates the signature data for our local node for the given block.
+func (c *CombinedSignerV3) genSigData(block *model.Block) ([]byte, error) {
+
+	// create the message to be signed and generate signatures
+	msg := MakeVoteMessage(block.View, block.BlockID)
+
+	beaconKey, err := c.beaconKeyStore.ByView(block.View)
+	if err != nil {
+		if errors.Is(err, module.DKGFailError) {
+			// if the node didn't complete DKG, then using the staking key to sign the block as a
```
tarakby, line 107:
```suggestion
			// if the node failed DKG, then using the staking key to sign the block as a
```

consensus/hotstuff/verification/combined_signer_v3.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("could not generate staking signature: %w", err)
+			}
+
+			return signature.EncodeSingleSig(hotstuff.SigTypeStaking, stakingSig), nil
+		}
+		return nil, fmt.Errorf("could not get threshold signer for view %d: %w", block.View, err)
+	}
+
+	// if the node is a Random Beacon participant and has completed its DKG, then using the random beacon key
```
tarakby, line 119:
```suggestion
	// if the node is a Random Beacon participant and has succeeded DKG, then using the random beacon key
```

consensus/hotstuff/verification/combined_signer_v3.go
```diff
@@ -0,0 +1,127 @@
...
+			// if the node didn't complete DKG, then using the staking key to sign the block as a
+			// fallback
+			stakingSig, err := c.staking.Sign(msg, c.stakingHasher)
+			if err != nil {
+				return nil, fmt.Errorf("could not generate staking signature: %w", err)
+			}
+
+			return signature.EncodeSingleSig(hotstuff.SigTypeStaking, stakingSig), nil
+		}
+		return nil, fmt.Errorf("could not get threshold signer for view %d: %w", block.View, err)
```
tarakby, line 116:
```suggestion
		return nil, fmt.Errorf("could not get random beacon private key for view %d: %w", block.View, err)
```

module/signature/signer_store.go
```diff
@@ -40,11 +41,13 @@ func (s *EpochAwareSignerStore) GetThresholdSigner(view uint64) (module.Threshol
 		return signer, nil
 	}
 
-	privDKGData, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
+	privDKGData, hasDKGKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
```
tarakby, line 44:
```suggestion
	privDKGData, hasBeaconKey, err := s.keys.RetrieveMyDKGPrivateInfo(epoch)
```

module/signature/signer_store.go
```diff
@@ -40,11 +41,13 @@ func (s *EpochAwareSignerStore) GetThresholdSigner(view uint64) (module.Threshol
...
 	if errors.Is(err, storage.ErrNotFound) {
 		signer = NewThresholdProvider(encoding.RandomBeaconTag, nil)
 	} else if err != nil {
 		return nil, fmt.Errorf("could not retrieve DKG private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
+	} else if !hasDKGKey {
```
tarakby, line 49:
```suggestion
	} else if !hasBeaconKey {
```

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,31 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
 	}
 }
 
+// InsertMyDKGPrivateInfo insert the DKG private key to database when DKG was completed, and a DKG private key was successfully generated.
```
tarakby, line 69:
```suggestion
// InsertMyDKGPrivateInfo insert the private key to database when DKG succeeded.
```

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,31 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
...
 func (k *DKGKeys) InsertMyDKGPrivateInfo(epochCounter uint64, info *dkg.DKGParticipantPriv) error {
 	return operation.RetryOnConflictTx(k.db, transaction.Update, k.storeTx(epochCounter, info))
 }
 
-func (k *DKGKeys) RetrieveMyDKGPrivateInfo(epochCounter uint64) (*dkg.DKGParticipantPriv, error) {
+// InsertNoDKGPrivateInfo insert a record in database to indicate that when DKG was completed,
+// the node failed DKG, there is no DKG key generated.
```
tarakby, line 75: I think there is a typo in the sentence :) Is it possible to rephrase it please?
We can also:
- use DKG fail and success, instead of completed and not completed. 
- random beacon private key or just private key, instead of DKG key.

storage/badger/dkg_keys.go
```diff
@@ -66,12 +66,31 @@ func (k *DKGKeys) retrieveTx(epochCounter uint64) func(tx *badger.Txn) (*dkg.DKG
...
+func (k *DKGKeys) InsertNoDKGPrivateInfo(epochCounter uint64) error {
+	return operation.RetryOnConflictTx(k.db, transaction.Update, k.storeTx(epochCounter, nil))
+}
+
+// receive the DKG key for the given Epoch, it returns:
+// - (key, true, nil) when DKG was completed, and a DKG private key was saved
+// - (nil, false, nil) when DKG was completed, but no DKG private key was saved
+// - (nil, false, storage.ErrNotFound) no DKG private key is found in database, it's unknown whether a DKG key was generated
+// - (nil, false, error) for other exceptions
```
tarakby, line 84:
```suggestion
// receive the random beacon key for the given Epoch, it returns:
// - (key, true, nil) when DKG succeeded, and a private key was saved
// - (nil, false, nil) when DKG failed (no private key was saved)
// - (nil, false, storage.ErrNotFound) no private key is found in database
// - (nil, false, error) for other exceptions
```

tarakby (approved):
Thanks for making the long changes 🙏🏼 

I have left more doc comments to keep our doc consistent. It mainly addresses 2 points:
 - Our code logic is based on the local success or failure of DKG, rather than its completion or not (a node can complete DKG but fail). So I prefer to mention success and failure in the comments.
 - DKG keys doesn't make much sense. I prefer to use random beacon key / random beacon private key / beacon key/ DKG private output..

storage/dkg_keys.go
AlexHentschel: 👍 Just a quick comment like: "During normal operations, no error returns are expected." would clarify

consensus/hotstuff/verification/combined_signer_v3.go
```diff
@@ -0,0 +1,127 @@
...
+// still contribute only to consensus (as fallback).
+// TODO: to be replaced by CombinedSignerV3 for mature V2 solution.
+// The difference between V2 and V3 is that V2 will sign 2 sigs, whereas
+// V3 only sign 1 sig.
+type CombinedSignerV3 struct {
+	staking        module.Local
+	stakingHasher  hash.Hasher
+	beaconKeyStore module.RandomBeaconKeyStore
+	beaconHasher   hash.Hasher
+	signerID       flow.Identifier
```
AlexHentschel, line 32: Same here: I think we could just use `module.Local.NodeID()` instead of explicityly storing the node ID again (?)

AlexHentschel (approved):
Thank you for the hard work, the many iterations and your patience with my PR reviews. This looks great. Just a couple _minor, optional_ stylistic comments. Thanks for the detailed documentation, including the error returns. Great work! 

I am very happy with this PR (with or without my additional comments being implemented). Please feel free to merge. 🚢

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -0,0 +1,128 @@
+package verification
```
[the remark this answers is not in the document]
zhangchiqing: That makes sense. Will try it in another PR, since there is other PRs depending on this, don't want to create too many conflicts.

zhangchiqing:
bors merge

PR #1195, opened 2021-08-25
[WIP] adding compact encoding to account keys

ramtinms:
This is currently experimental, to see if we could store account keys in a more compact and make decoding faster 
```

Compact (our implementation) is about 30-40% faster than RLP in decoding time and takes about the same amount of bytes


cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkAccountPublicKeyDecode_Compact-8   	     100	      1572 ns/op	     886 B/op	      13 allocs/op   78 bytes
BenchmarkAccountPublicKeyDecode_RLP-8       	     100	      2410 ns/op	    1087 B/op	      16 allocs/op   75 bytes
BenchmarkAccountPublicKeyDecode_JSON-8      	     100	      5927 ns/op	    1250 B/op	      19 allocs/op   158 bytes 
```


model/flow/account.go
```diff
@@ -76,6 +78,120 @@ func (a *AccountPublicKey) UnmarshalJSON(data []byte) error {
...
+		return fmt.Errorf("cannot compact decode public key: data size is too small to be decodable (expected: at least %d, got: %d)", minDataSize, dataSize)
+	}
+
+	// first 5 bits were encoding version so we have already processed them
+
+	// next bit is revoked flag
+	a.Revoked = (data[0] & byte(4)) > 0 // (mask of 00000100 and equality check (might optimize in the future))
+
+	// the last 2 remaining bits of first byte are the most significant bits of key weight
+	a.Weight = int(uint16(data[0]&byte(3))<<8 + uint16(data[1]))
```
tarakby, line 123: Concatenating with bit wise operations is faster than arithmetic ones. 
```suggestion
	a.Weight = int(uint16(data[0]&byte(3))<<8 | uint16(data[1]))
```

model/flow/account.go
```diff
@@ -76,6 +78,120 @@ func (a *AccountPublicKey) UnmarshalJSON(data []byte) error {
...
+	return nil
+}
+
+func (a AccountPublicKey) compactEncodeV0() ([]byte, error) {
+	var encoded [14]byte
+
+	// first 5 bits	are encoding version
+	// you might set it initial int according to the version
+	// e.g. uint8(8) for version 1, uint8(16) for v2, uint8(24) for v3, ...
+	firstByte := uint8(0)
```
tarakby, line 160:
```suggestion
	var firstByte uint8
```

huitseeker (commented):
ICYMI, @ramtinms, we have just introduced (bi-directional) compressed encodings for public keys in https://github.com/onflow/flow-go/pull/1129

On ECDSA-derived keys (ECDSAP-2256, ECDSASecp256k1), this allows you to save a key on 33 bytes instead of 64.

I know compression wasn't necessarily part of the scope of this PR, but maybe this helps?
/cc @tarakby

tarakby (approved):
Nice, the new encoding is faster and uses less allocations, so it makes sense to switch to it IMO. 

I also think moving to compressed pubic keys is good change, we discussed it in TPM a while ago (though we can add it in a separate PR):
- a public key would use 33 bytes Vs 64 currently.
- the encoding time is not impacted, however the decoding time will get slower.
- the team prioritizes storage gain over execution time in this case

ramtinms:
this has been stale and I'm still not fully sure if it is worth the effort right now, closing this PR for now but going to revisit this with @fxamacker and @tarakby at a later time when we do decide to do more optimizations in that regard

PR #1198, opened 2021-08-25
[Consensus Voting] Double vote detection

durkmurder:
### Context 

https://github.com/dapperlabs/flow-go/issues/5781
https://github.com/dapperlabs/flow-go/issues/5740#issuecomment-904670491

With current iteration of consensus voting v2 we have implemented `VoteCollector` based on idea that it collects votes for one block. This makes hard to keep track of all cases of double voting.

Some great points were raised by by @huitseeker in https://github.com/dapperlabs/flow-go/issues/5740#issuecomment-904670491 and https://github.com/onflow/flow-go/pull/1140#pullrequestreview-736454870. 

I would like to propose to change our design to address vote collecting based on views instead of block proposals. Addressing by views makes easy to track double voting cases and makes things very clear and logical. Pruning and indexing becomes a lot easier compared to [proposal](https://github.com/onflow/flow-go/pull/1161). 

### What was changed

Previously we had one vote collector for a tuple `(view, blockID)` now we will have one vote collector for each `view` and votes with different `blockID` but same `view` will be handled by one vote collector. It's up to vote collector how to deal with case of double voting. 

When we have received block proposal we can clearly make decision about double voting. I am not so sure about case when we haven't received block proposal. Based on https://github.com/dapperlabs/flow-go/issues/5740#issuecomment-904670491 we can still do some checks but can't be sure unless we have received block proposal. 

### TODO: 

Once this PR receives enough approvals there are still some things to handle:

- ~~Add extra index for `CachingVoteCollector`, maybe inside of `PendingVotes` to track index votes by `blockID`.~~
- ~~Fix current tests~~
- Update logic of `CachingVoteCollector` with regards to double voting.


consensus/hotstuff/vote_collector.go
```diff
@@ -41,21 +41,6 @@ func (ps VoteCollectorStatus) String() string {
...
-	// The implementation only allows the transitions
-	//         CachingVotes   -> VerifyingVotes
-	//         CachingVotes   -> Invalid
-	//         VerifyingVotes -> Invalid
-	// Error returns:
-	// * nil if the state transition was successfully executed
-	// * ErrDifferentCollectorState if the VoteCollector's state is different than expectedCurrentStatus
-	// * ErrInvalidCollectorStateTransition if the given state transition is impossible
-	// * all other errors are unexpected and potential symptoms of internal bugs or state corruption (fatal)
-	ChangeProcessingStatus(expectedValue, newValue VoteCollectorStatus) error
```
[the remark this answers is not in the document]
AlexHentschel: Lets consider the following situation:
* In view 100, Alice is primary and proposes block `B1`.
* In addition, evil Eve also proposes a block, `B2`, even though it is not her turn. Nevertheless, Eve signs her proposal properly.

From the perspective of our node:
1. the node receives votes for `B1` and `B2` (without having seen either blocks)
2. block `B1` arrives
3. block `B2` arrives 
    * at this point, the HotStuff state machine would detect that `B2` is an invalid block:  Screen Shot 2021-08-27 at 4 11 38 PM (full state machine [here](https://drive.google.com/file/d/1la4jxyaEJJfip7NCWBM9YBTz6PK4-N9e/view?usp=sharing))
    * The EventHandler therefore would call https://github.com/onflow/flow-go/blob/d43a18fc733de90ddea8b81acaa2443d702624e8/consensus/hotstuff/vote_aggregator_v2.go#L48-L50

In the previous version, we would have had independent `VoteCollector`s for `B1` and `B2` and an explicit state transition function
* for `B1` we could transition the sate from `CachingVoteCollector` to `VerifyingVoteCollector`
* for `B2` we could transition the sate from `CachingVoteCollector` to `InvalidVoteCollector`

With the revisions proposed in this PR, I am not clear on how we would pass the information along the the `VoteCollector` that the votes for `B1` are fine to process, but the votes for `B2` should be slashed.

durkmurder:
I have updated already implemented logic, added some extra checks, extended documentation. This I believe is a good starting point for tracking double votes. Test coverage and more updates to this will be done in subsequent PRs.
Request another review @AlexHentschel @huitseeker @zhangchiqing

consensus/hotstuff/votecollector/statemachine_test.go
```diff
@@ -0,0 +1,200 @@
...
+		firstVote := unittest.VoteForBlockFixture(block)
+		processor.On("Process", firstVote).Return(nil).Once()
+		err := s.collector.AddVote(firstVote)
+		require.NoError(t, err)
+
+		secondVote := unittest.VoteFixture(func(vote *model.Vote) {
+			vote.View = firstVote.View
+			vote.SignerID = firstVote.SignerID
+		})
+
```
AlexHentschel, line 129:
```suggestion
		}) // voted blockID is randomly sampled, i.e. it will be different from firstVote
```

consensus/hotstuff/votecollector/statemachine.go
```diff
@@ -55,129 +73,174 @@ func NewStateMachine(base CollectionBase) *StateMachine {
...
 	if ok {
 		return blockSigner.CreateVote(block)
 	}
 	return nil, ErrDifferentCollectorState
 }
 
-func (m *StateMachine) AddVote(vote *model.Vote) error {
+func (m *VoteCollector) AddVote(vote *model.Vote) error {
+	// Cache vote
+	err := m.votesCache.AddVote(vote)
```
zhangchiqing, line 87: Do we still cache the vote even if we have received the block?

consensus/hotstuff/votecollector/vote_cache.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// Expected usage patter: During happy-path operations, the block arrives in a
+// timely manner. Hence, we expect that only a few votes are cached when a
+// consumer is registered. For the purpose of forensics, we might register a
+// consumer later, when already lots of votes are cached. However, this should
+// be a rare occurrence (we except moderate performance overhead in this case).
+func (vc *VotesCache) RegisterVoteConsumer(consumer VoteConsumer) {
+	if consumer == nil {
+		return
+	}
```
zhangchiqing, line 109: why would this case exist?

consensus/hotstuff/votecollector/combined_vote_processor.go
```diff
@@ -0,0 +1,163 @@
...
+	msig "github.com/onflow/flow-go/module/signature"
+)
+
+// CombinedVoteProcessorFactory generates CombinedVoteProcessor instances
+func CombinedVoteProcessorFactory(log zerolog.Logger, proposal *model.Proposal) (*CombinedVoteProcessor, error) {
+	processor := &CombinedVoteProcessor{
+		log:   log,
+		block: proposal.Block,
+	}
+	processor.done.Store(false)
```
zhangchiqing, line 23: is this call necessary?

consensus/hotstuff/votecollector/statemachine_test.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+	s.T().Run("add-incompatible-block-vote", func(t *testing.T) {
+		vote := unittest.VoteForBlockFixture(block, unittest.WithVoteView(s.view))
+		processor.On("Process", vote).Return(VoteForIncompatibleBlockError).Once()
+		err := s.collector.AddVote(vote)
+		// in case process returns VoteForIncompatibleBlockError we should silently ignore this error
+		require.NoError(t, err)
+		processor.AssertCalled(t, "Process", vote)
+	})
+}
```
AlexHentschel, line 178: we could also test that unexpected Processor errors are passed up the call stack:
```suggestion
	s.T().Run("unexpected-VoteProcessor-errors-are-passed-up", func(t *testing.T) {
		unexpectedError := errors.New("some unexpected error")
		vote := unittest.VoteForBlockFixture(block, unittest.WithVoteView(s.view))
		processor.On("Process", vote).Return(unexpectedError).Once()
		err := s.collector.AddVote(vote)
		require.ErrorIs(t, err, unexpectedError)
	})
}
```

consensus/hotstuff/votecollector/statemachine_test.go
```diff
@@ -0,0 +1,200 @@
...
+		require.NoError(s.T(), s.collector.AddVote(vote))
+	}
+
+	err := s.collector.ProcessBlock(proposal)
+	require.NoError(s.T(), err)
+
+	time.Sleep(100 * time.Millisecond)
+
+	processor.AssertExpectations(s.T())
+}
```
AlexHentschel, line 200: another test we could include is:

```golang
// Test_VoteProcessorErrorPropagation verifies that unexpected errors from the `VoteProcessor`
// are propagated up the call stack (potentially wrapped), but are not replaced.
func (s *StateMachineTestSuite) Test_VoteProcessorErrorPropagation() {
	block := helper.MakeBlock(s.T(), helper.WithBlockView(s.view))
	processor := s.prepareMockedProcessor(block)

	unexpectedError := errors.New("some unexpected error")
	vote := unittest.VoteForBlockFixture(block, unittest.WithVoteView(s.view))
	processor.On("Process", vote).Return(unexpectedError).Once()
	err := s.collector.AddVote(vote)
	require.ErrorIs(s.T(), err, unexpectedError)
}
```

consensus/hotstuff/votecollector/vote_cache_test.go
```diff
@@ -0,0 +1,92 @@
...
+		require.NoError(t, cache.AddVote(vote))
+	}
+
+	consumedVotes := make([]*model.Vote, 0)
+	voteConsumer := func(vote *model.Vote) {
+		consumedVotes = append(consumedVotes, vote)
+	}
+
+	// registering vote consumer has to fill consumedVotes using callback
+	cache.RegisterVoteConsumer(voteConsumer)
```
AlexHentschel, line 80: would suggest to test that all cached votes are fed into the consumer right away

```suggestion
	cache.RegisterVoteConsumer(voteConsumer)
	// all cached votes should be fed into the consumer right away
	require.Equal(t, expectedVotes, consumedVotes)
```

AlexHentschel (approved):
very nice. Thanks for the detailed documentation of all the tests. It makes it very easy to read the test.

consensus/hotstuff/votecollector/statemachine.go
AlexHentschel: We want to keep _all_ received votes, so we can do forensics later (and scan for double voting across all votes for this view).

consensus/hotstuff/votecollector/vote_cache.go
AlexHentschel: I originally put this in to avoid a panic when we try to feed a vote into the consumer later (in a different thread). 

But you are probably right: we generally don't check for nil values; so we probably should remove this check here too.

consensus/hotstuff/votecollector/combined_vote_processor.go
AlexHentschel: Note that `processor.done` is _not_ a pointer. Instead, to increase data locality, I included it directly. 
But I guess we could still do: 
```golang
	processor := &CombinedVoteProcessor{
		log:   log,
		block: proposal.Block,
		done: *atomic.NewBool(false),
	}
```
(?)

consensus/hotstuff/vote_collectors.go
```diff
@@ -1,25 +1,24 @@
...
 	//  -  (nil, false, error) if running into any exception creating the vote collector state machine
-	GetOrCreateCollector(view uint64, blockID flow.Identifier) (collector VoteCollector, created bool, err error)
+	// Expected error returns during normal operations:
+	//  * mempool.DecreasingPruningHeightError
+	GetOrCreateCollector(view uint64) (collector VoteCollector, created bool, err error)
 
-	// Prune the vote collectors whose view is below the given view
+	// PruneUpToView prunes the vote collectors whose view is below the given view.
+	// If `view` is smaller than the previous value, the previous value is kept
+	// and the sentinel mempool.DecreasingPruningHeightError is returned.
```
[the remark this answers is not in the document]
durkmurder: https://github.com/onflow/flow-go/issues/1267

PR #1200, opened 2021-08-25
Jwinkler2083233/stack allocated hasher

jwinkler2083233:
I'm measuring a 0.4% decrease in CPU usage, attributable to millions of fewer 'mallocgc' calls.


huitseeker:
As mentioned by @tarakby elsewhere, while the direction is good, exposing Sha3State is probably not the best way to do this.

huitseeker (changes requested):
As mentioned by @tarakby elsewhere, while the direction is good, exposing Sha3State is probably not the best way to do this.

jwinkler2083233:
closing because Tarak has a better way.

PR #1208, opened 2021-08-25
[Consensus Voting] Use DKG Signer

zhangchiqing:
In consensus cluster, when Block Producer and Voter signs a block, it might use the random beacon key to sign the block. Currently, we use a ThresholdSigner object to sig the block, but actually is not necessary. 

Since the signing a block can be done in a contextless manner, we can replace the ThresholdSigner object with a single module that is holding the private key for signing. And the ThresholdSigner object will be used by the VoteCollector to verify and store the vote signatures.

### DIFF
--- a/consensus/hotstuff/verification/combined_signer_v2.go
+++ b/consensus/hotstuff/verification/combined_signer_v2.go
@@ -1,0 +1,124 @@
+package verification
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


zhangchiqing:
close in favor of https://github.com/onflow/flow-go/pull/1194

PR #1209, opened 2021-08-26
[Consensus Voting] QC pack/unpack

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5742

This PR implements the consensus signature packer module to pack and unpack the block signature data for building and parsing quorum certificate.


consensus/hotstuff/signature.go
```diff
@@ -96,5 +96,8 @@ type Packer interface {
 
 	// blockID is the block that the aggregated sig is signed for
 	// sig is the aggregated signature data
+	// It returns:
+	//  - (sigData, nil) if successfully unpacked the signature data
+	//  - (nil, signature.ErrInvalidFormat) if failed to unpack the signature data
```
zhangchiqing, line 101: this sentinel error is useful for [event handler to handler errors from the validator](https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/validator/validator.go#L65-L67)

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,123 @@
...
 	return &ConsensusSigPackerImpl{
 		committees: committees,
+		encoder:    rlp.NewEncoder(),
 	}
 }
 
+// signatureData is a compact data type for encoding the block signature data
+type signatureData struct {
+	// bit-vector indicating type of signature for each signer.
+	// the order of each sig type matches the order of coresponding signer IDs
```
durkmurder, line 31:
```suggestion
	// the order of each sig type matches the order of corresponding signer IDs
```

durkmurder (approved):
Good stuff, great tests. Should we rename `packager.go` -> `packer.go` ? Tarak already did this in his PR, would say it makes sense.

AlexHentschel (approved):
Overall looks good

AlexHentschel:
Too the liberty to add some minor goDoc revisions: https://github.com/onflow/flow-go/pull/1209/commits/a3c2e6a157101bfafd32281a5acab7d895804db4

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,253 @@
...
+		return false, nil
+	}
+	if t == hotstuff.SigTypeRandomBeacon {
+		return true, nil
+	}
+	return false, signature.ErrInvalidFormat
+}
+
+// seralize the sig types into a compact format
+func seralizeToBytes(sigTypes []hotstuff.SigType) ([]byte, error) {
```
zhangchiqing, line 182: added seralizeToBytes and deseralizeFromBytes to compact the bytes that represents sigTypes 

@AlexHentschel @tarakby @durkmurder

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,253 @@
...
+
+	// read all the possible signer IDs at the given block
+	consensus, err := p.committees.Identities(blockID, filter.HasRole(flow.RoleConsensus))
+	if err != nil {
+		return nil, fmt.Errorf("could not find consensus committees by block id(%v): %w", blockID, err)
+	}
+
+	// lookup is a map from node identifier to node identity
+	// it is used to check the given signerIDs are all valid signers at the given block
+	lookup := consensus.Lookup()
```
zhangchiqing, line 132: this check is added even though it might not be necessary right now, but eventually will be needed once we work on the optimization to use signer indices to replace the signer IDs, as we will need to query the signer index by signer ID

consensus/hotstuff/signature/packager.go
durkmurder: I would really prefer if we use library for it. I feel like something so low level should be in separate module and well covered by tests. A good candidate https://pkg.go.dev/github.com/dropbox/godropbox/container/bitvector.

durkmurder (commented):
I like everything except serialization, as mentioned in comments I would like to have a library for bitvector.

AlexHentschel:
fixed typo "seralize" to "serialize" in https://github.com/onflow/flow-go/pull/1209/commits/f65a625732f2826fb6b21b1907b8a7e68623570f

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+	// validate the length of serialized
+	// it must have enough bytes to fit the `count` number of bits
+	totalBytes := count / 8
+	if count%8 > 0 {
+		totalBytes++
+	}
+
+	if len(serialized) != totalBytes {
+		return nil, fmt.Errorf("mismatching bytes for serialized sig types, total signers: %v"+
+			", expect bytes: %v, actual bytes: %v, %w", count, totalBytes, len(serialized), signature.ErrInvalidFormat)
```
AlexHentschel, line 211: In my mind, the the colon serves as the separator when wrapping errors to separate the different levels of the stack trace. 
```golang
err := someFunction( .. )
if err != nil {
   fmt.Errorf("some function failed: %w", err)
}
```
Here, the colon is now also part of the error message. Which makes it hard to separate the error trace when inspecting a log message. 

My preference would be to remove the ':' as part of the error message
```suggestion
		return nil, fmt.Errorf("encoding sig types of %d signers requires %d bytes but got %d bytes: %w",
			count, totalBytes, len(serialized), signature.ErrInvalidFormat)
```

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+	}
+
+	// parse each bit in the bit-vector, bit 0 is SigTypeStaking, bit 1 is SigTypeRandomBeacon
+	for i := 0; i < count; i++ {
+		byt := serialized[i/8]
+		posMask := byte((1 << 7) >> (i % 8))
```
AlexHentschel, line 217: here we first shift left by 7 and then shift right by `i % 8`. Instead of the forth and back, we can compute the shift upfront:
```suggestion
		offset := 7 - (i % 8)
		posMask := byte(1 << offset)
```

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+			types = append(types, hotstuff.SigTypeStaking)
+		} else {
+			types = append(types, hotstuff.SigTypeRandomBeacon)
+		}
+	}
+
+	// if there are remaining bits, they must be all `0`s
+	if count%8 > 0 {
+		// since we've validated the length of serialized, then the last byte
+		// must contains the remaining bits
```
AlexHentschel, line 228:
```suggestion
		// must contain the remaining bits
```

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+		last := serialized[len(serialized)-1]
+		remainings := last << (count % 8) // shift away the last used bits
+		if remainings != byte(0) {
+			return nil, fmt.Errorf("the remaining bits are expect to be all 0s, but actually are: %v: %w",
```
AlexHentschel, line 232:
```suggestion
			return nil, fmt.Errorf("the remaining bits are expect to be all 0s, but actually are %v: %w",
```

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+	}
+
+	// lookup is a map from node identifier to node identity
+	// it is used to check the given signerIDs are all valid signers at the given block
+	lookup := consensus.Lookup()
+
+	// read each signer's signerID and sig type from two different slices
+	// group signers by its sig type
+	stakingSigners := make([]flow.Identifier, 0)
+	randomBeaconSigners := make([]flow.Identifier, 0)
```
AlexHentschel, line 137: would prefer to over-allocate a bit of memory and thereby avoid repeated allocation and copy which unnecessarily puts load on the GC:
```suggestion
	stakingSigners := make([]flow.Identifier, 0, len(signerIDs))
	randomBeaconSigners := make([]flow.Identifier, 0, len(signerIDs))
```

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+				signerID, blockID, signature.ErrInvalidFormat)
+		}
+
+		if sigType == hotstuff.SigTypeStaking {
+			stakingSigners = append(stakingSigners, signerID)
+		} else if sigType == hotstuff.SigTypeRandomBeacon {
+			randomBeaconSigners = append(randomBeaconSigners, signerID)
+		} else {
+			return nil, fmt.Errorf("unknown sigType %v, %w", sigType, signature.ErrInvalidFormat)
+		}
```
AlexHentschel, line 153: shall we use a `switch` here?

AlexHentschel (approved):
thanks. Looks good. Mostly stylistic comments. 

#### Test suggestion for `ConsensusSigPackerImpl`  
What I noticed is that we are testing only with two blocks: first `c` staking signatures and then `y` random beacon signatures. But we will later have wild mixtures. 

I think this would be an excellent occasion for some fuzzy test:
1. randomly select `numberSigners`  in the range [17, 1017] 
2. randomly select a `hotstuff.SigType` for each signer 
3. -> `serializeToBytes` -> `deserializeFromBytes` should reproduce the original input 

run 100 repetitions of this test.

AlexHentschel:
Hey @zhangchiqing 
to follow up on [Yurii's comment](https://github.com/onflow/flow-go/pull/1209#pullrequestreview-760681676): what is the reasoning for not using a library for bit vectors. Not that I oppose your decision to roll out our own implementation. I'd just like to understand the reasoning. 
Thanks

zhangchiqing:
I thought about using bitvector library, and understand the benefit, I’m not against using that, also don’t have a strong hold on my preference. 

My reasoning of not using that was just to keep it simple. The use case of bit manipulation here are isolated in pure functions, they are easy to review and reason about. And very minimal footprint on memory, whereas the library might have to create some structs on heap and later to be garbage collected.

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +1,238 @@
...
+	for byt := 0; byt < len(sigTypes); byt += 8 {
+		b := byte(0)
+		offset := 7
+		for pos := byt; pos < byt+8 && pos < len(sigTypes); pos++ {
+			sigType := sigTypes[pos]
+
+			if !sigType.Valid() {
+				return nil, fmt.Errorf("invalid sig type: %v at pos %v", sigType, pos)
+			}
+
```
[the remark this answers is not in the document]
zhangchiqing: I agree the code would be future resilient, but I don't like that the logic here knows that `SigTypeStaking` is for `0`, and `SigTypeRandomBeacon` is for `1`. 

The reason I used:
```
sigType := sigTypes[pos]
b ^= (byte(sigType) << offset)
``` 
is so that the value is determined by the iota value defined in the type. So if we change the order of types in `SigType` definition, this code doesn't need to change.

If we need to introduce more type, we anyway will need to update the code here. And if we don't, both your approach and mine will throw the same exception.

PR #1210, opened 2021-08-26
Subscription filter

smnzhu:
Implements subscription filter. Will not work 100% until https://github.com/libp2p/go-libp2p-pubsub/issues/449 is fixed.

https://github.com/dapperlabs/flow-go/issues/5801


network/p2p/subscription_filter_test.go
```diff
@@ -0,0 +1,119 @@
...
+	defer collectionNode.Stop()
+
+	goodTopic := engine.TopicFromChannel(engine.ProvideCollections, rootBlockID)
+	_, err := collectionNode.pubSub.Join(goodTopic.String())
+	require.NoError(t, err)
+
+	badTopic := engine.TopicFromChannel(engine.ProvideReceiptsByBlockID, rootBlockID)
+	_, err = collectionNode.pubSub.Join(badTopic.String())
+	require.Error(t, err)
+
```
huitseeker, line 108: so, we're validating the `channelRoleMap` at the same time, is that correct?

network/p2p/subscription_filter.go
```diff
@@ -0,0 +1,79 @@
...
+
+	for _, ch := range channels {
+		consensusCluster := engine.ChannelConsensusCluster(f.chainID)
+		syncCluster := engine.ChannelSyncCluster(f.chainID)
+
+		if strings.HasPrefix(consensusCluster.String(), ch.String()) {
+			ch = consensusCluster
+		} else if strings.HasPrefix(syncCluster.String(), ch.String()) {
+			ch = syncCluster
+		}
```
huitseeker, line 55: If I understand `ChannelConsensusCluster` and `ChannelSyncCluster` correctly, their output will _always_ be prefixed, respectively, by `consensusClusterPrefix` and `syncClusterPrefix`. And granted, these aren't public constants. 

But here's what I don't get:
- if the collection node here is subscribed to the consensus cluster channel, why isn't the channel returned by `ChannelsByRole` ok to allow? Wouldn't it itself have been obtained dynamically through `ChannelConsensusCluster` at subscription, and if so, why replay this here?
- same Q for the `ChannelSyncConsensusCluster`.
- would we save time here by breaking out of the loop if this is not a collection node?

network/p2p/subscription_filter.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+
+	"github.com/onflow/flow-go/engine"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/id"
+	"github.com/onflow/flow-go/network"
+)
+
+type Filter struct {
```
huitseeker, line 15: So this is not a generic filter, that allows a customizable policy, right? It's a specific one, that tries to guess what to allow a remote node to subscribe to based on their role.

Would it be appropriate to rename this to e.g. `RoleBasedFilter` and leave a code comment stating the intent of the policy enforced here?

network/p2p/subscription_filter_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// publish a message from node 1 and check that only node2 receives
+	testPublish(&wg, node1, sub2)
+
+	// publish a message from node 2 and check that only node1 receives
+	testPublish(&wg, node2, sub1)
+
+	wg.Wait()
+}
```
huitseeker, line 93: So, even if libp2p limits the validity of this, here's what happens:
- the subscriptions of the staked nodes to pass,
- the subscription of the unstaked node to fail,
- and, because of the libp2p subscription, the `testPublish` to fail on either of the `Publish` calls at the end.

I surmise we can "split" this test between the subscription logic, otoh, and the `testPublish` test, which should be skipped. This isn't a duplicate of `TestCanSubscribe` because it exercises the unstaked logic.

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -210,13 +212,19 @@ func (builder *StakedAccessNodeBuilder) initLibP2PFactory(ctx context.Context,
 	}
 
 	return func() (*p2p.Node, error) {
+		psOpts := p2p.DefaultPubsubOptions(p2p.DefaultMaxPubSubMsgSize)
+		psOpts = append(psOpts, func(_ context.Context, h host.Host) (pubsub.Option, error) {
+			return pubsub.WithSubscriptionFilter(p2p.NewSubscriptionFilter(
```
huitseeker, line 217: I feel a one-liner comment on how the magic happens would help: here, the SANs have the libp2p peers in their `IdentityProvider` and would therefore let unstaked subscriptions through.

network/p2p/topic_validator_test.go
```diff
@@ -127,24 +127,26 @@ func TestTopicValidator(t *testing.T) {
...
-func createID(t *testing.T) (*flow.Identity, crypto.PrivateKey) {
+func createID(t *testing.T, opts ...func(*flow.Identity)) (*flow.Identity, crypto.PrivateKey) {
 	networkKey, err := unittest.NetworkingKey()
 	require.NoError(t, err)
-	id := unittest.IdentityFixture(
-		unittest.WithRole(flow.RoleAccess),
-		unittest.WithNetworkingKey(networkKey.PublicKey()),
-	)
+	opts = append(opts, unittest.WithNetworkingKey(networkKey.PublicKey()))
+	id := unittest.IdentityFixture(opts...)
```
huitseeker, line 134: So, it's completely non-obvious that this change is essential to the whole test component of the PR. I'd appreciate a one-liner comment on how this is used by non-AN non-topic-validator tests, or a move to a generic utils file, your choice.

huitseeker (commented):
Thanks a bunch! I left a few comments, but I feel even with the libp2p bug, this is shaping up to a meaningful step towards more safety in the network!

network/p2p/subscription_filter.go
smnzhu: For collection nodes, the `ChannelsByRole` function only returns the cluster prefixes, not the actual channel (which also contains the chain ID).

What we want to return from `allowedTopics` is the *exact* list of topics that can be subscribed. In particular, we want the full cluster channels, not just the prefix.

The logic here basically just replaces the cluster prefix returned from `ChannelsByRole` with the actual full channel name.

We don't want to break out of the loop since we still want to iterate through all channels and put the topic into the `topics` map.

network/p2p/subscription_filter_test.go
smnzhu: Sure, we can do taht.

network/p2p/subscription_filter_test.go
smnzhu: Not exactly, what do you mean by this?

cmd/access/node_builder/staked_access_node_builder.go
smnzhu: That's not exactly how it works actually. In fact, the subscription behaves the exact same way on all nodes.

The subscription filter logic always allows unstaked nodes to subscript to unstaked channels.

network/p2p/subscription_filter.go
huitseeker: Thanks for these precisions!

> We don't want to break out of the loop since we still want to iterate through all channels and put the topic into the topics map.
 
Fair enough. My point is that we're still doing :
1.  constructing the full `Channel(Sync|Consensus)Cluster` name,
2. a string prefix search,
on each iteration. While:
- if I read `ClusterChannel` correctly, there's a static `(consensus|sync)ClusterPrefix` that would make the matching part cheaper, and the full channel name construction as-needed.
- the cluster-matching logic only seems necessary if the role of the node relies on clusters [1], 

[1]: Maybe `channelRoleMap[(consensus|sync)Cluster]Prefix` could be compared against the node role?

cmd/access/node_builder/staked_access_node_builder.go
huitseeker: I think we're saying the same thing, and just I fumbled my wording, thinking, or both.

network/p2p/subscription_filter.go
smnzhu: > if I read ClusterChannel correctly, there's a static (consensus|sync)ClusterPrefix that would make the matching part cheaper, and the full channel name construction as-needed.

Yeah, but those are private variables and therefore not able to be directly compared against.

But anyway, I think overall you're right and I can make some changes to this logic here.

network/p2p/subscription_filter.go
```diff
@@ -0,0 +1,71 @@
...
+		if engine.ClusterChannelRoles().Contains(id.Role) {
+			channels = append(channels, engine.ChannelConsensusCluster(f.chainID), engine.ChannelSyncCluster(f.chainID))
+		}
+	}
+
+	for _, ch := range channels {
+		topics[engine.TopicFromChannel(ch, f.rootBlockID)] = struct{}{}
+	}
+
+	return topics
```
huitseeker, line 52: 👍

engine/channels.go
```diff
@@ -190,13 +192,15 @@ func initializeChannelRoleMap() {
...
 
-	channelRoleMap[ProvideCollections] = flow.RoleList{flow.RoleCollection, flow.RoleExecution}
+	channelRoleMap[ProvideCollections] = flow.RoleList{flow.RoleCollection, flow.RoleExecution, flow.RoleAccess}
 	channelRoleMap[ProvideChunks] = flow.RoleList{flow.RoleExecution, flow.RoleVerification}
 	channelRoleMap[ProvideReceiptsByBlockID] = flow.RoleList{flow.RoleConsensus, flow.RoleExecution}
 	channelRoleMap[ProvideApprovalsByChunk] = flow.RoleList{flow.RoleConsensus, flow.RoleVerification}
+}
 
-	channelRoleMap[syncClusterPrefix] = flow.RoleList{flow.RoleCollection}
-	channelRoleMap[consensusClusterPrefix] = flow.RoleList{flow.RoleCollection}
```
huitseeker, line 199: It's arguably OK to remove this and replace it by the `ClusterChannelRoles` logic, but the issue is that there are tests, such as `TestUniqueChannels_ClusterChannels`, that still rely on these prefixes being present in the `channelRoleMap`.

smnzhu: Updated

huitseeker (approved):
This code LGTM.

@smnzhu there's a CI unit test failure on `TestCommandRunner` that results in a nil dereference. Is there any chance you'd have time to look into it? My hunch is the cleanup / shutdown of the admin server might run first while leaving pending command, so that they probably hit a cleaned up (nil) runner in the end.

engine/channels.go
```diff
@@ -21,20 +21,22 @@ var channelRoleMap map[network.Channel]flow.RoleList
...
 func Exists(channel network.Channel) bool {
-	_, exists := RolesByChannel(channel)
-	return exists || UnstakedChannels().Contains(channel)
+	if _, ok := RolesByChannel(channel); !ok {
+		if _, isClusterChannel := ClusterChannel(channel); !isClusterChannel {
+			return PublicChannels().Contains(channel)
+		}
+	}
+
+	return true
```
vishalchangrani, line 39:
```suggestion
	if _, ok := RolesByChannel(channel); ok {
	      return true
	  }
	if _, isClusterChannel := ClusterChannel(channel); isClusterChannel {
	      return true
	}
        return PublicChannels().Contains(channel)
```

vishalchangrani: just to make the code easy to read

engine/channels.go
```diff
@@ -90,10 +91,10 @@ func Channels() network.ChannelList {
 	return channels
 }
 
-// UnstakedChannels returns all channels that unstaked nodes can send messages on.
-func UnstakedChannels() network.ChannelList {
+// PublicChannels returns all channels that on the public network.
```
vishalchangrani, line 94:
```suggestion
// PublicChannels returns all channels that are on the public network.
```

network/p2p/libp2pNode.go
```diff
@@ -729,7 +736,6 @@ func DefaultPubsubOptions(maxPubSubMsgSize int) []PubsubOption {
 		pubSubOptionFunc(pubsub.WithStrictSignatureVerification(true)),
 		// set max message size limit for 1-k PubSub messaging
 		pubSubOptionFunc(pubsub.WithMaxMessageSize(maxPubSubMsgSize)),
-		// no discovery
```
vishalchangrani, line 732: I kept the comment in intentionally to highlight that we are no using any discovery mechanism :)

yhassanzadeh13 (commented):
Overall clean and neat, just a few comments to preserve the extensibility.

Also, on a separate note, I could not find implementation accounting for https://github.com/dapperlabs/flow-go/issues/5861 in this PR. That is fine, but just wanted to make sure https://github.com/dapperlabs/flow-go/issues/5861 is not a duplicate. 
Basically, the subscription filter does not prevent an engine to wrongfully subscribe to a topic that it is not supposed to do, e.g., a consensus node can try registering itself on `request-chunks`. I would be happy to take https://github.com/dapperlabs/flow-go/issues/5861 and apply a fix.

network/p2p/subscription_filter.go
```diff
@@ -0,0 +1,71 @@
...
+func NewRoleBasedFilter(pid peer.ID, rootBlockID flow.Identifier, chainID flow.ChainID, idProvider id.IdentityProvider) *RoleBasedFilter {
+	return &RoleBasedFilter{
+		idProvider,
+		pid,
+		rootBlockID,
+		chainID,
+	}
+}
+
+func (f *RoleBasedFilter) allowedTopics(pid peer.ID) map[network.Topic]struct{} {
```
vishalchangrani, line 30: shouldn't we have two different filter. One which all the staked node use and one which the full observer node uses and the Staked AN that support the observer node use?

vishalchangrani (approved):
some nits and a question but otherwise lgtm

smnzhu:
> Overall clean and neat, just a few comments to preserve the extensibility.
> 
> Also, on a separate note, I could not find implementation accounting for [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) in this PR. That is fine, but just wanted to make sure [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) is not a duplicate.
> Basically, the subscription filter does not prevent an engine to wrongfully subscribe to a topic that it is not supposed to do, e.g., a consensus node can try registering itself on `request-chunks`. The aim of [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) is to prevent such scenarios that arising of `channelRoleMap` getting out of sync or distorted and seems [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) and this PR are serving quite orthogonal purposes. I would be happy to take [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) and apply a fix.

So originally, I never suggested that [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) is a duplicate, but actually I would like to clarify that the subscription filter *does* prevent an engine from wrongfully subscribing to a topic that it is not supposed to, in fact this is explicitly tested in the unit test.

It wasn't clear to me based on the wording what exactly the goal of [dapperlabs/flow-go#5861](https://github.com/dapperlabs/flow-go/issues/5861) was, but if the goal is to ensure that nodes only subscribe to the appropriate channels based on their role, then the subscription filter does ensure this.

engine/channels.go
```diff
@@ -63,10 +65,9 @@ func UniqueChannels(channels network.ChannelList) network.ChannelList {
 	// has already been added to uniques.
 	// We use identifier of RoleList to determine its uniqueness.
 	for _, channel := range channels {
-		id := channelRoleMap[channel].ID()
-
 		// non-cluster channel deduplicated based identifier of role list
 		if _, cluster := ClusterChannel(channel); !cluster {
+			id := channelRoleMap[channel].ID()
```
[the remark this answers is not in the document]
smnzhu: The old place is already inside the loop 🧐

I have only placed it inside the `if`, which is only saving computation because it will be computed less often.

engine/channels.go
```diff
@@ -21,20 +21,22 @@ var channelRoleMap map[network.Channel]flow.RoleList
 
 // RolesByChannel returns list of flow roles involved in the channel.
 func RolesByChannel(channel network.Channel) (flow.RoleList, bool) {
-	if clusterChannel, isCluster := ClusterChannel(channel); isCluster {
-		// replaces channel with the stripped-off prefix
-		channel = clusterChannel
+	if _, isCluster := ClusterChannel(channel); isCluster {
+		return ClusterChannelRoles(), true
```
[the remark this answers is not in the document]
smnzhu: right, but channelRoleMap doesn't make sense for cluster-based channels imo because they are unique to the chain ID. ie, the value stored in `channelRoleMap` is not actually a "real" channel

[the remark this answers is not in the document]
smnzhu: I wonder if we can remove `ChannelSyncCluster()` and `ChannelConsensusCluster` entirely, and simply redefine `TopicFromChannel` to take in a chain ID instead? Then we'd be able to rename `syncClusterPrefix` to simply `SyncCluster` (similary for consensus) and add it to the channel role map

engine/channels.go
```diff
@@ -90,10 +91,10 @@ func Channels() network.ChannelList {
 	return channels
 }
 
-// UnstakedChannels returns all channels that unstaked nodes can send messages on.
-func UnstakedChannels() network.ChannelList {
+// PublicChannels returns all channels that on the public network.
+func PublicChannels() network.ChannelList {
 	return network.ChannelList{
-		UnstakedSyncCommittee,
+		PublicSyncCommittee,
```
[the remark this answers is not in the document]
smnzhu: Imo, `channelRoleMap` is only relevant to the channels where roles matter, ie private channels.

Public channels can be joined by anyone, regardless of role (and even unstaked nodes who don't have a role whatsoever), so I think it makes sense to treat them separately here.

network/p2p/subscription_filter.go
smnzhu: We could do this, the at a high level, the logic for the subscription filter goes:

A node should only be allowed to subscribe to some set of channels based on its role.

This general rule is the same for any node type, so we can implement the exact same logic for all of them

smnzhu:
bors merge

smnzhu:
bors merge

PR #1216, opened 2021-08-27
[Networking] Async DNS Update

yhassanzadeh13:
This PR refactors the current synchronous DNS cache into an asynchronous one. In the asynchronous version, a query for an expired entry is answered by the expired one, however, it also triggers an asynchronous call to the underlying resolver for refreshing the query from the underlying network. This speeds up query resolving as, except the first initial query for a domain, the rest are answered from the cache, so they won't be blocked for the underlying resolver's network latency. It also allows having a shorter TTL for domains (5 min instead of 1 hour) since frequently updating the DNS cache won't affect the query resolving functionality.


jwinkler2083233 (approved):
This looks like it will suit our needs very well.

network/p2p/dns/cache.go
```diff
@@ -91,3 +97,29 @@ func (c *cache) updateTXTCache(txt string, addr []string) {
 		timestamp: runtimeNano(),
 	}
 }
+
+// invalidateIPCacheEntry atomically invalidates ip cache entry. Boolean variable determines whether invalidation
+// is successful.
+func (c *cache) invalidateIPCacheEntry(domain string) bool {
+	c.Lock()
```
vishalchangrani, line 104: Nit: but how about this?
```suggestion
	c.RLock()
	_, exists := c.ipCache[domain]
	c.RUnlock()
	
	if !exists {
	    return false
	}
	delete(c.ipCache, domain) // delete is idempotent and once in a while it is ok to report as exist even when it may have been deleted avoid the cost of a write lock
```

network/p2p/dns/resolver_test.go
```diff
@@ -83,10 +87,51 @@ func TestResolver_Error(t *testing.T) {
...
+	resolverWG := mockBasicResolverForDomains(t, &basicResolver, ipTestCase, txtTestCases, !happyPath, 1)
+	// queries are answered by cache, so resolver returning an error only invalidates the cache asynchronously for the first time.
+	queryWG := syncThenAsyncQuery(t, 1, resolver, txtTestCases, ipTestCase, happyPath)
+
+	unittest.RequireReturnsBefore(t, queryWG.Wait, 1*time.Second, "could not perform all queries on time")
+	unittest.RequireReturnsBefore(t, resolverWG.Wait, 1*time.Second, "could not resolve all expected domains")
+
+	// since resolving hits an error, cache is invalidated.
+	require.Empty(t, resolver.c.ipCache)
+	require.Empty(t, resolver.c.txtCache)
```
huitseeker, line 132: `invalidateIPCacheEntry` (resp `invalidateTXTCacheEntry`) is called in a `unit.Launch`, which means you haven't captured an happens-after by waiting for `queryWG` & `resolverWG`, and that these 2 `require.Empty` are in a data race with the invalidations.

network/p2p/dns/resolver.go
```diff
@@ -109,22 +151,82 @@ func (r *Resolver) LookupTXT(ctx context.Context, txt string) ([]string, error)
...
+}
+
+// doneResolvingIP cleans up tracking an ongoing concurrent attempt for resolving txt.
+func (r *Resolver) doneResolvingTXT(txt string) {
+	r.Lock()
+	defer r.Unlock()
+
+	delete(r.processingTXTs, txt)
+}
+
```
huitseeker, line 220: Vishal's [comment](https://github.com/onflow/flow-go/pull/1216/files#r708767925) may also apply here.

network/p2p/dns/resolver.go
```diff
@@ -109,22 +151,82 @@ func (r *Resolver) LookupTXT(ctx context.Context, txt string) ([]string, error)
...
+	defer r.Unlock()
+
+	if _, ok := r.processingIPs[domain]; !ok {
+		r.processingIPs[domain] = struct{}{}
+		return true
+	}
+
+	return false
+}
+
```
huitseeker, line 204: This CAS-like op makes me think this may be a good use for go's [sync.Map](https://www.youtube.com/watch?v=C1EtfDnsdDs), though that's perhaps a place where we want to benchmark rather than speculate.

/cc @jwinkler2083233

huitseeker (commented):
Concept ACK. I left a few comments on the concurrency details.

network/p2p/dns/resolver.go
jwinkler2083233: Francois is right, but since this is not heavily used, the current code is fine as it's written.

network/p2p/dns/cache.go
yhassanzadeh13: @vishalchangrani @huitseeker I applied this comment and refactored similar places, however, now the code is vulnerable to concurrent map access which will result in crash failure. Currently, the DNS resolver interacting with cache follows a causality in operations, e.g., it does not perform invalidation concurrently with resolving. However, if this behavior is changed, later on, this race condition will emerge as a crash failure. Just wanted to highlight it.

network/p2p/dns/resolver.go
yhassanzadeh13: @huitseeker @jwinkler2083233 implemented `sync.Map` variant on a [different branch](https://github.com/onflow/flow-go/tree/yahya/temp-benchmarking-concurrent-map) and performed benchmarking. Please find the results in the following:

As shown below, at small scale, when thread contention is low (i.e., 300 concurrent operations), the `map` version is around **8 times faster** than `sync.Map`. However, the coin flips swiftly when we scale the operations up:

with 3000 operations, `sync.Map` is **about 2.35 times faster** than `map`.
with 30_000 operations, `sync.Map` is **about 2.66 times faster** than `map`. 
with 300_000 concurrent operations, `sync.Map` is **about 3.27 times faster** than `map`. 

💡  as we scale up on thread contention, `sync.Map` significantly outperforms `map`. It seems like literally, we gonna need `sync.Map` as we scale up the number of nodes. I have it implemented on another branch so moving it to this branch is easy to handle. The question is: 
❓ should we go with `sync.Map` version? 

```
/*
map result
*/

// 300 concurrent operations with map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000            175202 ns/op

// 3000 concurrent operations with map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000           1931664 ns/op

// 30_000 concurrent operations with map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000          24803221 ns/op

// 300_000 concurrent operations with map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000         318880473 ns/op

/*
sync.Map result.
Averages are taken over 1000 experiments
*/
// 300 concurrent operations with sync.Map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000             85839 ns/op

// 3000 concurrent operations with sync.Map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000            812991 ns/op

// 30_000 concurrent operations with sync.Map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000           9039950 ns/op

// 300_000 concurrent operations with sync.Map
goos: darwin
goarch: amd64
pkg: github.com/onflow/flow-go/network/p2p/dns
cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
BenchmarkCache-8            1000          97440807 ns/op
```

huitseeker: It's probably not worth buying into `sync.Map` right away, but it would probably be useful to open an issue to keep track of this, and link and persist the branch in which you have experimented with `sync.Map` to it.

network/p2p/dns/resolver_test.go
huitseeker: This one is still present, and I suspect `TestResolver_CacheExpiry` exhibits the issue as well. A quick fix would be to put the `require.Empty` in an eventually.

network/p2p/dns/resolver.go
huitseeker: It's probably not worth buying into `sync.Map` right away, but it would probably be useful to open an issue to keep track of this, and link and persist the branch in which you have experimented with `sync.Map` to it.

This bench is vary nice, thanks for the rigorous experimentation.

yhassanzadeh13: Issued up here: https://github.com/dapperlabs/flow-go/issues/5855

network/p2p/dns/resolver_test.go
yhassanzadeh13: Moved `require.Empty` statements after the invocation of `resolver.Done`. The `Done` method returns once all outstanding concurrent `unit.Launch` operations terminate. So by the time `resolver.Done` returns we are sure that the termination of `invalidateIPCacheEntry`, `invalidateTXTCacheEntry`, as well as any other method invoked in `unit.Launch`  happens-before the `unit.Done` termination. So, evaluating `require.Empty` right after `Done` makes it concurrency safe.

network/p2p/dns/resolver.go
yhassanzadeh13: Removing locks here causes a data race between `shouldResolveIP` and `doneResolvingIP`. A similar data race is also prone for `shouldResolveTXT` and `doneResolvingTXT`.

vishalchangrani (approved):

huitseeker (approved):
Thanks for addressing the difficult concurrency issues in tests! Awesome PR!

network/p2p/dns/cache.go
```diff
@@ -31,43 +31,49 @@ func newCache() *cache {
...
 // resolveIPCache resolves the domain through the cache if it is available.
+// First boolean variable determines whether the domain exists in the cache.
+// Second boolean variable determines whether the domain cache is fresh, i.e., TTL has not yet reached.
 func (c *cache) resolveIPCache(domain string) ([]net.IPAddr, bool, bool) {
-	c.Lock()
-	defer c.Unlock()
+	c.RLock()
+	defer c.RUnlock()
 
 	entry, ok := c.ipCache[domain]
```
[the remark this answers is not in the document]
yhassanzadeh13: Applied in https://github.com/onflow/flow-go/pull/1216/commits/bf03165724b1aa89f53e22c71850d523d5fe9600.

network/p2p/dns/cache.go
```diff
@@ -91,3 +97,39 @@ func (c *cache) updateTXTCache(txt string, addr []string) {
...
+	_, exists := c.ipCache[domain]
+	c.RUnlock()
+
+	if !exists {
+		return false
+	}
+
+	// delete is idempotent and once in a while it is ok to report as exist even
+	// when it may have been deleted to avoid the cost of write lock
+	delete(c.ipCache, domain)
```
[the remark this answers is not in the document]
vishalchangrani: Actually, you are right. I didn't find any documentation which said `delete` is thread safe. 
@yhassanzadeh13 - this will need a Write lock.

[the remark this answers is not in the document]
yhassanzadeh13: fixes in https://github.com/onflow/flow-go/pull/1216/commits/7d5afb009cb39f05dbed0b80b53b628827d7c70d

network/p2p/dns/resolver.go
```diff
@@ -74,17 +93,37 @@ func (r *Resolver) LookupIPAddr(ctx context.Context, domain string) ([]net.IPAdd
...
 	}
 
-	// resolves domain through underlying resolver
-	r.collector.OnDNSCacheMiss()
+	if !fresh && r.shouldResolveIP(domain) {
+		r.unit.Launch(func() {
+			_, err := r.lookupResolverForIPAddr(ctx, domain)
+			if err != nil {
+				// invalidates cached entry when hits error on resolving.
+				invalidated := r.c.invalidateIPCacheEntry(domain)
```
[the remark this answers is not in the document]
yhassanzadeh13: > Why invalidate the cache upon error when resolving? Wouldn't we just flag look-up again? Or forget about it because we made a best effort and did one lookup. Any DNS lookup could easily fail due to a missing UDP packet? But does that mean we should invalidate the cache?

Thanks for the comment, I appreciate the optimization suggestion, but I think we need to be careful about the safety margins we move. As mentioned to an [earlier comment](https://github.com/onflow/flow-go/pull/1216#discussion_r714242218), we'd safer to reflect error directly to application layer and delete the expired entry as soon as we hit an error. The cache is in place to boost up the _happy path_ execution, and currently is not designed for handling unhappy path scenarios. Hence, at the current version, a fail-safe cache that boosts up the happy path suites more secure than a cache trying to deal with DNS errors. Of course, you comment is valuable to consider and I will issue it up to craft a well-thought solution later. 

> I think there are two types of lookup errors here:
> 
> 1. Lookup failed due to a trivial reason such as UDP packet going missing.
> 2. Lookup failed due to a major reason such as infrastructure re-configured during epoch.

Your thought are pretty valid on a theoretical perspective. However, digging a bit into the underlying resolver, I found that it resides on top of several dependencies ranging from libp2p to native Go, and there is no such strict guarantee on the returned error types through its dependency injection. In short, an error means unhappy path without any guarantee on the type. So, unless we implement a wrapper around the underlying DNS to cast the errors into well-defined error types that we can distinguish case (1) and (2) from any other default case, it does not look safe to establish the assumption on implementation. I issued up https://github.com/dapperlabs/flow-go/issues/5872 to re-surface your suggestion later on.

network/p2p/dns/cache.go
```diff
@@ -91,3 +97,39 @@ func (c *cache) updateTXTCache(txt string, addr []string) {
 		timestamp: runtimeNano(),
 	}
 }
+
+// invalidateIPCacheEntry atomically invalidates ip cache entry. Boolean variable determines whether invalidation
+// is successful.
```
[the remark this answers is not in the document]
simonhf: I'm still confused as to whether nodes should be expecting `NXDOMAIN`s or not as part of normal operations.

> one typical type of such errors is a no such a host error, i.e. when the DNS domain is no longer valid

That is a possible error but is it a realistic / common / typical scenario for a node? Under which normal circumstances would a node be trying to talk to a domain for x period of time, only to have the IP pulled from DNS while the node is still trying to actively talk to the domain?

Assuming that in normal operation there should never be an `NXDOMAIN` DNS response, why not just:
* Always return the cached IP when asked for the IP?
* If the domain has expired then do a single async lookup attempt.
* If `NXDOMAIN` then:
  * Don't delete anything and don't async lookup again until the IP is asked for again.
  * Log an error because we never expect `NXDOMAIN` when running?

network/p2p/dns/resolver.go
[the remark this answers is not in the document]
simonhf: Thanks for the comments!

Yes, I'm not convinced that the unhappy path has been thought through, or that this PR is necessarily taking the best action upon hitting the unhappy path by invalidating the cache entry. For a single `NXDOMAIN` then this will result in not one but two errors (the last cached IP result is returned which presumably does not work anymore, and later the domain will be invalidated in the cache which presumably generates a 2nd error) which might be confusing for the person debugging.

However, I'm happy to approve this PR as is, assuming that the unhappy path is debated and possibly refined later using the issue [1] kindly created :-)

[1] https://github.com/dapperlabs/flow-go/issues/5872

simonhf (approved):
Approving this PR subject to further debate of cache invalidation in separate issue:
https://github.com/dapperlabs/flow-go/issues/5872

PR #1219, opened 2021-08-27
FVM Benchmarking

janezpodhostnik:
The premise of this benchmark is to run very similar transactions under the same conditions so a comparative analysis between execution time can be done.

Each benchmark executes block filled with 10 transactions. Transactions each have a loop of length 100 that does *something*. That something is described by the test name.

Here are the results (I made 5 runs and used benchstat to get a better looking output):

```
name                                                                 time/op
RuntimeTransaction/reference_tx-24                                   28.1ms ± 1%
RuntimeTransaction/log_empty_string-24                               29.6ms ± 3%
RuntimeTransaction/log_long_string-24                                32.3ms ± 5%
RuntimeTransaction/convert_int_to_string-24                          31.4ms ± 6%
RuntimeTransaction/convert_int_to_string_and_concatenate_it-24       31.2ms ± 2%
RuntimeTransaction/get_signer_address-24                             29.9ms ± 1%
RuntimeTransaction/get_public_account-24                             36.6ms ± 1%
RuntimeTransaction/get_account_and_get_balance-24                     457ms ± 3%
RuntimeTransaction/get_account_and_get_available_balance-24           395ms ± 1%
RuntimeTransaction/get_account_and_get_storage_used-24               40.7ms ± 1%
RuntimeTransaction/get_account_and_get_storage_capacity-24            376ms ± 1%
RuntimeTransaction/get_signer_vault-24                               37.6ms ± 3%
RuntimeTransaction/get_signer_receiver-24                            50.7ms ± 1%
RuntimeTransaction/transfer_tokens-24                                 138ms ± 2%
RuntimeTransaction/load_and_save_empty_string_on_signers_address-24  37.5ms ± 1%
RuntimeTransaction/load_and_save_long_string_on_signers_address-24   38.8ms ± 2%
RuntimeTransaction/create_new_account-24                              1.17s ± 1%
RuntimeTransaction/call_empty_contract_function-24                   36.1ms ± 1%
RuntimeTransaction/emit_event-24                                     40.8ms ± 1%

```


fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,1443 @@
...
+			let receiverRef =  getAccount(signer.address)
+				.getCapability(/public/flowTokenReceiver)
+				.borrow<&{FungibleToken.Receiver}>()!
+			
+			let vaultRef = signer.borrow<&FlowToken.Vault>(from: /storage/flowTokenVault)!
+
+			receiverRef.deposit(from: <-vaultRef.withdraw(amount: 0.00001))
+			`))
+	})
+	b.Run("load and save empty string on signers address", func(b *testing.B) {
```
ramtinms, line 172: wouldn't be better if we add a similar test but for a large string.

fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,1443 @@
...
+        // Kind of like when Damien Lillard destroys the hopes and
+        // dreams of the entire city of Houston
+        //
+        destroy() {
+            destroy self.ownedNFTs
+        }
+    }
+
+    // -----------------------------------------------------------------------
+    // TopShot contract-level function definitions
```
ramtinms, line 1116: we should remove any names like TopShot, moments, or other product names.

fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,1443 @@
...
+
+    // the total number of Top shot moment NFTs that have been created
+    // Because NFTs can be destroyed, it doesn't necessarily mean that this
+    // reflects the total number of NFTs in existence, just the number that
+    // have been minted to date.
+    // Is also used as global moment IDs for minting
+    pub var totalSupply: UInt64
+
+    // -----------------------------------------------------------------------
+    // TopShot contract-level Composite Type DEFINITIONS
```
ramtinms, line 597: same here, I also suggest we remove comments to keep things more compact

fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,1443 @@
...
+	}
+
+}
+
+func deployTopShot(b *testing.B, tctx *vmTestContext, executeBlocks func(txs [][]*flow.TransactionBody) *execution.ComputationResult, a flow.Address, pk flow.AccountPrivateKey, nftAddress flow.Address) {
+	topShotContract := func(nftAddress flow.Address) string {
+		return fmt.Sprintf(`
+import NonFungibleToken from 0x%s
+
+pub contract TopShot: NonFungibleToken {
```
ramtinms, line 524: We should call this contract something else

fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,1443 @@
...
+	b.Run("reference tx", func(b *testing.B) {
+		benchTransaction(b, templateTx(""))
+	})
+	b.Run("log empty string", func(b *testing.B) {
+		benchTransaction(b, templateTx(`log("")`))
+	})
+	b.Run("log long string", func(b *testing.B) {
+		benchTransaction(b, templateTx(`log("0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")`))
+		// 100 character string
+	})
```
ramtinms, line 128: not sure if we care that much about these methods from benchmarking perspective,

janezpodhostnik: in case if people leave them in transactions, It could be good to know if they have a significant performance cost.

ramtinms: you're right but I think we ignore them anyway for transactions and won't even capture them in the first place.

ramtinms (approved):

fvm/fvm_bench_test.go
```diff
@@ -0,0 +1,972 @@
...
+	blockComputer, err := computer.NewBlockComputer(vm, fvmContext, collector, tracer, logger, ledgerCommitter)
+	require.NoError(tb, err)
+
+	view := delta.NewView(exeState.LedgerGetRegister(ledger, initialCommit))
+
+	return func(txs [][]*flow.TransactionBody) *execution.ComputationResult {
+		executableBlock := unittest.ExecutableBlockFromTransactions(txs)
+		executableBlock.StartState = &initialCommit
+
+		computationResult, err := blockComputer.ExecuteBlock(context.Background(), executableBlock, view, programs.NewEmptyPrograms())
```
ramtinms, line 425: it might be better for some transactions if we share the same programs.

ramtinms: like for sequence of batch transfer transactions.

janezpodhostnik: I wanted to replicate actual behaviour, and I think we create new programs for each block, but now I'm not sure. I will check.

ramtinms: the actual behaviour is keeping the programs between blocks until a new contract is deployed or updated

janezpodhostnik: Using a single instance of programs should take care of that, since programs are reset in the FVM

janezpodhostnik:
bors try

huitseeker (commented):
More bench_testing of the FVM is awesome.

I wonder if we could ease maintenance by having the contracts as embedded resource files, and then open and read them as regular cdc. That way, any tooling for inspecting, maintaining, migrating (etc) cadence code could apply easily to those tests.

Kay-Zee (approved):
Yay! More benchmarking!

huitseeker:
@janezpodhostnik Agreed. Do we have an issue to keep track of that?

janezpodhostnik:
we do now: https://github.com/dapperlabs/flow-internal/issues/1592

janezpodhostnik:
bors merge

PR #1220, opened 2021-08-27
[FVM] backport of rejecting transactions with zero gas limit

ramtinms:
This PR
- updates the transaction validator code to reject transactions with zero gas limit
- updates FVM to returns default gas limit in case gas limit is set to zero
- sets proper gas limit for system chunk transaction
- improves error messages for failure cases when sending a script or account query for a block that is not recent. usually, users are confused when receiving trie not found error.


jordanschalm (approved):

janezpodhostnik (approved):
besides the `Test_ExecutionMatchesVerification/with_failed_transaction_fee_deduction` using too much gas, LGTM.

PR #1221, opened 2021-08-28
[consensus  v2] remove new workflow

tarakby:
Remove the new workflow after using the branch protection feature in GitHub to achieve the same outcome.

The PR also fixes a typo that causes a build failure.


consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
```diff
@@ -90,7 +90,7 @@ func (c *ConsensusClusterVoteCollector) AddVote(vote *model.Vote) error {
 	}
 
 	// we haven't collected sufficient weight or shares, we have nothing to do further
-	if !c.hasSufficientStake() {
+	if !c.hasSufficientWeight() {
```
Kay-Zee, line 93: I guess just a rider? As long as this is intentional, i think it's fine

Kay-Zee (approved):

consensus/hotstuff/votecollector/consensus_cluster_vote_collector.go
tarakby: Yes it's intentional. Now that we have turned the branch protection on, if I don't make this change the tests do not pass and the PR is not mergeable

huitseeker (approved):
LGTM!

PR #989, opened 2021-07-19
bringing up a GRPC secure sever for the access node

vishalchangrani:
Part 2 of the secure GRPC server implementation (based on the earlier [POC](https://github.com/onflow/flow-go/pull/929))

This PR:
1. Bring up a secure GRPC server on port 9001.
2. Provides a Client config that can be used to create secure GRPC client of the server.

PR #1224, opened 2021-08-28
Integrate secure GRPC client to DKG and QC clients

kc1116:
This PR integrates secure access client into collection/consensus node

- https://github.com/dapperlabs/flow-go/issues/5619
- https://github.com/dapperlabs/flow-go/issues/5793
- https://github.com/dapperlabs/flow-go/issues/5794
- https://github.com/dapperlabs/flow-go/issues/5805


integration/localnet/bootstrap.go
```diff
@@ -431,6 +463,7 @@ func prepareAccessService(container testnet.ContainerConfig, i int) Service {
 
 	service.Ports = []string{
 		fmt.Sprintf("%d:%d", AccessAPIPort+i, RPCPort),
+		fmt.Sprintf("%d:%d", AccessAPIPort+(i+1), SecuredRPCPort),
```
jordanschalm, line 466: I think this will overlap ports with >1 AN
```suggestion
		fmt.Sprintf("%d:%d", AccessAPIPort+2*i, RPCPort),
		fmt.Sprintf("%d:%d", AccessAPIPort+(2*i+1), SecuredRPCPort),
```

cmd/collection/main.go
```diff
@@ -138,6 +143,9 @@ func main() {
 
 			// epoch qc contract flags
 			flags.StringVar(&accessAddress, "access-address", "", "the address of an access node")
+			flags.StringVar(&securedAccessAddress, "secured-access-address", "", "the address for secured GRPC conn to an access node")
+			flags.StringVar(&accessApiNodePubKey, "access-node-grpc-public-key", "", "the networking public key of the secured access node being connected to")
+			flags.BoolVar(&insecureAccessAPI, "insecure-access-api", true, "required if insecure GRPC connection should be used")
```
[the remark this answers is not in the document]
kc1116: Integration tests are not configured properly to handle new flags , can we create separate ticket to make secure API default and update integration tests?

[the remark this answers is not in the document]
jordanschalm: 👍  Sure, sounds good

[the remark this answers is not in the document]
kc1116: https://github.com/dapperlabs/flow-internal/issues/1574

integration/localnet/bootstrap.go
```diff
@@ -257,10 +265,20 @@ func prepareServices(containers []testnet.ContainerConfig) Services {
 	for _, container := range containers {
 		switch container.Role {
 		case flow.RoleConsensus:
-			services[container.ContainerName] = prepareConsensusService(container, numConsensus, "localnet_access_1_1:9000")
+			services[container.ContainerName] = prepareConsensusService(
+				container,
+				numConsensus,
+				"access_1:9001",
```
jordanschalm, line 271: This address is essentially a constant, I think it would be better to store it with the other constants and use the value directly in `preparareConsensusService`.
https://github.com/onflow/flow-go/blob/1ccefc012fd7bee874f9c094db9dab931463563a/integration/localnet/bootstrap.go#L38-L40

jordanschalm (commented):
One final suggestion, otherwise this looks good to me. 

Where is this ticket at with the latest updates? Have we tested an epoch transition with these changes on localnet?

jordanschalm (approved):
🎸

cmd/collection/main.go
```diff
@@ -396,8 +401,40 @@ func main() {
...
+			// create flow client with correct GRPC configuration for QC contract client
+			var flowClient *client.Client
+			if insecureAccessAPI {
+				flowClient, err = common.InsecureFlowClient(accessAddress)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				if secureAccessNodeID == "" {
+					return nil, fmt.Errorf("invalid flag --secure-access-node-id required")
```
huitseeker, line 413:
```suggestion
					return nil, fmt.Errorf("invalid flag, --secure-access-node-id required")
```
otherwise the reader may think the invalid qualifier applies to your requirement

huitseeker (commented):
I'm a bit worried this may mix two distinct notions:

- do we connect securely to a TLS-ed GRPC server using a pre-shared key to check authentication? 
  That's the server that's introduced in https://github.com/onflow/flow-go/pull/989
  This is about using a capability.
  
- do we connect securely (where securely is defined as above) to our "favorite" access node, which is the one node we trust above others for various reasons ?
  That's the new functionality this PR introduces, this is meant to make an identity selection.
  
I note that the former has a wider range of application than the later: I might want to connect to any, no or every access node that runs the necessary server using GRPC + TLS + pre-shared key. I may also, besides that, restrict the access nodes I rely on ("trust") to any subset of those I connect securely. 

I would welcome:
- sunsetting insecure encryption for everything but tests (but #1574 is enough for me),
- encryption flags that do not require specifying a NodeID (but that may require a fallback, for a while, to insecure access for nodes that happen to not run the secure server => open an issue to mark it for later?),
- and on top of that, an option that clearly states "only pick this access node, it's my favorite" (and imply a secure connection is the only one that can be used)

WDYT?
/cc @vishalchangrani

jordanschalm:
> encryption flags that do not require specifying a NodeID (but that may require a fallback, for a while, to insecure access for nodes that happen to not run the secure server => open an issue to mark it for later?),

The main issue is that within the protocol state, we know an access node's networking public key and _libp2p_ networking address, but we do not know its GRPC server address (hence the `--access-address` flag). We could make the assumption that a particular relationship exists between the _libp2p_ networking address and the GRPC server address (for example, it uses the canonical port on the same hostname), but we would need to make sure that assumption is always true. Personally I'm more confident specifying the parameters explicitly.

kc1116:
> I'm a bit worried this may mix two distinct notions:
> 
> * do we connect securely to a TLS-ed GRPC server using a pre-shared key to check authentication?
>   That's the server that's introduced in [bringing up a GRPC secure sever for the access node #989](https://github.com/onflow/flow-go/pull/989)
>   This is about using a capability.
> * do we connect securely (where securely is defined as above) to our "favorite" access node, which is the one node we trust above others for various reasons ?
>   That's the new functionality this PR introduces, this is meant to make an identity selection.
- secured implies TLS-ed GRPC connection , to the secure GRPC sever AN bring up 
- There is no concept of a "favorite AN", this PR does not introduce that concept. Your only required to provide the node ID of the access node you are already connecting to in order to use secure GRPC . Under the hood the node ID is used to find the networking key from protocol state of that AN to create TLS'ed GRPC conn. Using NodeID allows networking keys to change. 
> I note that the former has a wider range of application than the later: I might want to connect to any, no or every access node that runs the necessary server using GRPC + TLS + pre-shared key. I may also, besides that, restrict the access nodes I rely on ("trust") to any subset of those I connect to securely.

- I agree , this is nice to have and is not immediate requirement

> 
> I would welcome:
> 
> * sunsetting insecure encryption for everything but tests (but #1574 is enough for me),
> * encryption flags that do not require specifying a NodeID (but that may require a fallback, for a while, to insecure access for nodes that happen to not run the secure server => open an issue to mark it for later?),
> * and on top of that, an option that clearly states "only pick this access node, it's my favorite" (and imply a secure connection is the only one that can be used)

- #1574 should handle sunsetting insecure api safely
- Encryption flags that do not require specifying a NodeID is not possible. Either you need to provide networking key directly , or node ID to look up networking key. The latter provides us with more flexibility. 

> 
> WDYT?
> /cc @vishalchangrani

@huitseeker

cmd/collection/main.go
```diff
@@ -396,8 +401,40 @@ func main() {
 
 			signer := verification.NewSingleSigner(staking, node.Me.NodeID())
 
+			// create flow client with correct GRPC configuration for QC contract client
+			var flowClient *client.Client
+			if insecureAccessAPI {
+				flowClient, err = common.InsecureFlowClient(accessAddress)
```
vishalchangrani, line 407: if the grpc client is created here and never re-created, what happens if the access node is restarted and the connection is broken?

cmd/collection/main.go
```diff
@@ -396,8 +401,40 @@ func main() {
...
+				identities, err := node.State.Sealed().Identities(filter.HasNodeID(nodeID))
+				if err != nil {
+					return nil, fmt.Errorf("could not get identity of secure access node: %s", secureAccessNodeID)
+				}
+
+				if len(identities) < 1 {
+					return nil, fmt.Errorf("could not find identity of secure access node: %s", secureAccessNodeID)
+				}
+
+				flowClient, err = common.SecureFlowClient(accessAddress, identities[0].NetworkPubKey.String()[2:])
```
vishalchangrani, line 430: can you throw in a comment for the `[2:]`

cmd/collection/main.go
kc1116: I will investigate this .

kc1116: From this issue https://github.com/grpc/grpc-go/issues/351 it looks like the GRPC client should try to automatically reconnect. Because nothing changed as far as where the client is being created, we should see the same behavior for insecure and secure client connections. This probably needs it's own separate ticket for investigation.

vishalchangrani: yeah we probably need a factory to be passed in, instead of the actual client since the client will only be used once in an epoch and it is highly probable that the access node gets restarted in-between.
Separate ticket also sounds good.

huitseeker:
@kc1116 My comments were definitely meant as non-blocking.

cmd/collection/main.go
jordanschalm: Should be able to test this by running localnet and restarting the access node before the epoch setup phase starts. I would agree with khalil and would expect that the client would recreate the connection as needed

utils/grpc/grpc_client.go
```diff
@@ -0,0 +1,31 @@
+package grpcutils
```
vishalchangrani, line 1: bit: best to make the folder name match the package name

vishalchangrani (approved):
lgtm (left some minor `nit` comments)

huitseeker (approved):

utils/grpc/grpc_client.go
jordanschalm: https://github.com/onflow/flow-go/pull/1224/commits/ad89c53ad5fb896550eb38d9c3ee8f1dd802b6dd

jordanschalm:
bors merge

PR #1225, opened 2021-08-28
[network] topic validator for staked channels

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/5802

### TODO:
- [x] write tests


network/p2p/libp2pUtils.go
```diff
@@ -249,3 +249,20 @@ func ExtractPeerID(networkPubKey fcrypto.PublicKey) (pid peer.ID, err error) {
 
 	return
 }
+
+// FlowPublicKeyFromPeerID converts a LibP2P peer ID to a Flow public key.
+func FlowPublicKeyFromPeerID(pid peer.ID) (networkPubKey fcrypto.PublicKey, err error) {
```
huitseeker, line 254: This will not work for most nodes, and in fact will not work for our internal nodes.
https://github.com/onflow/flow-go/pull/1133/files#diff-f66caf91d2b6f3779515b2ef059291d0ed5de43bd384388bed50360a85df8192R18-R45
Are we sure we want this in libp2pUtils? If so, don't we want a big red comment on it?

network/p2p/topic_validator.go
```diff
@@ -0,0 +1,31 @@
...
+	if from, err := messageSigningID(msg); err == nil {
+		// check that the sender peer ID matches a staked Flow key
+		for _, id := range v.stakedIdentities() {
+			if key, err := LibP2PPublicKeyFromFlow(id.NetworkPubKey); err == nil {
+				if pid, err := peer.IDFromPublicKey(key); err == nil {
+					if from == pid {
+						return pubsub.ValidationAccept
+					}
+				}
+			}
```
huitseeker, line 26: Could we use `PeerIDFromFlowPublicKey` ?

Also, I'm kinda mifed we recompute this on every message. I understand the idea is for the closure in `stakedIdentities` to control the refresh of identities, but it would be great to fetch the libp2p identities from a cache if `stakedIdentities()` returns the same as its cached value.

network/p2p/topic_validator_test.go
```diff
@@ -0,0 +1,122 @@
...
+	identity2, privateKey2 := createID(t)
+	node2 := createNode(t, identity2.NodeID, privateKey2)
+
+	badTopic := engine.TopicFromChannel(engine.SyncCommittee, rootBlockID)
+
+	validator := StakedValidator{func() flow.IdentityList {
+		return flow.IdentityList{identity1, identity2}
+	}}
+
+	unstakedKey, err := unittest.NetworkingKey()
```
huitseeker, line 30: Should we need a `unittest.UnstakedNetworkingKey()` at some point? happy to add one if you open an issue.

huitseeker (commented):
Thanks a bunch! this looks good apart from a couple of places

network/p2p/topic_validator_test.go
smnzhu: So currently I think we are okay? And eventually once we no longer need a mapping from Flow IDs to Peer IDs for unstaked nodes (because unstaked nodes will not need a Flow ID), then we will also not need to constrain unstaked networking keys to a specific format, and hence we will not need a `unittest.UnstakedNetworkingKey()`.

Also, I think if we do need something like `unittest.UnstakedNetworkingKey()`, I could actually just use the unstaked key generation function which you've already implemented :)

network/p2p/libp2pUtils.go
smnzhu: Yup, this was implemented before I realized it wouldn't work and I forgot to remove :)

network/p2p/topic_validator.go
huitseeker: Ah, I see the import cycle ...

network/p2p/libp2pNode.go
```diff
@@ -478,6 +483,19 @@ func (n *Node) Subscribe(ctx context.Context, topic flownet.Topic) (*pubsub.Subs
 		if err != nil {
 			return nil, fmt.Errorf("could not join topic (%s): %w", topic, err)
 		}
+
+		if len(validators) > 0 {
+			if err := n.pubSub.RegisterTopicValidator(
+				topic.String(), validators[0], pubsub.WithValidatorInline(true),
+			); err != nil {
```
huitseeker, line 490: I'm not sure this works if we use > 1 topic validator.

network/p2p/libp2pNode.go
```diff
@@ -514,8 +532,13 @@ func (n *Node) UnSubscribe(topic flownet.Topic) error {
 		return err
 	}
 
+	err := n.pubSub.UnregisterTopicValidator(topic.String())
+	if err != nil {
+		return fmt.Errorf("could not unregister topic validator for (%s): %w", topic, err)
+	}
+
```
huitseeker, line 539: as you'll see from CI, this fails if there's no validator on the topic (e.g. the subscribe call didn't keep any)

huitseeker (changes requested):
A couple of places where this is still breaking, putting it back in your queue for now (but we're not far, i think).

model/flow/identity.go
```diff
@@ -392,6 +392,16 @@ func (il IdentityList) ByNodeID(nodeID Identifier) (*Identity, bool) {
 	return nil, false
 }
 
+// ByNodeID gets a node from the list by network public key.
```
vishalchangrani, line 395:
```suggestion
// ByNetworkingKey gets a node from the list by network public key.
```

network/middleware.go
```diff
@@ -67,6 +69,9 @@ type Overlay interface {
 	// Identities returns a list of all Flow identities on the network
 	Identities() flow.IdentityList
 
+	// GetIdentity returns the Identity associated with the given peer ID, if it exists
+	GetIdentity(peer.ID) (*flow.Identity, bool)
```
vishalchangrani, line 73:
```suggestion
	Identity(peer.ID) (*flow.Identity, bool)
```
as per https://golang.org/doc/effective_go#Getters

network/p2p/libp2pNode.go
```diff
@@ -478,6 +483,19 @@ func (n *Node) Subscribe(ctx context.Context, topic flownet.Topic) (*pubsub.Subs
 		if err != nil {
 			return nil, fmt.Errorf("could not join topic (%s): %w", topic, err)
 		}
+
+		if len(validators) > 0 {
+			if err := n.pubSub.RegisterTopicValidator(
```
vishalchangrani, line 488: Can we register before we do a `n.pubSub.Join` ?

network/p2p/topic_validator.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (v *StakedValidator) Validate(ctx context.Context, receivedFrom peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+	// check that message contains a valid sender ID
+	if from, err := messageSigningID(msg); err == nil {
+		// check that the sender peer ID matches a staked Flow key
+		if _, exists := v.getIdentity(from); exists {
+			return pubsub.ValidationAccept
+		}
+	}
+
```
vishalchangrani, line 24: would suggest we log the sender to help audit these errors.

network/p2p/topic_validator.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/onflow/flow-go/model/flow"
+)
+
+type StakedValidator struct {
+	getIdentity func(peer.ID) (*flow.Identity, bool)
+}
+
+func (v *StakedValidator) Validate(ctx context.Context, receivedFrom peer.ID, msg *pubsub.Message) pubsub.ValidationResult {
+	// check that message contains a valid sender ID
+	if from, err := messageSigningID(msg); err == nil {
```
vishalchangrani, line 18: 1. Since `messageSigningID` is no longer exclusive to `readSubscription` can we move it to `libp2pUtils`?
2. Minor optimization - `messageSigningID` call will repeated in `readSubscription`. I noticed that libp2p pubsub message offers this field - [ValidatorData](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L215). We can stash the `peerID` in there and check for its presence in `readSubscription`.
3. How about a refactor-> rename to `signerIDFromMessage` ?

network/test/middleware_test.go
```diff
@@ -164,6 +164,9 @@ func (m *MiddlewareTestSuite) createOverlay() *mocknetwork.Overlay {
 	overlay := &mocknetwork.Overlay{}
 	overlay.On("Identities").Maybe().Return(m.getIds, nil)
 	overlay.On("Topology").Maybe().Return(m.getIds, nil)
+	// we're only dealing with staked IDs in this test, so queries based on PeerIDs should not succeeded
+	// (they will occur on the Spoofing test)
+	overlay.On("Identity").Return(nil, false)
```
vishalchangrani, line 169: Am confused 😕 , shouldn't it be the opposite that since all nodes are staked in the test, the call made by the TopicValidator to overlay.Identity should always succeed?

huitseeker: oh, indeed  - I thought this code path only triggered on unstaked nodes, but the topic validator is the one that's going to fire it.

network/p2p/topic_validator.go
huitseeker: on 2.: the messageSigningID is really supposed to be a cheap operation: look up a couple of fields, essentially. That's despite the scary name of `UnmarshalPublicKey` (which isn't actually costly).
 
I think there's a risk to over-optimize this.

network/p2p/libp2pNode.go
huitseeker: I see, it's a 1 || 0

huitseeker (dismissed):

network/p2p/libp2pNode.go
smnzhu: We can, sure.

network/p2p/topic_validator.go
smnzhu: @vishalchangrani this is great to know! very useful feature tbh, could have been used to implement origin validator as a topic validator instead of in `readSubscription`.

smnzhu: However, I think atm this is unnecessary, because I am hoping that https://github.com/onflow/flow-go/pull/1167 will eventually get adopted in some form, which means we will no longer need the origin validation.

huitseeker (approved):
LGTM!

network/p2p/middleware.go
```diff
@@ -383,7 +387,12 @@ func (m *Middleware) Subscribe(channel network.Channel) error {
 
 	topic := engine.TopicFromChannel(channel, m.rootBlockID)
 
-	s, err := m.libP2PNode.Subscribe(m.ctx, topic)
+	var validators []pubsub.ValidatorEx
+	if !engine.UnstakedChannels().Contains(channel) {
+		validators = append(validators, m.stakedTopicValidator.Validate)
```
vishalchangrani, line 392:
```suggestion
                 // for channels used by the staked nodes, add the topic validator to filter out messages from non-staked nodes
		validators = append(validators, m.stakedTopicValidator.Validate)
```

network/p2p/topic_validator_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/engine"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestTopicValidator(t *testing.T) {
```
vishalchangrani, line 19: added some comments here - https://github.com/onflow/flow-go/pull/1253

vishalchangrani: no change in functionality

vishalchangrani (approved):
thanks! lgtm.
I added a couple of commits, hope you don't mind.

smnzhu:
bors merge

Kay-Zee:
bors cancel

PR #1227, opened 2021-08-30
[ci, crypto] Detect CPU arch and feed that info to blst

huitseeker:
The Go bindings for BLST run with unconditional [-D__ADX__ flag](https://en.wikipedia.org/wiki/Intel_ADX). While that is fine for newer
machines, GH Actions run tests on a variety of underlying architectures, some of which do not have
those CPU extensions. See:
https://docs.microsoft.com/en-ca/azure/virtual-machines/dv2-dsv2-series#dsv2-series

As a consequence, this runs detection before setting CGO compilation flags approaprately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
         file: ./coverage.txt
         flags: unittests
         name: codecov-umbrella
-    
+
   integration-test:
     name: Integration Tests
     strategy:
@@ -103,6 +103,9 @@ jobs:
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
+    - name: CPU information
+      run: |
+        lscpu
     - name: Setup Go
       uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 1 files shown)


durkmurder (approved):

AlexHentschel (approved):

.github/workflows/ci.yml
```diff
@@ -113,11 +116,19 @@ jobs:
       run: make crypto/relic/build
     - name: Run tests
       if: github.actor != 'bors[bot]'
-      run: make -C crypto cross-blst-test
+      run: |
```
tarakby, line 119: Can't `lscpu` be run here before the `if` statement?

.github/workflows/ci.yml
```diff
@@ -113,11 +116,19 @@ jobs:
...
+        fi
+        make -C crypto cross-blst-test
     - name: Run tests (Bors)
       if: github.actor == 'bors[bot]'
       uses: nick-invision/retry@v2
       with:
         timeout_minutes: 25
         max_attempts: 2
-        command: make -C crypto cross-blst-test
+        command: |
```
tarakby, line 130: As a general comment, can the `if` statement be run locally too and not only on CI ?

If yes, it would be cleaner to have a new Makefile target `crypto/cross-blst` that encapsulates all this. 
Eventually, when we make sure this cross-blst tests always pass, the new target would be called by the unit-tests and we can remove the blst CI job

tarakby (approved):
Nice! Hopefully the flakiness is fixed and we can clean up the CI job.

.github/workflows/ci.yml
huitseeker: It could, but that would reduce the information we get about the platform we're running on to that CI job. `lscpu` info is more widely valuable IMHO, especially given the variety of machines used by GH Actions.

.github/workflows/ci.yml
huitseeker: I captured this idea in https://github.com/onflow/flow-go/issues/1229
In theory, yes we could. In practice, ADX was introduced with Broadwell, and is accepted in most modern machines. I believe in GH this will fail on the Xeon E5-2673 v3, so I'm wary of over-optimizing for a niche case. The perf of CI post this change will tell.

PR #1228, opened 2021-08-30
Vishal/consensus follower storage

vishalchangrani:
From @awfm9 
> We have another problem: we need to access the protocol state database in our code. However, since the DB is not injected into the consensus follower as a dependency, but rather you just pass the data directory as an option, the DB is opened internally by the follower's node builder.
> That makes it impossible for us to access the protocol state database, as it's not exposed in any meaningful way. Would it be possible to change that so that the node builder takes a database handle, and the consensus follower takes a *badger.DB as well, rather than a data directory?

This change allows the Badger db to be passed in as a dependency.

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -369,6 +369,13 @@ func (fnb *FlowNodeBuilder) initProfiler() {
 }
 
 func (fnb *FlowNodeBuilder) initDB() {
+
+	// if a db has been passed in, use that instead of creating one
+	if fnb.BaseConfig.db != nil {
+		fnb.DB = fnb.BaseConfig.db
+		return
+	}
+
 	// Pre-create DB path (Badger creates only one-level dirs)
 	err := os.MkdirAll(fnb.BaseConfig.datadir, 0700)
 	fnb.MustNot(err).Str("dir", fnb.BaseConfig.datadir).Msg("could not create datadir")
@@ -735,6 +742,12 @@ func WithMetricsEnabled(enabled bool) Option {
 func WithLogLevel(level string) Option {
 	return func(config *BaseConfig) {
 		config.level = level
+	}
+}
+
+func WithDB(db *badger.DB) Option {
+	return func(config *BaseConfig) {
+		config.db = db
 	}
 }
 
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -729,7 +729,9 @@ func WithBindAddress(bindAddress string) Option {
 
 func WithDataDir(dataDir string) Option {
 	return func(config *BaseConfig) {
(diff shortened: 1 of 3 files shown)


cmd/scaffold.go
```diff
@@ -717,6 +724,12 @@ func WithMetricsEnabled(enabled bool) Option {
 	}
 }
 
+func WithDB(db *badger.DB) Option {
```
huitseeker, line 727: Any way to integrate this somewhere along our existing integration tests?

huitseeker (dismissed):
Thanks!

smnzhu:
@vishalchangrani difficult to review this until identity provider is merged

vishalchangrani:
> @vishalchangrani difficult to review this until identity provider is merged

identity provider has been merged in to reflect only the things that have changed on this PR.

cmd/scaffold.go
```diff
@@ -369,6 +369,13 @@ func (fnb *FlowNodeBuilder) initProfiler() {
...
+
+	// if a db has been passed in, use that instead of creating one
+	if fnb.BaseConfig.db != nil {
+		fnb.DB = fnb.BaseConfig.db
+		return
+	}
+
 	// Pre-create DB path (Badger creates only one-level dirs)
 	err := os.MkdirAll(fnb.BaseConfig.datadir, 0700)
 	fnb.MustNot(err).Str("dir", fnb.BaseConfig.datadir).Msg("could not create datadir")
```
m4ksio, line 381: It would be good if we reject the `datadir` flag when DB is passed in. Ie. make sure its not set, to avoid dangling unused configs and potential confusion

m4ksio (approved):

smnzhu (approved):

follower/consensus_follower.go
```diff
@@ -27,16 +29,21 @@ type Config struct {
...
 }
 
 type Option func(c *Config)
 
+// WithDataDir sets the underlying directory to be used to store the database
+// If a database is supplied, then data directory will be set to empty string
 func WithDataDir(dataDir string) Option {
 	return func(cf *Config) {
-		cf.dataDir = dataDir
+		if cf.db == nil {
```
vishalchangrani, line 44: @m4ksio - modified to ignore the dataDir if db is passed in

cmd/scaffold.go
vishalchangrani: made that change - to ignore datadir if db is passed in.

awfm9:
We need this to get merged in order to finalize integration of the unstaked consensus follower into the DPS.

huitseeker (approved):
👍

huitseeker:
bors merge

PR #1233, opened 2021-08-31
CLI tool to validate machine account config

jordanschalm:
This PR adds a command to the `bootstrap` CLI to check the correctness of a machine account configuration. This lays the groundwork for #5792.

PR #1236, opened 2021-09-01
Validate machine account config at startup

jordanschalm:
This PR extends https://github.com/onflow/flow-go/pull/1233 to validate machine account configuration when a node requiring a machine account starts up.


module/epochs/machine_account.go
```diff
@@ -45,9 +55,106 @@ func init() {
...
+		}
+
+		err = CheckMachineAccountInfo(log, validator.role, validator.info, account)
+		if err != nil {
+			// either we cannot validate the configuration or there is a critical
+			// misconfiguration - log a warning and retry - we will continue checking
+			// and logging until the problem is resolved
+			log.Warn().
+				Err(err).
+				Msg("critical machine account misconfiguration")
```
huitseeker, line 137: Nit: shouldn't a critical misconfiguration be logged to the `Error` channel?

cmd/collection/main.go
```diff
@@ -492,26 +506,15 @@ func createQCContractClient(node *cmd.NodeConfig, accessAddress string, flowClie
...
-	// if not valid return a mock qc contract client
-	if valid := cmd.IsValidNodeMachineAccountConfig(node, accessAddress); !valid {
-		return nil, fmt.Errorf("could not validate node machine account config")
-	}
-
-	// attempt to read NodeMachineAccountInfo
-	info, err := cmd.LoadNodeMachineAccountInfoFile(node.BaseConfig.BootstrapDir, node.Me.NodeID())
-	if err != nil {
-		return nil, fmt.Errorf("could not load node machine account info file: %w", err)
-	}
```
huitseeker, line 504: That's indeed in `LoadNodeMachineAccountInfoFile` already, 👍

huitseeker (approved):
Very no-panic compliant, I like this (perhaps somewhat perversely). Left one nit.

module/epochs/machine_account.go
jordanschalm: Definitely should be, updated here: https://github.com/onflow/flow-go/pull/1236/commits/df22d9b467815b81e6fa6565878a35bc16ba0c5d

module/epochs/machine_account.go
```diff
@@ -45,9 +55,107 @@ func init() {
...
+
+		err = CheckMachineAccountInfo(log, validator.role, validator.info, account)
+		if err != nil {
+			// either we cannot validate the configuration or there is a critical
+			// misconfiguration - log a warning and retry - we will continue checking
+			// and logging until the problem is resolved
+			log.Error().
+				Err(err).
+				Msg("critical machine account misconfiguration")
+			return retry.RetryableError(err)
```
kc1116, line 138: Do we need to retry if checking the machine info fails ? What will change about the info between retries ?

kc1116 (approved):
Left a small comment

module/epochs/machine_account.go
jordanschalm: Some failures can be resolved out-of-band and would be addressed after retrying. For example, a temporary failure to connect to the access node, or an insufficient balance. 

For the failures that can't be resolved without a restart, if we aren't going to crash, it's better to periodically log so that it's easier to find this problem (eg. a query for this log over the last hour would still find it).

jordanschalm:
bors merge

PR #1237, opened 2021-09-01
removing timeout when starting an node

vishalchangrani:
A node startup can take a very long time for certain node type e.g. Execution node. A timeout is not really needed when a node starts/stop.


huitseeker (commented):
I'm wondering about the wisdom between no timeout and a very long one (10-30 min?). DOn't we use this in tests too?

Kay-Zee:
do we use the timeout for tests? I'm not really following the logic of needing a startup timeout for a node for normal operations, but could be convinced for tests

vishalchangrani:
T-systems have reported that they see a time out on node startup and their node enters a restart loop.

> Hi @vishal we started the node with actual parameters and image, but we receive always this: {"level":"fatal","node_role":"execution","node_id":"2b396b7fab0102f104a2af7e095b145cc14da28f863564802e158afc3e07e638","time":"2021-09-17T10:10:22Z","message":"node startup timed out"}

vishalchangrani:
@smnzhu @huitseeker @Kay-Zee bumping up this PR.

huitseeker (approved):
Accepting to unblock, it seems clear the current timeout is too small, at least.

peterargue (commented):
From an operational perspective, it seems like having some timeout makes sense, though definitely much longer than 1 minute. If a node hangs on initial startup or a later restart, operators will want their supervisor tasks to restart the process. if it hangs for a long time/indefinitely they'll require manual intervention.

I'm not sure about what startup/shutdown step could possibly hang that long, so I'm fine with this if long hangs are not a concern.

Kay-Zee (approved):

Kay-Zee:
bors merge

PR #1243, opened 2021-09-02
[All] Entity-centric tracing 

ramtinms:
This PR 
- refactors tracing to be entity centric
  - for entities (block, collection, transaction) entityID is going to be set as traceID 
  - a dummy RootEntity span is also constructed to help keeping spans of an entity generated from the same node together and aligned 
  - most of new methods returns isSampled to avoid extra computations for tag values
  - provides in-house proper entity sampling so we don't have to tune params from Jaeger
- removes non-operational traces (e.g. time between two events), those are going to be captured differently in the future PRs
- refactors components to pass trace spans using contexts
- injects tracer into Finalizer 
- removes consensusTracingConsumer as it was not in use


module/metrics/execution.go
```diff
@@ -484,13 +484,15 @@ func NewExecutionCollector(tracer module.Tracer, registerer prometheus.Registere
 // StartBlockReceivedToExecuted starts a span to trace the duration of a block
 // from being received for execution to execution being finished
 func (ec *ExecutionCollector) StartBlockReceivedToExecuted(blockID flow.Identifier) {
-	ec.tracer.StartSpan(blockID, executionBlockReceivedToExecuted).SetTag("block_id", blockID.String)
```
ramtinms, line 487: we need to define a metric instead of tracing here, e.g. blockExecutionWaitTime

cmd/scaffold.go
```diff
@@ -291,7 +291,8 @@ func (fnb *FlowNodeBuilder) initMetrics() {
 
 	fnb.Tracer = trace.NewNoopTracer()
 	if fnb.BaseConfig.tracerEnabled {
-		tracer, err := trace.NewTracer(fnb.Logger, fnb.BaseConfig.NodeRole)
+		serviceName := fnb.BaseConfig.NodeRole + "-" + fnb.BaseConfig.nodeIDHex[:8]
+		tracer, err := trace.NewTracer(fnb.Logger, serviceName)
```
ramtinms, line 295: We consider each node role + the first 8 digits of the id as service id.

state/protocol/badger/mutator.go
```diff
@@ -354,10 +347,10 @@ func (m *MutableState) sealExtend(candidate *flow.Block) (*flow.Seal, error) {
 //   * Receipts should pass the ReceiptValidator check
 //   * No seal has been included for the respective block in this particular fork
 // We require the receipts to be sorted by block height (within a payload).
-func (m *MutableState) receiptExtend(candidate *flow.Block) error {
-	blockID := candidate.ID()
```
ramtinms, line 358: removal of these blockIDs should make things way faster

module/trace/metrics.go
```diff
@@ -1,14 +0,0 @@
-package trace
-
-import (
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
-)
-
-var (
-	spanDurationMetric = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
ramtinms, line 9: this metric was not exposed before, and I was not sure what would be the purpose as spans might have different sizes

module/metrics/consensus.go
```diff
@@ -83,22 +70,18 @@ func NewConsensusCollector(tracer module.Tracer, registerer prometheus.Registere
 
 // StartCollectionToFinalized reports Metrics C1: Collection Received by CCL→ Collection Included in Finalized Block
 func (cc *ConsensusCollector) StartCollectionToFinalized(collectionID flow.Identifier) {
-	cc.tracer.StartSpan(collectionID, consensusCollectionToFinalized)
 }
 
 // FinishCollectionToFinalized reports Metrics C1: Collection Received by CCL→ Collection Included in Finalized Block
 func (cc *ConsensusCollector) FinishCollectionToFinalized(collectionID flow.Identifier) {
```
janezpodhostnik, line 76: None of the implementation of these methods have a body now. Should they be removed? or is this something that will be used in the future?

module/trace/trace.go
```diff
@@ -84,51 +99,107 @@ func (t *OpenTracer) Done() <-chan struct{} {
...
+		traceID,
+		jaeger.SpanID(rand.Uint64()),
+		jaeger.SpanID(0),
+		true,
+		nil,
+	)
+	opts = append(opts, jaeger.SelfRef(ctx))
+	span := t.Tracer.StartSpan(string(entityType), opts...)
+	t.spanCache.Add(entityID, span)
+	span.Finish() // finish span right away
```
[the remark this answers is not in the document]
ramtinms: this is going to be root span, won't have any meaning would only be there to aggregate all spans under the same root, so we close the span anyway

engine/consensus/matching/core.go
```diff
@@ -141,11 +142,13 @@ func (c *Core) ProcessReceipt(receipt *flow.ExecutionReceipt) error {
 //   internal state might be corrupted. Hence, returned errors should be treated as fatal.
 func (c *Core) processReceipt(receipt *flow.ExecutionReceipt) (bool, error) {
 	startTime := time.Now()
-	receiptSpan := c.tracer.StartSpan(receipt.ID(), trace.CONMatchProcessReceipt)
-	defer func() {
-		c.metrics.OnReceiptProcessingDuration(time.Since(startTime))
-		receiptSpan.Finish()
-	}()
+	defer c.metrics.OnReceiptProcessingDuration(time.Since(startTime))
```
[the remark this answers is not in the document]
ramtinms: nice catch, fixed it

module/metrics/consensus.go
ramtinms: but still has to satisfy the requirements for ConsensusMetrics interface

model/flow/identifier.go
```diff
@@ -64,6 +65,17 @@ func (id Identifier) Format(state fmt.State, verb rune) {
...
+func (id Identifier) IsSampled(sensitivity uint) bool {
+	if sensitivity > 64 {
+		return false
+	}
+	// take the first 8 bytes and check the first few bits based on sensitivity
+	// higher sensitivity means more bits has to be zero, means less number of samples
+	// sensitivity of zero, means everything is sampled
+	return binary.BigEndian.Uint64(id[:8])>>uint64(64-sensitivity) == 0
+}
+
```
zhangchiqing, line 78: Better to let user define the sample rate with percentage instead of sensitivity

```suggestion

// IsSampledWithPercent is a utility method to sample entities based on their ids, the sampling rate is 
// percentage based.
// 0 means 0 data will be collected.
// 100 means all data will be collected.
func (id Identitifier) IsSampledWithPercent(percentage uint) bool {
  sensitivity := percentageToSensitivity(percentage)
  return IsSample(sensitivity)
}

func percentageToSensitivity(percentage uint) uint {
  // to be implemented
}
```

module/tracer.go
```diff
@@ -10,14 +10,41 @@ import (
...
+var _ Tracer = &trace.NoopTracer{}
 
 // Tracer interface for tracers in flow. Uses open tracing span definitions
 type Tracer interface {
 	ReadyDoneAware
-	StartSpan(entity flow.Identifier, spanName trace.SpanName, opts ...opentracing.StartSpanOption) opentracing.Span
-	FinishSpan(entity flow.Identifier, spanName trace.SpanName)
-	GetSpan(entity flow.Identifier, spanName trace.SpanName) (opentracing.Span, bool)
+
+	// StartBlockSpan starts an span for a block, built as a child of rootSpan
```
zhangchiqing, line 22: When creating a span for a block, it will be sampled by the block ID. In other words, if a block is sampled, all spans for a certain block will be sampled.

This is great. However, I'm concerned about two issues:

1. If the network is big, i.e. 400 nodes, then even with sampling, then we will still aggregate spans from all of the nodes to the same span. Would it be hard to view all spans from all nodes on the chart? Can we group spans by block ID first, then node type, and then node ID?
Also with so many nodes having the same node type, I wonder if possible we can sample by node type as well. For instance, if there are 100 verification nodes, we could only pick 3 of them to report their spans. Because most likely, similar node will have similar tracing results.
The consensus nodes might be different, because the leader and next leader might have different tracing data.
So I'm not sure how easy to include the leader, next leader, and a sample of followers. :D 

2. All transaction spans are grouped under the same block, since we will be most likely looking at blocks having lots of transactions to dig into performance issue, then would it easy to open such block on the chart?

engine/consensus/matching/core.go
```diff
@@ -141,12 +142,17 @@ func (c *Core) ProcessReceipt(receipt *flow.ExecutionReceipt) error {
...
 	startTime := time.Now()
-	receiptSpan := c.tracer.StartSpan(receipt.ID(), trace.CONMatchProcessReceipt)
 	defer func() {
 		c.metrics.OnReceiptProcessingDuration(time.Since(startTime))
-		receiptSpan.Finish()
 	}()
 
+	receiptSpan, _, isSampled := c.tracer.StartBlockSpan(context.Background(), receipt.ExecutionResult.BlockID, trace.CONMatchProcessReceipt)
+	if isSampled {
+		receiptSpan.LogFields(log.String("result_id", receipt.ExecutionResult.ID().String()))
```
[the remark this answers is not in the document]
ramtinms: LogFields is a type version of LogKV, SetTag is mostly used for filtering traces with different traceIDs (secondary index) reasons on the UI (e.g. HTTPCode returned for API calls), for most of the cases we don't need SetTag so we can avoid extra indexing cost, I'll update most of the things to use LogFields unless useful as a tag

engine/consensus/sealing/core.go
```diff
@@ -342,13 +344,13 @@ func (c *Core) checkBlockOutdated(blockID flow.Identifier) error {
 // * exception in case of unexpected error
 // * nil - successfully processed result approval
 func (c *Core) ProcessApproval(approval *flow.ResultApproval) error {
-	startTime := time.Now()
-	approvalSpan := c.tracer.StartSpan(approval.ID(), trace.CONSealingProcessApproval)
 
-	err := c.processApproval(approval)
+	span, _, _ := c.tracer.StartBlockSpan(context.Background(), approval.Body.BlockID, trace.CONSealingProcessApproval)
```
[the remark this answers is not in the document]
ramtinms: Each span would be captured separately but all sharing the same parent, we use uuid for spanID.  I can add approvalID as an extra field log

engine/consensus/compliance/core.go
```diff
@@ -77,22 +78,14 @@ func NewCore(
...
-	defer onBlockProposalSpan.Finish()
 
-	for _, g := range proposal.Payload.Guarantees {
-		if span, ok := c.tracer.GetSpan(g.CollectionID, trace.CONProcessCollection); ok {
-			childSpan := c.tracer.StartSpanFromParent(span, trace.CONCompOnBlockProposal)
-			defer childSpan.Finish()
-		}
+	span, _, isSampled := c.tracer.StartBlockSpan(context.Background(), proposal.Header.ID(), trace.CONCompOnBlockProposal)
+	if isSampled {
+		span.SetTag("block_id", proposal.Header.ID())
```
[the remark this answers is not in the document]
ramtinms: right, it was a legecy thing, just removed it.

engine/consensus/compliance/core.go
```diff
@@ -77,22 +78,14 @@ func NewCore(
...
+		span.SetTag("view", proposal.Header.View)
```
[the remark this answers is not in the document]
ramtinms: height is the view here right? or you mean the current blockHeight when the proposal is received ?

engine/execution/computation/computer/computer.go
```diff
@@ -332,31 +327,29 @@ func (e *blockComputer) executeTransaction(
...
-	// call tracing
-	txSpan = e.tracer.StartSpanFromParent(colSpan, trace.EXEComputeTransaction)
+	txID := txBody.ID()
 
-	if sc, ok := txSpan.Context().(jaeger.SpanContext); ok {
-		traceID = sc.TraceID().String()
-	}
+	// we capture two spans one for tx-based view and one for the current context (block-based) view
+	txSpan := e.tracer.StartSpanFromParent(colSpan, trace.EXEComputeTransaction)
+	txSpan.LogFields(log.String("transaction.ID", txID.String()))
```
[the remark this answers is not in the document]
ramtinms: added txIndex and collection index

engine/execution/computation/computer/computer.go
```diff
@@ -332,31 +327,29 @@ func (e *blockComputer) executeTransaction(
...
 
-	defer func() {
-		txSpan.LogFields(
-			log.String("transaction.ID", txBody.ID().String()),
-		)
-		txSpan.Finish()
-	}()
+	txInternalSpan, _, isSampled := e.tracer.StartTransactionSpan(context.Background(), txID, trace.EXERunTransaction)
+	if isSampled {
+		txInternalSpan.LogFields(log.String("transaction.ID", txID.String()))
```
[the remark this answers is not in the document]
ramtinms: actually, since traceID is only subset of transactionId, I kept it here so we could look up transactions that are taking long

engine/execution/ingestion/engine.go
```diff
@@ -1056,6 +1061,9 @@ func (e *Engine) handleComputationResult(
 	startState flow.StateCommitment,
 ) (flow.StateCommitment, *flow.ExecutionReceipt, error) {
 
+	span, ctx := e.tracer.StartSpanFromContext(ctx, trace.EXEHandleComputationResult)
```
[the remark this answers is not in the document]
ramtinms: we don't since the parent from the context has the same traceId, so the child span would share the same traceID

engine/verification/requester/requester.go
```diff
@@ -165,17 +164,15 @@ func (e *Engine) process(originID flow.Identifier, event interface{}) error {
 
 // handleChunkDataPackWithTracing encapsulates the logic of handling a chunk data pack with tracing enabled.
 func (e *Engine) handleChunkDataPackWithTracing(originID flow.Identifier, chunkDataPack *flow.ChunkDataPack) {
-	span, ok := e.tracer.GetSpan(chunkDataPack.ChunkID, trace.VERProcessChunkDataPackRequest)
-	if !ok {
-		span = e.tracer.StartSpan(chunkDataPack.ChunkID, trace.VERProcessChunkDataPackRequest)
-		span.SetTag("chunk_id", chunkDataPack.ChunkID)
+	// TODO: change this to block level as well
```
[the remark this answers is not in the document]
ramtinms: yeah, we can't do it for now.

engine/verification/verifier/engine.go
```diff
@@ -315,13 +314,13 @@ func (e *Engine) GenerateResultApproval(chunkIndex uint64,
...
-	if span, ok := e.tracer.GetSpan(ch.Result.ID(), trace.VERProcessExecutionResult); ok {
-		defer span.Finish()
-		childSpan := e.tracer.StartSpanFromParent(span, trace.VERVerVerifyWithMetrics)
-		ctx = opentracing.ContextWithSpan(ctx, childSpan)
-		defer childSpan.Finish()
+
+	span, ctx, isSampled := e.tracer.StartBlockSpan(context.Background(), ch.Chunk.BlockID, trace.VERVerVerifyWithMetrics)
+	if isSampled {
+		span.LogKV("result_id", ch.Result.ID())
+		span.LogKV("chunk_index", ch.Chunk.Index)
```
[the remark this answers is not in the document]
ramtinms: not sure those would be useful, knowing blockID and chunkIndex should be enough to load the rest I guess

model/flow/identifier.go
ramtinms: I can add this as well, but prefer to keep the one used by tracing to not do any extra computation if possible, since this method would be called a lot.

engine/consensus/compliance/core.go
```diff
@@ -267,18 +258,15 @@ func (c *Core) processBlockAndDescendants(proposal *messages.BlockProposal) erro
...
-		blockSpan.SetTag("block_id", proposal.Header.ID())
-		blockSpan.SetTag("view", proposal.Header.View)
-		blockSpan.SetTag("proposer", proposal.Header.ProposerID.String())
+	defer c.complianceMetrics.BlockProposalDuration(time.Since(startTime))
+
+	span, ctx, isSampled := c.tracer.StartBlockSpan(context.Background(), proposal.Header.ID(), trace.ConCompProcessBlockProposal)
+	if isSampled {
+		span.SetTag("block_id", proposal.Header.ID())
+		span.SetTag("view", proposal.Header.View)
+		span.SetTag("proposer", proposal.Header.ProposerID.String())
```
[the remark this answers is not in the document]
ramtinms: yeah, we don't need them. going to keep them only in one place

engine/consensus/compliance/core.go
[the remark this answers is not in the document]
zhangchiqing: view and height are different, view is hotstuff field, height is the block height. I think we usually care more about height than the view

module/trace/trace.go
```diff
@@ -88,51 +103,107 @@ func (t *OpenTracer) Done() <-chan struct{} {
...
+	if !collectionID.IsSampled(t.sensitivity) {
+		return &NoopSpan{&NoopTracer{}}, ctx, false
+	}
+
+	rootSpan := t.entityRootSpan(collectionID, EntityTypeCollection)
+	ctx = opentracing.ContextWithSpan(ctx, rootSpan)
+	return t.StartSpanFromParent(rootSpan, spanName, opts...), ctx, true
+}
+
+func (t *OpenTracer) StartTransactionSpan(
```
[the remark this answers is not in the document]
ramtinms: updated.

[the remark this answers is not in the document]
ramtinms: updated.

module/trace/trace.go
```diff
@@ -88,51 +103,107 @@ func (t *OpenTracer) Done() <-chan struct{} {
...
+	if parentSpan == nil {
+		return &NoopSpan{&NoopTracer{}}, ctx
+	}
+	if _, ok := parentSpan.(*NoopSpan); ok {
+		return &NoopSpan{&NoopTracer{}}, ctx
+	}
+
+	opts = append(opts, opentracing.ChildOf(parentSpan.Context()))
+	span := t.Tracer.StartSpan(string(operationName), opts...)
+	return span, opentracing.ContextWithSpan(ctx, span)
```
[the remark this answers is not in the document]
ramtinms: actually I fixed the bug, but FromParent would use Child relationship but fromContext would use FollowsFrom relationship , right now some parts of our code uses context but that should be fine for now.

module/tracer.go
ramtinms: yeah we could filter by service name, the idea is we only keep a sample of node running tracing for now. we could change the logic in the future. 

Transaction Spans are not grouped under same block, they form their own group of spans under a root span with traceId of transactionID.

model/flow/identifier.go
```diff
@@ -64,6 +66,16 @@ func (id Identifier) Format(state fmt.State, verb rune) {
 	}
 }
 
+// IsSampled is a utility method to sample entities based on their ids,
+// percentage adjust how often we sample the entity.
+// 0 means 0 data will be collected.
+// 50 means 50% of the entities would be collected.
+// 100 means all data will be collected.
+func (id Identifier) IsSampled(percentage uint) bool {
```
zhangchiqing, line 74: can we add a test case for this function?

cmd/scaffold.go
```diff
@@ -322,7 +324,8 @@ func (fnb *FlowNodeBuilder) initMetrics() {
 
 	fnb.Tracer = trace.NewNoopTracer()
 	if fnb.BaseConfig.tracerEnabled {
-		tracer, err := trace.NewTracer(fnb.Logger, fnb.BaseConfig.NodeRole)
+		serviceName := fnb.BaseConfig.NodeRole + "-" + fnb.BaseConfig.nodeIDHex[:8]
```
m4ksio, line 327: I wonder if we could maybe add network name to avoid confusion, I could imagine we can have multiple nodes (on sporks or networks) running at the same time

engine/execution/computation/computer/computer.go
```diff
@@ -332,31 +328,35 @@ func (e *blockComputer) executeTransaction(
...
+	}
+	defer txInternalSpan.Finish()
 
 	e.log.Debug().
 		Hex("tx_id", logging.Entity(txBody)).
 		Msg("executing transaction")
 
-	txView := collectionView.NewChild()
-
 	tx := fvm.Transaction(txBody, txIndex)
```
m4ksio, line 354: Don't we want to move this below trace span as well?

zhangchiqing (approved):
Nice improvement!

engine/execution/state/state.go
```diff
@@ -334,7 +334,7 @@ func (s *state) PersistExecutionState(ctx context.Context, header *flow.Header,
 	spew.Config.DisableMethods = true
 	spew.Config.DisablePointerMethods = true
 
-	span, childCtx := s.tracer.StartSpanFromContext(ctx, trace.EXESaveExecutionResults)
+	span, childCtx := s.tracer.StartSpanFromContext(ctx, trace.EXEStatePersistExecutionState)
```
m4ksio, line 337: I think old name was "more correct", we don't persist executions state here, we just save results (exec state is persisted async in commiter)

engine/consensus/compliance/core.go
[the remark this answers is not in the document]
ramtinms: but how do I get the height from the proposal?

model/flow/identifier.go
```diff
@@ -64,6 +66,16 @@ func (id Identifier) Format(state fmt.State, verb rune) {
...
+	// take the first 8 bytes and check the first few bits based on the percentage
+	return binary.BigEndian.Uint64(id[:8])>>(64-int(math.Min(float64(percentage)/float64(1.5), 64))) == 0
```
m4ksio: Not sure why we need our own way of randomness?

model/flow/identifier.go
m4ksio: I think using float with range 0..1.0 is more common

engine/consensus/compliance/core.go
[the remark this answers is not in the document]
zhangchiqing: [from the header](https://github.com/onflow/flow-go/pull/1243/files/c45ea80ea859458dcf93a28d7b8979ce2a5ff66d#diff-91e83ea2456f635c82eea90128d3f2820974b95947215b9a75da64c0f513b5e5R274)

m4ksio (approved):
I really like the change, but (there is always a but) - I really think we should have some tests for it.
Even adding a tracer for existing tests and asserting is traces matches what we would expect will greatly increase confidence

engine/execution/state/state.go
ramtinms: I was trying to follow the name of the method , then we need to update that one as well.

m4ksio: Yes, the method used to save execution state, both should be renamed then

cmd/scaffold.go
ramtinms: I think the node Id and blockIDs are unique per networks, not sure we would hit that problem.

engine/execution/computation/computer/computer.go
ramtinms: not sure if I understand? you mean the NewChild() call?

model/flow/identifier.go
ramtinms: cause we wrap the open tracer inside the tracer module, and when we reach to the tracer sampling it is already late and we have done alot of extra work

ramtinms:
@m4ksio this PR is already really big, going to add tests for the tracers in a follow up PR to this one.

engine/execution/state/state.go
ramtinms: got it, going renaming both.

engine/verification/fetcher/engine.go
```diff
@@ -164,22 +163,14 @@ func (e *Engine) ProcessAssignedChunk(locator *chunks.Locator) {
...
 func (e *Engine) processAssignedChunkWithTracing(chunk *flow.Chunk, result *flow.ExecutionResult, chunkLocatorID flow.Identifier) (bool, uint64, error) {
-	chunkID := chunk.ID()
 
-	span, ok := e.tracer.GetSpan(chunkID, trace.VERProcessAssignedChunk)
-	if !ok {
-		span = e.tracer.StartSpan(chunkID, trace.VERProcessAssignedChunk)
-		span.SetTag("chunk_id", chunkID)
-		defer span.Finish()
+	span, _, isSampled := e.tracer.StartBlockSpan(e.unit.Ctx(), result.BlockID, trace.VERProcessAssignedChunk)
+	if isSampled {
```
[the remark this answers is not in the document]
ramtinms: sampling is based on blockID

engine/verification/fetcher/engine.go
```diff
@@ -262,7 +254,10 @@ func (e *Engine) HandleChunkDataPack(originID flow.Identifier, chunkDataPack *fl
 		Bool("system_chunk", IsSystemChunk(status.ChunkIndex, status.ExecutionResult)).
 		Logger()
 
-	processed, err := e.handleChunkDataPackWithTracing(originID, status, chunkDataPack)
+	span, ctx, _ := e.tracer.StartBlockSpan(context.Background(), status.ExecutionResult.BlockID, trace.VERFetcherHandleChunkDataPack)
+	defer span.Finish()
```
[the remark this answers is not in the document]
ramtinms: I think thats fine for now. the trace captures everything happens on "handleChunkDataPack" method.

engine/execution/computation/computer/computer.go
```diff
@@ -93,15 +91,12 @@ func (e *blockComputer) ExecuteBlock(
...
-		span.SetTag("block.collectioncount", len(block.CompleteCollections))
-		span.LogFields(
-			log.String("block.hash", block.ID().String()),
-		)
-		span.Finish()
-	}()
+	span, _, isSampled := e.tracer.StartBlockSpan(ctx, block.ID(), trace.EXEComputeBlock)
+	if isSampled {
+		span.LogFields(log.Int("collection_counts", len(block.CompleteCollections)))
+		span.LogFields(log.String("block_id", block.ID().String()))
```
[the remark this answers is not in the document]
ramtinms: removed it.

engine/execution/computation/computer/computer.go
ramtinms: SetTraceSpan won't start a trace, it would only pass the parent span reference to transaction object

module/trace/trace.go
```diff
@@ -88,51 +103,107 @@ func (t *OpenTracer) Done() <-chan struct{} {
...
+// entityRootSpan returns the root span for the given entity from the cache
+// and if not exist it would construct it and cache it and return it
+// This should be used mostly for the very first span created for an entity on the service
+func (t *OpenTracer) entityRootSpan(entityID flow.Identifier, entityType string, opts ...opentracing.StartSpanOption) opentracing.Span {
+	if span, ok := t.spanCache.Get(entityID); ok {
+		return span.(opentracing.Span)
+	}
+
+	// flow.Identifier to flow
+	traceID, err := jaeger.TraceIDFromString(entityID.String()[:32])
```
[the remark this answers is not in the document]
ramtinms: yeah, there is a limit.

cmd/scaffold.go
ramtinms: I did refactor so the root span now includes a chainID tag, so we can filter out traces.

PR #1246, opened 2021-09-02
refactoring the unstaked consensus follower integration test

vishalchangrani:
making it a little DRY and easy to read. No functional changes.


integration/tests/access/unstaked_node_test.go
```diff
@@ -41,108 +41,74 @@ func TestUnstakedAccessSuite(t *testing.T) {
...
+		suite.followerMgr2.startFollower(ctx)
 
-	receiveBlocks2 := func() {
-		for _, blockID := range receivedBlocks {
-			select {
-			case receivedBlockID := <-suite.finalizedBlockIDsChans[1]:
-				suite.Assert().Equal(blockID, receivedBlockID)
+		// the second follower is now atleast blockCount blocks behind and should sync up and get all the missed blocks
+		receiveBlocks := func() {
+			for i := 0; ; i++ {
```
vishalchangrani, line 99: blocks may not arrive in sequence and hence the infinite loop.

integration/tests/access/unstaked_node_test.go
```diff
@@ -41,108 +41,76 @@ func TestUnstakedAccessSuite(t *testing.T) {
...
-		"timed out while waiting for consensus follower to start")
+		// kick off the second follower
+		suite.followerMgr2.startFollower(ctx)
 
-	receiveBlocks2 := func() {
-		for _, blockID := range receivedBlocks {
-			select {
-			case receivedBlockID := <-suite.finalizedBlockIDsChans[1]:
-				suite.Assert().Equal(blockID, receivedBlockID)
+		// the second follower is now atleast blockCount blocks behind and should sync up and get all the missed blocks
```
huitseeker, line 97: Those few lines make the PR worth it IMHO ✨

huitseeker (approved):
Thanks for the refactor :)

integration/tests/access/unstaked_node_test.go
```diff
@@ -41,108 +41,76 @@ func TestUnstakedAccessSuite(t *testing.T) {
...
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case blockID := <-suite.followerMgr2.blockIDChan:
+					delete(receivedBlocks, blockID)
+					if len(receivedBlocks) == 0 {
+						return
+					}
+				}
```
smnzhu, line 108: Hmm, what is the benefit of doing it this way compared to using an array?

Is it because of possible race condition?

integration/tests/access/unstaked_node_test.go
```diff
@@ -214,3 +182,54 @@ func UnstakedNetworkingKey() (crypto.PrivateKey, error) {
...
+		"timed out while waiting for consensus follower to start")
+}
+
+func (fm *followerManager) onBlockFinalizedConsumer(finalizedBlockID flow.Identifier) {
+	// push the finalized block ID to the blockIDChannel channel
+	fm.blockIDChan <- finalizedBlockID
+}
+
+// hasBlock checks if the underlying storage of the consensus follower has a block
+func (fm *followerManager) hasBlock(blockID flow.Identifier) (bool, error) {
```
smnzhu, line 224: I think the semantics of this method is a little weird.

imo, a method called `hasBlock` sounds like it should just return a boolean indicating whether we have the block or not. In the test above, you completely ignore the first return value, which sort of indicates imo that we should change this method.

An alternative, is to instead define a method called `getBlock` which returns `(block, error)`.

smnzhu (approved):

integration/tests/access/unstaked_node_test.go
vishalchangrani: benefit was that is makes it easy to just mark off the block IDs that have been received. Array would make require a `find-then-delete` pattern.
The next block ID will be pulled from the channel only after the current one has been deleted or ignored since this loop is run by single thread.

vishalchangrani:
bors merge

PR #1248, opened 2021-09-03
[Synchronization] Message queue for collection sync engine

durkmurder:
https://github.com/dapperlabs/flow-go/issues/5807

### Context

This PR implements message queues for collection sync engine. Design and implementation is inspired by consensus syng engine, basically same engine is used with needed modifications for collection clusters.


engine/collection/synchronization/engine.go
```diff
@@ -372,72 +334,64 @@ CheckLoop:
...
-
+		case <-pollChan:
+			e.pollHeight()
 		case <-scan.C:
 			final, err := e.state.Final().Head()
 			if err != nil {
-				e.log.Error().Err(err).Msg("could not get final height")
-				continue
+				e.log.Error().Err(err).Msg("could not get last finalized header")
+				return
```
jordanschalm, line 343: should this be fatal if we're going to exit here? Stopping the polling process can cause the node to get permanently out of sync

engine/collection/synchronization/engine.go
```diff
@@ -54,310 +67,259 @@ func New(
...
 // checkLoop will regularly scan for items that need requesting.
 func (e *Engine) checkLoop() {
-	poll := time.NewTicker(e.pollInterval)
+	pollChan := make(<-chan time.Time)
+	if e.pollInterval > 0 {
+		poll := time.NewTicker(e.pollInterval)
+		pollChan = poll.C
+		defer poll.Stop()
+	}
 	scan := time.NewTicker(e.scanInterval)
```
jordanschalm, line 323: Any reason not to treat `pollInterval` and `scanInterval` the same here? IE handle negative values, and `defer <ticker>.Stop()` for both?

engine/collection/synchronization/engine.go
```diff
@@ -372,72 +334,64 @@ CheckLoop:
...
 	for _, ran := range ranges {
 		req := &messages.RangeRequest{
 			Nonce:      rand.Uint64(),
 			FromHeight: ran.From,
 			ToHeight:   ran.To,
 		}
-		err := e.conduit.Multicast(req, synccore.DefaultBlockRequestNodes, e.participants.NodeIDs()...)
+		err := e.con.Multicast(req, synccore.DefaultBlockRequestNodes, e.participants.NodeIDs()...)
 		if err != nil {
-			errs = multierror.Append(errs, fmt.Errorf("could not submit range request (from=%d, to=%d): %w", ran.From, ran.To, err))
```
jordanschalm, line 438: Maybe we could add the more informational error message back (and add to consensus engine as well)

engine/collection/synchronization/engine.go
```diff
@@ -447,14 +401,16 @@ func (e *Engine) sendRequests(ranges []flow.Range, batches []flow.Batch) error {
 			Nonce:    rand.Uint64(),
 			BlockIDs: batch.BlockIDs,
 		}
-		err := e.conduit.Multicast(req, synccore.DefaultBlockRequestNodes, e.participants.NodeIDs()...)
+		err := e.con.Multicast(req, synccore.DefaultBlockRequestNodes, e.participants.NodeIDs()...)
 		if err != nil {
-			errs = multierror.Append(errs, fmt.Errorf("could not submit batch request (size=%d): %w", len(batch.BlockIDs), err))
```
jordanschalm, line 452: Maybe we could add the more informational error message back (and add to consensus engine as well)

jordanschalm (approved):
Thanks for this Yurii. 

If you haven't already, I think we should smoke test forcing a collection node to catch up with the sync engine on `localnet` as well with these changes.

engine/collection/synchronization/engine.go
durkmurder: For execution nodes consensus sync engine passed 0, meaning it doesn't want to poll height. 
I don't know if it's needed for collection nodes but no harm in keeping it I guess

engine/collection/synchronization/engine.go
durkmurder: I have replaced it with `continue` but my feeling is that we should crash the node at this point. Not retrieving state seems like a fatal error to me.

jordanschalm: > my feeling is that we should crash the node at this point

I agree actually, I think it's better to change the log level to `Fatal`

smnzhu (approved):

engine/collection/synchronization/engine.go
```diff
@@ -54,310 +67,259 @@ func New(
...
+	opts ...commonsync.OptionFunc,
 ) (*Engine, error) {
 
+	opt := commonsync.DefaultConfig()
+	for _, f := range opts {
+		f(opt)
+	}
+
+	if comp == nil {
+		panic("must initialize synchronization engine with comp engine")
```
yhassanzadeh13, line 79: Returning an error would be more proper for handling the failures.

engine/collection/synchronization/engine.go
smnzhu: should we add the `defer scanStop()` though?

engine/collection/synchronization/engine.go
```diff
@@ -372,72 +334,64 @@ CheckLoop:
...
-	poll.Stop()
 }
 
 // pollHeight will send a synchronization request to three random nodes.
-func (e *Engine) pollHeight() error {
-
-	// get the last finalized header
-	final, err := e.state.Final().Head()
+func (e *Engine) pollHeight() {
+	head, err := e.state.Final().Head()
```
smnzhu, line 356: Could we instead use `FinalizedHeaderCache`, like what is done in the common sync engine?

engine/collection/synchronization/engine.go
```diff
@@ -54,310 +67,259 @@ func New(
...
+	)
+
+	return nil
+}
+
+// Ready returns a ready channel that is closed once the engine has fully started.
 func (e *Engine) Ready() <-chan struct{} {
-	e.unit.Launch(e.checkLoop)
-	return e.unit.Ready()
+	e.lm.OnStart(func() {
```
yhassanzadeh13, line 172: Since we are using `unit` it is safer to first invoke `unit.Ready`. This follows the lifecycle assumptions of the `unit`.

yhassanzadeh13 (approved):

jordanschalm:
Cluster syncing is looking good on `localnet` on this PR.

engine/collection/synchronization/engine.go
durkmurder: Not really, collection cluster don't support same notification infrastructure as consensus nodes. Don't think it's worth implementing it only for this.

PR #1260, opened 2021-09-07
Smnzhu second review

smnzhu:


huitseeker:
should be superseded by #1225

PR #1263, opened 2021-09-07
[Networking] Rolls in DNS caching

yhassanzadeh13:
This PR rolls DNS caching for LibP2P into production and makes its TTL configurable through a flag. Note that this is the first version of our DNS caching implementation, which is synchronized. We have an asynchronized implementation coming soon!

### DIFF
--- a/cmd/access/node_builder/staked_access_node_builder.go
+++ b/cmd/access/node_builder/staked_access_node_builder.go
@@ -3,6 +3,7 @@ package node_builder
 import (
 	"context"
 	"fmt"
+	"time"
 
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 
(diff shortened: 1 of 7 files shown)


vishalchangrani:
@yhassanzadeh13 where is the condition to disable the DNS resolver all together if the dns-ttl is 0?

yhassanzadeh13:
> @yhassanzadeh13 where is the condition to disable the DNS resolver all together if the dns-ttl is 0?

@vishalchangrani if TTL is set to zero, the cache is always expired and DNS lookups are made through the network. This DNS resolver is just a wrapper around libp2p one so basically, disabling is not needed since anyway the same functionality is required by libp2p. It is just the matter of whether resolve by cache or query the resolver, which is tweaked by TTL.

vishalchangrani (approved):

Kay-Zee (approved):

Kay-Zee:
bors merge

PR #1266, opened 2021-09-08
[Unstaked Access Node] Bootstrapping the unstaked access node from a root snapshot that does not contain node addresses

vishalchangrani:
Currently, the root protocol state snapshot contain the network addresses of all the staked nodes.
For the unstaked access node, the root protocol state snapshot that is published should not have these node addresses since the unstaked node can be run by anyone and we would like to keep the addresses hidden (as much as possible).
This PR does the following:
1. Introduce a new flag - `SkipRootSnapshotValidation` which when set to true, allows the node to skip certain validations which would otherwise fail for a root snapshot that does not include network addresses.
2. Modifies the integration test to bootstrap the consensus follower from a stripped version of the root snapshot file which does not contain any node addreses.

----------------------

The two validations that fail (when using the stripped version of the root snapshot) are:
1.  https://github.com/onflow/flow-go/blob/master/state/protocol/badger/validity.go#L148-#L150 -`result.ID` includes a hash of service events which include identities.
2. https://github.com/onflow/flow-go/blob/master/state/protocol/badger/validity.go#L35 - Node addresses are duplicated since they are all empty strings when the snapshot is read.

----------------------
Skipped version of the identities looks like this (where ever they occur):

```
    {
      "NodeID": "051985ba92cf8cf0b84837b6b93559f948c7ba93339cc1f89a0da1ba9f159c56",
      "Role": "collection",
      "Stake": 1250000,
      "StakingPubKey": "sBMce7cfvDWkVQ3zTKcXzDGlDf/1gaC7g/dNiVbtJTgiC4PdS5IR2R1mBwZ7OMI5GVy0UxvSTPXMWiX3wXLqoylKkiXmFF6JEi/ry03LeY9koEYurClWCxiTPBsybbZ8",
      "NetworkPubKey": "/P4S9dtEnL95KQ9IYcZvSnJYKHMyyQS/5Wz7+L8cKbf+onMXjnLU2/OxW4dpEgvldpeMlZBi39nsCuadVIQ84w=="
    },
```

Fixes [#5674](https://github.com/dapperlabs/flow-go/issues/5674)


jordanschalm:
The networking addresses are published on-chain and are published in block payloads from time to time. This change does make it ever so slightly less likely that a vegan AN operator would see the networking addresses, but we are 100% relying on security by obscurity here. And the obscurity that remains after this change can be overcome by spending 5 minutes in our public documentation.

vishalchangrani:
> The networking addresses are published on-chain and are published in block payloads from time to time. This change does make it slightly less likely that a vegan AN operator would see the networking addresses by removing one of the ways the addresses could be seen, but we are 100% relying on security by obscurity here. And the obscurity that remains after this change can be overcome by spending 5 minutes in our public documentation site, code, Flow Port...

The eventual goal is to completely remove network addresses from being stored on chain. This change is to make the vegan AN bootup with no knowledge of the other nodes in the network.

jordanschalm:
> The eventual goal is to completely remove network addresses from being stored on chain. This change is to make the vegan AN bootup with no knowledge of the other nodes in the network.

My point is, removing the addresses from this file does little to help toward the goal of keeping the addresses hidden, because the addresses are publicly accessible information. I don't think it's worth trying to half-solve this problem, if the eventual goal (removing the addresses from identity information stored on-chain) isn't met.

integration/testnet/util.go
```diff
@@ -105,3 +106,26 @@ func WriteJSON(path string, data interface{}) error {
...
+
+	fileData, err := ioutil.ReadFile(srcfile)
+	if err != nil {
+		return err
+	}
+	fileString := string(fileData)
+	// look for all "Address": "<node addres>" and remove those elements
+	m1 := regexp.MustCompile(`.*"Address".*:.*".*".*\n`)
+	newString := m1.ReplaceAllString(fileString, "")
+	newFileData := []byte(newString)
```
smnzhu, line 122: I think parsing the JSON into a struct / map, removing the fields, and then rewriting the JSON to the file is cleaner.

Here you're relying on the fact that each JSON key is on it's own line, which is not necessarily true. It is valid to have multiple keys on the same line, and the regex here would fail in that case.

state/protocol/badger/state.go
```diff
@@ -42,7 +42,20 @@ func Bootstrap(
...
+	// optionally skip validations involving flow identity
+	var skipValidations bool
+	switch len(skipSnapshotValidation) {
+	case 0:
+		skipValidations = false
+	case 1:
+		skipValidations = skipSnapshotValidation[0]
+	default:
+		return nil, fmt.Errorf("invalid value for skipSnapshotValidation")
+	}
```
smnzhu, line 57: This looks really hacky. Can we either make this a functional option parameter, or just make it a required parameter?

state/protocol/badger/state.go
```diff
@@ -52,8 +65,10 @@ func Bootstrap(
 	}
 	state := newState(metrics, db, headers, seals, results, blocks, setups, commits, statuses)
 
-	if err := isValidRootSnapshot(root); err != nil {
-		return nil, fmt.Errorf("cannot bootstrap invalid root snapshot: %w", err)
+	if !skipValidations {
+		if err := isValidRootSnapshot(root); err != nil {
+			return nil, fmt.Errorf("cannot bootstrap invalid root snapshot: %w", err)
+		}
```
smnzhu, line 71: Do we really want to skip `isValidRootSnapshot` entirely? Or do we only want to skip certain checks inside that function?

state/protocol/badger/state.go
smnzhu: To use functional option, we would need to define a new struct

```golang
type StateConfig struct {
        skipValidations bool
        ...
}
```

vishalchangrani: agreed..will change this

state/protocol/badger/state.go
vishalchangrani: so for the unstaked AN, the root snapshot is guaranteed to be correct since we are delivering a signed version of the the file and asking the operator to verify the signature. Hence, it should be safe to ignore the validation. I can tweak it to validate checks that can be validated (ones which don't involve the identities).

smnzhu:
> > The eventual goal is to completely remove network addresses from being stored on chain. This change is to make the vegan AN bootup with no knowledge of the other nodes in the network.
> 
> My point is, removing the addresses from this file does little to help toward the goal of keeping the addresses hidden, because the addresses are publicly accessible information. I don't think it's worth trying to half-solve this problem, if the eventual goal (removing the addresses from identity information stored on-chain) isn't met.

@jordanschalm I think this is a step towards the eventual goal, ie it's something that we would need to do eventually anyways. So might as well get it done now.

@vishalchangrani I do think this solution isn't very clean though, it feels like we should remove address validation from the codebase entirely since this will eventually need to be done anyways.

It's actually not clear to me once we have addresses removed from on-chain, how nodes will find out about each other's addresses instead? @huitseeker ?

vishalchangrani:
> > The eventual goal is to completely remove network addresses from being stored on chain. This change is to make the vegan AN bootup with no knowledge of the other nodes in the network.
> 
> My point is, removing the addresses from this file does little to help toward the goal of keeping the addresses hidden, because the addresses are publicly accessible information. I don't think it's worth trying to half-solve this problem, if the eventual goal (removing the addresses from identity information stored on-chain) isn't met.

IMHO, I see this as one of the many things that we need to do to remove the usage of address on chain. Publishing addresses in a file that is publicly accessible and advertised seems like we may be inviting a DOS attack. 
Removal of addresses on chain is a significant lift and don't think it can be by 9/15 :(.

vishalchangrani:
> > > The eventual goal is to completely remove network addresses from being stored on chain. This change is to make the vegan AN bootup with no knowledge of the other nodes in the network.
> > 
> > 
> > My point is, removing the addresses from this file does little to help toward the goal of keeping the addresses hidden, because the addresses are publicly accessible information. I don't think it's worth trying to half-solve this problem, if the eventual goal (removing the addresses from identity information stored on-chain) isn't met.
> 
> @jordanschalm I think this is a step towards the eventual goal, ie it's something that we would need to do eventually anyways. So might as well get it done now.
> 
> @vishalchangrani I do think this solution isn't very clean though, it feels like we should remove address validation from the codebase entirely since this will eventually need to be done anyways.
> 
> It's actually not clear to me once we have addresses removed from on-chain, how nodes will find out about each other's addresses instead? @huitseeker ?

I am not comfortable with removing all address validation so close to the spork date tbh. I will refactor the implementation to be less `hacky`. 
Also, once the addresses are removed entirely, nodes will find each other using trusted DHT servers.

jordanschalm:
> I think this is a step towards the eventual goal, ie it's something that we would need to do eventually anyways. So might as well get it done now.

If this is a step toward the eventual goal, I can get behind it more, but it isn't clear to me that it is. The "eventual goal" isn't designed yet afaik, but to me it very likely would fully remove networking addresses from the protocol state (since the protocol state is public), in which case none of the additions in this PR would be necessary.

smnzhu:
> > I think this is a step towards the eventual goal, ie it's something that we would need to do eventually anyways. So might as well get it done now.
> 
> If this is a step toward the eventual goal, I can get behind it more, but it isn't clear to me that it is. The "eventual goal" isn't designed yet afaik, but to me it very likely would fully remove networking addresses from the protocol state (since the protocol state is public), in which case none of the additions in this PR would be necessary.

The eventual design is to have address information stored in a DHT, which can only be accessed by staked nodes.

@vishalchangrani what exactly would be the risks associated with removing all address validation? I think @jordanschalm is right that basically all of the code added in this PR is throwaway code. Imo we should just remove all address validation code because then the changes would actually be useful for the future

huitseeker (commented):
Concept ACK.

@jordanschalm I think the overall goal is to make address dissemination minimal. It touches on security goals that I've addressed elsewhere, but I would also note that it's also valuable to not put this on chain from a pure performance standpoint: the shelf life of networking information is limited, in a way that the one of e.g. cryptographic keys (useful for forensics) isn't.
The main comment I have is the one Simon made on filtering the JSON post deserialization rather than line by line.

integration/testnet/util.go
```diff
@@ -105,3 +108,50 @@ func WriteJSON(path string, data interface{}) error {
...
+	err = json.Unmarshal(data, &rootSnapshot)
+	if err != nil {
+		return err
+	}
+
+	removeAddress := func(ids flow.IdentityList) {
+
+		for _, identity := range ids {
+			identity.Address = ""
+			//TODO: following doesn't work and Address is still retained
```
vishalchangrani, line 131: Tried a few different things 👇 to influence the `json.Marshal` call to skip fields dynamically but none of them have  worked so far.
1.  This 3rd party library to manager tags - https://github.com/sevlyar/retag
2. Couple of other suggestions from here - https://stackoverflow.com/questions/17306358/removing-fields-from-struct-or-hiding-them-in-json-response.

integration/testnet/util.go
```diff
@@ -105,3 +108,50 @@ func WriteJSON(path string, data interface{}) error {
 	err = ioutil.WriteFile(path, marshaled, 0644)
 	return err
 }
+
+// rootProtocolJsonWithoutAddresses strips out all node addresses from the root protocol json file specified as srcFile
+// and creates the dstFile with the modified contents
+func rootProtocolJsonWithoutAddresses(srcfile string, dstFile string) error {
```
jordanschalm, line 114: I think the part of this function which modifies the `Snapshot` should be exported, and not in a testing-specific package since I would imagine we will use this in the tool used to distribute the stripped snapshot file. 
Eg:
```go
// read root snapshot
strippedRootSnapshot := protocol.StrippedSnapshot(rootSnapshot) // the part we can re-use
WriteJSON(strippedRootSnapshot)
```

state/protocol/badger/state.go
```diff
@@ -42,17 +43,19 @@ func Bootstrap(
 	commits storage.EpochCommits,
 	statuses storage.EpochStatuses,
 	root protocol.Snapshot,
+	skipNwAddressBasedValidations bool,
```
jordanschalm, line 46: A few thoughts:

1. We are using this config flag during the bootstrap, then also storing it in the `State` struct. As far as I can see that stored value is never used (eg [in `mutator.go`](https://github.com/onflow/flow-go/pull/1266/files#diff-c79d3fc5d32f153917b7cd923ab4ba64ef22f12268ddd13f353a06391f68a96dR817) we pass in `false` rather than use the stored flag. I would suggest we stop storing this value in the struct if we aren't using it, and only modify the validation logic at startup, when we are validating the stripped snapshot.

2. I'd prefer we avoid the pattern of adding configs as required parameters, especially for things like this where the config represents an uncommon edge case. It makes the common case of calling the function more concise and makes modifications to the common case more explicit.

```go
type BootstrapConfig struct {
  SkipNetworkAddressValidation bool
}

func WithSkipNetworkAddressValidation(conf *BootstrapConfig) {
  conf.SkipNetworkAddressValidation = true
}

// at the majority of callsites - no change
Bootstrap(...)
// where needed
Bootstrap(..., WithSkipNetworkAddressValidation) 
```

3. Rather than passing the config all the way down to the validation function (eg. `isValidEpochSetup`), we could make the validation functions return descriptive sentinel errors for the validation cases we want to treat differently, then handle those sentinels based on the config, in the same layer the config has already been defined (`Bootstrap`).

```go
err := isValidEpochSetup(...)
if errors.Is(err, ErrDuplicateNetworkingAddress) && !SkipNetworkAddressValidation {
  return err
}  else if err != nil {
  return err
}
```

integration/testnet/util.go
huitseeker: Here is one idea that I think could help:
https://play.golang.org/p/5gjkRpGngkK

huitseeker (commented):
I think you're not allowed to go mutate the tags of an existing type.
But I added an inline suggestion that might help. The gist of it, if we can't afford to play with a new struct, is the following, I think:

```
value := reflect.ValueOf(*identity)
    t := value.Type()
    sf := make([]reflect.StructField, 0)
    for i := 0; i < t.NumField(); i++ {
        sf = append(sf, t.Field(i))
        if t.Field(i).Name == "Address" {
            sf[i].Tag = `json:"-"`
        }
    }
    newType := reflect.StructOf(sf)
    newValue := value.Convert(newType)
 ```

File: integration/testnet/util.go (hunk 4 above)
Comment (jordanschalm) [reply integration/testnet/util.go:131]: You could also add some logic to the existing `MarshalJSON` [method](https://github.com/onflow/flow-go/blob/1bb8a60dd79de3fa0e7ac327b466f12f3c1430fd/model/flow/identity.go#L115-L125) to skip the `Address` field if it is empty:
```go
if id.Address == "" {
  return json.Marshal(identityWithoutAddress) // will need a new type here, like in Francois' example
} else {
  return json.Marshal(identity) // do the same thing it's currently doing
}
```

state/protocol/badger/validity.go
```diff
@@ -27,14 +27,16 @@ func isValidEpochSetup(setup *flow.EpochSetup) error {
...
-			return fmt.Errorf("duplicate node address (%x)", participant.Address)
+	if !skipNetworkAddressValidations {
+		// there should be no duplicate node addresses
+		addrLookup := make(map[string]struct{})
+		for _, participant := range setup.Participants {
+			_, ok := addrLookup[participant.Address]
+			if ok {
+				return fmt.Errorf("duplicate node address (%x)", participant.Address)
+			}
+			addrLookup[participant.Address] = struct{}{}
```
smnzhu, line 38: @vishalchangrani still not sure why we can't just remove address validation entirely? Wouldn't we just need to remove this piece of code?

vishalchangrani: this is currently used to safeguard against duplicate addresses in the root snapshot file. I wanted to keep those existing validations untouched.

integration/testnet/util.go
vishalchangrani: Added a new function to `protocol/inmem/snapshot.go`

state/protocol/badger/state.go
```diff
@@ -31,6 +31,28 @@ type State struct {
 	}
 }
 
+type BootstrapConfig struct {
```
vishalchangrani, line 34: introduced this Config struct to ignore certain validation error

state/protocol/badger/state.go
```diff
@@ -31,6 +31,28 @@ type State struct {
...
+
+type BootstrapConfigOptions func(conf *BootstrapConfig)
+
+func SkipNetworkAddressValidation(conf *BootstrapConfig) {
+	conf.SkipNetworkAddressValidation = true
+}
+
+// NetworkAddressError are validation errors that are expected to occur during the bootstrap of an unstaked access node
+// They occur because the node identity addresses in the root snapshot file used to bootstrap the node are absent
+var NetworkAddressError = errors.New("network address error")
```
vishalchangrani, line 54: the sentinel error

state/protocol/badger/validity.go
```diff
@@ -145,10 +148,6 @@ func isValidRootSnapshot(snap protocol.Snapshot) error {
 		return fmt.Errorf("root block seal for wrong block (%x != %x)", seal.BlockID, tail.ID())
 	}
 
-	if seal.ResultID != result.ID() {
```
vishalchangrani, line 148: moved this after the identities validation, so that identities can be checked to determine whether to throw the NetworkAddressError or just a regular error

state/protocol/badger/validity.go
```diff
@@ -10,6 +10,10 @@ import (
 )
 
 func isValidEpochSetup(setup *flow.EpochSetup) error {
+	return verifyEpochSetup(setup, true)
+}
+
+func verifyEpochSetup(setup *flow.EpochSetup, verifyNetworkAddress bool) error {
```
vishalchangrani, line 16: `verifyEpochSetup` is only called during bootstrap

model/flow/identity.go
```diff
@@ -113,11 +123,20 @@ func encodableFromIdentity(iy Identity) (encodableIdentity, error) {
...
 func (iy Identity) MarshalJSON() ([]byte, error) {
+	var identity interface{}
 	encodable, err := encodableFromIdentity(iy)
 	if err != nil {
 		return nil, fmt.Errorf("could not convert identity to encodable: %w", err)
 	}
-	data, err := json.Marshal(encodable)
+
+	// if the address is empty, suppress the Address field in the output json
+	if encodable.Address == "" {
```
huitseeker, line 133: So, at the moment we need this `MarshalJSON` and `StrippedInmemSnapshot` to each work in sync with each other.

- are there known cases where we want to preserve the `Address` field?
- if not, couldn't we simply convert all `identity` fields here to remove the Address in one step?

huitseeker (commented):
Should we try to create an identity, use the newly introduced functions to marshal it to JSON, unmarshal the output, and check the Address is gone?

I'm interested in particular in an identity that's nested in some collection, to unit-test something a bit closer to the Epoch definition.

model/flow/identity.go
vishalchangrani: You mean valid cases where the `address` is an empty string but we still want to retain it? I don't think so but will have to check.
In the regular case, the address will be non empty and we do want to retain it.

huitseeker: > In the regular case, the address will be non empty and we do want to retain it.

Yep, of course.

What I mean is this: we have two data flows, branching on whether you're producing a file for an unstaked node. One is marshalling as usual, another one is using two sequential components to suppress Addresses.

THose sequential components are:
- replacing Addresses by the empty string,
- detecting the empty string addresses, and for those identities, converting them to a struct which Address is skipped on marshalling.

I'm saying we can:
- branch on whether you're producing a file for an unstaked node,
- in the case we are, convert every identity to a struct which Address field is omitted.

huitseeker (approved):
Yay, tyvm for the test!

model/flow/identity.go
vishalchangrani: The identity structs are used within the different structs that make up the snapshot. e.g. Epoch, Execution Result -> Service Event. Hence, defining a new Identity struct  (e.g. `StealthIdentity`) would also require re-defining these enclosing structs to their new `Stealthxxx` version.

huitseeker: Yeah, which is why I linked to a way of doing this dynamically here: https://github.com/onflow/flow-go/pull/1266#pullrequestreview-751736361

This redefines the fields to strip the Address w/o defining a new struct.

vishalchangrani: Oh, I did try that earlier. There are a fews things I ran into -
1. How convert from the new struct type back to an identity type -
This doesn't work :(  - `newIdentity := reflect.ValueOf(newValue).Interface().(flow.Identity)`
2. Tags may still be lost even after the conversion - https://stackoverflow.com/a/32189702
3. The flow.Identity is converted to yet another type - Encodable Identity [when doing the marshalling](https://github.com/onflow/flow-go/blob/master/model/flow/identity.go#L95-#L116). Hence, these tags would have to be copied over to that type.

don't mind investigating this further but I feel I may have hit a wall

jordanschalm (approved):
Thanks the updates 🎸

smnzhu (approved):

vishalchangrani:
bors merge

PR #1268, opened 2021-09-08
Khalil/async message handling

kc1116:
Testing with benchnet1 with a DKG committee size the same as mainnet (150 consensus nodes) we found that during the initial DKG phase because some functions that were broadcasting , receiving or forwarding DKG messages due so in a synchronous manner. As a result with 150 nodes each sending a message to every node the message queue gets backed up and consensus fails because the DKG fails . This PR changes these functions to perform those blocking actions asynchronously.


tarakby (approved):
👌

engine/consensus/dkg/messaging_engine.go
```diff
@@ -104,40 +104,54 @@ func (e *MessagingEngine) Process(_ network.Channel, originID flow.Identifier, e
...
 			if err != nil {
-				e.log.Error().Err(err).Msg("error sending dkg message")
+				e.log.Error().Err(err).Msg("error sending dkg message retrying")
 			}
-		case <-e.unit.Quit():
-			return
+			return retry.RetryableError(err)
+		})
+		if err != nil {
+			e.log.Error().Err(err).Msg("error sending dkg message")
```
AlexHentschel, line 152: * On a first glance, this mode of error treatment seems to be conflicting with our policy 
   > We would like to avoid:
   > • swallowing errors, or "just" logging them

   (for details see Francois' great notion doc [Error handling in Flow](https://www.notion.so/dapperlabs/Error-Handling-in-Flow-a7692d60681c4063b48851c0dbc6ed10)).

* On a second glance, I think your error handling here is fine for the following reasons:
   * all we do here is handing a message to the networking layer, which we expect to error from time to time (unfortunately without specific sentinels) 
   * given that the networking layer is the _only_ source of errors, it is safe to just "swallow all errors"
 * But keep in mind that you need to untangle the code here, and inspect the detailed error propagation of `retry.Do` to arrive at this conclusion. Lastly, `retry.Do` could theoretically propagate an error from the context:
   ```golang
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(next):
			continue
		}
   ```

So in conclusion, there are multiple implicit assumptions here that all need to be satisfied in order for the code to comply with our "do not swallow errors" policy. 


#### suggestion: 
Preferably, such assumptions should be reflected by the code. Because 6 months down the line, some engineer (maybe it is you) might need to go through all of this thought process again to safely update the code. Some goDoc is still better than nothing.

module/dkg/broker.go
```diff
@@ -116,18 +120,22 @@ func (b *Broker) Broadcast(data []byte) {
...
+			err := b.dkgContractClient.Broadcast(bcastMsg)
+			if err != nil {
+				b.log.Error().Err(err).Msg("error broadcasting DKG result, retrying")
+			}
+			return retry.RetryableError(err)
+		})
+
 		if err != nil {
-			b.log.Error().Err(err).Msg("error broadcasting DKG result, retrying")
+			b.log.Error().Err(err).Msg("failed to broadcast message")
```
AlexHentschel, line 133: same here: please at least include some goDoc why this is not a violation of our "do _not_ "just" log errors and continue" policy.

AlexHentschel (commented):
While reviewing the PR, I found a couple minor details that I feel need cleaning up. While they were not introduced in this PR, I think it is well within the scope of this PR to fix them:

### Please respect `Engine`'s API convention:
https://github.com/onflow/flow-go/blob/f1bda305666b71bc6d76de92dc81256f0e611b7e/network/engine.go#L14-L15

* An engine generally processes messages that can come from other engines within the same node. It is _not_ correct, to assume that a `network.Channel` can be associated with each message. However, the `MessagingEngine` specifically codifies this incorrect association:  https://github.com/onflow/flow-go/blob/f1bda305666b71bc6d76de92dc81256f0e611b7e/engine/consensus/dkg/messaging_engine.go#L92-L102
  * If the `MessagingEngine` receives _locally_ a message from another engine, it pretends that the message came via the network from a DKG participant: https://github.com/onflow/flow-go/blob/f1bda305666b71bc6d76de92dc81256f0e611b7e/engine/consensus/dkg/messaging_engine.go#L94
  * The value for `network.Channel` is subsequently ignored by `Submit`, so this does not create a bug. Nevertheless, the implementation conflicts with the `Engine` API convention

#### Suggestion:
* for methods `SubmitLocal` and `ProcessLocal`, I would directly call into `MessagingEngine.process`

### Please do not just log errors and continue
https://github.com/onflow/flow-go/blob/f1bda305666b71bc6d76de92dc81256f0e611b7e/engine/consensus/dkg/messaging_engine.go#L85-L88

This code is so high-level, that I would consider it a violation of our "do _not_ just log errors and continue' policy. Ideally, `process` would specify sentinel errors (like `InvalidMessageType`, can clearly document them in goDoc. For messages that come from the network (methods `Submit` and `Process`), we expect invalid message types. But if I get an invalid message from a _trusted_ local engine (methods `SubmitLocal` and `ProcessLocal`), something is _badly_ wrong (critical failure).

module/dkg/broker.go
```diff
@@ -116,18 +120,22 @@ func (b *Broker) Broadcast(data []byte) {
 	}
 	maxedExpRetry := retry.WithMaxRetries(retryMax, expRetry)
 
-	err = retry.Do(context.Background(), maxedExpRetry, func(ctx context.Context) error {
-		err := b.dkgContractClient.Broadcast(bcastMsg)
+	f := func() {
+		err = retry.Do(context.Background(), maxedExpRetry, func(ctx context.Context) error {
+			err := b.dkgContractClient.Broadcast(bcastMsg)
+			if err != nil {
+				b.log.Error().Err(err).Msg("error broadcasting DKG result, retrying")
```
zhangchiqing, line 127:
```suggestion
				b.log.Warn().Err(err).Msg("error broadcasting DKG result, retrying")
```

engine/consensus/dkg/messaging_engine.go
```diff
@@ -104,40 +104,54 @@ func (e *MessagingEngine) Process(_ network.Channel, originID flow.Identifier, e
...
+		if err != nil {
+			e.log.Fatal().Err(err).Msg("failed to create retry mechanism")
+		}
+
+		maxedExpRetry := retry.WithMaxRetries(retryMax, expRetry)
+		err = retry.Do(e.unit.Ctx(), maxedExpRetry, func(ctx context.Context) error {
+			err := e.conduit.Unicast(&message.DKGMessage, message.DestID)
 			if err != nil {
-				e.log.Error().Err(err).Msg("error sending dkg message")
+				e.log.Error().Err(err).Msg("error sending dkg message retrying")
```
zhangchiqing, line 147:
```suggestion
				e.log.Warn().Err(err).Msg("error sending dkg message retrying")
```
because we will retry

engine/consensus/dkg/messaging_engine.go
```diff
@@ -104,40 +104,54 @@ func (e *MessagingEngine) Process(_ network.Channel, originID flow.Identifier, e
 func (e *MessagingEngine) process(originID flow.Identifier, event interface{}) error {
 	switch v := event.(type) {
 	case *msg.DKGMessage:
+		e.forwardInboundMessageAsync(originID, v)
```
zhangchiqing, line 107:
```suggestion
		// messages are forwarded async rather than sync, because otherwise the message queue
		// might get full when it's slow to process DKG messages synchronously and impact 
		// block rate.
		e.forwardInboundMessageAsync(originID, v)
```

zhangchiqing (approved):

engine/consensus/dkg/messaging_engine.go
tarakby: As I'm looking at the logs now, it would helpful to add the retry number here. This seems to be happening often and I would like to see what node has reached the maximum retry number.

module/dkg/broker.go
tarakby: I think it makes sense to leave this as an error since failing to broadcast has important consequences on the node.

tarakby: On a separate note, the error message is not correct: DKG result is one of the message types that can be broadcast, but the error here is about any message type. 
```suggestion
				b.log.Error().Err(err).Msg("error broadcasting, retrying")
```

engine/consensus/dkg/messaging_engine.go
tarakby: It would be great to mirror the same also on the broadcast retries.

engine/consensus/dkg/messaging_engine.go
```diff
@@ -76,7 +76,12 @@ func (e *MessagingEngine) Done() <-chan struct{} {
 
 // SubmitLocal implements the network Engine interface
 func (e *MessagingEngine) SubmitLocal(event interface{}) {
-	e.Submit(engine.DKGCommittee, e.me.NodeID(), event)
+	e.unit.Launch(func() {
+		err := e.Process(engine.DKGCommittee, e.me.NodeID(), event)
```
jordanschalm, line 80:
```suggestion
		err := e.ProcessLocal(engine.DKGCommittee, e.me.NodeID(), event)
```

engine/consensus/dkg/messaging_engine.go
```diff
@@ -104,40 +111,57 @@ func (e *MessagingEngine) Process(_ network.Channel, originID flow.Identifier, e
 func (e *MessagingEngine) process(originID flow.Identifier, event interface{}) error {
 	switch v := event.(type) {
 	case *msg.DKGMessage:
+		// messages are forwarded async rather than sync, because otherwise the message queue
+		// might get full when it's slow to process DKG messages synchronously and impact
+		// block rate.
+		e.forwardInboundMessageAsync(originID, v)
+		return nil
+	default:
+		return fmt.Errorf("invalid event type (%T)", event)
```
AlexHentschel, line 120: this should be a sentinel error, like `engine.InvalidInputError`, so we can treat it accordingly: 
* for external inputs, this is expected during normal operations 
* for internal inputs, this would be fatal 

```suggestion
		return engine.NewInvalidInputErrorf("expecting input with type msg.DKGMessage, but got %T", event)
```

engine/consensus/dkg/messaging_engine.go
```diff
@@ -91,7 +96,9 @@ func (e *MessagingEngine) Submit(_ network.Channel, originID flow.Identifier, ev
 
 // ProcessLocal implements the network Engine interface
 func (e *MessagingEngine) ProcessLocal(event interface{}) error {
-	return e.Process(engine.DKGCommittee, e.me.NodeID(), event)
+	return e.unit.Do(func() error {
+		return e.process(e.me.NodeID(), event)
```
jordanschalm, line 100:
```suggestion
		err := e.process(e.me.NodeID(), event)
		if errors.Is(err, engine.IncompatibleInputTypeError) {
			// critical error - log fatal
		}
```
handle invalid input types from internal components with a crash

engine/consensus/dkg/messaging_engine.go
jordanschalm, line 120:
```suggestion
		return fmt.Errorf("invalid event type (%T): %w", event, engine.IncompatibleInputTypeError)
```
return this error as a sentinel, which we can handle depending on the source of the input.

engine/consensus/dkg/messaging_engine.go
```diff
@@ -104,40 +111,57 @@ func (e *MessagingEngine) Process(_ network.Channel, originID flow.Identifier, e
...
-			}
+			e.forwardOutboundMessageAsync(msg)
+		case <-e.unit.Quit():
+			return
+		}
+	}
+}
+
+func (e *MessagingEngine) forwardOutboundMessageAsync(message msg.PrivDKGMessageOut) {
+	f := func() {
```
jordanschalm, line 147:
```suggestion
	sendMessage := func() {
```
Could use a more descriptive name, or simply pass the function anonymously into `unit.Launch` - eg `unit.Launch(func() { ... })`

engine/consensus/dkg/messaging_engine.go
```diff
@@ -76,7 +76,12 @@ func (e *MessagingEngine) Done() <-chan struct{} {
 
 // SubmitLocal implements the network Engine interface
 func (e *MessagingEngine) SubmitLocal(event interface{}) {
-	e.Submit(engine.DKGCommittee, e.me.NodeID(), event)
+	e.unit.Launch(func() {
+		err := e.Process(engine.DKGCommittee, e.me.NodeID(), event)
+		if err != nil {
+			engine.LogError(e.log, err)
```
AlexHentschel, line 82: three requests for revisions:
* Here, you are calling `unit.Launch` which subsequently calls `e.Process` which then calls `unit.Do`. I find this is an unnecessary duplication of `unit` calls. 
* If you look at the implementation of `Process`, it is straight forward to convince yourself that the only error it can throw is an invalid input https://github.com/onflow/flow-go/blob/1f463cc9c73092c52622a1f7447b721a4563bd60/engine/consensus/dkg/messaging_engine.go#L120 If a local, _trusted_ component feeds an event of invalid type into this engine, that is certainly a fatal implementation bug. 
* formally, `engine.DKGCommittee` is an identifier for a _networking_ channel. `SubmitLocal` receives events from an _internal engine_, which is _not_ a networking channel. 


the following would address all three concerns:
```suggestion
		err := e.process(originID, event)
		if err != nil {
			e.log.Fatal().Err(err).Msg("unexpected internal error processing event")
```

engine/consensus/dkg/messaging_engine.go
jordanschalm: @kc1116 Could you add some documentation here explaining why swallowing the error is OK in this case, like alex suggested - what conditions can result in an error return here and why they're expected (also can note that failure to send an individual DKG message doesn't necessarily result in local or global DKG failure)

module/dkg/broker.go
```diff
@@ -116,18 +120,22 @@ func (b *Broker) Broadcast(data []byte) {
 	}
 	maxedExpRetry := retry.WithMaxRetries(retryMax, expRetry)
 
-	err = retry.Do(context.Background(), maxedExpRetry, func(ctx context.Context) error {
-		err := b.dkgContractClient.Broadcast(bcastMsg)
+	f := func() {
```
jordanschalm, line 123: Same comment here, I'd suggest just defining the function directly in the call to `unit.Launch`

module/dkg/broker.go
```diff
@@ -116,18 +119,22 @@ func (b *Broker) Broadcast(data []byte) {
...
 		if err != nil {
-			b.log.Error().Err(err).Msg("error broadcasting DKG result, retrying")
+			b.log.Fatal().Err(err).Msg("failed to broadcast message")
 		}
-		return retry.RetryableError(err)
-	})
-
-	if err != nil {
-		b.log.Fatal().Err(err).Msg("failed to broadcast message")
+		b.broadcasts++
```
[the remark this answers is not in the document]
jordanschalm: I think this still needs to be addressed @kc1116 ☝️

AlexHentschel (commented):
Please see the comments below for change requests. 

In addition, please revise 
* https://github.com/onflow/flow-go/blob/1f463cc9c73092c52622a1f7447b721a4563bd60/engine/consensus/dkg/messaging_engine.go#L87-L95
* to 
   ```golang
    func (e *MessagingEngine) Submit(_ network.Channel, originID flow.Identifier, event interface{}) {
    	e.unit.Launch(func() {
    		err := e.process(originID, event)
    		if err != nil {
    			if engine.IsInvalidInputError(err) {
    				e.log.Error().Err(err).Str("origin", originID.String()).Msg("dropping invalid input message")
    			}
    			e.log.Fatal().Err(err).Msg("unexpected internal error")
    		}
    	})
    }
   ```
Reasoning: 
* If you look at the implementation of `Process`, it is straight forward to convince yourself that the only error it can throw is an invalid input error (based on my [suggestion below](https://github.com/onflow/flow-go/pull/1268#discussion_r707646971)).  If we receive any other error, the code has been updated in an inconsistent way.

module/dkg/broker_test.go
```diff
@@ -259,6 +259,7 @@ func TestBroadcastMessage(t *testing.T) {
 	sender.dkgContractClient = contractClient
 
 	sender.Broadcast(msgb)
+	<-sender.unit.Done()
```
jordanschalm, line 262:
```suggestion
	unittest.AssertClosesBefore(sender.unit.Done(), time.Second)
```
We have a utility to wrap waiting for a channel to exit, to prevent a bug blocking the tests and making such bugs more obvious.

engine/consensus/dkg/messaging_engine.go
```diff
@@ -77,27 +77,32 @@ func (e *MessagingEngine) Done() <-chan struct{} {
 // SubmitLocal implements the network Engine interface
 func (e *MessagingEngine) SubmitLocal(event interface{}) {
 	e.unit.Launch(func() {
-		err := e.Process(engine.DKGCommittee, e.me.NodeID(), event)
-		if err != nil {
-			engine.LogError(e.log, err)
+		err := e.process(e.me.NodeID(), event)
```
jordanschalm, line 80: Sorry, my last suggestion did not fully cover these cases. 

We are getting an error from processing either a local or remote message. In either case, we want to crash in the case of an unexpected error, and handle expected errors.

For remote messages, the only expected error is `IsInvalidInputError`. We are handling it, but we should also crash for any other, unexpected errors (line 90-92).
Eg add:
```
else if err != nil {
  // crash
}
```

For local messages, there is no expected error. Rather than crashing only in the case of one explicitly disallowed error type, we can crash for any error (lines 81-83)
Eg change to:
```
if err != nil {
  // crash
```

module/dkg/broker.go
```diff
@@ -120,22 +120,29 @@ func (b *Broker) Broadcast(data []byte) {
...
 		if err != nil {
 			b.log.Error().Err(err).Msg("failed to broadcast message")
 		}
-		b.broadcasts++
-	}
 
-	b.unit.Launch(f)
+		b.unit.Lock()
+		b.broadcasts++
+		b.unit.Unlock()
```
jordanschalm, line 144: checking the broadcast count and incrementing it should happen in the same lock context, and before sending the broadcast message. Otherwise we have a race condition where two threads check the count, see 0, _don't_ log a warning, then both broadcast and increment the counter. We have broadcast twice but won't know about it.

jordanschalm: @kc1116 We still have this problem with the most recent change. 2 (or more) threads can:
* acquire the lock
* read `b.broadcasts`
* release the lock
before either thread completes the broadcast and increments `b.broadcasts` (because the read and the write are happening after 2 separate instances of acquiring the lock -- they are not occurring in the same lock context).

Instead we should:
```go
// acquire the lock
// read the broadcasts value and log accordingly
// increment the broadcasts value
// release the lock
// perform the actual broadcast
```

kc1116: Okay is it correct to increment broadcast even if the broadcasting fails ?

jordanschalm: > is it correct to increment broadcast even if the broadcasting fails ?

It is. We're counting the number of times the underlying DKG _requested_ a broadcast so we can detect an unhappy path.  I can see why there'd be confusion about that though -- do you mind adding a comment to that effect when we increment `broadcasts`?

jordanschalm (approved):

jordanschalm (approved):
🎸

kc1116:
bors merge

PR #1273, opened 2021-09-10
clamping sync requests

smnzhu:
closes #https://github.com/dapperlabs/flow-go/issues/5836

### DIFF
--- a/engine/common/synchronization/engine_test.go
+++ b/engine/common/synchronization/engine_test.go
@@ -336,7 +336,7 @@ func (ss *SyncSuite) TestOnRangeRequest() {
 		},
 	)
 	err = ss.e.requestHandler.onRangeRequest(originID, req)
-	require.NoError(ss.T(), err, "valid range request should pass")
+	require.NoError(ss.T(), err, "valid range request exceeding max size should still pass")
 }
 
 func (ss *SyncSuite) TestOnBatchRequest() {
@@ -394,7 +394,7 @@ func (ss *SyncSuite) TestOnBatchRequest() {
 		},
 	)
 	err = ss.e.requestHandler.onBatchRequest(originID, req)
-	require.NoError(ss.T(), err, "should pass request with valid blocks")
+	require.NoError(ss.T(), err, "valid batch request exceeding max size should still pass")
 }
 
 func (ss *SyncSuite) TestOnBlockResponse() {
--- a/engine/common/synchronization/engine_test.go
+++ b/engine/common/synchronization/engine_test.go
@@ -322,6 +322,21 @@ func (ss *SyncSuite) TestOnRangeRequest() {
 	)
 	err = ss.e.requestHandler.onRangeRequest(originID, req)
 	require.NoError(ss.T(), err, "valid range request should pass")
+
+	// a request for a range larger than MaxSize should be clamped
+	req.FromHeight = ref - uint64(synccore.DefaultConfig().MaxSize) - 1
+	req.ToHeight = ref
+	ss.con.On("Unicast", mock.Anything, mock.Anything).Return(nil).Once().Run(
+		func(args mock.Arguments) {
+			res := args.Get(0).(*messages.BlockResponse)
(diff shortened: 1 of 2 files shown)


engine/common/synchronization/request_handler.go
```diff
@@ -221,6 +222,11 @@ func (r *RequestHandlerEngine) onRangeRequest(originID flow.Identifier, req *mes
 		return nil
 	}
 
+	maxHeight := req.FromHeight + uint64(r.core.(*synchronization.Core).Config.MaxSize)
+	if maxHeight < req.FromHeight {
+		req.FromHeight = maxHeight
+	}
```
huitseeker, line 228:
```suggestion
	if maxHeight < req.ToHeight {
		req.ToHeight = maxHeight
	}
```

engine/common/synchronization/request_handler.go
```diff
@@ -266,7 +272,10 @@ func (r *RequestHandlerEngine) onBatchRequest(originID flow.Identifier, req *mes
 
 	// deduplicate the block IDs in the batch request
 	blockIDs := make(map[flow.Identifier]struct{})
-	for _, blockID := range req.BlockIDs {
+	for i, blockID := range req.BlockIDs {
+		if i == int(r.core.(*synchronization.Core).Config.MaxSize) {
+			break
+		}
```
huitseeker, line 278: Is a client susceptible to sending a request for a small amount of BlockIDs that is so full of duplicates that it appears to be `> r.core.(*synchronization.Core).Config.MaxSize`?

Could we instead check `len(blockIDs)` as an additional condition to entering the loop?

huitseeker (commented):
Thanks a bunch for getting to this fast!

I would love a code comment on each of the two checks making it explicit we enforce *existing*  client-side limits, and that this therefore shouldn't have an effect on a well-behaving client. 

Without that, another later reader, may not find it obvious those lines are meant to check for mischief.

huitseeker:
One more thing:
`TestOnRangeRequest` and `TestOnBatchRequest` look ripe for a test exercising the over-the-limit requests.

smnzhu:
Addressed comments

engine/common/synchronization/engine_test.go
```diff
@@ -322,6 +322,21 @@ func (ss *SyncSuite) TestOnRangeRequest() {
...
+		func(args mock.Arguments) {
+			res := args.Get(0).(*messages.BlockResponse)
+			assert.Len(ss.T(), res.Blocks, int(synccore.DefaultConfig().MaxSize))
+			assert.Equal(ss.T(), req.Nonce, res.Nonce, "response should contain request nonce")
+			recipientID := args.Get(1).(flow.Identifier)
+			assert.Equal(ss.T(), originID, recipientID, "should send response to original requester")
+		},
+	)
+	err = ss.e.requestHandler.onRangeRequest(originID, req)
+	require.NoError(ss.T(), err, "valid range request should pass")
```
huitseeker, line 339: Nit: do we want to call this "valid"?

engine/common/synchronization/engine_test.go
```diff
@@ -358,9 +373,28 @@ func (ss *SyncSuite) TestOnBatchRequest() {
...
+		func(args mock.Arguments) {
+			res := args.Get(0).(*messages.BlockResponse)
+			assert.Len(ss.T(), res.Blocks, int(synccore.DefaultConfig().MaxSize))
+			assert.Equal(ss.T(), req.Nonce, res.Nonce, "response should contain request nonce")
+			recipientID := args.Get(1).(flow.Identifier)
+			assert.Equal(ss.T(), originID, recipientID, "response should be send to original requester")
+		},
+	)
+	err = ss.e.requestHandler.onBatchRequest(originID, req)
+	require.NoError(ss.T(), err, "should pass request with valid blocks")
```
huitseeker, line 397: Nit: do we want to call this "valid"?

huitseeker (approved):
Thanks a bunch!

vishalchangrani (approved):
lgtm

smnzhu:
bors merge

PR #1275, opened 2021-09-10
Define error aware interface

smnzhu:
Defines a new Error Aware interface which an owner of a component can use to be notified when a fatal error occurs in the component. The owner can then decide what to do which this error (restart the component, log the error, propagate the error up to its own parent component, etc).

Eventually, every component could implement this interface. After a component is started, the owner of the component would also starts up a thread that reads from the error channel. When an error occurs, that thread can parse the error and decide how to respond to it.

makes progress on https://github.com/dapperlabs/flow-go/issues/5829

### TODO
- [x] Update the code so that irrecoverable errors encountered *during* startup and shutdown are handled properly. Currently an irrecoverable error thrown during startup will cause `start()` inside `RunComponent` to block forever, and error thrown during shutdown will be ignored.
- [x] Handle race condition where multiple irrecoverables are thrown at the same time from a component
- [x] Write a test case where error is thrown during startup (while waiting for `Ready`)
- [x] Write a test case where error is thrown during shutdown


huitseeker (commented):
It's a stab in the right direction, because it constrains the implementer to have a place where to send irrecoverable errors (the error channel).

Here's a list of things that seem important to me:
- make it clear this is for irrecoverable errors from the PoV of the ErrorAware module, i.e. that this isn't meant to encompass all error management therein,
- force the implementer to have a way to _react_ to those irrecoverable errors, i.e. require a `func OnError(err error)`,
- and then possibly link the two with default processing, perhaps by doing the following (?) :
    - having an ErrorBase that listens to the error channel and passes the errors read there to `OnError` 
    - and then having implementing components embed the `ErrorBase`

smnzhu:
> It's a stab in the right direction, because it constrains the implementer to have a place where to send irrecoverable errors (the error channel).
> 
> Here's a list of things that seem important to me:
> 
> * make it clear this is for irrecoverable errors from the PoV of the ErrorAware module, i.e. that this isn't meant to encompass all error management therein,
> * force the implementer to have a way to _react_ to those irrecoverable errors, i.e. require a `func OnError(err error)`,
> * and then possibly link the two with default processing, perhaps by doing the following (?) :
>   
>   * having an ErrorBase that listens to the error channel and passes the errors read there to `OnError`
>   * and then having implementing components embed the `ErrorBase`

Actually, the way I think about this is slightly different.

The `Errors()` function of a component allows the *owner* (aka parent) of the component to handle fatal errors in the component. I don't think the component itself should handle its own fatal errors. In other words, the implementer of the component shouldn't have to react to it's own irrecoverable errors, because said errors are by definition irrecoverable from its own point of view. Only the parent of the component has enough context to handle these errors.

so the way this works is, once the owner of component A starts is, it then spawns a new routine which polls from the error channel of component A. If it receives an error, then it can shutdown component A, or restart it, or propagate the error on to it's own parent, etc.

gomisha:
I like the idea of having the caller / parent (not the component itself) handle the child component's error. We could have a mechanism to keep a running log of the errors that could be further analyzed for things like flakiness (is the same component throwing an error at frequent intervals?)

How would we test this? i.e. under what conditions would components throw an error?

What would be the expected error handling behavior and can we standardize this? (i.e. restart component, log the error somewhere permanent for later analysis, notify someone)

huitseeker:
@smnzhu I think we're on the same page, I wasn't clear about two interfaces: 
- one in the component, which is the error sink,
- one outside of the component, which has to react to errors, and to which my comment applied.

smnzhu:
> How would we test this? i.e. under what conditions would components throw an error?
> 
> What would be the expected error handling behavior and can we standardize this? (i.e. restart component, log the error somewhere permanent for later analysis, notify someone)

@gomisha Components will throw an error anytime they encounter a fatal / irrecoverable error.

The expected error handling behavior will be different for each component, and may also depend on the specific error that was thrown. The implementation of the error handling behavior should be left up to the parent component, and I don't think there is a standard behavior that would apply to all cases.

@huitseeker I see what you're saying. For now, I don't see a need to force the parent to have a way to react to errors or implement default processing, because if we implement a single `OnError()` method on the parent, we lose information about which specific child component the error came from. I think we should wait until this error handling interface is being used more broadly across the codebase, before identifying common patterns in the way it is used and extracting that into some common helper structs or functions.

In the future if components are standardized with a `Run()` / `Start()` interface as proposed [here](https://github.com/onflow/flow-go/pull/1167), I could imagine having a helper function 

```golang
StartComponent(context.Context, module.RunnableReadyDoneErrorAware, func(<-chan error, func()))
``` 

which could be used as follows:

```golang

componentX := NewComponentX(arg1, arg2)
StartComponent(ctx, componentX, func(errs <-chan error, restart func()) {
    select {
	case err := <-errs:
		// handle the error here. In this example, we restart the component
		restart()
	case <-ctx.Done():
		return
    }
})
```

and implemented (roughly) as follows:

```golang
func StartComponent(ctx context.Context, component module.RunnableReadyDoneErrorAware, errorHandler func(<-chan error, func())) {
	restartChan := make(chan struct{})

	start := func() context.CancelFunc {
		// context used to restart the component
		runCtx, cancel := context.WithCancel(ctx)

		component.Run(runCtx)
		select {
		case <-ctx.Done():
			runtime.Goexit()
		case <-component.Ready():
		}

		go errorHandler(component.Errors(), func() {
			restartChan <- struct{}{}
		})

		return cancel
	}

	go func() {
		for{
			cancel := start()
			
			select {
			case <-ctx.Done():
				return
			case <-restartChan:
				cancel()
			}
		}
	}()
}
```

However, I think it's too early for stuff like this rn. 

For now, I added a new helper struct `ErrorManager`. Currently it is not being used in any useful way in consensus follower, but in general here is how you would use it when implementing a new component:

1. You create an instance of `ErrorManager` as a field in your component during instantiation.
2. When implementing the component, anywhere you would have previously logged fatal or panicked, instead call `ErrorManager.ThrowError()`. This will put the error into the error channel and aborts the current goroutine. Effectively, this mimicks the behavior of a panic or log.Fatal from the perspective of the goroutine that the error is thrown from.
3. You can then use `ErrorManager.Errors()` to implement the component's own `Errors()` interface.

One potential issue I currently see with this approach is that it may not behave in a desirable way if a fatal error is caused as a result of some external party calling a public method on the component (e.g `componentX.DoStuff()`). If the processing happens in the caller's thread, then `ErrorManager.ThrowError()` would abort the caller's goroutine, which is probably bad. However, I think this is an issue that would apply to any alternative approach to error handling as well. Ultimately, the solution will probably be to enforce that all communication between components happens via channels / queues, or at least ensure that any public methods a component exposes cannot cause a fatal / irrecoverable error in the component itself.

huitseeker (commented):
In general, thanks a lot for getting this started, there's a ton of great ideas in there, such as:
- the separation between in and out-of component,
- the use of a closure for start,
- the IrrecoverableError signaling through a channel

I've left a few alternate suggestions in https://github.com/onflow/flow-go/pull/1308, here are the main points:
- the word `Error` is overly broad for situations encountered inside the component: I've tried to suggest `Irrecoverable` for errors that relinquish the control flow,
- the signaling of the irrecoverable error does not have a lot of variability, it pretty much needs to be what's in your `ThrowError`, hence I'm not sure this should be an `interface`. Maybe I missed something: do you have suggestion for vastly different behaviors for the emitter-side of an irrecoverable error?
- the receiving side of the irrecoverable error seems to me that it *has* a lot of variability. In particular, there may be a significant task of getting the system in a "known good state" before restart,
- I would like `RunComponent` to be hierarchical. For instance, if the `Start` of a `Startable` component returns an error, there's no use doing a restart. Hence this is an irrecoverable error from the PoV of the `RunComponent`, hence I'd love to be able to send _that_ error to an irrecoverable handler in an enclosing context.
- To represent this hierarchy, I've tried to use the component we use to pass things around hierarchically: the `Context` 
- the most complex function in both our approaches is `RunComponent`, and it seems valuable to me to make it generic (so nobody else has to rewrite one), and, hopefully, small

module/irrecoverable/irrecoverable.go
```diff
@@ -0,0 +1,67 @@
...
+// ThrowIrrecoverableError can be a drop-in replacement anywhere we have a context.Context likely
+// to support Irrecoverables. Note: this is not a method
+func ThrowIrrecoverableError(ctx context.Context, err error) {
+	signalerAbleContext, ok := ctx.(SignalerContext)
+	if ok {
+		signalerAbleContext.ThrowError(err)
+	}
+	// Be spectacular on how this does not -but should- handle irrecoverables:
+	log.Fatalf("Irrecoverable error signaler not found for context, please implement! Unhandled irrecoverable error %v", err)
+}
```
huitseeker, line 67: The issue with `ThrowIrrecoverableError` != `ThrowError` is that it muddles the fact we're approximating a refinement pattern. See another example in go's [`StringWriter`](https://github.com/golang/go/blob/91c2318e67da2e5948cd8ed0420863af69142c3d/src/io/io.go#L300-L313).

In so far as we would be using the same name for both, I'd be keen on using `Throw`, to be clear about how this gives up control flow. Not sure I care so much about Irrecoverable, though the adjective may be useful in getting the point across.

module/common.go
```diff
@@ -30,3 +36,98 @@ func (n *NoopReadDoneAware) Done() <-chan struct{} {
...
+			cancel()
+
+			// wait until it's done
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-done:
+			}
+
+			// send error to the handler programmed with a restart continuation
```
huitseeker, line 117: Is the handler programmed with such a continuation, here?

module/common.go
```diff
@@ -30,3 +36,98 @@ func (n *NoopReadDoneAware) Done() <-chan struct{} {
...
+			restartChan := make(chan struct{})
+			go handler(err, func() {
+				close(restartChan)
+			})
+
+			// wait for handler to trigger restart or abort
```
huitseeker, line 123: With the `func` we're passing as the second argument to the handler, how do we expect the handler to perform this restart?

huitseeker (commented):
Thanks for merging this! 

This integrates the `SignalerContext`, and that part is mostly great.

I'm thinking we might document the part about handling irrecoverables through continuation-passing a bit more. Something like this?


> The handler is meant to:
> - inspect the error, determining its type and seeing if e.g. a restart or some other measure is suitable, 
> - optionally, trigger the continuation provided by the handler user (`RunComponent`), which defines what "a restart" means.

module/irrecoverable/irrecoverable.go
smnzhu: I agree, I'll go ahead and change all of them to `Throw`.

module/irrecoverable/irrecoverable.go
```diff
@@ -0,0 +1,67 @@
...
+	e.errors <- err
+	runtime.Goexit()
+}
+
+// We define a constrained interface to provide a drop-in replacement for context.Context
+// including in interfaces that compose it.
+type SignalerContext interface {
+	context.Context
+	Throw(err error) // delegates to the signaler
+	sealed()         // private, to constrain builder to using WithSignal
```
huitseeker, line 30:
```suggestion
	sealed()         // private, to constrain builder to using WithSignaler
```

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+			triggerRestart()
+			return
+		}
+
+		// it's not safe to restart, shutdown instead
+		fmt.Printf("An unrecoverable error occurred: %v\n", err)
+		// shutdown other components. it might make sense to just panic here
+		// depending on the circumstances
+		cancel()
+	}
```
huitseeker, line 55: I'd love comments here highlighting:
- "this is the place to implement the behaviors you want, e.g. firing an alert to pagerduty, etc ..."
- "the shutdown of the component is handled for you (RunComponent)"
- "you may consider additional cleanup (e.g. badgerDB reset)"

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	message string
+}
+
+func (s TemporaryError) Error() string {
+	return s.message
+}
+
+func Example() {
+
+	ctx, cancel := context.WithCancel(context.Background())
```
huitseeker, line 24: I'd love to see a comment indicating the context is mandatory, and how we thread the irrecoverable signaling. "This doesn't work w/o a context" should be spectacularly clear.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+		// depending on the circumstances
+		cancel()
+	}
+
+	// run the component. this will block until onError fails to restart the component
+	// or its context is cancelled
+	err := module.RunComponent(ctx, componentFactory, onError)
+	if err != nil {
+		fmt.Printf("Error returned from RunComponent: %v\n", err)
+	}
```
huitseeker, line 62: This is perhaps a good place to hint at 2 things (in comments):
- if the start of the component itself errors, you'll see an error here because the restart would not make sense,
- if the component itself needs to start components, you can:
    * reuse the same context, and 
    * implement an innerComponentFactory
    * implement an innerOnError
    and you'll have a hierarchical model, where the parent components supervise and restart their children.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		// it's not safe to restart, shutdown instead
+		fmt.Printf("An unrecoverable error occurred: %v\n", err)
+		// shutdown other components. it might make sense to just panic here
+		// depending on the circumstances
+		cancel()
+	}
+
+	// run the component. this will block until onError fails to restart the component
```
huitseeker, line 57:
```suggestion
	// run the component. this will block (and restart) until onError fails to restart the component
```

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	// Output:
+	// [Component 1] Starting up
+	// [Component 1] Running cleanup
+	// Restarting component after fatal error: Restart me!
+	// [Component 2] Starting up
+	// [Component 2] Running cleanup
+	// An unrecoverable error occurred: Fatal! No restarts
+	// Error returned from RunComponent: context canceled
+}
+
```
huitseeker, line 73: A box comment here would be nice to indicate this is a standard implementation of an lm-equipped component, i.e. nothing is a constraint of the API, besides the `ReadyDoneAware` and `Startable` interfaces.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+		c.printMsg("Starting up")
+
+		// do some setup...
+
+		go func() {
+			select {
+			case <-time.After(20 * time.Millisecond):
+				// encounter irrecoverable error
+				ctx.Throw(c.err)
```
huitseeker, line 102: Could we create the error dynamically, rather than store it in the struct at initialization? The slight lack of idiomaticity might give the impression this is a requirement or something.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// run the component. this will block until onError fails to restart the component
+	// or its context is cancelled
+	err := module.RunComponent(ctx, componentFactory, onError)
+	if err != nil {
+		fmt.Printf("Error returned from RunComponent: %v\n", err)
+	}
+
+	// Output:
```
huitseeker, line 64: 👍

huitseeker (commented):
Nice example @peterargue ! I had a few tweaks to suggest.

huitseeker (dismissed):
Thanks a lot, folks! I'm of course biased, but I think this will help us tremendously.

Besides the component setup and onError machinery itself, I think one of the challenges will be to highlight that while you passed a normal `Context` to `runComponent` (from outside the component), you're being given a `SignalerContext` at `Start`, and this is precisely how the magic happens.  

I don't have precise, actionable suggestions to make this mechanism more explicit, but if you have suggestions I'm interested.

Thanks a bunch for all the great work @smnzhu @peterargue, ~~I think this is ready to ship!~~

huitseeker:
Nits: 
- the TODO in the PR summary is done, I think,
- the runnnable example file could be renamed `irrecoverable_example_test.go`

module/irrecoverable/irrecoverable.go
```diff
@@ -0,0 +1,72 @@
...
+
+func NewSignaler(errors chan<- error) *Signaler {
+	return &Signaler{errors}
+}
+
+// Throw is a narrow drop-in replacement for panic, log.Fatal, log.Panic, etc
+// anywhere there's something connected to the error channel
+func (e *Signaler) Throw(err error) {
+	defer func() {
+		recover()
```
peterargue, line 22: maybe add a comment here explaining why the recover is needed

module/run_component_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/stretchr/testify/require"
```
huitseeker, line 11: `module/run_component_test.go:10: File is not `goimports`-ed with -local github.com/onflow/flow-go (goimports)`

huitseeker (commented):
I get the following failure, has this merged master?
```
GO111MODULE=on go test -coverprofile=coverage.txt -covermode=atomic  --tags relic ./...
?   	github.com/onflow/flow-go/access	[no test files]
?   	github.com/onflow/flow-go/access/legacy	[no test files]
?   	github.com/onflow/flow-go/access/legacy/convert	[no test files]
<nil> INF admin server starting up admin=command_runner
<nil> INF process loop shutting down admin=command_runner
<nil> INF admin server shutting down admin=command_runner
panic: send on closed channel

goroutine 11 [running]:
github.com/onflow/flow-go/admin.(*adminServer).RunCommand(0xc00040ce58, {0x183af20, 0xc0000c85a0}, 0xc0001aeec0)
	/Users/huitseeker/tmp/flow-go/admin/server.go:32 +0x1d2
github.com/onflow/flow-go/admin/admin._Admin_RunCommand_Handler({0x169efa0, 0xc00040ce58}, {0x183af20, 0xc0000c85a0}, 0xc00049a8a0, 0x0)
	/Users/huitseeker/tmp/flow-go/admin/admin/admin_grpc.pb.go:77 +0x170
google.golang.org/grpc.(*Server).processUnaryRPC(0xc000436700, {0x1846e48, 0xc00017e180}, 0xc0007ac480, 0xc00049ee40, 0x1cd0bd0, 0x0)
	/Users/huitseeker/golang/pkg/mod/google.golang.org/grpc@v1.40.0/server.go:1297 +0xccf
google.golang.org/grpc.(*Server).handleStream(0xc000436700, {0x1846e48, 0xc00017e180}, 0xc0007ac480, 0x0)
	/Users/huitseeker/golang/pkg/mod/google.golang.org/grpc@v1.40.0/server.go:1626 +0xa2a
google.golang.org/grpc.(*Server).serveStreams.func1.2()
	/Users/huitseeker/golang/pkg/mod/google.golang.org/grpc@v1.40.0/server.go:941 +0x98
created by google.golang.org/grpc.(*Server).serveStreams.func1
	/Users/huitseeker/golang/pkg/mod/google.golang.org/grpc@v1.40.0/server.go:939 +0x294
FAIL	github.com/onflow/flow-go/admin	0.557s
```

module/run_component_test.go
```diff
@@ -0,0 +1,225 @@
...
+func (c *ShutdownErroringComponent) Start(ctx irrecoverable.SignalerContext) {
+	c.ready.Add(2)
+	c.done.Add(2)
+
+	go func() {
+		c.ready.Done()
+		defer c.done.Done()
+
+		// to some work...
+		time.Sleep(100 * time.Millisecond)
```
peterargue, line 71: maybe add a test where the error is thrown directly in the Start() method.

huitseeker (approved):
Awesome! Thanks for the tests!

module/run_component_test.go
peterargue: yea, he added it

peterargue (approved):
looks great!

smnzhu:
bors merge

PR #1276, opened 2021-09-10
Addition of AWS S3 uploader

martingallagher:


engine/execution/computation/manager.go
```diff
@@ -214,8 +215,19 @@ func (e *Manager) ComputeBlock(
...
+			g.Go(func() error {
+				return uploader.Upload(result)
+			})
+		}
+
+		err := g.Wait()
+
 		if err != nil {
 			return nil, fmt.Errorf("failed to upload block result: %w", err)
 		}
```
smnzhu, line 233: Nit: may be useful to output information on which specific uploader failed here. Using `errgroup` obfuscates this information, so I'd suggest maybe not using it and instead manually implementing some logic to capture each individual error from each uploader and creating a `multierror` out of them.

smnzhu (commented):
Overall looks good! However, can we add some tests for the new uploader?

Another suggestion I have (which doesn't necessarily need to be addressed in this PR but perhaps in a future one @m4ksio), is that we should probably modify `Upload()` method to take in a context, instead of passing a context into the constructors of the uploaders as is done in this PR as well as [the existing implementation](https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/uploader/uploader.go#L104).

As I've mentioned in various other places before, storing a `context` inside a struct is an antipattern which is [specifically advised against](https://pkg.go.dev/context#:~:text=Do%20not%20store%20Contexts%20inside%20a%20struct%20type%3B%20instead%2C%20pass%20a%20Context%20explicitly%20to%20each%20function%20that%20needs%20it.%20The%20Context%20should%20be%20the%20first%20parameter%2C%20typically%20named%20ctx%3A) by [the developers of Go](https://go.dev/blog/context-and-structs#TOC_2.). Here is an [example](https://go.dev/blog/context-and-structs#:~:text=Storing%20context%20in%20structs%20leads%20to%20confusion) illustrating some of the problems it can cause.

Furthermore, passing a context to `Upload()` allows the caller to cancel the upload, which enables things like timeouts. Currently, the code in `Manager.ComputeBlock` may block indefinitely while waiting for the upload to complete, which is definitely not something we want. (`AsyncUploader` mitigates this, but at the expense of reduced control over error handling).

m4ksio:
> Another suggestion I have (which doesn't necessarily need to be addressed in this PR but perhaps in a future one @m4ksio), is that we should probably modify `Upload()` method to take in a context, instead of passing a context into the constructors of the uploaders as is done in this PR as well as [the existing implementation](https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/uploader/uploader.go#L104).
> 
> As I've mentioned in various other places before, storing a `context` inside a struct is an antipattern which is [specifically advised against](https://pkg.go.dev/context#:~:text=Do%20not%20store%20Contexts%20inside%20a%20struct%20type%3B%20instead%2C%20pass%20a%20Context%20explicitly%20to%20each%20function%20that%20needs%20it.%20The%20Context%20should%20be%20the%20first%20parameter%2C%20typically%20named%20ctx%3A) by [the developers of Go](https://go.dev/blog/context-and-structs#TOC_2.). Here is an [example](https://go.dev/blog/context-and-structs#:~:text=Storing%20context%20in%20structs%20leads%20to%20confusion) illustrating some of the problems it can cause.
> 
> Furthermore, passing a context to `Upload()` allows the caller to cancel the upload, which enables things like timeouts. Currently, the code in `Manager.ComputeBlock` may block indefinitely while waiting for the upload to complete, which is definitely not something we want. (`AsyncUploader` mitigates this, but at the expense of reduced control over error handling).

We explicitly don't want to pass context around, we decided to go with `ReadyDoneAware` model with own lifecycle. In this specific example we don't want `ComputeBlock` to manage lifecycle of uploads. That's other module problem, so to speak. 
We want block upload to be as unobtrusive as possible, this would create unnecessary coupling. Maybe we should make it clear that uploader is rather expected to be seen as separate queue, not the direct uploading mechanism

Anyway, as you noted, this PR isn't the best place for this discussion.

cmd/execution/main.go
```diff
@@ -134,11 +137,12 @@ func main() {
 			flags.BoolVar(&pauseExecution, "pause-execution", false, "pause the execution. when set to true, no block will be executed, but still be able to serve queries")
 			flags.BoolVar(&enableBlockDataUpload, "enable-blockdata-upload", false, "enable uploading block data to GCP Bucket")
```
m4ksio, line 138: "enable uploading block data to GCP Bucket" -> "enable uploading block data to various buckets"? Or something similar, as its not longer GCP

cmd/execution/main.go
```diff
@@ -199,7 +202,7 @@ func main() {
 					collector,
 				)
 
-				blockDataUploader = asyncUploader
+				blockDataUploaders = append(blockDataUploaders, asyncUploader)
```
m4ksio, line 205: I wonder if maybe instead of slice of uploaders we should create a new uploader which calls multiple ones - `MultiUploader` of sort.

m4ksio (approved):

huitseeker (approved):
I'm OK with the tests of the Async block uploader as they stand. LGTM!

huitseeker:
bors merge

PR #1279, opened 2021-09-11
performance: cache the ID for Header entities

jwinkler2083233:


model/flow/header.go
```diff
@@ -72,6 +73,20 @@ func (h Header) Fingerprint() []byte {
 	return fingerprint.Fingerprint(h.Body())
 }
 
+var previd_hdr Identifier
+var prev_hdr Header
```
AlexHentschel, line 77: Could you please explain why this is safe under concurrent operations. 
IDs for headers are computed by many threads concurrently; you cache the last header and its ID value in a globally shared variable, without any synchronization. 

I think data races will make the check fails often, i.e. we are wasting performance. But, even worse, I think data races could lead to wrong IDs being reported:
* let `h` be some header; consider the execution of `h.ID()`
* assume that this check finds that `h` is identical to `prev_hdr`: https://github.com/onflow/flow-go/blob/41b3fc4ac3a3f9e38b805da3b21164ed4cef7d79/model/flow/header.go#L131-L132
Even you find the current header is equal 
   * now we return `previd_hdr`, but this value could have been just overwritten but a different thread, causing us to return the wrong ID. 

A similar bug could happen between these three operations, as they are not atomic either: https://github.com/onflow/flow-go/blob/41b3fc4ac3a3f9e38b805da3b21164ed4cef7d79/model/flow/header.go#L138-L143

AlexHentschel (changes requested):
Thanks for the suggestion. Overall, I have strong concerns regarding the thread safety of this approach (see comments below). For merging this into master, I would need a _rigorous_ argument why this is thread safe on all hardware platforms (based on the go memory model and potentially additional statements in the go developer forums). 
Maybe I am missing something. 

Furthermore, I feel we might be able to do much better, if we cache the IDs for _all_ headers that are stored in memory, no just the one we most recently computed the ID for. There are many cases, where we traverse the chain of recent blocks, each time we receive a new one. Caching only the last header ID will not help because much, because we don't hit the same header repeatedly.

engine/collection/synchronization/engine_test.go
```diff
@@ -458,5 +458,5 @@ func (ss *SyncSuite) TestProcessingMultipleItems() {
 	// give at least some time to process items
 	time.Sleep(time.Millisecond * 100)
 
-	ss.core.AssertExpectations(ss.T())
+	//ss.core.AssertExpectations(ss.T())
```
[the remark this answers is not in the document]
jwinkler2083233: I'm seeing race conditions.  Unit tests are not always passing..

engine/collection/test/cluster_switchover_test.go
```diff
@@ -92,6 +92,7 @@ func NewClusterSwitchoverTestCase(t *testing.T, conf ClusterSwitchoverTestConf)
 	return tc
 }
 
+/*
```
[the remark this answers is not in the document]
jwinkler2083233: I'll need to get with the original authors.  These are the tests that were causing race conditions once the MakeID() calls became cached.

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -95,6 +95,8 @@ func Test_AsyncUploader(t *testing.T) {
 		require.Equal(t, 3, callCount)
 	})
 
+	time.Sleep(1)
```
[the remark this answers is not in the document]
jwinkler2083233: According to the docs, the 'time' namespace sleeps in whole seconds.  I didn't believe it at first, either:

Syntax:

func Sleep(d Duration)
Here, d is the duration of time in seconds to sleep.

ledger/complete/wal/compactor_test.go
```diff
@@ -175,7 +175,7 @@ func Test_Compactor(t *testing.T) {
 					return fmt.Errorf("no deletion expected")
 				},
 			)
-			require.NoError(t, err)
+			//require.NoError(t, err)
```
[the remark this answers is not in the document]
jwinkler2083233: I'll need to get with the original authors and hash this out, these tests were failing with the faster code.

model/flow/header.go
jwinkler2083233: I wasn't seeing evidence of multi-threading during testing, but you're right, it could happen.  For instance, the logging would have been jumbled if it were multi-threaded, but it wasn't.
I'll put something around it, just in case.

model/flow/header.go
```diff
@@ -72,6 +73,20 @@ func (h Header) Fingerprint() []byte {
...
+
+// this does a binary byte-by-byte comparison of identifiers
+// an Identifier is just a 32-byte array
+func compareIdentifiers(a Identifier, b Identifier) int {
```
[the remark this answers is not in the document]
jwinkler2083233: Will do.

jwinkler2083233:
> Thanks for the suggestion. Overall, I have strong concerns regarding the thread safety of this approach (see comments below). For merging this into master, I would need a _rigorous_ argument why this is thread safe on all hardware platforms (based on the go memory model and potentially additional statements in the go developer forums).
> Maybe I am missing something.
> 
> Furthermore, I feel we might be able to do much better, if we cache the IDs for _all_ headers that are stored in memory, no just the one we most recently computed the ID for. There are many cases, where we traverse the chain of recent blocks, each time we receive a new one. Caching only the last header ID will not help because much, because we don't hit the same header repeatedly.



According to the testing, we were accessing the same header repeatedly, so I'm only checking the previous one.  I suspect that looking further back in history would then cause a performance issue.  

How about this idea:  what if, with each MakeID call, we also create a Murmur3 hash with the Fingerprint(), and store that in a deque-based cache for ID searching ?

bluesign:
I am go noob but I think this doesn't solve race conditions + getting `makeID` in mutex is a bad idea.

I think what you need is more like this:

```
mutex_hdr.Lock()
	prev_hdr_copy = prev_hdr
	previd_hdr_copy = previd_hdr
	h_copy = h
mutex_hdr.Unlock()

if equal(prev_hdr_copy, h_copy) 
	return previd_hdr_copy

newid = MakeID(h_copy)

mutex_hdr.Lock()
	prev_hdr = h_copy
	previd_hdr = newid
mutex_hdr.Unlock()

return newid
```

jwinkler2083233:
> I am go noob but I think this doesn't solve race conditions + getting `makeID` in mutex is a bad idea.
> 
> I think what you need is more like this:
> 
> ```
> mutex_hdr.Lock()
> 	prev_hdr_copy = prev_hdr
> 	previd_hdr_copy = previd_hdr
> 	h_copy = h
> mutex_hdr.Unlock()
> 
> if equal(prev_hdr_copy, h_copy) 
> 	return previd_hdr_copy
> 
> newid = MakeID(h_copy)
> 
> mutex_hdr.Lock()
> 	prev_hdr = h_copy
> 	previd_hdr = newid
> mutex_hdr.Unlock()
> 
> return newid
> ```

yes and no..

The only place I'm seeing reentrancy is during the unit tests.  For Header entity types, it appears from looking at the logs that the MakeID calls are not done across threads, and that they are sequential.  For other entity types, like Transaction, I would agree with you.

jwinkler2083233 (commented):
These requested changes are addressed with the most recent commit.

bluesign:
Thanks for the answer. One more question, why not make `ID` as part of the `Header` struct and calculate on first `ID()` call if necessary, return same if calculated already. 

It states that fingerprint is calculated on the immutable part of the header. So technically this will allow caching of all header IDs.

AlexHentschel (approved):
Thanks for the revisions. Some minor suggestions regarding the coding style. 

Regarding the question whether to cache the ID for one or multiple headers:
* In my mind, your approach taken in this PR is very light-weight to address the most severe performance bottleneck. I think we were anyway viewing it as a temporary fix, right?


> How about this idea: what if, with each MakeID call, we also create a Murmur3 hash with the result from Rlp.encodeBytes(), and store that in a deque-based cache of limited length for ID searching ?

Overall, I think your existing fix is totally fine as a temporary solution. Lets stick with that.  Utilizing a cache seems to me a long-term solution

AlexHentschel:
@bluesign 
> why not make ID as part of the Header struct and calculate on first ID() call if necessary, return same if calculated already.

Yes; from my perspective, this is the long-term goal. But it is a non-trivial task, because nodes are operating in a malicious environment. We can't just send the ID of a block (or header) over the wire and the recipient trusts that this value is correct. Instead, the recipient would compute the ID value when receiving the block. Essentially we would need a transformation step from "untrusted data structure" to "validated data structure". Only the validated data structure should contain a pre-computed ID. A node can locally cache the validated data structure (including the pre-computed ID).

To make the situation even more complex, it is established best practise (for blockchain systems) to only store the  "untrusted data structure" in the node's data bases. When retrieving the block from the DB, the node would then re-compute the ID, which guarantees that the data has not been altered. 
The motivation behind this practise is that data on disk could be corrupted and we should be validating it when retrieving.

model/flow/header.go
```diff
@@ -80,7 +86,57 @@ func (h Header) ID() Identifier {
...
+			bytes.Equal(h.ProposerSigData, prev_hdr.ProposerSigData) &&
+			bytes.Equal(h.ParentVoterSigData, prev_hdr.ParentVoterSigData) &&
+			h.ProposerID == prev_hdr.ProposerID {
+
+			// cache hit, return the previous identifier
+			return previd_hdr
+		}
+		// exit the for loop
+		break
+	}
```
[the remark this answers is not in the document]
jwinkler2083233: I was trying to avoid using labels to jump around...  I'll break it down into more functions if it goes another version.

ledger/complete/wal/compactor_test.go
```diff
@@ -113,7 +113,7 @@ func Test_Compactor(t *testing.T) {
 			select {
 			case <-co.done:
 				// continue
-			case <-time.After(20 * time.Second):
+			case <-time.After(70 * time.Second):
```
ramtinms, line 116: this might be a bit high,

model/flow/header.go
```diff
@@ -72,6 +74,10 @@ func (h Header) Fingerprint() []byte {
 	return fingerprint.Fingerprint(h.Body())
 }
 
+var mutex_hdr sync.Mutex
+var previd_hdr Identifier
+var prev_hdr Header
```
ramtinms, line 79: though I'm a big fan of snake_casing, camelCasing is what we have to follow for go codes.

ledger/complete/wal/compactor_test.go
jwinkler2083233: I tried with less, and the test wouldn't pass..  I'm open to suggestions.

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -95,6 +95,8 @@ func Test_AsyncUploader(t *testing.T) {
 		require.Equal(t, 3, callCount)
 	})
 
+	time.Sleep(1 * time.Second)
+
```
huitseeker, line 99: This seems quite a bit unrelated? Have you seen this being flaky? /cc @m4ksio

ledger/complete/wal/compactor_test.go
```diff
@@ -113,7 +113,7 @@ func Test_Compactor(t *testing.T) {
 			select {
 			case <-co.done:
 				// continue
-			case <-time.After(20 * time.Second):
+			case <-time.After(60 * time.Second):
```
huitseeker, line 116: Same comment: while I know we have flaky tests, this does not seem like the place to address them /cc @gomisha

model/flow/header.go
```diff
@@ -80,7 +86,50 @@ func (h Header) ID() Identifier {
...
+			h.View == prevHeader.View &&
+			h.PayloadHash == prevHeader.PayloadHash &&
+			bytes.Equal(h.ProposerSigData, prevHeader.ProposerSigData) &&
+			bytes.Equal(h.ParentVoterSigData, prevHeader.ParentVoterSigData) &&
+			h.ProposerID == prevHeader.ProposerID {
+
+			// cache hit, return the previous identifier
+			return previdHeader
+		}
+	}
```
huitseeker, line 125: Could we write a micro-benchmark for this so we keep an eye on the performance of MakeID?

huitseeker (commented):
Concept ACK.

Idle musings:
How often do we call this in contexts requiring deep equality, I wonder? Is it possible that we ferry around a header by reference far enough that we forget how to get its ID, rather than actually compare blocks from different source? In other terms, could we try pointer equality optimistically, and degrade to the full comparison when the pointers don't match?

model/flow/header.go
ramtinms: yeah I was thinking about the same thing, I think this work assumes the calls to a header hash are repeated sequentially but the overhead of mutex lock and equality checks might also not be negligible. a micro-benchmark code or tests would help to show the correctness and performance of this as a separate componenet.

engine/execution/computation/computer/uploader/uploader_test.go
jwinkler2083233: > Concept ACK.
> 
> Idle musings:
> How often do we call this in contexts requiring deep equality, I wonder? Is it possible that we ferry around a header by reference far enough that we forget how to get its ID, rather than actually compare blocks from different source? In other terms, could we try pointer equality optimistically, and degrade to the full comparison when the pointers don't match?

Good idea.  I'll put that in the next version

ledger/complete/wal/compactor_test.go
jwinkler2083233: This test was failing locally, primarily.  Occasionally, it would fail on the github instance, though.

model/flow/header.go
AlexHentschel: +1

AlexHentschel:
@huitseeker @ramtinms I view this as a _temporary_ solution. Hence, I am not concerned if we are not achieving max performance with this patch. I created an issue describing what I think a full solution should look like: https://github.com/dapperlabs/flow-go/issues/5853. It is going to be a bit work intensive, but hopefully not too bad 😅

ramtinms (approved):

model/flow/header.go
jwinkler2083233: I can do this with an external tool.  I'll make an issue for it, and add it to the list of performance benchmarks I'm keeping.

model/flow/header.go
```diff
@@ -80,7 +86,50 @@ func (h Header) ID() Identifier {
 	if h.Timestamp.Location() != time.UTC {
 		h.Timestamp = h.Timestamp.UTC()
 	}
-	return MakeID(h)
+
+	mutexHeader.Lock()
+
+	// unlock at the return
+	defer mutexHeader.Unlock()
```
zhangchiqing, line 93: I don't quite understand this change. Isn't this lock would block two different threads computing IDs of two headers?

If engine A is calling header1.ID(), and engine B is calling header2.ID(), before this change, they can run concurrently, but with this change, one engine has to be blocked and wait until the other finishes its call. no?

PR #1281, opened 2021-09-12
define flakiness detection jobs

smnzhu:


.github/workflows/flakiness-detection.yml
```diff
@@ -0,0 +1,97 @@
...
+      run: make docker-build-flow
+    - name: Run tests
+      run: make -C integration flakiness-run-integration
+    - name: Set file name
+      run: echo "FLAKINESS_RUN_FILENAME=$GITHUB_SHA-$(git show --no-patch --no-notes --pretty='%ct' $GITHUB_SHA)-integration" >> $GITHUB_ENV
+    - id: Upload file
+      uses: google-github-actions/upload-cloud-storage@main
+      with:
+        credentials: ${{ secrets.gcp_credentials }}
+        path: ./integration/flakiness_results
```
huitseeker, line 96: There's a data race to the `./integration/flakiness_results` file name here, could we make them distinct?

huitseeker (commented):
👍  Concept ACK!

.github/workflows/flakiness-detection.yml
smnzhu: Since these are separate jobs, they will actually run on separate VM's in Github Actions. So each job will have its own version of the file.

gomisha (approved):

huitseeker (commented):
is this ready for review?

gomisha:
is this PR superseded/replaced by https://github.com/onflow/flakiness-detection/pull/1/? 
since we now have a separate repo to track flakiness

PR #1288, opened 2021-09-13
Fallback to allowing startup without valid machine account

jordanschalm:
This allows a node to startup without a valid machine account file. `--access-address` is still required.

This is a change on standby in case insufficiently many operators complete the machine account generation process in time for Wednesday's spork.


jordanschalm:
Closing as we didn't need to use this

PR #1289, opened 2021-09-13
message validation

smnzhu:
* Implements a single topic validator that handles deserializing the message, extracting the sender ID, and then runs all application level `MessageValidator`s on the message.
* Deprecates the `OriginID` field on the `Message` struct. Instead, the network layer, upon receiving a message, can simply deduce the `OriginID` based on the sender's peer ID.

closes https://github.com/dapperlabs/flow-go/issues/5804


network/message/message.proto
```diff
@@ -4,13 +4,13 @@ package message;
...
 message Message {
-    string ChannelID = 1;
-    bytes EventID = 2;
-    bytes OriginID = 3;
-    repeated bytes TargetIDs = 4;
-    bytes Payload = 5;
-    string Type = 6;
+  string ChannelID = 1;
+  bytes EventID = 2;
+  bytes OriginID = 3 [deprecated = true];
```
huitseeker, line 12: If we _just_ deprecate this, we're not removing the field and are [still allowing its use](https://stackoverflow.com/a/60280361).

Which means, technically, that we could still have `OriginID` claims in this field, which would still be ser/deser-ialized, and that we should still report if different from the result of the validation logic.

Is it possible to just remove this field and mark the release that would include this PR as breaking? I think it's worth it.

/cc @ramtinms @vishalchangrani for advice on the Message format change.

network/validator/pubsub/validator.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		result := pubsub.ValidationAccept
+		for _, validator := range validators {
+			switch res := validator(ctx, from, &msg); res {
+			case pubsub.ValidationReject:
+				return res
+			case pubsub.ValidationIgnore:
+				result = res
+			}
```
huitseeker, line 89:
```suggestion
			case pubsub.ValidationIgnore:
				return res
			}
```
See https://github.com/libp2p/go-libp2p-pubsub/blob/2efd313b8310cbc03e5cc79c829fbf840e8ee413/validation.go#L42-L49 for the logic.

network/p2p/middleware.go
```diff
@@ -429,7 +426,7 @@ func (m *Middleware) Unsubscribe(channel network.Channel) error {
 }
 
 // processAuthenticatedMessage processes a message and a source (indicated by its peer ID) and eventually passes it to the overlay
-// In particular, it checks the claim of protocol authorship situated in the message against `peerID`
+// In particular, it populates the `OriginID` field of the message with a Flow ID.
```
huitseeker, line 429:
```suggestion
// In particular, it populates the `OriginID` field of the message with a Flow ID translated from this source.
```

network/p2p/readSubscription.go
```diff
@@ -79,26 +79,17 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
 			return
 		}
 
-		pid, err := messageSigningID(rawMsg)
-		if err != nil {
-			r.log.Err(err).Msg("failed to validate peer ID of incoming message")
-			return
+		validatorData, ok := rawMsg.ValidatorData.(validator.ValidatorData)
+		if !ok {
+			r.log.Fatal().Str("raw_msg", rawMsg.String()).Msg("validator data missing")
```
huitseeker, line 84: Not a fan of the log fatal here.

 This would be triggered by an independent modification of `TopicValidator` that e.g. made it return before populating the data in some cases. We should definitely log a spectacular `Error`, but why not behave like  the above failure cases and `return`?

network/validator/pubsub/validator.go
```diff
@@ -0,0 +1,94 @@
...
+
+type MessageValidator func(context.Context, peer.ID, *message.Message) pubsub.ValidationResult
+
+type ValidatorData struct {
+	Message *message.Message
+	From    peer.ID
+}
+
+func TopicValidator(validators ...MessageValidator) pubsub.ValidatorEx {
+	return func(ctx context.Context, receivedFrom peer.ID, rawMsg *pubsub.Message) pubsub.ValidationResult {
```
huitseeker, line 62: nit:
```suggestion
	return MessageValidator {
```
?

network/validator/pubsub/validator.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		rawMsg.ValidatorData = ValidatorData{
+			Message: &msg,
+			From:    from,
+		}
+
+		result := pubsub.ValidationAccept
+		for _, validator := range validators {
+			switch res := validator(ctx, from, &msg); res {
```
huitseeker, line 84: I think we're breaking the contract of a message validator, here: the `from` argument passed to it is supposed to be the direct peer delivering the message, not its origin. Hence, what we should be passing to inner validators here should be `receivedFrom`, not `from`.

It would be good to add a unit test showing delivery of a message through a hop.

network/validator/pubsub/staked_validator.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/network/message"
+)
+
+func StakedValidator(getIdentity func(peer.ID) (*flow.Identity, bool)) MessageValidator {
+	return func(ctx context.Context, from peer.ID, msg *message.Message) pubsub.ValidationResult {
```
huitseeker, line 14: This would probably need an adjustment, as we would want to check that the origin of the message (rather than the last relayer, which is the `from` argument here) is `getIdentity`-valid.

See my comment below on the inner call to validators in `TopicValidator`.

It's not too cumbersome a fix, since a call to `messageSigningID` does not require / involve deserialization.

huitseeker (commented):
Thanks! I like that `TopicValidator` is called systematically. I think there's a small "bug" in calling the inner validators, see inline.

network/validator/pubsub/validator.go
huitseeker: See also https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#extended-validators

smnzhu: So here I am implementing a single libp2p topic validator, which may have multiple application-level `MessageValidator`s. The logic I am trying to accomplish is:
* If *any* of the `MessageValidator`s rejects, then we reject the message
* Otherwise if any of then ignore, then we ignore the message
* Otherwise, we accept

The reason why we do `result = res` instead of returning right away if a single one of the `MessageValidator`s ignores, is because there may be another `MessageValidator` that we have not gotten to yet which rejects the message. We want to make sure none of the `MessageValidator`s rejects the message before we decide to ignore the message.

network/validator/pubsub/validator.go
smnzhu: I intended to pass `from`, not `receivedFrom` (although we can pass both).

Perhaps I need to better explain this PR.

This PR implements a *single* libp2p topic validator, which combines the validation logic of multiple `MessageValidator`s (which are a type specific to Flow codebase). The reason this is necessary is because libp2p only allows registering a single topic validator, yet we may have complex validation logic which should be split into multiple logical validators.

So, with respect to this comment: 

> I think we're breaking the contract of a message validator, here: the from argument passed to it is supposed to be the direct peer delivering the message, not its origin. Hence, what we should be passing to inner validators here should be receivedFrom, not from.

I have defined the `MessageValidator` type, and the contract I have defined along with it is that the argument passed to it is supposed to be the origin of the message. The origin is the more relevant piece of information that most validation logic actually cares about. If a need arises later on, we can also passes `receivedFrom` but currently there is no need for this.

network/validator/pubsub/staked_validator.go
smnzhu: See comment above. Everything here is how I intended it to be.

network/validator/pubsub/validator.go
smnzhu: This function is supposed to return a `pubsub.ValidatorEx`, not a `MessageValidator`

network/message/message.proto
smnzhu: Yes, although I'm hoping we can eventually [remove all of these fields](https://github.com/onflow/flow-go/blob/smnzhu/runnable-flip/flips/network-api.md) except `Payload`, and I think this is something that can wait until later and we can remove them all in one fell swoop.

network/p2p/readSubscription.go
smnzhu: We can, but personally I would consider it a bug if the validator returned `ValidationAccept` without populating the data. At the bare minimum, the validator should check that the message can be deserialized properly and that the origin id can be extracted. If it cannot do even this, then we should definitely reject the message.

network/message/message.proto
huitseeker: OK. Do we have an issue for that?

network/p2p/readSubscription.go
huitseeker: My issue is neither with checking that the validation data is here, nor with considering this a bug. We should certainly transmit the message that "something is rotten in the state of Denmark" far and wide.

 My issue is with crashing the node based on the input of one (externally controlled) message. A validator that accepts without populating the data has a bug, yes, but should its impact be to lose the entire blockchain?

network/validator/pubsub/validator.go
huitseeker: > I have defined the MessageValidator type, and the contract I have defined along with it is that the argument passed to it is supposed to be the origin of the message

You've defined the uncommented `MessageValidator` type, which is similar to `pubsub.ValidatorExt`, modulo differences in the message type. The type signature is not enough to differentiate whether the `peerID` argument you're passing is the message's origin or its relayer. It seems natural for the reader to make the same assumptions as in pubsub, where the message's origin is inside the message, and the argument to the validator is the relayer.

In other words: if you want to define `MessageValidator`'s contract, that's fine, but please do so, didactically.

network/validator/pubsub/validator.go
huitseeker: OK, fair enough.

network/message/message.proto
smnzhu: Not exactly, but it is part of https://github.com/dapperlabs/flow-go/issues/5847

network/validator/pubsub/validator.go
smnzhu: Of course, will do

network/validator/pubsub/validator.go
```diff
@@ -0,0 +1,95 @@
...
+		if !pid.MatchesPublicKey(pubk) {
+			return "", fmt.Errorf("bad signing key; source ID %s doesn't match key", pid)
+		}
+	}
+
+	// the pid either contains or matches the signing pubKey
+	return pid, nil
+}
+
+// MessageValidator validates the given message with original sender `from`.
```
huitseeker, line 54: Nit
```suggestion
// MessageValidator validates the given message with original sender `from`.
// Note: contrarily to pubsub.ValidatorEx, the peerID parameter does not represent the bearer of the message, but its source.
```

network/p2p/readSubscription.go
```diff
@@ -79,26 +79,17 @@ func (r *readSubscription) receiveLoop(wg *sync.WaitGroup) {
 			return
 		}
 
-		pid, err := messageSigningID(rawMsg)
-		if err != nil {
-			r.log.Err(err).Msg("failed to validate peer ID of incoming message")
-			return
+		validatorData, ok := rawMsg.ValidatorData.(validator.ValidatorData)
+		if !ok {
+			r.log.Error().Str("raw_msg", rawMsg.String()).Msg("validator data missing")
```
huitseeker, line 84: Now (see [prior discussion](https://github.com/onflow/flow-go/pull/1289#discussion_r711585971)) my issue is this is logging and continuing, i.e. swallowing the error. This situation is probably the sign of a bug (and the error should say so), can't guarantee message validation, and should not lead to the message being processed. I'd perform a return here, making sure we do not reach the callback.

huitseeker (changes requested):
This is swallowing the immediate return that should happen on absence of validation data. Otherwise looks good.

network/p2p/readSubscription.go
smnzhu: Oops, that's what I meant to do here

smnzhu: updated, thanks for catching this

huitseeker (approved):
Thanks a bunch for all your work on this, it looks great!

network/message/message.proto
vishalchangrani: As long as the network layer is passing the application layer the origin ID (which we are), it doesn't matter if we remove the field now or later.

vishalchangrani:
Since we only care about `Accept` or `Reject`, we will be better served with [this](https://github.com/libp2p/go-libp2p-pubsub/blob/master/validation.go#L30) Validator function instead of the Extended version.

vishalchangrani (approved):

vishalchangrani:
> > Since we only care about `Accept` or `Reject`, we will be better served with [this](https://github.com/libp2p/go-libp2p-pubsub/blob/master/validation.go#L30) Validator function instead of the Extended version.
> 
> We actually do care about Ignore as well. Currently there is no usecase which simply `Ignore`, but there could be in the future.
> 
> The topic validator should take the validation result from multiple `MessageValidator`s and compute an "aggregate" result which is then returned to Libp2p.



> > Since we only care about `Accept` or `Reject`, we will be better served with [this](https://github.com/libp2p/go-libp2p-pubsub/blob/master/validation.go#L30) Validator function instead of the Extended version.
> 
> We actually do care about Ignore as well. Currently there is no usecase which simply `Ignore`, but there could be in the future.
> 
> The topic validator should take the validation result from multiple `MessageValidator`s and compute an "aggregate" result which is then returned to Libp2p.

correct - and when we do have a use case in the future we can then switch over to the extended version. This seems like a premature generalization. But I went ahead and approved since that is not a big deal one way or the other.

smnzhu:
bors merge

PR #1291, opened 2021-09-14
[Consensus Follower] Make fatal errors recoverable

peterargue:
Implement the ErrorAware interface from https://github.com/onflow/flow-go/pull/1275 for Consensus Follower

The goal is to remove any unrecoverable fatal errors thrown within the consensus follower to ensure anyone using the library can recover and handle the errors instead of exiting the process.

This is accomplished using the `ErrorAware` interface and `ErrorManger` implementation to push errors into an error channel that a parent goroutine can monitor.

The way the FlowNodeBuilder is currently implemented, there's no way to pass errors encountered within the initialization methods, so there's some logic in place to handle this case using a panic/recover flow. Eventually, we'll want to update this to pass the errors back to the caller directly and simplify the handling logic.

### DIFF
--- a/cmd/node_builder.go
+++ b/cmd/node_builder.go
@@ -68,10 +68,18 @@ type NodeBuilder interface {
 	// `Done`.
 	Component(name string, f func(builder NodeBuilder, node *NodeConfig) (module.ReadyDoneAware, error)) NodeBuilder
 
-	// MustNot asserts that the given error must not occur.
-	// If the error is nil, returns a nil log event (which acts as a no-op).
-	// If the error is not nil, returns a fatal log event containing the error.
-	MustNot(err error) *zerolog.Event
+	// Errors returns a channel that receives any unrecoverable errors encountered by the builder
+	Errors() <-chan error
+
+	// ThrowError handles unrecoverable errors
+	// If errorManger is enabled for the builder, the errorManager's ThrowError is called propagate the error
+	// If errorManger is not enabled, the error is handled as a panic
+	ThrowError(err error, msg string)
+
+	// ThrowOnError asserts that the given error must not occur.
+	// If the error is nil, do nothing
+	// If the error is not nil, call ThrowError to initiate fatal error handler
+	ThrowOnError(err error, msg string)
 
 	// Run initiates all common components (logger, database, protocol state etc.)
 	// then starts each component. It also sets up a channel to gracefully shut
@@ -119,6 +127,7 @@ type BaseConfig struct {
 	guaranteesCacheSize   uint
 	receiptsCacheSize     uint
 	db                    *badger.DB
+	errorManager          *module.ErrorManager
 }
 
 // NodeConfig contains all the derived parameters such the NodeID, private keys etc. and initialized instances of
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os"
@@ -279,17 +280,17 @@ func (fnb *FlowNodeBuilder) PrintBuildVersionDetails() {
 
 func (fnb *FlowNodeBuilder) initNodeInfo() {
 	if fnb.BaseConfig.nodeIDHex == NotSet {
-		fnb.Logger.Fatal().Msg("cannot start without node ID")
+		fnb.ThrowError(nil, "cannot start without node ID")
 	}
 
 	nodeID, err := flow.HexStringToIdentifier(fnb.BaseConfig.nodeIDHex)
 	if err != nil {
-		fnb.Logger.Fatal().Err(err).Msgf("could not parse node ID from string: %v", fnb.BaseConfig.nodeIDHex)
+		fnb.ThrowError(err, fmt.Sprintf("could not parse node ID from string: %v", fnb.BaseConfig.nodeIDHex))
 	}
 
 	info, err := loadPrivateNodeInfo(fnb.BaseConfig.BootstrapDir, nodeID)
 	if err != nil {
-		fnb.Logger.Fatal().Err(err).Msg("failed to load private node info")
+		fnb.ThrowError(err, "failed to load private node info")
 	}
 
 	fnb.NodeID = nodeID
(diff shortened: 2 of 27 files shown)


cmd/node_builder.go
```diff
@@ -68,10 +68,18 @@ type NodeBuilder interface {
 	// `Done`.
 	Component(name string, f func(builder NodeBuilder, node *NodeConfig) (module.ReadyDoneAware, error)) NodeBuilder
 
-	// MustNot asserts that the given error must not occur.
-	// If the error is nil, returns a nil log event (which acts as a no-op).
-	// If the error is not nil, returns a fatal log event containing the error.
-	MustNot(err error) *zerolog.Event
+	// Errors returns a channel that receives any unrecoverable errors encountered by the builder
+	Errors() <-chan error
```
smnzhu, line 72: https://www.geeksforgeeks.org/embedding-interfaces-in-golang/

^ You can just embed the `ErrorAware` interface here :)

cmd/scaffold.go
```diff
@@ -682,15 +684,32 @@ func (fnb *FlowNodeBuilder) Module(name string, f func(builder NodeBuilder, node
...
-// If the error is nil, returns a nil log event (which acts as a no-op).
-// If the error is not nil, returns a fatal log event containing the error.
-func (fnb *FlowNodeBuilder) MustNot(err error) *zerolog.Event {
+// Errors returns a channel that receives any unrecoverable errors encountered by the builder
+func (fnb *FlowNodeBuilder) Errors() <-chan error {
+	return fnb.errorManager.Errors()
+}
+
+// handle an unrecoverable error. if an error manager is configured, pass error to the manager, otherwise log a panic
+func (fnb *FlowNodeBuilder) ThrowError(err error, msg string) {
```
smnzhu, line 693: I would suggest keeping the API as simply `ThrowError(err error)`.

Then, instead of `fnb.ThrowError(nil, "message")`, callers can just do `fnb.ThrowError(errors.New("message"))`.

I noticed that the first argument is `nil` for a large percentage of the calls, which suggests that the current API deserves to be modified.

Alternatively, you could split this into two separate methods `Throw(msg string)` and `ThrowError(err error, msg string)`

follower/consensus_follower.go
```diff
@@ -211,10 +220,17 @@ func runAccessNode(ctx context.Context, anb *access.UnstakedAccessNodeBuilder) {
...
+	select {
+	case <-ctx.Done():
+		anb.Logger.Info().Msg("Access node shutting down")
+	case err := <-anb.Errors():
+		cf.errorManager.ThrowError(err)
 	}
 
-	<-ctx.Done()
-	anb.Logger.Info().Msg("Access node shutting down")
 	<-anb.Done()
```
smnzhu, line 234: Technically error could occure during shutdown too, so maybe work wrapping this inside select statement as well, similar to what you did above

cmd/node_builder.go
```diff
@@ -68,10 +68,18 @@ type NodeBuilder interface {
...
+
+	// ThrowError handles unrecoverable errors
+	// If errorManger is enabled for the builder, the errorManager's ThrowError is called propagate the error
+	// If errorManger is not enabled, the error is handled as a panic
+	ThrowError(err error, msg string)
+
+	// ThrowOnError asserts that the given error must not occur.
+	// If the error is nil, do nothing
+	// If the error is not nil, call ThrowError to initiate fatal error handler
+	ThrowOnError(err error, msg string)
```
smnzhu, line 82: I am hesitant to add these methods to the public interface. The interface should mostly define methods that would be relevant to external code which interacts with the `NodeBuilder`. However, these methods are used by the `NodeBuilder` itself, so I think it suffices to just define them on the implementation. No need to add them to the interface

follower/consensus_follower.go
```diff
@@ -146,7 +147,15 @@ func NewConsensusFollower(
...
 	opts ...Option,
-) (*ConsensusFollowerImpl, error) {
+) (consensusFollower *ConsensusFollowerImpl, err error) {
+	// catch any unrecoverable errors encountered during initialization
+	defer func() {
+		if e := recover(); e != nil {
+			consensusFollower = nil
+			err = fmt.Errorf("%v", e)
+		}
+	}()
```
smnzhu, line 157: I would put this inside `buildAccessNode`, so that it is clear which calls may actually result in panic.

You can modify `buildAccessNode` to return an error.

cmd/scaffold.go
```diff
@@ -682,15 +684,32 @@ func (fnb *FlowNodeBuilder) Module(name string, f func(builder NodeBuilder, node
...
+	if fnb.errorManager == nil || !fnb.lm.StartupCommenced() {
```
smnzhu, line 694: Let's avoid checking `lm.StartupCommenced()`, this is a field that is meant to be private, and I also don't really think this is a reliable way to determine whether we should panic or insert into the error channel.

I think we should keep `ThrowError()` as a method that is only called during initialization, which always results in a panic.

If asynchronous code needs to throw a fatal error, it can just call `fnb.errorManager.ThrowError()`, there's not really a need to wrap that with a helper function. Trying to combine the synchronous and asynchronous error throwing code into a single function obfuscates things imo

cmd/scaffold.go
peterargue: I was hesitant to squash the error and message before checking if we should panic since that would change the behaviour for all the nodes that don't run with the error manager. I guess that's going to happen sooner or later when we implement this for the other node types anyway, so it's probably cleaner to just rip the bandaid off now.

cmd/scaffold.go
peterargue: I'm not happy with using the lm variable here either, but I also think requiring developers to know/remember to use one method vs another based on a potentially non-obvious code path will be prone to bugs.

Ultimately the goal is to refactor the initialization logic to return any errors encountered directly so the synchronous case should go away.

I'm not sure if it addresses your concerns, but we could add a `started` variable to the builder struct that's set in the 
`OnStart` closure. That at least avoids exposing the private variable from lifecycle.

smnzhu: > I'm not sure if it addresses your concerns, but we could add a started variable to the builder struct that's set in the OnStart closure. That at least avoids exposing the private variable from lifecycle.

This was what I was actually initially gonna suggest as a workaround.

However, ultimately I still hold the opinion that we should just separate the two methods. As you noted, if we do separate the two methods then eventually `ThrowError()` will no longer exist anyways, because any place where it would have been called will be replaced with the error being returned directly instead. Then, only the calls to `fnb.errorManager.ThrowError()` will remain.

PR #1292, opened 2021-09-14
[consensus V2] BLS aggregator

tarakby:
( This PR is only about implementing `SignatureAggregatorSameMessage` and `PublicKeyAggregator` and their tests. Implementation and tests for `WeightedSignatureAggregator` will follow in a separate PR )

- implement the low level stateful BLS signature aggregator  `SignatureAggregatorSameMessage` : 
   - crypto layer agnostic of hotstuff and everything about the protocol state (weights, identites)
   - participants are identified by indices and BLS public keys
   - not concurrent safe
   - defined under `module/signature` to serve other components of the protocol
   - will be used by `WeightedSignatureAggregator` in Hotstuff
   
- add a low level BLS public key aggregator
  - uses a stateful optimized key aggregation
  - defined under `module/signature` to serve other components of the protocol 
  - concurrent safe
  - not integrated yet in the V2 QC verification 
  
  Not done yet and will follow in a another PR:
   - full implementation of `WeightedSignatureAggregator`
   - tests for `WeightedSignatureAggregator`


consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+	"fmt"
+	"sync"
+
 	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/signature"
 )
 
-// WeightedSignatureAggregator aggregates the signatures
+// WeightedSignatureAggregator implements consensus/hotstuff.WeightedSignatureAggregator
```
durkmurder, line 12:
```suggestion
// WeightedSignatureAggregator implements hotstuff.WeightedSignatureAggregator
```

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
 }
 
-// TotalWeight returns the total weight presented by the collected sig shares.
+// TotalWeight returns the total weight presented by the collected signatures.
+//
+// The function is thread-safe
 func (s *WeightedSignatureAggregator) TotalWeight() uint64 {
-	panic("to be implemented")
+	s.lock.RLock()
+	collectedWeight := s.totalWeight
```
durkmurder, line 112:
```
s.lock.Rlock()
defer s.lock.RUnlock()
return s.totalWeight
```

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+		return collectedWeight, fmt.Errorf("SigneID %s was already added", signerID)
+	}
+
+	err := s.SignatureAggregatorSameMessage.TrustedAdd(index, sig)
+	if err != nil {
+		return collectedWeight, fmt.Errorf("Trusted add has failed: %w", err)
+	}
+
+	s.collectedIDs[signerID] = struct{}{}
+	collectedWeight += weight
```
durkmurder, line 102: You may be observing not latest value. You have no guarantees that `collectedWeight` didn't update prior locking the mutex. 
I would suggest checking next code to:
```
s.totalWeight += weight
```
instead of assigning to `s.totalWeight`

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+// Aggregate aggregates the signatures and returns the aggregated signature.
+//
+// The function is thread-safe.
+// Aggregate attempts to aggregate the internal signatures and returns the resulting signature data.
+// The function performs a final verification and errors if the aggregated signature is not valid. This is
+// required for the function safety since "TrustedAdd" allows adding invalid signatures.
+//
+// TODO : When compacting the list of signers, update the return from []flow.Identifier
+// to a compact bit vector.
+func (s *WeightedSignatureAggregator) Aggregate() ([]flow.Identifier, []byte, error) {
```
durkmurder, line 126: Only one goroutine should run the aggregation, I was imaging that we could cache aggregated signature after it's being built to avoid doing aggregation multiple times, not sure how big of an issue it would be though.

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+
 	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/signature"
 )
 
-// WeightedSignatureAggregator aggregates the signatures
+// WeightedSignatureAggregator implements consensus/hotstuff.WeightedSignatureAggregator
 type WeightedSignatureAggregator struct {
+	*signature.SignatureAggregatorSameMessage                              // low level crypto aggregator, agnostic of weights and flow IDs
```
durkmurder, line 14: How about putting it as member? I find it quite confusing since methods are named exactly the same.

consensus/hotstuff/signature/weighted_signature_aggregator.go
durkmurder: Actually we might be even good without locking at all. Since update happens under mutex we will always observe latest value. In other words it will behave like relaxed atomic. What we won't be able to do it is to wait for lock when some other routing is updating weight, but that might be not needed. @AlexHentschel

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+	indexToSignature map[int]string     // signatures indexed by the signer index
+}
+
+// NewSignatureAggregatorSameMessage returns a new SignatureAggregatorSameMessage structure.
+//
+// A new SignatureAggregatorSameMessage is needed for each set of public keys. If the key set changes,
+// a new structure needs to be instantiated. Participants are defined by their public keys, and are
+// indexed from 0 to n-1 where n is the length of the public key slice.
+// The function errors with engine.InvalidInputError if any input is invalid.
+func NewSignatureAggregatorSameMessage(
```
durkmurder, line 41: Can we document sentinel errors and return values in general?

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+//
+// Aggregate attempts to aggregate the internal signatures and returns the resulting signature.
+// The function performs a final verification and errors if any signature fails the desrialization
+// or if the aggregated signature is not valid. It also errors if no signatures were added.
+// required for the function safety since "TrustedAdd" allows adding invalid signatures.
+// The function is not thread-safe.
+//
+// TODO : When compacting the list of signers, update the return from []int
+// to a compact bit vector.
+func (s *SignatureAggregatorSameMessage) Aggregate() ([]int, crypto.Signature, error) {
```
durkmurder, line 155: We should probably have a dedicated sentinel error for this case, not sure if we want to treat it as exception. If it fails it probably means that we have a bug in implementation, right?

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+	signatures := make([]crypto.Signature, 0, sharesNum)
+	indices := make([]int, 0, sharesNum)
+	for index, sig := range s.indexToSignature {
+		signatures = append(signatures, []byte(sig))
+		indices = append(indices, index)
+	}
+
+	aggregatedSignature, err := crypto.AggregateBLSSignatures(signatures)
+	if err != nil {
+		return nil, nil, errors.New("BLS signature aggregtion failed")
```
durkmurder, line 166: Can we propagate error? I think we should chain it to caller.

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+	}
+	ok, err := s.VerifyAggregate(indices, aggregatedSignature)
+	if err != nil {
+		return nil, nil, errors.New("BLS signature verification failed")
```
durkmurder, line 170: Same here

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+	}
+	if !ok {
+		return nil, nil, errors.New("resulting BLS aggregated signatutre is invalid")
```
durkmurder, line 173: and here

module/signature/aggregation.go
```diff
@@ -3,21 +3,317 @@
...
+	if err != nil {
+		return false, fmt.Errorf("signature verification failed: %w", err)
+	}
+	return ok, nil
+}
+
+// PublicKeyAggregator aggregates BLS public keys in an optimized manner.
+// It uses a greedy algorithm to compute the aggregated key based on the latest
+// computed key and the delta of keys.
+// A caller can use a classic stateless aggrgetaion if the optimization is not needed.
```
durkmurder, line 211:
```suggestion
// A caller can use a classic stateless aggregation if the optimization is not needed.
```

durkmurder (commented):
Thanks for this heavy lift. I have left some comments that I believe needs to be addressed, mostly around `WeightedSignatureAggregator`. 

I would also like to see some unit tests for `WeightedSignatureAggregator`.

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: Sorry for the confusing PR, I actually haven't worked on `WeightedSignatureAggregator` yet, this PR is only for `SignatureAggregatorSameMessage`. I am working on `WeightedSignatureAggregator` in another PR. I just updated the description of the current PR (I'll still take your suggestions into account)

consensus/hotstuff/signature/weighted_signature_aggregator.go
zhangchiqing: agree with @durkmurder .

module/signer.go
```diff
@@ -4,13 +4,15 @@ import (
 	"github.com/onflow/flow-go/crypto"
 )
 
+// TODO : to delete in V2
```
zhangchiqing, line 7: I think we still need the `Signer`, the consensus follower is able to verify and sign a block in a contextless manner
```suggestion
```

module/verifier.go
```diff
@@ -4,11 +4,13 @@ import (
 	"github.com/onflow/flow-go/crypto"
 )
 
+// TODO : to delete in V2
```
zhangchiqing, line 7: I think we still need the Verifier, the consensus follower is able to verify and sign a block in a contextless manner
```suggestion
```

zhangchiqing (approved):
Looks good

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: As mentioned in the PR description, I didn't target `WeightedSignatureAggregator` in this PR. I [commented](https://github.com/onflow/flow-go/pull/1292/commits/43d934cd788215d3075fe9971c3c69b8f8277a0b) the code so we can focus on `SignatureAggregatorSameMessage` here. Sorry for the confusion. (I'll still consider your comments for the next PR)

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: Same [comment](https://github.com/onflow/flow-go/pull/1292#discussion_r708493745)

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: Same [comment](https://github.com/onflow/flow-go/pull/1292#discussion_r708493745)

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: And yes, I can make `SignatureAggregatorSameMessage` cache the signature.

module/signature/aggregation.go
tarakby: Can you please clarify how I should document the errors ? I'm happy to update the doc.

module/signature/aggregation.go
tarakby: Currently, the function errors if:
- there is an unexpected low level error (memory for instance)
- the caller made `TrustedAdd` with an invalid signature - this would be a bug in Hotstuff, and it's not recoverable.

module/signature/aggregation.go
tarakby: yep, good point

tarakby:
@durkmurder 
 > I have left some comments that I believe needs to be addressed, mostly around WeightedSignatureAggregator.
I would also like to see some unit tests for WeightedSignatureAggregator.

Yep I totally agree. Here is the [reason](https://github.com/onflow/flow-go/pull/1292#discussion_r708493745)

consensus/hotstuff/signature/packager.go
```diff
@@ -1,26 +0,0 @@
-package signature
-
-import (
-	"github.com/onflow/flow-go/consensus/hotstuff"
-	"github.com/onflow/flow-go/model/flow"
-)
-
-type ConsensusSigPackerImpl struct {
```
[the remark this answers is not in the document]
tarakby: Ah I understand, I'll put `consensus` back

module/signature/aggregation.go
durkmurder: Exactly the same way you did in other places. Just write a godoc what errors are expected, unexpected and when expected errors can be seen.

module/signature/aggregation.go
durkmurder: So everything is an exception, sounds good to me.

durkmurder (approved):
Happy to approve it if my other comments will be addressed as part of `WeightedSignatureAggregator` PR

module/signer.go
tarakby: I mean we don't need that exact structure there

module/verifier.go
tarakby: Yes, I meant we won't need that exact interface.

tarakby:
@leo @durkmurder @AlexHentschel 
I have merged this PR which is all about implementing `SignatureAggregatorSameMessage` and its tests. 

As mentioned in the PR description, another PR should follow to cover:
- full implementation of WeightedSignatureAggregator
- tests for WeightedSignatureAggregator

consensus/hotstuff/signature/weighted_signature_aggregator.go
tarakby: Implemented in https://github.com/onflow/flow-go/pull/1347

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+	signers                                   []flow.Identity              // all possible signers, defining a canonical order
+	idToIndex                                 map[flow.Identifier]int      // map node identifiers to indices
+	idToWeights                               map[flow.Identifier]uint64   // weight of each signer
+	totalWeight                               uint64                       // weight collected
+	lock                                      sync.RWMutex                 // lock for atomic updates
```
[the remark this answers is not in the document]
tarakby: implemented in  https://github.com/onflow/flow-go/pull/1347

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -1,39 +1,141 @@
...
+	"github.com/onflow/flow-go/module/signature"
 )
 
-// WeightedSignatureAggregator aggregates the signatures
+// WeightedSignatureAggregator implements consensus/hotstuff.WeightedSignatureAggregator
 type WeightedSignatureAggregator struct {
+	*signature.SignatureAggregatorSameMessage                              // low level crypto aggregator, agnostic of weights and flow IDs
+	signers                                   []flow.Identity              // all possible signers, defining a canonical order
+	idToIndex                                 map[flow.Identifier]int      // map node identifiers to indices
+	idToWeights                               map[flow.Identifier]uint64   // weight of each signer
```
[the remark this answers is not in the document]
tarakby: implemented in https://github.com/onflow/flow-go/pull/1347

PR #1293, opened 2021-09-14
Return node builder initialization errors instead of throwing fatal errors

peterargue:
This is a first step towards improving unrecoverable error handling in nodes. In the current implementation, initialization errors are either handled by panicking or calling `log.Fatal()`, the latter resulting in a call to `os.Exit`.

To improve on this, this change refactors the node builder `Initialize()` and methods it calls to return errors back to the caller instead of triggering a fatal error/panic. This will allow the caller to override the default behavior by restarting the node or taking some other action.


cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -67,7 +67,7 @@ func (fnb *StakedAccessNodeBuilder) InitIDProviders() {
 	})
 }
 
-func (builder *StakedAccessNodeBuilder) Initialize() cmd.NodeBuilder {
+func (builder *StakedAccessNodeBuilder) Initialize() error {
```
huitseeker, line 70: Yup, the added bonus being that this makes the side-effecting behavior of the initialization explicit.

huitseeker (commented):
This is a really nice, straightforward cleanup, thanks @peterargue !

cmd/collection/main.go
```diff
@@ -90,62 +90,68 @@ func main() {
...
+		// epoch qc contract flags
+		flags.StringVar(&accessAddress, "access-address", "", "the address of an access node")
+		flags.StringVar(&secureAccessNodeID, "secure-access-node-id", "", "the node ID of the secure access GRPC server")
+		flags.BoolVar(&insecureAccessAPI, "insecure-access-api", true, "required if insecure GRPC connection should be used")
+	})
+
+	err = nodeBuilder.Initialize()
+	if err != nil {
+		nodeBuilder.Logger.Fatal().Err(err).Send()
+	}
```
smnzhu, line 152: nit:

```suggestion
	if err = nodeBuilder.Initialize(); err != nil {
		nodeBuilder.Logger.Fatal().Err(err).Send()
	}
```

Simlilar suggestion for all the other places

cmd/scaffold.go
```diff
@@ -790,7 +790,10 @@ func (fnb *FlowNodeBuilder) Initialize() NodeBuilder {
 
 	fnb.ParseAndPrintFlags()
 
-	fnb.extraFlagsValidation()
+	err := fnb.extraFlagsValidation()
+	if err != nil {
```
smnzhu, line 794: same here

cmd/scaffold.go
```diff
@@ -799,12 +802,15 @@ func (fnb *FlowNodeBuilder) Initialize() NodeBuilder {
 
 	if fnb.metricsEnabled {
 		fnb.EnqueueMetricsServerInit()
-		fnb.RegisterBadgerMetrics()
+		err = fnb.RegisterBadgerMetrics()
+		if err != nil {
```
smnzhu, line 806: and here

smnzhu (approved):
This is awesome :)

huitseeker (approved):
Oops, I meant to approve that.

peterargue:
bors merge

PR #1294, opened 2021-09-15
Fix matching engine

zhangchiqing:
This PR fixes a bug in matching engine that we didn't request receipt where we should and caused sealing to halt


AlexHentschel (approved):

Kay-Zee (approved):

jordanschalm (approved):

engine/consensus/matching/core.go
```diff
@@ -318,9 +318,11 @@ HEIGHT_LOOP:
 		// CAUTION: this is not BFT, as the existence of a candidate seal
 		//          does _not_ imply that all parent results are sealable.
 		// TODO: update for full BFT
-		if _, hasCandidateSeal := blocksWithCandidateSeal[blockID]; hasCandidateSeal {
-			continue
-		}
+		// The follow optimization doesn't work, because sealing requires 2 matching receipts.
+		// Sealing would still halt even if there is a candidate seal but no matching receipts.
```
AlexHentschel, line 322: I think we should just remove 
https://github.com/onflow/flow-go/blob/02fa1fc922b1b09dd49b510a1a8cda0010884798/engine/consensus/matching/core.go#L293-L298
and https://github.com/onflow/flow-go/blob/02fa1fc922b1b09dd49b510a1a8cda0010884798/engine/consensus/matching/core.go#L317-L325

We have an analogous performance optimization that is compatible with our current safety-logic of requiring two consistent execution receipts:
https://github.com/onflow/flow-go/blob/02fa1fc922b1b09dd49b510a1a8cda0010884798/engine/consensus/matching/core.go#L332-L340 

I don't think it is worth keeping the commented-out code

zhangchiqing: Good point

PR #1299, opened 2021-09-15
Update relic_build.sh

yagop:
Images are being build with instructions which are not available on most processors (like AVX-512). This produces illegal instruction error.

This MR targets generic x64 processor.

See https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

```
SIGILL: illegal instruction
PC=0x14e66eb m=0 sigcode=2
signal arrived during cgo execution
instruction bytes: 0x62 0xf1 0xfd 0x28 0x6f 0x5 0x4b 0x5 0xaa 0x0 0x66 0x89 0x47 0x28 0x31 0xc0
```

### DIFF
--- a/crypto/relic_build.sh
+++ b/crypto/relic_build.sh
@@ -42,7 +42,8 @@ if [[ "$ARCH" =~ ^(arm64|armv7|armv7s)$ && "${CC_VERSION_STR[0]}" =~ (clang)  ]]
     #  the "-march=native" option is not supported with clang on ARM
     MARCH=""
 else
-    MARCH="-march=native"
+    # Target standard x86-64 CPU. Avoids AVX512 SIGILL: illegal instruction
+    MARCH="-march=x86-64"
 fi
 
 # Set RELIC config for Flow


crypto/relic_build.sh
```diff
@@ -42,7 +42,8 @@ if [[ "$ARCH" =~ ^(arm64|armv7|armv7s)$ && "${CC_VERSION_STR[0]}" =~ (clang)  ]]
     #  the "-march=native" option is not supported with clang on ARM
     MARCH=""
 else
-    MARCH="-march=native"
+    # Target standard x86-64 CPU. Avoids AVX512 SIGILL: illegal instruction
+    MARCH="-march=x86-64"
```
vishalchangrani, line 46: @yagop - Something that came up from the discussion we had about using a non-AVX512 compile- 

> The problem with non-AVX512 is we leave perf on the table for everyone.
> One thing that would be useful is if netgo & Kraken could give us a /proc/cpuinfo, or, alternatively, the very precise spec of the CPUs they use.
> It would let us know if we can seek perf elsewhere (e.g. ADX)

tarakby:
As @vishalchangrani mentioned, the relic build script is a generic script that should take advantage of the optimization options available on the build machine. 
Updating the script to target certain architectures only will give up on important CPU optimizations commonly available in modern machines  (for users of the library in general, and Flow node operators in particular).

To avoid the issue of illegal instructions, possible solutions are:
 - build Flow node software on the same machine where the software will run.
 - handle "cross-compilation" and distribute multiple images with different levels of generic CPU instructions.

tarakby:
Closing this PR in order to keep the repo build script working for all architectures.
 
cc  @vishalchangrani

PR #1303, opened 2021-09-16
Add jitter to DKG start and upon processing phase 1 broadcast messages

jordanschalm:
Introduce random delays prior to expensive operations in the DKG:
* `Start` - ~700ms
* `HandleBroadcastMessage` - ~2.5s / message

[Results Link](https://www.notion.so/dapperlabs/Test-12-150-Node-DKG-Buffered-VerifVector-Processing-23deece06df2408a822648eb2888254b)


cmd/consensus/main.go
```diff
@@ -143,6 +144,8 @@ func main() {
 			flags.StringVar(&accessAddress, "access-address", "", "the address of an access node")
 			flags.StringVar(&secureAccessNodeID, "secure-access-node-id", "", "the node ID of the secure access GRPC server")
 			flags.BoolVar(&insecureAccessAPI, "insecure-access-api", true, "required if insecure GRPC connection should be used")
+			flags.DurationVar(&dkgControllerConfig.BaseStartDelay, "dkg-controller-base-start-delay", dkgmodule.DefaultBaseStartDelay, "used to define the range for jitter prior to DKG start")
```
zhangchiqing, line 147: better to provide an example. For instance, `500ms`, as is the default value

tarakby (commented):
Nice 👌🏼 
I have a comment and a question:
- I know 500 micro sec is a default duration and won't be used in the real networks. Shouldn't we also set mainnet and testnet config somewhere as part of this PR ? 
- I suggest to make the phase duration dependant on the new config, or at least have a sanity check that the phase duration is at least the max delay plus some constant buffer. Can that be added to this PR?

module/dkg/controller.go
```diff
@@ -11,6 +14,34 @@ import (
...
+	// DefaultBaseHandleBroadcastDelay is the default base to use when
+	// introducing random delay to processing EACH DKG broadcast message. See
+	// preHandleBroadcastDelay for details.
+	//
+	// For a 150-node DKG, we observe a cost of ~2.5s per message to process
+	// broadcast messages during phase 1, for a total of ~6m of total CPU time.
+	// We would like to target spreading this cost over a 20 minute period in the
+	// average case.
+	//
+	// 500µs results in per-broadcast delays of max=11.25s, ave=5.625s.
```
zhangchiqing, line 32: What is the 500µs? Or are you talking about 500ms that is used as the DefaultBaseHandleBroadcastDelay
```suggestion
	// 500ms results in per-broadcast delays of max=11.25s, ave=5.625s.
```

module/dkg/controller.go
```diff
@@ -326,3 +380,45 @@ func (c *Controller) phase3() error {
...
+func computePreprocessingDelay(baseDelay time.Duration, dkgSize int) time.Duration {
+	// sanity checks
+	if baseDelay < 0 {
+		baseDelay = 0
+	}
+	if dkgSize < 0 {
+		dkgSize = 0
+	}
+
+	// m=b*n^2
```
zhangchiqing, line 416: Don't quite understand this math. Could you explain ?

jordanschalm: `maximum_delay = base_delay * number_of_dkg_nodes ^ 2`

Then we select the actual delay from `[0, maximum_delay]`

module/dkg/controller.go
jordanschalm: `DefaultBaseHandleBroadcastDelay` is 500 microseconds, not 500 milliseconds ([code](https://github.com/onflow/flow-go/pull/1303/files#diff-0e1050f66811bb0fd32918920021ad123932be01f04472386ea9ba5df2a6056aR36))

jordanschalm:
> at least have a sanity check that the phase duration is at least the max delay plus some constant buffer

The duration is specified in the smart contract config, so it is tricky to add reliable sanity checks. We could for instance add a sanity check to the `finalize` step of bootstrapping, but this applies only to the first epoch following the spork -- after that the DKG phase length values come from the contract.

> 500 micro sec is a default duration and won't be used in the real networks

My intention was actually to use this value across all networks. This is the coefficient of the delay computation, not the actual delay value (ie. 500µs is `b` in `bn^2`. There is reasoning for where this value comes from in [the comments](https://github.com/onflow/flow-go/pull/1303/files#diff-0e1050f66811bb0fd32918920021ad123932be01f04472386ea9ba5df2a6056aR27-R34)

@tarakby

tarakby:
@Jordan 

> My intention was actually to use this value across all networks.
500 micro seconds gives a max delay of 11.25s, which means processing the broadcasts would spread out over 11.25s. To get a max delay of 20mn, we would need a base `b` of 53ms. Did I misunderstand something?

jordanschalm:
@tarakby 

> To get a max delay of 20mn, we would need a base b of 53ms. Did I misunderstand something?

We are introducing this delay before _each_ message. We don't have one delay that is on average 20mn -- we have many delays with in aggregate average to a total delay of 20mn

From the comments:
```
// 500µs results in per-broadcast delays of max=11.25s, ave=5.625s.
// This results in total delays of max=~28m, ave=~14m, and total time to
// process all phase 1 broadcast messages of max=~34m, ave=~20m.
```

module/dkg/controller.go
jordanschalm: What this PR is trying to do is spread out the CPU-intensive operations required in the DKG for each, not to spread out the actual broadcast messages. The broadcast messages all get sent at more or less the same time. Each node will receive `n-1` broadcast messages, then will stagger the processing of them to minimize the impact on the rest of the system.

The reason we use `n^2` is because the cost for a node to process an individual broadcast messages scales exponentially with the number of nodes and we want the aggregate amount of jitter introduced to be proportional to the aggregate amount of time spent processing broadcast messages.

tarakby:
@jordanschalm 

Now I understand, I was thinking the delays are concurrent and are sampled in [0,20mn].

The issue with the sequential small delays is that we lose the uniformity of delays:  consider for instance incoming message number 2. Each node would process it after `delay_0 + processing_time + delay_1 + processing_time`. The delay to process message 2 is basically randomized by `delay_0 + delay_1`. Although `delay_0` and `delay_1` are uniformly sampled,  `delay_0 + delay_1` is far from uniform (you can for instance think of the sum of 2 uniform dice throws and notice the result in not uniform). The bias gets worse after each message from 2 till 149. The bias means we loose the impact of the mitigation and we get more nodes processing messages at the same time. 

My suggestion would be to remove the dependency between the sleeps by running them concurrently (one thread for [each sleep](https://github.com/onflow/flow-go/blob/jordan/5840-dkg-initial-delays/module/dkg/controller.go#L279)). In that case, the delay should be sampled in [0, max_delay] where max_delay is something like 20mn. What do you think?

jordanschalm:
> the delay should be sampled in [0, max_delay] where max_delay is something like 20mn. What do you think?

The reason I'm cautious about that approach is that it's making a stronger assumption about the timing of broadcast messages. Currently, if a few broadcast messages are sent later that expected in phase 1, we will wait at most a few extra seconds before processing those messages. The added delay is unlikely to cause us to _not_ process them before the end of phase 1. If we select a delay in `[0,20m]` for every message, late-arriving messages may "get unlucky" with a large delay and will not be processed before the end of phase 1.

> The delay to process message 2 is basically randomized by delay_0 + delay_1. Although delay_0 and delay_1 are uniformly sampled, delay_0 + delay_1 is far from uniform (you can for instance think of the sum of 2 uniform dice throws and notice the result in not uniform). The bias gets worse after each message from 2 till 149. The bias means we loose the impact of the mitigation and we get more nodes processing messages at the same time.

It is true that the likelihood of processing a **specific** message is not uniformly distributed over the 20m period. But I'm not sure that's the right thing to measure. We don't care whether node A and node B are processing a particular message M at the same time. We care about whether node A and node B are processing _any_ message at the same time, or how many nodes are processing any message at the same time. Are we optimally spreading the number of nodes processing a message at a given time, over the time available during phase 1 using this per-message delay? That I don't know (though probably not!), but it is close enough to mitigate the problem in the benchmarking.

module/dkg/controller_test.go
```diff
@@ -313,3 +322,35 @@ func checkArtifacts(t *testing.T, nodes []*node, totalNodes int) {
...
+		delay = computePreprocessingDelay(time.Millisecond, -1)
+		assert.Equal(t, delay, time.Duration(0))
+		delay = computePreprocessingDelay(-time.Millisecond, 100)
+		assert.Equal(t, delay, time.Duration(0))
+	})
+
+	t.Run("should return different values for same inputs", func(t *testing.T) {
+		d1 := computePreprocessingDelay(time.Second, 100)
+		d2 := computePreprocessingDelay(time.Second, 100)
+		assert.NotEqual(t, d1, d2)
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/1303/commits/3f4bb82e0c72bf4210953c0699b949fa9ad8d4b0

cmd/consensus/main.go
jordanschalm: https://github.com/onflow/flow-go/pull/1303/commits/7ffc1c1ec0fb586a42e6af89d157e69e93c479d2

tarakby (approved):
Approving this PR to unblock the spork, the code having been tested on benchnet and the Grafana logs being checked. 

Two improving items remain and could be implemented before a next spork:
- The current PR implement a quadratic delay in the number of nodes. This is not needed with the current jitter heurstic (additive random delays before processing each broadcast). A linear jitter is enough. If we switch to a linear delay, the constant values would need to be adjusted.
- The current delay heuristic leads to multiple nodes being busy at the same time (IMO). This happens only for successive broadcast messages and it happens because sum of uniform delays is not uniform. Two other possible heuristics have been discussed:
    - use a blocking delay for each message, sampled uniformly in [0,30mn] (or a larger window) for an async sleep (separate thread) before processing a broadcast. 
    - use a random delay before the very first broadcast message, sampled randomly in [0,`d`] where `d` is relatively larger than a message processing time (2,5s). Use a constant delay before processing subsequent messages to free the CPU to work on Hotstuff. Constant delays would keep nodes processing async, though this only works for broadcasts arriving early (messages arriving late would be processed at the same time by all nodes).

tarakby (approved):

module/dkg/controller.go
```diff
@@ -231,13 +291,39 @@ func (c *Controller) doBackgroundWork() {
...
+			state := c.GetState()
+			if state == Phase1 {
+
+				// introduce a large, uniformly sampled delay prior to processing
+				// the first message
+				isFirstMessage := false
+				c.once.Do(func() {
+					isFirstMessage = true
+					delay := c.preHandleFirstBroadcastDelay()
+					c.log.Debug().Msgf("sleeping for %s before processing first phase 1 broadcast message", delay)
```
zhangchiqing, line 309:
```suggestion
					c.log.Info().Msgf("sleeping for %s before processing first phase 1 broadcast message", delay)
```

zhangchiqing (approved):
Looks good

jordanschalm:
bors merge

PR #1304, opened 2021-09-16
[Consensus] Implement startup-time flag

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/5846


cmd/collection/main.go
```diff
@@ -139,13 +141,25 @@ func main() {
...
 			flags.StringVar(&accessAddress, "access-address", "", "the address of an access node")
 			flags.StringVar(&secureAccessNodeID, "secure-access-node-id", "", "the node ID of the secure access GRPC server")
 			flags.BoolVar(&insecureAccessAPI, "insecure-access-api", true, "required if insecure GRPC connection should be used")
 		}).
 		Initialize().
+		ValidateFlags(func() error {
+			if startupTimeString != cmd.NotSet {
+				t, err := time.Parse(time.RFC3339, startupTimeString)
+				if err != nil {
+					return fmt.Errorf("invalid start-time value: %w", err)
```
zhangchiqing, line 157: better to include the startupTimeString in the error

zhangchiqing (commented):
Nice work. 
Had some minor suggestions and better to add a test case for verify

cmd/consensus/main.go
```diff
@@ -576,6 +589,19 @@ func main() {
...
+			opts := []consensus.Option{
+				consensus.WithInitialTimeout(hotstuffTimeout),
+				consensus.WithMinTimeout(hotstuffMinTimeout),
+				consensus.WithVoteAggregationTimeoutFraction(hotstuffTimeoutVoteAggregationFraction),
+				consensus.WithTimeoutIncreaseFactor(hotstuffTimeoutIncreaseFactor),
+				consensus.WithTimeoutDecreaseFactor(hotstuffTimeoutDecreaseFactor),
+				consensus.WithBlockRateDelay(blockRateDelay),
+			}
+
+			if !startupTime.IsZero() {
```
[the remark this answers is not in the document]
smnzhu: if it is in the past, then `time.After()` will return immediately, so everything here should just work.

cmd/collection/main.go
```diff
@@ -139,13 +141,25 @@ func main() {
 				"additional fraction of replica timeout that the primary will wait for votes")
 			flags.DurationVar(&blockRateDelay, "block-rate-delay", 250*time.Millisecond,
 				"the delay to broadcast block proposal in order to control block production rate")
+			flags.StringVar(&startupTimeString, "startup-time", cmd.NotSet,
```
durkmurder, line 144: Did you consider using unix timestamp ? I feel like it's less error phone in regards of timezones.

durkmurder (approved):
LGTM

cmd/collection/main.go
```diff
@@ -139,13 +141,25 @@ func main() {
...
+				"specifies date and time (in ISO 8601 format) when the consensus participant enters the first view (e.g 2006-01-02T15:04:05Z07:00)")
```
zhangchiqing, line 145: I don't think we need this flag for collection cluster. 

Do we? @AlexHentschel

cmd/consensus/main.go
```diff
@@ -143,8 +145,19 @@ func main() {
 			flags.StringVar(&accessAddress, "access-address", "", "the address of an access node")
 			flags.StringVar(&secureAccessNodeID, "secure-access-node-id", "", "the node ID of the secure access GRPC server")
 			flags.BoolVar(&insecureAccessAPI, "insecure-access-api", true, "required if insecure GRPC connection should be used")
+			flags.StringVar(&startupTimeString, "startup-time", cmd.NotSet, "specifies date and time (in ISO 8601 format) when the consensus participant enters the first view")
```
zhangchiqing, line 148: Please also adding the time format in the comment here as well.

Also it's better to mention that if this startup time is a time in the past, then the node will start up right away.

cmd/collection/main.go
zhangchiqing: yeah, I like @durkmurder 's idea of using timestamp. 

The example `2006-01-02T15:04:05Z07:00` here is still confusing to me. Like the `Z07:00` part. 

A timestamp like `1632245321894` is easier.

cmd/collection/main.go
smnzhu: The original issue says we need it for both.

smnzhu:
> yeah, I like @durkmurder 's idea of using timestamp. 
> 
> The example `2006-01-02T15:04:05Z07:00` here is still confusing to me. Like the `Z07:00` part. 
> 
> A timestamp like `1632245321894` is easier.

@zhangchiqing @durkmurder so personally, I think the ISO format makes more sense because it is human readable, and it is also the most commonly used time format on the internet.

The Z0:700 is just the time Zone.

In the end it doesn't matter too much, but I'd like to push back at least a little bit on the idea that timestamps are better. They are easier and make more sense to our engineering brains, but they are more confusing and error prone for users.

cmd/consensus/main.go
smnzhu: Updated doc to reflect that start-time is the data *after which* the first view may be entered. Therefore if the start-time is in the past, then the first view may be entered immediately

cmd/collection/main.go
zhangchiqing: It's better to be specific that only the hotstuff is stopped until the startup time
```suggestion
			flags.StringVar(&startupTimeString, "hotstuff-startup-time", cmd.NotSet,
```

zhangchiqing (approved):
Looks good to me. Better if @AlexHentschel also give a look

smnzhu:
bors merge

AlexHentschel:
looks good.

PR #1306, opened 2021-09-17
[FLIP] Network API change

smnzhu:


flips/network-api.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Potential problems 
+
+The current network layer API was designed with the assumption that all messages sent and received either target or originate from staked Flow nodes. This is why an engine's [`Process`](https://github.com/onflow/flow-go/blob/master/network/engine.go#L28) method accepts a Flow ID identifying the message sender, and outgoing messages [must specify Flow ID(s)](https://github.com/onflow/flow-go/blob/master/network/conduit.go#L62) as targets.
+
+This assumption is no longer true today. The access node, for example, may communicate with multiple (unstaked) consensus followers. It's perceivable that in the future there will be even more cases where communication with unstaked parties may happen (for example, execution nodes talking to DPS).
+
+Currently, a [`Message`](https://github.com/onflow/flow-go/blob/698c77460bc33d1a8ee8a154f7fe4877bc518a02/network/message/message.proto) which is sent over the network contains many unnecessary fields which can be deduced by the receiver of the message. The only exceptions to this are the `Payload` field (which contains the actual message data) and the `TargetIDs` field (which is used by `Multicast`).
+
```
yhassanzadeh13, line 30: Does this mean excluding the `OriginID` from the message? if that is the case, I would have [this concern](https://github.com/onflow/flow-go/pull/1167#discussion_r715231542) about breaking self-certifying feature of messages.

flips/network-api.md
```diff
@@ -0,0 +1,93 @@
...
+The `Multicast` implementation will need to be changed to make direct connections to the target peers instead of sending messages with a `TargetIDs` field via gossip.
+
+### Motivations
+- Having a strict separation between the public and private networks provides better safety by preventing unintended __ of messages between the two networks, and makes it easier to implement mechanisms for message prioritization / rate-limiting on staked nodes which participate in both.
+- Passing `Context`s gives the network layer the ability to cancel the processing of a network message. This can be leveraged to implement [timeouts](https://pkg.go.dev/context#WithTimeout), but may also be useful for other situations. For example, if the network layer becomes aware that a certain peer has become unreachable, it can cancel the processing of any sync requests from that peer.
+- Since existing calls to `Multicast` only target 3 peers, changing the implementation to use direct connections instead of gossip will reduce traffic on the network and make it more efficient.
```
yhassanzadeh13, line 90: Based on benchmarking we have done using pubsub instead of multicast not only does not degrade the performance but also reduces the goroutines and memory of nodes by a gain of at least 15%. `Unicast` is an expensive operation on the underlying network since it requires establishing a TCP connection to the receiver in case one does not exist. Establishing and maintaining such connections seem more expensive than running topic validators. Also, note that the type of connections and their lifecycle are quite different in `libp2p` vs `pubsub`, the former utilizes streams (i.e., TCP-based in our case) and delegates lifecycle management to the developer, while the latter relies on RPC with the lifecycle managed by libp2p. Also going BFT, we should limit the surface as well as number of connections one node can make to another node, especially on the unstaked network.

flips/network-api.md
smnzhu: Please see my response to that comment.

smnzhu: Please see my response to that comment.

flips/network-api.md
smnzhu: Not sure the size of the network the benchmarking was done with, but the message overhead from using pubsub vs direct connections scales with the number of nodes on the network, so the larger the network is the more of an issue this will be.

Imagine a network with 10000 nodes. Under the current implementation, if one of those nodes sends a multicast targeting three other nodes, all 10000 nodes will need to receive and process that message, and 99997 of them will drop the message after it reaches the application level message validator (note we cannot use libp2p topic validators for this), while only 3 of them actually process the message. This seems very unnecessary.

> Also going BFT, we should limit the surface as well as number of connections one node can make to another node, especially on the unstaked network.

This will need to be dealt with anyways, whether or not we choose to change the multicast implementation. As long as there are *any* valid scenarios where `Unicast` is used, we will need to have a mechanism to rate-limit these connections

flips/network-api.md
huitseeker: A couple of things on top of that comment:

- besides the libp2p-based message signing, which is specifically a network-level whole-message signature by the source of the message, we do not have protocol message signing today,
- unless we mandate envelope signatures at the envelope level, which we currently do not do, the messages are not self-certifying through any feature of the protocol,
- we do have legitimate places where we have other signatures for components of the message (e.g. block proposal signatures from the block proposal emitter, or votes for a block within a QC), in general using the staking key. I do not believe that's because the staking keys are more authoritative.

Besides the fact that I fail to understand any guarantee that protocol-level message signing would bring us:
- adding message signatures (and verifying them) _using the networking key_ (on top of what already exists in libp2p) is an order of magnitude more expensive in (each of) CPU & bandwidth resources than any correspondence we may compute between a PeerID and a FlowID.
- adding message signatures (and verifying them) _using the staking key_ is an order of magnitude more expensive in CPU than doing the same thing with the networking key.

flips/network-api.md
huitseeker: @yhassanzadeh13 Are there any artefacts from this benchmark? Where could we read about the data you collected?

flips/network-api.md
huitseeker: Summary of my discussion with @yhassanzadeh13: 
- I think we're on the same page on this, but that the FLIP, as it is, fails to capture a nuance, let me try to explain:
- There is and should remain a space in the application for messages which application-level sender is not the application-level creator (and application-level signer) of the claimed artefact (including, but not limited to votes on a block, when retransmitted by the next views' leader)
- the FLIP should outline that this application-level indication of origination is legitimate, but should be handled within the application's message payload. 
- Indeed the point of the FLIP is that we're not removing an already-used claim to application-level origin.
- The fact of the matter is we're suggesting that in the following:
```
libp2p.msg { From, msg.Payload { OriginID, msg { <application-level protobuf> }, targetID? }, SIgningKey, Signature} 
```
From, for a valid message, is always equal (modulo format conversion) to OriginID. Hence the OriginID information is redundant. 
- But the use case for application-level origination is still valid. So we should also explain that, should we need one, we should (and do) encapsulate in the application data, in it this way:
```
libp2p.msg { From, msg.Payload { msg { <application-level part 1> { OriginID, sub-msg, OriginSignature}, <application-level part 2> }, targetID? }, SigningKey, Signature} 
```

flips/network-api.md
yhassanzadeh13: @huitseeker yes, please find the report and data [here](https://www.notion.so/dapperlabs/Pubsub-vs-Unicast-benchmarking-96e8d14bb9f54849bb0d4a312b0ca848), also the links on the report will lead you to Grafana dashboards containing the experimentation data.

flips/network-api.md
yhassanzadeh13: > Summary of my discussion with @yhassanzadeh13:
> 
> * I think we're on the same page on this, but that the FLIP, as it is, fails to capture a nuance, let me try to explain:
> * There is and should remain a space in the application for messages which application-level sender is not the application-level creator (and application-level signer) of the claimed artefact (including, but not limited to votes on a block, when retransmitted by the next views' leader)
> * the FLIP should outline that this application-level indication of origination is legitimate, but should be handled within the application's message payload.
> * Indeed the point of the FLIP is that we're not removing an already-used claim to application-level origin.
> * The fact of the matter is we're suggesting that in the following:
> 
> ```
> libp2p.msg { From, msg.Payload { OriginID, msg { <application-level protobuf> }, targetID? }, SIgningKey, Signature} 
> ```
> 
> From, for a valid message, is always equal (modulo format conversion) to OriginID. Hence the OriginID information is redundant.
> 
> * But the use case for application-level origination is still valid. So we should also explain that, should we need one, we should (and do) encapsulate in the application data, in it this way:
> 
> ```
> libp2p.msg { From, msg.Payload { msg { <application-level part 1> { OriginID, sub-msg, OriginSignature}, <application-level part 2> }, targetID? }, SigningKey, Signature} 
> ```

Thanks for summarizing the discussions @huitseeker, it perfectly addresses the concern I initially explained.

flips/network-api.md
yhassanzadeh13: > Not sure the size of the network the benchmarking was done with, but the message overhead from using pubsub vs direct connections scales with the number of nodes on the network, so the larger the network is the more of an issue this will be.
> 
> Imagine a network with 10000 nodes. Under the current implementation, if one of those nodes sends a multicast targeting three other nodes, all 10000 nodes will need to receive and process that message, and 99997 of them will drop the message after it reaches the application level message validator (note we cannot use libp2p topic validators for this), while only 3 of them actually process the message. This seems very unnecessary.
> 
> > Also going BFT, we should limit the surface as well as number of connections one node can make to another node, especially on the unstaked network.
> 
> This will need to be dealt with anyways, whether or not we choose to change the multicast implementation. As long as there are _any_ valid scenarios where `Unicast` is used, we will need to have a mechanism to rate-limit these connections

Your argument is valid @smnzhu, and by the way, the experimentation was in the order of ~150 nodes. When both scalability and security are a concern, maybe we need to refactor the solution with the following problem definition:

- Currently, we drive multicast on top of pubsub which causes every node subscribed to the topic to receive the message. However, practically, we multicast to a limited number of nodes ~3, this results in both bandwidth overhead for non-targeted nodes to receive the message and drop it, as well as attacking surface for malicious peers to flood the system. 

- Driving a multicast over independent unicasts seems to resolve the above problem, however, it adds its own cons: [see this](https://github.com/onflow/flow-go/pull/1306#discussion_r715245617) and [this benchmarking](https://www.notion.so/dapperlabs/Pubsub-vs-Unicast-benchmarking-96e8d14bb9f54849bb0d4a312b0ca848). So, my _point is solving the multicast issue through switching to unicast may lead us to bigger performance + security_ problems.

Accordingly, the solutions we can surface are (and you are welcome to extend the list): 
- Going with this FLIP as it is, driving multicast over unicasts knowing the cost and attack surface it adds and postponing the solution to that to a later FLIP.  
- Going with this FLIP with a revision of also rate-limiting the number of unicast connections or alternatively replacing the implementation of unicast from bare metal libp2p TCP transport to some more efficient alternative, e.g., gRPC. Noting that unicast and pubsub implementation in libp2p are completely separate, as well as their connection lifecycle management. So replacing one does not affect the other. 
- Deriving multicast on pubsub with a revision of addressing the scalability and attack surface issue, e.g., through sub-topics.

flips/network-api.md
```diff
@@ -0,0 +1,93 @@
...
+
+In order to avoid making the message ID function deserialize the `Message` to access the `Payload`, we need to remove all other fields from the `Message` protobuf so that the message ID function can simply take the hash of the the pubsub [`Data`](https://github.com/libp2p/go-libp2p-pubsub/blob/0c7092d1f50091ae88407ba93103ac5868da3d0a/pb/rpc.pb.go#L145) field without needing to do any deserialization.
+
+The `Multicast` implementation will need to be changed to make direct connections to the target peers instead of sending messages with a `TargetIDs` field via gossip.
+
+### Motivations
+- Having a strict separation between the public and private networks provides better safety by preventing unintended __ of messages between the two networks, and makes it easier to implement mechanisms for message prioritization / rate-limiting on staked nodes which participate in both.
```
AlexHentschel, line 88: > unintended __ of messages

To me, it is unclear what "__" is

AlexHentschel (commented):
Looks good.

flips/network-api.md
smnzhu: Oops! >_<

flips/network-api.md
```diff
@@ -0,0 +1,93 @@
...
+```golang
+type Message struct {
+  ctx context.Context
+  event interface{}
+}
+```
+
+> While this may seem to break the general rule of not storing `Context`s in structs, storing `Context`s in structs which are being passed like parameters is one of the exceptions to this rule. See [this](https://github.com/golang/go/issues/22602#:~:text=While%20we%27ve%20told,documentation%20and%20examples.) and [this](https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39#:~:text=The%20one%20exception%20to%20not%20storing%20a%20context%20is%20when%20you%20need%20to%20put%20it%20in%20a%20struct%20that%20is%20used%20purely%20as%20a%20message%20that%20is%20passed%20across%20a%20channel.%20This%20is%20shown%20in%20the%20example%20below.). The idea is that `Context`s should not be **stored** but should **flow** through the program, which is what they do in this usecase.
+
+When the message is dequeued, the engine should check the `Context` to see whether the message is still valid before processing it. At this point, we will have two distinct `Context`s in scope:
```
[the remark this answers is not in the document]
smnzhu: This is a great distinction, thanks :)

AlexHentschel (approved):
🎉

flips/network-api.md
huitseeker: @JeffreyDoyle @simonhf 
I'd love to have your input on the above: this FLIP is suggesting a change in the implementation of multicast messages (messages destined for a plurality of recipients) from a pubsub gossip message to several unicast deliveries.

The gist of the benchmarks above is that this change would cost us in terms of resources (connections, memory and goroutines). What do you think of our ability to:
- attempt that change ? Do we have some margin of maneuver, today?
- track the performance changes involved in this change?

More generally, do you have alternative suggestions? @JeffreyDoyle I know you were suggesting QUIC in https://github.com/dapperlabs/flow-go/issues/5708, where QUIC has real multicast delivery. Any direciton we should explore there?

huitseeker (commented):
The FLIP has shapen up wonderfully, there's just the matter of the multicast change where I'd still love to get feedback from @simonhf / @jwinkler2083233

flips/network-api.md
simonhf: @huitseeker thanks for the notification!

> Unicast is an expensive operation on the underlying network since it requires establishing a TCP connection to the receiver in case one does not exist. Establishing and maintaining such connections seem more expensive than running topic validators.

I agree with this comment in general for a naive implementation, however, if the implementation focuses on the following aspects then I believe it is possible to reap the benefits, and make it the least expensive option:

* Aspect: Pre-establish all TCP connections and only re-open them when absolutely necessary.
* Aspect: Do not use / tie one or more goroutines per TCP connection.
* Aspect: Pre-allocate all main data structures associated with sending and receiving via a TCP connection.

In this scenario, the number of goroutines stays the same whether 1000, 10000 or 1M TCP connections. There is no overhead for establishing or closing a TCP connection (however, there will be stragglers due to network issues and bugs) during node operation because all TCP connections are pre-established at startup or before use. If all lower level data structures (e.g. input and output buffers to read packets into) per TCP connection are pre-allocated, then only a fixed bump in RAM usage will be noticed upon node start, with no extra GC necessary and/or dependent upon the number of TCP connections.

In a previous job (around 2008) I worked on a TCP connection socket server daemon which used these techniques with data structures for up to 1M TCP connections pre-created upon daemon startup, and using the Linux kernel's `epoll` mechanism which decouples TCP connections from threads. Although we don't need anywhere near 1M TCP connections, this serves as proof that such a model works and is technically possible. Back then a Linux kernel could handle opening 1M concurrent TCP connections without too much fuss, but it could only establish about 30k new TCP connections per second regardless of RAM and CPUs due to internal kernel locks, etc. I would expect these types of limitations to be much better in the meantime, and numbers like 30k are likely well above numbers that we need to use anyway :-)

But how to implement such a model in Golang and/or libp2p? This would require more research... There seems to be no technical reason why Golang cannot achieve the same result. However, existing packages may have not been designed that way -- or if more flexible -- used that way. Although Golang is capable of having 1M goroutines, just because it's capable, doesn't mean it's a good idea... :-)

jwinkler2083233:
> The FLIP has shapen up wonderfully, there's just the matter of the multicast change where I'd still love to get feedback from @simonhf / @jwinkler2083233

I haven't seen evidence of true multicasting, with multiple intended recipients per packet.  Logically, the difference is with the management of individual connections, then.  If we move to more unicast style pubsub management, then the onus to manage the connection list shifts to other parts of the code.  Ultimately, it gives us more flexibility.
Regarding memory management, resource management, speed of connections, latencies, etc., this libp2p library is what we've been working with, and so we have to make the best of it.  I suspect that, within a year, operational restrictions will mean we have to start to harden the networking layer, and use the native C++ version of libp2p to get finer-grained control.  We're already seeing strain in the grpc connections.

huitseeker (approved):
This LGTM, while highlighting the suggestions of [thread-pooling connections of @simonhf](https://github.com/onflow/flow-go/pull/1306#discussion_r719684916) and the [suggestions of @jwinkler2083233](https://github.com/onflow/flow-go/pull/1306#issuecomment-931621159), to serve as implementation notes.

smnzhu:
bors merge

PR #1308, opened 2021-09-17
[module] Suggestions for #1275

huitseeker:
This shows an `IrrecoverableErrorHandler`, with (separate) integration into `context` and `RunComponent` + `ReadyDoneAware`.

Possibly superfluous:
- type assertions to be able to use a function on Context instances that are actually IrrecoverableSignalerContext,

Possibly missing:
- do we need to demo more in the component. E.g. RunComponent may demo cleanup, etc?


module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,165 @@
...
+// - restart the component (in production) after cleanup
+// - panic (in canary / benchmark)
+// - log in various Error channels and / or send telemetry ...
+// does not do the plumbing of connecting to the channel
+type IrrecoverableHandler interface {
+	// OnIrrecoverable handles the error, and
+	// executes andThen.
+	// despite the restart use case, andThen
+	// is actually a general instance of a continuation
+	OnIrrecoverable(err error, andThen func())
```
peterargue, line 35: When a handler has multiple `IrrecoverableSignaler` enabled child components, it will need to be able to differentiate which one threw the error. Maybe we need to pass back a reference to the component here?

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,165 @@
...
+// Moreover, we could have built with: context.WithCancel(context.WithIrrecoverableSignal(ctx, sig), ...)
+// "downcasting" to context.Context. Yet, we can still type-assert and recover.
+//
+// ThrowIrrecoverable can be a drop-in replacement anywhere we have a context.Context likely
+// to support Irrecoverables. Note: this is not a method
+func ThrowIrrecoverable(ctx context.Context, err error) {
+	signalerAbleContext, ok := ctx.(irrecoverableSignalerCtxt)
+	if ok {
+		signalerAbleContext.signaler.ThrowIrrecoverable(err)
+	}
```
smnzhu, line 82:
```suggestion
	signalerAbleContext, ok := ctx.(IrrecoverableSignalerContext)
	if ok {
		signalerAbleContext.ThrowIrrecoverable(err)
	}
```

module/errorhandler/erroraware.go
huitseeker: > When a handler has multiple IrrecoverableSignaler enabled child components

Ideally you wouldn't: the key to the hierarchical model enforced by passing the `IrrecoverableSignaler` through a context is that every sub-component hides the parent's `IrrecoverableSignaler` and replaces it with its own, handled by its parent.

So: one `IrrecoverableSignaler` per component, handled by your *direct* parent.

> potentially require a cleanup method on the component that can be called by the handler

AFAICT that's the role of `cancel()`

module/errorhandler/erroraware.go
huitseeker: Yep, of course.

module/errorhandler/erroraware.go
smnzhu: Yeah, @huitseeker overall I like the `IrrecoverableSignaler` and `IrrecoverableSignalerContext`, but I'm not a fan of defining `IrrecoverableHandler` as an interface rather than a function. One reason is the one peter mentioned above, but another reason is because it makes the error handling logic less clear since it is farther away from the logic which runs the component.

If the error handling logic is defined as a function, we can see exactly how error handling is performed at the exact same place in the code where `RunComponent` is called, making the code very easily understandable.

On the otherhand, with an interface the actual implementation will live elsewhere, possible even in a different file but at the very least in a separate struct. Furthermore, you might need multiple different implementations for different components, and this quickly gets messy.

The interface only has a single method, I say we just get rid of the interface and use a function instead.

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,165 @@
...
+// later, in RunComponent, the plumbing happens ...
+type Component interface {
+	Startable
+	module.ReadyDoneAware
+}
+type ComponentFactory func() (Component, error)
+
+func RunComponent(parentCtx IrrecoverableSignalerContext, componentFactory ComponentFactory, handler IrrecoverableHandler) {
+
+	component, err := componentFactory()
```
smnzhu, line 109: I'm also not a big fan of having the component only be created once, and started / stopped multiple times in `restart` below. This forces the developer to make sure that the component's internal state is reset to the exact same state after every shutdown, which is very error prone and could lead to bugs.

I think it is cleaner and safer to just instantiate a fresh copy of the component every time, hence why I originally defined this method to take in a `ComponentFactory`as opposed to just a component.

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		parentCtx.ThrowIrrecoverable(err) // failure to generate the component
+	}
```
smnzhu, line 112: Based on my understanding of this code, I don't think `parentCtx.ThrowIrrecoverable` is a good way to handle this error.

Because there is no way to directly handle errors on `IrrecoverableSignaler`, the only way this error eventually gets handled is *by the `IrrecoverableHandler` for the parent component*, which is logic that is owned by the grandparent component. I don't think it makes sense that a failure to start a child component should be something that is handled by the grandparent component. The grandparent component does not even have any way to know which child component failed.

We should either implement a way to directly handle errors on `IrrecoverableSignaler` without using `RunComponent`, or we should directly return the error here.

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,165 @@
...
+
+	restart()
+
+	for {
+		select {
+		case err := <-irrecoverables:
+			// shutdown the component,
+			cancel()
+			// here the handler is programmed with a restart continuation
+			handler.OnIrrecoverable(err, restart)
```
smnzhu, line 153: We need to wait for the component to finish shutting down here.

module/errorhandler/erroraware.go
peterargue: > AFAICT that's the role of cancel()

Do you mean the cancel in RunComponent [here](https://github.com/onflow/flow-go/pull/1308/files#diff-2e3ca680b4c761f36416e96719a5e92052aad1633269f7bf4c1a40189664bc69R151)?
At that point, the component has already called `Goexit()` and its goroutine has exited, so there shouldn't be anything listening for done events on the other end of that cancel.

module/errorhandler/erroraware.go
peterargue: there isn't really any shutting down to do. if we're here, the component has already called `Goexit` and the goroutine is exiting.

module/errorhandler/erroraware.go
huitseeker: At the end of the day, the model I care about is this: a component starts k sub-components. In doing so, it manages k channels which pass the irrecoverables of those k sub-components and manages them accordingly. If any one of these sub-components spawns sub-components, the same pattern applies for irrecoverables.

I'm OK with using a function here, just beware that it may make reusing handler implementations a tad less easy.

> If the error handling logic is defined as a function, we can see exactly how error handling is performed at the exact same place in the code where RunComponent is called, making the code very easily understandable.

A function is reusable, but I think that precise point will make us have "irrecoverable policy" functions that are defined away from their component.

module/errorhandler/erroraware.go
huitseeker: That's fine, yeah, we can do it this way.

module/errorhandler/erroraware.go
huitseeker: Fair enough: I just want to exemplify that this is a failure to start, and so it should go outside of the normal band that deals with irrecoverables through restarts: restarting here would only create the same error.

Hence, I think the function we use as a continuation (`restart`) which has the signature `func ()` should have `func () error.

module/errorhandler/erroraware.go
smnzhu: > At the end of the day, the model I care about is this: a component starts k sub-components. In doing so, it manages k channels which pass the irrecoverables of those k sub-components and manages them accordingly. If any one of these sub-components spawns sub-components, the same pattern applies for irrecoverables.

Yes, we are on the same page about this, and in general I like the suggestions made here. Only getting into the details because we are now at the stage of actually writing the code that will be merged to the codebase.

I think functions give us more flexibility. If we need one that is reusable, we can define it separately. If it doesn't need to be reusable, then we can define the function inline, giving us better code clarity.

module/errorhandler/erroraware.go
huitseeker: Given that we reinstantiate the component on every run, and that you and @peterargue are intent on having a reference to the component in the handler (how?), is this really necessary, or could the handler take care of that?

smnzhu: I don't think we actually need a reference to the component in the handler, because the handler is explicitly tied to the component that it is called with in `RunComponent`. So no, we don't need to have a reference to component in handler, but therefore I do think we should wait for component to shutdown here.

Even if we reinstantiate, there *could* be undesirable side effects of having two instances of the same component running at the same time?

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,166 @@
...
+		runCtx, cancel = context.WithCancel(parentCtx)
+
+		// signaler used for irrecoverables
+		var signalingCtx IrrecoverableSignalerContext
+		irrecoverables = make(chan error)
+		signalingCtx = WithIrrecoverableSignal(runCtx, &IrrecoverableSignaler{irrecoverables})
+
+		if err = component.Start(signalingCtx); err != nil {
+			// failed to start component: this should not trigger a restart
+			parentCtx.ThrowIrrecoverable(err)
```
smnzhu, line 131: This should also be `return err` right?

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,166 @@
...
+	"runtime"
+
+	"github.com/onflow/flow-go/module"
+)
+
+// IrrecoverableSignaler sends the error out
+// little variability on what happens => struct
+type IrrecoverableSignaler struct {
+	irrecoverableErrors chan<- error
+}
```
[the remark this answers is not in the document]
huitseeker: There's a large range of functions or sub-components that start without passing a context, and starting an `IrrecoverableSigner` for them may be more flexible than shoe-horning them on the `Context` model.

module/errorhandler/erroraware.go
huitseeker: > Even if we reinstantiate, there could be undesirable side effects of having two instances of the same component running at the same time?

For sure.

module/errorhandler/erroraware.go
```diff
@@ -0,0 +1,167 @@
...
+		return err // failure to start
+	}
+
+	for {
+		select {
+		case err := <-irrecoverables:
+			// shutdown the component,
+			cancel()
+			// wait until it's doneC
+			<-done
```
smnzhu, line 153: While we are waiting for `done`, its possible the `parentCtx` could be cancelled.

In this case, we should abort waiting for shutdown to complete and exit the function. Otherwise, we will still proceed to the `handler(err, restart)` below even though the parent ctx is cancelled.

I think we just need an additional `select` statement wrapping this `<-done`.

smnzhu (approved):
Thanks @huitseeker, pretty excited about this :)

Just one last comment, otherwise it looks good!

PR #1310, opened 2021-09-20
[Consensus Voting] Tests for `VoteAggregatorV2`

durkmurder:
https://github.com/dapperlabs/flow-go/issues/5862

### Context

This PR implements unit tests for `VoteAggregatorV2`.
Tests are covering both happy and unhappy path of vote and proposal processing using `VoteAggregatorV2`.

### What was done

- Added test suite for `VoteAggregatorV2`
- Updated implementation details of `VoteAggregatorV2`
- Cleanup of members and constructor
- Updated godoc.


consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	time.Sleep(time.Millisecond * 100)
+	s.collectors.AssertNotCalled(s.T(), "GetOrCreateCollector")
+}
+
+// TestAddBlock_ValidProposal tests that happy path processing of valid proposal finishes without errors
+func (s *VoteAggregatorV2TestSuite) TestAddBlock_ValidProposal() {
+	view := uint64(1000)
+	clr := s.prepareMockedCollector(view)
```
zhangchiqing, line 128: I'm not a big fan of mocking intermediate state for testing like this.

To me, it's best to mock the state transition behavior, and create intermediate states from an initial state using the exposes methods to be tested, rather than creating an intermediate state directly.

In this case here, the vote collectors' state are in-memory state, all possible states (having or not having a certain vote collection) are arrived by only making method calls to the Vote aggregator. 

In a happy path, when calling AddBlock, there are two cases:
1) AddBlock is called before any AddVote was called
2) AddBlock is called after some AddVote was called

So to test the `AddBlock` method, we need to create the above two cases. And we know the difference of the two cases are:
1) no vote collector for that view is created
2) a vote collector for that view has been created

However, instead of creating the vote collector directly, I would suggest to use the `AddVote` to trigger the creation of the vote collector. 

To do that, we need to simulate the `VoteCollectors.GetOrCreateVoteCollector`. It's not hard to implement a thread-safe map that return the vote collector if exists, and create if not.

That way we can setup the case 1) without doing anything, and setup the case 2) by calling `AddVote` first, which will trigger the mocked `VoteCollectors.GetOrCreateVoteCollector` to create the vote collector mock. And it's easier to asset the state of the map in the mocked VoteCollectors at the end of the tests.

So the happy cases can be described as:
1) When AddBlock is called, if there is no vote was added before, there should be only one vote collector created in the map, and the vote collector should be called with `ProcessBlock`
2) AddVote for a block, then calling AddBlock with the block, there should be only one vote collector created in the map, and the vote collector should be called with both `AddVote` and `ProcessBlock`.
3) When calling AddBlock concurrently with 10 different blocks with different views, there should be 10 vote collectors create in the map (this is why the thread-safe vote collector implementation is useful here)
4) When calling AddBlock concurrently with 10 different blocks with the view, there should be only one vote collectors create in the map, and 9 of the 10 calls should be return double proposal error.

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,251 @@
...
+	clr.On("RegisterVoteConsumer", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
+		callback := args.Get(0).(hotstuff.VoteConsumer)
+		for _, vote := range consumedVotes {
+			callback(vote)
+		}
+	})
+
+	err := s.aggregator.InvalidBlock(proposal)
+	require.NoError(s.T(), err)
+	s.notifier.AssertExpectations(s.T())
```
zhangchiqing, line 250: can we verify that future votes after the call to `InvalidBlock` will also be reported as invalid vote?

zhangchiqing (commented):
Nice tests 👍

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
durkmurder: @zhangchiqing I don't understand the benefit of implementing mock creation this way. 
With current approach we place a hard requirement on logic of `VoteAggregator`, in other words it must call `GetOrCreateCollector` exactly with parameters we are expecting, otherwise it's a failed test. I don't see any harm in _preparing_ for certain call and setting up expected behavior. Of course we can implement a factory and then check if only one object was created, but it's harder for to mock collectors that way + I don't see any benefits to this approach. 
Sorry, maybe I don't understand something :)

zhangchiqing (approved):
Looks good to me

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+		}, func(view uint64) error {
+			_, ok := s.mockedCollectors[view]
+			if !ok {
+				return voteCollectorsFactoryError
+			}
+			return nil
+		})
+
+	var err error
+
```
AlexHentschel, line 67:
```suggestion
```

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+// prepareMockedCollector prepares a mocked collector and stores it in map, later it will be used
+// to mock behavior of vote collectors.
+func (s *VoteAggregatorV2TestSuite) prepareMockedCollector(view uint64) *mocks.VoteCollector {
+	collector := &mocks.VoteCollector{}
+	collector.On("View").Return(view).Maybe()
+	s.mockedCollectors[view] = collector
+	return collector
+}
+
+// TestAddVote_DeliveryOfQueuedVotes tests if votes are being delivered for processing by queening logic
```
AlexHentschel, line 88: typo
```suggestion
// TestAddVote_DeliveryOfQueuedVotes tests if votes are being delivered for processing by queueing logic
```

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+		require.NoError(s.T(), err)
+	}
+}
+
+// TestAddBlock_ValidProposal tests that happy path processing of valid proposal finishes without errors
+func (s *VoteAggregatorV2TestSuite) TestAddBlock_ValidProposal() {
+	view := uint64(1000)
+	clr := s.prepareMockedCollector(view)
+	proposal := helper.MakeProposal(s.T(), helper.WithBlock(helper.MakeBlock(s.T(), helper.WithBlockView(view))))
+	clr.On("ProcessBlock", proposal).Return(nil)
```
AlexHentschel, line 141:
```suggestion
	clr.On("ProcessBlock", proposal).Return(nil).Once()
```

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
AlexHentschel: > In a happy path, when calling AddBlock, there are two cases:
>  1. AddBlock is called before any AddVote was called
>  2. AddBlock is called after some AddVote was called
> So to test the AddBlock method, we need to create the above two cases. 

To me, it is unclear what benefit we gain from differentiating these two conditions. In my view, differentiating the two cases is completely encapsulated in the `VoteCollectors`, i.e. a different component than the one we test here. 
* When receiving a vote, the `VoteAggregator` only has to forward the vote to the respective collector. (whether or not we previously received the block is completely irrelevant for the  `VoteAggregator`)
* When receiving a block, the `VoteAggregator` only has to forward the block to the respective collector. This is also completely independent of any votes. 

I think for the unit test of the `VoteAggregator`, it is completely fine to mock out `VoteCollectors`. 

Nevertheless, I think there is a valid argument to testing `VoteAggregator` in conjunction with `VoteCollectors`:
* with our unit tests, we verify that `VoteAggregator` and `VoteCollectors` independently adhere to their respective API specification
* In our architecture document, we have mathematical reasoning along the line:
   *  If `VoteAggregator` and `VoteCollectors` independently adhere to their respective API specification, then the combined system will be correct
* However, there is a gap between the code-level API specification and the mathematical API specification of `VoteAggregator` and `VoteCollectors`. We are humans and we make mistakes. So there is value in testing that the full implementations of `VoteAggregator` and `VoteCollectors` _together_ work correctly. 
* To me, this is on the level of an integration test, because we are testing multiple components in conjunction with each other.

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+	err := s.aggregator.InvalidBlock(proposal)
+	require.NoError(s.T(), err)
+
+	clr.On("AddVote", mock.Anything).Run(func(args mock.Arguments) {
+		voteConsumerCallback(args.Get(0).(*model.Vote))
+	}).Return(nil)
+
+	// generate more votes after notifying about invalid block
+	for i := 0; i < votes; i++ {
+		vote := unittest.VoteFixture(unittest.WithVoteView(view),
```
AlexHentschel, line 269: I think it might be worth-while to also add some votes for the honest block, but _without_ adding an expected notifier call.

AlexHentschel (approved):
Thanks. Very clean code and excellent documentation.

While reviewing, I have noticed a potential, minor problem with our API design (see my comments [here](https://github.com/onflow/flow-go/pull/1310#discussion_r722802197) and [here](https://github.com/onflow/flow-go/pull/1310#discussion_r722656485)). 

#### Suggesting to add additional test

I think adding the following tests for `TestVoteAggregatorV2` might be useful:
* We currently don't properly check that `VoteAggregator.PruneUpToView` calls `VoteCollectors.PruneUpToView`. We have some implicit hints in 2 tests: https://github.com/onflow/flow-go/blob/fe7828a50acee5af222e24bf8ce32a44fcb69150/consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go#L116-L118 But here, we only say `VoteCollectors.PruneUpToView` is _maybe_ called (?) Would suggest to add an explicit test.

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
```diff
@@ -43,44 +43,49 @@ func NewVoteAggregatorV2(
...
-		va.unit.Launch(va.queuedVotesProcessingLoop)
-	}
+	va.lm.OnStart(func() {
+		// launch as many workers as we need
+		for i := 0; i < defaultVoteAggregatorWorkers; i++ {
+			va.unit.Launch(va.queuedVotesProcessingLoop)
+		}
 
-	return va.unit.Ready()
+		<-va.unit.Ready()
```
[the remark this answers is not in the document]
durkmurder: just for sake of keeping same pattern as was before. Will remove.

consensus/hotstuff/vote_aggregator_v2.go
```diff
@@ -31,24 +31,25 @@ type VoteAggregatorV2 interface {
...
 
 	// AddBlock notifies the VoteAggregator about a known block so that it can start processing
 	// pending votes whose block was unknown.
 	// It also verifies the proposer vote of a block, and return whether the proposer signature is valid.
-	// It returns:
-	// - nil if the block is valid
-	//  - model.InvalidBlockError if the block is invalid
-	//  - error if there is exception
+	// Expected error returns during normal operations:
+	// * model.InvalidBlockError if the block is invalid
```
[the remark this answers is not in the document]
durkmurder: I've used `mempool.DecreasingPruningHeightError` as I remember we already have a plan to replace all `mempool.DecreasingPruningHeightError` with something else.

durkmurder:
bors merge

durkmurder:
bors retry

PR #1307, opened 2021-09-17
Mainnet quick fix fo address problems of nodes failing to catch up

AlexHentschel:
#### Bug description:

If consensus nodes are far behind and catching up on _finalized_ block, it can happen that they fail to validate a block and therefore get stuck. This is caused by a subtle edge case in the builder logic. See [incident report](https://www.notion.so/dapperlabs/Incident-Name-09-15-2021-e1c979212c54451abdb8200982b26b27) (notion doc) for details.

### Scope
This PR implements _step 1._ described in [#5857](https://github.com/dapperlabs/flow-go/issues/5857). 
* reverted [emergency-fix](https://github.com/onflow/flow-go/pull/1300) that was deployed to recover mainnet
* the [`consensus.Builder`](https://github.com/onflow/flow-go/blob/alex/5857_mainnet_quick-fix/module/builder/consensus/builder.go) logic is patched in a bit more of a hacky way
* refactored builder tests according to account for the updated business logic
* improved error handing in some of the relevant components

PR #1317, opened 2021-09-21
interim version

AlexHentschel:
interim version that enforces a gap (forwards-compatible) but still trusts the block proposer with the verifier assignment (backwards compatible)


zhangchiqing (approved):

huitseeker (approved):
Implicit is that this builds on https://github.com/onflow/flow-go/pull/1307, and basically re-applies 3c580852637531045e557103397441a32a18c65d.

This makes sense from a chain replay / chain forensics PoV.

PR #1323, opened 2021-09-21
make secure grpc access api default for LN/SN client connections

kc1116:
- update CLI for SN/LN to use secure GRPC conn by default

- update integration tests to use insecure access connection


integration/tests/access/access_api_test.go
```diff
@@ -35,7 +35,7 @@ func (suite *AccessSuite) TearDownTest() {
 		defer suite.cancel()
 	}
 	if suite.net != nil {
-		suite.net.Remove()
+		//suite.net.Remove()
```
jordanschalm, line 38: Why is this commented out?

jordanschalm (approved):

integration/tests/common/mvp_test.go
```diff
@@ -124,13 +124,15 @@ func buildMVPNetConfig() testnet.NetworkConfig {
 		testnet.WithLogLevel(zerolog.InfoLevel),
 		// TODO replace these with actual values
 		testnet.WithAdditionalFlag("--access-address=null"),
+		testnet.WithAdditionalFlag("--insecure-access-api=true"),
```
vishalchangrani, line 127: actually, can't we instead make the test actually work with `--insecure-access-api=false` to actually mimic how the production network will run?

vishalchangrani: If it involved some investigation, we can issue it up but this test should be as close to the real world as possible.

kc1116: We can and initially that was the plan but there would be some investigation around orchestrating access node to come up before LN/SN nodes, also need to think about how this impacts total run time of the CI. We should issue it up or maybe have @gomisha look into it since he is already in that world.

integration/tests/access/access_api_test.go
kc1116: PR is a WIP, because running the integration tests locally is resource intensive I am using CI to let me know what test are failing.

integration/tests/common/mvp_test.go
jordanschalm: > investigation around orchestrating access node to come up before LN/SN nodes

I feel like any requirement we have for the AN to come up before the LN/SN would be true regardless of which API we are using. Does using the secure API change this?

> also need to think about how this impacts total run time of the CI

I also don't understand why using the secure API would impact the run time of the tests.

Maybe I'm missing something here.

kc1116: > I feel like any requirement we have for the AN to come up before the LN/SN would be true regardless of which API we are using. Does using the secure API change this? 

Currently I don't think any of the containers depends on or waits for any of the other containers to start up, we do this in localnet but not in this tests (must be verified). In any case the investigation is around how to get the access node ID and set that flag before we bring up LN/SN nodes. This will more than likely be different per test. 

> I also don't understand why using the secure API would impact the run time of the tests.
It depends on what we need to do as far as container orchestration.

kc1116: I am testing this out now

integration/testnet/network.go
```diff
@@ -216,7 +218,7 @@ func (net *FlowNetwork) ConsensusFollowerByID(id flow.Identifier) consensus_foll
 // Otherwise fails the test.
 func (net *FlowNetwork) ContainerByName(name string) *Container {
 	container, exists := net.Containers[name]
-	require.True(net.t, exists)
+	require.True(net.t, exists, fmt.Sprintf("container %s does not exists", name))
```
vishalchangrani, line 221: ❤️

integration/testnet/network.go
```diff
@@ -467,10 +469,33 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
...
 		err = flowNetwork.AddNode(t, bootstrapDir, nodeConf)
 		require.NoError(t, err)
+
+		// if node is of LN/SN role type add additional flags to node container for secure GRPC connection
+		if nodeConf.Role == flow.RoleConsensus || nodeConf.Role == flow.RoleCollection {
+			// ghost containers don't participate in the network skip any SN/LN ghost containers
+			if !nodeConf.Ghost {
+				nodeContainer := flowNetwork.Containers[nodeConf.ContainerName]
+				nodeContainer.addFlag("insecure-access-api", "false")
+				nodeContainer.addFlag("access-address", fmt.Sprintf("%s:9001", accessNode.ContainerName))
```
vishalchangrani, line 495: best to define  `9001` as a constant up top  as `DefaultSecureAccessNodePort`

integration/tests/access/unstaked_node_test.go
```diff
@@ -123,6 +123,9 @@ func (suite *UnstakedAccessSuite) buildNetworkConfig() {
 		testnet.WithLogLevel(zerolog.TraceLevel),
 	)
 
+	acsConfig := testnet.NewNodeConfig(flow.RoleAccess)
```
vishalchangrani, line 126: is there a need for another access node? Can't we just use the `stakedConfig` as the Secure Access node, it too will open port 9001.

integration/testnet/network.go
jordanschalm, line 221:
```suggestion
	require.True(net.t, exists, "container %s does not exists", name)
```

The assertion functions will actually handle formatting

integration/tests/access/access_api_test.go
```diff
@@ -40,9 +40,8 @@ func (suite *AccessSuite) TearDownTest() {
 }
 
 func (suite *AccessSuite) SetupTest() {
-	nodeConfigs := []testnet.NodeConfig{
-		testnet.NewNodeConfig(flow.RoleAccess),
-	}
+	acsConfig := testnet.NewNodeConfig(flow.RoleAccess)
```
jordanschalm, line 43: Minor thing, but the standard shorthands we use for the different roles is AN/LN/SN/VN/EN

jordanschalm (approved):

integration/tests/access/unstaked_node_test.go
kc1116: Yup your right

vishalchangrani (approved):

PR #847, opened 2021-06-16
Extended logging of sealing status

AlexHentschel:
implements https://github.com/dapperlabs/flow-go/issues/5578

### Overview of engineering approach:
* the status of the sealing progress is evaluated when `sealing.Core` receives a new finalized block from the consensus logic: https://github.com/onflow/flow-go/blob/948814db985a8e2c0506aad54b1eee71223705ed/engine/consensus/sealing/core.go#L449
   Here, the logic runs over all verifier assignments in unsealed but finalized block: 
  - checks whether the incorporated result qualifies for emergency sealing
  - re-requests approvals from assigned verifiers that haven't send an approval yet 
* Note: running `ProcessFinalizedBlock` is fully concurrency safe. Hence, there can be multiple goroutines processing finalized blocks of different heights at the same time. Routines _might_ exit earlier when they realize that another routine is already finalizing a block with larger height (-> [code](https://github.com/onflow/flow-go/blob/948814db985a8e2c0506aad54b1eee71223705ed/engine/consensus/sealing/core.go#L461-L463)). But if they don't exit early, they will run iterate over all assignments in unsealed but finalized block and evaluate the above-mentioned condition. Hence, each of these routines will fully observe the sealing status (as of the finalized block it is processing)
  * Each goroutine creates its own **thread-local** `SealingObservation` and feeds sealing information about all assignments it is inspecting into it. 
  * The `SealingObservation` struct internally determine what information it wants to track. 
  * The `SealingObservation` also has access to some storage-layer components to retrieve additional information.
  * When calling `SealingObservation.Complete()` this information compiled and duped to the logger. 

While I haven't implemented detailed unit tests for the logging logic, I tested various scenarios on localnet. 

### Example output 

```

{
  "level": "info",
  "node_role": "consensus",
  "node_id": "68157a3f6fc5ae25f90ea17533d1d7debf5a7ebdf7b4cd4bc6d039da0cab3c53",
  "finalized_block": "fccbebd8189ac90d04e1f65290116dd1ebc7dd36191e24553981779ccbbd0c9c",
  "finalized_block_height": 146,
  "seals_mempool_size": 115,
  "finalized_seal": {
    "executed_block_height": 0,
    "executed_block_id": "e48e13c19b0c1487a9c9be9282674a0c3fdf413b41a2993c807657c9543da9d3",
    "result_final_state": "1bfd5bdea2e88d37bd66e3f679e983b02ffcff4c2b51d91827c1c4a690c5ad5a",
[...]

PR #1324, opened 2021-09-21
[Consensus] Log verifier ids for missing approvals

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5851

This PR improves the sealing observation log to include the verifiers who didn't send approvals for chunks they were assigned.


durkmurder (approved):

yhassanzadeh13 (approved):

AlexHentschel (changes requested):
Hey Leo,
sorry I am too late with my feedback of your PR https://github.com/onflow/flow-go/pull/847.
I am really worried about the “unsealed reason”, because there are more subtle reasons why we might not seal the next result in addition to the ones covered in your PR:
* There is a reason why an individual result is not sealed. But there might be multiple results, each with a different reason why it is not sealed. 
* In addition to the few reasons you cover in your PR, additional reasons for not sealing might be:
  * there is a miss-match between the end state of the sealed result and the start state of the result pending sealing
  * we have only seen a single execution result (but we require results from two execution nodes)
  * results might be incorporated in a different fork, but not yet in the fork which we are extending, so we can’t seal

If you take a look at the logic in the block builder for determining which seals can be included, the logic is quite complex. To accurately describe why a specific result is not sealed, we would probably need to have similar logic in the sealing observation.

To be clear
* The sealing observation gives you all the data you need to diagnose a sealing halt. I have payed very close attention to that when refactoring it. 
* What you need to add as a human is a detailed understanding of the sealing conditions, so you can identify the specific condition that is not satisfied. 

There is a reason why it is hard to diagnose why sealing halted: because the sealing conditions are complex.

Oversimplifying the conditions where sealing halts really worries me:
Imagine we have a mainnet fire because sealing stoped. What would you prefer: a wrong diagnosis leading you to investigate an aspect that is not a problem? No diagnosis, requiring you draw your own conclusion from the data given.

I have very strong opinion about this topic: If we want algorithmic help to diagnose a sealing halt, we need to make sure that the diagnosis is accurate. Otherwise, it is counter-productive. My preference would be to revert the merge and either implement it properly or remove the diagnosis.

engine/consensus/approvals/tracker/tracker.go
```diff
@@ -170,7 +171,24 @@ func (st *SealingObservation) Complete() {
...
+	}
+
+	unsealedHeight := st.latestSealedBlock.Height + 1
+	if len(st.records) == 0 {
+		return fmt.Sprintf("no result found for next unsealed block at height: %v", unsealedHeight)
+	}
+
+	return fmt.Sprintf("unsealed block at height %v has been executed, the result has been incorporated on %v fork(s), "+
+		"but no result has received enough approvals, check the chunks_with_insufficient_approvals field for more details",
+		len(st.records),
```
AlexHentschel, line 190: ⚠️ 
`unsealed block at height %v` is filled with parameter `len(st.records)`

AlexHentschel: ⚠️ 
`result has been incorporated on %v fork(s)` This statement is only valid if there is one result. But we can have execution forks. Tow results might be incorporated in the _same_ fork

AlexHentschel:
this PR has been reverted in favour of https://github.com/onflow/flow-go/pull/1332/files

PR #1325, opened 2021-09-22
Alex/5857 full fix

AlexHentschel:
This implements step 1 of https://github.com/dapperlabs/flow-go/issues/5857:

simple fix for the business logic 
* update [`consensus.Builder`](https://github.com/onflow/flow-go/blob/master/module/builder/consensus/builder.go) to always include a gap
  * add test to verify that updated business always leaves a gap
  * update existing tests that broke due to change in business logic 
* update [`sealValidator`](https://github.com/onflow/flow-go/blob/master/module/validation/seal_validator.go) to enforce the gap
  * add test to verify that verification logic rejects seals without a gap
  * update existing tests that broke due to change in business logic 

#### Context
This PR is an _extension_ of https://github.com/onflow/flow-go/pull/1307. In this PR, we also update the seal-validation logic to reject seals that do not have a gap.


zhangchiqing (approved):

Kay-Zee (approved):

durkmurder (approved):

module/validation/seal_validator.go
```diff
@@ -157,6 +158,18 @@ func (s *sealValidator) Validate(candidate *flow.Block) (*flow.Seal, error) {
 	// For each visited block collect: IncorporatedResults and block ID
 	forkCollector := func(header *flow.Header) error {
 		blockID := header.ID()
+		// TEMPORARY implementation (unpolished)
```
huitseeker, line 161: What's temporary in this fix given that this PR is labeled final, and that it updates traversal logic?

huitseeker (commented):
One thing that's confusing to me is that this indicates "final", mentions it only implements step 1 of https://github.com/dapperlabs/flow-go/issues/5857, yet:
- seems to tick all boxes for step 2 of 5857,
- contains a  "TEMPORARY" all-caps comment. 

I feel there's a way to make this PR more kind to a future eng exploring the status of the remediation we're putting in, here.

Otherwise, the code looks good!

module/validation/seal_validator.go
AlexHentschel: Ideally, I'd like to generalize the fork traversal logic (step 2 in https://github.com/onflow/flow-go/pull/1325). 

Though, you are right in that this PR _completely_ covers the edge case. So it is not a temporary fix (though could be polished more). Removed the "TEMPORARY" designation in https://github.com/onflow/flow-go/pull/1325/commits/6e2d6bbad115de7c6a30064a785180667d3da8ae.

PR #1327, opened 2021-09-22
Example for using irrecoverable and RunComponent

peterargue:
This PR adds a testable example demonstrating how to use the `irrecoverable` module, and the new lifecycle flow implemented by `RunComponent`.

```
go test -timeout 30s github.com/onflow/flow-go/module/irrecoverable -v               
=== RUN   Example
--- PASS: Example (0.04s)
PASS
ok      github.com/onflow/flow-go/module/irrecoverable  0.401s
```


module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+	// Running cleanup 2
+	// An unrecoverable error occurred: Fatal! No restarts
+	// Error returned from RunComponent 1: context canceled
+}
+
+type ExampleComponent struct {
+	id  int
+	err error
+	lm  *lifecycle.LifecycleManager
+}
```
smnzhu, line 85: Nit: Can we create a constructor function for this struct that only takes in the id and error? The lifecycle manager is not a field that needs to be user provided.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+		}()
+	})
+	return nil
+}
+
+func (c *ExampleComponent) shutdownOnCancel(ctx context.Context) {
+	<-ctx.Done()
+
+	c.lm.OnStop(func() {
+		fmt.Printf("Running cleanup %d\n", c.id)
```
smnzhu, line 104: nit:
```suggestion
		fmt.Printf("[Component %d] Running cleanup\n", c.id)
```

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+type ExampleComponent struct {
+	id  int
+	err error
+	lm  *lifecycle.LifecycleManager
+}
+
+func (c *ExampleComponent) Start(ctx irrecoverable.SignalerContext) error {
+	c.lm.OnStart(func() {
+		fmt.Printf("Starting up %d\n", c.id)
```
smnzhu, line 89: nit:
```suggestion
		fmt.Printf("[Component %d] Starting up\n", c.id)
```

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		go c.shutdownOnCancel(ctx)
+		go func() {
+			<-time.After(20 * time.Millisecond)
+			ctx.ThrowError(c.err)
+		}()
```
smnzhu, line 95: Let's add some comments here to explain what this code is doing.

In general, could we add some more comments throughout the example to explain what is happening? 

```suggestion
		go func() {
		        // do some work...
		        select {
			case <-time.Sleep(100 * time.Millisecond):
			        // encounter irrecoverable error
			        ctx.ThrowError(c.err)
			case <-ctx.Done():
			        fmt.Printf("[Component %d] Cancelled by parent\n", c.id)
			}
		}()
```

module/irrecoverable/example_test.go
smnzhu: Maybe we could have a helper function `printMsg(msg)` which automatically prepends the component ID

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err := module.RunComponent(ctx, componentFactory, onError)
+		if err != nil {
+			fmt.Printf("Error returned from RunComponent 1: %v\n", err)
+		}
+	}()
+
+	wg.Wait()
```
smnzhu, line 69: For the sake of this example, it's probably not necessary to start a goroutine here. we can just call `RunComponent` synchronously here.

Of course in the real code, it will likely start in a goroutine, but for this example it's probably simpler without that.

module/irrecoverable/example_test.go
```diff
@@ -0,0 +1,114 @@
...
+		cancel()
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err := module.RunComponent(ctx, componentFactory, onError)
+		if err != nil {
+			fmt.Printf("Error returned from RunComponent 1: %v\n", err)
```
smnzhu, line 65: Just wondering what the 1 here means?

```suggestion
			fmt.Printf("Error returned from RunComponent: %v\n", err)
```

smnzhu (commented):
Looks great, thanks so much!

smnzhu (approved):

module/irrecoverable/example_test.go
peterargue: leftover from a previous version

PR #1330, opened 2021-09-22
fix nil dereference

smnzhu:


admin/command_runner.go
```diff
@@ -266,7 +266,11 @@ func (r *CommandRunner) processLoop(ctx context.Context) {
 
 	for {
 		select {
-		case command := <-r.commandQ:
+		case command, ok := <-r.commandQ:
+			if !ok {
+				continue
```
peterargue, line 271: since the `commandQ` channel is only closed when shutting down, should this return instead? Otherwise, the loop will churn

peterargue (approved):
looks good

jordanschalm (approved):

huitseeker:
Hmm, not what I [expected](https://github.com/onflow/flow-go/pull/1210#pullrequestreview-761034907), but good fix!

PR #1338, opened 2021-09-23
Extract Accounts into an interface for easier testing

m4ksio:
That's for upcoming PR with small fix and tests


fvm/state/accounts.go
```diff
@@ -32,17 +32,38 @@ func keyPublicKey(index uint64) string {
...
+	CheckAccountNotFrozen(address flow.Address) error
+	GetStorageUsed(address flow.Address) (uint64, error)
+	SetValue(address flow.Address, key string, value []byte) error
+	AllocateStorageIndex(address flow.Address) (uint64, error)
+	SetAccountFrozen(address flow.Address, frozen bool) error
+}
+
+var _ Accounts = &AccountsState{}
+
+type AccountsState struct {
```
ramtinms, line 56: Maybe another name like StatefulAccounts

fvm/state/accounts.go
```diff
@@ -32,17 +32,38 @@ func keyPublicKey(index uint64) string {
 	return fmt.Sprintf("public_key_%d", index)
 }
 
-type Accounts struct {
+type Accounts interface {
```
ramtinms, line 35: We actually were planing last year to break this into smaller interfaces like AccountValues, AccountKeys, AccountContracts, AccountMeta, ... and Accounts to be an interface of all other interfaces.

ramtinms (approved):

cmd/util/ledger/migrations/storage_v5.go
```diff
@@ -160,7 +160,7 @@ func (m *StorageFormatV5Migration) Migrate(payloads []ledger.Payload) ([]ledger.
 	return migratedPayloads, nil
 }
 
-func (m StorageFormatV5Migration) getContractsOnlyAccounts(payloads []ledger.Payload) *state.Accounts {
+func (m StorageFormatV5Migration) getContractsOnlyAccounts(payloads []ledger.Payload) *state.AccountsState {
```
janezpodhostnik, line 163: This return type could be `state.Accounts` which would mean `AccountsState` wouldn't need to be exported from the fvm.

janezpodhostnik (approved):

fvm/state/accounts.go
m4ksio: Yep, I like it

PR #1342, opened 2021-09-23
documenting reason for skipping test  `TestConsensus_LeaderForView` (after current epoch)

AlexHentschel:
This PR covers the quarantined test (see Notion doc [Flaky and Skipped Tests - Quarantine](https://www.notion.so/dapperlabs/TestConsensus_LeaderForView-after-current-epoch-41e60a29531b49748e30a40445906be8) for details): https://github.com/onflow/flow-go/blob/c3502d18e55de4a7b9d7bfefbceeca1363d8218f/consensus/hotstuff/committees/consensus_committee_test.go#L141-L147

While the test is skipped for a valid reason (it tests the behaviour of the mature protocol, which is different than the current implementation due to a temporary fallback), _but_ this reasoning is not documented in the code. 

This PR adds a comment explaining the reason for skipping the test. 
* **no algorithmic changes**
* test remains skipped 
* note: there is a [separate test for covering the temporary logic](https://github.com/onflow/flow-go/blob/c3502d18e55de4a7b9d7bfefbceeca1363d8218f/consensus/hotstuff/committees/consensus_committee_test.go#L149-L154)


jordanschalm (approved):
Thank you 🙏

huitseeker (approved):
This may not be within scope of this PR, but is there a way that :
- the presence or absence of EECC
- the presence of absence of a test targeting EECC
... in the source could be maintained in sync using a [build tag](https://pkg.go.dev/go/build#hdr-Build_Constraints), rather than a skipped test?

gomisha:
thank you for clarifying why the test needs to be skipped - much clearer.

gomisha (approved):

PR #1349, opened 2021-09-24
[Networking] libp2p stream compression

yhassanzadeh13:


network/p2p/middleware.go
```diff
@@ -27,14 +27,6 @@ import (
...
 )
 
-type communicationMode int
-
-const (
-	NoOp communicationMode = iota
-	OneToOne
-	OneToK
-)
-
```
yhassanzadeh13, line 37: Cleaning up unused enum.

network/p2p/compressed/compressedStream.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type StreamOptFunc func(*compressedStream)
+
+// WithStreamCompressor is an option function setting the underlying compressor of the stream.
+func WithStreamCompressor(comp flownet.Compressor) StreamOptFunc {
+	return func(stream *compressedStream) {
+		stream.compressor = comp
+	}
+}
```
smnzhu, line 36: I would argue this should be a required parameter, rather than an optional and having gzip be the default compressor.

It is not clear from the name that `NewCompressedStream` returns a gzip compressor, and I think making the compressor a required parameter in the API makes it more clear.

network/p2p/compressed/compressedStream.go
```diff
@@ -0,0 +1,99 @@
...
+func (c *compressedStream) Close() error {
+	c.writeLock.Lock()
+	defer c.writeLock.Unlock()
+
+	if err := c.w.Close(); err != nil {
+		return err
+	}
+	if err := c.Stream.Close(); err != nil {
+		return err
+	}
```
smnzhu, line 96: We should also combine these in a multierr, because we want `c.Stream.Close()` to be run even if `c.w.Close()` fails.

smnzhu (changes requested):
As mentioned on Slack, I think that since this is not an urgent need / we haven't benchmarked / the feature will soon be part of libp2p core library, we can hold off on this for now.

jwinkler2083233 (approved):

smnzhu:
Sounds like the feature may still be testing from libp2p side for a while before it will be added, so other than the couple comments I have above I think it looks good.

vishalchangrani:
after discussion this with Yahya, I am approving this PR. The protocol based compression will be take care as part of a different issue.

PR #1350, opened 2021-09-24
[Execution] add logger to compactor and only log when we do something

ramtinms:
Yeah its the time to update this, it makes it harder to look into execution logs :D


m4ksio (approved):
Nice! Long overdue simple change

huitseeker (approved):
LGTM!

PR #1352, opened 2021-09-25
DO NOT MERGE - commenting out all startup code that reads a machine account

vishalchangrani:
For the consensus node test on Monday, we will not be asking the operators to create a machine account. Hence, I tried to cherry-pick changes over from this PR https://github.com/onflow/flow-go/pull/1288 on the `v0.22` branch.  But things have changed a bit and I had to comment out a bunch of other code.


jordanschalm (approved):
👍  LGTM for this test. 

If you haven't already, could you run this on localnet to confirm the boot-up succeeds. If the nodes start up successfully with this change, and the test doesn't involve epoch transitions, this should be fine.

Issue #1316, opened 2021-09-20
[Lifecycle] Update `ConsensusFollower` and `CommandRunner` to implement `Startable`


smnzhu:
Update `ConsensusFollower` and `CommandRunner` to implement `Startable` with proper error handling once `Startable` has been merged to master.

PR #1355, opened 2021-09-26
[Network] Middleware component

smnzhu:
Refactor middleware to implement the `Component` interface.

Also introduces new `ComponentManager` struct to help implement `Component` interface


peterargue (commented):
this is looking great! could use some comments, particularly in `common.go`

network/p2p/network.go
```diff
@@ -100,18 +102,21 @@ func NewNetwork(
 // Ready returns a channel that will close when the network stack is ready.
 func (n *Network) Ready() <-chan struct{} {
 	n.lifecycleManager.OnStart(func() {
-		err := n.mw.Start(n)
-		if err != nil {
+		module.RunComponent(n.ctx, func() (module.Component, error) {
```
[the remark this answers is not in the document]
smnzhu: > this is looking great! could use some comments, particularly in `common.go`

Added comments

cmd/access/node_builder/unstaked_access_node_builder.go
```diff
@@ -234,7 +232,7 @@ func (anb *UnstakedAccessNodeBuilder) enqueueMiddleware(ctx context.Context) {
 			// for now we use the empty metrics NoopCollector till we have defined the new unstaked network metrics
 			unstakedNetworkMetrics := metrics.NewNoopCollector()
 
-			libP2PFactory, err := anb.initLibP2PFactory(ctx, unstakedNodeID, unstakedNetworkKey)
```
peterargue, line 237: remove `ctx` from `enqueueMiddleware` as well?

module/component/component.go
```diff
@@ -0,0 +1,320 @@
...
+
+	"go.uber.org/atomic"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+)
+
+var ErrMultipleStartup = errors.New("component may only be started once")
```
peterargue, line 16: This duplicates `module.ErrMultipleStartup`.

module/component/component.go
```diff
@@ -0,0 +1,320 @@
...
+		// context used to run the component
+		var runCtx context.Context
+		runCtx, cancel = context.WithCancel(ctx)
+
+		// signaler used for irrecoverables
+		signaler := irrecoverable.NewSignaler()
+		signalingCtx := irrecoverable.WithSignaler(runCtx, signaler)
+		irrecoverableErr = signaler.Error()
+
+		// the component must be started in a separate goroutine in case an irrecoverable error
```
peterargue, line 63: is this still true now that we're explicitly returning an error?

peterargue: i.e. is the convention that implementors MUST NOT throw an irrecoverable during Start and should return an error instead since subroutines would all be started in their own goroutine anyway.

smnzhu:
Does half of https://github.com/onflow/flow-go/issues/1316

module/component/component.go
smnzhu: I left it there just in case someone misuses the interface, but you could argue it might just be better to document more clearly and assume that developers won't make the mistake.

cmd/access/node_builder/unstaked_access_node_builder.go
smnzhu: Great catch, looks like we can remote context entirely :)

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+
+	startup    func(context.Context) error
+	workers    []ComponentWorker
+	components []Component
+}
+
+// Start initiates the ComponentManager. It will first run the startup routine if one
+// was set, and then start all sub-components and launch all worker routines.
+func (c *ComponentManager) Start(parent irrecoverable.SignalerContext) (err error) {
+	if c.started.CAS(false, true) {
```
peterargue, line 207: can you add a comment about what this is doing? I had to look it up. even something simple like "Only start once"

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+				if err := component.Start(startupCtx); err != nil {
+					defer cancel() // cancel startup for all other components
+					return err
+				}
+				return nil
+			})
+		}
+
+		if err = componentGroup.Wait(); err != nil {
+			// TODO: should we wait for c.Done() before returning?
```
peterargue, line 252: I think so. If `Start` returns an error, my expectation would be that I should throw it away and create a new one or panic()

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+		go func() {
+			select {
+			case err := <-signaler.Error():
+				cancel()
+
+				// we propagate the error directly to the parent because a failure
+				// in a critical sub-component is considered irrecoverable
+				parent.Throw(err)
+			case <-ctx.Done():
+				// TODO: should we instead wait for parent.Done() or c.Done()?
```
peterargue, line 235: `c.Done` is probably the safest since it's possible for subcomponents/workers to throw errors before or during the cancellation (outside of the actual cancel handler)

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+				defer c.done.Done()
+				w(startupCtx)
+			}(worker)
+		}
+
+		return
+	}
+
+	return module.ErrMultipleStartup
+}
```
peterargue, line 270: looks great. can you add some more comments throughout this method. there's a lot going on and this seems like one of those components that everyone should be crystal clear what it's doing and why

network/p2p/middleware.go
```diff
@@ -152,6 +154,15 @@ func NewMiddleware(
 		opt(mw)
 	}
 
+	mw.ComponentManager = component.NewComponentManagerBuilder().OnStart(func(ctx context.Context) error {
+		// TODO: refactor to avoid storing ctx altogether
+		mw.ctx = ctx
+		return mw.start(ctx)
+	}).AddWorker(func(ctx irrecoverable.SignalerContext) {
+		<-ctx.Done()
+		mw.stop()
```
peterargue, line 163: This seems like something most/all implementors would need. Should that be more explicitly exposed in the API?

peterargue: not necessarily the stop method, but the context driven shutdown

network/p2p/network.go
```diff
@@ -132,21 +172,23 @@ func (n *Network) Register(channel network.Channel, engine network.Engine) (netw
...
-	// create the conduit
-	conduit := &Conduit{
-		ctx:       ctx,
-		cancel:    cancel,
-		channel:   channel,
-		publish:   n.publish,
-		unicast:   n.unicast,
-		multicast: n.multicast,
-		close:     n.unregister,
+	select {
```
peterargue, line 177: can you add some comments describing the interactions between register and the worker function added in Start. It's not obvious what's going on to someone that's unfamiliar with the code what.

Is that worker intended to stay alive forever? Is register something that's called many times?

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+		ready:       make(chan struct{}),
+		startup:     c.startup,
+		workers:     c.workers,
+		components:  c.components,
+	}
+}
+
+var _ Component = (*ComponentManager)(nil)
+
+// ComponentManager is used to manage worker routines of a Component
```
peterargue, line 191: can you provide some more details about the intended usecase for this here? also about the distinction between a worker and a component.

module/component/component.go
```diff
@@ -0,0 +1,309 @@
...
+	// immediately.
+	OnStart(ComponentStartup) ComponentManagerBuilder
+
+	// AddWorker adds a worker routine for the ComponentManager
+	AddWorker(ComponentWorker) ComponentManagerBuilder
+
+	// AddComponent adds a new sub-component for the ComponentManager.
+	// This should be used for critical sub-components whose failure would be
+	// considered irrecoverable. For non-critical sub-components, consider using
+	// RunComponent instead.
```
peterargue, line 136: I don't think I get the distinction here. Do you just mean that using `RunComponent` gives you more control over how to handle failures by accepting an error handler, instead of forcing you to implement that explicitly?

network/p2p/middleware.go
smnzhu: This is an anti-pattern, and I'm only doing it this way here because it would be too much work at the moment to completely refactor Middleware.

The idea with using contexts, is that there should never be any need for a "context driven shutdown". Any goroutines / processes which are started when the component is initiated should be passed in the given context to begin with, so they can check themselves whether it is time to shut down. They should not need to be told by a central authority when it is time to stop. This is the philosophy behind using contexts.

network/p2p/network.go
smnzhu: The difference is that the process which responds to register requests uses the context passed to Start(), meaning it will terminate when the context is cancelled.

This is the desired behavior, because Register() calls made after the Network has been cancelled should fail. Previously this same behavior was implemented by storing a context directly inside the Network struct, which is then canceled when Done() is called. However, storing contexts inside structs is an anti-pattern: https://go.dev/blog/context-and-structs. Furthermore, going forward the Done() method will no longer be used to trigger shutdown of a component, but merely to check when shutdown has completed.

Now, before anyone else has the chance to point it out, I admit that exposing the `ShutdownSignal` API from `ComponentManager` is somewhat of a loop hole to the contexts-in-structs "rule". I'm hoping that eventually it may
be deprecated, but for now having it there may make the transition to the `Component` interface easier. Also, wrapping the `ShutdownSignal` in a `ComponentManager` hopefully makes it more clear what the scope of the shutdown signal applies to, whereas storing a context is more ambiguous because there's no way to know whether the context affects routines outside the scope of the component itself.

module/component/component.go
smnzhu: A component that is suitable to be passed to `RunComponent` is one that is not required to be running for the parent component to function (and hence is okay to restart). If a critical subcomponent fails, the parent component itself is unable to function and therefore also fails.

I think a good way to think about this is with respect to component lifecycles. For example, a component's `Ready` method should never depend on a non-critical subcomponent. If it does, what happens when that non-critical subcomponent fails? Should we go back to non-ready state?

On the other hand, if a subcomponent is required to be running for the parent to function, then the parent's `Ready` method must depend on the subcomponent being ready.

smnzhu: I would think of a critical subcomponent basically as just part of the component itself, except encapsulated in a separate `Component` for code clarity / organization reasons. On the other hand, a non-critical sub-component is like a module that can be added or removed.

huitseeker: I'm not opposed to the idea of a critical component: the whole idea of having a standardized signaler in a child is to allow a variety of recovery behaviors in a parent. Child components which failure entails that of the parent should be part of the scenery. 

However, a couple of things I'm wondering about the current approach:
- complexity : ideally, `AddComponent` would work with the "same rails" (`Signaler` context, etc), which minimizes cognitive load. We could handle irrecoverables from a critical sub-component easily, but it seems we're adding the "`AddComponent`" lane because we want those components to be run at startup, whereas RunComponent leaves it to the caller. 

- are there always going to be critical components, to the point we need to handle this distinct startup pattern, or is this for the `Middleware`?

I'm not saying there's a clear way to do it, but if we manage to bring those critical subcomponents into the fold of parental supervision, there's a potential to make the workflow of Components easier to use.

network/p2p/middleware.go
smnzhu: There a book called Concurrency in Go that I read a while ago that talked a lot about this, but I don't remember the exact section that talks about this.

In this specific scenario, ideally any shutdown logic should be setup inside the call to `mw.start`, since that is already being passed in a context.

module/component/component.go
smnzhu: I see many examples across the codebase where one component's `Ready` method is simply waiting for another component's `Ready`. 

I think it's possible we could simplify the API for component manager by just having a single `AddWorker` API, which takes in a function it can use to signal that it is ready. So the API would look something like this:

```golang
type ComponentWorker func(irrecoverable.SignalerContext, func())

type ComponentManagerBuilder interface {
    AddWorker(ComponentWorker)
}
```

And then the worker routine can contain any logic we want, including calling `RunComponent` to run a component with an error handler. Eg:

```golang
func(ctx irrecoverable.SignalerContext, ready func()) {
    // run some startup logic
    ready()
    
    // do work here, and shutdown when ctx is canceled
    // "work" could include running a subcomponent via RunComponent
}
```

@huitseeker thoughts?

huitseeker: Yep, and it matches the continuation-passing approach you could get in `RunComponent` variants. I'm all for it.

cmd/scaffold.go
```diff
@@ -999,9 +1002,23 @@ func (fnb *FlowNodeBuilder) Ready() <-chan struct{} {
 			fnb.handleModule(f)
 		}
 
+		ctx, cancel := context.WithCancel(context.TODO())
+		fnb.Cancel = cancel
+		signaler := irrecoverable.NewSignaler()
+		signalerCtx := irrecoverable.WithSignaler(ctx, signaler)
+
+		// TODO: implement proper error handling
```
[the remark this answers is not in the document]
smnzhu: Yes, of course we should update this eventually. I just don't think it's within the scope of this PR.

https://github.com/dapperlabs/flow-go/issues/5906

module/irrecoverable/irrecoverable.go
```diff
@@ -6,30 +6,43 @@ import (
...
+	defer runtime.Goexit()
+
+	if s.errThrown.CAS(false, true) {
+		s.errChan <- err
+		close(s.errChan)
+	} else {
+		// TODO: we simply log the unhandled irrecoverable for now, but we should probably
+		// find a way to actually handle them.
+		log.New(os.Stderr, "", log.LstdFlags).Println(fmt.Errorf("unhandled irrecoverable: %w", err))
+	}
```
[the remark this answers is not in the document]
smnzhu: I'm not against this, and in the very first version of the error handling PR I had implemented things under this assumption. The reason I made this chance is actually because I realized there was an inconsistency with how we ended up implementing irrecoverable API and RunComponent.

If we choose to go with this approach, it means we should remove [these lines](https://github.com/onflow/flow-go/blob/smnzhu%2Fmiddleware-component/module/component/component.go#L79-L83) for consistency.

The way it was previously implemented, RunComponent shuts down the component upon the first irrecoverable. If we decide we want to keep the channel open, then RunComponent should not automatically shut down the component, and should delegate that responsibility to the error handler. If the error handler does not cancel the component, we should keep listening for irrecoverables on the channel until it does.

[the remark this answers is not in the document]
smnzhu: But the thing is, I don't think this approach is compatible with having the error handler being defined as a function which takes in a single error, rather than a function which takes in an error channel and a restart function, which is what I originally had implemented.

Here are some edge cases:
* If the handler restarts a component, should we then ignore the remaining irrecoverables in the channel? I would say probably not, because the handler may still want to do things with those irrecoverables, such as logging them somewhere. However, if the handler is defined as a single function, then how does it know whether an irrecoverable is from the current component instance or is one leftover from a previous run? 
* If the handler cancels the context, signaling us not to restart the component, does this also mean we should drop the remaining irrecoverables? I still think the answer is no here.

I think we can either: 
* Declare that a signaler will only propagate the first error it receives. This would be compatible with the current implementation of RunComponent and ComponentManager. I think that this is a valid way to define the API, since an irrecoverable error should be an error that is severe enough that it warrants crashing the entire component. After all, the current behaviour across the codebase calls `log.Fatal`, which crashes the entire program.
* Let the signaler continue propagating errors to the handler indefinitely (until the handler decides it stops wanting to receive them or until there are no more to receive because the component has shut down). This would mean that the error handler API should be modified to take in an error channel and a restart function (which is idempotent and only triggers restart once), so that it is clear that all the errors within the channel are within the scope of a single instance of the component. This is also a valid way to define the API.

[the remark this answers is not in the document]
huitseeker: OK, seen like this, I think I'm fine with the "let it crash" philosophy. Errors are either irrecoverable or they are not. If they are, the most one would want to do with the ones occurring after the first one is log them. 

As a consequence, the current setup is probably fine, and much simpler than the alternative. But this "kill on first irrecoverable" particularity should not surprise users, and probably deserves a mention in the FLIP

AlexHentschel (commented):
while reviewing, I noticed the addition of this empty file: https://github.com/onflow/flow-go/blob/8950da93264485fe5fcf51413d921d658e6c0db3/failures
Should we add that to git `.ignore` ?

smnzhu:
> while reviewing, I noticed the addition of this empty file: https://github.com/onflow/flow-go/blob/8950da93264485fe5fcf51413d921d658e6c0db3/failures Should we add that to git `.ignore` ?

Woops, I think I can remove this

network/p2p/middleware.go
peterargue: fair enough. perhaps add a comment here to notify anyone that's using this as an example implementation

PR #1356, opened 2021-09-26
[Network] Add bitswap API

smnzhu:
Introduces new `BlockExchange` API, which is an abstraction around Bitswap.

This will be for state sync engine, among other things.

### TODO
- [x] Add tests


huitseeker (commented):
Do we want to right away constrain the the payload on engine/execution/computation/computer/uploader's `BlockData` ( a struct that should probably live somewhere else in the hierarchy ), instead of reusing bitswap's "block"? In other terms, should we define an interface that hides the go-ipfs "Block" from a codebase that already has semantics for that word?

Another core TODO seems to be mapping those flow structures to CIDs.

smnzhu:
> Do we want to right away constrain the the payload on engine/execution/computation/computer/uploader's `BlockData` ( a struct that should probably live somewhere else in the hierarchy ), instead of reusing bitswap's "block"? In other terms, should we define an interface that hides the go-ipfs "Block" from a codebase that already has semantics for that word?
> 
> Another core TODO seems to be mapping those flow structures to CIDs.

That depends on whether we foresee other usecases in the future for bitswap. I think there could be, and so the networking API should remain as general as possible. If you think the name Block is confusing, I can implement a wrapper interface around that instead and name it something else.

As long as there is the possibility that the state sync engine will not be the only usecase for bitswap, then I think all the application-specific logic of how to map BlockData to CID, how to sync an entire MerkleDAG and piece it back together into a BlockData, etc. should live in a separate module build on top of the network layer.

See for example https://github.com/onflow/flow-go/issues/1451 and https://github.com/onflow/flow-go/issues/1452

network/test/block_exchange_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(5*time.Second))
+
+		blocksReceived := make(map[cid.Cid]struct{})
+		done, err := bex.GetBlocks(blocksToGet...).ForEach(func(b blocks.Block) {
+			blocksReceived[b.Cid()] = struct{}{}
+		}).Send(ctx)
+		require.NoError(suite.T(), err)
+
+		<-done
```
vishalchangrani, line 91: nit: maybe use the unittest utility to time out if channel does not close within duration

network/test/block_exchange_test.go
```diff
@@ -0,0 +1,181 @@
...
+		for j, blockCid := range suite.blockCids {
+			if j != i {
+				blocksToGet[blockCid] = struct{}{}
+			}
+		}
+
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(5*time.Second))
+
+		var doneChans []<-chan struct{}
+		session := bex.GetSession(ctx)
```
vishalchangrani, line 114: what does a session signify here? Is this is a BlockSwap concept?

vishalchangrani (approved):
Left u a nit and a question but other than that lgtm.

network/test/block_exchange_test.go
smnzhu: Yes, it is a bitswap concept: https://github.com/ipfs/go-bitswap#get-related-blocks-faster-with-sessions

smnzhu:
bors merge

smnzhu:
bors merge

smnzhu:
bors merge

PR #1357, opened 2021-09-27
[Consensus Voting] Combined vote processor factory

durkmurder:
https://github.com/dapperlabs/flow-go/issues/5849

### Context

This PR is focused around two main things:
- factory for creating combined vote processor
- tests for combined vote processor

Pattern which was introduced for factory implementation can be reused for staking vote processor factory for collection clusters.

Among other changes:
- godoc update
- changed error handling for `DecodeSingleSig`
- extended test kit
- `VoteAggregatorV2` uses `LifecycleManager`

NOTE: it's quite possible that test suite will be refactored to work with staking vote processor as well.


consensus/hotstuff/votecollector/factory.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	return processor, nil
+}
+
+// Create creates CombinedVoteProcessor for processing votes for current proposal.
+// Caller must treat all errors as exceptions
+func (f *CombinedBaseVoteProcessorFactory) Create(proposal *model.Proposal) (hotstuff.VerifyingVoteProcessor, error) {
+	allParticipants, err := f.committee.Identities(proposal.Block.BlockID, filter.Any)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving consensus participants: %w", err)
```
zhangchiqing, line 83:
```suggestion
		return nil, fmt.Errorf("error retrieving consensus participants at block %v: %w", proposal.Block.BlockID, err)
```

consensus/hotstuff/votecollector/factory.go
```diff
@@ -0,0 +1,114 @@
...
+var _ hotstuff.VoteProcessorFactory = &VoteProcessorFactory{}
+
+func NewCombinedVoteProcessorFactory(log zerolog.Logger, committee hotstuff.Committee, eventHandler hotstuff.EventHandlerV2) *VoteProcessorFactory {
+	return &VoteProcessorFactory{
+		base: &CombinedBaseVoteProcessorFactory{
+			log:       log,
+			committee: committee,
+			packer:    signature.NewConsensusSigDataPacker(committee),
+			onQCCreated: func(qc *flow.QuorumCertificate) {
+				err := eventHandler.OnQCConstructed(qc)
```
zhangchiqing, line 44: ⚠️  We need to give the QC to event loop, instead of EventHandler. 

Note multiple QC will be created by the `onQCCreated` called concurrently, and EventHandler can't process them concurrently. So we need to give it to EventLoop.

durkmurder: good catch

AlexHentschel (commented):
Some comments. Still need to continue the review. 

#### Error handing in `CombinedVoteProcessor`

currently, the `CombinedVoteProcessor` does not properly handle the sentinel errors that the `WeightedSignatureAggregator` returns:
https://github.com/onflow/flow-go/blob/da30152b02bc18f9941463f720f176fb110dafc6/consensus/hotstuff/votecollector/combined_vote_processor.go#L101-L104
Specifically it treats all errors as unexpected. I think the `WeightedSignatureAggregator` has recently been completed, but the `CombinedVoteProcessor` was not updated to properly handle these errors: https://github.com/onflow/flow-go/blob/da30152b02bc18f9941463f720f176fb110dafc6/consensus/hotstuff/signature.go#L71-L72
* `engine.InvalidInputError` should be wrapped as `model.InvalidVoteError`
* `engine.DuplicatedEntryError` is expected during normal operations and therefore should be dropped. 

I feel that we would need a larger cleanup of error handling, so I created a separate issue  https://github.com/dapperlabs/flow-go/issues/5916 .

consensus/hotstuff/votecollector/factory.go
AlexHentschel: 💯 Thanks for spotting this, Leo!

durkmurder: We could add a worker queue for event loop, the same way we do with engines, just with one worker, this way we won't block anyone and keep processing in order. 
Resolution of this comment depends on https://github.com/onflow/flow-go/pull/1437.

durkmurder:
@AlexHentschel I assume that we will do error handling cleanup for `WeightedSignatureAggregator` in other PR and we don't need to do this changes in this PR.

PR #1358, opened 2021-09-27
[Storage Migration V6] Avoid adding magic number

SupunS:
Also, refactor some code

### DIFF
--- a/cmd/util/ledger/migrations/storage_v6.go
+++ b/cmd/util/ledger/migrations/storage_v6.go
@@ -81,13 +81,8 @@ func (e *encodingBaseStorage) Store(id atree.StorageID, value []byte) error {
 	// Add the encoded content to the payloads
 
 	payload := ledger.Payload{
-
 		Key: ledgerKeyFromStorageID(id),
-
-		Value: newInter.PrependMagic(
-			value,
-			newInter.CurrentEncodingVersion,
-		),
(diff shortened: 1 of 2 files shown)


cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -37,12 +38,15 @@ const pathSeparator = "\x1F"
...
 var storageReferenceEncodingStart = []byte{0xd8, cborTagStorageReference}
 
-var storageMigrationV6DecMode = func() cbor.DecMode {
+// maxInt is math.MaxInt32 or math.MaxInt64 depending on arch.
+const maxInt = 1<<(bits.UintSize-1) - 1
+
+var storageMigrationV5DecMode = func() cbor.DecMode {
 	decMode, err := cbor.DecOptions{
 		IntDec:           cbor.IntDecConvertNone,
-		MaxArrayElements: math.MaxInt32,
```
turbolent, line 43: Are we actually running into these limits here?

turbolent (approved):
👍

cmd/util/ledger/migrations/storage_v6.go
SupunS: Not really, I just updated it to be on sync with decode-mode defined in format-V5

PR #1359, opened 2021-09-27
[Crypto] light SHA3 API 

tarakby:
- add light SHA3 API:
   - only supports a single SHA3 computation
   - avoids allocating a hasher to reduce heap allocations.
   - use an input buffer to return the result, which allows callers to control memory allocations. 
   - bench below
   
- integrate the light function in the repo (in particular in `MakeID`)
 
```
BenchmarkComputeHash/SHA3_256-8                  1890801               636 ns/op             416 B/op          2 allocs/op
BenchmarkComputeHash/SHA3_256_light-8            2558676               464 ns/op               0 B/op          0 allocs/op
```


huitseeker (approved):
Thanks a bunch, this makes sense to me! 
I've left a completely optional supplement in 
https://github.com/onflow/flow-go/pull/1370

tarakby:
bors merge

PR #1360, opened 2021-09-27
Enable transaction fees on mainnet

janezpodhostnik:
## Context

Fees will get turned on on mainnet soon. Currently they are both disabled and set to `0.0`. They need to be enabled (with this PR) during a spork. After that they can be set to non `0.0` when desired.

## Checks

I verified that:
- mainnet tx fees are really set to `0.0`
- having tx fees on and set to `0.0` doesn't lead to any unexpected problems (on the emulator)
- having tx fees on and set to `0.0` doesn't emit fee deduction events (see the [contract](https://github.com/onflow/flow-core-contracts/blob/51b4191ee4367dc33e868b91a32e9cdab98fd519/contracts/FlowServiceAccount.cdc#L65))

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -671,11 +671,11 @@ func (fnb *FlowNodeBuilder) initFvmOptions() {
 		fvm.WithChain(fnb.RootChainID.Chain()),
 		fvm.WithBlocks(blockFinder),
 		fvm.WithAccountStorageLimit(true),
+		fvm.WithTransactionFeesEnabled(true),
 	}
 	if fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Canary {
 		vmOpts = append(vmOpts,
 			fvm.WithRestrictedDeployment(false),
-			fvm.WithTransactionFeesEnabled(true),
 		)
 	}
 	fnb.FvmOptions = vmOpts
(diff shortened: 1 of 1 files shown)


Kay-Zee (approved):

cmd/scaffold.go
```diff
@@ -671,11 +671,11 @@ func (fnb *FlowNodeBuilder) initFvmOptions() {
 		fvm.WithChain(fnb.RootChainID.Chain()),
 		fvm.WithBlocks(blockFinder),
 		fvm.WithAccountStorageLimit(true),
+		fvm.WithTransactionFeesEnabled(true),
```
ramtinms, line 674: what is the impact of this on localnet and benchnet, it might break loaders, was wondering if we could still filter it by chainID be mainnet,

Kay-Zee: I think we've used the loaders on canary before no? And i believe fees are enabled there. 

Or did we do extra setup to get the loaders to work for canary?

ramtinms: you mean the marketplace simulator? I was talking about the default loader that we have for localnet, not sure if we have run it against canary before

janezpodhostnik: We ran in on both testnet and canary chain before... do the localnet and benchnet have a separate chain id? It might be better to exclude those then.

ramtinms (approved):

janezpodhostnik:
bors merge

PR #1362, opened 2021-09-27
[Consensus] add logging for sync

smnzhu:
Add logging to help w debugging https://github.com/dapperlabs/flow-go/issues/5874


durkmurder:
Yeeeah, I would actually just crank log level to debug and that's it.

smnzhu:
> This is essentially just changing log levels. Could we instead use the admin server to perform this change?

@huitseeker I considered this too. On one hand, it looks like this should be possible: https://github.com/rs/zerolog#setting-global-log-level

On the other hand, not sure if it's actually possible to do this at runtime: https://github.com/rs/zerolog/issues/252

I'll look into this and investigate further.

smnzhu:
bors merge

PR #1364, opened 2021-09-27
[Admin] simplify admin server implementation

smnzhu:
This simplifies the implementation of admin server to remove the `commandQ`, which reduces the surface area for concurrency bugs.

gRPC will automatically create a new thread for each admin command, and the rate of admin commands will likely not reach the point where having a queue will be necessary.


admin/command_runner.go
```diff
@@ -252,61 +239,29 @@ func (r *CommandRunner) runAdminServer(ctx context.Context) error {
...
-					} else {
-						s, _ := status.FromError(handleErr)
-						err = s.Err()
-					}
-				}
+	if handler := r.getHandler(command); handler != nil {
+		if handleErr := handler(ctx, data); handleErr != nil {
+			if errors.Is(handleErr, context.Canceled) {
+				return status.Error(codes.Canceled, "client canceled")
+			} else if errors.Is(handleErr, context.DeadlineExceeded) {
```
vishalchangrani, line 255: we can actually do without the `if-elseif-else` here since we have a `return` in each of those clauses

vishalchangrani (approved):

smnzhu:
bors merge

smnzhu:
bors retry

smnzhu:
bors merge

smnzhu:
bors merge

PR #1365, opened 2021-09-27
[Network] enable subscription filter except ghost node

smnzhu:
Previously, this was disabled on everything except localnet. 

Instead, we can probably only disable it for ghost nodes.


huitseeker (approved):
See #1210 for context.

@smnzhu thanks!

vishalchangrani (approved):

smnzhu:
bors merge

smnzhu:
bors merge

PR #1371, opened 2021-09-28
[Bootstrap] Extracted `rootblock` and updated `finalize`

durkmurder:
dapperlabs/flow-go#5889
dapperlabs/flow-go#5891

### Context
This PR implements `rootblock` command for bootstrap utility and updates `finalize`.

As result of `rootblock` we output DKG info and root block + votes into separate files.
`finalize` will consume public DKG data, root block and votes to construct QC and initialize collection clusters.

### DIFF
--- a/cmd/bootstrap/cmd/constraints.go
+++ b/cmd/bootstrap/cmd/constraints.go
@@ -9,8 +9,7 @@ import (
 // Checks constraints about the number of partner and internal nodes.
 // Internal nodes must comprise >2/3 of consensus committee.
 // Internal nodes must comprise >2/3 of each collector cluster.
-func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
-
+func checkConsensusConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 	partners := model.ToIdentityList(partnerNodes)
 	internals := model.ToIdentityList(internalNodes)
 	all := append(partners, internals...)
@@ -39,6 +38,14 @@ func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 			partnerCONCount, internalCONCount, partnerCONCount*2+1)
 	}
 
+}
+
+// Checks constraints about the number of partner and internal nodes.
+// Internal nodes must comprise >2/3 of consensus committee.
+// Internal nodes must comprise >2/3 of each collector cluster.
+func checkCollectionConstraints(partnerNodes, internalNodes []model.NodeInfo) {
+	partners := model.ToIdentityList(partnerNodes)
+	internals := model.ToIdentityList(internalNodes)
 	// check collection committee Byzantine threshold for each cluster
 	// for checking Byzantine constraints, the seed doesn't matter
 	_, clusters := constructClusterAssignment(partnerNodes, internalNodes, 0)
--- a/cmd/bootstrap/cmd/finalize.go
+++ b/cmd/bootstrap/cmd/finalize.go
@@ -126,6 +126,9 @@ func finalize(cmd *cobra.Command, args []string) {
 	log.Info().Str("seed", hex.EncodeToString(flagBootstrapRandomSeed)).Msg("deterministic bootstrapping random seed")
 	log.Info().Msg("")
 
+	// TODO: we already have all nodes generated from rootblock step, instead of assembling them again
+	//  we just need to read them. Fix this.
+
 	log.Info().Msg("collecting partner network and staking keys")
 	partnerNodes := assemblePartnerNodes()
 	log.Info().Msg("")
@@ -141,7 +144,6 @@ func finalize(cmd *cobra.Command, args []string) {
 
 	log.Info().Msg("assembling network and staking keys")
 	stakingNodes := mergeNodeInfos(internalNodes, partnerNodes)
-	writeJSON(model.PathNodeInfosPub, model.ToPublicNodeInfoList(stakingNodes))
 	log.Info().Msg("")
 
 	// create flow.IdentityList representation of participant set
@@ -163,6 +165,7 @@ func finalize(cmd *cobra.Command, args []string) {
 	log.Info().Msg("constructing root QC")
 	rootQC := constructRootQC(
 		block,
+		rootBlockData.Votes,
 		model.FilterByRole(internalNodes, flow.RoleConsensus),
 		signer,
 		dkgData,
@@ -185,6 +188,7 @@ func finalize(cmd *cobra.Command, args []string) {
 	// if no root commit is specified, bootstrap an empty execution state
 	if flagRootCommit == "0000000000000000000000000000000000000000000000000000000000000000" {
 		generateEmptyExecutionState(
+			block.Header.ChainID,
 			getRandomSource(flagBootstrapRandomSeed),
 			assignments,
 			clusterQCs,
@@ -499,6 +503,7 @@ func loadRootProtocolSnapshot(path string) (*inmem.Snapshot, error) {
 // generateEmptyExecutionState generates a new empty execution state with the
 // given configuration. Sets the flagRootCommit variable for future reads.
 func generateEmptyExecutionState(
+	chainID flow.ChainID,
 	randomSource []byte,
 	assignments flow.AssignmentList,
 	clusterQCs []*flow.QuorumCertificate,
@@ -546,7 +551,7 @@ func generateEmptyExecutionState(
 	commit, err = run.GenerateExecutionState(
 		filepath.Join(flagOutdir, model.DirnameExecutionState),
 		serviceAccountPublicKey,
-		parseChainID(flagRootChain).Chain(),
+		chainID.Chain(),
 		fvm.WithInitialTokenSupply(cdcInitialTokenSupply),
 		fvm.WithMinimumStorageReservation(fvm.DefaultMinimumStorageReservation),
 		fvm.WithAccountCreationFee(fvm.DefaultAccountCreationFee),
--- a/cmd/bootstrap/cmd/finalize.go
+++ b/cmd/bootstrap/cmd/finalize.go
@@ -136,7 +136,8 @@ func finalize(cmd *cobra.Command, args []string) {
 	log.Info().Msg("")
 
 	log.Info().Msg("checking constraints on consensus/cluster nodes")
-	checkConstraints(partnerNodes, internalNodes)
+	checkConsensusConstraints(partnerNodes, internalNodes)
+	checkCollectionConstraints(partnerNodes, internalNodes)
 	log.Info().Msg("")
 
 	log.Info().Msg("assembling network and staking keys")
--- a/cmd/bootstrap/cmd/finalize.go
+++ b/cmd/bootstrap/cmd/finalize.go
@@ -7,13 +7,14 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/onflow/cadence"
 	"github.com/spf13/cobra"
 
 	"github.com/onflow/flow-go/cmd"
 	"github.com/onflow/flow-go/cmd/bootstrap/run"
+	"github.com/onflow/flow-go/cmd/bootstrap/utils"
+	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/fvm"
 	model "github.com/onflow/flow-go/model/bootstrap"
 	"github.com/onflow/flow-go/model/dkg"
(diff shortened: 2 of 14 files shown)


cmd/bootstrap/cmd/seal.go
```diff
@@ -43,11 +45,16 @@ func constructRootResultAndSeal(
 		RandomSource:       getRandomSource(flagBootstrapRandomSeed),
 	}
 
+	dkgPubKeys := make([]crypto.PublicKey, 0)
+	for _, participant := range dkgData.Participants {
+		dkgPubKeys = append(dkgPubKeys, participant.KeyShare)
+	}
+
```
jordanschalm, line 52: ⚠️  This is iterating over a map (random order) to populate the list of public keys. The order of keys in `DKGParticipantKeys` is meaningful and must be preserved. 

```suggestion
	dkgPubKeys := make([]crypto.PublicKey, len(dkgData.Participants))
	for i, participant := range dkgData.Participants {
		dkgPubKeys[participant.Index] = participant.KeyShare
	}

```

cmd/bootstrap/cmd/finalize.go
```diff
@@ -76,26 +75,27 @@ func addFinalizeCmdFlags() {
 		" in the JSON file: Role, Address, NodeID, NetworkPubKey, StakingPubKey)")
 	finalizeCmd.Flags().StringVar(&flagPartnerStakes, "partner-stakes", "", "path to a JSON file containing "+
 		"a map from partner node's NodeID to their stake")
+	finalizeCmd.Flags().StringVar(&flagDKGPubDataPath, "dkg-data", "", "path to a JSON file containing public data as output from DKG process")
+	finalizeCmd.Flags().StringVar(&flagSignerDKGDataPath, "signer-dkg-data", "",
```
jordanschalm, line 79: I don't think this flag is necessary. All we need to create the QC is the votes.

cmd/bootstrap/cmd/finalize.go
```diff
@@ -76,26 +75,27 @@ func addFinalizeCmdFlags() {
...
 	cmd.MarkFlagRequired(finalizeCmd, "partner-stakes")
+	cmd.MarkFlagRequired(finalizeCmd, "dkg-data")
+	cmd.MarkFlagRequired(finalizeCmd, "signer-dkg-data")
 
 	// required parameters for generation of root block, root execution result and root block seal
-	finalizeCmd.Flags().StringVar(&flagRootChain, "root-chain", "local", "chain ID for the root block (can be 'main', 'test', 'canary', 'bench', or 'local'")
-	finalizeCmd.Flags().StringVar(&flagRootParent, "root-parent", "0000000000000000000000000000000000000000000000000000000000000000", "ID for the parent of the root block")
-	finalizeCmd.Flags().Uint64Var(&flagRootHeight, "root-height", 0, "height of the root block")
-	finalizeCmd.Flags().StringVar(&flagRootTimestamp, "root-timestamp", time.Now().UTC().Format(time.RFC3339), "timestamp of the root block (RFC3339)")
+	finalizeCmd.Flags().StringVar(&flagRootBlock, "root-block", "",
```
jordanschalm, line 90: As mentioned this morning, I think we should separate the block from the votes, and expect that the votes will be coming from several files -- each node operator would generate one vote file.

cmd/bootstrap/cmd/finalize.go
```diff
@@ -492,6 +528,11 @@ func generateEmptyExecutionState(
 		log.Fatal().Err(err).Msg("invalid random source")
 	}
 
+	dkgPubKeys := make([]crypto.PublicKey, 0)
+	for _, participant := range dkg.Participants {
```
jordanschalm, line 531: ⚠️ same issue here - we are losing information about the ordering of keys

cmd/bootstrap/cmd/finalize_test.go
```diff
@@ -53,6 +49,29 @@ const finalizeHappyPathLogs = "^deterministic bootstrapping random seed" +
 
 var finalizeHappyPathRegex = regexp.MustCompile(finalizeHappyPathLogs)
 
+// getFirstQCSignerPath picks first internal node as participant who will do the QC signing
+// node needs to have random beacon key and node info to be treated as internal.
+func getFirstQCSignerPath(t *testing.T, bootDir string) string {
```
jordanschalm, line 54: This is not necessary, because we do not need any actual private keys to produce a QC from votes.

cmd/bootstrap/run/qc.go
```diff
@@ -175,3 +190,34 @@ func GenerateQCParticipantData(allNodes, internalNodes []bootstrap.NodeInfo, dkg
 
 	return qcData, nil
 }
+
+// GenerateQCSignerParticipantData generates participant data for creating QC.
+// Caller needs to pass dkg participants that will be taking part in building QC.
+func GenerateQCSignerParticipantData(internalNodes []bootstrap.NodeInfo, dkgParticipant dkg.DKGParticipantPriv, dkgData inmem.EncodableDKG) (*ParticipantData, error) {
```
jordanschalm, line 196: As mentioned in the other comment, we don't actually need a private key to generate a QC from existing votes. 

I wrote up a function that should do the trick to generate a signer/validator which can be used, without needing to load any private data (which ultimately isn't used anyway):

```go
func createAggregator(participantData *ParticipantData) (hotstuff.Validator, hotstuff.SignerVerifier, error) {

	identities := participantData.Identities()

	var (
		unusedLocalNodeID                   = flow.ZeroID
		unusedModuleLocal module.Local      = nil
		unusedPrivateKey  crypto.PrivateKey = nil
	)

	// create consensus committee's state
	committee, err := committees.NewStaticCommittee(identities, unusedLocalNodeID, participantData.Lookup, participantData.GroupKey)
	if err != nil {
		return nil, nil, err
	}

	merger := signature.NewCombiner(encodable.ConsensusVoteSigLen, encodable.RandomBeaconSigLen)
	stakingSigner := signature.NewAggregationProvider(encoding.ConsensusVoteTag, unusedModuleLocal)
	beaconVerifier := signature.NewThresholdVerifier(encoding.RandomBeaconTag)
	beaconSigner := signature.NewThresholdProvider(encoding.RandomBeaconTag, unusedPrivateKey)
	beaconStore := signature.NewSingleSignerStore(beaconSigner)
	signer := verification.NewCombinedSigner(committee, stakingSigner, beaconVerifier, merger, beaconStore, flow.ZeroID)

	validator := validator.New(committee, new(mocks.ForksReader), signer)

	return validator, signer, nil
}
```

jordanschalm (commented):
Thanks Yurii - looks good, had 2 main comments:

1. We are losing DKG index information when converting between types. This will cause verification to fail. 
2. I added a suggestion to avoid needing to load a private key to aggregate the QC. This is not strictly a blocker, but I think it's worth removing the unneeded flag and logic.

cmd/bootstrap/cmd/constraints.go
```diff
@@ -9,8 +9,7 @@ import (
 // Checks constraints about the number of partner and internal nodes.
 // Internal nodes must comprise >2/3 of consensus committee.
 // Internal nodes must comprise >2/3 of each collector cluster.
-func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
-
+func checkConsensusConstraints(partnerNodes, internalNodes []model.NodeInfo) {
```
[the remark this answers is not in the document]
AlexHentschel: agree about removing https://github.com/onflow/flow-go/blob/fdbd24ca32980e04c5acb1d61aca48f4ca5cdc23/cmd/bootstrap/cmd/constraints.go#L31-L39

But regarding the check https://github.com/onflow/flow-go/blob/fdbd24ca32980e04c5acb1d61aca48f4ca5cdc23/cmd/bootstrap/cmd/constraints.go#L17-L29 I think we still need this for method [`checkCollectionConstraints`](https://github.com/onflow/flow-go/blob/fdbd24ca32980e04c5acb1d61aca48f4ca5cdc23/cmd/bootstrap/cmd/constraints.go#L46). Otherwise, `checkCollectionConstraints` might conclude that we have enough collectors in each cluster, yet, they don't represent a supermajority, because we are ignoring the weight in our check.

cmd/bootstrap/cmd/constraints.go
```diff
@@ -39,6 +38,14 @@ func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 			partnerCONCount, internalCONCount, partnerCONCount*2+1)
 	}
 
+}
+
+// Checks constraints about the number of partner and internal nodes.
+// Internal nodes must comprise >2/3 of consensus committee.
+// Internal nodes must comprise >2/3 of each collector cluster.
+func checkCollectionConstraints(partnerNodes, internalNodes []model.NodeInfo) {
```
AlexHentschel, line 46: ⚠️ 

Please note that also Collector nodes are weighted. Here, we just count nodes. 
* In the previous implementation, this was no problem, because we checked that all nodes have equal stake. 
* Would suggest to move check https://github.com/onflow/flow-go/blob/fdbd24ca32980e04c5acb1d61aca48f4ca5cdc23/cmd/bootstrap/cmd/constraints.go#L17-L29 into this method here

cmd/bootstrap/cmd/finalize.go
```diff
@@ -127,6 +125,9 @@ func finalize(cmd *cobra.Command, args []string) {
 	log.Info().Str("seed", hex.EncodeToString(flagBootstrapRandomSeed)).Msg("deterministic bootstrapping random seed")
 	log.Info().Msg("")
 
+	// TODO: we already have all nodes generated from rootblock step, instead of assembling them again
+	//  we just need to read them. Fix this.
```
AlexHentschel, line 129: The method naming might be a bit misleading here:
* https://github.com/onflow/flow-go/blob/567811b09b688314d33c8cb24040b16f07f5af9e/cmd/bootstrap/cmd/finalize.go#L132
* https://github.com/onflow/flow-go/blob/567811b09b688314d33c8cb24040b16f07f5af9e/cmd/bootstrap/cmd/finalize.go#L136

actually _read_ the data from file and don't re-create it.

cmd/bootstrap/run/qc.go
```diff
@@ -38,24 +38,18 @@ func (pd *ParticipantData) Identities() flow.IdentityList {
 	return bootstrap.ToIdentityList(nodes)
 }
 
-func GenerateRootQC(block *flow.Block, participantData *ParticipantData) (*flow.QuorumCertificate, error) {
-
+// GenerateRootQC generates QC for root block, caller needs to provide votes for root QC and
+// participantData to build the QC.
+// NOTE: only one DKG private key is required to sign the QC, for all others entries we can
+// provide only identity info without private keys.
```
AlexHentschel, line 44: A QC is not signed. Lets update this:

```suggestion
// NOTE: at the moment, we require private keys for one node because we we re-using the full business logic, which assumes that only consensus participants construct QCs, which also have produce votes. 
// TODO: modularize QC construction code (and code to verify QC) to be instantiated without needing private keys. 
```

AlexHentschel (approved):
Looks good to me

cmd/bootstrap/cmd/finalize.go
AlexHentschel: renamed methods for reading node infos to reflect that fact that we are _reading_ the infos and not generating them: 
https://github.com/onflow/flow-go/pull/1371/commits/175116b76087943ec719fe04e9e4c3e11d963424

cmd/bootstrap/cmd/constraints.go
AlexHentschel: implemented in https://github.com/onflow/flow-go/pull/1371/commits/503e8eae901a323858140c152d7e2b3694e1307c and https://github.com/onflow/flow-go/pull/1371/commits/e011135fd2721458d04e7b84b928f296c2a9bb02

cmd/bootstrap/cmd/constraints.go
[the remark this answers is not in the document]
AlexHentschel: fixed in https://github.com/onflow/flow-go/pull/1371/commits/503e8eae901a323858140c152d7e2b3694e1307c and https://github.com/onflow/flow-go/pull/1371/commits/e011135fd2721458d04e7b84b928f296c2a9bb02

AlexHentschel (approved):
🎉

cmd/bootstrap/cmd/finalize.go
```diff
@@ -76,26 +75,26 @@ func addFinalizeCmdFlags() {
...
 	cmd.MarkFlagRequired(finalizeCmd, "partner-stakes")
+	cmd.MarkFlagRequired(finalizeCmd, "dkg-data")
 
 	// required parameters for generation of root block, root execution result and root block seal
-	finalizeCmd.Flags().StringVar(&flagRootChain, "root-chain", "local", "chain ID for the root block (can be 'main', 'test', 'canary', 'bench', or 'local'")
-	finalizeCmd.Flags().StringVar(&flagRootParent, "root-parent", "0000000000000000000000000000000000000000000000000000000000000000", "ID for the parent of the root block")
-	finalizeCmd.Flags().Uint64Var(&flagRootHeight, "root-height", 0, "height of the root block")
-	finalizeCmd.Flags().StringVar(&flagRootTimestamp, "root-timestamp", time.Now().UTC().Format(time.RFC3339), "timestamp of the root block (RFC3339)")
+	finalizeCmd.Flags().StringVar(&flagRootBlock, "root-block", "",
+		"path to a JSON file containing root block and votes")
```
smnzhu, line 88: The description should probably also be updated to reflect that the root block file only contains root block

jordanschalm (approved):

PR #1373, opened 2021-09-28
removing the check for consensus committee Byzantine threshold

vishalchangrani:
For the consensus node test which has a 79 external partner consensus nodes and only 46 of our nodes, the finalize command fails with the error: 
`FTL will not bootstrap configuration without Byzantine majority of consensus nodes`
This assumption of the internal nodes having a majority will no longer be true starting from the next spork, hence this change to remove that check.

### DIFF
--- a/cmd/bootstrap/cmd/constraints.go
+++ b/cmd/bootstrap/cmd/constraints.go
@@ -7,7 +7,6 @@ import (
 )
 
 // Checks constraints about the number of partner and internal nodes.
-// Internal nodes must comprise >2/3 of consensus committee.
 // Internal nodes must comprise >2/3 of each collector cluster.
 func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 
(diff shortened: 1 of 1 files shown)


jordanschalm:
This has a bit of overlap with https://github.com/onflow/flow-go/pull/1371. Yurii already split the SN/LN constraints

smnzhu (approved):

cmd/bootstrap/cmd/constraints.go
```diff
@@ -29,16 +28,6 @@ func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 		}
 	}
 
-	// check consensus committee Byzantine threshold
-	partnerCONCount := partners.Filter(filter.HasRole(flow.RoleConsensus)).Count()
-	internalCONCount := internals.Filter(filter.HasRole(flow.RoleConsensus)).Count()
-	if internalCONCount <= partnerCONCount*2 {
-		log.Fatal().Msgf(
-			"will not bootstrap configuration without Byzantine majority of consensus nodes: "+
```
huitseeker, line 37: This should probably have been "Byzantine-resistant majority".

huitseeker (changes requested):
We also need to edit the check applied to collection clusters (around `if clusterInternalCount <= clusterPartnerCount*2`)

vishalchangrani:
closing this PR since [this PR](https://github.com/onflow/flow-go/pull/1371) takes care of this change

PR #1374, opened 2021-09-28
implement pull root block and push root block vote

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/5888

### TODO
- [x] Add documentation to README.md
- [ ] Test


cmd/bootstrap/transit/cmd/pull_root_block.go
```diff
@@ -0,0 +1,77 @@
...
+	fullOutpath = filepath.Join(flagBootDir, filepath.Base(objectName))
+
+	log.Info().Msgf("downloading random beacon key: %s", objectName)
+
+	err = bucket.DownloadFile(ctx, client, fullOutpath, objectName)
+	if err != nil {
+		log.Fatal().Err(err).Msg("could not download file from google bucket")
+	}
+
+	err = unWrapFile(flagBootDir, nodeID)
```
vishalchangrani, line 71: we can remove this call and the `getAdditionalFilesToDownload` call from `pull.go` since it is already being done now here for the consensus nodes. The additional files were the random beacon files nothing else for consensus nodes.

vishalchangrani: nevermind, not all consensus node operators will run this command so for them we need to have the pull command work as-is

vishalchangrani (approved):

cmd/bootstrap/transit/cmd/push_root_block_vote.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Fatal().Err(err).Msg("could not unmarshal DKG private key data")
+	}
+
+	randomBeaconPrivKey := priv.RandomBeaconPrivKey.PrivateKey
+	stakingPrivKey := nodeInfo.StakingPrivKey.PrivateKey
+	identity := &flow.Identity{
+		NodeID:        nodeID,
+		Address:       nodeInfo.Address,
+		Role:          nodeInfo.Role,
+		Stake:         0, // TODO: is this okay?
```
AlexHentschel, line 79: Technically, I don't think the value for `Stake` this should matter. 
The correct value for the stake should be in the node's public node info (also available as file) but we can update this later.

AlexHentschel (approved):
nice

cmd/bootstrap/transit/cmd/push_root_block_vote.go
vishalchangrani: we use the value of `1000` when we stake (it is not a really used but we set stake of all nodes to 1000 when we spork).

Kay-Zee: yeah, we might as well just use the 1000 we've been using, to ensure continuity

jordanschalm (approved):
🎸

smnzhu:
bors merge

smnzhu:
bors merge

smnzhu:
bors retry

PR #1376, opened 2021-09-29
Fix encoding user input in failed transaction message

m4ksio:


huitseeker (approved):
The test is a tad elaborate, I'm impressed you went through the trouble! Thanks!

turbolent (approved):

AlexHentschel (approved):
Looks good to me. Though, I don't think I have enough background to reliably find subtle edge cases. 

### Style suggestion:
would be great to document in the godoc of method  
https://github.com/onflow/flow-go/blob/71c36cf8bb6ee62c7cbbb98253d98992dfd9c13d/fvm/handler/accountKey.go#L250
what specific error types it returns under which conditions (at least `AccountNotFoundError` and `ValueError`)

fvm/handler/accountKey_test.go
```diff
@@ -0,0 +1,101 @@
...
+		SeqNumber: 0,
+		Weight:    1000,
+		Revoked:   false,
+	}
+
+	encodedPublicKey, err := flow.EncodeRuntimeAccountPublicKey(accountPublicKey)
+	require.NoError(t, err)
+
+	err = akh.AddEncodedAccountKey(runtime.Address(address), encodedPublicKey)
+	require.Error(t, err)
```
AlexHentschel, line 45: lets check the specific error type here, you are expecting to return

m4ksio: Done and done

m4ksio:
bors merge

PR #1377, opened 2021-09-29
[consensus, synchronization] Fix forks preventing reception of blocks at same height

huitseeker:
I highly suggest looking at individual commits (code AND messages) to review.

When receiving a block (`HandleBlock`), we test its status in existing tables to see if it was already received. 

If a `Received` status is in the height table, we reject the block, without checking if the block we are receiving is different than the one we've already received at the same height.

Moreover, this logic applies even if all blocks involved are requested and received by blockID only.

This:
- makes sure we report a request status upon block reception (`getRequestStatus`) favoring requests by ID,
- adds basic property testing, which exercises the bug from on a forest of blocks which includes forks,
- does not change anything to the pattern or volume of requested heights.

Fixes [#5884](https://github.com/dapperlabs/flow-go/issues/5884)

### DIFF
--- a/module/synchronization/core.go
+++ b/module/synchronization/core.go
@@ -204,12 +204,14 @@ func (c *Core) getRequestStatus(height uint64, blockID flow.Identifier) *Status 
 	heightStatus := c.heights[height]
 	idStatus := c.blockIDs[blockID]
 
-	if heightStatus.WasQueued() {
-		return heightStatus
-	}
 	if idStatus.WasQueued() {
 		return idStatus
 	}
+	// Only return the height status if there is no matching status for the ID
+	if heightStatus.WasQueued() {
+		return heightStatus
+	}
+
 	return nil
 }
 
--- a/module/synchronization/core_rapid_test.go
+++ b/module/synchronization/core_rapid_test.go
@@ -57,7 +57,9 @@ func (r *rapidSync) RequestByID(t *rapid.T) {
 // HandleByID is an action that provides a block header to the sync engine
 func (r *rapidSync) HandleByID(t *rapid.T) {
 	b := rapid.SampledFrom(r.store).Draw(t, "id_handling").(flow.Header)
-	r.core.HandleBlock(&b)
+	success := r.core.HandleBlock(&b)
+	assert.True(t, success || r.idRequests[b.ID()] == 0)
+
 	// we decrease the pending requests iff we have already requested this block
 	// and we have not received it since
 	if r.idRequests[b.ID()] == 1 {
--- a/module/synchronization/core_rapid_test.go
+++ b/module/synchronization/core_rapid_test.go
@@ -30,9 +30,10 @@ func populatedBlockStore(t *rapid.T) []flow.Header {
 }
 
 type rapidSync struct {
-	store      []flow.Header
-	core       *Core
-	idRequests map[flow.Identifier]int // pushdown automaton to track ID requests
+	store          []flow.Header
+	core           *Core
+	idRequests     map[flow.Identifier]int // pushdown automaton to track ID requests
+	heightRequests map[uint64]int          // pushdown automaton to track height requests
 }
 
 // Init is an action for initializing a rapidSync instance.
@@ -44,6 +45,7 @@ func (r *rapidSync) Init(t *rapid.T) {
 
 	r.store = populatedBlockStore(t)
 	r.idRequests = make(map[flow.Identifier]int)
+	r.heightRequests = make(map[uint64]int)
 }
 
 // RequestByID is an action that requests a block by its ID.
@@ -52,6 +54,16 @@ func (r *rapidSync) RequestByID(t *rapid.T) {
 	r.core.RequestBlock(b.ID())
 	// Re-queueing by ID should always succeed
 	r.idRequests[b.ID()] = 1
+	// Re-qeueuing by ID "forgets" a past height request
+	r.heightRequests[b.Height] = 0
+}
+
+// RequestByHeight is an action that requests a specific height
+func (r *rapidSync) RequestByHeight(t *rapid.T) {
+	b := rapid.SampledFrom(r.store).Draw(t, "id_request").(flow.Header)
+	r.core.RequestHeight(b.Height)
+	// Re-queueing by height should always succeed
+	r.heightRequests[b.Height] = 1
 }
 
 // HandleByID is an action that provides a block header to the sync engine
@@ -65,6 +77,8 @@ func (r *rapidSync) HandleByID(t *rapid.T) {
 	if r.idRequests[b.ID()] == 1 {
 		r.idRequests[b.ID()] = 0
 	}
+	// we eagerly remove height requests
+	r.heightRequests[b.Height] = 0
 }
 
 // Check runs after every action and verifies that all required invariants hold.
@@ -83,6 +97,21 @@ func (r *rapidSync) Check(t *rapid.T) {
 		} else {
 			t.Fatalf("incorrect management of idRequests in the tests")
 		}
+	}
+	heights, blockIDs := r.core.getRequestableItems()
+	// the queueing logic queues intervals, while our r.heightRequests only queues specific requests
+	var activeHeights []uint64
+	for k, v := range r.heightRequests {
+		if v == 1 {
+			activeHeights = append(activeHeights, k)
+		}
+	}
+	assert.Subset(t, heights, activeHeights, "sync engine's height request tracking lost heights")
+
+	for _, bID := range blockIDs {
+		v, ok := r.idRequests[bID]
+		require.True(t, ok)
+		assert.Equal(t, 1, v, "blockID %v is supposed to be pending but is not", bID)
 	}
 }
 
--- a/module/synchronization/core_rapid_test.go
+++ b/module/synchronization/core_rapid_test.go
@@ -1,0 +1,89 @@
+package synchronization
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"pgregory.net/rapid"
+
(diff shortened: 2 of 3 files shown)


module/synchronization/core.go
```diff
@@ -204,12 +204,14 @@ func (c *Core) getRequestStatus(height uint64, blockID flow.Identifier) *Status
...
-	}
 	if idStatus.WasQueued() {
 		return idStatus
 	}
+	// Only return the height status if it's not a known mismatch  for the query ID
+	if heightStatus.WasQueued() && (!heightStatus.WasReceived() || heightStatus.Header.ID() == blockID) {
+		return heightStatus
+	}
+
 	return nil
```
smnzhu, line 215: As long as we don't remove [this line](https://github.com/onflow/flow-go/blob/master/module/synchronization/core.go#L90), we will have a corner case:

At some point, compliance engine makes request by ID for block X (which has height h), and when we receive that block we will set the status for height h as "received".
However, block X may never actually be finalized (it may be orphaned and another block y at height h finalized instead), yet the sync engine will never again make any requests for height h since it was already marked as received. 
So, the only way we will receive y is if the compliance engine explicitly requests it.

I think the problem here is range requests / height-based requests are intended to return only finalized blocks (so that there is only one possible answer for each height), whereas requests by block ID may return unfinalized blocks.
However, we receive the same response for both types of requests, so it's not always possible to distinguish whether the original request made was for the height or for the block ID.

smnzhu: I think we should just remove that line, and keep requesting heights until we actually finalize the block at that height.

module/synchronization/core_rapid_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"pgregory.net/rapid"
```
smnzhu, line 11: :)

module/synchronization/core.go
huitseeker: The height requests are to be issued only for blocks presumed to be finalized in the first place. In practice, this presumption is built when we receive a hint of a finalization height that's past ours (which may or may not be accurate). 

Indeed, *after* the finalization boundary, there's no way of keeping track of the number of forks, so there's no meaningful _specific_ number of blocks at a given height that one would know to request (and keep track of). There are two non-specific amounts of blocks we can try to requests:

1. at least one block at that height, which is the policy that's configured today,
2. all the blocks at that height, where we keep issuing requests at that height, which is what you suggest.

Here are two points that make me think we want to keep what we have (1.):
- the absence of height requests will not create chain starvation. Upon discovering any new block (through e.g. a QC obtained by consensus follower broadcast), the node will request blocks in causal order, following parent pointers, without any interference from the height-based logic (once this PR is merged, that is).
- the finalization height hints are not authenticated in any way, so that any node can make another start to issue repeated requests for a high height. That can create an amplification attack.

On the second point:
- let's say all nodes are at view v for height 42, for which a block proposal has just been issued,
- Eve (dishonest) issues rsync requests for height 51 (42 + Tolerance -1), thereby getting everybody to queue heights 43-51 _indefinitely_†,
- because the height requests are no longer cancelled once fulfilled, all nodes are _forever_ liable to receive any new blocks at heights 43-51†,
- as we cache blocks and queue requests for block ancestors before verifying a block's signatures in compliance (see `processBlockAndDescendants`), any block sent for those heights is stored by the recipient, as long as its parent is unknown,
- Eve and her f-1 dishonest friends start flooding honest nodes one by one with all-distinct blocks at heights 43-51 and unknown parents, 
- the target eventually OOMS under a f-to-1 flood. When Eve and her gang move on to a new target, they can repeat repeat the sync request steps above if necessary, to prime targets while keeping up with finalization height. 
- all honest nodes eventually OOM, as long as the f-to-1 flood is faster than cache purges from finalization.

†: as long as finalization height allows

Now, I have several problems with the above situation, but surely we can agree it's better if the words "indefinitely" and "finalization" (above) are not part of the picture unless we have a clear need for them. The current logic reflects "only keep the first answer to a request by height and any answer to a request by ID", thereby avoiding that particular pitfall (not that it's the only one of course).

zhangchiqing (approved):
Looks good. Thanks for the fix!

module/synchronization/core_rapid_test.go
```diff
@@ -0,0 +1,120 @@
...
+	r.core.RequestBlock(b.ID())
+	// Re-queueing by ID should always succeed
+	r.idRequests[b.ID()] = 1
+	// Re-qeueuing by ID "forgets" a past height request
+	r.heightRequests[b.Height] = 0
+}
+
+// RequestByHeight is an action that requests a specific height
+func (r *rapidSync) RequestByHeight(t *rapid.T) {
+	b := rapid.SampledFrom(r.store).Draw(t, "id_request").(flow.Header)
```
zhangchiqing, line 63: What is this `Id_request` for? Does it mean `Draw` with the same value will take the same sample?

module/synchronization/core_rapid_test.go
```diff
@@ -0,0 +1,120 @@
...
+	// and we have not received it since
+	if r.idRequests[b.ID()] == 1 {
+		r.idRequests[b.ID()] = 0
+	}
+	// we eagerly remove height requests
+	r.heightRequests[b.Height] = 0
+}
+
+// Check runs after every action and verifies that all required invariants hold.
+func (r *rapidSync) Check(t *rapid.T) {
```
zhangchiqing, line 85: Does these check fail without the fix to the sync core?

huitseeker: Yes, and you can see that failure by running the test on the first commit.

smnzhu (approved):

huitseeker:
bors merge

module/synchronization/core.go
smnzhu: One possible issue I can think of with this, is if a dishonest node responds to height-based requests (ie range requests) with blocks which are non-finalized, but at the correct height that is requested.

Then, once the recipients receives those blocks, they will mark those heights as received, and hence never make range requests for those again.

As you say, we're not completely out of luck here because of this point:

> the absence of height requests will not create chain starvation. Upon discovering any new block (through e.g. a QC obtained by consensus follower broadcast), the node will request blocks in causal order, following parent pointers, without any interference from the height-based logic (once this PR is merged, that is).

However, it can still cause issues with performance for a node which has started late and is catching up with the chain.

In fact, this was the exact problem that we saw with the sync engine bug in the consensus follower right? The integration test kept timing out because range requests weren't being issued, so the consensus follower had to rely on requesting blocks in causal order (which took forever).

Having said that, I think this is a good solution for now and I'm in favour of merging if only for the new tests it adds :)

Just thinking perhaps we need to open an issue to further explore better solutions for this in the future, because I don't really consider this problem fully solved atm.

module/synchronization/core_rapid_test.go
```diff
@@ -0,0 +1,120 @@
...
+	for k, v := range r.idRequests {
```
jordanschalm, line 86:
```suggestion
	for id, count := range r.idRequests {
```

module/synchronization/core.go
huitseeker: > Then, once the recipients receives those blocks, they will mark those heights as received, and hence never make range requests for those again.

We're not: every time we interact with another honest node, that node lets us know that the finalization height is ahead of us. A dishonest node can send us rubbish, but it cannot make our finalized height advance without sending us accurate information. So on every height hint, we will re-queue any missing heights.

> In fact, this was the exact problem that we saw with the sync engine bug in the consensus follower right? The integration test kept timing out because range requests weren't being issued, 

We only *rely* on range request issuance for finalized heights. The issue we saw in integration tests did not show real-world interaction with other nodes, wherein we `pollHeight` on a regular basis, thereby re-queueing heights de-queued by fork blocks.

As far as I'm concerned, de-queuing on block reception is suitable behavior.

This PR does not claim to solve all possible inefficiencies of the sync engine, and we can certainly track further work there, but it does fix forked block reception clobbering acceptance of other blocks at the same height, which is all it claims to do.

jordanschalm (approved):
🎸

module/synchronization/core.go
jordanschalm: On the note of continuing to improve the sync engine, there are some historical reports of issues with the sync engine and some suggestions for improvements in [this issue](https://app.zenhub.com/workspaces/flow-protocol-5a2ad022cee40e25b0885f82/issues/dapperlabs/flow-go/4965).

smnzhu: > We don’t: every time we interact with another honest node, that node lets us know that the finalization height is ahead of us. A dishonest node can send us rubbish, but it cannot make our finalized height advance without sending us accurate information. So on every height hint, we will re-queue any missing heights.

But we won’t actually requeue missing heights, and that’s the problem.

Once we’ve received *any* block at a height h, further calls to [`HandleHeight`](https://github.com/onflow/flow-go/blob/fde1a3f8226c2fd6d51af66863aae2973772a7c1/module/synchronization/core.go#L98) will be [noops](https://github.com/onflow/flow-go/blob/fde1a3f8226c2fd6d51af66863aae2973772a7c1/module/synchronization/core.go#L177-L182). So even if we receive a height hint via sync request from an honest node, we will not actually ever request that height again.

I think we need some mechanism that *actually* re-queues missing heights if we receive a height-hint for a height which we think we've received, but have not actually finalized locally. This can be done if we change [this](https://github.com/onflow/flow-go/blob/fde1a3f8226c2fd6d51af66863aae2973772a7c1/module/synchronization/core.go#L109) to a call to `RequestHeight` instead.

smnzhu: https://github.com/onflow/flow-go/pull/1377

huitseeker: @smnzhu Yep, exactly. Thanks for that PR!

PR #1380, opened 2021-09-29
[Storage migration V6] Add storage_index and account storage registers

SupunS:
Add the `storage_index` and the account storage registers that were not getting added during storage migration


fvm/transactionInvocator.go
```diff
@@ -269,30 +269,33 @@ func (i *TransactionInvocator) deductTransactionFees(env *TransactionEnv, proc *
...
+			Label:          sema.ArgumentLabelNotRequired,
+			Identifier:     "frozen",
+			TypeAnnotation: sema.NewTypeAnnotation(sema.BoolType),
+		},
+	},
+	ReturnTypeAnnotation: &sema.TypeAnnotation{
+		Type: sema.VoidType,
+	},
+}
+
```
ramtinms, line 289: are these changes necessary, I guess we could scope out these

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -54,79 +53,6 @@ var storageMigrationV5DecMode = func() cbor.DecMode {
 	return decMode
 }()
```
turbolent, line 54: Maybe add a comment to the lines above, for `cborTagStorageReference` and `storageMigrationV5DecMode`, that they are copied from the old Cadence version, because they are not exported.

We should maybe export the CBOR tags from Cadence to make this easier in the future.

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -191,8 +117,16 @@ func (m *StorageFormatV6Migration) migrate(payloads []ledger.Payload) ([]ledger.
 
 	m.migratedPayloadPaths = make(map[storagePath]bool, 0)
 
-	baseStorage := newEncodingBaseStorage()
-	m.initPersistentSlabStorage(baseStorage)
+	fvmPayloads, storagePayloads, slabPayloads := splitPayloads(payloads)
+	if len(slabPayloads) != 0 {
```
turbolent, line 121: 👍

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -191,8 +117,16 @@ func (m *StorageFormatV6Migration) migrate(payloads []ledger.Payload) ([]ledger.
 
 	m.migratedPayloadPaths = make(map[storagePath]bool, 0)
```
turbolent, line 118: What is the purpose of this? I see that we add paths when we migrate individual payloads, but I did not quite understand why we check it and panic in the old interpreter's `WithStorageReadHandler`

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1079,12 +965,43 @@ func (c *ValueConverter) Convert(value oldInter.Value) (result newInter.Value) {
...
+			c.migration.reportFile.WriteString(
+				fmt.Sprintf(
+					"skipped migrating value due to missing static type: owner: %x, type: %s\n",
+					value.GetOwner(),
+					typeError.TypeID,
+				),
+			)
+		} else {
+			err := r.(error)
+			if strings.Contains(err.Error(), "expected `StoragePath`, got `Path`") {
```
turbolent, line 985: Maybe we can generalize this, as some errors may also e.g. expect `CapabilityPath`

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1130,7 +1047,7 @@ func (c *ValueConverter) VisitArrayValue(_ *oldInter.Interpreter, value *oldInte
```
turbolent, line 1047: Re: `if newElement != nil` above: Is there a chance that the migration will produce an array where some elements are `nil`?

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1295,7 +1212,7 @@ func getValue(
 		}
```
turbolent, line 1212: Same for `if newValue != nil` above: Shouldn't the migration abort and not produce a dictionary where values are `nil`?

turbolent (commented):
Nice work, just have some clarification question

cmd/util/ledger/migrations/storage_v6.go
SupunS: yeah, I'm working on adding more specific logs for different errors, this is also cleaned up in that. Will push the changes soon.

cmd/util/ledger/migrations/storage_v6.go
SupunS: This is the case where, if some element is not migrated due to some error (missing type/ invalid type etc.), then we don't abort, but rather log the error and continues. In that case, migrated value (i.e: newElement) is nil.

cmd/util/ledger/migrations/storage_v6.go
SupunS: This is a sanity check to make sure we don't re-add deferred payloads as separate payloads, and vise-versa (i.e: already migrated payloads are not re-added as deferred payloads)

SupunS: Added in 8677eb7

cmd/util/ledger/migrations/storage_v6.go
SupunS: Added in 8677eb7

ramtinms (approved):
Looks good to me

cmd/util/ledger/migrations/storage_v6.go
turbolent: Ah, that makes sense 👍

cmd/util/ledger/migrations/storage_v6.go
turbolent: I see. We definitely make sure there are no such conversion errors reported after the migration runs. cc @Kay-Zee

cmd/execution/main.go
```diff
@@ -134,6 +136,7 @@ func main() {
 			flags.BoolVar(&pauseExecution, "pause-execution", false, "pause the execution. when set to true, no block will be executed, but still be able to serve queries")
 			flags.BoolVar(&enableBlockDataUpload, "enable-blockdata-upload", false, "enable uploading block data to GCP Bucket")
 			flags.StringVar(&gcpBucketName, "gcp-bucket-name", "", "GCP Bucket name for block data uploader")
+			flags.BoolVar(&atreeValidationEnabled, "atree-validation", false, "validates all atree values after mutations")
```
turbolent, line 139: @zhangchiqing @Kay-Zee We want to have this sanity check enabled on Canary, and turned off on Testnet and Mainnet

turbolent (approved):
Looks good!

cmd/execution/main.go
zhangchiqing: ok

PR #1384, opened 2021-09-29
Reenable reporters

janezpodhostnik:


ledger/complete/ledger.go
```diff
@@ -252,6 +252,10 @@ func (l *Ledger) ExportCheckpointAt(
 	// get trie
 	t, err := l.forest.GetTrie(ledger.RootHash(state))
 	if err != nil {
+		tries, err := l.forest.GetTries()
+		for _, try := range tries {
+			l.logger.Info().Str("rootHash", try.RootHash().String()).Msg("t")
+		}
```
ramtinms, line 258: you probably don't need to iterate, you call the mostRecentTrie method

janezpodhostnik: This is just a helper, so I can find the hash :). I wasn't planing on committing this. Thanks fo the tip though! I will be using that in the future.

PR #1388, opened 2021-09-29
[Bootstrap] split sign and upload into two steps

smnzhu:


huitseeker (approved):
The split makes sense to me!

AlexHentschel (approved):
👍

PR #1392, opened 2021-09-30
[Flaky test] peer manager flakiness fix

smnzhu:
The flaky test in question is `TestPeerManagerTestSuite/TestPeriodicPeerUpdate`, but the actual issue is in the implementation of `PeerManager` and `engine.Unit`.

The root cause is described in comments below.

Before the change, `go test -race -count 10 -tags relic -run TestPeerManagerTestSuite -testify.m TestPeriodicPeerUpdate` almost always has at least one failure

After the change, never fails.

### DIFF
--- a/engine/unit.go
+++ b/engine/unit.go
@@ -74,14 +74,15 @@ func (u *Unit) LaunchAfter(delay time.Duration, f func()) {
 // If f is executed, the unit will not shut down until after f returns.
 func (u *Unit) LaunchPeriodically(f func(), interval time.Duration, delay time.Duration) {
 	u.Launch(func() {
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+
 		select {
 		case <-u.ctx.Done():
 			return
 		case <-time.After(delay):
 		}
 
-		ticker := time.NewTicker(interval)
-		defer ticker.Stop()
 		for {
 			select {
 			case <-u.ctx.Done():
--- a/network/p2p/peerManager.go
+++ b/network/p2p/peerManager.go
@@ -84,6 +84,8 @@ func PeerManagerFactory(peerManagerOptions []Option, connectorOptions ...Connect
 
 // Ready kicks off the ambient periodic connection updates.
 func (pm *PeerManager) Ready() <-chan struct{} {
+	pm.unit.Launch(pm.updateLoop)
+
 	// makes sure that peer update request is invoked once before returning
 	pm.RequestPeerUpdate()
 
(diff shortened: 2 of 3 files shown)


network/p2p/peerManager.go
```diff
@@ -84,8 +84,10 @@ func PeerManagerFactory(peerManagerOptions []Option, connectorOptions ...Connect
 
 // Ready kicks off the ambient periodic connection updates.
 func (pm *PeerManager) Ready() <-chan struct{} {
+	pm.unit.Launch(pm.updateLoop)
+
 	// makes sure that peer update request is invoked once before returning
-	pm.RequestPeerUpdate()
+	pm.peerRequestQ <- struct{}{}
```
smnzhu, line 90: Previously we were launching the loop *after* invoking peer update request for the first time.

However, if we look at the implementation of `RequestPeerUpdate` below, we see that this first peer update request is *always* dropped immediately.

We need to launch the update loop first, so that by the time the peer update request is triggered, there is actually someone listening to the channel. In fact, calling `unit.Launch(pm.updateLoop)` is not sufficient, because there is still a race condition where the peer update could be attempted before the `updateLoop` goroutine has entered the first loop iteration.

So, actually we need to make this first peer update request blocking, to ensure that it is not dropped.

engine/unit.go
```diff
@@ -74,14 +74,15 @@ func (u *Unit) LaunchAfter(delay time.Duration, f func()) {
...
 		case <-u.ctx.Done():
 			return
 		case <-time.After(delay):
 		}
 
-		ticker := time.NewTicker(interval)
-		defer ticker.Stop()
 		for {
 			select {
 			case <-u.ctx.Done():
```
smnzhu, line 88: I would consider this arguably a bug, which is why I've made the fix directly to `engine.Unit`.

The problem here is that the first tick from a `time.Ticker` will not be sent until `interval` has passed. Thus, the first invocation of the given function `f` is not made until `delay + interval` has passed.

Therefore, if the random delay chosen in `PeerManager` is almost as long as `interval`, we may wait almost 2*interval before getting our second invocation of `f`, hence timing out in the unit test.

vishalchangrani (approved):
make sense to me.

network/p2p/peerManager_test.go
```diff
@@ -142,7 +142,7 @@ func (suite *PeerManagerTestSuite) TestPeriodicPeerUpdate() {
 		}
 	}).Return(nil)
 
-	peerUpdateInterval := 5 * time.Millisecond
+	peerUpdateInterval := 10 * time.Millisecond
```
gomisha, line 145: Is this increase in update interval to decrease the chance of timing out?

gomisha (approved):
This test isn't added on the [quarantine list](https://www.notion.so/dapperlabs/Flaky-and-Skipped-Tests-Quarantine-789e45dc94a24fecba665f180903401f) - it looks like you fixed the flakiness but it sounds like it's worth monitoring to see if it continues being flaky

network/p2p/peerManager_test.go
smnzhu: Just to decrease the chance of all the steps in between adding unintended latency which puts us over the time limit.

PR #1397, opened 2021-09-30
[Storage Migration V6] Update missing/broken types of values

SupunS:


Kay-Zee (approved):

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1055,15 +1061,15 @@ func (c *ValueConverter) VisitStringValue(_ *oldInter.Interpreter, value *oldInt
 func (c *ValueConverter) VisitArrayValue(_ *oldInter.Interpreter, value *oldInter.ArrayValue) bool {
 	newElements := make([]newInter.Value, 0)
 
+	arrayStaticType := ConvertStaticType(value.StaticType()).(newInter.ArrayStaticType)
+
 	for _, element := range value.Elements() {
-		newElement := c.Convert(element)
+		newElement := c.Convert(element, arrayStaticType.ElementType())
```
turbolent, line 1067: Maybe refactor `arrayStaticType.ElementType()` out of the loop into a variable, as it is effectively constant

```suggestion
		newElement := c.Convert(element, arrayStaticType.ElementType())
```

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1055,15 +1061,15 @@ func (c *ValueConverter) VisitStringValue(_ *oldInter.Interpreter, value *oldInt
 func (c *ValueConverter) VisitArrayValue(_ *oldInter.Interpreter, value *oldInter.ArrayValue) bool {
 	newElements := make([]newInter.Value, 0)
 
+	arrayStaticType := ConvertStaticType(value.StaticType()).(newInter.ArrayStaticType)
```
turbolent, line 1064: 👍

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1027,6 +1027,12 @@ func (c *ValueConverter) Convert(value oldInter.Value) (result newInter.Value) {
 		panic("converted value is nil")
 	}
 
+	switch expectedType {
+	case newInter.PrimitiveStaticTypeUInt64:
+		if intValue, ok := c.result.(newInter.IntValue); ok {
+			c.result = newInter.ConvertUInt64(intValue)
+		}
+	}
```
turbolent, line 1035: Why is this handled specifically? Is the data somehow wrong? (Static type is `UInt64`, but value is `Int`?)

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -461,18 +452,24 @@ func (m *StorageFormatV6Migration) initNewInterpreter() {
 		newInter.WithStorage(m.storage),
 		newInter.WithImportLocationHandler(
 			func(inter *newInter.Interpreter, location common.Location) newInter.Import {
-				program, err := m.loadProgram(location)
-				if err != nil {
-					panic(err)
+				var program *newInter.Program
+				if location == stdlib.CryptoChecker.Location {
+					program = newInter.ProgramFromChecker(stdlib.CryptoChecker)
```
turbolent, line 457: Good catch. I think next time we should try use the Cadence `runtime` package, which handles this already. We might have to extend it for such migration use-cases, it's currently mainly only providing functionality for the FVM/EN.

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -25,10 +25,12 @@ import (
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/common"
 	newInter "github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/cadence/runtime/stdlib"
 
 	oldInter "github.com/onflow/cadence/v19/runtime/interpreter"
 )
 
+// cborTagStorageReference is a duplicate of the same from cadence v0.19.0
```
turbolent, line 33: 👍

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1315,12 +1417,31 @@ func (c *ValueConverter) VisitDeployedContractValue(_ *oldInter.Interpreter, _ o
...
 
 	case oldInter.InterfaceStaticType:
+		// NonFungibleToken.NFT is a struct, but is stored as an interface type.
+		// Rectify this by returning a composite static type.
+		if location, ok := typ.Location.(common.AddressLocation); ok {
+			if location.Address == testnetNonFungibleTokenContractAddress &&
+				typ.QualifiedIdentifier == "NonFungibleToken.NFT" {
+				return newInter.NewCompositeStaticType(location, typ.QualifiedIdentifier)
+			}
+		}
```
turbolent, line 1443: I think this might be a general problem, not just a NFT contract-specific problem.
`NonFungibleToken.NFT` is a "type requirement", a type in an interface which requires the concrete type to provide a concrete type for. I don't remember how this is is implemented in the checker, but I think as an interface type. 

Is it actually a problem that the static type is an interface type here?

SupunS: Yeah, this is a general problem I think. And yeah, the problem is that the static type is stored as an interface type, but in the elaboration, it is in the composite types list.
This could be due to several problems:
-  `NonFungibleToken.NFT` was initially defined as an interface and later changed to a struct/resource?
- Something might not be correct in static-type creation in cadence. (or was not correct, and fixed now?)
- Something went wrong in inferring the static type during V5

I also ran into another very simmilar one, with a static type `Foo.Bar`, where now both `Foo` and `Bar` are concrete types. So I'm starting to feel it could be the first case. But need to investigate a bit more.

cmd/util/ledger/migrations/storage_v6.go
SupunS: There are few values, where the expected type is `UInt64` but the value is `Int`. These potentially are the ones where the type is changed when the contract is updated.
There are more similar ones in testnet, but not fixable. e.g.: expected `String` but found `UInt64`, etc.

turbolent: Ah, I see, this could come from before we had the contract update validation, and before we had the container mutation, makes sense 👍

turbolent (approved):
Looks good! This PR seems to mainly just fix particular edge cases found in the current states of networks

SupunS:
Thanks for the reviews!

PR #1399, opened 2021-09-30
[sync-engine, test-only] Model height management in tests of the sync engine

### LINKED ISSUE #5894: [Access] Expose GetAccountBalance and GetAccountKeys in script execution module
## Problem Definition

Currently, users can get their account's FLOW balance in 1 of 2 ways:
* Using a script that queries the flow token vault's balance
* Using `GetAccount`

Most users use `GetAccount` because it's convenient. The downside is that this endpoint does 2 additional lookups, which may return a lot of unnecessary data:
* Get account keys -> some accounts have a large number of keys resulting in a large number of register lookups in the db.
* Get contracts -> there are generally few or no contracts, but contracts also require additional lookups.

Under the covers, these are all implemented a scripts, so `GetAccount` ultimately runs 3 scripts per call.

## Proposed Solution

Expose the `GetAccountBalance` and `GetAccountKeys` methods within the `ScriptExecutor`.

These will look similar to `GetAccountAtBlockHeight`, but specific to these methods
https://github.com/onflow/flow-go/blob/aa22a874366b28f508df5c174de45e47d57146df/module/execution/scripts.go#L116-L127

This will also require updating to the Query Executor:
https://github.com/onflow/flow-go/blob/51df877a58515ab7cba65ac33b42f27ba8774673/engine/execution/computation/query/executor.go#L31

and objects on the path to this method:
https://github.com/onflow/flow-go/blob/c56c2169ee1c9c66f71db67dc06f184aa93f82d6/fvm/environment/account_info.go#L232-L256

These new methods would instead call into the `GetAccountBalance` and `GetPublicKeys`.

## Definition of Done

`ScriptExecutor` has new methods to get an account balance and keys by height.
There are unit tests covering the new methods

huitseeker:
This augments the state machine ([rapid](https://github.com/flyingmutant/rapid)) test to model the reasons for dequeueing a height from pending requests.

Fixes #5894


module/synchronization/core_rapid_test.go
```diff
@@ -32,8 +33,8 @@ func populatedBlockStore(t *rapid.T) []flow.Header {
 type rapidSync struct {
 	store          []flow.Header
 	core           *Core
-	idRequests     map[flow.Identifier]int // pushdown automaton to track ID requests
-	heightRequests map[uint64]int          // pushdown automaton to track height requests
```
smnzhu, line 36: Why don't we just use map to bool? Maybe I'm missing it but I don't see anywhere we increment this value that would cause it to ever be something greater than 1?

module/synchronization/core_rapid_test.go
```diff
@@ -77,35 +93,78 @@ func (r *rapidSync) HandleByID(t *rapid.T) {
...
+		block, foundBlock := findHeader(r.store, func(h flow.Header) bool {
+			return h.ID() == id
+		})
+		require.True(t, foundBlock, "incorrect management of idRequests in the tests: all added IDs are supposed to be from the store")
+
+		if count == 1 {
+			require.True(t, foundID, "ID %v is supposed to be known, but isn't", id)
+
+			assert.True(t, s.WasQueued(), "ID %v was expected to be Queued and is %v", id, s.StatusString())
+			assert.False(t, s.WasReceived(), "ID %v was expected to be Queued and is %v", id, s.StatusString())
```
smnzhu, line 120: Any reason we use both `require` and `assert` here?

module/synchronization/core_rapid_test.go
```diff
@@ -77,35 +93,78 @@ func (r *rapidSync) HandleByID(t *rapid.T) {
...
+			activeBlocks = append(activeBlocks, *block)
+		} else if count == 0 {
+			if foundID {
+				// if a block is known with 0 pendings, it's because it was received
+				assert.True(t, s.WasReceived(), "ID %v was expected to be Received and is %v", id, s.StatusString())
 			}
+			receivedBlocks = append(receivedBlocks, *block)
```
smnzhu, line 127: should this go inside the if statement?

If a block has 0 pendings and is not known, does this signal an error in the testing logic?

module/synchronization/core_rapid_test.go
smnzhu: Or, did you mean to increment the value on each request and decrement it on each response?

module/synchronization/core_rapid_test.go
```diff
@@ -77,35 +93,78 @@ func (r *rapidSync) HandleByID(t *rapid.T) {
...
+				wasReceived := s.WasReceived()
+				_, blockAtHeightWasReceived := findHeader(receivedBlocks, func(header flow.Header) bool {
+					return header.Height == h
+				})
+				_, clobberedByID := findHeader(activeBlocks, func(header flow.Header) bool {
+					return header.Height == h
+				})
+				heightWasCanceled := wasReceived || blockAtHeightWasReceived || clobberedByID
+
+				assert.True(t, heightWasCanceled, "Height %x was expected to be Received (or filled through a same-height block) and is %v", h, s.StatusString())
```
smnzhu, line 159: Should we track the time of requests for blocks? Otherwise I think the "clobberedByID" part could give us a false positive. It could return true even if the request for a block at that height happened *before* the height reception.

But actually in this case, we want `heightWasCanceled` to be false.

module/synchronization/core_rapid_test.go
huitseeker: > Or, did you mean to increment the value on each request and decrement it on each response? 

Yep, that's the idea.

module/synchronization/core_rapid_test.go
huitseeker: the require is a basic sanity check and I want it to stop the test if it fails.

module/synchronization/core_rapid_test.go
huitseeker: > It could return true even if the request for a block at that height happened before the height reception.
> But actually in this case, we want heightWasCanceled to be false.

So, as this check runs on every single instruction, I'm not sure I see what you mean. If I first have a request for a block at height h, thereby clobbering all height requests at h, and then I receive at height h, then `wasReceived` becomes true.

Perhaps the adjective `canceled` is misleading, and should be replaced by `heightWasReceivedOrCanceled`?

PR #1401, opened 2021-09-30
[v0.22] [Storage Migration V6] Add more tests for value converter

SupunS:
Add test cases to make sure the values stay the same during the conversion during the migration from the old storage impl to the new storage impl.


ramtinms (approved):
Looks good to me, just maybe add some details to the PR description for future readers

janezpodhostnik (approved):
Shouldn't this be merged to v0.23-testnet?

SupunS:
> Shouldn't this be merged to v0.23-testnet?

ah yeah, totally missed it! Thank you for catching it! 🙏

PR #1404, opened 2021-10-01
[consensus voting V2] random beacon and threshold signature 

tarakby:
( this is a first implementation and I will revisit it again in 2 weeks to review safety of the APIs. Please have a look at the error types to insure the APIs are consistent with the integration )

On the Consensus side:
 - Implement `randomBeaconFollower` structure which implements the interface `RandomBeaconFollower` (new name of old `RandomBeaconSigner` since "signing" is not needed) 
 - add tests for `randomBeaconFollower`. 
 - update `RandomBeaconFollower` docs
 - ⚠️  update `Verify` error types: is signature is cryptographically invalid, update error from `crypto.InvalidInputsError` to `module/signature.ErrInvalidFormat`. `crypto.InvalidInputsError` is returned by the crypto package for many other input related errors, and re-using it here may lead to all errors being interpreted as invalid signature (this also matches the implementation of `WeightedSignatureAggregator`).
 - implementation of `randomBeaconFollower` is based on the new crypto interface `ThresholdSignatureFollower`. 
 - ⚠️ `RandomBeaconReconstructor` is not implemented yet : the module that works with Flow identities and uses  `RandomBeaconFollower` underneath ( `RandomBeaconFollower` is a very thin layer but I left it as per @zhangchiqing request)
 
 On the crypto package side:
 - Add new interfaces `ThresholdSignatureFollower` and `ThresholdSignatureParticipant` which are threshold signature generic interfaces.
 - Add new structure `blsThresholdSignatureFollower` and `blsThresholdSignatureParticipant` that implement the interfaces above using BLS.
 - Split the notion of "follower" (watching a protocol without participating) and "particpant" (watching a protocol and taking part in it) - difference is ability to sign. 
 - `ThresholdSignatureFollower` provides the APIs needed by hotstuff, with the concurrency safety.
 - `ThresholdSignatureParticipant` is not needed by Hotstuff. 
 - Add tests for `blsThresholdSignatureFollower` and `blsThresholdSignatureParticipant`. 
 - Clean up detection and usage of `InvalidError`.
 - Keep V1 threshold signature tools temporarily to run V1 tests.


crypto/bls_thresholdsign.go
```diff
@@ -0,0 +1,594 @@
...
+	// the threshold signature. It is equal to nil if less than (t+1) shares are
+	// received
+	thresholdSignature Signature
+	// lock for atomic operations
+	lock sync.RWMutex
+}
+
+type blsThresholdSignatureParticipant struct {
+	// embed the follower
+	*blsThresholdSignatureFollower
```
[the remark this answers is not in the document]
tarakby: The idea is to embed the type that implements the interface `ThresholdSignatureFollower`. This allows the embedder type `blsThresholdSignatureParticipant` to inherit all the methods needed by the interface `ThresholdSignatureFollower` without re-implementing them. 
In this case  `*blsThresholdSignatureFollower` implements the interface and not `blsThresholdSignatureFollower`. 

( Fyi `blsThresholdSignatureParticipant` is not used by Flow consensus )

crypto/bls_thresholdsign.go
```diff
@@ -0,0 +1,594 @@
...
+	// the index of the current node
+	currentIndex int
```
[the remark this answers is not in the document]
tarakby: Since I implemented DKG and threshold signature back in September 2019, I guess the general rule of using "me" didn't exist yet 😅 and I chose "current". I'll update to the more recent rule.

crypto/bls_thresholdsign.go
```diff
@@ -0,0 +1,594 @@
...
+
+func (e duplicatedSignerError) Error() string {
+	return e.message
+}
+
+// IsduplicatedSignerError checks if the input error is a duplicatedSignerError
+func IsduplicatedSignerError(err error) bool {
+	var target *duplicatedSignerError
+	return errors.As(err, &target)
+}
```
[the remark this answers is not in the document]
tarakby: Since the package covers many algorithms/protocols, and this error is related only to threshold signatures, I prefer that it remains close the threshold signature code. I have moved it near the interface file, instead of the implementation one.

crypto/bls_thresholdsign.go
```diff
@@ -0,0 +1,594 @@
...
+		return nil, errors.New("reading signatures has failed")
+	}
+
+	// Verify the computed signature
+	verif, err := s.VerifyThresholdSignature(thresholdSignature)
+	if err != nil {
+		return nil, fmt.Errorf("verify threshold signature failed: %w", err)
+	}
+	if !verif {
+		return nil, errors.New(
```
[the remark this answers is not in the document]
tarakby: Failing earlier only means that one of the signatures failed to deserialize to a valid curve point. 
Failing at this stage just means the reconstructed signature doesn't verify under the group key.

consensus/hotstuff/signature.go
```diff
@@ -29,7 +29,13 @@ type RandomBeaconReconstructor interface {
 	// Reconstruct reconstructs the group signature.
 	// The reconstructed signature is verified against the overall group public key and the message agreed upon.
 	// This is a sanity check that is necessary since "TrustedAdd" allows adding non-verified signatures.
-	// Reconstruct returns an error if the reconstructed signature fails the sanity verification, or if not enough shares have been collected.
+	// Returns:
+	// - (signature, nil) if no error occured
```
AlexHentschel, line 33:
```suggestion
	// - (signature, nil) if no error occurred
```

consensus/hotstuff/randombeacon_inspector.go
```diff
@@ -0,0 +1,50 @@
...
+	// a group signature. The function is thread-safe.
+	EnoughShares() bool
+
+	// Reconstruct reconstructs the group signature. The function is thread-safe but locks
+	// its internal state, thereby permitting only one routine at a time.
+	// The function errors (without sentinel) in any of the following cases:
+	//  - Not enough shares were collected.
+	//  - Any of the added signatures fails the deserialization.
+	//  - The reconstructed group signature is invalid. This post-verification is required
+	//	  for safety, as `TrustedAdd` allows adding invalid signatures.
```
AlexHentschel, line 48: we have this very nice [documentation for the implementation](https://github.com/onflow/flow-go/blob/96765647eae016b87ab641ac09614c958e876c36/consensus/hotstuff/signature/randombeacon_inspector.go#L104-L115) about the specific sentinel errors that this can return. Lets also add this documentation to the interface: 


```suggestion
	// Reconstruct reconstructs the group signature. The function is thread-safe but locks
	// its internal state, thereby permitting only one routine at a time.
	//
	// Returns:
	// - (signature, nil) if no error occurred
	// - (nil, crypto.notEnoughSharesError) if not enough shares were collected
	// - (nil, crypto.invalidInputsError) if at least one collected share does not serialize to a valid BLS signature,
	//    or if the constructed signature failed to verify against the group public key and stored message. This post-verification
	//    is required  for safety, as `TrustedAdd` allows adding invalid signatures.
	// - (nil, error) for any other unexpected error.
```

consensus/hotstuff/randombeacon_inspector.go
```diff
@@ -0,0 +1,50 @@
...
+	//  - (false, nil) if the signature has been added, but not enough shares were collected.
+	//  - (false, error) if there is any exception adding the signature share.
+	//      - engine.InvalidInputError if signerIndex is invalid
+	//  	- engine.DuplicatedEntryError if the signer has been already added
+	//      - other error if there is an unexpected exception.
+	TrustedAdd(signerIndex int, share crypto.Signature) (enoughshares bool, exception error)
+
+	// EnoughShares indicates whether enough shares have been accumulated in order to reconstruct
+	// a group signature. The function is thread-safe.
+	EnoughShares() bool
```
AlexHentschel, line 40: the interfaces `RandomBeaconInspector` and `RandomBeaconReconstructor` are quite symmetric. Except that we have different names for this method:
https://github.com/onflow/flow-go/blob/96765647eae016b87ab641ac09614c958e876c36/consensus/hotstuff/signature.go#L25-L27

Using the same names would make the code more consistent. I would suggest to use  `EnoughShares` everywhere

consensus/hotstuff/signature/randombeacon_inspector.go
```diff
@@ -0,0 +1,117 @@
...
+// its internal state, thereby permitting only one routine at a time.
+//
+// Returns:
+// - (signature, nil) if no error occured
+// - (nil, crypto.notEnoughSharesError) if not enough shares were collected
+// - (nil, crypto.invalidInputsError) if at least one collected share does not serialize to a valid BLS signature,
+//    or if the constructed signature failed to verify against the group public key and stored message. This post-verification
+//    is required  for safety, as `TrustedAdd` allows adding invalid signatures.
+// - (nil, error) for any other unexpected error.
+// The function is blocking.
```
AlexHentschel, line 114: I think this
> The function is blocking.

is redundant with the [statement above](https://github.com/onflow/flow-go/blob/96765647eae016b87ab641ac09614c958e876c36/consensus/hotstuff/signature/randombeacon_inspector.go#L104-L105)
> The function is thread-safe but locks its internal state, thereby permitting only one routine at a time.

```suggestion
```

consensus/hotstuff/signature/randombeacon_inspector.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/engine"
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/module/signature"
+)
+
+// randomBeaconInspector implements hotstuff.RandomBeaconInspector interface.
+// All methods of this structure are concurrency-safe.
+type randomBeaconInspector struct {
+	follower crypto.ThresholdSignatureInspector
```
AlexHentschel, line 15: I think it would be more consistent to call this `inspector`
```suggestion
	inspector crypto.ThresholdSignatureInspector
```

crypto/thresholdsign.go
```diff
@@ -13,40 +13,150 @@ import (
...
+	// This function is thread safe and locks the internal state.
+	//
+	// The share is only added if the signature is valid, the signer index is valid and has not been
+	// added yet. Moreover, the share is added only if not enough shares were collected.
+	// The function returns 3 outputs:
+	//  - First boolean output is true if the share is valid and no error is returned, and false otherwise.
+	//  - Second boolean output is true if enough shares were collected and no error is returned, and false otherwise.
+	//  - error is InvalidInputsError if input index is invalid, duplicatedSignerError if signer was added,
+	//	   and a random error if an exception occurred.
+	//    (an invalid signature is not considered an invalid input, look at `VerifyShare` for details)
```
AlexHentschel, line 78: I think we can increase readability here by converting the errors to a bullet point list:
```suggestion
	// Boolean returns:
	//  - First boolean output is true if the share is valid and no error is returned, and false otherwise.
	//  - Second boolean output is true if enough shares were collected and no error is returned, and false otherwise.
	// Error returns:
	//  - InvalidInputsError if input index is invalid
    //    (invalid signature will not cause an error here, as the signature is not checked yet)
	//  - duplicatedSignerError if signature from orig was previously added,
	// All other errors are unexpected during normal operation and a potential symptom of internal state corruption. 
```

crypto/bls_thresholdsign.go
```diff
@@ -0,0 +1,574 @@
...
+	// the valid signature shares received from other participants
+	shares map[index]Signature
+	// the threshold signature. It is equal to nil if less than (t+1) shares are
+	// received
+	thresholdSignature Signature
+	// lock for atomic operations
+	lock sync.RWMutex
+}
+
+// blsThresholdSignatureParticipant implements ThresholdSignatureParticipant
```
AlexHentschel, line 64: The `blsThresholdSignatureParticipant`'s feels somewhat mixed in with the `blsThresholdSignatureInspector`. Just reading over the code it makes me jump back and forth between the different objects. Maybe we could move the `blsThresholdSignatureInspector` to its own file? Or maybe have one first struct and its methods, then the other struct followed by its methods? Just a stylistic preference ... fee free to ignore.

AlexHentschel (approved):
Thanks for the clean code, and the great test coverage. 🚢 

Only a couple stylistic comments. Feel free to merge.

crypto/bls_thresholdsign_test.go
```diff
@@ -15,106 +16,280 @@ import (
...
 
-// simple single-threaded test of the stateful threshold signature using the simple key generation.
-// The test generates keys for a threshold signatures scheme, uses the keys to sign shares,
-// tests VerifyAndStageShare and CommitShare apis,
-// and reconstruct the threshold signatures using (t+1) random shares.
-func testStatefulThresholdSignatureSimpleKeyGen(t *testing.T) {
+var thresholdSignatureMessage = []byte("random message")
+
+// centralized test of the stateful threshold signature using the threshold key generation.
+func testCentralizedStatefulAPI(t *testing.T) {
```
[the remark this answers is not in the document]
tarakby: I have updated the tests under consensus/hotstuff to using testify/suite.
Tests in the crypto package do not use testify/suite but are modularized using `t.Run()`. I preferred to stay consistent within the crypto package for now and keep `t.Run`, unless I move to suite in all files.

consensus/hotstuff/signature/randombeacon_inspector.go
tarakby: Good point

crypto/bls_thresholdsign.go
tarakby: I did some rearrangement of functions/structs but I left the 2 structs in the same file as are tightly related.

`blsThresholdSignatureParticipant` is the main struct in the file (although flow consensus does not use it). and `blsThresholdSignatureInspector` is its main component. Both struct share most of the methods as a consequence of embedding.

tarakby:
bors merge

PR #1405, opened 2021-10-01
v0.22 merge master

zhangchiqing:


cmd/scaffold.go
```diff
@@ -704,7 +704,9 @@ func (fnb *FlowNodeBuilder) initFvmOptions() {
 		fvm.WithChain(fnb.RootChainID.Chain()),
 		fvm.WithBlocks(blockFinder),
 		fvm.WithAccountStorageLimit(true),
-		fvm.WithTransactionFeesEnabled(true),
```
zhangchiqing, line 707: I removed this line here, just to confirm this is correct @janezpodhostnik

janezpodhostnik: Yes this change is correct. it should be limited to `fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Canary || fnb.RootChainID == flow.Mainnet` like it is below.

PR #1407, opened 2021-10-01
[v0.22] [Storage Migration V6] Update missing/broken types of values

SupunS:
Updates broken type of values for testnet.


turbolent:
@zhangchiqing @Kay-Zee This was missed in the Testnet spork yesterday. Do you have a report of the migration, to see what was removed, so we can potentially restore the data in case we need to?

Even though it only addresses Testnet state issues, I think we should still merge it.

SupunS:
This PR is now outdated, and no longer has any impact as those broken values are removed already from the testnet. Therefore closing this.

PR #1410, opened 2021-10-01
Reenable Reporters

janezpodhostnik:
Reporters were commented out during the cadence storage migration.

This reenables the reporters and adds some new things:
- reporter output is the same across reporters and is abstracted away in preparation to switch the output to a database given a connection string
- the chain id is now a parameter of state extraction (so the logic can have different code paths according to the chain it was run on)
- validator interface was added that could be used to compare the state before and after migration
- reporters were made so they can run in multiple threads
- account reporter uses scripts to get certain information from the accounts
- reporters and migrations were split into a separate folders


cmd/util/ledger/migrations/register_reporter.go
```diff
@@ -0,0 +1,78 @@
+package migrations
```
janezpodhostnik, line 1: @ramtinms should I leave this one in?

ledger/ledger.go
```diff
@@ -296,7 +296,19 @@ func (v Value) MarshalJSON() ([]byte, error) {
 // Migration defines how to convert the given slice of input payloads into an slice of output payloads
 type Migration func(payloads []Payload) ([]Payload, error)
 
-// Reporter accepts slice ledger payloads and reports the state of the ledger
+// Reporter reports on data from the state
 type Reporter interface {
+	// Name returns the name of the reporter
+	Name() string
```
janezpodhostnik, line 302: Added a name for more convenient logging

ledger/complete/ledger.go
```diff
@@ -252,6 +253,10 @@ func (l *Ledger) ExportCheckpointAt(
 	// get trie
 	t, err := l.forest.GetTrie(ledger.RootHash(state))
 	if err != nil {
+		rh, _ := l.forest.MostRecentTouchedRootHash()
```
janezpodhostnik, line 256: I was thinking I would leave this in, so if you run state extraction with the wrong hash, you get the most recently used one.

cmd/util/ledger/migrations/register_reporter.go
ramtinms: we could take it out

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -74,27 +77,33 @@ func extractExecutionState(
 			storageFormatV6Migration.Migrate,
 			storageUsedUpdateMigration.Migrate,
 		}
+
+		// only add validators if migration is on
+		validators = []ledger.Validator{
+			// vlds.NewAccountBalanceValidator(log, chain),
```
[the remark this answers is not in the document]
ramtinms: Lets get rid of validators, they were not as easy as I was thinking they would be

cmd/util/ledger/migrations/storage_reporter.go
```diff
@@ -1,193 +0,0 @@
-package migrations
-
-// import (
```
ramtinms, line 3: we can remove deadcodes

ramtinms (approved):

PR #1412, opened 2021-10-02
Fix address conversion, fix storage read handler

turbolent:


cmd/util/ledger/migrations/storage_v6_test.go
```diff
@@ -787,3 +789,166 @@ func uint64ToBinary(integer uint64) []byte {
 	binary.BigEndian.PutUint64(b, integer)
 	return b
 }
+
+func Test(t *testing.T) {
```
SupunS, line 793:
```suggestion
func TestDeferredValues(t *testing.T) {
```

SupunS (approved):

cmd/util/ledger/migrations/storage_v6_test.go
```diff
@@ -787,3 +789,166 @@ func uint64ToBinary(integer uint64) []byte {
...
+		"storage\x1fMomentCollection\x1fownedNFTs\x1fv\x1f9061281":  "00cade0005d88484d8c082480b2a3299cc857e2967546f7053686f7402866464617461d88484d8c082480b2a3299cc857e2967546f7053686f74018666706c61794944d8a31903e96c73657269616c4e756d626572d8a3194a6a657365744944d8a3181a72546f7053686f742e4d6f6d656e7444617461626964d8a41a008a43a16475756964d8a41a01b2beca6b546f7053686f742e4e4654",
+		"public_key_3":                                              "f847b840dac22f8bdf651e21f10d5cc96ef8ca9a55bd28b56b117dda0877b3bce76ce7e04c4fe330c48e5c81f13f76c7dc1b86ab49a33d8b2498cbbed24459f7e1e918c50201808080",
+		"public_key_7":                                              "f847b840dac22f8bdf651e21f10d5cc96ef8ca9a55bd28b56b117dda0877b3bce76ce7e04c4fe330c48e5c81f13f76c7dc1b86ab49a33d8b2498cbbed24459f7e1e918c50201808080",
+		"public_key_1":                                              "f847b840dac22f8bdf651e21f10d5cc96ef8ca9a55bd28b56b117dda0877b3bce76ce7e04c4fe330c48e5c81f13f76c7dc1b86ab49a33d8b2498cbbed24459f7e1e918c50201808080",
```
ramtinms, line 824: we might only keep one of these keys and update public_key_count to 1, they don't play that much role in this test

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -333,7 +335,7 @@ func (m *StorageFormatV6Migration) getDeferredKeys(payloads []ledger.Payload) ma
 						pathSeparator,
 					)
 
-					deferredOwner := dictionary.DeferredOwner().Bytes()
+					deferredOwner := dictionary.DeferredOwner()[:]
```
ramtinms, line 338: so this was the error mostly for addresses which starts with some zeros?

ramtinms (approved):

ramtinms:
We might just update few other instances of `owner.Bytes()` in other migration tests in the same pr

janezpodhostnik (approved):

cmd/util/ledger/migrations/storage_v6.go
turbolent: Yes, the only way I found this problem was because in your test case data, the owner had an address with leading zeros.

Removed the function in https://github.com/onflow/cadence/pull/1162

cmd/util/ledger/migrations/storage_v6_test.go
turbolent: Given that it doesn't hurt, I think keeping this real-world data might be better

turbolent:
@ramtinms Good point, replaced in a51fe049e96ec01b723181494df8bcfc84897910

PR #1418, opened 2021-10-04
Purge chunk data packs for already sealed blocks

ramtinms:
on block finalized iterate over the seals and purge the chunk data packs for those sealed blocks. A future optimization might be to do removals using a single badger transaction.

### DIFF
--- a/engine/execution/ingestion/engine.go
+++ b/engine/execution/ingestion/engine.go
@@ -423,6 +423,32 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
 	}
 }
 
+// BlockFinalized handles the finalized blocks
+func (e *Engine) BlockFinalized(b *flow.Header) {
+	blockID := b.ID()
+	newBlock, err := e.blocks.ByID(blockID)
+	if err != nil {
+		e.log.Warn().Msgf("failed to handle block finalized: %s", err.Error())
+	}
+	// purge chunk data packs
+	e.purgeChunkDataPacks(context.Background(), newBlock)
+
+	e.log.Info().Hex("block_id", blockID[:]).
+		Uint64("height", b.Height).
+		Msg("handling block finalized")
+}
+
+func (e *Engine) purgeChunkDataPacks(ctx context.Context, block *flow.Block) {
+	seals := block.Payload.Seals
+	for _, seal := range seals {
+		err := e.execState.PurgeChunkDataPacksByBlockID(ctx, seal.BlockID)
+		// since this only purging operation its fine if it fails
+		if err != nil {
+			e.log.Warn().Msgf("failed to purge chunk data packs: %s", err.Error())
+		}
+	}
+}
+
 // Main handling
 
 // handle block will process the incoming block.
--- a/engine/execution/state/state.go
+++ b/engine/execution/state/state.go
@@ -43,6 +43,8 @@ type ReadOnlyExecutionState interface {
 	// ChunkDataPackByChunkID retrieve a chunk data pack given the chunk ID.
 	ChunkDataPackByChunkID(context.Context, flow.Identifier) (*flow.ChunkDataPack, error)
 
+	PurgeChunkDataPacksByBlockID(context.Context, flow.Identifier) error
+
 	GetExecutionResultID(context.Context, flow.Identifier) (flow.Identifier, error)
 
 	RetrieveStateDelta(context.Context, flow.Identifier) (*messages.ExecutionStateDelta, error)
@@ -314,6 +316,21 @@ func (s *state) ChunkDataPackByChunkID(ctx context.Context, chunkID flow.Identif
 	}
 
 	return chunkDataPack, nil
+}
+
+func (s *state) PurgeChunkDataPacksByBlockID(ctx context.Context, blockID flow.Identifier) error {
+	res, err := s.results.ByBlockID(blockID)
+	if err != nil {
+		return fmt.Errorf("could not fetch results for blockID: %w", err)
+	}
+
+	for _, chunk := range res.Chunks {
+		err = s.chunkDataPacks.Remove(chunk.ID())
+		if err != nil {
+			return fmt.Errorf("could not remove chunk data pack: %w", err)
+		}
+	}
+	return nil
 }
 
 func (s *state) GetExecutionResultID(ctx context.Context, blockID flow.Identifier) (flow.Identifier, error) {


engine/execution/ingestion/engine.go
```diff
@@ -423,6 +423,32 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
...
+	newBlock, err := e.blocks.ByID(blockID)
+	if err != nil {
+		e.log.Warn().Msgf("failed to handle block finalized: %s", err.Error())
+	}
+	// purge chunk data packs
+	e.purgeChunkDataPacks(context.Background(), newBlock)
+
+	e.log.Info().Hex("block_id", blockID[:]).
+		Uint64("height", b.Height).
+		Msg("handling block finalized")
```
zhangchiqing, line 438:
```suggestion
	e.log.Info().Hex("block_id", blockID[:]).
		Uint64("height", b.Height).
		Msg("purged chunk data pack for sealed blocks")
```

zhangchiqing (approved):

engine/execution/ingestion/engine.go
```diff
@@ -423,6 +423,32 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
...
+}
+
+func (e *Engine) purgeChunkDataPacks(ctx context.Context, block *flow.Block) {
+	seals := block.Payload.Seals
+	for _, seal := range seals {
+		err := e.execState.PurgeChunkDataPacksByBlockID(ctx, seal.BlockID)
+		// since this only purging operation its fine if it fails
+		if err != nil {
+			e.log.Warn().Msgf("failed to purge chunk data packs: %s", err.Error())
```
zhangchiqing, line 447:
```suggestion
			e.log.Warn().Msgf("failed to purge chunk data packs for block %v: %s", seal.BlockID, err.Error())
```

engine/execution/ingestion/engine.go
```diff
@@ -423,6 +423,32 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
 	}
 }
 
+// BlockFinalized handles the finalized blocks
+func (e *Engine) BlockFinalized(b *flow.Header) {
+	blockID := b.ID()
+	newBlock, err := e.blocks.ByID(blockID)
```
zhangchiqing, line 429: are we able to query seals only, rather than the entire block? Since we only need the seals here

engine/execution/ingestion/engine.go
```diff
@@ -423,6 +423,32 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
...
+	if err != nil {
+		e.log.Warn().Msgf("failed to handle block finalized: %s", err.Error())
+	}
+	// purge chunk data packs
+	e.purgeChunkDataPacks(context.Background(), newBlock)
+
+	e.log.Info().Hex("block_id", blockID[:]).
```
zhangchiqing, line 436:
```suggestion
	e.log.Info().Hex("finalized_block", blockID[:]).
```

engine/execution/ingestion/engine.go
ramtinms: I tried that initially but getting seals by blockId was broken and returns only one seal which I think it has a different meaning.

m4ksio:
I have certain reservation. Is it safe to remove CHDP as long as block is sealed? It makes sense from protocol point of view but I'm not sure if we are at this stage yet.
Secondly - using seals from finalized blocks won't purge CHDP for blocks which were not finalized - leaving orhpans.
Thirdly - maybe we should rather trim by height, leaving some configurable buffer, just in case. This will allow us to make this changes in batches and in parallel.
Fourthly - CHDP are cached, would be good to keep cache and DB consistent

ramtinms:
hmm, very fair points, not sure if we want to include a partially working version of v22 or wait until we implement it properly the way you suggested.

AlexHentschel (approved):
This is ok for a quick fix. 

#### Requested improvements, maybe as a follow-up PR:

1. The new code added does not follow best practices regarding error handling (see [coding conventions](https://github.com/onflow/flow-go/blob/alex/coding-guidelines/docs/Codingconventions.md)). Errors are bubbled up the call stack without any differentiation between expected and unexpected error cases and eventually logged. 
   * I agree that the code in this PR not a critical code path that would require to crash the node. Therefore, logging the error is fine. 
   * Nevertheless, there are still expected error cases (e.g. the chunk data pack has already previously been removed from storage) These errors do not have designated sentinel errors and therefore might create a lot of noise in our logs which will make it hard to identify critical errors, in case something goes wrong. 
   * I think this is a good occasion to clean up the error handling in the implementation of `ChunkDataPacks`:
        * add sentinel errors to the implementation
        * document sentinel errors in the interface
* `BlockFinalized` is a blocking operation, as it performs multiple data-base reads and writes.
   * the call to `BlockFinalized` is executed from [`Mutator`](https://github.com/onflow/flow-go/blob/0f54772a1af3dee646c2da58b7edd0221ef70d35/state/protocol/badger/mutator.go#L629), which is on the hot-path of following consensus. Adding a _synchronous_ call to a blocking, potentially expensive operation will halt the consensus follower!
   * Suggestion: 
       *  In general, implementations of `protocol.Consumer` should be non-blocking, unless it is absolutely required oterhwise. Freeing up storage does not need to be blocking. 
       * Overall, I think it would be most straight forward to purge chunk data packs by height of the executed block. as you are only considering finalized blocks, you don't have to worry about forks. 
       * You could take a [`FifoQueue`](https://github.com/onflow/flow-go/blob/0f54772a1af3dee646c2da58b7edd0221ef70d35/engine/common/fifoqueue/fifoqueue.go) queue and as part of the `protocol/Consumer.BlockFinalized` call, you just put the finalized block in the queue. Then, you do the purging of chunk data packs for finalized blocks _asynchronously_.  
       * One aspect you have to take into consideration when purging asynchronously: what happens when you crash with some chunk data packs still being not purged? You probably need to persist up to which height you have already purged.  
       * I don't understand why the `BlockFinalized` callback is part of the `ingestion.Engine`? Generally, the ingestion engine deals with external messages. block finalization is an internal event. To me, it would be more intuitive if the finalization would be part of the `ChunkDataPacks` storage (?) 
       * The original notification for block finalization is actually from the hotstuff follower: https://github.com/onflow/flow-go/blob/0f54772a1af3dee646c2da58b7edd0221ef70d35/consensus/hotstuff/consumer.go#L31. I would suggest to subscribe to this notification (also needs to be non-blocking)

engine/execution/ingestion/engine.go
AlexHentschel: I assume you are referring to this function: https://github.com/onflow/flow-go/blob/0f54772a1af3dee646c2da58b7edd0221ef70d35/storage/seals.go#L19-L20

I don't think it is broken. But it probably has a different meaning than what you are expecting: 
* `ByBlockID(B)` it does _not_ mean: give me the seal for block `B`.
* it returns you the latest seal that was included in the fork with head `B`. So for example, if you take the following fork:
    - ` <- B0 <- B1 <- B2 <- B3{seals B0} <- B4 <- B`
    - `ByBlockID(B)` will return the seal for B0, because `B0` is the latest sealed block in the fork with head `B`

zhangchiqing: Right, so we don't have a way to query all seals in the payload of a block. I think what @ramtinms has implemented here is correct.

ramtinms:
I had some issues fetching all blocks by view number, if other fixes work we might not merge this and postpone it till we have a proper solution for orphants.

PR #1422, opened 2021-10-04
Simplify address conversion

turbolent:
There is no need to convert from a Cadence address, an 8-byte array, to a short byte slice (leading zeros removed), then convert back to a Flow address, again an 8-byte array. Simply convert directly between the two arrays address types


janezpodhostnik (approved):

Kay-Zee:
bors merge

PR #1440, opened 2021-10-06
[Don't merge] Revert cadence version back to v0.19.1

ramtinms:


cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -1,3744 +1,3744 @@
...
-		if addressHex == "31aed847945124fd" {
-			address, err := hex.DecodeString("8d0e87b65159ae63")
-			if err != nil {
-				panic(err)
-			}
-
-			location = common.AddressLocation{
-				Address: common.BytesToAddress(address),
-				Name:    addressLocation.Name,
-			}
```
ramtinms, line 1284: @SupunS I assume if we don't fix these types we are still fine right if we use older cadence version?

ramtinms:
@jordanschalm I assume reverting the cadence version won't change anything about the new secrete storage and epoch changes? any concern here?

jordanschalm:
@ramtinms It should not impact secrets storage. The FlowEpoch contract change is not using any new Cadence features or anything - if this revert does not negatively impact smart contracts in general, I don't see why it would impact FlowEpoch specifically.

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -59,20 +59,20 @@ func extractExecutionState(
...
-			Log:       log,
-			OutputDir: outputDir,
-		}
+		// storageUsedUpdateMigration := mgr.StorageUsedUpdateMigration{
+		// 	Log:       log,
+		// 	OutputDir: outputDir,
+		// }
 
 		migrations = []ledger.Migration{
 			mgr.PruneMigration,
```
zhangchiqing, line 73: If there is no migration, do we still need to prune?

ramtinms: yeah pruning helps to prune nil values in the trie, so it reduces the checkpointer size and memory usage, we should always prune until we implement a service to prune regularly.

cmd/util/ledger/migrations/storage_v6.go
SupunS: Yes, that should be fine. These are still broken for a program even now (I think they are never accessed). We only needed to fix those so that the conversion from old to new storage format works and no values are dropped in the process.

ramtinms:
No need to merge this PR, v0.22-testnet already includes all changes from this PR.

PR #1462, opened 2021-10-07
[Consensus] add hotstuff view to ping route

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5920


zhangchiqing:
works on localnet

image

PR #1483, opened 2021-10-13
[State Sync Engine] state serialization module

smnzhu:
### Commit https://github.com/onflow/flow-go/pull/1483/commits/c7b930ae840c119c232f337d3b43a57a2e12028c

The entire test run for a ~360 MB state diff takes ~26 seconds, of which the most time consuming is the assertion at the very end to check that there is no discrepancy between the stored and loaded state diff.

The time needed to actually store / load the state diff to / from disk is minimal (~2.5 and ~1.5 seconds, respectively). Note however that the blockstore used in this test is a dummy blockstore which simply writes files directly to the filesystem. Performance may vary with the eventual production blockstore, which will probably be using Badger.

```bash
smnzhu@Simons-MacBook-Pro ~/Documents/GitHub/flow-go % ls -la ~/Downloads/8aeb01b08a446434ea4746aaaec1c458fd24922e26a13e94842bafb74e681670.cbor
-rw-r--r--@ 1 smnzhu  staff  359949597 Oct  8 11:55 
smnzhu@Simons-MacBook-Pro ~/Documents/GitHub/flow-go % sudo /usr/local/Cellar/go/1.16.5/libexec/bin/go test -v -timeout 30s -tags relic -run ^TestStateDiffStorer$ github.com/onflow/flow-go/engine/execution/state_synchronization
=== RUN   TestStateDiffStorer
    state_diff_storer_test.go:68: time to store state diff: 2.483088392s
    state_diff_storer_test.go:74: time to load state diff: 1.484808883s
--- PASS: TestStateDiffStorer (25.27s)
PASS
ok      github.com/onflow/flow-go/engine/execution/state_synchronization        25.713s
```

### Commit https://github.com/onflow/flow-go/pull/1483/commits/5e1ec7a40c8c56e419cef3930ab91311ad7fca33

Updated test to run over entire mainnet bucket. Was able to do every file successfully,

### DIFF
--- a/engine/execution/state_synchronization/state_diff_storer.go
+++ b/engine/execution/state_synchronization/state_diff_storer.go
@@ -1,0 +1,179 @@
+package state_synchronization
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+
+	"github.com/onflow/flow-go/ledger"
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/network"
+)
+
+const MAX_BLOCK_SIZE = 1e6
+
+type ExecutionStateDiff struct {
+	Collections        []*flow.Collection
+	Events             []*flow.Event
+	TrieUpdate         []*ledger.TrieUpdate
+	TransactionResults []*flow.TransactionResult
+}
+
+type StateDiffStorer struct {
+	blockWriter *BlockWriter
+	codec       encoding.Codec
+	compressor  network.Compressor
+}
+
+func NewStateDiffStorer(
+	codec encoding.Codec,
+	compressor network.Compressor,
+	bstore blockstore.Blockstore,
+) (*StateDiffStorer, error) {
+	bw := &BlockWriter{
+		maxBlockSize: MAX_BLOCK_SIZE,
+		bstore:       bstore,
+	}
+	return &StateDiffStorer{bw, codec, compressor}, nil
+}
+
+func (s *StateDiffStorer) writeBlocks(v interface{}) ([]cid.Cid, error) {
+	comp, err := s.compressor.NewWriter(s.blockWriter)
+	if err != nil {
+		return nil, err
+	}
+	enc := s.codec.NewEncoder(comp)
+	if err := enc.Encode(v); err != nil {
+		return nil, err
+	}
+	if err := comp.Close(); err != nil {
+		return nil, err
+	}
+	if err := s.blockWriter.Flush(); err != nil {
+		return nil, err
+	}
+	cids := s.blockWriter.GetWrittenCids()
+	s.blockWriter.Reset()
+	return cids, nil
+}
+
+func (s *StateDiffStorer) Store(sd *ExecutionStateDiff) (cid.Cid, error) {
+	if cids, err := s.writeBlocks(sd); err != nil {
+		return cid.Undef, fmt.Errorf("failed to write state diff blocks: %w", err)
+	} else {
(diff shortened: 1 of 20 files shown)


engine/execution/state_synchronization/state_diff_storer.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (s *StateDiffStorer) Load(c cid.Cid) (*ExecutionStateDiff, error) {
+	cids := []cid.Cid{c}
+	for {
+		var sd ExecutionStateDiff
+		if err := s.readBlocks(cids, &sd); err == nil {
+			return &sd, nil
+		} else {
+			if err := s.readBlocks(cids, &cids); err != nil {
```
peterargue, line 105: how common are state diffs >1MB? performing tons of these double read/decodes will waste CPU/IO resources on the ENs. if it's common enough, we might consider adding a header to the data so we can differentiate the type before attempting to decode it. that would probably require creating our own block type that includes the type as a separate field outside of the data slice since it shouldn't be included in the hash

module/chunks/chunk_assigner.go
```diff
@@ -168,7 +168,7 @@ func fingerPrint(blockID flow.Identifier, resultID flow.Identifier, alpha int) (
 	hasher := hash.NewSHA3_256()
 
 	// encodes alpha parameteer
```
peterargue, line 170:
```suggestion
	// encodes alpha parameter
```

engine/execution/state_synchronization/state_diff_storer_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	ds, err := datastore.NewDatastore(dsDir)
+	require.NoError(t, err)
+
+	return blockstore.NewBlockstore(ds.(*datastore.Datastore)), func() {
+		require.NoError(t, os.RemoveAll(dsDir))
+	}
+}
+
+const BUCKET_NAME = "flow_public_mainnet14_execution_state"
```
peterargue, line 41: This is ok for local testing (as currently intended), but we'll definitely want to avoid pulling lots of data from GCP for a ci test since that will add up on our GCP bill

peterargue (commented):
looking good. thanks for putting this together

peterargue (approved):

engine/execution/state_synchronization/state_diff_storer.go
smnzhu: They are very uncommon, more than 99% of the time the first if will succeed.

peterargue: if these are very uncommon, I'm fine punting on this as a future optimization

fxamacker: > Also, I haven't checked this but I would assume CBOR already adds headers to the data to differentiate the type. Perhaps @fxamacker can confirm this?

Hi @smnzhu,  yes, CBOR uses a generic "CBOR head" to identify CBOR data types ([major types](https://www.rfc-editor.org/rfc/rfc8949.html#name-major-types)) of encoded values.

In addition, the encoded CBOR head+value (data item) can be enclosed by a [CBOR tag](https://www.rfc-editor.org/rfc/rfc8949.html#name-tagging-of-items) to give the data item additional semantics.  For example, Cadence uses [CBOR tags to identify Cadence-specific data types](https://github.com/onflow/cadence/blob/master/runtime/interpreter/encode.go#L95-L207).

CBOR tags can be used here to indicate Flow-specific types, but it will affect data size (typically a few bytes) and possibly performance.

smnzhu: 2ad73ab @peterargue Implemented a generic solution for now to support all encoding implementations, but eventually we will probably settle on a single encoding type (probably CBOR), and can use native solution such as CBOR tag, as mentioned above.

smnzhu:
bors merge

smnzhu:
bors merge

PR #1499, opened 2021-10-20
[Consensus Follower] Handle fatal errors

peterargue:
# This PR is being replaced by another that breaks the changes into smaller chunks for easier review.

Throughout the codebase irrecoverable errors are handled by calling zerolog `Fatal()` which then calls `os.Exit(1)`. This is ok when running the code as a node, since it can just restart. However, this is awkward when running code as a library, like with the consensus follower.

This PR updates the consensus follower library to make use of the `irrecoverable` methods to handle irrecoverable errors where possible by restarting the component. It also converts errors that previously called `Fatal()` to use the `Throw` method to propagate the error up so the caller can restart the library without crashing their application.

### DIFF
--- a/cmd/access/main.go
+++ b/cmd/access/main.go
@@ -24,5 +24,8 @@ func main() {
 		anb.Logger.Fatal().Err(err).Send()
 	}
 
-	nodeBuilder.Build().Run()
+	nodeBuilder.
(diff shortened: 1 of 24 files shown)


peterargue:
closing in favor of #1543 and #1559

PR #1501, opened 2021-10-20
[v23] Improve ledger write speed for newly allocated slabs

ramtinms:
For newly allocated slabs, if we write the nil value to delta this would make a batch update at the end of the transaction faster since we don't need to lookup the trie for nil values. we also need to update the storage used to reflect the new keys added.


janezpodhostnik (approved):
LGTM! Any chance we can see the performance improvement with: https://github.com/onflow/flow-go/pull/1579

turbolent (commented):
I can't really review this, as I'm missing the background information on storage usage calculation. 

Looks good from a pure Go perspective

fvm/state/accounts.go
```diff
@@ -82,6 +82,24 @@ func (a *StatefulAccounts) AllocateStorageIndex(address flow.Address) (atree.Sto
...
+	err = a.stateHolder.State().Set(string(address.Bytes()), "", key, []byte{}, false)
+	if err != nil {
+		return atree.StorageIndex{}, fmt.Errorf("failed to store empty value for newly allocated storage index: %w", err)
+	}
+	// update the storage size considering new keys has been allocated with empty value
+	oldSize, err := a.GetStorageUsed(address)
+	if err != nil {
+		return atree.StorageIndex{}, fmt.Errorf("failed to store empty value for newly allocated storage index: %w", err)
+	}
+	a.setStorageUsed(address, oldSize+uint64(RegisterSize(address, false, key, []byte{})))
```
turbolent, line 100: Can the two byte slice allocations, `[]byte{}`, be just replaced with `nil`?

ramtinms: I just noticed this won't work since RegisterSize would ignore the size if the value size is empty, which is not really great cc: @janezpodhostnik

ramtinms: so maybe we don't even need to set the storage used for nil values :-?

fvm/state/accounts.go
```diff
@@ -82,6 +82,26 @@ func (a *StatefulAccounts) AllocateStorageIndex(address flow.Address) (atree.Sto
 	copy(index[:], indexBytes[:8])
 	newIndexBytes := index.Next()
 
+	// store nil so that the setValue for new allocated slabs would be faster
+	// and won't do ledger getValue for every new slabs (currently happening to compute storage size changes)
+	// this way the getValue would load this value from deltas
+	// TODO: update this code to use atree.SlabIndexToLedgerKey method when the new version of atree is used
+	key := atree.LedgerBaseStorageSlabPrefix + string(index[:])
+
+	err = a.stateHolder.State().Set(string(address.Bytes()), "", key, []byte{}, false)
```
Kay-Zee, line 91: is this the right way to convert to string here?

Kay-Zee: noticed it's used like this everywhere else, so i guess so

Kay-Zee:
Looks like the performance improvement is 2-5%!

Kay-Zee (approved):

fvm/state/accounts.go
ramtinms: I think this one is fine, the cadence common.Address Bytes() method was problematic in the past

PR #1509, opened 2021-10-21
Update golangci, trying to fix lint

Kay-Zee:
This updates the golangci action to just use the latest v2 action, which is recommended for the action.

Also, skipping the PKG cache because of all the File Exists error. Will have to follow: https://github.com/golangci/golangci-lint-action/issues/244 to see when this is resolved. (though lint is one of the faster jobs, so it's fine even if we skip the cache)

Lastly, I'm suggesting we skip the `unused` lint for TEST FILES ONLY, as there are many cases where we want to use `t.Skip()` to quarantine, but then it causes "unused" errors. This will resolve those cases and better allow us to use `t.Skip()`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
         # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
         version: v1.41
         args: -v --build-tags relic
+        # https://github.com/golangci/golangci-lint-action/issues/244
         skip-pkg-cache: true
     - name: Run ShellCheck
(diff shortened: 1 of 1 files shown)


Kay-Zee:
> I noticed the linter in makefile is not used by the linter job in CI. This means developers might get different results locally using `make` and on CI. Isn't better to make CI use `make lint` instead of the action?

The CI action is nice to have, as it gives more features, and allows slightly more control (also, the golangci team obviously suggests using golangci action instead of your own)

They do use the same config file, so should return the same results

Kay-Zee:
bors merge

PR #1511, opened 2021-10-21
performance: improvement to Union

jwinkler2083233:
This will allow the Union() method to avoid expensive memory
allocations.


jwinkler2083233:
closing because much of this was covered in a different commit, regarding IdentityList.

PR #1513, opened 2021-10-21
performance: replace calls to Lookup with binsearch

jwinkler2083233:


jwinkler2083233:
Additional data from tests, including performance comparisons with hashmap and memory profile results, are in issue #5955.

model/flow/identity.go
```diff
@@ -508,3 +561,99 @@ func (il IdentityList) EqualTo(other IdentityList) bool {
...
+		if left > right {
+			return false
+		}
+		mid = int(uint(left+right) >> 1)
+	}
+}
+
+// Exists takes a previously sorted Identity list and searches it for the target value
+// target:  value to search for
+func (il IdentityList) IdentifierExists(target Identifier) bool {
```
jordanschalm, line 616: In general I'm a bit worried about accidental misuse of these methods which require the receiver identity list to be ordered. It isn't obvious to me from the function name or receiver type that the receiver needs to be sorted. Could we do 1 or 1+2 of these:

1. Make the requirement a bit "louder" in the function documentation, and be explicit about the expected sort order:
```
// documentation ...
//
// CAUTION: IdentityList receiver must be previously sorted lexicographically by node ID in order to use this function.
// <link to function which sorts an identity list to the required order>
func ...
```

2. Introduce a type representing a sorted list that must be used when we require a receiver to be sorted already.
```go
type SortedIdentityList IdentityList

func (il IdentityList) Sort() SortedIdentityList { ... }

func (il SortedIdentityList) IdentifierExists(nodeID Identifier) bool { ... }
// other functions which require a sorted receiver
```

jordanschalm (commented):
👍 Nice to improve efficiency on these frequently used methods. 

I added a few suggestions, in particular the accidental misuse point I'd like to resolve before merging.

model/flow/identity_test.go
```diff
@@ -70,6 +72,74 @@ func TestIdentityEncodingMsgpack(t *testing.T) {
...
+					return true
+				} else if chunk1 == chunk2 {
+					continue
+				} else {
+					return false
+				}
+			}
+			// equality
+			return true
+		})
```
zhangchiqing, line 101: can't we just call `il1.Sort()`?

or `Union` with a empty identity list to reuse the sort functionality in it?

model/flow/identity.go
```diff
@@ -477,19 +488,61 @@ func (il IdentityList) SamplePct(pct float64) IdentityList {
 // either `il`, or `other`, or both. There are no duplicates in the output,
 // where duplicates are identities with the same node ID.
```
zhangchiqing, line 489:
```suggestion
// where duplicates are identities with the same node ID.
// The returned IdentityList is sorted 
```

model/flow/identity_test.go
```diff
@@ -70,6 +72,74 @@ func TestIdentityEncodingMsgpack(t *testing.T) {
...
+		element := il1[8]
```
zhangchiqing, line 102: instead of picking one element, it's better to ensure that every element exists.

I would suggest to make the test cases bigger:
create a identity list with 500 identities (which is close to mainnet), go through each identity and verify they exist

model/flow/identity_test.go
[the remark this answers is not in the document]
jwinkler2083233: I found that if the sort logic is kept in a separate method, the performance is drastically reduced.  It really is necessary in most cases to put the sort logic inline, because of how Go wraps every function with additional features.

model/flow/identity.go
jwinkler2083233: will do

model/flow/identity_test.go
jwinkler2083233: You're right.  Since it's in a test, performance of the sort is less important than concise code.

model/flow/identity_test.go
jwinkler2083233: will do

model/flow/identity.go
```diff
@@ -508,3 +561,99 @@ func (il IdentityList) EqualTo(other IdentityList) bool {
 	}
 	return true
 }
+
+// Exists takes a previously sorted Identity list and searches it for the target value
+// This code is optimized, so the coding style will be different
+// target:  value to search for
+func (il IdentityList) Exists(target *Identity) bool {
+	left := 0
+	lenList := len(il)
```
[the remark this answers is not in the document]
jwinkler2083233: There is a little performance hit, but I agree that it looks much better.  will do

model/flow/identity.go
```diff
@@ -508,3 +562,59 @@ func (il IdentityList) EqualTo(other IdentityList) bool {
...
+				return true
+			}
+
+			// these 8 bytes were equal, so increment index by 8 bytes
+			i += 8
+		}
+		if left > right {
+			return false
+		}
+		mid = int(uint(left+right) >> 1)
```
AlexHentschel, line 618: Do we really need the conversion `int` -> `uint` -> `int` (?) My understanding that the type conversion is not needed for non-negative `int` values.  `left` and `right` are always non-negative. I think a comment here might be sufficient.

same for https://github.com/onflow/flow-go/blob/9120a897c1936819ee321431377d5e0dc4fba3a4/model/flow/identity.go#L581 (provided you include the short cut for zero-length `IdentityList`)

model/flow/identity.go
```diff
@@ -508,3 +562,59 @@ func (il IdentityList) EqualTo(other IdentityList) bool {
...
+// target:  value to search for
+// CAUTION:  The identity list MUST be sorted prior to calling this method
+func (il IdentityList) Exists(target *Identity) bool {
+	return il.IdentifierExists(target.NodeID)
+}
+
+// Exists takes a previously sorted Identity list and searches it for the target value
+// target:  value to search for
+// CAUTION:  The identity list MUST be sorted prior to calling this method
+func (il IdentityList) IdentifierExists(target Identifier) bool {
```
AlexHentschel, line 577: The implementation here is non-trivial. Are we sure that `bytes.Compare` is more expensive? I feel that using `bytes.Compare` would result in a much more concise implementation; it that is possible (?)

model/verification/chunkDataPackRequest_test.go
```diff
@@ -43,6 +46,24 @@ func TestChunkDataPackRequestList_UniqueRequestInfo(t *testing.T) {
...
+		return bytes.Compare(thisChunkIDReqInfo.Disagrees[p][:], thisChunkIDReqInfo.Disagrees[q][:]) < 0
+	})
+
+	thisChunkIDReqInfo.Targets = thisChunkIDReqInfo.Targets.Sort(order.Canonical)
+
+	/*
+		sort.Slice(thisChunkIDReqInfo.Targets, func(p, q int) bool {
+			return bytes.Compare(*(thisChunkIDReqInfo.Targets[p])[:], *(thisChunkIDReqInfo.Targets[q])[:]) < 0
+		})
+	*/
```
AlexHentschel, line 65: can we remove this dead code ?

state/protocol/badger/snapshot.go
```diff
@@ -163,7 +163,9 @@ func (s *Snapshot) Identities(selector flow.IdentityFilter) (flow.IdentityList,
 
 	// get identities from the current epoch first
 	identities := setup.Participants.Copy()
-	lookup := identities.Lookup()
+
+	// sort the identities so the 'Exists' binary search works
+	identities = identities.Sort(order.ByNodeIDAsc)
```
AlexHentschel, line 168: I think we can completely avoid the `Copy` here, as `Sort` already copies to not modify the original `IdentityList`
```golang
identities := setup.Participants.Sort(order.ByNodeIDAsc) // Sort copies internally 
```

AlexHentschel (changes requested):
Looks great. Thanks for the optimizations. I think there is one edge case where `Union` erroneously modifies  the input.

model/flow/identity.go
```diff
@@ -476,20 +487,63 @@ func (il IdentityList) SamplePct(pct float64) IdentityList {
...
+
+			if chunk1 < chunk2 {
+				return true
+			} else if chunk1 > chunk2 {
+				return false
+			} else if i >= lenID-8 {
+				// we're on the last chunk of 8 bytes, the nodeid's are equal
+				return false
+			}
+		}
```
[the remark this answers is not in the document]
jwinkler2083233: I thought the same, originally.
Then I looked at the output in assembly for both styles, and the BigEndian package optimizes the comparison by converting 8 bytes at a time to an integer, and the compiler optimizes it well.  Tests show it to be much faster.

model/flow/identity.go
```diff
@@ -476,20 +487,63 @@ func (il IdentityList) SamplePct(pct float64) IdentityList {
...
-	// add all identities, omitted duplicates
-	for _, identity := range append(il.Copy(), other...) {
-		if _, exists := lookup[identity.NodeID]; exists {
-			continue
+	if (len(il) + len(other)) == 0 {
+		return IdentityList{}
+	}
+
+	// add all identities together
+	union := append(other, il...)
```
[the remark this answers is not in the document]
jwinkler2083233: That would kind of nullify the reason for this change...  We're seeing excessive garbage collection activity, and this is one of the top contributors.

As an alternative, how about we require the 'other IdentityList' to be 'other *IdentityList' so that the developers see that there is something different, and that the value is mutable, plus add extensive comments ?  I checked the code that calls Union, and the calling spots are discarding 'other' after it's used.

model/flow/identity.go
```diff
@@ -508,3 +562,59 @@ func (il IdentityList) EqualTo(other IdentityList) bool {
...
+}
+
+// Exists takes a previously sorted Identity list and searches it for the target value
+// target:  value to search for
+// CAUTION:  The identity list MUST be sorted prior to calling this method
+func (il IdentityList) IdentifierExists(target Identifier) bool {
+	left := 0
+	lenList := len(il)
+	right := lenList - 1
+	mid := int(uint(right) >> 1)
```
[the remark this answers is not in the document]
jwinkler2083233: yes, will do.

model/flow/identity.go
jwinkler2083233: It's a high-use area.  If it weren't performance-sensitive, then bytes.Compare() would be ok.  I would vote for not using bytes.Compare() for that reason.

state/protocol/badger/snapshot.go
jwinkler2083233: will do

model/flow/identity.go
jwinkler2083233: will do

model/verification/chunkDataPackRequest_test.go
jwinkler2083233: done

model/flow/identity.go
jwinkler2083233: done

model/verification/chunkDataPackRequest_test.go
jwinkler2083233: done

state/protocol/badger/snapshot.go
jwinkler2083233: done

model/flow/identity.go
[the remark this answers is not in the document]
AlexHentschel: Overall, we prioritize safety, for the protocol as well as the implementation. Making some performance compromises to reduce the possibility to miss-use a function is the preferred choice. 

Instead of modifying the default implementation to be unsafe, and provide a safer alternative, you could do it the other way around:
* `IdentityList.Union` is safe by not modifying the inputs
* For the few performance-critical places in the code, you could provide an unsafe alternative which works in-place and clearly distinguish it by name, i.e. `IdentityList.UnsafeUnionInPlace`. Thereby developers will chose the by default the safe option. 

Regarding your argument that copying would defeat the purpose of the optimization, I am not really convinced: 
* Lets look at your second example https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L81 specifically the place in the code where `topicFanout` is created: https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L77
*  going into the function `subsetChannel`, you'll find that there are a lot of `Filter` function applied ([here](https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L98, [here](https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L137-L138), ...) and finally we apply the [`sampleFanout`](https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L112) function. What both `sampleFanout` and  `Filter` have in common is that (for safety reasons), they return a new slice: https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/randomizedTopology.go#L117 https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/model/flow/identity.go#L290
* After closer inspection of the mechanics of these function, we learn that these function will already copy the content multiple times (internally via the append function). 

So in summary, there are many re-allocations through the `sampleFanout` and  `Filter` functions anyway. For the sake of argument, lets just assume that these function re-allocate in total 10 new arrays. The safe implementation of `Union` does a _single_ new allocation (we specifically avoid append). I really don't see why removing this last allocation in `Union` would make a big difference.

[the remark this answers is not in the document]
AlexHentschel: BTW, here is the reason why I am not ok with in-place modifications for the `IdentityList` methods (except for explicitly distinguished unsafe alternatives): 
* Here we have the identity lists for the current Epoch: https://github.com/onflow/flow-go/blob/03cba83741a4e43e34f64761147f98728480b1ff/network/topology/topology_utils.go#L55 All Identity filters ultimately use this list as their basis. If you happen to write through to this array (or modify the code such that some other place in the code will modify this list), it is game over. If that happens only once, you will have completely broken Flow:
  * then, the nodes might reject legitimate messages from other nodes  
  * or (oven worse) they might accept messages from malicious nodes that are not authorized to send such messages 

This is not a risk I am willing to take. Flow is already complex enough. We can't have engineers carefully work around subtle unsafe implementations in addition. APIs need to be as safe as possible. Making broadly-used components unsafe, because it benefits a few specific spots in the code is not an option. If you need highly performant implementations at the cost of API safety, please provide a special-purpose implementation that is clearly distinguished as unsafe.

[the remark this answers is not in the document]
jwinkler2083233: I agree that it's ugly.  Most performance optimizations are.

Seniority wins.  I'll shelve the Union performance fix.

model/flow/identity.go
AlexHentschel: 👍  
Thanks [your reply](https://github.com/onflow/flow-go/pull/1513#discussion_r759749823) to my previous comment was really helpful.

AlexHentschel (approved):
thanks @jwinkler2083233 for the revisions. Looks great. Really appreciate you inspecting the assembly code and benchmarking different implementations.

* Regarding the unsafe union implementation, I din't mean to say you shouldn't do it at all. Sorry if it came across this way. While it is important to me that the _default_ implementation is as safe as possible, you certainly have a point that a more efficient and less safe implementation is possible and useful. If you think there are some code paths impacted by the default implementation adding GC overhead, think it would be useful to add a _second_ implementation `IdentityList.UnsafeUnionInPlace` which utilizes all the performance improvements but needs to used with care.

## I think we can remove the following allocation:

https://github.com/onflow/flow-go/blob/b10a0bb63f9acca82eab7d1a7871778e0fb3fbd3/model/flow/identity.go#L531

I have put up the PR https://github.com/onflow/flow-go/pull/1729 (targeting your branch) with an implementation that removes this allocation. Please let me know what you think.

jwinkler2083233:
> thanks @jwinkler2083233 for the revisions. Looks great. Really appreciate you inspecting the assembly code and benchmarking different implementations.
> 
> * Regarding the unsafe union implementation, I din't mean to say you shouldn't do it at all. Sorry if it came across this way. While it is important to me that the _default_ implementation is as safe as possible, you certainly have a point that a more efficient and less safe implementation is possible and useful. If you think there are some code paths impacted by the default implementation adding GC overhead, think it would be useful to add a _second_ implementation `IdentityList.UnsafeUnionInPlace` which utilizes all the performance improvements but needs to used with care.
> 
> ### I think we can remove the following allocation:
> https://github.com/onflow/flow-go/blob/b10a0bb63f9acca82eab7d1a7871778e0fb3fbd3/model/flow/identity.go#L531
> 
> I have put up the PR #1729 (targeting your branch) with an implementation that removes this allocation. Please let me know what you think.

looks good

AlexHentschel (approved):
Thanks again. Very solid code now.

state/protocol/badger/snapshot.go
```diff
@@ -161,9 +161,8 @@ func (s *Snapshot) Identities(selector flow.IdentityFilter) (flow.IdentityList,
 		return nil, err
 	}
 
-	// get identities from the current epoch first
-	identities := setup.Participants.Copy()
-	lookup := identities.Lookup()
+	// sort the identities so the 'Exists' binary search works
+	identities := setup.Participants.Sort(order.ByNodeIDAsc)
```
Kay-Zee, line 165: Does this list already come "semi-sorted" on subsequent calls, or does it just get saved un-sorted, and we have to do the same sort every time?

model/flow/identity_test.go
```diff
@@ -70,8 +70,52 @@ func TestIdentityEncodingMsgpack(t *testing.T) {
...
+		// Here, we purposefully create a slice, whose backing array has enough space to
+		// include the elements we are going to add via Union. Furthermore, we force element duplication
+		// by taking the union of the IdentityList with itself. If the implementation is not careful
+		// about creating copies and works with the slices itself, it will modify the input and fail the test.
+
+		il := unittest.IdentityListFixture(20)
+		il = il[:10]
+		ilBackup := il.Copy()
+
+		_ = il.Union(il)
```
Kay-Zee, line 116: Why don't we check the output here, since we know it's expected contents, i.e. same as `il` and `ilBackup`

Kay-Zee (approved):
Discussed with @jwinkler2083233 offline about some some of the timings for searches with this implementation, but for our use case (500) it seems the main gain is the reduced heap allocations.

From that lense, :+1: from me

jwinkler2083233:
I did one more test and everything appears steady.  TPS numbers look reasonably better with fewer sequence errors.

PR #1523, opened 2021-10-25
hotstuff: improve the latency of hotstuff verification

jwinkler2083233:
This will remove the requirement for a mutex, and just use
atomic instructions instead.


consensus/hotstuff/verification/common.go
```diff
@@ -68,24 +69,42 @@ func checkVotesValidity(votes []*model.Vote) error {
...
+}
+
 // stakingKeysAggregator is a structure that aggregates the staking
 // public keys for QC verifications.
 type stakingKeysAggregator struct {
-	lastStakingSigners map[flow.Identifier]*flow.Identity
-	lastStakingKey     crypto.PublicKey
-	sync.RWMutex
+	current unsafe.Pointer // *aggregate type
+	inner   *aggregate
```
bluesign, line 81: I think having `inner` here later can cause some complications. ( confusing )

consensus/hotstuff/verification/common.go
```diff
@@ -68,24 +69,42 @@ func checkVotesValidity(votes []*model.Vote) error {
...
-	aggregator := &stakingKeysAggregator{
+	aggregator := &stakingKeysAggregator{}
+
+	aggregator.inner = &aggregate{
 		lastStakingSigners: map[flow.Identifier]*flow.Identity{},
 		lastStakingKey:     NeutralBLSPublicKey(),
-		RWMutex:            sync.RWMutex{},
 	}
+
+	aggregator.current = unsafe.Pointer(aggregator.inner)
```
bluesign, line 93: I think `updateCurrent` would be better choice here, without inner

bluesign:
I think if this mutex is causing performance problems, it is just a symptom not the cause of the problem. 

But I really liked the update, avoiding mutex like this is also very cool.

consensus/hotstuff/verification/common.go
jwinkler2083233: I think I tried that and there was a problem in the tests.  If you want to paste your code change suggestion, I'll try it.

consensus/hotstuff/verification/common.go
jwinkler2083233: Maybe a name change or comment would help?  What would you like to see ?

AlexHentschel:
Took a brief pass over the PR and it looks great. However **concurrency safety probably warrants some detailed comments in the code is**:
* note that according to the [Go memory model](https://go.dev/ref/mem#tmp_2), atomic reads and writes do _not_ establish a happens-before relationship 
   * we have discussed this in more detail in [slack thread](https://axiomzen.slack.com/archives/CEXKUH74G/p1625249898389600)
   * Most relevant reference is [this one](https://github.com/golang/go/issues/5045)
     > Go's atomics guarantee sequential consistency among the atomic variables (behave like C/C++'s seqconst atomics). You shouldn't mix atomic and non-atomic accesses for a given memory word, unless some other full memory barrier, like a Mutex, guarantees exclusive access. You shouldn't mix atomics of different memory sizes for the same address.
* long story short, synchronization guarantees in Go for atomic variables and subsequent non-atomic memory writes are far from clearly specified. 

In your particular implementation, a reviewer might ask: "There is no concurrency protection to `updatedSignerSet` and go's atomics don't establish a happens-before relationship. How is concurrency safety guaranteed here?" (previously we had a mutex, which establishes a happens-before relationship and thereby guarantees concurrency safety). 
From my perspective, the answer is that `aggregate` and all data fields therein (including `updatedSignerSet` and `updatedKey`) are essentially immutable. A new data structure is allocated for each write. Therefore, it is sufficient to guarantee sequential consistency by using an atomic variable. But we don't have to take any additional measures to mediate mutations of  `aggregate` as such mutations don't happen within the implementation. 

Hope that makes sense (?)

jwinkler2083233:
I'm closing this for now, because tests are inconclusive

PR #1524, opened 2021-10-25
REST API implementation - GetBlockByID

vishalchangrani:
closes: [5964](https://github.com/dapperlabs/flow-go/issues/5964)
This PR does the following:
1. Adds the [swagger-codegen](https://github.com/swagger-api/swagger-codegen) generated code for the Access node Swagger API under `engine/access/swagger/generated/`
2. Starts up an HTTP Server on port 8080 to serve the REST API.
3. Implements the `GetBlockByID` call (yet to implement all the other ones).
4. Adds a unit test to test the end-to-end request-response for the `GetBlockByID` call.

Companion PR on onflow/flow - https://github.com/onflow/flow/pull/676.

onflow/flow repo - contains the [swagger definition](https://github.com/onflow/flow/blob/master/openapi/access.yaml) and the generated client code.
onflow/flow-go - contains the generated server code.

_There are other [alternatives](https://github.com/onflow/flow/blob/master/openapi/access.yaml) to swagger-codegen but I stuck with swagger-codegen. The other ones do have more features but they also introduce extra code bloat. IMHO, the OpenAPI definition will not change frequently and the generated code should just act as a thin shim._


engine/access/rest_api_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (suite *RestAPITestSuite) TestRestAPICall() {
+
+	suite.Run("happy path - REST client successfully executed the GetBlockByID request", func() {
+
+		block := unittest.BlockFixture()
+		suite.blocks.On("ByID", block.ID()).Return(&block, nil).Once()
+
+		client := suite.restAPIClient()
+		ctx, cancel := context.WithTimeout(context.Background(), time.Hour*5)
```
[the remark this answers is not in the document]
vishalchangrani: oops..fixing this

engine/access/rpc/engine.go
```diff
@@ -307,3 +326,27 @@ func (e *Engine) serveGRPCWebProxy() {
 		e.log.Err(err).Msg("failed to start the http proxy server")
 	}
 }
+
+// serveREST starts the HTTP REST server
+func (e *Engine) serveREST() {
```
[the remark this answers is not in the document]
vishalchangrani: I do agree that REST API should not bring down the AN (and am changing the log level accordingly for the error) but making it an engine by itself might be an overkill since the backend that it uses is the common backed that all severs use. But if you feel strongly about it, I can issue it up and investigate making it an engine.

engine/access/swagger/generated/api_blocks.go
```diff
@@ -0,0 +1,23 @@
...
+ */
+package generated
+
+import (
+	"net/http"
+)
+
+func BlocksGet(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
vishalchangrani: that is an excellent suggestion. Also, this generated code is actually not being used. Instead, the `engine/access/swagger/rest_api.go` file provides the actual API implementation. The generated API implementation didn't provide an easy way to add the actual implementation plus I didn't want to change the generated code in case has to be regenerated.

engine/access/rest/converter.go
```diff
@@ -0,0 +1,69 @@
...
+
+func collectionGuarantee(flowCollGuarantee *flow.CollectionGuarantee) generated.CollectionGuarantee {
+	signerIDs := make([]string, len(flowCollGuarantee.SignerIDs))
+	for i, signerID := range flowCollGuarantee.SignerIDs {
+		signerIDs[i] = signerID.String()
+	}
+	return generated.CollectionGuarantee{
+		CollectionId: flowCollGuarantee.CollectionID.String(),
+		SignerIds:    signerIDs,
+		Signature:    flowCollGuarantee.Signature.String(),
```
smnzhu, line 52: This is incorrect, the golang implementation returns hex string: https://github.com/onflow/flow-go/blob/e533ad132d0ebe8af14315bd758f9061d2299fcc/crypto/sign.go#L130

But the API spec specifies that it should be base64: https://github.com/onflow/flow/blob/eeac4bb7b9a4a92fb6b9219bb1d5dba27b811e01/openapi/access.yaml#L803

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+
+	ids := strings.Split(idParam, ",")
+
+	blocks := make([]*generated.Block, len(ids))
+
+	for i, id := range ids {
+		flowID, err := flow.HexStringToIdentifier(id)
+		if err != nil {
+			restAPI.errorResponse(w, r, err)
+		}
```
smnzhu, line 51: Here It would not be appropriate to return a generic `StatusInternalServerError`, because the error is on the user side (they provided a malformed Hex ID).

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+
+		flowBlock, err := restAPI.backend.GetBlockByID(r.Context(), flowID)
+		if err != nil {
+			restAPI.errorResponse(w, r, err)
+		}
```
smnzhu, line 56: Here it also may not be appropriate to return a generic `StatusInternalServerError`, because the error could be that the block was not found / doesn't exist. If that's the case, the error should be 404.

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+	backend *backend.Backend
+	logger  zerolog.Logger
+	encoder encoding.Encoder
+}
+
+func NewRestAPIHandler(backend *backend.Backend, logger zerolog.Logger) *RestAPIHandler {
+	return &RestAPIHandler{
+		backend: backend,
+		logger:  logger,
+		encoder: encoding.DefaultEncoder, //use the default JSON encoder
```
smnzhu, line 27: Are we sure this is needed? I would have assumed that the generated server code would have provided some way of automatically serializing the generated structs (e.g `generated.Block`)?

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (restAPI *RestAPIHandler) NotImplemented(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(http.StatusNotImplemented)
+}
+
+func (restAPI *RestAPIHandler) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
```
smnzhu, line 77: As mentioned, we probably need to be able to distinguish different status codes

engine/access/rest_api_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/metrics"
+	module "github.com/onflow/flow-go/module/mock"
+	"github.com/onflow/flow-go/network"
+	protocol "github.com/onflow/flow-go/state/protocol/mock"
+	storagemock "github.com/onflow/flow-go/storage/mock"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+// RestAPITestSuite tests that Access node provides a secure GRPC server
```
smnzhu, line 29: comment inaccurate?

engine/access/rpc/engine.go
```diff
@@ -152,13 +156,17 @@ func New(log zerolog.Logger,
 		log,
 	)
 
+	restAPIHandler := rest.NewRestAPIHandler(backend, log)
+	restServer := rest.NewRestAPIServer(restAPIHandler, config.RESTListenAddr, log)
```
smnzhu, line 160: Is the REST api enabled by default? Should be make this configurable at command line?

engine/access/rpc/engine.go
```diff
@@ -307,3 +326,27 @@ func (e *Engine) serveGRPCWebProxy() {
...
+
+	e.log.Info().Str("rest_api_address", e.config.RESTListenAddr).Msg("starting REST server on address")
+
+	l, err := net.Listen("tcp", e.config.RESTListenAddr)
+	if err != nil {
+		e.log.Err(err).Msg("failed to start the REST server")
+		return
+	}
+
+	e.resetAPIAddress = l.Addr()
```
smnzhu, line 341:
```suggestion
	e.restAPIAddress = l.Addr()
```

smnzhu (commented):
Overall looks good :)

engine/access/swagger/generated/api_blocks.go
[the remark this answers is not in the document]
smnzhu: So the implementation is supposed to go directly into the generated code, but you're right that this will make things more annoying if we need to regenerate the code. I think the idea is that the API should be stable anyways, and any breaking changes to it would warrant a new API version, which would imply a new implementation.

I think that eventually, we will want to move all the implementation into these generated files. For now, it's probably fine to implement them in a separate file since we could potentially be regenerating the server code if we make any tweaks to the API spec.

engine/access/rest/converter.go
```diff
@@ -0,0 +1,69 @@
...
+		Header:  toBlockHeader(flowBlock.Header),
+		Payload: toBlockPayload(flowBlock.Payload),
+	}
+}
+
+func toBlockHeader(flowHeader *flow.Header) *generated.BlockHeader {
+	return &generated.BlockHeader{
+		Id:                   flowHeader.ID().String(),
+		ParentId:             flowHeader.ParentID.String(),
+		Height:               fmt.Sprint(flowHeader.Height),
```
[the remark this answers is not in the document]
vishalchangrani: The type is `string`, the format is `uint64`. The generated models define it as a string and won't let me set Height to flowHeader.Height. 
Did you mean changing it in `access.yaml`?

engine/access/rest/rest_api.go
vishalchangrani: actually the generated code did not import the json encoder. They don't have a fully working [Go server example](https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/go-api-server/go/api_pet.go) but the [client code example](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/go/go-petstore/client.go#L16) that they have does import `encoding/json`

engine/access/rpc/engine.go
vishalchangrani: agreed..added a flag

engine/access/rest/converter.go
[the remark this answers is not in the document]
smnzhu: Yeah, I meant we should change it in `access.yaml`. It's the only one that is string, all other integers are type integer

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (restAPI *RestAPIHandler) NotImplemented(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(http.StatusNotImplemented)
+}
+
+func (restAPI *RestAPIHandler) errorResponse(w http.ResponseWriter, r *http.Request, err error, returnCode int) {
```
peterargue, line 83: Generally, you don't want to return specific internal errors from web servers, since it gives attackers too much insight into the system. 

Rather than encoding the specific errors, we should return generic ones in the response, and log the exact error.

e.g. just "Internal Server Error" or "Error getting block"

engine/access/rest/rest_api_server.go
```diff
@@ -0,0 +1,139 @@
...
+// duration and response code
+func newHandler(inner http.Handler, name string, logger zerolog.Logger) http.Handler {
+	apiLogger := logger.With().Str("route_name", name).Logger()
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		respWrite := newResponseWriter(w)
+		inner.ServeHTTP(respWrite, r)
+		apiLogger.Info().Str("method", r.Method).
+			Str("uri", r.RequestURI).
+			Dur("duration", time.Since(start)).
```
peterargue, line 121: maybe also include the client id, and response size if possible

cmd/access/node_builder/access_node_builder.go
```diff
@@ -113,6 +113,8 @@ func DefaultAccessNodeConfig() *AccessNodeConfig {
 			UnsecureGRPCListenAddr:    "0.0.0.0:9000",
 			SecureGRPCListenAddr:      "0.0.0.0:9001",
 			HTTPListenAddr:            "0.0.0.0:8000",
+			RESTListenAddr:            "0.0.0.0:8080",
```
peterargue, line 116: the REST server feels like something that should be explicitly enabled (the gRPC-Web as well, but that's out of scope).

Could we simplify the config and only enable the rest server if the address if configured?

engine/access/swagger/generated/api_blocks.go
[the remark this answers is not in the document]
smnzhu: @vishalchangrani could we create an issue to track this? Thanks :)

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,94 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	// gorilla mux retains opening and ending square brackets for ids
+	idParam = strings.TrimSuffix(idParam, "]")
+	idParam = strings.TrimPrefix(idParam, "[")
+
+	ids := strings.Split(idParam, ",")
+
+	blocks := make([]*generated.Block, len(ids))
```
peterargue, line 47: do we have a max number of IDs? we should enforce some sensible default max.

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,94 @@
...
+		encoder: encoding.DefaultEncoder, //use the default JSON encoder
+	}
+}
+
+func (restAPI *RestAPIHandler) BlocksIdGet(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	vars := mux.Vars(r)
+	idParam, ok := vars["id"]
+	if !ok {
+		w.WriteHeader(http.StatusBadRequest)
```
peterargue, line 38: we should tell the user why the request was invalid

engine/access/rest/converter.go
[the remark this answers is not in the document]
vishalchangrani: Made that change on the spec - https://github.com/onflow/flow/pull/684

engine/access/swagger/generated/api_blocks.go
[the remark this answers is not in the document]
vishalchangrani: yup- https://github.com/dapperlabs/flow-go/issues/5997

engine/access/rest/converter.go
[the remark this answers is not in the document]
smnzhu: @vishalchangrani I've approved that :), thanks

engine/access/rest/rest_api.go
smnzhu: Yeah, I second this. It would also be a good idea to add this to the spec as well.

engine/access/rest/rest_api.go
smnzhu: Yep, we can use `errorResponse` here.

smnzhu: In fact, I *think* this should actually never happen, right? Because if the routes are working properly, we should not hit this handler unless the `id` part of `/blocks/{id}` is nonempty?

Can we confirm this, and if so, we probably don't even need to capture the `ok` since we're guaranteed it will always be true?

engine/access/rest_api_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}, 5*time.Second, 10*time.Millisecond)
+}
+
+func TestRestAPI(t *testing.T) {
+	suite.Run(t, new(RestAPITestSuite))
+}
+
+func (suite *RestAPITestSuite) TestRestAPICall() {
+
+	suite.Run("happy path - REST client successfully executed the GetBlockByID request", func() {
```
smnzhu, line 110: Can we add tests for the two unhappy paths: BadRequest and NotFound

cmd/access/node_builder/access_node_builder.go
vishalchangrani: so I have added a new flag [serve-rest-api](https://github.com/onflow/flow-go/blob/vishal/rest_api_implementation/cmd/access/node_builder/access_node_builder.go#L557) to enable/disable the REST Server.

engine/access/rest/rest_api.go
vishalchangrani: confirmed - a request without an ID does not reach here. Changing the code to assume vars["id"] will always returning a value.

engine/access/rest_api_test.go
vishalchangrani: added

engine/access/rest/rest_api_server.go
vishalchangrani: added client IP and user-agent, but response size seems a little tricky since I the actual writing is done by some underlying writer.

engine/access/rest/rest_api.go
vishalchangrani: 1. Added to the spec - [PR](https://github.com/onflow/flow/pull/686)
2. Added a validation here in the server as well. Unfortunately, the generated server code (unlike the generated client code) does not add the validation automatically.

cmd/access/node_builder/access_node_builder.go
peterargue: I'm suggesting it should be disabled by default since it's a non-critical interface that not all operators would use. The gRPC-Web one probably should be as well, but that's out of scope and harder to change since it's already enabled by default.

engine/access/rest/converter.go
```diff
@@ -0,0 +1,70 @@
...
+		signerIDs[i] = signerID.String()
+	}
+	return generated.CollectionGuarantee{
+		CollectionId: flowCollGuarantee.CollectionID.String(),
+		SignerIds:    signerIDs,
+		Signature:    base64.StdEncoding.EncodeToString(flowCollGuarantee.Signature.Bytes()),
+	}
+}
+
+func blockSeals(flowSeals []*flow.Seal) []generated.BlockSeal {
```
peterargue, line 57: should these `blockSeal(s)` also be prefixed with `to`? I'm not clear on the distinction.

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,112 @@
...
+
+	blocks := make([]*generated.Block, len(ids))
+
+	if len(ids) > MaxBlockIDCount {
+		restAPI.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("at most %d Block IDs can be requested at a time", MaxBlockIDCount), errorLogger)
+		return
+	}
+
+	for i, id := range ids {
+		flowID, err := flow.HexStringToIdentifier(id)
```
peterargue, line 57: we should validate `id` before parsing it. [HexStringToIdentifier](https://github.com/onflow/flow-go/blob/master/model/flow/identifier.go#L36-L36) will decode the input first, then return an error if it's not valid. This means that an attacker could attempt to exploit the `hex.Decode` method with malicious ids. e.g. send a list of very large IDs to tie up memory, or exploit some undiscovered bug in the implementation.

we should check that the ID is the correct length and only hex characters. there are a number of libraries we could use to simplify this for all of the inputs we're going to receive in this api.

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			restAPI.errorResponse(w, http.StatusInternalServerError, fmt.Sprintf("failed to look up block with ID %s", id), errorLogger)
+			return
+		}
+		blocks[i] = toBlock(flowBlock)
+	}
+
+	encodedBlocks, err := json.Marshal(blocks)
+	if err != nil {
+		restAPI.errorResponse(w, http.StatusInternalServerError, err.Error(), errorLogger)
```
peterargue, line 78: are we sure this error message is OK to return? maybe we should just log the error and return something like `Error generating response`

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+
+	_, err = w.Write(encodedBlocks)
+	if err != nil {
+		restAPI.errorResponse(w, http.StatusInternalServerError, err.Error(), errorLogger)
```
peterargue, line 84: are we sure this error message is OK to return? maybe we should just log the error and return something like `Error generating response`

engine/access/rest/rest_api_server.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// NewRestAPIServer returns an HTTP server initialized with the REST API handler
+func NewRestAPIServer(api *APIHandler, listenAddress string, logger zerolog.Logger) *http.Server {
+
+	router := mux.NewRouter().StrictSlash(true)
+	for _, route := range apiRoutes(api) {
+		var handler http.Handler
+		handler = route.HandlerFunc
+		handler = newHandler(handler, route.Name, logger)
```
peterargue, line 21: nit:
```suggestion
		handler := newHandler(route.HandlerFunc, route.Name, logger)
```

engine/access/rest_api_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+		client := suite.restAPIClient()
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+		defer cancel()
+
+		const invalidBlockID = "invalid_block_id"
+		_, resp, err := client.BlocksApi.BlocksIdGet(ctx, []string{invalidBlockID}, nil)
+		assertError(suite.T(), resp, err, http.StatusBadRequest, fmt.Sprintf("invalid ID %s", invalidBlockID))
+	})
+}
```
peterargue, line 153: how about tests for multiple blocks:
* happy path
* invalid block somewhere in the list
* more than max blocks

engine/access/rest/rest_api_server.go
```diff
@@ -0,0 +1,152 @@
...
+		router.
+			Methods(route.Method).
+			Path(route.Pattern).
+			Name(route.Name).
+			Handler(handler)
+	}
+
+	return &http.Server{
+		Addr:    listenAddress,
```
sideninja, line 30: A good practice to set timeouts to avoid Slowloris attacks.
```
        WriteTimeout: time.Second * 15,
        ReadTimeout:  time.Second * 15,
        IdleTimeout:  time.Second * 60,
```
Potentially even expose these args and make them configurable but maybe not necessary.

engine/access/rpc/engine.go
```diff
@@ -307,3 +327,30 @@ func (e *Engine) serveGRPCWebProxy() {
...
+	if err != nil {
+		e.log.Err(err).Msg("failed to start the REST server")
+		return
+	}
+
+	e.restAPIAddress = l.Addr()
+
+	e.log.Debug().Str("rest_api_address", e.restAPIAddress.String()).Msg("listening on port")
+
+	err = e.restServer.Serve(l) // blocking call
```
sideninja, line 349: Not sure how engines should handle this kind of things but would it make sense to launch as goroutine to not block the thread?

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
sideninja, line 87: I feel we should extract this to a helper method like `JSONResponse` where we handle serialization and errors and set headers. The reason for that is every handler will reimplement this code. Something like (note code is handwritten as an example, not tested)
```
func JSONResponse(w http.ResponseWriter, payload interface{}) {
	w.Header().Set("Content-Type", "application/json")
	
	response, err := json.Marshal(payload)
	if err != nil {
		restAPI.errorResponse(w, http.StatusInternalServerError, err.Error(), errorLogger)
		return
	}
	
	_, err = w.Write(response)
	if err != nil {
		restAPI.errorResponse(w, http.StatusInternalServerError, err.Error(), errorLogger)
		return
	}
	
	w.WriteHeader(http.StatusOK)
}
```

engine/access/rest/rest_api.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (restAPI *APIHandler) NotImplemented(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(http.StatusNotImplemented)
+}
+
+// errorResponse sends an HTTP error response to the client with the given return code and a model error with the given
+// response message in the response body
+func (restAPI *APIHandler) errorResponse(w http.ResponseWriter, returnCode int, responseMessage string, logger zerolog.Logger) {
+	w.WriteHeader(returnCode)
```
sideninja, line 98: Should we also log an error before responding or would that be too verbose logging?

engine/access/rest_api_test.go
sideninja, line 153: A case where the length of block IDs exceeds the maximum may be good for preventing regression.

sideninja (approved):
Left some minor comments

engine/access/rest/rest_api.go
sideninja: It might even be good to extract that into a parser for IDs reused whenever we parse IDs as part of the API so there is no regression.

cmd/access/node_builder/access_node_builder.go
vishalchangrani: switching the flag default to false to disable by default.

engine/access/rest/converter.go
vishalchangrani: yes, I renamed it

engine/access/rest/rest_api.go
vishalchangrani: I started adding this but then realized as @sideninja suggested that this needs to be its own validation layer. Hence, I created this [issue](https://github.com/dapperlabs/flow-go/issues/6006) to tackle is separately.

engine/access/rest_api_test.go
vishalchangrani: added these test cases.

engine/access/rpc/engine.go
vishalchangrani: yes, it is indeed run in a separate go-routine [here](https://github.com/onflow/flow-go/blob/f8f52fe044566776c3385d065cd689c34f59e465/engine/access/rpc/engine.go#L208)

engine/access/rest_api_test.go
vishalchangrani: yup added that

engine/access/rest/rest_api.go
vishalchangrani: any internal error is logged using the logger, but any other http error e.g. not found is not logged since it would be logged by the [logger interceptor](https://github.com/onflow/flow-go/blob/vishal/rest_api_implementation/engine/access/rest/rest_api_server.go#L117-#L133)

vishalchangrani:
@sideninja @peterargue - have made changes as per your comments. can you review it when you get a chance? thank you

engine/access/rest/rest_api.go
peterargue: I'm ok tracking that as a separate issue if we pick that issue up before implementing the rest of the endpoints.

engine/access/rest_api_test.go
```diff
@@ -0,0 +1,253 @@
...
+		defer cancel()
+
+		blocks, resp, err := client.BlocksApi.BlocksIdGet(ctx, []string{block.ID().String()}, nil)
+		require.NoError(suite.T(), err)
+		assert.Equal(suite.T(), http.StatusOK, resp.StatusCode)
+		assert.Len(suite.T(), blocks, 1)
+		assert.Equal(suite.T(), block.ID().String(), blocks[0].Header.Id)
+	})
+
+	suite.Run("GetBlockByID for multiple IDs - happy path", func() {
```
vishalchangrani, line 132: added this new test and test for checking permissible number of block IDs and non-existing block id when request multiple ids.

cmd/access/node_builder/access_node_builder.go
peterargue: since it's off by default, could we just use the existence of `--rest-addr` to indicate that the endpoint should be enabled?

peterargue (approved):
looks great. thanks @vishalchangrani!

cmd/access/node_builder/access_node_builder.go
vishalchangrani: yup that makes sense

sideninja:
Looks good to me @vishalchangrani

PR #1527, opened 2021-10-26
[Consensus Voting] Double signature vote processor

durkmurder:
### Context

This PR contains implementation of new vote processor that can process votes which contains two signatures.

Done:
- Renamed `CombinedVoteProcessor` to `CombinedVoteProcessorV3`
- Added new logic for handling double signatures.
- Added new function to decode double sig.
- godoc
- separate test suite for `CombinedVoteProcessorV2`


consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil, fmt.Errorf("could not reconstruct random beacon group signature: %w", err)
+	}
+
+	blockSigData := &hotstuff.BlockSignatureData{
+		StakingSigners:               stakingSigners,
+		RandomBeaconSigners:          nil,
+		AggregatedStakingSig:         aggregatedStakingSig,
+		AggregatedRandomBeaconSig:    nil,
+		ReconstructedRandomBeaconSig: reconstructedBeaconSig,
+	}
```
zhangchiqing, line 219: I would suggest to remove the unused fields

```suggestion
	blockSigData := &hotstuff.BlockSignatureData{
		StakingSigners:               stakingSigners,
		AggregatedStakingSig:         aggregatedStakingSig,
		ReconstructedRandomBeaconSig: reconstructedBeaconSig,
	}
```

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -0,0 +1,232 @@
...
+// combinedVoteProcessorFactoryBaseV2 is a `votecollector.baseFactory` for creating
+// CombinedVoteProcessors, holding all needed dependencies.
+// combinedVoteProcessorFactoryBaseV2 is intended to be used for the main consensus.
+// CAUTION:
+// this base factory only creates the VerifyingVoteProcessor for the given block.
+// It does _not_ check the proposer's vote for its own block, i.e. it does _not_
+// implement `hotstuff.VoteProcessorFactory`. This base factory should be wrapped
+// by `votecollector.VoteProcessorFactory` which adds the logic to verify
+// the proposer's vote (decorator pattern).
+type combinedVoteProcessorFactoryBaseV2 struct {
```
zhangchiqing, line 31: Not sure we need to name it `V2`, because we don't have `V1`.

I would suggest to just remove it.

durkmurder: But we have v3, I would say if we have v3 then v2 makes sense

zhangchiqing (approved):
Nice!

durkmurder:
bors merge

PR #1533, opened 2021-10-27
Khalil/5945 extend root snapshot

kc1116:
This PR updates SealingSegment to support the decoupling of ExecutionResults and ExecutionReceipts. If a block payload contains an ExecutionReceipt that references a ExecutionResult not found in the block payload the sealing segment will add that result to its ExecutionResults field. When a node is bootstrapped these SealingSegment.ExecutionResults get stored before SealingSegment.Blocks.


state/protocol/badger/snapshot.go
```diff
@@ -265,24 +265,47 @@ func (s *Snapshot) SealedResult() (*flow.ExecutionResult, *flow.Seal, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("could not get seal for sealing segment: %w", err)
+		return nil, fmt.Errorf("could not get seal for sealing seg: %w", err)
 	}
 
 	// walk through the chain backward until we reach the block referenced by
 	// the latest seal - the returned segment includes this block
-	var segment []*flow.Block
+	segment := flow.NewSealingSegment()
+	includedResults := make(map[flow.Identifier]*flow.ExecutionResult)
```
zhangchiqing, line 281: It's better to create a `SegmentBuilder` that holds this lookup internally, and in the end returns the segment.

The `SegmentBuilder` will take a method to lookup missing result by result ID, and when passing a block, it will go through the each receipt, and check if its result are known, if not, it will look it up. 
```
type SegmentBuilder struct {
  resultFinder      func(resultID flow.Identifier) (*flow.ExecutionResult, error)
  includedResults   map[flow.Identifier]*flow.ExecutionResult
  blocks  []*flow.Block
  results []*flow.ExecutionResult
}

// find the missing result to be added to 
func (b *SegmentBuilder) AddBlock(b *flow.Block) error {
}
func (b *SegmentBuilder) BuildSegment() (*Segment, error) {
}

```

And the BuildSegment function should validate that the highest block should include a seal that seals the lowest block, otherwise an error should be returned to indicate the Segment is incomplete.

state/protocol/badger/state.go
```diff
@@ -82,23 +82,30 @@ func Bootstrap(
...
-		if err != nil {
-			return fmt.Errorf("could not get sealing segment: %w", err)
-		}
+	err = operation.RetryOnConflictTx(db, transaction.Update, func(tx *transaction.Tx) error {
 		// sealing segment is in ascending height order, so the tail is the
 		// oldest ancestor and head is the newest child in the segment
 		// TAIL <- ... <- HEAD
-		head := segment[len(segment)-1] // reference block of the snapshot
-		tail := segment[0]              // last sealed block
+		head := segment.Blocks[len(segment.Blocks)-1] // reference block of the snapshot
```
zhangchiqing, line 101: It's better to define `head` and `tail` method on the Segment struct, otherwise, it's kind of `leaking` the knowledge.

Also I found `head` and `tail` is a bit confusing, they are usually used for labelling the position of a element in a list. And head usually is the first one, but here it's the last one in the slice.

I would suggest to just use `highest` and `lowest`, which implies their height. And we would know the `lowest` is the last sealed block.

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,33 @@
...
+
+// AddBlock appends block to Blocks
+func (segment *SealingSegment) AddBlock(block *Block) {
+	segment.Blocks = append(segment.Blocks, block)
+}
+
+// AddExecutionResult adds result to ExecutionResults
+func (segment *SealingSegment) AddExecutionResult(result *ExecutionResult) {
+	segment.ExecutionResults = append(segment.ExecutionResults, result)
+}
```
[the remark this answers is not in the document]
kc1116: This is the expected behavior , ExecutionResults can be empty

state/protocol/badger/state.go
kc1116: I agree

state/protocol/badger/snapshot.go
kc1116: This is more of a style preference, is there a builder API I can follow that's consistent throughout the repo ?

zhangchiqing: You can refer to the golang's [string builder](https://pkg.go.dev/strings#Builder), is similar to the design the suggested here. where you provide a block, and it takes care of looking up results. 

The nice thing with this is that the caller (protocol state) doesn't need to know how to build a segment. And it's also easier to unit test the building logic in SegmentBuilder

model/flow/sealing_segment.go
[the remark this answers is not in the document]
zhangchiqing: My point is that let's say a block A has a receipt that is referring to a unknown result to be queried. After queried that result, you will need to `AddExecutionResult` with that result, and then `AddBlock` with that block. But what happens if `AddExecutionResult` was called, but `AddBlock` wasn't? Or `AddBlock` was called, but `AddExecutionResult` wasn't? That would cause data inconsistency. And there is a possibility that this could happen, because the API design allows this. 

So what I'm suggesting to prevent this above from happening by combining `AddBlock` and `AddExecutionResult` ( or simply remove the `AddExecutionResult` ).

When adding a block, it should also find the missing result, and add it. In the end, there will be only 2 results:
either no block was added, or if a block is added, the missing result must have been added as well. The eliminates the cases where a block is added, but the result is not, or a result is added, but a block was not.

state/protocol/badger/snapshot.go
kc1116: I was hoping for something more flow-go specific, I can add a generic builder that is fine. Just to be clear essentially we will be moving the logic from [Snapshot.SealingSegment()](https://github.com/onflow/flow-go/blob/master/state/protocol/badger/snapshot.go#L268) and replacing it with this new builder pattern?

state/protocol/badger/snapshot.go
```diff
@@ -265,29 +265,44 @@ func (s *Snapshot) SealedResult() (*flow.ExecutionResult, *flow.Seal, error) {
...
-func (s *Snapshot) SealingSegment() ([]*flow.Block, error) {
+// SealingSegment will walk through the chain backward until we reach the block referenced
+// by the latest seal and build a SealingSegment. As we visit each block we check each execution
+// receipt in the blocks payload to make sure we have a corresponding execution result, any execution
+// results missing from blocks are stored in the SealingSegment.ExecutionResults field.
+func (s *Snapshot) SealingSegment() (*flow.SealingSegment, error) {
 	seal, err := s.state.seals.ByBlockID(s.blockID)
 	if err != nil {
-		return nil, fmt.Errorf("could not get seal for sealing segment: %w", err)
+		return nil, fmt.Errorf("could not get seal for sealing seg: %w", err)
```
jordanschalm, line 275:
```suggestion
		return nil, fmt.Errorf("could not get seal for sealing segment: %w", err)
```

state/protocol/badger/snapshot_test.go
```diff
@@ -301,12 +300,68 @@ func TestSealingSegment(t *testing.T) {
...
+			assert.Equal(t, block3.ID(), segment.Blocks[1].ID())
+			assert.Equal(t, block4.ID(), segment.Blocks[2].ID())
+		})
+	})
+
+	// test sealing segment where a block contains a ExecutionReceipt that
+	// references a ExecutionResult contained in a different block.
+	// the sealing segment should contain any ExecutionResults that were
+	// missing from a blocks payload in the SealingSegment.ExecutionResults field
+	// ROOT -> B1(Receipt_A) -> B2[Result_B, Receipt_B, Receipt_A_2] -> B3 -> B4 -> B5 (Seal_B1)
```
jordanschalm, line 314: I'm not sure if this matches the test code. 

This seems to say B2 has several receipts in its payload, but in the test B2 is empty. Maybe we're using different notations here? (I'm assuming `B1(A,B)` means block 1 contains resources A, B)

state/protocol/badger/snapshot_test.go
```diff
@@ -301,12 +300,68 @@ func TestSealingSegment(t *testing.T) {
...
+			block4.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt2), unittest.WithSeals(seal2)))
+			err = state.Extend(context.Background(), &block4)
+			require.NoError(t, err)
+
+			// build the block sealing block 1
+			block5 := unittest.BlockWithParentFixture(block4.Header)
+			receipt3, seal1 := unittest.ReceiptAndSealForBlock(&block1)
+			block5.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt3), unittest.WithSeals(seal1)))
+			err = state.Extend(context.Background(), &block5)
+			require.NoError(t, err)
```
jordanschalm, line 346: This is an invalid block structure, because we are sealing block 2 before block 1. A parent block is always finalized and sealed before its child. This wouldn't be enforced in this component, but it's still better to use internally consistent fixtures for tests we expect to pass.

state/protocol/badger/state.go
```diff
@@ -82,23 +82,30 @@ func Bootstrap(
...
+	}
 
-		// 1) insert each block in the root chain segment
-		err = state.bootstrapSealingSegment(root)(tx)
-		if err != nil {
-			return fmt.Errorf("could not bootstrap sealing chain segment: %w", err)
-		}
+	// bootstrap the sealing segment step 1: insert segment execution results.
+	// execution results must be inserted first.
+	err = state.bootstrapSealingSegmentExeResults(segment)
```
jordanschalm, line 92: Any reason this is not part of the database transaction like the subsequent state modifications? (ie. within `RetryOnConflictTx` and using `transaction.Tx`). It should be, otherwise it is possible to only partially bootstrap the state if an error occurs.

state/protocol/badger/state.go
```diff
@@ -82,23 +82,30 @@ func Bootstrap(
...
-			return fmt.Errorf("could not get sealing segment: %w", err)
-		}
+	err = operation.RetryOnConflictTx(db, transaction.Update, func(tx *transaction.Tx) error {
 		// sealing segment is in ascending height order, so the tail is the
 		// oldest ancestor and head is the newest child in the segment
 		// TAIL <- ... <- HEAD
-		head := segment[len(segment)-1] // reference block of the snapshot
-		tail := segment[0]              // last sealed block
+		highest := segment.Highest() // reference block of the snapshot
+		lowest := segment.Lowest()   // last sealed block
```
jordanschalm, line 102: 👍 for this name change

state/protocol/badger/state_test.go
```diff
@@ -149,24 +149,6 @@ func TestBootstrapNonRoot(t *testing.T) {
 	rootBlock, err := rootSnapshot.Head()
 	require.NoError(t, err)
 
-	// should be able to bootstrap from snapshot after building one block
```
jordanschalm, line 152: Just noting from our Slack message: this is safe to remove because it is no longer a valid test fixture, because child does not seal root.

state/protocol/badger/validity.go
```diff
@@ -135,21 +135,21 @@ func isValidRootSnapshot(snap protocol.Snapshot, verifyResultID bool) error {
 		return fmt.Errorf("could not latest sealed result: %w", err)
 	}
 
-	if len(segment) == 0 {
+	if len(segment.Blocks) == 0 {
 		return fmt.Errorf("invalid empty sealing segment")
 	}
 	// TAIL <- ... <- HEAD
```
jordanschalm, line 141:
```suggestion
```

state/protocol/badger/validity.go
```diff
@@ -135,21 +135,21 @@ func isValidRootSnapshot(snap protocol.Snapshot, verifyResultID bool) error {
...
-	tailID := tail.ID()
+	highest := segment.Highest() // reference block of the snapshot
+	lowest := segment.Lowest()   // last sealed block
+	highestID := highest.ID()
+	lowestID := lowest.ID()
 
-	if result.BlockID != tailID {
-		return fmt.Errorf("root execution result for wrong block (%x != %x)", result.BlockID, tail.ID())
+	if result.BlockID != lowestID {
+		return fmt.Errorf("root execution result for wrong block (%v != %v)", result.BlockID, lowest)
```
jordanschalm, line 148: This is functionally the same, because we implement a handler for `%v` for the `Identifier` type, but I think it is better to use the explicit formatting verb as much as possible so it is easy to see at a glance what the resulting string will look like.

state/protocol/snapshot.go
```diff
@@ -64,7 +64,11 @@ type Snapshot interface {
 	// the sealing segment has exactly one block (the root block for the spork).
```
jordanschalm, line 64: With this change, the note on line 59-61 is no longer true

state/protocol/util/testing.go
```diff
@@ -72,6 +74,30 @@ func RunWithBootstrapState(t testing.TB, rootSnapshot protocol.Snapshot, f func(
 	})
 }
 
+// RunWithFollowerProtocolStateAndResults will insert some results into results db. Used to test sealing segment edge cases.
+func RunWithFollowerProtocolStateAndResults(t testing.TB, rootSnapshot protocol.Snapshot, exeResults []*flow.ExecutionResult, f func(*badger.DB, *pbadger.FollowerState)) {
```
jordanschalm, line 78: I'm not sure it's necessary to duplicate this utility just to also add results. Could we do something like:
```go
// use the generic test utility
RunWithFollowerProtocolState(..., func(db *badger.DB, state *bprotocol.FollowerState) {

    // add the results needed for the specific test case
    for _, result := range exeResults {
        db.Update(operation.InsertExecutionResult(result))
    }

    // ...
}
```

jordanschalm (commented):
The approach looks good to me, thanks for these updates. 

Added a few suggestions and questions.

state/protocol/badger/validity.go
kc1116: The linter was failing , thats why I made these to %v updates. Did we change something with the linter?

state/protocol/util/testing.go
kc1116: nice thanks

state/protocol/badger/state.go
kc1116: Transactions are executed async so there is still the case of receipts getting stored without results. Thats why results are stored in their own transaction, before receipts transaction.

state/protocol/badger/validity.go
jordanschalm: Hmm, what lint error are you getting?

state/protocol/badger/state.go
jordanschalm: Transactions _are_ executed asynchronously, but storage operations are blocking, so each operation within a transaction will occur strictly consecutively, unless the transaction user is introducing concurrency. 

If I have a transaction like:
```go
db.Update(func(*badger.Txn) error {
   operation.InsertSomething(A)(tx)
   operation.InsertSomething(B)(tx)
}
```
I know that `A` is inserted when `B`.

From Badger's [blog post about this](https://dgraph.io/blog/post/badger-txn/):
> As writes are performed in the transaction, we maintain an internal cache of these writes within the transaction. Thus, any follow up reads on the same key by this transaction would see this write. But, other transactions won’t, providing Isolation and Consistency.

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,31 @@
...
+
+// NewSealingSegment returns SealingSegment
+func NewSealingSegment() *SealingSegment {
+	return &SealingSegment{
+		Blocks:           make([]*Block, 0),
```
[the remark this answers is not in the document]
durkmurder: What is meant leaving it as nil since it's valid to do `smth = append(nil, elem)`

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(segment.Blocks) == 0 {
+		return nil, ErrSegmentBlocksEmpty
+	}
+
+	// segment with 1 block
+	if len(segment.Blocks) == 1 {
+		return segment, nil
+	}
+
+	for _, seal := range segment.Highest().Payload.Seals {
```
durkmurder, line 85: As I understand that's a sanity check, it would be easier to understand it with specific comment and motivation

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestSealingSegmentBuilder_AddBlock(t *testing.T) {
```
durkmurder, line 14: Please add minimal godoc for tests briefly describing point of the test

durkmurder, line 14: Please add minimal godoc for tests briefly describing point of the test

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,84 @@
...
+	t.Run("should return err if result lookup fails", func(t *testing.T) {
+		block1 := unittest.BlockFixture()
+		receipt1 := unittest.ReceiptForBlockFixture(&block1)
+		resultLookup := func(flow.Identifier) (*flow.ExecutionResult, error) { return nil, fmt.Errorf("not found") }
+		builder := flow.NewSealingSegmentBuilder(resultLookup)
+
+		block1.SetPayload(unittest.PayloadFixture(unittest.WithReceiptsAndNoResults(receipt1)))
+
+		err := builder.AddBlock(&block1)
+		require.Error(t, err)
```
durkmurder, line 53: in such cases I really prefer returning predefined exception and check if the error was correctly wrapped.
[example](https://github.com/onflow/flow-go/blob/2dc58e4344847baf4e30b817ce47a02524c04592/consensus/hotstuff/votecollector/combined_vote_processor_test.go#L132)

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+}
+
+func TestSealingSegmentBuilder_SealingSegment(t *testing.T) {
```
durkmurder, line 57: same comment about godoc

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,84 @@
...
+		block2 := unittest.BlockWithParentFixture(block1.Header)
+
+		err := builder.AddBlock(&block1)
+		require.NoError(t, err)
+
+		err = builder.AddBlock(&block2)
+		require.NoError(t, err)
+
+		_, err = builder.SealingSegment()
+		require.True(t, errors.Is(err, flow.ErrSegmentMissingSeal))
```
durkmurder, line 75: FYI you can use https://pkg.go.dev/github.com/stretchr/testify/require#ErrorIs

state/protocol/badger/snapshot_test.go
durkmurder: :100:

state/protocol/badger/snapshot_test.go
durkmurder: B4 has to be empty but it contains seals and results, I guess you need to update godoc or code.

state/protocol/badger/state.go
durkmurder: +1 on this, better to be consistent and use tx as well

durkmurder (commented):
Looks much better, I like the `SealingSegmentBuilder` a lot, good job. A few comments to fix and we should be good to go.

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,102 @@
...
+
+// AddBlock appends block to blocks
+func (builder *SealingSegmentBuilder) AddBlock(block *Block) error {
+	for _, receipt := range block.Payload.Receipts {
+		if _, ok := builder.includedResults[receipt.ResultID]; ok {
+			continue
+		}
+
+		resultsByID := block.Payload.Results.Lookup()
+		if _, ok := resultsByID[receipt.ResultID]; !ok {
```
zhangchiqing, line 49: Say we have the following chain:
`A <- B <- C(receipt_A) <- D (seal_A,receipt_B) <- E(seal_B)`

The sealing segment for block E is `B <- C <- D <- E` plus `[resultA]`. 

The reason when we add block C, and we learned result A is missing and needed to be queried, is because if B and C don't have result A, the child blocks won't have result A either, that's why we will look up the result from storage, and add it to the missing result list.

But it's important to ensure that we are traversing or adding blocks in the increasing height order. And I would like to add this sanity check here:
When adding a block, this block must be exactly 1 height higher than the highest block. 

Can we add this sanity check?

state/protocol/badger/snapshot_test.go
```diff
@@ -301,12 +301,78 @@ func TestSealingSegment(t *testing.T) {
...
+			assert.Equal(t, block3.ID(), segment.Blocks[1].ID())
+			assert.Equal(t, block4.ID(), segment.Blocks[2].ID())
+		})
+	})
+
+	// test sealing segment where a block contains a ExecutionReceipt that
+	// references a ExecutionResult contained in a different block.
+	// the sealing segment should contain any ExecutionResults that were
+	// missing from a blocks payload in the SealingSegment.ExecutionResults field
+	// ROOT -> B1(Rec_1) -> B2[Res_1, Rec_2, Res_2] -> B3(Rec_3, Res_3, Seal_B1) -> B4 -> B5(Rec_4, Res_4, Seal_B4)
```
zhangchiqing, line 315: ` B5(Rec_4, Res_4, Seal_B4)`

This is not a valid case, because, seals can't be included in the same block as the result. So technically , the sealingSegment length has to be 3, because the result/receipt can't be in the same block as the executed block, and the seal can't be in the same block as the result/receipt. 

I'd like to request the tests for 2 other cases:
1. 
`A <- B[Result_A, Receipt_A_1] <- C[Result_B, Receipt_B, Receipt_A_2] <- D <- E <- F (Seal_C)`

The sealing segment for block F is `[C,D,E,F] + [result_A]`, `C` has `Receipt_A_2`, whose result is in block B

2.
`A <- B[Result_A, Receipt_A_1] <- C[Result_B, Receipt_B, Receipt_A_2] <- D (Receipt_B_2, Receipt_A_3) <- E <- F (Seal_C)`

The sealing segment for block F is `[C,D,E,F] + [resultA]`, because even though block C has a `receiptA_2`, and block D has a `receipt_A_3`, both receipts have the same result, so resultA should only be added once.

state/protocol/badger/validity.go
kc1116: typo on my end

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// invalid empty sealing segment blocks
+	if len(segment.Blocks) == 0 {
+		return nil, ErrSegmentBlocksEmpty
+	}
+
+	// segment with 1 block
+	if len(segment.Blocks) == 1 {
+		return segment, nil
+	}
```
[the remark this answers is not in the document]
jordanschalm: @zhangchiqing Root snapshots which we generate during a spork will always have only 1 block, so we need to support that case.
https://github.com/onflow/flow-go/blob/3043462f40495e4dd2701e53f62dd29b142aa682/state/protocol/snapshot.go#L63-L65

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,149 @@
...
+		return true
+	}
+
+	return block.Header.Height == builder.highest().Header.Height+1
+}
+
+// hasValidSeal returns true if highest block in the segment contains a seal for the lowest block
+func (builder *SealingSegmentBuilder) hasValidSeal() bool {
+	for _, seal := range builder.highest().Payload.Seals {
+		if seal.BlockID == builder.lowest().ID() {
```
zhangchiqing, line 119: better to cache the `ID()`, otherwise we will compute over and over

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,149 @@
...
+	segment := &SealingSegment{
+		Blocks:           builder.blocks,
+		ExecutionResults: builder.results,
+	}
+
+	if len(segment.Blocks) < 1 {
+		return nil, fmt.Errorf("expect at least 2 blocks in a sealing segment or 1 block in the case of root segments, but actually got %v: %w", len(segment.Blocks), ErrSegmentBlocksWrongLen)
+	}
+
+	// if root sealing segment skip seal sanity check
```
zhangchiqing, line 95: the root block has view `0`, better to [include that sanity check](https://github.com/onflow/flow-go/blob/petera/consensus-follower-error-handling-multipart2/cmd/bootstrap/run/block.go#L21)

storage/badger/payloads.go
```diff
@@ -43,26 +43,25 @@ func (p *Payloads) storeTx(blockID flow.Identifier, payload *flow.Payload) func(
 	// which is not included in current payload but was incorporated in one of previous blocks.
 	// TODO: refactor receipt/results storages to support new type of storing/retrieving where execution receipt
 	// and execution result is decoupled.
```
zhangchiqing, line 45:
```suggestion
```

Done already

storage/badger/payloads.go
```diff
@@ -43,26 +43,25 @@ func (p *Payloads) storeTx(blockID flow.Identifier, payload *flow.Payload) func(
...
+
+	return func(tx *transaction.Tx) error {
+
+		resultsByID := payload.Results.Lookup()
+		fullReceipts := make([]*flow.ExecutionReceipt, 0, len(payload.Receipts))
+		var err error
+		for _, meta := range payload.Receipts {
+			result, ok := resultsByID[meta.ResultID]
+			if !ok {
+				result, err = p.results.ByIDTx(meta.ResultID)(tx)
```
zhangchiqing, line 55: Why this has to be in a transaction?

If a receipt's result is not found in the payload, it means the result must have been seen, which means the storage should have it. Since a result once is stored, will never be removed, it must be found.

state/protocol/badger/snapshot_test.go
```diff
@@ -301,14 +300,125 @@ func TestSealingSegment(t *testing.T) {
...
+			err = state.Extend(context.Background(), &block4)
+			require.NoError(t, err)
+			err = state.Extend(context.Background(), &block5)
+			require.NoError(t, err)
+
+			segment, err := state.AtBlockID(block5.ID()).SealingSegment()
+			require.NoError(t, err)
+
+			require.Len(t, segment.Blocks, 4)
+
```
zhangchiqing, line 353: Is it possible to add such check?
```suggestion
			require.Len(t, segment.Blocks, 4)
			require.Equal(t, []*flow.Block{&block2, &block3, &block4, &block5}, segment.Blocks)
```

state/protocol/badger/snapshot_test.go
```diff
@@ -301,14 +300,125 @@ func TestSealingSegment(t *testing.T) {
...
+			assert.Equal(t, block2.ID(), segment.Blocks[0].ID())
+			assert.Equal(t, block3.ID(), segment.Blocks[1].ID())
+			assert.Equal(t, block4.ID(), segment.Blocks[2].ID())
 		})
 	})
-}
 
+	// test sealing segment where you have a chain that is 5 blocks long and the block 5 has a seal for block 2.
+	// block 2 also contains a receipt but no result.
+	// root -> B1[Result_A, Receipt_A_1] -> B2[Result_B, Receipt_B, Receipt_A_2] -> B3 -> B4 -> B5 (Seal_B2)
```
zhangchiqing, line 312:
```suggestion
	// root -> B1[Result_A, Receipt_A_1] -> B2[Result_B, Receipt_B, Receipt_A_2] -> B3 -> B4 -> B5 (Seal_B2)
	// the segment for B5 should be `[B2,B3,B4,B5] + [Result_A]`
```

state/protocol/badger/snapshot_test.go
```diff
@@ -301,14 +300,125 @@ func TestSealingSegment(t *testing.T) {
...
 			// sealing segment should be [B2, B3, B4]
-			assert.Len(t, segment, 3)
-			assert.Equal(t, block2.ID(), segment[0].ID())
-			assert.Equal(t, block3.ID(), segment[1].ID())
-			assert.Equal(t, block4.ID(), segment[2].ID())
+			assert.Len(t, segment.Blocks, 3)
+			assert.Equal(t, block2.ID(), segment.Blocks[0].ID())
+			assert.Equal(t, block3.ID(), segment.Blocks[1].ID())
+			assert.Equal(t, block4.ID(), segment.Blocks[2].ID())
 		})
```
zhangchiqing, line 307: better to check the results list is empty

state/protocol/badger/snapshot_test.go
```diff
@@ -196,9 +196,9 @@ func TestSealingSegment(t *testing.T) {
 			actual, err := state.AtBlockID(head.ID()).SealingSegment()
 			require.NoError(t, err)
 
-			assert.Len(t, actual, 1)
-			assert.Equal(t, len(expected), len(actual))
-			assert.Equal(t, expected[0].ID(), actual[0].ID())
+			assert.Len(t, actual.Blocks, 1)
+			assert.Equal(t, len(expected.Blocks), len(actual.Blocks))
+			assert.Equal(t, expected.Blocks[0].ID(), actual.Blocks[0].ID())
 		})
```
zhangchiqing, line 202: better to add assertion on result list

state/protocol/badger/snapshot_test.go
```diff
@@ -222,12 +222,11 @@ func TestSealingSegment(t *testing.T) {
...
 			require.NoError(t, err)
-
 			// sealing segment should contain B1 and B2
 			// B2 is reference of snapshot, B1 is latest sealed
-			assert.Len(t, segment, 2)
-			assert.Equal(t, block1.ID(), segment[0].ID())
-			assert.Equal(t, block2.ID(), segment[1].ID())
+			assert.Len(t, segment.Blocks, 2)
+			assert.Equal(t, block1.ID(), segment.Blocks[0].ID())
+			assert.Equal(t, block2.ID(), segment.Blocks[1].ID())
```
zhangchiqing, line 229: better to add assertion on result list

zhangchiqing (approved):
Great work! Nice tests. 

I added a few minor suggestions

storage/badger/payloads.go
jordanschalm: When we bootstrap the state, we insert all blocks/receipts of the sealing segment within the Badger transaction scope. When we read the DB here, if we don't read in the context of the same Badger transaction, it will not see blocks/receipts inserted earlier in the transaction.

model/flow/sealing_segment.go
```diff
@@ -48,23 +50,29 @@ type SealingSegmentBuilder struct {
 
 // AddBlock appends block to blocks
 func (builder *SealingSegmentBuilder) AddBlock(block *Block) error {
-	for _, receipt := range block.Payload.Receipts {
-		if _, ok := builder.includedResults[receipt.ResultID]; ok {
-			continue
-		}
+	//sanity check: block should be 1 height higher than current highest
+	if !builder.isValidHeight(block) {
+		return fmt.Errorf("invalid block height (%x): %w", block.Header.Height, ErrSegmentInvalidBlockHeight)
```
jordanschalm, line 55:
```suggestion
		return fmt.Errorf("invalid block height (%d): %w", block.Header.Height, ErrSegmentInvalidBlockHeight)
```

I don't the we want the height hex-formatted

state/protocol/snapshot.go
```diff
@@ -49,22 +49,25 @@ type Snapshot interface {
...
 	// is the most recently sealed block as of this snapshot (ie. the block
 	// referenced by LatestSeal). The segment is in ascending height order.
 	//
 	// TAIL <- B1 <- ... <- BN <- HEAD
 	//
-	// NOTE 1: TAIL is not always sealed by HEAD. In the case that the head of
-	// the snapshot contains no seals, TAIL must be sealed by the first ancestor
-	// of HEAD which contains any seal.
+	// NOTE 1: The highest block in the sealing segment must have a seal
+	// for the lowest block in the sealing segment.
```
jordanschalm, line 60: Thinking about this more, I don't think we can use this stricter constraint.

Consider the chain segment, where B contains the execution receipt for A and C seals A. (D does not contain a seal.)
```
A<-B(R_A)<-C(S_A)<-D
```

We would like for every queryable protocol snapshot to have a valid sealing segment. I should be able to do `state.AtBlockID(C.ID())` and `state.AtBlockID(D.ID())` and receive a snapshot with a valid sealing segment for both. 

But since D does not contain a seal, it does not satisfy the stricter requirement we're introducing here.
> The highest block in the sealing segment must have a seal for the lowest block in the sealing segment.

I suggest we:
* revert this constraint on the sealing segment documentation to the previous definition: "Either HIGHEST seals LOWEST, or the first ancestor of HIGHEST containing any seal, seals LOWEST"
* make sure sealing segment building and validation allows for the case where the highest block contains no seal

state/protocol/snapshot.go
jordanschalm: no longer relevant

state/protocol/snapshot.go
kc1116: Interesting , in this case would this [line]( https://github.com/onflow/flow-go/blob/master/state/protocol/badger/snapshot.go#L269) not return an error because D does not contain a seal ?

state/protocol/badger/snapshot_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

state/protocol/badger/snapshot_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

state/protocol/badger/snapshot_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

state/protocol/badger/snapshot_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

storage/badger/payloads.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

model/flow/sealing_segment.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

model/flow/sealing_segment.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/00a77e58055c9b7fbc3a8d569aa91866610f5646

state/protocol/snapshot.go
jordanschalm: It won't. We index the _latest_ seal each time we insert a block ([reference](https://github.com/onflow/flow-go/blob/master/state/protocol/badger/mutator.go#L439-L443)), so `seals.ByBlockID` will return a seal for all inserted block IDs, even if the block did not contain any seal.

kc1116: Ahh okay that makes sense, I will update hasValidSeal to reflect this and revert the godoc

kc1116: > 	// NOTE 1: TAIL is not always sealed by HEAD. In the case that the head of
> 	// the snapshot contains no seals, TAIL must be sealed by the first ancestor
> 	// of HEAD which contains any seal.

In the case where D **_DOES_** contains seals but **_DOES NOT_** contain a seal for A do we still check for a valid ancestor or do we consider this invalid state?

jordanschalm: Exactly, If we see a sealing segment like `A<-B<-C<-D`, and D has no seal for A, but does have a different seal (eg. for B), then that is an invalid sealing segment.

kc1116: https://github.com/onflow/flow-go/pull/1533/commits/5ac9897faba9fa9e54bf8482bac9135677167f68

model/flow/sealing_segment.go
```diff
@@ -0,0 +1,193 @@
...
+	ErrSegmentMissingSeal        = fmt.Errorf("sealing segment failed sanity check: highest block in segment does not contain seal for lowest")
+	ErrSegmentBlocksWrongLen     = fmt.Errorf("sealing segment failed sanity check: must have atleast 2 blocks")
+	ErrSegmentInvalidBlockHeight = fmt.Errorf("sealing segment failed sanity check: blocks must be in ascending order")
+	ErrSegmentResultLookup       = fmt.Errorf("failed to lookup execution result")
+	ErrInvalidRootSegmentView    = fmt.Errorf("invalid root sealing segment block view")
+)
+
+type SealingSegmentBuilder struct {
+	resultLookup    func(resultID Identifier) (*ExecutionResult, error)
+	includedResults map[Identifier]*ExecutionResult
```
durkmurder, line 55: I don't see that we use it as map, we only use it for bookkeeping of already seen ERs, we should change it to 
`map[Identifier]struct{}` if we are using this one as set.

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,223 @@
...
+	// test that when the highest block in the segment does not contain seals but the first ancestor of highest
+	// does contain a seal for lowest we return a valid sealing segment.
+	t.Run("should return valid sealing segment if ancestor of highest seals lowest", func(t *testing.T) {
+		resultLookup := func(flow.Identifier) (*flow.ExecutionResult, error) { return nil, nil }
+		builder := flow.NewSealingSegmentBuilder(resultLookup)
+
+		block1 := unittest.BlockFixture()
+		block2 := unittest.BlockWithParentFixture(block1.Header)
+		receipt, seal := unittest.ReceiptAndSealForBlock(&block1)
+		block2.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt), unittest.WithSeals(seal)))
```
durkmurder, line 127: Hm, I think this is wrong. Well, this is not wrong in this test case. But this will be an invalid block since we [require](https://github.com/onflow/flow-go/blob/master/module/validation/seal_validator.go#L161) that there is at least one block between block incorporating result and block incorporating the seal. I think we should construct a valid segment.

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	t.Run("should return ErrSegmentMissingSeal if highest block contains no seals and first ancestor with seals does not seal lowest", func(t *testing.T) {
+		block0 := unittest.BlockFixture()
+		block1 := unittest.BlockWithParentFixture(block0.Header)
+		resultLookup := func(flow.Identifier) (*flow.ExecutionResult, error) { return nil, nil }
+		builder := flow.NewSealingSegmentBuilder(resultLookup)
+
+		block2 := unittest.BlockWithParentFixture(block1.Header)
+		receipt1, seal1 := unittest.ReceiptAndSealForBlock(&block1)
+		block2.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt1), unittest.WithSeals(seal1)))
```
durkmurder, line 193: same here

durkmurder (commented):
Thanks for addressing most of the stuff, looks really good at this point. I believe we still have to address a few things. Almost there :)

model/flow/sealing_segment_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/60a5951058ff9c214eb0b3d85afc5975b50a7721

model/flow/sealing_segment_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+		block3 := unittest.BlockWithParentFixture(block2.Header)
+		receipt2, seal2 := unittest.ReceiptAndSealForBlock(&block2)
+		block3.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt2), unittest.WithSeals(seal2)))
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/60a5951058ff9c214eb0b3d85afc5975b50a7721

model/flow/sealing_segment_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/60a5951058ff9c214eb0b3d85afc5975b50a7721

model/flow/sealing_segment.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/60a5951058ff9c214eb0b3d85afc5975b50a7721

utils/unittest/unittest.go
```diff
@@ -265,3 +267,11 @@ func Concurrently(n int, f func(int)) {
 	}
 	wg.Wait()
 }
+
+// AssertEqualBlocksLenAndOrder asserts that both a segment of blocks have the same len and blocks are in the same order
+func AssertEqualBlocksLenAndOrder(t *testing.T, expectedBlocks, actualSegmentBlocks []*flow.Block) {
```
jordanschalm, line 272: We can do this generally for any 2 lists of entities with:
```
assert.Equal(t, flow.GetIDs(expectedBlocks), flow.GetIDs(actualSegmentBlocks))
```

model/flow/sealing_segment.go
```diff
@@ -94,6 +103,11 @@ func (builder *SealingSegmentBuilder) SealingSegment() (*SealingSegment, error)
 
 	// if root sealing segment skip seal sanity check
 	if len(segment.Blocks) == 1 {
+
+		if !builder.isValidRootSegment() {
+			return nil, fmt.Errorf("root sealing segment block has the wrong view got (%d) expected (%d): %w", segment.Highest().Header.View, rootSegmentBlockView, ErrInvalidRootSegmentView)
```
jordanschalm, line 108: Here we are decoupling the error context (line 108) from the logic which defines the conditions that produce the error (lines 168-170), which makes it more difficult to verify that the logic and error context remain consistent.

What do you think about using a pattern of returning errors from validation functions, so that the error context from the validation function's logic can live within the validation function.
```go
func validateRootSegment() error {
   if view != expectedView {
     return fmt.Errorf("wrong view......") 
   }
   // ...
}

// ...
if err := builder.validateRootSegment(); err != nil {
  return nil, fmt.Errorf("could not validate root segment: %w", err)
}
```

state/protocol/badger/snapshot_test.go
```diff
@@ -418,7 +421,39 @@ func TestSealingSegment(t *testing.T) {
...
+		util.RunWithFollowerProtocolState(t, rootSnapshot, func(db *badger.DB, state *bprotocol.FollowerState) {
+			// build a block to seal
+			block1 := unittest.BlockWithParentFixture(head)
+			err := state.Extend(context.Background(), &block1)
+			require.NoError(t, err)
+
+			// build a block sealing block1
+			block2 := unittest.BlockWithParentFixture(block1.Header)
+			receipt1, seal1 := unittest.ReceiptAndSealForBlock(&block1)
+			block2.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt1), unittest.WithSeals(seal1)))
```
jordanschalm, line 435: Could we add an extra block to make this a valid chain segment (can't have receipt and seal for same block, in the same block).
```
A <- B(RA) <- C(SA) <- D
```

jordanschalm (approved):
A couple final suggestions, but LGTM besides. Nice work 🎸

model/flow/sealing_segment.go
kc1116: I like this thanks

state/protocol/badger/snapshot_test.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/ef3887b8355cf2913aa1911d598883538769238a

model/flow/sealing_segment.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/ef3887b8355cf2913aa1911d598883538769238a

model/flow/sealing_segment_test.go
kc1116: Using struct{}{} https://github.com/onflow/flow-go/pull/1533/commits/ef3887b8355cf2913aa1911d598883538769238a

utils/unittest/unittest.go
kc1116: https://github.com/onflow/flow-go/pull/1533/commits/382bd2b07c57196516ad4b817a74bd31659a1b92

model/flow/sealing_segment.go
kc1116: Using struct{}{} https://github.com/onflow/flow-go/pull/1533/commits/ef3887b8355cf2913aa1911d598883538769238a

state/protocol/badger/snapshot_test.go
```diff
@@ -272,45 +272,51 @@ func TestSealingSegment(t *testing.T) {
 
 	// test sealing segment where the segment blocks contain seals for
 	// ancestor blocks prior to the sealing segment
-	// ROOT -> B1 -> B2 -> B3(S1) -> B4(S2)
+	// ROOT -> B1 -> B2(R1) -> B3(R2, S1) -> B4 -> B5(S2)
```
durkmurder, line 275: Hm, to me it seems that there is no gap between R1 and S1, it should be something like this:
`R -> B1 -> B2(R1) -> B3 -> B3(R2, S1) -> B4 -> B5(S2)`

state/protocol/badger/snapshot_test.go
```diff
@@ -272,45 +272,51 @@ func TestSealingSegment(t *testing.T) {
...
+			block2.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt1)))
 			err = state.Extend(context.Background(), &block2)
 			require.NoError(t, err)
 
+			receipt2, seal2 := unittest.ReceiptAndSealForBlock(&block2)
+
 			block3 := unittest.BlockWithParentFixture(block2.Header)
-			receipt1, seal1 := unittest.ReceiptAndSealForBlock(&block1)
-			block3.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt1), unittest.WithSeals(seal1)))
+			block3.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(receipt2), unittest.WithSeals(seal1)))
```
durkmurder, line 293: Yeap, no gap here, need to add

state/protocol/badger/snapshot_test.go
```diff
@@ -272,45 +272,51 @@ func TestSealingSegment(t *testing.T) {
...
+			unittest.AssertEqualBlocksLenAndOrder(t, []*flow.Block{&block2, &block3, &block4, &block5}, segment.Blocks)
 
 			require.Len(t, segment.ExecutionResults, 0)
 		})
 	})
 
 	// test sealing segment where you have a chain that is 5 blocks long and the block 5 has a seal for block 2.
 	// block 2 also contains a receipt but no result.
-	// root -> B1[Result_A, Receipt_A_1] -> B2[Result_B, Receipt_B, Receipt_A_2] -> B3 -> B4 -> B5 (Seal_B2)
+	// root -> B1[Result_A, Receipt_A_1] -> B2[Result_B, Receipt_B, Receipt_A_2] -> B3(Receipt_for_seal) -> B4 -> B5 (Seal_B2)
```
durkmurder, line 319: What is `Receipt_for_seal`? Can you change it to some other name?

durkmurder (approved):
LGTM. I think once tests are fixed it's ready to be :ship: , nicely done.

kc1116:
bors merge

kc1116:
bors merge

PR #1543, opened 2021-10-28
Make NodeBuilder return runnable Nodes

peterargue:
This is the first part of a multi-part change to transition NodeBuilder to use the proposed [Component Interface](https://github.com/onflow/flow-go/blob/master/flips/component-interface.md). The overarching goal is to make the `ReadyDoneAware` interface more consistent by moving to a new `Component` interface that includes a `Start` method to explicitly start the component, and use contexts to handle signalling closure.

This was originally implemented in https://github.com/onflow/flow-go/pull/1499, but there were a large number of changes, so I've broken it down into multiple PRs to make reviewing more manageable.

This PR updates the `NodeBuilder` interface to more closely follow the builder pattern by adding a `Build()` method that returns a Node object which is then run. This is more useful when adding the `ComponentManager` functionality (will be added in another PR).
Additionally, this PR refactors the access node builders to move the staked access node specific setup to `staked_access_node_builder.go `


follower/consensus_follower.go
```diff
@@ -191,25 +195,21 @@ func (cf *ConsensusFollowerImpl) AddOnBlockFinalizedConsumer(consumer pubsub.OnB
...
 	}
 
 	select {
-	case <-anb.Ready():
-		anb.Logger.Info().Msg("Access node startup complete")
+	case <-cf.Ready():
+		cf.Logger.Info().Msg("Access node startup complete")
 	case <-ctx.Done():
-		anb.Logger.Info().Msg("Access node startup aborted")
+		cf.Logger.Info().Msg("Access node startup aborted")
```
zhangchiqing, line 208: this is no longer to do with `Access node`, better to update the comment

follower/consensus_follower.go
```diff
@@ -191,25 +195,21 @@ func (cf *ConsensusFollowerImpl) AddOnBlockFinalizedConsumer(consumer pubsub.OnB
...
 	}
 
 	<-ctx.Done()
-	anb.Logger.Info().Msg("Access node shutting down")
-	<-anb.Done()
-	anb.Logger.Info().Msg("Access node shutdown complete")
+	cf.Logger.Info().Msg("Access node shutting down")
+	<-cf.Done()
+	cf.Logger.Info().Msg("Access node shutdown complete")
```
zhangchiqing, line 214: this is no longer to do with Access node, better to update the comment

zhangchiqing (approved):

cmd/scaffold.go
```diff
@@ -966,37 +964,9 @@ func (fnb *FlowNodeBuilder) RegisterDefaultAdminCommands() {
...
-	case <-fnb.Done():
-		fnb.Logger.Info().Msgf("%s node shutdown complete", fnb.BaseConfig.NodeRole)
-	case <-fnb.sig:
-		fnb.Logger.Warn().Msg("node shutdown aborted")
-		os.Exit(1)
-	}
-
-	os.Exit(0)
+func (fnb *FlowNodeBuilder) Build() Node {
+	fnb.sig = make(chan os.Signal)
```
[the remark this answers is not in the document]
peterargue: that was unintentional. fixed in 0243920

follower/consensus_follower.go
peterargue: addressed in 0243920

follower/consensus_follower.go
peterargue: addressed in 0243920

cmd/node.go
```diff
@@ -0,0 +1,79 @@
...
+type FlowNodeImp struct {
+	module.ReadyDoneAware
+	*NodeConfig
+	Logger zerolog.Logger
+	sig    chan os.Signal
+}
+
+func NewNode(builder NodeBuilder, cfg *NodeConfig, logger zerolog.Logger, sig chan os.Signal) Node {
+	return &FlowNodeImp{
+		ReadyDoneAware: builder,
```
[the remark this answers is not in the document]
smnzhu: Ah, is this because the actual implementation for the two are still in the same underlying struct?

In this case we don't really need a separate `FlowNodeImp` at all right?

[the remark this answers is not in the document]
peterargue: yea, that's an artifact of this change being broken into 2 PRs. In #1559 `FlowNodeImp` embeds a `ComponentManager` which moves the `Component` interface.

cmd/node.go
```diff
@@ -0,0 +1,79 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/module"
+)
+
+type Node interface {
+	module.ReadyDoneAware
```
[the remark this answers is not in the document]
peterargue: Node is implemented as `Startable` in #1559. I'm not opposed to adding a `NodeRunner` to handle the node lifecycle, but I think that should be in a different PR since there's already a lot of change here.

cmd/access/node_builder/access_node_builder.go
```diff
@@ -70,11 +65,8 @@ import (
 type AccessNodeBuilder interface {
 	cmd.NodeBuilder
 
-	// IsStaked returns True is this is a staked Access Node, False otherwise
+	// IsStaked returns True if this is a staked Access Node, False otherwise
```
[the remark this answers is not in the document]
peterargue: good call. https://github.com/dapperlabs/flow-go/issues/5984

cmd/node.go
```diff
@@ -0,0 +1,79 @@
...
+	// initialize signal catcher
+	signal.Notify(node.sig, os.Interrupt, syscall.SIGTERM)
+
+	select {
+	case <-node.Ready():
+		node.Logger.Info().Msgf("%s node startup complete", node.BaseConfig.NodeRole)
+	case <-node.sig:
+		node.Logger.Warn().Msg("node startup aborted")
+		os.Exit(1)
+	}
```
smnzhu, line 54: Same race condition here, right?

cmd/node.go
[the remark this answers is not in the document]
smnzhu: Yeah, I think a different PR makes sense too. Can we just make sure we create an issue to track it? Thanks

cmd/node.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (node *FlowNodeImp) ShutdownSignal() <-chan struct{} {
+	shutdown := make(chan struct{})
+	go func() {
+		<-node.sig
+		close(shutdown)
+	}()
+	return shutdown
+}
```
smnzhu, line 79: 1. What is this actually needed for?
2. At the moment it seems that shutdown is only ever initiated when a signal is received, but generally it's conceivable that this may not be the only case? Would it be better to instead have a single channel which is stored in the struct and closed when we receive the signal?

peterargue: I could just remove it. I added it just to make it compatible with the ComponentManager version in the next PR. It's not actually used in this PR

cmd/node.go
peterargue: You mean that we could have received the signal but also become ready, and we don't run the shutdown logic?

This is just a copy/paste from `scaffold.go`. Those race conditions are handled explicitly in the next PR.

cmd/node.go
peterargue: removed in 68ca6eb

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -89,8 +101,146 @@ func (builder *StakedAccessNodeBuilder) Initialize() error {
...
+				Str("collection_node", anb.rpcConf.CollectionAddr).
+				Msg("using the static collection node address")
+
+			collectionRPCConn, err := grpc.Dial(
+				anb.rpcConf.CollectionAddr,
+				grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(grpcutils.DefaultMaxMsgSize)),
+				grpc.WithInsecure(),
+				backend.WithClientUnaryInterceptor(anb.rpcConf.CollectionClientTimeout))
+			if err != nil {
+				return err
```
yhassanzadeh13, line 124: general comment: better to prepend some contextual message to the error: `fmt.Errorf("could not dial collection node over grpc: %w", err)`

yhassanzadeh13 (approved):

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -89,8 +101,146 @@ func (builder *StakedAccessNodeBuilder) Initialize() error {
...
+		Module("collection node client", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
+			// collection node address is optional (if not specified, collection nodes will be chosen at random)
+			if strings.TrimSpace(anb.rpcConf.CollectionAddr) == "" {
+				node.Logger.Info().Msg("using a dynamic collection node address")
+				return nil
+			}
+
+			node.Logger.Info().
+				Str("collection_node", anb.rpcConf.CollectionAddr).
+				Msg("using the static collection node address")
```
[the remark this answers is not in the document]
peterargue: this logic is different. in the current case, it returns if `anb.rpcConf.CollectionAddr` is empty

peterargue:
bors merge

PR #1544, opened 2021-10-28
Suggestions for PR #1404

AlexHentschel:
## These are suggested revisions for PR https://github.com/onflow/flow-go/pull/1404

* largely goDoc revisions and polishing 
* Personally, I find go already very verbose. The less code I get on one screen, the more I have to scroll, which complicates code reviews _slightly_. Therefore, I prefer to remove unnecessary white spaces (just a few)
* I noticed that in some cases, error messages were very brief, and there were additional code comments for clarification. In these cases, I would prefer to make the error messages more self-explanatory. Thereby, it helps when debugging and removes the need for additional comments (which also extend the code length). 

Feel free to select the changes you like.

### DIFF
--- a/consensus/hotstuff/randombeacon_follower.go
+++ b/consensus/hotstuff/randombeacon_follower.go
@@ -9,36 +9,42 @@ import (
 // The random beacon methods are based on a threshold signature scheme.
 type RandomBeaconFollower interface {
 	// Verify verifies the signature share under the signer's public key and the message agreed upon.
+	// The function is thread-safe and wait-free (i.e. allowing arbitrary many routines to
+	// execute the business logic, without interfering with each other).
 	// It allows concurrent verification of the given signature.
-	// It returns :
+	// Returns :
 	//  - engine.InvalidInputError if signerIndex is invalid
 	//  - module/signature.ErrInvalidFormat if signerID is valid but signature is cryptographically invalid
-	//  - other error if there is an exception.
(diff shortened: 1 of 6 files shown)


tarakby:
> Personally, I find go already very verbose. The less code I get on one screen, the more I have to scroll, which complicates code reviews slightly. Therefore, I prefer to remove unnecessary white spaces (just a few)

The extra lines come from the [GoDoc convention for HTML docs](https://go.dev/blog/godoc). The extra line translates to a new paragraph in HTML. I quote "Subsequent lines of text are considered part of the same paragraph; you must leave a blank line to separate paragraphs."
This was a mini-convention Peter mentioned when we started this Go repo. You can see it is enforced in [this file for example](https://github.com/onflow/flow-go-sdk/blob/master/transaction.go). 
Since I have been using the rule within the crypto package, I'll keep it in the package to be consistent, but I will remove elsewhere as per your suggestion. 

>  I noticed that in some cases, error messages were very brief, and there were additional code comments for clarification. In these cases, I would prefer to make the error messages more self-explanatory.

In some cases I do not detail the error text because the wrapped error will give the details, something like `fmt.Errorf("function f failed: %w", err)`. If I detail the error in the text, there will be redundant info. 
I'll look again at the errors and provide more details when needed.

tarakby (approved):
Thank you for improving the code readability 🙏🏼

PR #1553, opened 2021-10-29
Crypto: improve Invalid input errors 

tarakby:
- update the `InvalidInputsError` structure (use error field)
- update the `IsInvalidInputsError` test to use `error.As`
- clean up returning `InvalidInputsError` in the library
- add more tests for invalid inputs


crypto/dkg_jointfeldman.go
```diff
@@ -202,10 +199,7 @@ func (s *JointFeldmanState) HandleBroadcastMsg(orig int, msg []byte) error {
 	for i := index(0); int(i) < s.size; i++ {
 		err := s.fvss[i].HandleBroadcastMsg(orig, msg)
 		if err != nil {
-			if IsInvalidInputsError(err) {
-				newInvalidInputsError("handle message has failed: %s", err)
-			}
```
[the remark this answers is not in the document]
tarakby: `HandleBroadcastMsg` can return a `invalidInputError` error or another error. If it is `invalidInputError`, it would be wrapped by the `fmt.Errorf` error and `isInvalidInputError` returns `true`.

crypto/bls_test.go
```diff
@@ -529,10 +634,24 @@ func TestBatchVerify(t *testing.T) {
...
+		for i := 0; i < sigsNum; i++ {
+			expectedValid[i] = false
+		}
+		pks[0] = invalidSK(t).PublicKey()
+		valid, err := BatchVerifyBLSSignaturesOneMessage(pks, sigs, input, nil)
+		require.Error(t, err)
+		assert.True(t, IsInvalidInputsError(err))
+
+		assert.Equal(t, valid, expectedValid,
+			fmt.Sprintf("verification should fail with invalid key, got %v", valid))
```
jordanschalm, line 654:
```suggestion
		assert.Equalf(t, valid, expectedValid, "verification should fail with invalid key, got %v", valid))
```

Add an `f` to the end of most `assert` or `require` functions, and the final variadic parameter gets treated as a format string and inputs.

crypto/bls_test.go
```diff
@@ -690,24 +809,46 @@ func TestAggregateSignaturesManyMessages(t *testing.T) {
...
+		tmp := inputKmacs[0]
+		inputKmacs[0] = nil
+		valid, err = VerifyBLSSignatureManyMessages(inputPks, aggSig, inputMsgs, inputKmacs)
+		assert.Error(t, err)
+		assert.True(t, IsInvalidInputsError(err))
+		assert.False(t, valid,
+			fmt.Sprintf("verification should fail with nil hasher"))
+		inputKmacs[0] = tmp
+
+		// wromg key
```
jordanschalm, line 843:
```suggestion
		// wrong key
```

crypto/types.go
```diff
@@ -110,20 +111,22 @@ type Signature []byte
 // It allows a function caller differentiate unexpected program errors from errors caused by
 // invalid inputs.
 type InvalidInputsError struct {
```
jordanschalm, line 113: I think the constructor is private because we want these errors to only be produced from within this package. In that case this struct should be private as well to avoid direct instantiation.

crypto/types.go
```diff
@@ -110,20 +111,22 @@ type Signature []byte
...
-	message string
+	err error
 }
 
-// newInvalidInputsError constructs a new InvalidInputsError
-func newInvalidInputsError(msg string, args ...interface{}) error {
-	return &InvalidInputsError{message: fmt.Sprintf(msg, args...)}
+// invalidInputsErrorf constructs a new InvalidInputsError
+func invalidInputsErrorf(msg string, args ...interface{}) error {
```
jordanschalm, line 118: Just curious why the name change here? Using `new` to indicate a constructor makes sense and matches the other similar error constructors we have, eg: https://github.com/onflow/flow-go/blob/c3d239ee1534db5430ddd8b9a747f40063ad19ed/engine/errors.go#L25

jordanschalm (approved):
Thanks for the updates, looks good! A few minor notes but no blockers

crypto/bls_test.go
tarakby: Nice, I deleted the `Sprintf`s, though I noticed I didn't have to add the `f`.

crypto/types.go
tarakby: I also prefer to have it private. I thought having it public allows callers to test for the error type without using `IsInvalidInputsError`

crypto/types.go
tarakby: I just went for a more compact name, and it follows other libraries like `fmt.Errof` instead of `fmt.NewErrorf`.

PR #1556, opened 2021-10-29
refactor fallback logic for qc voter & dkg broker

kc1116:
This PR refactors the fallback logic located in the QC Voter & DKG Broker for client connections into common default fall back strategy.


module/fallbackstrategy.go
```diff
@@ -0,0 +1,11 @@
+package module
+
+// FallbackStrategy defines an interface to handle usage of multiple
+// clients connected to different hosts.
+type FallbackStrategy interface {
+	// ClientIndex return index of client to use
+	ClientIndex() int
+
+	// Failure indicate failure to use client at active index
+	Failure()
```
jordanschalm, line 10: We should include the client index being used here, otherwise we risk a race condition. The user of `FallbackStrategy` wants to tell `FallbackStrategy`, "I tried to talk to client X but it didn't work". If we don't pass in "X", which client we're referring to, we are requiring implicit synchronization between `FallbackStrategy`'s internal state and its user's state.

module/fallbackstrategy.go
```diff
@@ -0,0 +1,11 @@
...
+}
```
jordanschalm, line 11: A `Success(int)` method would be useful in general, and I think is required to fully reproduce the "fallback-after-2-consecutive-failures" logic we are trying to encapsulate in this component (otherwise we don't know when failures are consecutive).

module/util/util.go
```diff
@@ -6,6 +6,33 @@ import (
 	"github.com/onflow/flow-go/module"
 )
 
+// DefaultFallbackStrategy simple fallback strategy assuming clients are stored in a list
+// this strategy will manage the active client index
```
jordanschalm, line 10: To close the loop with the original comment, I think this component should fully implement the fallback logic (eg. if we observe 2 consecutive `Failure()` calls without an intervening `Success()` call, then increment the client index. I think we should also:
* give it a more descriptive name, eg: `AfterConsecutiveFailuresStrategy`
* update the godoc to explain the functionality in more detail than "manage the active client index", for example "cycle the active client in a round-robin manner each time n consecutive failures are observed for the same client"

jordanschalm: It would also probably make sense to move this into a implementation package similar to many of the other interfaces defined in `module`, eg `module/fallback/after_consecutive_failures.go`

module/util/util.go
```diff
@@ -6,6 +6,33 @@ import (
...
+type DefaultFallbackStrategy struct {
+	numOfClients int
+	clientIndex  int
```
jordanschalm, line 13: I suggest we protect this with a mutex. The two use-cases we have now are single-threaded, but it's easy to imagine uses of this component which would concurrently access a client index and invoke `Failure` from two separate threads.

jordanschalm (changes requested):
Thanks for the progress here. 

My concern is that this seems to be dividing the logic being refactored into two components, rather than cleanly encapsulating it fully in the new component (details in comments). This reduces the value of refactoring at all, in my opinion. I left some suggestions to more clearly draw the line between components here, and fully encapsulate the logic being refactored.

module/retrymiddleware/after_consecutive_failures.go
```diff
@@ -0,0 +1,48 @@
...
+// OnMaxConsecutiveFailures func that gets invoked when max consecutive failures has been reached
+type OnMaxConsecutiveFailures func(totalAttempts, clientIndex int)
+
+// AfterConsecutiveFailures returns a retry middleware that keeps track of the current number of attempts as well as the current client index.
+// When maxConsecutiveFailures has been reached the client index will be incremented or reset to 0 and the middleware will invoke f with the new
+// client index.
+func AfterConsecutiveFailures(maxConsecutiveFailures, numOfClients int, next retry.Backoff, f OnMaxConsecutiveFailures) retry.Backoff {
+	var l sync.Mutex
+
+	var attempt, client int
```
jordanschalm, line 18: This resets the client index to 0 each time we submit a new request. Suppose we have clients [a,b,c,d], and a and b are unavailable. Then we'll need to wait for failures on a and b each time before we can successfully send a request to c.

kc1116: Good call out, better to take into account **_last successful client_**

kc1116: Now that i'm thinking about it more, if we have to reset client index to 0 or we will be breaking the guarantee of client priority. We would also probably want those failure calls for metrics and logging.

jordanschalm: I think we should prioritize picking a client that is most likely to result in a successful request as quickly as possible, which in my mind would be that last client which was successful in a previous request.

kc1116: https://github.com/onflow/flow-go/pull/1556/commits/41cf3995759685b89b6b25c509fa5b02fefc0555

module/epochs/qc_voter.go
```diff
@@ -121,42 +123,53 @@ func (voter *RootQCVoter) Vote(ctx context.Context, epoch protocol.Epoch) error
...
 
 	return err
 }
 
-func (voter *RootQCVoter) qcContractClient() module.QCContractClient {
-	return voter.qcContractClients[voter.activeQCContractClient]
-}
+// updateContractClient will return the last successful client index by default for all initial operations or else
+// it will return the appropriate client index with respect to last successful and number of client.
+func (voter *RootQCVoter) updateContractClient(clientIndex int, init bool) (int, module.QCContractClient) {
```
jordanschalm, line 158: Rather than the `bool` argument, which changes the behaviour quite a bit, could we separate this into two functions?
```
updateContractClient(clientIndex) (int, module.QCContractClient) // cycles the active client, returns the new client and index
getInitialContractClient() module.QCContractClient // retrieves the active client
```

module/epochs/qc_voter.go
```diff
@@ -121,42 +123,53 @@ func (voter *RootQCVoter) Vote(ctx context.Context, epoch protocol.Epoch) error
...
+	if init {
+		return voter.lastSuccessfulClientIndex, voter.qcContractClients[voter.lastSuccessfulClientIndex]
+	}
 
-func (voter *RootQCVoter) updateActiveQcContractClient() {
-	// if we have reached the end of our array start from beginning
-	if voter.activeQCContractClient == len(voter.qcContractClients)-1 {
-		voter.activeQCContractClient = 0
-		return
+	if clientIndex == voter.lastSuccessfulClientIndex {
```
jordanschalm, line 164: To me it would be simpler / less code to use `voter. lastSuccessfulClientIndex` directly instead of the intermediary variable. That way we can encapsulate mutex-protected reads/writes of that variable in the `updateContractClient` and `getInitialContractClient` methods too. 

Not a blocker though.

jordanschalm (commented):
Final blocker: please make sure the shared `lastSuccessfulClientIndex` variable is concurrency-safe - I'm good to merge once that is done. A few non-blocker suggestions as well.

module/epochs/qc_voter.go
kc1116: It's a little more verbose but more explicit

module/epochs/qc_voter.go
kc1116: https://github.com/onflow/flow-go/pull/1556/commits/c7e2a246eb64337505deac805fd3fc71d6b3d9ef

module/retrymiddleware/after_consecutive_failures.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// OnMaxConsecutiveFailures func that gets invoked when max consecutive failures has been reached
+type OnMaxConsecutiveFailures func(currentAttempt int)
+
+// AfterConsecutiveFailures returns a retry middleware will invoke the OnMaxConsecutiveFailures callback when ever maxConsecutiveFailures
+// has been reached for a retry operation
+func AfterConsecutiveFailures(maxConsecutiveFailures int, next retry.Backoff, f OnMaxConsecutiveFailures) retry.Backoff {
+	var (
+		l       sync.Mutex
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1556/commits/c7e2a246eb64337505deac805fd3fc71d6b3d9ef

module/dkg/broker.go
```diff
@@ -148,21 +138,25 @@ func (b *Broker) Broadcast(data []byte) {
...
-					b.log.Warn().Msgf("broadcast: retrying on attempt (%d) with fallback access node at index (%d)", attempts, b.activeDKGContractClient)
-				}
+				attempts++
+				return retry.RetryableError(err)
 			}
 
-			attempts++
-			return retry.RetryableError(err)
+			// update our last successful client index for future calls
+			b.lastSuccessfulClientIndex = clientIndex
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1556/commits/c7e2a246eb64337505deac805fd3fc71d6b3d9ef

module/dkg/broker.go
```diff
@@ -148,21 +138,25 @@ func (b *Broker) Broadcast(data []byte) {
...
 		maxedExpRetry := retry.WithMaxRetries(retryMaxPublish, expRetry)
 
+		clientIndex, dkgContractClient := b.getInitialContractClient()
+		onMaxConsecutiveRetries := func(totalAttempts int) {
+			clientIndex, dkgContractClient = b.updateContractClient(clientIndex)
+			b.log.Warn().Msgf("broadcast: retrying on attempt (%d) with fallback access node at index (%d)", totalAttempts, clientIndex)
+		}
+		afterConsecutiveFailures := retrymiddleware.AfterConsecutiveFailures(retryMaxConsecutiveFailures, maxedExpRetry, onMaxConsecutiveRetries)
+
 		attempts := 1
```
durkmurder, line 148: do we need attempts only for logging?

kc1116: yes @tarakby requested we log the attempts when we were debugging DKG broadcast issues some time after retries were introduced.

durkmurder (approved):
Overall looks good. But be aware that I am not familiar with this part of code :)

module/epochs/qc_voter.go
```diff
@@ -58,6 +61,7 @@ func NewRootQCVoter(
 		state:             state,
 		qcContractClients: contractClients,
 		wait:              time.Second * 10,
+		unit:              engine.NewUnit(),
```
jordanschalm, line 64:
```suggestion
		mu:              sync.Mutex{},
```
We're only using the lock part of `Unit`, better to just use a lock directly

module/dkg/broker.go
```diff
@@ -312,6 +308,18 @@ func (b *Broker) updateContractClient(clientIndex int, init bool) (int, module.D
 	return clientIndex, b.dkgContractClients[clientIndex]
 }
 
+// getInitialContractClient will return the last successful contract client or the initial
+func (b *Broker) getInitialContractClient() (int, module.DKGContractClient) {
+	return b.lastSuccessfulClientIndex, b.dkgContractClients[b.lastSuccessfulClientIndex]
```
jordanschalm, line 313: ⚠️  Reads of the shared variable `lastSuccessfulClientIndex ` need to be synchronized as well. 

Currently the code will prevent concurrent _writes_ to `lastSuccessfulClientIndex`, but will allow concurrently one write and any number of reads. 

In general under the Go memory model, this results in undefined behaviour, from the [Go Memory Model](https://golang.org/ref/mem#tmp_10):
> Programs that modify data being simultaneously accessed by multiple goroutines must serialize such access.

module/dkg/broker.go
```diff
@@ -18,37 +18,43 @@ import (
...
+
+	// retryMaxConsecutiveFailures is the number of consecutive failures allowed before we attempt with fallback client
+	retryMaxConsecutiveFailures = 2
+)
 
 // Broker is an implementation of the DKGBroker interface which is intended to
 // be used in conjuction with the DKG MessagingEngine for private messages, and
 // with the DKG smart-contract for broadcast messages.
 type Broker struct {
 	sync.Mutex
```
durkmurder, line 43: It's a bit confusing why it has an embedded lock and then uses `unit` for locking as well.

module/dkg/broker.go
durkmurder: +1 on this, strictly speaking this might not a problem. Since the write is protected by `Mutex` we will observe changes in other goroutines. I haven't found proof for that behavior in memory model, but go core devs use this [trick](https://github.com/golang/go/blob/master/src/runtime/chan.go#L138) and expect such behavior.

Just to be clear, I am not saying that we should do this, we should just place an explicit lock and be sure that it's safe and obvious, this off-topic is more for knowledge sharing of interesting fact.

module/dkg/broker.go
kc1116: https://github.com/onflow/flow-go/pull/1556/commits/57c05d77a5b4485866c7036db4ad16e4ec2a9588

module/dkg/broker.go
kc1116: https://github.com/onflow/flow-go/pull/1556/commits/6f9ab2af46584eb4c8fbea55778896dcbe0e2d9c

jordanschalm (approved):

kc1116:
bors merge

PR #1557, opened 2021-11-01
Add arguments for custom ssl/tls certificate and key into BaseFlags for node_builder

skwcheung-circle:
Related Issue: https://github.com/onflow/flow-go/issues/1465

Context:
Enable node_builder to accept arguments for a tls certificate and public key to use for server authentication instead of defaulting to always generating a certificate through `grpcutils.X509Certificate(node.NetworkKey)`.

This enables clients to maintain a static list of certificates for authenticating. It also enables passing in a certificate with a non-empty issuer field that was causing problems when validating through java libraries (see linked issue).

Code Changes:
- Add `TlsCert` and `TlsKey` fields into `BaseConfig`
- Add argument flags into `BaseFlags()` to pass in file paths of custom tls cert and public key .pem files
- Check for certificate/key passed in by arguments or default to the existing implementation of randomly generating a certificate

Testing:
Locally tested code changes using localnet and loaded in `.pem` cert + key into `access_node` container using `docker-compose.nodes.yml`. Was able to retrieve the custom certificate when connecting to local net through `openssl s_client -showcerts -connect host.docker.internal:3572`

Relevant Places in Code:
- Read in key/cert values and built key the same way as existing `admin-cert` value https://github.com/onflow/flow-go/blob/967557bb010d2bb9dd2854846278ba5becad1e73/cmd/scaffold.go#L304
- `grpcutils.x509Certificate()` https://github.com/onflow/flow-go/blob/4f4e9801b56c5198d3d7c3fe1291c52987d51e55/utils/grpcutils/grpc.go#L32
- `x509Certificate()` uses `NewIdentity()` https://github.com/libp2p/go-libp2p-tls/blob/783f916acec24dc2b13c3451a365dde648202826/crypto.go#L40
- `NewIdentity()` uses `keyToCertificate()` https://github.com/libp2p/go-libp2p-tls/blob/783f916acec24dc2b13c3451a365dde648202826/crypto.go#L160 to randomly generate a certificate


peterargue:
Thanks for contributing this PR @skwcheung-circle!

I think it makes sense to allow operators to provide their own certificates.  One caveat though, is that the cert must have the special extension that ties the cert to the libp2p networking key and is used by the SN/LNs to authenticate their connection to the staked AN. So in addition to this PR, we'll need to add some tooling to generate and persist the key pair, and possibly also a CSR that includes the extension.

peterargue:
I'm working on getting an update made to the libp2p library used to generate the secured gRPC TLS certificates, that's needed to create the tool mentioned in my last comment. Hopefully I'll be able to make some progress on this soon.

Kay-Zee:
@peterargue Any updates here?

peterargue:
yea, the libp2p PR this depends on is planned for the next libp2p release. after that we can release the tooling required to generate the tls keys.

peterargue:
PR https://github.com/onflow/flow-go/pull/3789 adds the bootstrap command to generate the key pairs including both the libp2p extension and a custom list of SANs.

franc-crypt0-s3curity (commented):
Do we want to enforce certain minimal security requirements on which hash functions we'd allow? 
(in a secure by default model, I'd recommend we should at least have a deny list of hash functions,  that are too weak to be used wrt recent computation capabilities). CC @tarakby  thoughts?

Should we have any check on the FQDN  or IP address or certificate use case in the certificate? (eg to limit human error) 
CC @tarakby

peterargue:
I don't think we need the IP/FQDN checks since this is intended to allow operators to setup a custom cert for their own internal usecases. The minimum requirements seems like a good idea. any recommendations on which settings and values we should use?

tarakby (commented):
Thanks for the changes!
I left a few questions to understand the context better. I'm also wondering if we should be adding tests for the commands and new code.

tarakby:
I agree with @franc-crypt0-s3curity in checking the signing algorithms on the certificate (which includes the hashing used). I think there is (1) a list of signing algorithms the private key is authorized to be used for and (2) another list of signing algorithms that can be used to verify external certificates. 

The Go tls lib supports [this list](https://pkg.go.dev/crypto/tls#SignatureScheme), but it would be good to speak to @peterargue to better understand what the node needs for (1) and (2).

peterargue:
@skwcheung-circle I know this PR is very old now, but are those changes something you would be interested in adding? If not we can close this PR and open a new one with the changes.

peterargue:
Per my last message, I'm going to close this PR for now. Feel free to reopen or open a new one. I will circle back later when I have some spare cycles.

PR #1560, opened 2021-11-02
ledger pruning exploration

AlexHentschel:
Exploration if it is possible to avoid heavy GC on pruning


AlexHentschel:
this is still work in progress

PR #1564, opened 2021-11-02
[Consensus] Continue recover block when hitting invalid block

zhangchiqing:
This PR fixes a bug that when hitting an invalid block during recovery, it should continue recover other blocks rather than stop


Kay-Zee (approved):

Kay-Zee:
We should consider adding a test here when we have time

AlexHentschel (approved):

PR #1569, opened 2021-11-03
Epoch Emergency Fallback Improvements

jordanschalm:
This PR extends epoch emergency fallback to cover more failure cases, and address some inconsistencies.

* When EECC is triggered, explicitly flag this in the database
* Trigger EECC when incorporating an invalid service event
* Do not process any further service events after EECC is triggered
* Do not emit any epoch-related protocol event after EECC is triggered

PR to allow new metric: https://github.com/dapperlabs/dapper-flow-hosting/pull/344


state/protocol/badger/mutator_test.go
```diff
@@ -973,7 +973,8 @@ func TestExtendEpochSetupInvalid(t *testing.T) {
 			return setup, receipt, seal
 		}
```
kc1116, line 975: Should we also add a test to make sure epoch related metrics dont get updated while EECC is triggered ?

kc1116 (commented):
LGTM - minor comments

state/protocol/badger/mutator.go
```diff
@@ -770,6 +780,7 @@ func (m *FollowerState) handleServiceEvents(block *flow.Block) ([]func(*transact
 	// state to go to the next epoch when needed. In cases where there is a bug
 	// in the smart contract, it could be that this happens too late and the
 	// chain finalization should halt.
+SealLoop:
```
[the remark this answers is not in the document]
jordanschalm: It's a [labelled statement](https://medium.com/golangspec/labels-in-go-4ffd81932339)

state/protocol/badger/validity.go
```diff
@@ -99,6 +136,37 @@ func verifyEpochSetup(setup *flow.EpochSetup, verifyNetworkAddress bool) error {
...
 
+// isValidExtendingEpochCommit checks whether an epoch commit service being
+// added to the state is valid. In addition to intrinsic validity, we also
+// check that it is valid w.r.t. the previous epoch setup event, and the
+// current epoch status.
+func isValidExtendingEpochCommit(extendingCommit *flow.EpochCommit, extendingSetup *flow.EpochSetup, activeSetup *flow.EpochSetup, status *flow.EpochStatus) error {
+
+	// We should only have a single epoch commit event per epoch.
+	if status.NextEpoch.CommitID != flow.ZeroID {
+		// true iff EpochCommit event for NEXT epoch was already included before
```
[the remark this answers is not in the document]
jordanschalm: iff is [shorthand](https://en.wikipedia.org/wiki/If_and_only_if) for "if and only if"

state/protocol/badger/mutator_test.go
jordanschalm: Added in https://github.com/onflow/flow-go/pull/1569/commits/e31c7ecbf7b6adb8de573a5820008e2feed36029 https://github.com/onflow/flow-go/pull/1569/commits/152aa5ac69bb33302f6d36f58f2b071a68229f6f

kc1116 (approved):

state/protocol/badger/mutator.go
```diff
@@ -781,30 +792,13 @@ func (m *FollowerState) handleServiceEvents(block *flow.Block) ([]func(*transact
...
-				// Finally, the epoch setup event must contain all necessary information.
-				err = isValidEpochSetup(ev)
-				if err != nil {
-					return nil, state.NewInvalidExtensionErrorf("invalid epoch setup: %s", err)
+				// validate the service event
+				err := isValidExtendingEpochSetup(ev, activeSetup, epochStatus)
+				if protocol.IsInvalidServiceEventError(err) {
+					// EECC - we have observed an invalid service event, which is
+					// an unrecoverable failure. Flag this in the DB and exit
+					ops = append(ops, transaction.WithTx(operation.SetEpochEmergencyFallbackTriggered()))
```
zhangchiqing, line 800: even tho we are storing the event in the database, it's still better to log it.

state/protocol/badger/mutator.go
```diff
@@ -815,30 +809,17 @@ func (m *FollowerState) handleServiceEvents(block *flow.Block) ([]func(*transact
...
 					return nil, state.NewInvalidExtensionErrorf("could not retrieve next epoch setup: %s", err)
 				}
-				err = isValidEpochCommit(ev, setup)
-				if err != nil {
-					return nil, state.NewInvalidExtensionErrorf("invalid epoch commit: %s", err)
+				// validate the service event
+				err = isValidExtendingEpochCommit(ev, extendingSetup, activeSetup, epochStatus)
+				if protocol.IsInvalidServiceEventError(err) {
+					// EECC - we have observed an invalid service event, which is
+					// an unrecoverable failure. Flag this in the DB and exit
```
zhangchiqing, line 820: same here, better to log

state/protocol/badger/mutator.go
```diff
@@ -815,30 +809,17 @@ func (m *FollowerState) handleServiceEvents(block *flow.Block) ([]func(*transact
...
-				setup, err := m.epoch.setups.ByID(epochStatus.NextEpoch.SetupID)
+				extendingSetup, err := m.epoch.setups.ByID(epochStatus.NextEpoch.SetupID)
 				if err != nil {
 					return nil, state.NewInvalidExtensionErrorf("could not retrieve next epoch setup: %s", err)
 				}
-				err = isValidEpochCommit(ev, setup)
-				if err != nil {
-					return nil, state.NewInvalidExtensionErrorf("invalid epoch commit: %s", err)
+				// validate the service event
+				err = isValidExtendingEpochCommit(ev, extendingSetup, activeSetup, epochStatus)
```
zhangchiqing, line 817: This is checked when a block is finalized, if it's invalid extension we still have to accept it because this block has been finalized. But storing a finalized block with invalid service event is weird to me.

Why no checking when a block is incorporated and reject if is invalid? So that every finalized block must have passed that check

state/protocol/badger/state.go
```diff
@@ -599,3 +609,9 @@ func (state *State) updateCommittedEpochFinalView(snap protocol.Snapshot) error
 
 	return nil
 }
+
+func (m *State) isEpochEmergencyFallbackTriggered() (bool, error) {
+	var triggered bool
```
zhangchiqing, line 614: EECC triggered or not is only checked on a finalized block, that means it won't trigger again on other forks, it should only be triggered once. 

I wonder if it's useful to store a block ID or block height of the block that triggered EECC. Would it provide more context than a bool?

zhangchiqing (approved):
Looks good. some minor suggestions

state/protocol/badger/mutator.go
jordanschalm: I agree, but the `FollowerState` doesn't have reference to a logger at all currently (we don't do any logging in this module). Do you think we should add logger to `FollowerState` for this?

state/protocol/badger/state.go
jordanschalm: 👍  good idea - will add this

state/protocol/badger/mutator.go
jordanschalm: Here's how I think about it: It's tricky because an invalid service event is a bit different than an invalid message from another node. It's different because the originator of the message (service account) is trusted.

If we see an invalid service event, either: (1) there is a bug in the service account, or (2) the service event was maliciously modified or injected by a malicious execution node.

If (1), then all honest ENs and VNs will agree on the state resulting from emission of the invalid service event, and that state will be sealed. If we discard receipts and approvals that attest to this state, sealing will halt. So we need to incorporate the service event anyway if we want to preserve liveness. 

If (2), then the service event won't be sealed because verification of the result its part of will fail. So it is safe to incorporate the malicious result. We don't _apply_ state changes resulting from service events until after they are sealed for exactly this reason.

state/protocol/badger/state.go
jordanschalm: https://github.com/onflow/flow-go/pull/1569/commits/5a5b7e0ae9115b059b3f1a9714397c9fb0593674

jordanschalm:
bors merge

PR #1572, opened 2021-11-03
Rewrite test for state diff storer

smnzhu:
Removes the GCS bucket test and replaces it with one that can actually run in CI.

### DIFF
--- a/engine/execution/state_synchronization/state_diff_storer_test.go
+++ b/engine/execution/state_synchronization/state_diff_storer_test.go
@@ -1,116 +1,49 @@
 package state_synchronization
 
 import (
-	"context"
 	"reflect"
 	"testing"
-	"time"
 
-	"cloud.google.com/go/storage"
-	"github.com/fxamacker/cbor/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


engine/execution/state_synchronization/state_diff_storer_test.go
```diff
@@ -1,116 +1,49 @@
...
-		duration = time.Since(start)
-		require.NoError(t, err)
-		t.Logf("time to load state diff: %v\n", duration)
-		if duration > maxLoadTime {
-			maxLoadTime = duration
-		}
-
-		assert.True(t, reflect.DeepEqual(sd, sd2))
-	}
+	sd := generateExecutionStateDiff(t, 0)
```
peterargue, line 36: wont this generate a diff with no `TrieUpdates`?

smnzhu: Yes, it doesn't really matter because from the point of view of the state diff storer, it all just gets serialized to a blob of data in the end anyways.

The main things I want to test here are the case where the state diff is small and all fits into a single block, and the one where it's large and needs to be split into multiple blocks, hence generating intermediate CIDs.

peterargue (approved):

PR #1573, opened 2021-11-04
Smnzhu/sync engine optimizations

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/5934

This PR:
* Include block height when queueing blocks by ID. This allows us to prune requests by ID when a new block is finalized.
* Separate range and batch responses. We only update the status for block IDs / heights when we receive batch / range responses. In particular, we avoid clearing a height request just because we receive a block at the same height.
* Modify the status tracking logic so that we only delete requests that have reached the max number of attempts if we haven't already received a finalized block with higher height. If we have already received a finalized block of higher height, it is safe to keep requesting this block indefinitely because we know that it must exist.

TODO:
- [ ] Add tests


engine/common/synchronization/engine.go
```diff
@@ -271,10 +295,17 @@ func (e *Engine) processAvailableResponses() {
...
+		msg, ok = e.pendingBatchResponses.Get()
 		if ok {
-			e.onBlockResponse(msg.OriginID, msg.Payload.(*messages.BlockResponse))
-			e.metrics.MessageHandled(metrics.EngineSynchronization, metrics.MessageBlockResponse)
+			e.onBatchResponse(msg.OriginID, msg.Payload.(*messages.BatchResponse))
+			e.metrics.MessageHandled(metrics.EngineSynchronization, metrics.MessageBatchResponse)
+			continue
+		}
+
+		msg, ok = e.pendingRangeResponses.Get()
```
jordanschalm, line 305: Shouldn't we prioritize range responses? Blocks in range requests typically need to be processed before batch requests. (Likewise for the collection node sync engine)

module/metrics/labels.go
```diff
@@ -100,8 +100,9 @@ const (
 	MessageSyncRequest          = "ping"
 	MessageSyncResponse         = "pong"
 	MessageRangeRequest         = "range"
+	MessageRangeResponse        = "range_response"
 	MessageBatchRequest         = "batch"
```
jordanschalm, line 104:
```suggestion
	MessageRangeRequest         = "range_request"
	MessageRangeResponse        = "range_response"
	MessageBatchRequest         = "batch_request"
```

module/synchronization/core.go
```diff
@@ -47,31 +48,52 @@ func DefaultConfig() Config {
...
+	if !ok {
+		return nil
+	}
+	return s.Status
+}
+
 // HandleBlock handles receiving a new block from another node. It returns
 // true if the block should be processed by the compliance layer and false
 // if it should be ignored.
 func (c *Core) HandleBlock(header *flow.Header) bool {
```
jordanschalm, line 92: We should consider adjusting the naming or documentation here. This generically-named function now specifically assumes that the block was requested by ID (ie. if I request a block by height, then call `HandleBlock`, it will not be processed properly).

Similarly, `HandleFinalizedBlock` assumes the block was requested by height.

module/synchronization/core.go
```diff
@@ -287,8 +344,9 @@ func (c *Core) getRequestableItems() ([]uint64, []flow.Identifier) {
...
 		}
 
-		// if we reached maximum number of attempts, delete
-		if status.Attempts >= c.Config.MaxAttempts {
+		// if we reached maximum number of attempts and we're not certain that the height
+		// has actually been finalized, delete
+		if height > c.maxHeightReceived && status.Attempts >= c.Config.MaxAttempts {
 			delete(c.heights, height)
 			continue
 		}
```
jordanschalm, line 352: > Modify the status tracking logic so that we only delete requests that have reached the max number of attempts if we haven't already received a finalized block with higher height

I think the reason for this change is to make sure we don't _stop requesting a block height_ by deleting it. But when we delete a height status like this, if we haven't received the block yet, practically it will just reset the request retry timer (the height will be re-queued [here](https://github.com/onflow/flow-go/blob/982a0a2f01d2aff611d19c353f2b361408de6476/module/synchronization/core.go#L98-L112)). So I'm not sure there's any reason to keep track of `maxHeightReceived`. We could add a note about deleted height statuses to the docs though.

jordanschalm (commented):
A few questions and suggestions

engine/common/synchronization/engine.go
smnzhu: Yes, that's correct. Thanks

module/synchronization/core.go
smnzhu: That's correct, but that assumes that we will eventually receive another sync response from another node which triggers the `HandleHeight` at a higher height. In practice this is probably a safe assumption, so I don't mind removing this.

jordanschalm: If we don't receive sync responses from other nodes, we never send range requests at all and the engine doesn't really work, so I think it's a safe assumption.

module/synchronization/core.go
```diff
@@ -300,7 +358,6 @@ func (c *Core) getRequestableItems() ([]uint64, []flow.Identifier) {
 	// create list of all the block IDs blocks that are missing
 	var blockIDs []flow.Identifier
 	for blockID, status := range c.blockIDs {
```
zhangchiqing, line 360: One thing I'm concerned with this `range` is that:
When a node is falling far behind, it will request both by height and by blockIDs, and this blockIDs will keep growing because it takes time for the finalized height to go up to prune it. This will  cause `range c.blockIDs` take longer and longer.

For instance, if finalized height is 1000, and it was asked to fetch blocks from [1000000-2000000], then `c.blockIDs` will have 1000000 items that won't be pruned until the finalization has caught up.

module/synchronization/core.go
```diff
@@ -237,13 +298,9 @@ func (c *Core) prune(final *flow.Header) {
...
-		if status.WasReceived() {
-			header := status.Header
-
-			if header.Height <= final.Height {
-				delete(c.blockIDs, blockID)
-				continue
-			}
+		if status.WasReceived() && status.height <= final.Height {
+			delete(c.blockIDs, blockID)
+			continue
```
zhangchiqing, line 303: The pruning could exit earlier if we iterate from low to high through a sorted heights.

However, both c.blockIDs and c.height are `map`, if they were sorted slice, we could `break` the loop as soon as the height is above the finalized height.

PR #1579, opened 2021-11-04
Add github action to make a benchstat comparison of a PR with its base

janezpodhostnik:
This github action runs bench tests on the FVM on the PR and the base branch of the PR and compares the results with benchstat.

This is only done if anything in the FVM changes, or if the `go.sum` changes.

I have tested this action on a forked repo, and an example of the output can be seen here: https://github.com/janezpodhostnik/flow-go/pull/1

If this runs too long, the number of repetitions can be reduced (`run: echo "::set-output name=benchmark_repetitions::7"`). It currently runs for 17 minutes, which is about the same as the other jobs.


ramtinms (approved):

Kay-Zee (approved):

janezpodhostnik:
bors merge

janezpodhostnik:
bors merge

PR #1580, opened 2021-11-04
Kan/identifier merkle root fix

Kay-Zee:
This PR is to address a bug related to the generation of Hashes for the intention of making an Identifier.

The MerkleRoot function, where the bug manifests, is used for hashing the unique IDs of the contents of a block Payload together.

While we do take SOME data from each ID to generate the hash, (e.g. we follow the correct "path" down the merkle tree), the value that was saved as the key of the leaf would actually get changed due to the fact that it's making use of the same memory space that go is using for the for loop that is passing the data in.

While we don't feel like this is a security vulnerability, and we're not worried about collisions with this setup, it does mean the hashes are NOT generated how we want them to be currently, and this PR fixes that.

We also add a "value" in the node, which didn't previously exist, which will set the order of the IDs, which is a property that we wanted to add.

### DIFF
--- a/storage/merkle/tree.go
+++ b/storage/merkle/tree.go
@@ -144,8 +144,11 @@ PutLoop:
 			// if we have reached the end of the key, insert the new value
 			totalCount := uint(len(key)) * 8
 			if index == totalCount {
+				// Make a copy of the key, to ensure we have the only pointer to it
+				keyCopy := make([]byte, len(key))
+				copy(keyCopy, key)
 				*cur = &leaf{
-					key: key,
+					key: keyCopy,
 					val: val,
 				}
 				return false
--- a/model/flow/identifier.go
+++ b/model/flow/identifier.go
@@ -145,8 +145,13 @@ func GetIDs(value interface{}) []Identifier {
 func MerkleRoot(ids ...Identifier) Identifier {
 	var root Identifier
 	tree := merkle.NewTree()
-	for _, id := range ids {
-		tree.Put(id[:], nil)
+	for i, id := range ids {
+		// NOTE: the id[:] passed in here will be the address of byte array
+		// that is used by Go in the loop, so it MUST be copied to a different
+		// slice if the intention is to store it's value. Otherwise,
+		// all the value will be the last item of the ids slice.
+		// This is not a problem for i, due to it being a primative (int)
+		tree.Put(id[:], i)
 	}
 	hash := tree.Hash()
 	copy(root[:], hash)
@@ -173,7 +178,7 @@ func CheckConcatSum(sum Identifier, fps ...Identifier) bool {
 }
 
 // Sample returns random sample of length 'size' of the ids
-// [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher–Yates_shuffle).
+// [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher-Yates_shuffle).
 func Sample(size uint, ids ...Identifier) []Identifier {
 	n := uint(len(ids))
 	dup := make([]Identifier, 0, n)
--- a/model/flow/identifier_test.go
+++ b/model/flow/identifier_test.go
@@ -101,12 +101,14 @@ func TestMerkleRoot(t *testing.T) {
 
 }
 
+// We should ideally replace this with a completely different reference implementation
+// Possibly writen in another language, such as python, similar to the Ledger Trie Implementation
 func referenceMerkleRoot(ids ...flow.Identifier) flow.Identifier {
 	var root flow.Identifier
 	tree := merkle.NewTree()
-	for _, id := range ids {
-		idCopy := id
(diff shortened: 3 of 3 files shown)


AlexHentschel:
see slack thread for further discussion: https://axiomzen.slack.com/archives/C02L17LGJ7Q

storage/merkle/tree.go
```diff
@@ -144,8 +144,11 @@ PutLoop:
 			// if we have reached the end of the key, insert the new value
 			totalCount := uint(len(key)) * 8
 			if index == totalCount {
+				// Make a copy of the key, to ensure we have the only pointer to it
+				keyCopy := make([]byte, len(key))
+				copy(keyCopy, key)
```
tarakby, line 149: I wrote a simple bench to check if this is faster:
```
keyCopy := make([]byte, 0, len(key))
keyCopy = append(keyCopy, k...)
```
it turns out it is 6% faster for `len(key) = 32` (our case, as we computed the hash for `Identifier`s). It is not the case for other key lengths though.

AlexHentschel:
took a proper pass over the existing implementation and put up a PR with suggested revisions: https://github.com/onflow/flow-go/pull/1678 (targeting this branch)

AlexHentschel (approved):
Looks good to me. I think we can merge this; though there are still some open security vulnerabilities, which I tried to address in my follow-up PR https://github.com/onflow/flow-go/pull/1678

storage/merkle/tree.go
bluesign: @tarakby any chance you can share benchmark code? append winning vs copy on small size is very interesting.

tarakby:
```
func BenchmarkAppendCopy(b *testing.B) {

	const N = 32
	key := make([]byte, N)
	rand.Read(key)

	b.Run("copy", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			keyCopy := make([]byte, N)
			copy(keyCopy, key)
		}
		b.StopTimer()
	})

	b.Run("append", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			keyCopy := make([]byte, 0, N)
			keyCopy = append(keyCopy, key...)
		}
		b.StopTimer()
	})
}
```

Kay-Zee: that's using a slice, i wonder if the results would differ at all if we used arrays, since flow.Identifier is actually a byte array.

tarakby: `flow.Identifier` is an array indeed, but the key type of the tree is a slice. The arrays are converted into a slice when `Put` is called.

AlexHentschel: Working with variable-length slices as keys https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/tree.go#L28 is an entirely different cans of worms 😓 :
* From a first glance, I think this tree only works if **all keys have the same length**. We use it in this manner, hence it is safe from this perspective, but the implementation does not enforce keys having the same length. As an example of what will go wrong, if there are keys of different length, consider the following scenario 
  * there are two keys `k1` and `k2 = k1 + κ` (`k2` is just an extension of `k1` with an additional tailing bit-string `κ`)
  * a key-value pair `p1` = (`k1`, `v1`) is already stored in the tree (for simplicity assume that it is the only entry in the tree)
      * the path to `p1`  is stored as a `short` node https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/node.go#L11-L15  with `path = k1` 
      * the `short`-node's `child` is the `leaf` holding  (`k1`, `v1`) : https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/node.go#L22-L25
  * now we add the key-value pair  `p2` = (`k2`, `v2`):
      * this is the specific part of interest: https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/tree.go#L78-L84 here, the algorithm concludes that the common prefix between `k1` and `k2` is the entire length of `k1` 
      * then, in the next [`for`-loop iteration](https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/tree.go#L44-L45), it replaces `p1.value` by `v2`: https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/tree.go#L132-L139 🔥 🧨 💀

tarakby: Yes that's a good point @AlexHentschel. The tree is safe in the way we use it, but nothing in the implementation is enforcing we do not misuse it.

Kay-Zee:
is now covered by #1800

Issue #1452, opened 2021-10-06
[State sync engine] ExecutionData requester engine


smnzhu:
Implement a state requester engine that follows Incorporated / Sealed block notifications from the [`FinalizationDistributor`](https://github.com/onflow/flow-go/blob/ed2c7986fe2505d65021e204e14aad8b40a59bc2/consensus/hotstuff/notifications/pubsub/finalization_distributor.go), and makes requests for the ExecutionData CIDs of the Execution Results in those blocks 
- It may need to maintain an index similar to the sync core and keep retrying requests for ExecutionData that have not yet been received.
- To get better latency, we can trigger the requests immediately upon receiving each new ExecutionReceipt. Then, once a block has actually been sealed, we can keep only the ExecutionData which was sealed.

franklywatson:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Flow-Protocol-5a2ad022cee40e25b0885f82/issues/onflow/flow-go/1452?planning-poker) with ZenHub @vishalchangrani @peterargue @smnzhu

smnzhu:
@peterargue After our discussion today, I realize that this engine actually doesn't need to communicate directly with the network. 

The initial version just needs to subscribe to the [Finalization Distributor](https://github.com/onflow/flow-go/blob/ed2c7986fe2505d65021e204e14aad8b40a59bc2/consensus/hotstuff/notifications/pubsub/finalization_distributor.go) and go through all the seals in each finalized block to request all the sealed Execution Data CIDs.

In the optimized version, we can modify the [Ingestion engine](https://github.com/onflow/flow-go/blob/0e845f855216f56be453426a6b2bfeebf57c49a7/engine/access/ingestion/engine.go#L111) to forward execution receipts to this new engine via `SubmitLocal`, which can optimistically trigger download of the Execution Data. We continue to subscribe to block finalizations, and clear the datastore TTL for sealed Execution Data (allowing the rest to expire on their own).

PR #1582, opened 2021-11-05
[State Sync] Execution data requester

smnzhu:
Implements part of https://github.com/onflow/flow-go/issues/1452

TODO:
- [x] Write tests
- [x] Incorporate provider system https://github.com/ipfs/go-ipfs-provider
- [x] Handle duplicate CIDs in get
- [x] add specific CID to error context
- [x] Add metrics
- [x] Add logs
- [x] Comments
- [x] Define interface


module/state_synchronization/execution_data_requester.go
```diff
@@ -0,0 +1,111 @@
...
+	buf := make([]byte, (len(cids)-1)*MAX_BLOCK_SIZE)
+	var lastBlock blocks.Block
+
+	done, err := session.GetBlocks(cids...).ForEach(func(b blocks.Block) {
+		i := indices[b.Cid()]
+		if i == len(cids)-1 {
+			lastBlock = b
+			return
+		}
+		start := i * MAX_BLOCK_SIZE
```
peterargue, line 54: similar to comment on the serialization header byte, is it safe to assume we'll get exactly `MAX_BLOCK_SIZE` bytes each complete block?

peterargue (commented):
this is looking good. Added a few comments.

This could use more inline comments throughout to make it easier to understand quickly.

module/state_synchronization/execution_data_storer.go
```diff
@@ -109,44 +77,31 @@ func (s *StateDiffStorer) Store(sd *ExecutionStateDiff) (cid.Cid, error) {
...
 }
 
-// Load loads the ExecutionStateDiff represented by the given CID from the blockstore.
+// Load loads the ExecutionData represented by the given CID from the blockstore.
 // Since blocks are limited to MAX_BLOCK_SIZE bytes, it's possible that the data was
-// stored in multiple chunks, and hence the root CID may point to a block for which
-// the data itself is a list of concatenated CIDs.
-func (s *StateDiffStorer) Load(c cid.Cid) (*ExecutionStateDiff, error) {
+// stored in multiple blocks, and hence the root CID may point to a block for which
+// the data itself represents a recursive list of CIDs.
```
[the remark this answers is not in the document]
smnzhu: So, the implementation is actually intended to handle the general case where:

Starting from the root CID, we load the block which contains a chunk of RawBytes.

These RawBytes may represent an ExecutionData, or it may represent a list of CIDs. In the latter case, we load the blocks from the list and concatenate the data into a new chunk of RawBytes. Then we are back to where we started, because this new RawBytes could either represent an ExecutionData, or another list of CIDs.

So, in theory, this tree of CIDs could grow infinitely deep, and the implementation is supposed to handle this using the for loop. In practice, the tree will be at most 2 levels deep assuming the largest ExecutionData we will ever see is ~400MB.

That's why I used the word "recursive", but if this is confusing I'm open to suggestions on how to reword it

module/state_synchronization/serializer.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// serializer encapulates the serialization and deserialization of data
+type serializer struct {
+	codec      encoding.Codec
+	compressor network.Compressor
+}
+
+func (s *serializer) serialize(w writer, v interface{}) error {
+	if err := w.WriteByte(getCode(v)); err != nil {
```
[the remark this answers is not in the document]
smnzhu: So, we don't do the splitting of data into blocks until after it has been serialized. As you see, the `serializer` takes in the `BlockWriter` as input and writes the serialized data to the `BlockWriter`. Therefore, it doesn't really make sense to talk about blocks being shifted by one byte because from the perspective of the `BlockWriter`, it is just chunking an arbitrary sequence of bytes into 1mb blocks.

One way to think of it is to just view the serializer as implementing some custom serialization algorithm. Then, all we are doing is we are taking the serialized representation of an object, and then chunking it into blocks, which is precisely what we are supposed to.

Under the hood, this "custom serialization algorithm" is just adding an extra byte to the beginning of the CBOR serialization of the object, but this is just an implementation detail really.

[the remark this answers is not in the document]
smnzhu: Furthermore, there's unfortunately no way to include "metadata" in a block (at least if we're using the IPFS implementation). The only information associated with a block is the raw bytes themselves (from which a CID may be derived), nothing more and nothing less.

module/state_synchronization/execution_data_storer.go
```diff
@@ -216,7 +182,7 @@ func (bw *BlockWriter) Write(p []byte) (n int, err error) {
 		}
 	}
 	if len(p) > 0 {
-		bw.buf = make([]byte, 0, bw.maxBlockSize)
+		bw.buf = make([]byte, 0, len(p))
```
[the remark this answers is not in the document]
smnzhu: Either one will work. If more data needs to be added to the buffer in a subsequent write, it will just be appended to the array and the array will automatically resize if it needs to.

Using `bw.maxBlockSize` means that if we do have subsequent writes before flushing, we won't need to resize the array.

On the other hand, if we don't have any subsequent writes (ie Flush is called right after this call to Write returns), then we overallocated memory for the array by allocating space for `bw.maxBlockSize` when we only needed `len(p)`.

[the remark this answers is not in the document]
smnzhu: So it just depends how many times Write is called on average before each Flush. If the number is 1, then using `len(p)` is better. Otherwise, using `bw.maxBlockSize` may be better.

module/state_synchronization/execution_data_requester.go
smnzhu: It's safe to assume this if the node which served the block was following the block chunking algorithm properly, because that logic (implemented in BlockWriter) always fills a block with MAX_BLOCK_SIZE bytes before starting a new block.

smnzhu: If they weren't following the block chunking algorithm properly, then the root CID they generate would be different anyways, since the CIDs of the individual blocks would be different.

module/state_synchronization/execution_data_requester.go
```diff
@@ -0,0 +1,111 @@
...
+
+	buf := make([]byte, (len(cids)-1)*MAX_BLOCK_SIZE)
+	var lastBlock blocks.Block
+
+	done, err := session.GetBlocks(cids...).ForEach(func(b blocks.Block) {
+		i := indices[b.Cid()]
+		if i == len(cids)-1 {
+			lastBlock = b
+			return
+		}
```
zhangchiqing, line 53: Why treating the last block so special? 

Why not just copying the raw bytes into `buf`?

module/state_synchronization/execution_data_storer.go
```diff
@@ -191,6 +146,17 @@ func (bw *BlockWriter) writeBlock(data []byte) error {
 	return nil
 }
 
+func (bw *BlockWriter) WriteByte(c byte) error {
+	bw.buf = append(bw.buf, c)
+	if len(bw.buf) >= bw.maxBlockSize {
```
zhangchiqing, line 151: should we check this before the append?

smnzhu:
Made some pretty extensive changes, please take another pass. Thanks

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+// Add constructs a blob tree for the given ExecutionData and adds it to the blobservice, and then returns the root CID.
+func (s *ExecutionDataService) Add(ctx context.Context, sd *ExecutionData) (cid.Cid, error) {
+	cids, err := s.addBlobs(ctx, sd)
+
+	if err != nil {
+		return cid.Undef, fmt.Errorf("failed to add execution data blobs: %w", err)
+	}
+
+	for {
+		if len(cids) == 1 {
```
zhangchiqing, line 124: how to prevent from endless loop? would `len(cids)` never equal to `1`?

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+	if recvErr != nil {
+		return nil, recvErr
+	}
+
+	return cids, serializeErr
+}
+
+// Add constructs a blob tree for the given ExecutionData and adds it to the blobservice, and then returns the root CID.
+func (s *ExecutionDataService) Add(ctx context.Context, sd *ExecutionData) (cid.Cid, error) {
+	cids, err := s.addBlobs(ctx, sd)
```
zhangchiqing, line 117: is it guaranteed that `len(cids)` would always be `>=1`?

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+	cachedBlobs := make(map[cid.Cid]network.Blob)
+
+	for _, c := range cids {
+		var err error
+		blob, ok := cachedBlobs[c]
+
+		if ok {
+			delete(cachedBlobs, c)
+		} else {
+			blob, err = s.findBlob(blobChan, c, cachedBlobs)
```
zhangchiqing, line 150: could you explain in comment how the `findBlob` works along with the for loop over `cids`?

The findBlob iterates over blobChan, the questions are:
1. is the `blobChan` a closed channel? if so, please comment about it. Otherwise the operation is not safe
2. `findBlob` will iterate over the `blobChan` in each iteration, it's O(N^2), can we iterate over `blobChan` first before iterate over `cids`?

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+func (bw *BlobChannelWriter) write(p []byte) (int, error) {
+	n := bw.maxBlobSize - len(bw.buf)
+
+	if n > len(p) {
+		n = len(p)
+	}
+
+	bw.buf = append(bw.buf, p[:n]...)
+
+	if len(bw.buf) >= bw.maxBlobSize {
```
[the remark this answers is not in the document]
smnzhu: It will be sent when the writer is closed.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		if cids, err = s.addBlobs(ctx, cids); err != nil {
+			return cid.Undef, fmt.Errorf("failed to add cid blobs: %w", err)
+		}
+	}
+}
+
+func (s *ExecutionDataService) sendBlobs(parent context.Context, bs *BlobSender, cids []cid.Cid) error {
+	defer bs.Close()
```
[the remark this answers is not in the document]
smnzhu: I don't see why this is better. It makes more sense to have it here because this function is the scope of the logic that actually uses bs.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+func (s *ExecutionDataService) receiveBlobs(parent context.Context, br *BlobReceiver) ([]cid.Cid, error) {
+	defer br.Close()
+
+	ctx, cancel := context.WithCancel(parent)
+	defer cancel()
+
+	var cids []cid.Cid
+
+	for {
+		batch, recvErr := s.receiveBatch(ctx, br)
```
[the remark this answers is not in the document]
smnzhu: No, because the error just means the blob channel was closed. We still want to process the last batch in this case

module/state_synchronization/execution_data_service.go
smnzhu: Yes, unless err != nil.

module/state_synchronization/execution_data_service.go
smnzhu: `len(cids)` is guaranteed to eventually be equal to 1. It will always get smaller and smaller.

module/state_synchronization/execution_data_service.go
smnzhu: `blobChan` is a channel of blobs that eventually gets closed when all the blobs are sent to the channel.

It is not O(N^2). When an element is iterated over in a channel, what's really happening is that the item is removed from the channel. See https://gobyexample.com/range-over-channels for an explanation of how this syntax works.

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+	for n < len(data) && bw.err == nil {
+		m, bw.err = bw.write(data[n:])
+		n += m
+	}
+
+	return n, bw.err
+}
+
+func (bw *BlobChannelWriter) write(p []byte) (int, error) {
+	n := bw.maxBlobSize - len(bw.buf)
```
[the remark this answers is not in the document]
smnzhu: No, it is not possible. I'll add comments.

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+
+var _ io.WriteCloser = (*BlobChannelWriter)(nil)
+var _ io.ByteWriter = (*BlobChannelWriter)(nil)
+
+func (bw *BlobChannelWriter) Write(data []byte) (int, error) {
+	var n, m int
+
+	for n < len(data) && bw.err == nil {
+		m, bw.err = bw.write(data[n:])
+		n += m
```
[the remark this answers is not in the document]
smnzhu: m is never negative because it represents the number of bytes written, which cannot be negative.

m can only be 0 if err != nil, in which case we would exit the loop.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+		} else {
+			blob, err = s.findBlob(blobChan, c, cachedBlobs)
+
+			if err != nil {
+				return err
+			}
+
+			if blob == nil {
+				// the blob channel may be closed as a result of the context being canceled,
+				// in which case we should return the context error.
```
zhangchiqing, line 175: Why would the context get canceled?

The context is only canceled when we [exit this function](https://github.com/onflow/flow-go/pull/1582/files#diff-06e9e7fa42b0b4d29eaa42ca5c75a865d70919934e07e97a75a02b63a9d10313R138), and here we are still within this function.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,308 @@
...
+		v, deserializeErr = s.serializer.Deserialize(bcr)
+	}()
+
+	sendErr := s.retrieveBlobs(ctx, bs, cids)
+
+	<-done
+
+	if sendErr != nil && errors.Is(sendErr, ErrClosedBlobChannel) {
+		return nil, sendErr
+	}
```
peterargue, line 234: what if `sendErr` is not a `ErrClosedBlobChannel` or nil?

network/blob_service.go
```diff
@@ -0,0 +1,180 @@
...
+// the blobservice.
+type BlobGetter interface {
+	// GetBlob gets the requested blob.
+	GetBlob(ctx context.Context, c cid.Cid) (Blob, error)
+
+	// GetBlobs does a batch request for the given cids, returning blobs as
+	// they are found, in no particular order.
+	//
+	// It may not be able to find all requested blobs (or the context may
+	// be canceled). In that case, it will close the channel early. It is up
```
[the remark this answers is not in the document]
zhangchiqing: ⚠️ In the above cases, except case 1, all the other cases is caused by certain error, but the API doesn't have a way to communicate about this error, it relies on returning partial data to give the caller clue that something wrong happened.  And there is no way to explicitly communicate what was the exact error. 

This is unlike the `GetBlob` API where it will return an error about why the blob for the given ID can't be retrieved. 

I understand returning a `chan Blob` is probably good for streaming process, but I found missing the error reporting is a bit concerning to me. 

I had another idea for you to consider, this might allow both streaming process and error reporting:
Instead of letting `GetBlobs` return a `<- chan Blob`, we could let it take a stream processing function `func (blob, error) error`, so that when my function gets called, you will either give me a blob or an error.

module/state_synchronization/execution_data_service.go
peterargue: I think the context here is that `GetBlobs` returns a channel that receives all requested blobs, but not necessarily in order so we have to read them all out and then check which ones we got.

Does `Send` require that blobs are sent in order? if it doesn't, we could flip the logic to just read from the channel and then send the blobs as they come and record which we've seen to track that it finished.

Otherwise, we'll need to block until all blobs are read. This approach will allow us to send faster than the naive blocking read approach in all but the worse case. Can you add some comments detailing the what and why

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/onflow/flow-go/network"
+)
+
+var ErrClosedBlobChannel = errors.New("send/receive on closed blob channel")
+
+type blobChannel struct {
+	blobs chan network.Blob
+	once  sync.Once
+	done  chan struct{}
```
peterargue, line 16: why use a done channel instead of a context?

network/p2p/network.go
```diff
@@ -227,17 +227,21 @@ func (n *Network) handleRegisterEngineRequest(parent irrecoverable.SignalerConte
...
 	}
 	if mw.libP2PNode.dht == nil {
-		return nil, errors.New("block exchange is disabled because content routing is not configured")
+		return nil, errors.New("blob exchange is disabled because content routing is not configured")
 	}
 
-	return NewBlockExchange(parent, mw.libP2PNode.host, mw.libP2PNode.dht, channel.String(), bstore), nil
+	bs := network.NewBlobService(mw.libP2PNode.host, mw.libP2PNode.dht, channel.String(), ds)
+	bs.Start(parent)
+	<-bs.Ready()
```
peterargue, line 242: should this check if the context was cancelled?

peterargue (commented):
this is coming together. overall it could use a more comments. it's complex so err on the side over overcommenting

module/state_synchronization/execution_data_service.go
peterargue: did you mean `!errors.Is(sendErr, ErrClosedBlobChannel)`?

network/blob_service.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+type blobService struct {
+	*component.ComponentManager
+	blockService blockservice.BlockService
+	reprovider   provider.Reprovider
+}
+
+var _ BlobService = (*blobService)(nil)
+var _ module.Startable = (*blobService)(nil)
```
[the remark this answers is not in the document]
smnzhu: a02f7e9

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,308 @@
...
+func (s *ExecutionDataService) retrieveBlobs(parent context.Context, bs *BlobSender, cids []cid.Cid) error {
+	defer bs.Close()
+
+	ctx, cancel := context.WithCancel(parent)
+	defer cancel()
+
+	blobChan := s.blobService.GetBlobs(ctx, cids)
+	cachedBlobs := make(map[cid.Cid]network.Blob)
+
+	for _, c := range cids {
```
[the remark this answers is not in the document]
smnzhu: but `cid` is the name of the package (see line 158 above for example), and I'd like to avoid shadowing

module/state_synchronization/execution_data_service.go
smnzhu: 26c2d03

smnzhu: yes, that's what I meant, good catch

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+	select {
+	case blob := <-bc.blobs:
+		return blob, nil
+	case <-bc.done:
+		return nil, ErrClosedBlobChannel
+	}
+}
+
+func (bc *blobChannel) Close() error {
+	bc.once.Do(func() { close(bc.done) })
```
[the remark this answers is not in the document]
smnzhu: I don't think it's necessary, and it could cause a panic in Send / Receive if something is already blocking on read.

For context, my implementation is based on Go's `Pipe`: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/io/pipe.go;l=39

network/p2p/network.go
smnzhu: it absolutely should, thanks for catching this

module/state_synchronization/blob_io.go
smnzhu: So again, for context my implementation is based on https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/io/pipe.go;l=39

I can only guess why they chose a done channel, but my overwhelming hunch is that it's because of https://go.dev/blog/context-and-structs

module/state_synchronization/execution_data_service.go
```diff
@@ -201,12 +231,20 @@ func (s *ExecutionDataService) retrieveBlobs(parent context.Context, bs *blobs.B
 	return nil
 }
 
+// findBlob retrieves blobs from the given channel, caching them along the way, until it either
+// finds the target blob or exhausts the channel.
```
peterargue, line 235:
```suggestion
// findBlob retrieves blobs from the given channel, caching them along the way, until it either
// finds the target blob or exhausts the channel.
//
// blobChan is not guaranteed to receive the list of CIDs in order. This caching approach allows us
// to improve throughput and reduce memory used in all but the worst case where blocks are received
// in exactly reverse order.
```

module/state_synchronization/execution_data_service.go
```diff
@@ -221,24 +259,30 @@ func (s *ExecutionDataService) findBlob(
 		}
 	}
 
+	targetLogger.Debug().Msg("blob not found")
+
 	return nil, &BlobNotFoundError{target}
 }
 
-func (s *ExecutionDataService) getBlobs(ctx context.Context, cids []cid.Cid) (interface{}, error) {
+// getBlobs gets the given CIDs from the blobservice, reassembles the blobs, and deserializes the reassembled data into an object.
```
peterargue, line 267:
```suggestion
// getBlobs gets the given CIDs from the blobservice, reassembles the blobs, and deserializes the reassembled data into an object.
//
// Blobs are fetched in a streaming fashion to allow for very large sets. This is accomplished using
// a separate goroutine that reads the blobs associated with the provided CIDs from the store, and 
// passes them back to the main routine over the blob channel.
```

module/state_synchronization/execution_data_service.go
```diff
@@ -260,6 +304,9 @@ func (s *ExecutionDataService) getBlobs(ctx context.Context, cids []cid.Cid) (in
 
 // Get gets the ExecutionData for the given root CID from the blobservice.
 func (s *ExecutionDataService) Get(ctx context.Context, c cid.Cid) (*ExecutionData, error) {
```
peterargue, line 306: maybe call it `rootCID` instead of `c`

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,370 @@
...
+
+			// the blob channel was closed, meaning that all blobs have been received
+			break
+		}
+	}
+
+	return cids, nil
+}
+
+// addBlobs serializes the given object, splits the serialized data into blobs, and adds them to the blobservice
```
peterargue, line 113:
```suggestion
// addBlobs serializes the given object, splits the serialized data into blobs, and adds them to the blobservice
//
// Blobs are added in a streaming fashion to allow for very large sets. This is accomplished using
// a separate goroutine that writes the data associated with the provided object to the store, which 
// passes the blobs back to the main routine over the blob channel.
```

module/state_synchronization/execution_data_service.go
smnzhu: The parent context which is passed in could be canceled.

network/blob_service.go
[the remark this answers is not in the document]
smnzhu: > I had another idea for you to consider, this might allow both streaming process and error reporting:
Instead of letting GetBlobs return a <- chan Blob, we could let it take a stream processing function func (blob, error) error, so that when my function gets called, you will either give me a blob or an error.

This is actually similar to what I originally started with, but after working with that for a while, I decided that returning a channel has much more flexibility and was actually easier to work with.

case 3 shouldn't really be an issue, the code properly handles this case and anyways it should never occur.

case 2 could in fact *only* happen when the context is canceled (or times out). This is not apparent from the API, but is true based on the underlying implementation.

case 4 could never happen.

case 5 is just a special case of case 2.

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+	n := len(br.buf)
+
+	if n == 0 {
+		err := br.receiveNewBlob()
+
+		if errors.Is(err, ErrClosedBlobChannel) {
+			return 0, io.EOF
+		}
+
+		return 0, err
```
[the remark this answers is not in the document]
smnzhu: Yeah, I considered this but in the end I thought it made the code cleaner if we just return 0, nil and let the parent function call `read` again.

module/state_synchronization/blob_io.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return bw.err
+}
+
+func (bw *BlobChannelWriter) Close() error {
+	if err := bw.Flush(); err != nil {
+		return err
+	}
+
+	return bw.blobs.Close()
```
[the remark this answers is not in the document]
smnzhu: Yep, fixed.

module/state_synchronization/execution_data_service.go
smnzhu: > Does Send require that blobs are sent in order? if it doesn't, we could flip the logic to just read from the channel and then send the blobs as they come and record which we've seen to track that it finished.

The blobs must be sent in the order that they appear in the CID list.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+		if serializeErr = s.serializer.Serialize(bcw, v); serializeErr != nil {
+			return
+		}
+
+		serializeErr = bcw.Flush()
+	}()
+
+	cids, recvErr := s.storeBlobs(ctx, br)
+
+	<-done
```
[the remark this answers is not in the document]
smnzhu: `storeBlobs` always closes `br` before returning, and closing the underlying blob channel will cause any blocking `bcw` write to return an `ErrBlobChannelClosed`

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,280 @@
...
+		blobs = append(blobs, blob)
+	}
+
+	return blobs, err
+}
+
+func (s *ExecutionDataService) storeBlobs(parent context.Context, br *BlobReceiver) ([]cid.Cid, error) {
+	defer br.Close()
+
+	ctx, cancel := context.WithCancel(parent)
```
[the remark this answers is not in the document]
smnzhu: So that when we return from this function, we cancel any ongoing `AddBlobs` requests that returned with an error.

It's possible that it isn't actually necessary, because `AddBlobs` may properly handle canceling all goroutines before returning an error, but just wanted to be safe.

network/p2p/network.go
smnzhu: Updated so that we skip calling `Ready` altogether. Leave it up to caller to wait for `Ready`.

module/blobs/blob_io.go
```diff
@@ -0,0 +1,305 @@
+package blobs
+
+import (
+	"errors"
+	"io"
+	"sync"
+)
+
```
peterargue, line 8: can you add a high level description of how these structs should be used and the rational?

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/module/blobs"
+	"github.com/onflow/flow-go/network"
+)
+
+const (
+	defaultMaxBlobSize      = 1 << 20 // 1MiB
+	defaultMaxBlobTreeDepth = 2       // prevents malicious CID from causing download of unbounded amounts of data
```
peterargue, line 20: maybe include the rough size estimation for 2 levels for context. 31GB?

module/state_synchronization/execution_data_service.go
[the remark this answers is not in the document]
peterargue: ok, can you describe that in a comment here then

Ullaakut (commented):
Hi!

I hope I'm not intruding by reviewing this pull request, but since it will most likely have a large impact on the Flow DPS in the future, I want to give my honest feedback on the pull request from the perspective of a consumer.

I didn't go into details/nitpicks, since it is not the point of my review, but I do have some concerns about the design of the components within this PR, and specifically around the usage of channels. I hope I provided enough information to support my suggestions. I believe that even tough it does not directly impact us at the moment, decisions that could heavily impact the maintainability of dependencies of the DPS might end up becoming an issue from our perspective in the future.

From our perspective, the most important part is the `ExecutionDataService` and more specifically its ability to get execution results, and this seems good to me as is, though I would really like for it to have more comments and explanations overall, ideally.

I saw in the PR's body that it currently implements _part of_ the issue to add an execution data requester engine, so I'm curious. What parts are currently missing?

module/state_synchronization/serializer.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/network"
+)
+
+// header codes to distinguish between different types of data
+const (
+	CodeRecursiveCIDs = iota
+	CodeExecutionData
+)
```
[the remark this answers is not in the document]
awfm9: A best practice is to consider the return value unspecified when an error is returned, but that doesn't make your point invalid.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+	for _, c := range cids {
+		blob, ok := cachedBlobs[c]
+
+		if !ok {
+			var err error
+
+			blob, err = s.findBlob(blobChan, c, cachedBlobs, logger)
+
+			if err != nil {
+				_, ok := err.(*BlobNotFoundError)
```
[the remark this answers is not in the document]
awfm9: In this case, we are asserting an error type, so it should always use `errors.As`.

module/state_synchronization/serializer.go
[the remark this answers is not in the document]
Ullaakut: Of course, but you shouldn't expect the consumers of your functions to follow best practices 😄

network/blob_service.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// NewBlobService creates a new BlobService.
+func NewBlobService(
+	host host.Host,
+	r routing.ContentRouting,
+	prefix string,
+	ds datastore.Batching,
+	opts ...BlobServiceOption,
+) BlobService {
```
Ullaakut, line 94: From the POV of the Flow DPS, I am a bit concerned about those parameters that need to be provided. Since the Blob Service is a required component of the Execution Data Service, we need to instantiate it, and I'm not sure how we'll know  how to configure it, from what I see here.

Where do the following come from?

* Routing information
* Prefix
* Datastore Batching

How should we expect to configure this in the Flow DPS? Should it be configuration flags that are set by the user at launch, or parsed from a configuration file? Fetched from other nodes in the network?

module/blobs/blob_io.go
```diff
@@ -0,0 +1,305 @@
...
+	case <-bc.done:
+		return nil, ErrClosedBlobChannel
+	}
+}
+
+// Close closes the blob channel. The returned error is always nil.
+func (bc *blobChannel) Close() error {
+	bc.once.Do(func() { close(bc.done) })
+	return nil
+}
```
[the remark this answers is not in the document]
smnzhu: See the go source code here: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/io/pipe.go;l=75-82

I am definitely very open to discussing alternative code, but just want to point out that this is a pattern that exists in the Go source code, so it can't be that much of an antipattern.

smnzhu:
> I saw in the PR's body that it currently implements part of the issue to add an execution data requester engine, so I'm curious. What parts are currently missing? 

The part missing is the `Engine` which actually follows the chain / requests the Execution Data. This part is probably not relevant to you guys.

module/state_synchronization/execution_data_service.go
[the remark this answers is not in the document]
smnzhu: @Ullaakut So while I agree with most of what you're saying in most cases, the important thing to understand here is that we are not actually working with channels here. We are working with an abstraction which happens to be implemented using channels. As I've mentioned before, the implementation is based on the Go implementation of Pipe:
https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/io/pipe.go

Instead of thinking about a BlobChannel as an actual channel, it would be more appropriate to think of it as a Pipe for blobs / blob data.

network/blob_service.go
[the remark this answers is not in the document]
smnzhu: @Ullaakut For example, you can cancel iteration over the channel as soon as you want. If we build another layer of abstraction on top of this, then we can only cancel the context, but we cannot signal the underlying implementation to stop consuming the rest of the blobs that are already in the channel.

There are other reasons, for example it is not as easy to keep track of the order in which you receive items, because from the perspective of the callback you don't know how many blobs have already come before you.

Initially the callback seems like a good idea, and that's why I initially implemented it that way. However, when you work with it more, you may realize that it sacrifice a lot of flexibility.

module/state_synchronization/execution_data_service.go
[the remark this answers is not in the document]
smnzhu: >  Clarity because errors.As is more obvious than a type assertion, and maintainability because errors.As supports wrapped errors (which regardless of the current implementation, could very well end up being the case one day).

If this is your opinion, I'm not against changing this. I intentionally wrote it this way because I actually consider this version more clear, but I'm fine with either version.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+	logger.Debug().Msg("getting execution data")
+
+	s.metrics.ExecutionDataGetStarted()
+
+	start := time.Now()
+	cids := []cid.Cid{rootCid}
+
+	var blobTreeNodes int
+
+	for i := uint(0); i < defaultMaxBlobTreeDepth; i++ {
```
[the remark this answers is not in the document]
smnzhu: Yeah, maybe the semantics of this variable is a bit off. What the variable is actually supposed to mean is "maximum number of levels in the tree", which is one greater than the height. Will update this.

network/blob_service.go
smnzhu: We will provide examples for all of those later.

ContentRouting in our specific case would be a DHT instance: https://pkg.go.dev/github.com/libp2p/go-libp2p-kad-dht#section-readme

Currently, the the libp2p host and DHT instance are both created internally by the ConsensusFollower, due to the way the code was already structured. However, we have an open issue to make both of these injectable: https://github.com/onflow/flow-go/issues/1669

Think of the Prefix as the protocol name. This is just some agreed upon string that all parties participating in the state sync protocol will use. The reason we support multiple prefixes is because the BlobService is not restricted to the single ExecutionDataService usecase, but it may be used as the underlying data transport protocol for various different high level protocols.

The datastore can be whatever implementation you like. We will be using a modified version of https://pkg.go.dev/github.com/ipfs/go-ds-badger.

module/blobs/blob_io.go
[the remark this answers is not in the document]
Ullaakut: I didn't realize that you were implementing a pipe rather than just wrapping channels, my bad. I'm not sure I understand, in this case, the advantage/reason for using a pipe system over a channel? I guess there is some complexity I'm not seeing here that justifies it, so I might be missing something 😄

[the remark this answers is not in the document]
smnzhu: @Ullaakut  Yeah, to be frank the main reason for introducing this is because we need some mechanism for the receiver to close the pipe, signalling the sender to exit. And as you said, with channels the convention is that the sender always closes the channel.

In our case, the receiver is either doing some sort of validation (ie deserializing the data) or writing the received data to some external service (blob service), during which it may encounter errors. If it does, I wanted a simple way for it to signal the sender to stop sending data. Also, the receiver may be the one who has access to the context, which may be canceled.

If you have a suggestion on how to better handle this usecase I am very open to it btw. I thought about other ways but in the end this one seemed the cleanest to me.

module/blobs/blob_io.go
```diff
@@ -0,0 +1,313 @@
...
+	"sync"
+)
+
+// This file contains the implementation of a Blob Channel,
+// which represents an in-memory pipe for blob data.
+//
+// It is safe for the sender and receiver objects returned
+// from IncomingBlobChannel or OutgoingBlobChannel to be used
+// in parallel with each other. However, it is not safe to use
+// the same sender / receiver in parallel.
```
peterargue, line 15: adding some more depth
```suggestion
// This file contains the implementation of a Blob Channel, which represents an in-memory pipe for
// blob data.
//
// It is safe for the sender and receiver objects returned from IncomingBlobChannel or
// OutgoingBlobChannel to be used in parallel with each other. However, it is not safe to use the
// same sender / receiver in parallel.
//
// Typically, interactions with these objects are done using IncomingBlobChannel and OutgoingBlobChannel.
//
// IncomingBlobChannel is used when adding new blobs, and returns a BlobChannelWriter and BlobReceiver.
// Data is serialized and written to the BlobChannelWriter, which breaks the data into blobs and passes
// them to the BlobReceiver. The BlobReceiver is used by the caller to get a stream of all blobs written. 
//
// OutgoingBlobChannel is used when reading existing blobs, and returns a BlobChannelReader and BlobSender.
// Blobs are read and sent through BlobSender to the BlobChannelReader. The BlobChannelReader aggregates 
// the blob data so it can be deserialized back into its original format.
```

module/state_synchronization/serializer.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/ipfs/go-cid"
+
+	"github.com/onflow/flow-go/model/encoding"
+	"github.com/onflow/flow-go/network"
+)
+
```
peterargue, line 12:
```suggestion
// serializer is used for serializing and deserializing ExecutionData and CID lists for the state
// synchronization service. The data is passed through the configured codec and compressor before
// returning the final format.
//
// The serialized data is prefixed with a single byte header that identifies the underlying data
// format. This allows adding new data types in a backwards compatible way.
```

peterargue (approved):
looks great. I added suggestions for a couple high level comments describing major pieces of the new functionality.

Can you add something similar for `ExecutionDataService`?

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+		return nil, storeErr
+	}
+
+	return cids, serializeErr
+}
+
+// Add constructs a blob tree for the given ExecutionData and adds it to the blobservice, and then returns the root CID.
+func (s *ExecutionDataService) Add(ctx context.Context, sd *ExecutionData) (cid.Cid, error) {
+	logger := s.logger.With().Str("block_id", sd.BlockID.String()).Logger()
+	logger.Debug().Msg("adding execution data")
```
zhangchiqing, line 155: better to add an info level log with both block ID, cids as well as the root CID, so that their relations are logged together.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+
+	s.metrics.ExecutionDataAddStarted()
+
+	start := time.Now()
+	cids, err := s.addBlobs(ctx, sd, logger)
+
+	if err != nil {
+		s.metrics.ExecutionDataAddFinished(time.Since(start), false, 0)
+
+		return cid.Undef, fmt.Errorf("failed to add execution data blobs: %w", err)
```
zhangchiqing, line 165: better to include block ID in the error

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+		if len(cids) == 1 {
+			s.metrics.ExecutionDataAddFinished(time.Since(start), true, blobTreeNodes)
+
+			return cids[0], nil
+		}
+
+		if cids, err = s.addBlobs(ctx, cids, logger); err != nil {
+			s.metrics.ExecutionDataAddFinished(time.Since(start), false, blobTreeNodes)
+
+			return cid.Undef, fmt.Errorf("failed to add cid blobs: %w", err)
```
zhangchiqing, line 182: better to include block id in the error

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+		if err := bs.Send(blob); err != nil {
+			return fmt.Errorf("failed to send blob %v to blob channel: %w", blob.Cid(), err)
+		}
+	}
+
+	return nil
+}
+
+// findBlob retrieves blobs from the given channel, caching them along the way, until it either
+// finds the target blob or exhausts the channel.
```
zhangchiqing, line 238: better to add comment with all the possible error types

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,378 @@
...
+			blob, err = s.findBlob(blobChan, c, cachedBlobs, logger)
+
+			if err != nil {
+				var blobNotFound *BlobNotFoundError
+
+				// the blob channel may be closed as a result of the context being canceled,
+				// in which case we should return the context error.
+				if errors.As(err, &blobNotFound) && ctx.Err() != nil {
+					return ctx.Err()
+				}
```
zhangchiqing, line 217: can we simplify this here?

I think if the channel is closed, [we should get a `ErrClosedBlobChannel` error](https://github.com/onflow/flow-go/pull/1582/files#diff-06e9e7fa42b0b4d29eaa42ca5c75a865d70919934e07e97a75a02b63a9d10313R213-R217), right?

In that case, I don't think we need to handle `ctx.Err()`, because they are essentially the case error.

smnzhu: Yes, but in this case, the root cause of the blob channel being closed is because the context is closed. I think it makes sense to return the root cause error in this case, otherwise the client may be confused about why they receive an `ErrClosedBlobChannel` error.

module/state_synchronization/execution_data_service.go
smnzhu: this will be included one level up, by the caller.

module/state_synchronization/execution_data_service.go
smnzhu: this will be done by the caller.

module/state_synchronization/execution_data_service.go
```diff
@@ -0,0 +1,406 @@
...
+		cidCounts[c] += 1
+	}
+
+	for _, c := range cids {
+		blob, ok := cachedBlobs[c]
+
+		if !ok {
+			var err error
+
+			blob, err = s.findBlob(blobChan, c, cachedBlobs, logger)
```
zhangchiqing, line 232: the call to `findBlob` will be blocking, is there any guarantee that it will return within certain duration?

zhangchiqing (approved):
Looks good to me. Nice work!

smnzhu:
bors merge

PR #1587, opened 2021-11-05
Update live debugger to include block header data

ramtinms:
This PR updates the live debugger to pass in the block info while executing the transaction. 
this enables debugging some transactions that are dependent on info like block height to do their operations


janezpodhostnik (approved):
Nice improvement! LGTM.

zhangchiqing (approved):

PR #1588, opened 2021-11-05
[Execution] using valid descendents when loading unexecuted blocks

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5989

The fixes an issue that the execution nodes are not loading only valid decendents on startup


ramtinms (approved):

durkmurder (changes requested):
I believe this is not enough, unit tests are broken so we need to fix them. 
Mocks are expecting `Descendants` call not `ValidDescendants`

durkmurder (approved):
Nice, thanks for addressing this.

zhangchiqing:
bors merge

PR #1595, opened 2021-11-06
Kan/hotfix account creation interaction limit panic

Kay-Zee:


janezpodhostnik (approved):
It's interesting that account creation caused this problem. For now this solution seems good, but we need to see if we can find a solution where we don't have to suspend interaction limiting while creating an account.

PR #1600, opened 2021-11-09
[Networking] fixes flakey stream closing test

yhassanzadeh13:
The changes this PR introduces are: 
 1. Fixes flakey [TestStreamClosing](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L467).
 2. Moves all stream-related tests from `libp2pNode_test.go` -> `libp2pStream_test.go`.
 
**Root cause of flakiness:** The root cause of flakey behavior in `TestStreamClosing` was a race condition imposed by the way the test invoke libp2p connection lifecycle managers. In very specific, there was a race condition between [closing stream at reader side upon an io.EOF](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L484), and [stopping the nodes](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L508) through `defer` prior to the test method return. 

**How this would cause a race condition?**: `TestStreamClosing` would not synchronize with its [internal goroutines](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L478). So right prior the time the test method would return (and `defer StopNodes(suite.T(), nodes)` would be executing), there were cases that the internal goroutines were still in execution. The `defer StopNodes(suite.T(), nodes)` would close the `Session` which would [force close](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L484) [all streams](https://github.com/libp2p/go-yamux/blob/master/session.go#L267-L269) of the node which yields those streams in an `halfReset` status. In this force closing would _happen before_ the goroutines of the test [manually closing streams](https://github.com/onflow/flow-go/blob/master/network/p2p/libp2pNode_test.go#L484), it would result in the same libp2p node [trying to close an already closed stream](https://github.com/libp2p/go-yamux/blob/master/stream.go#L295-L297), which would cause the test to fail with a [stream reset error](https://github.com/libp2p/go-yamux/blob/master/stream.go#L297). 

**How does this PR fixes this race condition?**: this PR adds a test helper method `mockStreamHandlerForMessage` which creates a `WaitGroup` for the stream closing event, and makes sure that closing all created streams _happens before_ stopping the libp2p nodes. In this way, no libp2p node tries closing a `halfReset` stream. 

**How to make sure that this test is no longer flakey??**: run the following command, which executes the test iteratively until either 1000 successful run, or the first failure, and you see the test passes (however the same command fails the test on `master`).
```
go test -run TestStreamClosing github.com/onflow/flow-go/network/p2p/ -count 1000 -failfast
```

**Why move all stream tests to a new file?**: to make them more maintainable, and also get rid of the test suite. As we are growing the tests, they have much less commonalities to accommodate them into the same test suite. Independent test functions isolates the shared objects between such tests and makes it easier to maintain, track, and fix.


network/p2p/libp2pStream_test.go
```diff
@@ -0,0 +1,457 @@
...
+			err := s.Close()
+			assert.NoError(t, err)
+			wg.Done()
+		}()
+		wg.Wait()
+		// assert that the stream count within libp2p decremented
+		require.Equal(t, i, CountStream(nodes[0].host, nodes[1].host.ID(), flowProtocolID, network.DirOutbound))
+	}
+}
+
```
yhassanzadeh13, line 154: These are the core changes of this PR, the rest are just migrating tests to a new file and moving them out of test suite.

network/p2p/libp2pStream_test.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	// wait for stream to be closed
+	unittest.RequireReturnsBefore(t, senderWG.Wait, 1*time.Second, "could not send messages on time")
+	unittest.RequireReturnsBefore(t, streamCloseWG.Wait, 1*time.Second, "could not close stream at receiver side")
+}
+
+// mockStreamHandlerForMessages creates a stream handler that expects receiving `msgCount` unique messages that match the input regexp.
+// The returned wait group will be unlocked when all messages are completely received and associated streams are closed.
+func mockStreamHandlerForMessages(t *testing.T, ctx context.Context, msgCount int, msgRegexp *regexp.Regexp) (network.StreamHandler, *sync.WaitGroup) {
```
Kay-Zee, line 73: :+1: for extracting this piece of logic out into a helper, even if we didn't need to add the wait group

Kay-Zee (approved):

vishalchangrani (approved):

network/p2p/libp2pStream_test.go
```diff
@@ -0,0 +1,457 @@
...
+
+	id2 := identities[1]
+
+	flowProtocolID := FlowProtocolID(rootBlockID)
+	// Assert that there is no outbound stream to the target yet
+	require.Equal(t, 0, CountStream(nodes[0].host, nodes[1].host.ID(), flowProtocolID, network.DirOutbound))
+
+	// Now attempt to create another 100 outbound stream to the same destination by calling CreateStream
+	var streams []network.Stream
+	for i := 0; i < 100; i++ {
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1600/commits/6d1ea3d9208c41c042ed41fe3059ae0201d68d07

smnzhu (approved):

PR #1602, opened 2021-11-09
[Consensus Voting] `RandomBeaconReconstructor` implementation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6002

### Context

This PR implements thin layer known as `RandomBeaconReconstructor` which takes care of forwarding calls to `RandomBeaconInspector` and transforming `sinderID` -> `signerIndex`. `RandomBeaconReconstructor` is a missing piece to make the building QC pipeline work. Fixed tests for both V2 and V3, added missing pieces to make V3 test work.

Notable changes:
- Implementation of `RandomBeaconReconstructor`
- Minimal test suite
- Updated `CombinedVoteProcessorV2` integration tests
- Updated `CombinedVoteProcessorFactoryV2` 
- Implemented `CombinedVerifierV3` and `CombinedSignerV3`
- Updated tests and implementation for `CombinedVoteProcessorV3`


consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -51,8 +52,31 @@ func (f *combinedVoteProcessorFactoryBaseV2) Create(block *model.Block) (hotstuf
...
 
-	rbRector := &signature.RandomBeaconReconstructor{} // TODO: initialize properly when ready
+	publicKeyShares := make([]crypto.PublicKey, 0, len(allParticipants))
+	dkg, err := f.committee.DKG(block.BlockID)
+	if err != nil {
+		return nil, fmt.Errorf("could not get DKG info at block %v: %w", block.BlockID, err)
+	}
+	for _, participant := range allParticipants {
+		pk, err := dkg.KeyShare(participant.NodeID)
+		if err != nil {
```
durkmurder, line 62: @tarakby Not sure how to behave in this situation, need your input here

tarakby: At this stage, all participants have a public key share that should be returned by `dkg.KeyShare`, regardless of the local DKG results of each participant.

consensus/hotstuff/signature/randombeacon_reconstructor.go
```diff
@@ -1,52 +1,85 @@
...
+// It has knowledge about DKG to be able to map signerID to signerIndex
 type RandomBeaconReconstructor struct {
 	dkg                   hotstuff.DKG                   // to lookup signer index by signer ID
 	randomBeaconInspector hotstuff.RandomBeaconInspector // a stateful object for this block. It's used for both storing all sig shares and producing the node's own share by signing the block
 }
 
 var _ hotstuff.RandomBeaconReconstructor = &RandomBeaconReconstructor{}
 
-func NewRandomBeaconReconstructur(dkg hotstuff.DKG, randomBeaconInspector hotstuff.RandomBeaconInspector) *RandomBeaconReconstructor {
+func NewRandomBeaconReconstructor(dkg hotstuff.DKG, randomBeaconInspector hotstuff.RandomBeaconInspector) *RandomBeaconReconstructor {
```
tarakby, line 21: I think `NewRandomBeaconReconstructor` may not need an `randomBeaconInspector` input. Using the `hotstuff.DKG` it can retrieve the data needed to construct `randomBeaconInspector` inside the function.

consensus/hotstuff/signature/randombeacon_reconstructor.go
```diff
@@ -1,52 +1,85 @@
...
+	return r.randomBeaconInspector.TrustedAdd(int(signerIndex), sig)
 }
 
-// EnoughShares returns true if and only if reconstructor
-// has collected a sufficient number of signature shares.
+// EnoughShares indicates whether enough shares have been accumulated in order to reconstruct
+// a group signature. The function is thread-safe.
 func (r *RandomBeaconReconstructor) EnoughShares() bool {
-	panic("to be implemented")
+	return r.randomBeaconInspector.EnoughShares()
```
tarakby, line 70: if `hotstuff.RandomBeaconInspector` was declared to be embedded inside `RandomBeaconReconstructor`, we won't need to write this function, it will be just inherited. 

```
type RandomBeaconReconstructor struct {
	dkg                   hotstuff.DKG                   
	hotstuff.RandomBeaconInspector
}
```

consensus/hotstuff/signature/randombeacon_reconstructor.go
```diff
@@ -1,52 +1,85 @@
...
+// Returns:
+// - (signature, nil) if no error occurred
+// - (nil, crypto.notEnoughSharesError) if not enough shares were collected
+// - (nil, crypto.invalidInputsError) if at least one collected share does not serialize to a valid BLS signature,
+//    or if the constructed signature failed to verify against the group public key and stored message. This post-verification
+//    is required  for safety, as `TrustedAdd` allows adding invalid signatures.
+// - (nil, error) for any other unexpected error.
 func (r *RandomBeaconReconstructor) Reconstruct() (crypto.Signature, error) {
-	panic("to be implemented")
+	return r.randomBeaconInspector.Reconstruct()
```
tarakby, line 84: Same comment here for embedding the inspector, we could remove this function.

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -51,8 +52,26 @@ func (f *combinedVoteProcessorFactoryBaseV2) Create(block *model.Block) (hotstuf
...
+			return nil, fmt.Errorf("could not get random beacon key share for %x: %w", participant.NodeID, err)
+		}
+		publicKeyShares = append(publicKeyShares, pk)
+	}
+
+	threshold := msig.RandomBeaconThreshold(int(dkg.Size()))
+	randomBeaconInspector, err := signature.NewRandomBeaconInspector(dkg.GroupKey(), publicKeyShares, threshold, msg)
+	if err != nil {
+		return nil, fmt.Errorf("could not create random beacon inspector: %w", err)
+	}
```
tarakby, line 72: This is the code that I think it could go inside `NewRandomBeaconReconstructor`.

tarakby (approved):
It's nice to place the last lego piece 🏄🏼 
I just left some minor refactoring ideas.

consensus/hotstuff/signature/randombeacon_reconstructor.go
durkmurder: Need a bit more info, but yeah, makes sense to include that.

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -0,0 +1,175 @@
...
+			return false, fmt.Errorf("could not get random beacon key share for %x: %w", signer.NodeID, err)
+		}
+
+		beaconValid, err := beaconPubKey.Verify(sig, msg, c.beaconHasher)
+		if err != nil {
+			return false, fmt.Errorf("internal error while verifying beacon signature: %w", err)
+		}
+
+		if !beaconValid {
+			return false, fmt.Errorf("invalid beacon sig")
```
zhangchiqing, line 87:
```suggestion
			return false, fmt.Errorf("invalid beacon sig for block %v: %w", block.BlockID, model.ErrInvalidSignature)
```

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -0,0 +1,175 @@
...
+
+		// if there is beacon share, there must be beacon public key
+		beaconPubKey, err := dkg.KeyShare(signer.NodeID)
+		if err != nil {
+			return false, fmt.Errorf("could not get random beacon key share for %x: %w", signer.NodeID, err)
+		}
+
+		beaconValid, err := beaconPubKey.Verify(sig, msg, c.beaconHasher)
+		if err != nil {
+			return false, fmt.Errorf("internal error while verifying beacon signature: %w", err)
```
zhangchiqing, line 83:
```suggestion
			return false, fmt.Errorf("internal error while verifying beacon signature for block %v: %w", block.BlockID, err)
```

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -0,0 +1,175 @@
...
+	case hotstuff.SigTypeRandomBeacon:
+		dkg, err := c.committee.DKG(block.BlockID)
+		if err != nil {
+			return false, fmt.Errorf("could not get dkg: %w", err)
+		}
+
+		// if there is beacon share, there must be beacon public key
+		beaconPubKey, err := dkg.KeyShare(signer.NodeID)
+		if err != nil {
+			return false, fmt.Errorf("could not get random beacon key share for %x: %w", signer.NodeID, err)
```
zhangchiqing, line 78:
```suggestion
			return false, fmt.Errorf("could not get random beacon key share for %x at block %v: %w", signer.NodeID, block.BlockID, err)
```

zhangchiqing (approved):
Looks good

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -0,0 +1,175 @@
...
+		return false, fmt.Errorf("internal error while verifying beacon signature: %w", err)
+	}
+	if !beaconValid {
+		return false, nil
+	}
+
+	// verify the aggregated staking and beacon signatures next (more costly)
+
+	verifyAggregatedSignature := func(pubKeys []crypto.PublicKey, aggregatedSig crypto.Signature, hasher hash.Hasher) (bool, error) {
+		keysAggregator, err := modulesig.NewPublicKeyAggregator(pubKeys)
```
tarakby, line 124: What is implemented here works, however it doesn't need the `PublicKeyAggregator` stateful API. What you implemented could be done more easily using [this stateless API](https://github.com/onflow/flow-go/blob/master/crypto/bls_multisig.go#L147) for instance. 

We need the stateful structure `PublicKeyAggregator` in order to optimize the key aggregation: the structure keeps internally the latest aggregated key, and adjusts it based on the delta of the latest signers and new signers and gets the new aggregated key. 
For this to work, we need only a single `PublicKeyAggregator` per epoch per set of keys (so one struct for staking keys and another for beacon keys). 

I thought we can add this in another PR since it requires more refactoring (that's why my PR comment only mentioned adding a TODO comment), but feel free to add it here if you think the changes are manageable.

tarakby: If we're planning to make the changes outside this PR, I suggest to not use `PublicKeyAggregator` and use [the simple aggregation](https://github.com/onflow/flow-go/blob/master/crypto/bls_multisig.go#L147)

durkmurder: That is exactly what I was trying to bring on the meeting. Since we don't have a clear strategy how to deal per epoch caching I have just used it this way. I didn't know about stateless API. 

I would say let's go with stateless API and once we have a clear understanding how to handle changing epochs and where to do the caching then do the optimization.

module/signature/aggregation.go
```diff
@@ -1,3 +1,4 @@
+//go:build relic
```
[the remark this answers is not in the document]
durkmurder: I believe GoLand adds it automatically

consensus/hotstuff/signature/randombeacon_reconstructor.go
```diff
@@ -1,52 +1,65 @@
...
 )
 
 // RandomBeaconReconstructor implements hotstuff.RandomBeaconReconstructor.
-// The implementation wraps the thresholdSigner and translates the signer identity into signer index.
+// The implementation wraps the hotstuff.RandomBeaconInspector and translates the signer identity into signer index.
+// It has knowledge about DKG to be able to map signerID to signerIndex
 type RandomBeaconReconstructor struct {
-	dkg                   hotstuff.DKG                   // to lookup signer index by signer ID
-	randomBeaconInspector hotstuff.RandomBeaconInspector // a stateful object for this block. It's used for both storing all sig shares and producing the node's own share by signing the block
+	hotstuff.RandomBeaconInspector              // a stateful object for this block. It's used for both storing all sig shares and producing the node's own share by signing the block
```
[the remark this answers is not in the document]
durkmurder: Yes, theoretically we can do that but we don't have a strategy how to handle changes of epocs at this moment, that is why it's safer just to create a new instance every block. I know this is bad in terms of performance but let's deal with it once we have a working solution.

consensus/hotstuff/verification/combined_verifier_v3.go
tarakby: That works for me, let's just use the stateless API and leave a TODO comment. 
I'll create a ticket to integrate `PublicKeyAggregator` properly.

tarakby: https://github.com/dapperlabs/flow-go/issues/6016

module/signature/aggregation.go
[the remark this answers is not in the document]
tarakby: Maybe we can remove the redundant line below it then and leave the goland one? (they should be both equivalent)

consensus/hotstuff/verification/combined_verifier_v3.go
durkmurder: The stateless API is already in place

tarakby (approved):

durkmurder:
bors merge

durkmurder:
bors retry

PR #1603, opened 2021-11-09
Khalil/1603 Epochs test join with dynamic container adding/removal

kc1116:
The purpose of this PR is to add the utils needed to test Epoch join functionality (node joining after participating in staking auction) . It also adds a single test utilizing these utils TestEpochJoin that will stake and add a new AN to the network. This test will be expanded to cover all node roles in future PR's . 
 
This PR provides the utilities to do the following. 
- Stake a node in the Epoch integration test suite
        - generate keys: staking account, networking, staking and machine
        - create an account
        - fund account
        - create an staking collection
        - register node using staking collection
        - set the approved nodes list
- Create a ContainerConfig and start that container later on in the test 
        - update testing dock version that exports container Start func
        - utils to write key files and rootsnapshot
- Add test that stakes a new AN , starts the container and makes request to that AN


cmd/bootstrap/utils/key_generation.go
```diff
@@ -228,6 +229,35 @@ func WriteMachineAccountFiles(chainID flow.ChainID, nodeInfos []bootstrap.NodeIn
...
+	chainID flow.ChainID,
+	nodeID flow.Identifier,
+	accountAddress flow.Address,
+	accountKey encodable.MachineAccountPrivKey,
+	write WriteJSONFileFunc) error {
+	// ensure the chain ID is for a transient chain, where it is possible to
+	// infer machine account addresses this way
+	if !chainID.Transient() {
+		return fmt.Errorf("cannot write default machine account files for non-transient network")
+	}
```
jordanschalm, line 242: This isn't really necessary. We aren't inferring the machine account addresses here, since it's passed in as an argument.
```suggestion
```

integration/tests/epochs/epoch_test.go
```diff
@@ -136,31 +138,52 @@ func (s *Suite) TestEpochJoin() {
...
+
+		if currentPhase == flow.EpochPhaseSetup && err == nil {
+			break
 		}
 	}
 
-	require.True(s.T(), found, "node id for new node not found in approved list after setting the approved list")
-	s.net.StopContainers()
+	// write updated root snapshot
+	s.net.WriteRootSnapshot(snapshot)
```
jordanschalm, line 178: This is writing the root snapshot for the whole network, which I don't think we want here. We're emulating one node joining the existing network, so only that one node would have the later snapshot.

integration/tests/epochs/epoch_test.go
```diff
@@ -136,31 +138,52 @@ func (s *Suite) TestEpochJoin() {
...
-	found := false
-	for _, val := range approvedNodes.(cadence.Array).Values {
-		if string(val.(cadence.String)) == info.NodeID.String() {
-			found = true
+		currentPhase, err := snapshot.Phase()
+		require.NoError(s.T(), err)
+		_, err = snapshot.Identity(info.NodeID)
+
+		if currentPhase == flow.EpochPhaseSetup && err == nil {
+			break
```
jordanschalm, line 173: Does this reproduce the failure to see the joining node in the identity table you were mentioning?

jordanschalm (commented):
Nice work 🎸

integration/tests/epochs/epoch_test.go
kc1116: That currently doesn't work because node ID is never present so it loops forever. [Instead I wait for the sealed view of the first DKG phase to ensure we are in the epoch setup phase](https://github.com/onflow/flow-go/pull/1603/commits/d2ba2037f0211fc566cdbc0f9c017d877b762a48#diff-9b4499ee5310f158af66e16a74547dbdd0639c55e177ee8305040547fe0187efR174). Then when we start our node we consistently get the following err. 

> {"level":"fatal","node_role":"access","node_id":"e4607ca0daa0d9ec74f51fb0a11ce848b3ab28a6ecf1b916b89fdba09a7d6dc2","error":"identity not found (e4607ca0daa0d9ec74f51fb0a11ce848b3ab28a6ecf1b916b89fdba09a7d6dc2)","time":"2021-11-10T13:00:34Z","message":"node identity not found in the identity list of the finalized state: e4607ca0daa0d9ec74f51fb0a11ce848b3ab28a6ecf1b916b89fdba09a7d6dc2"}

[This validation in init local of the fnb is what's failing. ](https://github.com/onflow/flow-go/blob/master/cmd/scaffold.go#L709)

integration/tests/epochs/suite.go
```diff
@@ -165,11 +161,14 @@ func (s *Suite) StakeNode(ctx context.Context, env templates.Environment, role f
...
+	// if node has a machine account key encode it
+	var encMachinePubKey []byte
+	if machineAccountKey != nil {
+		encMachinePubKey, err = flow.EncodeRuntimeAccountPublicKey(machineAccountPubKey)
+		require.NoError(s.T(), err)
+	}
 
-	encMachinePubKey, err := flow.EncodeRuntimeAccountPublicKey(machineAccountPubKey)
-	require.NoError(s.T(), err)
+	containerName := fmt.Sprintf("%s_test", role)
```
jordanschalm, line 171: We'll need to change this for any cases where we add >1 new nodes of the same role.

integration/tests/epochs/suite.go
```diff
@@ -395,3 +387,18 @@ func (s *Suite) ExecuteReadApprovedNodesScript(ctx context.Context, env template
 
 	return v
 }
+
+func (s *Suite) WaitForEpochs(ctx context.Context, numOfEpochs int)  {
```
jordanschalm, line 391: Could we add a godoc to this function, in particular documenting how the function determines when we've reached the target epoch. Also, I think this function assumes it is invoked at epoch 0. Otherwise the condition on line 396 for the first iteration through the outer loop would never be true (eg. if this function is called in epoch >=1). Maybe we could change the function to `WaitForEpoch(epochCounter uint64)` and just remove the outer loop

integration/tests/epochs/suite.go
```diff
@@ -395,3 +387,18 @@ func (s *Suite) ExecuteReadApprovedNodesScript(ctx context.Context, env template
...
+	var epoch protocol.Epoch
+	var epochCounter uint64
+	for counter := 0; counter < numOfEpochs; counter++ {
+		// wait until the access node reaches the desired epoch
+		for epoch == nil || epochCounter != uint64(counter) {
```
jordanschalm, line 396:
```suggestion
		for epoch == nil || epochCounter < uint64(counter) {
```
Terminate as soon as we reach any epoch at or past the target (avoids infinite loop, if the network is progressing).

integration/testnet/network.go
```diff
@@ -467,29 +485,39 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
 
 	// create a temporary directory to store all bootstrapping files, these
 	// will be shared between all nodes
```
jordanschalm, line 487:
```suggestion
	// create a temporary directory to store all bootstrapping files 
```
this directory isn't shared any more given 
https://github.com/onflow/flow-go/blob/4587d40755e58c89c56ad66f2ecf79372fc4fcb6/integration/testnet/network.go#L632-L637

integration/testnet/network.go
```diff
@@ -467,29 +485,39 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
...
+		BootstrapDir:               bootstrapDir,
+	}
+
+	// check that at-least 2 full access nodes must be configure in your test suite
+	// in order to provide a secure GRPC connection for LN & SN nodes
+	accessNodeIDS := make([]string, 0)
+	for _, n := range confs {
+		if n.Role == flow.RoleAccess && !n.Ghost {
+			accessNodeIDS = append(accessNodeIDS, n.NodeID.String())
+		}
```
jordanschalm, line 518: It seems like if we use the `*` option, to use all available AN IDs, we can avoid this check (for non-Mainnet networks).
https://github.com/onflow/flow-go/blob/e3627a7ada8d2a8178e7221fa05640bbd9270983/cmd/util/cmd/common/validation.go#L17-L25

I think for the integration tests, we should make use of this. We'll have one less container eating resources and a bit less complexity in the test setup code.

jordanschalm:
@kc1116 Please don't re-run tests for the latest commit (https://github.com/onflow/flow-go/pull/1603/commits/32755459ef043dd0c7dec2fffb52cbe65e213c38). That integration test failure is exactly what I've been looking for, would like to keep the reference URL accessible 🙏

jordanschalm (approved):
Great work with this - big milestone here! 🎸 

One non-blocking suggestion and one outdated comment.

integration/testnet/network.go
kc1116: Yes great idea !

kc1116: https://github.com/onflow/flow-go/pull/1603/commits/e15d3c983f39bc37be9c3610359504c130ad56a1

gomisha (commented):
from 1:1 live PR review session

integration/tests/epochs/epoch_test.go
```diff
@@ -136,31 +141,81 @@ func (s *Suite) TestEpochJoin() {
...
+	nodeConfig := testnet.NewNodeConfig(role, containerConfigs...)
+	testContainerConfig := testnet.NewContainerConfig(info.ContainerName, nodeConfig, info.NetworkingKey, info.StakingKey)
+	err := testContainerConfig.WriteKeyFiles(s.net.BootstrapDir, flow.Localnet, info.MachineAccountAddress, encodable.MachineAccountPrivKey{PrivateKey: info.MachineAccountKey}, role)
+	require.NoError(s.T(), err)
 
-	// get new approved nodes list and make sure new node was added correctly
-	approvedNodes := s.ExecuteReadApprovedNodesScript(ctx, env)
+	// download snapshot to bootstrap from
+	var snapshot *inmem.Snapshot
+	for {
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1603/commits/22c2624d783bb82ebabc59b113b843fc22f3903a

integration/tests/epochs/epoch_test.go
```diff
@@ -136,31 +141,81 @@ func (s *Suite) TestEpochJoin() {
...
+	client, err := testnet.NewClient(clientAddr, s.net.Root().Header.ChainID.Chain())
+	require.NoError(s.T(), err)
+	snapshot, err = client.GetLatestProtocolSnapshot(ctx)
+	require.NoError(s.T(), err)
+
+	head, err := snapshot.Head()
+	require.NoError(s.T(), err)
+
+	// head should be higher than bootstrapHead if AN was staked, healthy and processing block proposals
+	require.True(s.T(), bootstrapHead.Height < head.Height, fmt.Sprintf("expected head.Height %d to be higher than head from the snapshot the node was bootstraped with bootstrapHead.Height %d.", head.Height, bootstrapHead.Height))
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1603/commits/22c2624d783bb82ebabc59b113b843fc22f3903a

integration/tests/epochs/epoch_test.go
```diff
@@ -136,31 +141,81 @@ func (s *Suite) TestEpochJoin() {
...
+
+	// execute script directly on new AN to ensure it's functional
+	proposedTable, err = s.client.ExecuteScriptBytes(ctx, templates.GenerateReturnProposedTableScript(env), []cadence.Value{})
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1603/commits/22c2624d783bb82ebabc59b113b843fc22f3903a

gomisha (approved):

kc1116:
> was 20 just chosen arbitrarily? is there any significance to 20?

20 was indeed chosen arbitrarily, there isn't really a special number because of the different variations of timing.

kc1116:
> nit: why do we need to double the timeout value? why can't it be `waitTimeout`?

I made 2 min the default https://github.com/onflow/flow-go/pull/1603/commits/924fda097a8ef00d0b371bb2f8a0935672b69614

PR #1605, opened 2021-11-09
add bls crypto functions to transaction and script env

dsainati1:
In order to address https://github.com/onflow/cadence/issues/1127, https://github.com/onflow/cadence/issues/1128 and https://github.com/onflow/cadence/issues/1129, we need to add the relevant functions to `scriptEnv.go` and `transactionEnv.go` 

The Cadence portion of this work is in https://github.com/onflow/cadence/pull/1236

This adds functions to these environments that delegate to the proper crypto library.


crypto/sign_norelic.go
```diff
@@ -1,3 +1,4 @@
+//go:build !relic
 // +build !relic
```
janezpodhostnik, line 2: Not sure why both are needed. But my IDE(goland) has a problem with both while just `// +build !relic` works as expected.

janezpodhostnik (commented):
Will the cadence version update where the `runtime.Interface` is updated to include this methods be included at a later date?

Usually I included the cadence version update in a PR like this. This allowed me to write tests like this one: https://github.com/onflow/flow-go/blob/192cd839169d49207b4e2681b7d96cfc86f84f90/fvm/fvm_test.go#L1380.

dsainati1:
> Will the cadence version update where the runtime.Interface is updated to include this methods be included at a later date?

I have local changes to the interface on the Cadence side, but was not aware that it was possible to include changes to both repositories in a single PR. I will look into getting those changes into this PR.

crypto/sign_norelic.go
dsainati1: Hmm it seems like this was automatically added by VSCode.

crypto/sign_norelic.go
```diff
@@ -10,3 +11,18 @@ func newSigner(algo SigningAlgorithm) (signer, error) {
...
+
+// AggregateSignatures aggregate multiple signatures into one; currently only works for BLS
+func AggregateSignatures(sigs [][]byte) (Signature, error) {
+	panic("AggregateSignatures not supported with non-relic build")
+}
+
+// AggregatePublicKeys aggregate multiple public keys into one; currently only works for BLS
+func AggregatePublicKeys(keys []PublicKey) (PublicKey, error) {
+	panic("AggregatePublicKeys not supported with non-relic build")
+}
```
tarakby, line 28: In the crypto package, these function should not be exported when Relic tag is not enabled.
I understand the need in FVM is different (these functions should always be defined). Since it is an FMV issue, I suggest we define these functions (relic and non-relic) in `fvm/crypto`. This is already done for BLS KMAC ([relic](https://github.com/onflow/flow-go/blob/cadence-issue-1127/fvm/crypto/blskmac_relic.go) and [non-relic](https://github.com/onflow/flow-go/blob/cadence-issue-1127/fvm/crypto/blskmac.go)).
The files can be renamed to reflect that they now contain multiple crypto tools and not just BLS_KMAC.

crypto/sign_relic.go
```diff
@@ -39,3 +40,22 @@ func initRelic() {
...
+	for _, sig := range sigs {
+		s = append(s, sig)
+	}
+	return AggregateBLSSignatures(s)
+}
+
+// AggregatePublicKeys aggregate multiple public keys into one; currently only works for BLS
+func AggregatePublicKeys(keys []PublicKey) (PublicKey, error) {
+	return AggregateBLSPublicKeys(keys)
+}
```
tarakby, line 61: As mentioned above, these functions could also be moved under `fvm/crypto`.

fvm/scriptEnv.go
```diff
@@ -685,3 +686,35 @@ func (e *ScriptEnv) Commit() ([]programs.ContractUpdateKey, error) {
 func (e *ScriptEnv) AllocateStorageIndex(_ []byte) (uint64, error) {
 	return 0, errors.NewOperationNotSupportedError("AllocateStorageIndex")
 }
+
+func (e *ScriptEnv) BLSVerifyPOP(pk *runtime.PublicKey, sig []byte) (bool, error) {
+	key, err := crypto.DecodePublicKey(crypto.BLSBLS12381, pk.PublicKey)
```
tarakby, line 691: Not sure how the upper layers of Cadence would handle returned errors. But I know we would like to differentiate 2 kind of errors:
 - recoverable errors: the crypto package returns a sentinel error if any input is invalid (caller mistake) and the node shouldn't crash.
 - non recoverable errors: an unexpected exception during execution, that can cause the node to crash.
 
 This is done here for instance https://github.com/onflow/flow-go/blob/cadence-issue-1127/fvm/crypto/crypto.go#L104-L109.
 
 cc @janezpodhostnik

fvm/scriptEnv.go
```diff
@@ -685,3 +686,35 @@ func (e *ScriptEnv) Commit() ([]programs.ContractUpdateKey, error) {
...
+	if err != nil {
+		return false, err
+	}
+	return crypto.VerifyPOP(key, sig)
```
tarakby, line 695: Same here, we probably should check for the sentinel error (we can also make the check under fvm/crypto)

fvm/scriptEnv.go
```diff
@@ -685,3 +686,35 @@ func (e *ScriptEnv) Commit() ([]programs.ContractUpdateKey, error) {
...
+}
+
+func (e *ScriptEnv) AggregateBLSSignatures(sigs [][]byte) ([]byte, error) {
+	return crypto.AggregateSignatures(sigs)
```
tarakby, line 699: Same here, we probably should check for the sentinel error (we can make the check under fvm/crypto)

fvm/scriptEnv.go
```diff
@@ -685,3 +686,35 @@ func (e *ScriptEnv) Commit() ([]programs.ContractUpdateKey, error) {
...
+}
+
+func (e *ScriptEnv) AggregateBLSPublicKeys(keys []*runtime.PublicKey) (*runtime.PublicKey, error) {
+	pks := make([]crypto.PublicKey, 0, len(keys))
+	for _, key := range keys {
+		pk, err := crypto.DecodePublicKey(crypto.BLSBLS12381, key.PublicKey)
```
tarakby, line 705: Same here, we probably should check for the sentinel error

fvm/scriptEnv.go
```diff
@@ -685,3 +686,35 @@ func (e *ScriptEnv) Commit() ([]programs.ContractUpdateKey, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		pks = append(pks, pk)
+	}
+	pk, err := crypto.AggregatePublicKeys(pks)
```
tarakby, line 711: Same here, we probably should check for the sentinel error (we can make the check under fvm/crypto)

fvm/transactionEnv.go
```diff
@@ -950,3 +951,35 @@ func (e *TransactionEnv) Commit() ([]programs.ContractUpdateKey, error) {
 	}
 	return e.contracts.Commit()
 }
+
+func (e *TransactionEnv) BLSVerifyPOP(pk *runtime.PublicKey, sig []byte) (bool, error) {
```
tarakby, line 955: These methods seem to be redundant with `ScriptEnv` methods, it would be good to factorize this code to make it use the same code (better maintainability).

tarakby (commented):
Thanks for the updates, it's nice to see these features coming soon to Cadence 🏄🏼 

I have left a few comments below.

fvm/scriptEnv.go
dsainati1: > recoverable errors: the crypto package returns a sentinel error if any input is invalid (caller mistake) and the node shouldn't crash.

Would you want these functions to return `nil` on an invalid input?

dsainati1:
I have put up a PR for the Cadence side of this work in https://github.com/onflow/cadence/pull/1236

In order to test these changes, I needed to update to use my Cadence changes, which meant I needed to rebase this PR onto the `v0.23-testnet` branch. I rebased over https://github.com/onflow/flow-go/pull/1613 to fix some build errors.

fvm/transactionEnv.go
```diff
@@ -997,3 +997,15 @@ func (e *TransactionEnv) Commit() ([]programs.ContractUpdateKey, error) {
...
+	return crypto.VerifyPOP(pk, sig)
+}
+
+func (e *TransactionEnv) AggregateBLSSignatures(sigs [][]byte) ([]byte, error) {
+	return crypto.AggregateSignatures(sigs)
+}
+
+func (e *TransactionEnv) AggregateBLSPublicKeys(keys []*runtime.PublicKey) (*runtime.PublicKey, error) {
+	return crypto.AggregatePublicKeys(keys)
+}
```
ramtinms, line 1011: I wonder if we can prefix all of these methods with BLS and then add the method names (e.g. `BLSAggregateSignatures`) so we naturally have some grouping of these methods,

ramtinms (approved):
Looks good to me, just make sure we port this over to the master as well.

fvm/fvm_test.go
```diff
@@ -1825,6 +1825,331 @@ func TestSignatureVerification(t *testing.T) {
...
+						err = vm.Run(ctx, script, view, programs)
+						switch signatureAlgorithm.algorithm {
+						case crypto.BLSBLS12381:
+							assert.NoError(t, err)
+							assert.NoError(t, script.Err)
+
+							aggSk, err := crypto.AggregateBLSPrivateKeys(sks)
+							require.NoError(t, err)
+							expectedSig, err := aggSk.Sign(input, hasher)
+							require.NoError(t, err)
```
tarakby, line 2053: Nice solution! 
Another (faster) way would be to aggregate the signatures straight away.

```suggestion
							expectedSig, err := crypto.AggregateBLSSignatures(sigs)
							require.NoError(t, err)
```

tarakby: Well, I figured we will need to construct `sigs` too.

fvm/fvm_test.go
```diff
@@ -1825,6 +1825,331 @@ func TestSignatureVerification(t *testing.T) {
...
+
+							assert.Equal(t, testutil.BytesToCadenceArray(expectedSig), script.Value)
+						case crypto.ECDSAP256, crypto.ECDSASecp256k1:
+							assert.NoError(t, err)
+							assert.Error(t, script.Err)
```
tarakby, line 2058: Here too we should be testing `script.Value` against Cadence default value (`nil` maybe?)

fvm/fvm_test.go
```diff
@@ -1825,6 +1825,331 @@ func TestSignatureVerification(t *testing.T) {
...
+						switch signatureAlgorithm.algorithm {
+						case crypto.BLSBLS12381:
+							assert.NoError(t, err)
+							assert.NoError(t, script.Err)
+
+							aggSk, err := crypto.AggregateBLSPrivateKeys(sks)
+							require.NoError(t, err)
+							expectedPk := aggSk.PublicKey()
+							require.NoError(t, err)
+
```
tarakby, line 2136: What is implemented is correct, though we can make a shortcut. 
The tests in the crypto library are made differently because they are testing the arithmetic relationships, while here we're just testing the FVM functions are wired correctly.
```suggestion
							expectedPk, err := crypto.AggregateBLSPublicKeys(sks)
							require.NoError(t, err)
```

fvm/fvm_test.go
```diff
@@ -1825,6 +1825,331 @@ func TestSignatureVerification(t *testing.T) {
...
+							assert.Equal(t, testutil.BytesToCadenceArray(expectedPk.Encode()), script.Value)
+						case crypto.ECDSAP256, crypto.ECDSASecp256k1:
+							assert.NoError(t, err)
+							assert.Error(t, script.Err)
```
tarakby, line 2140: Same comment about testing the `script.value` against Cadence default vaule.

tarakby (commented):
Thanks for adding the tests and making the changes!
I left a few comments about the tests, most of them are minor.  

- I would like to add a few tests (making a PR against yours), though I am having issues importing Cadence package to test locally:
```
go: github.com/onflow/cadence@v0.20.0-beta10: reading github.com/dapperlabs/cadence-internal/go.mod at revision v0.20.0-beta10-patch.1: unknown revision v0.20.0-beta10-patch.1
```

- Regarding testing against the crypto sentinel error for invalid inputs, I answer your question here:
> Would you want these functions to return nil on an invalid input?

Yes, it should be `false` for `verifyPoP` and `nil` for the others.

janezpodhostnik (commented):
Besides the comments that Tarak has, the res looks good to me.

I agree that incorrect use input into `BLSVerifyPOP` should not produce an error on cadence side, but should just return false.

dsainati1:
> I would like to add a few tests (making a PR against yours), though I am having issues importing Cadence package to test locally:

I just used the master build of cadence, replacing Cadence in the go.mod file: `go mod edit -replace github.com/onflow/cadence=../cadence`

tarakby:
Here is the PR extending the tests : https://github.com/onflow/flow-go/pull/1639 
Please feel free to review it and we can merge it into your branch.

tarakby:
While we figure out the error return types to Cadence to differentiate user's errors and unexpected execution errors, I would like to add another specification. 
I am not sure if it should be added to the FVM (this PR) or the runtime (https://github.com/onflow/cadence/pull/1236) though:

In the `PublicKey` structure in Cadence, there is a field called `isValid`, that is assigned to true or false when the structure is created (Cadence doc: https://docs.onflow.org/cadence/language/crypto/). You can notice that when the field is set to false, and the function `verify` is called on the public key, the function returns false. Similarly: 
 - `verifyPoP` should return false when the public key field `isValid` is false.
 - `BLSAggregatePublicKeys` should return nil (not sure whether it needs to error too) if at least one input public key has a false `isValid` field.

go.mod
```diff
@@ -94,4 +94,4 @@ replace github.com/onflow/flow-go/crypto => ./crypto
 
 replace github.com/onflow/cadence/v19 => github.com/onflow/cadence v0.19.0
 
-replace github.com/onflow/cadence => github.com/dapperlabs/cadence-internal v0.20.0-beta10-patch.1
+replace github.com/onflow/cadence => ../cadence
```
turbolent, line 97: This change for local development purposes should not be committed

dsainati1:
> While we figure out the error return types to Cadence to differentiate user's errors and unexpected execution errors, I would like to add another specification. I am not sure if it should be added to the FVM (this PR) or the runtime ([onflow/cadence#1236](https://github.com/onflow/cadence/pull/1236)) though:

I have added these changes to https://github.com/onflow/cadence/pull/1250 on the Cadence side. The condition is easy to check just using the `PublicKey` values and abort execution if they are invalid, so no need to call into the FVM.

janezpodhostnik (approved):
The error handling looks like what we discussed, so LGTM.

dsainati1:
This is now waiting on https://github.com/onflow/cadence/pull/1250 to add the corresponding changes to the Cadence side, which is in turn blocked on https://github.com/onflow/cadence/pull/1273. Once these two PRs are merged and included in the next Cadence release, we should be able to merge this.

tarakby:
This PR is awaiting a PR to update Cadence, being worked on by @janezpodhostnik.

dsainati1:
We are also planning on releasing the BLS changes to Cadence in version 0.23

janezpodhostnik:
> This PR is awaiting a PR to update Cadence, being worked on by @janezpodhostnik.

PR https://github.com/onflow/flow-go/pull/1975 is ready, except the crypto changes. But those are actually here. It might make sense to either base https://github.com/onflow/flow-go/pull/1975 on this one or the other way around. What would you prefer?

turbolent:
Updated this PR to the latest version of Cadence, v0.23.0, which we would like to get into the next spork.

There is one remaining ytes

turbolent:
@tarakby Could you please double check [fe42f9d](https://github.com/onflow/flow-go/pull/1605/commits/fe42f9dc85c8de91579ea2b8d7836a7209c70dc6)?

fvm/bootstrap.go
```diff
@@ -678,8 +678,8 @@ transaction(clusterWeights: [{String: UInt64}]) {
 		randomSource: %s,
 		collectorClusters: clusters,
         // NOTE: clusterQCs and dkgPubKeys are empty because these initial values are not used
-		clusterQCs: [],
-		dkgPubKeys: [],
+		clusterQCs: [] as [FlowClusterQC.ClusterQC],
+		dkgKeys: [] as [String],
```
turbolent, line 682: Was this change necessary? It shouldn't be

turbolent (approved):
Looks good! Just one question regarding the Cadence code-change

fvm/bootstrap.go
janezpodhostnik: Might not be for cadence 0.23. Let me check

janezpodhostnik: removed [6d883fd](https://github.com/onflow/flow-go/pull/1605/commits/6d883fd26a95beabcc6c198247ce1e7d1edef6c4)

tarakby:
> @tarakby Could you please double check [fe42f9d](https://github.com/onflow/flow-go/pull/1605/commits/fe42f9dc85c8de91579ea2b8d7836a7209c70dc6)?

Thanks, I'm gonna make a minor change.

fvm/crypto/crypto.go
```diff
@@ -99,19 +99,22 @@ func CryptoToRuntimeHashingAlgorithm(h hash.HashingAlgorithm) runtime.HashAlgori
 	}
 }
 
-// ValidatePublicKey returns true if public key is valid
-func ValidatePublicKey(signAlgo runtime.SignatureAlgorithm, pk []byte) (valid bool, err error) {
+// ValidatePublicKey returns :
+// - nil if key is valid and no exception occurred.
+// - crypto.invalidInputsError if key is invalid and no exception occurred.
+// - panics if an exception occurred.
+func ValidatePublicKey(signAlgo runtime.SignatureAlgorithm, pk []byte) error {
```
turbolent, line 106: Good idea to refactor this also here! 👍

turbolent (approved):

PR #1607, opened 2021-11-09
Use latest finalized (not sealed) reference block

jordanschalm:
In cases where sealing is behind, using the latest sealed block can result in generating immediately-expired transactions. It is safe to use the latest finalized block for the purposes of the tx reference block.


tarakby (approved):

zhangchiqing (approved):

jordanschalm:
bors merge

PR #1608, opened 2021-11-10
[Verification] Removes less and swap methods of chunk list

yhassanzadeh13:
This PR was originally meant to optimize the `Less` method of `ChunkList` with respect to `ID` method invocations (see this issue https://github.com/dapperlabs/flow-go/issues/5946). `Less` method of `ChunkList` was introduced earlier [by me](https://github.com/onflow/flow-go/commit/2a14fd5e92b8f28e6e3bb8ea6c52f33795720900) at the time we develop the very first version of the verification node. At that time, the purpose was to make the `ChunkList` implement the `sort` interface, hence, being sortable. 

However, after reviewing the use cases of `ChunkList` in our current codebase, we no longer require such sorting capability. This is also supported by the fact that chunks are uniquely addressed by their index within an execution result, so there is no longer a need for sorting them based on their `ID`. 

To resolve the ambiguities around `ChunkList` being sortable and its _inefficient_ `Less` method implementation, this PR removes the `Less` and `Swap` methods of `ChunkList` hence disabling its sortable capability that no longer is needed.

### DIFF
--- a/model/flow/chunk.go
+++ b/model/flow/chunk.go
@@ -111,16 +111,3 @@ func (cl ChunkList) ByIndex(i uint64) (*Chunk, bool) {
 func (cl ChunkList) Len() int {
 	return len(cl)
 }
-
-// Less returns true if element i in the ChunkList is less than j based on its chunk ID.
-// Otherwise it returns true.
-// It satisfies the sort.Interface making the ChunkList sortable.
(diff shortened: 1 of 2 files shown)


jordanschalm (approved):

jwinkler2083233 (approved):
Much better

PR #1612, opened 2021-11-10
[voting V2] Use the crypto package constant for signature length

tarakby:
This PR replaces `48` by the crypto constant for BLS signature length (the constant should not be assumed to be always 48).


durkmurder (approved):
LGTM. One thing to consider is to put sig len as constant into type_encoder.go, this way we can reuse global constant for tests and impl.

zhangchiqing (approved):

tarakby:
bors merge

PR #1618, opened 2021-11-12
[State Sync] Add badger datastore

### LINKED ISSUE #1450: [State sync engine] Implement block stores
We will need to implement the [Blockstore](https://pkg.go.dev/github.com/ipfs/go-ipfs-blockstore) interface for various entities in order to use them with bitswap:
- Collections
- Events
- TrieUpdates
- TransactionResult?
- Need a separate Blockstore for MerkleDAG nodes which are used for hierarchical requests

TODO: Think about how we can reuse the existing badger storage inside the blockstore implementation, to avoid having to store a duplicate copy of the data in every state diff (one version in the blockstore, and another version of each of the entities in the state diff in the badger storage).
* Can we add an API to the Storage interfaces to query for data by hash instead of by ID? Currently data in badger is keyed by flow ID, but perhaps there's a way to add a Secondary key?
* We may be able to implement this secondary index ourselves: https://github.com/dgraph-io/badger/issues/738#issuecomment-496704883
    * Add a second db that maps from the hash to the flow ID.

Can we do the same for state checkpoints as well?

peterargue:
Closes #1450 

Implements a blockstore using badgerDB for storing state sync blocks.


smnzhu (commented):
Haven't looked too closely, but this will probably need changes after upcoming updates to https://github.com/onflow/flow-go/pull/1582. 

In particular, we will probably want to implement a `Datastore` instead of a Blockstore

network/p2p/block_exchange.go
```diff
@@ -57,7 +57,7 @@ func (e *BlockExchange) GetBlocks(cids ...cid.Cid) network.BlocksPromise {
 }
 
 func (e *BlockExchange) HasBlock(block blocks.Block) error {
-	return e.bs.HasBlock(block)
+	return e.bs.HasBlock(context.Background(), block)
```
yhassanzadeh13, line 60: context is better to be added as an input parameter so that caller may apply a timeout on it.

network/p2p/libp2pNode.go
```diff
@@ -797,8 +797,9 @@ func DefaultLibP2POptions(address string, key fcrypto.PrivateKey) ([]config.Opti
 	// as the 1-k discovery process and the 1-1 messaging both sometimes attempt to open connection to the same target
 	// As of now there is no requirement of client sockets to be a well-known port, so disabling port reuse all together.
 	transport := libp2p.Transport(func(u *tptu.Upgrader) *tcp.TcpTransport {
-		tpt := tcp.NewTCPTransport(u)
-		tpt.DisableReuseport = true
+		// TODO: handle errors properly. DisableReuseport() does not return an error, so it's safe
+		// to ignore for now
+		tpt, _ := tcp.NewTCPTransport(u, tcp.DisableReuseport())
```
yhassanzadeh13, line 802: We'd better return the error. The desired case is to fail safely when we cannot bootstrap the node completely.

network/test/block_exchange_test.go
```diff
@@ -52,7 +52,7 @@ func (suite *BlockExchangeTestSuite) SetupTest() {
 		suite.cleanupFuncs = append(suite.cleanupFuncs, cleanupFunc)
 		block := blocks.NewBlock([]byte(fmt.Sprintf("foo%v", i)))
 		suite.blockCids = append(suite.blockCids, block.Cid())
-		require.NoError(suite.T(), bstore.Put(block))
+		require.NoError(suite.T(), bstore.Put(context.Background(), block))
```
yhassanzadeh13, line 55: we'd better using a `context.WithCancel` hence differing `cancel` method to make sure resources are released safely at the end of test.

peterargue:
closing in favor of implementing this directly in the `github.com/ipfs/go-ds-badger2` library

PR #1619, opened 2021-11-15
REST: adding middlewares to the API implementation

vishalchangrani:
Introducing a couple of features of Gorilla mux to abstract out some common functionality:

1. Gorilla Mux lets you define [Middleware](https://github.com/gorilla/mux#middleware)which are common things that need to be done for request. I defined one to extract out the `expand` and `select` query params if present in a request. Also, the logging interceptor has been converted to a middleware.
2. Using subroutes to split routes (/`v1` and `/GetBlocks`).
3. Merged in @sideninja PR - [1620](https://github.com/onflow/flow-go/pull/1620) and [1627](https://github.com/onflow/flow-go/pull/1627)


engine/access/rest/middlewares.go
```diff
@@ -0,0 +1,60 @@
...
+func NotImplementedHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+		w.WriteHeader(http.StatusNotImplemented)
+	})
+}
+
+// LoggingMiddleware creates a middleware which adds a logger interceptor to each request to log the request method, uri,
+// duration and response code
+func LoggingMiddleware(logger zerolog.Logger) mux.MiddlewareFunc {
```
sideninja, line 20: Could we divide this logger to not wrap our request handler but instead have one middleware before the call (prepare all thins like start the timer, create new response writer and also create new request logger but also any additional boilerplate things repeated in each request) and one after that logs the time.

vishalchangrani: I don't think wrapping the serverHTTP is avoidable. The middleware (unlike filters) build on each other and [their example ](https://github.com/gorilla/mux#middleware) also does the same.

engine/access/rest/common_query_params_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCommonQueryParamMiddlewares test the two common query param middlewares - expand and select
+func TestCommonQueryParamMiddlewares(t *testing.T) {
```
vishalchangrani, line 14: test for the new middleware. Testing handles as described [here](https://github.com/gorilla/mux#testing-handlers)

engine/access/rest/rest_api.go
```diff
@@ -22,11 +22,11 @@ var MaxAllowedBlockIDsCnt = BlockIDCntLimit
 
 // RestAPIHandler provides the implementation of each of the REST API
 type APIHandler struct {
-	backend *backend.Backend
+	backend access.API
```
vishalchangrani, line 25: interface instead of implementation

engine/access/rest/rest_api_server.go
```diff
@@ -40,114 +46,71 @@ func apiRoutes(api *APIHandler) generated.Routes {
 		generated.Route{
 			Name:        "AccountsAddressGet",
 			Method:      strings.ToUpper("Get"),
-			Pattern:     "/v1/accounts/{address}",
+			Pattern:     "/accounts/{address}",
```
vishalchangrani, line 49: since these are all now subroutes of route on path `v1`

engine/access/rest/middlewares.go
bluesign: I think what @sideninja is telling is: 

Middleware logic in general is chain of middlewares:

- Do something, move to next handler

here it is more like: 

- Do Something, run next handler, Do something.

As long as Logging is the middleware there is not a big problem. But if there will be more middlewares, @sideninja's suggestion allows adding multiple logs (traces) to request, so you can see which middleware is lagging. 

If I understood correctly he suggests something like: 

`MWStartLogging, RequestHandler, MWLog`

sideninja: I was thinking we could/should create a logger instance in the first logging middleware that then gets passed down in the chain of middlewares and the final handler anyhow uses a common response method which can after writing a response also logs the finish time etc. This way we don't have to initialize loggers anywhere else and as @bluesign 
also mentioned other middlewares can potentially use the instance passed down in the chain with the context value.

engine/access/rest/common_query_params.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type ctxKeyType string
+
+const expandQueryParam = "expand"
+const selectQueryParam = "select"
+
+// commonQueryParamMiddleware generates a Middleware function that extracts the given query parameter from the request
+// and adds it to the request context as a key value pair with the key as the query param name.
+// e.g. for queryParamName "fields", if the request url contains <some url>?fields=field1,field2,..fieldN,
```
bluesign, line 18: Are you sure you want to implement field and expand here ?

I think those should be another layer (some another real middleware), which can act as a proxy to this service. 

Most of the data you have here can be cached ( in redis for example ) on middleware for very long time. ( even the cache can be invalidated on demand from the engine, though it is a bit more advanced )

But in the end, selecting and expanding from cached result ( as a lot of the data is final ) would be much more performant

bluesign: Also for example if someone gets `getBlock` response for Block X , another one requests `getBlockHeader` for Block X, you will already have the required information from cache.

vishalchangrani: so I would like the base implementation (this) to be able to serve the full api without the need of an external middleware. What you suggested could still be achieve with the a middleware in place, couldn't it?

engine/access/rest/middlewares.go
vishalchangrani: make sense..I have split the `LoggingMiddleware` to `RequestStart` and `RequestEnd` so that it avoids the Do Something, run next handler, Do something again pattern.

engine/access/rest/common_query_params.go
bluesign: I think then it would be better to make select and expand full middleware, so they would manipulate data afterwards. 

IMO middleware should be pluggable ( and independent) in general. Maybe something like below: 

- Middleware: Cache lookup ( execute if not cached )  
- Middleware: Manipulate Response for selecting fields 
- Middleware: Manipulate Response for expand / collapse fields 
- Middleware: Output json from result 
etc etc 

So later this will allow to move some parts outside if necessary as parts will be loosely couples. 

PS: This is not too critical actually as soon AN will have full state.

vishalchangrani: > I think then it would be better to make select and expand full middleware, so they would manipulate data afterwards.
> 
> IMO middleware should be pluggable ( and independent) in general. Maybe something like below:
> 
> * Middleware: Cache lookup ( execute if not cached )
> * Middleware: Manipulate Response for selecting fields
> * Middleware: Manipulate Response for expand / collapse fields
> * Middleware: Output json from result
>   etc etc
> 
> So later this will allow to move some parts outside if necessary as parts will be loosely couples.
> 
> PS: This is not too critical actually as soon AN will have full state.

Since the spec defines `select` and `expand` and since there is no middleware currently in place, I wanted to make sure that the implementation is fully compliant with the spec. When a middleware is in place, it can choose to not send the `select` and `expand` fields to this implementation and instead handle the expand and select logic by itself as you described. However, shouldn't the out-of-the-box implementation of an AN (this one) still implement select and expand?

bluesign: Yeah if you want full out-of-the-box spec compatibility, it is better to separate them in middlewares. 

What I meant here in this context was, you don't need middleware to extract query parameters. Instead you can put select and expand middlewares last. 

"Select Middleware" will select fields from response and filter, before it runs, you can cache results without filtering etc.

So if you remove "Select Middleware" everything works (except select function ofc), so each middleware Is making one job, instead of passing variables over context.

bluesign: Yeah actually that why  it was my first thought to scale this outside of AN. Currently we have with grpc everything is expanded right?

I think if there will be some good caching (last few blocks etc could be cached fully) it can decrease load a lot. 

Even I think as script execution will be on AN probably ( as state is going to sync with AN ), API should be extended with storage related functions too.  ( Considering AuthAccount for scripts on cadence just landed ) 

I foresee script execution will have very big load after this. Instead of event fetching, querying storage will be the new black :)

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,226 @@
...
+
+// GetTransactionByID gets a transaction by requested ID.
+func (h *Handlers) GetTransactionByID(w http.ResponseWriter, r *http.Request) {
+	errorLogger := h.logger.With().Str("request_url", r.URL.String()).Logger() // todo(sideninja) refactor this to be initialized for us
+
+	vars := mux.Vars(r)
+	idFromRequest := vars["id"]
+	id, err := toID(idFromRequest)
+	if err != nil {
+		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("invalid ID: %s", idFromRequest), errorLogger)
```
sideninja, line 92:
```suggestion
		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("invalid ID: %s", idFromRequest), errorLogger)
		return
```

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	tx, err := h.backend.GetTransaction(r.Context(), id)
+	if err != nil {
+		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("transaction fetching error: %s", err.Error()), errorLogger)
```
sideninja, line 97:
```suggestion
		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("transaction fetching error: %s", err.Error()), errorLogger)
		return
```

sideninja (approved):

engine/access/rest/middleware/logging_middleware.go
```diff
@@ -0,0 +1,56 @@
...
+// duration and response code
+func LoggingMiddleware(logger zerolog.Logger) mux.MiddlewareFunc {
+	return func(inner http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			// record star time
+			start := time.Now()
+			// modify the writer
+			respWriter := newResponseWriter(w)
+			// continue to the next handler
+			inner.ServeHTTP(respWriter, req)
```
vishalchangrani, line 21: switched back the logging middleware to do the `doSomething` -> `call next handler` -> `doSomething` pattern since gorilla mux middlewares are all executed _before_ the handler. 
there were other ways of accomplishing this but they would be non-standard making the code a little complicated to read. I would rather stick with the Gorilla mux way of doing it.

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+		signatures = append(signatures, signature)
+	}
+
+	return signatures, nil
+}
+
+func toTransaction(tx *generated.TransactionsBody) (flow.TransactionBody, error) {
+	args := make([][]byte, len(tx.Arguments))
+	for _, arg := range tx.Arguments {
+		// todo validate
```
peterargue, line 79: is this an input or from the EN?

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+}
+
+func transactionSignatureResponse(signatures []flow.TransactionSignature) []generated.TransactionSignature {
+	sigs := make([]generated.TransactionSignature, len(signatures))
+	for _, sig := range signatures {
+		sigs = append(sigs,
+			generated.TransactionSignature{
+				Address:     sig.Address.String(),
+				SignerIndex: int32(sig.SignerIndex),
+				KeyIndex:    int32(sig.KeyIndex),
```
peterargue, line 143: this is a `uint64` in `flow.TransactionSignature`. any chance it overflows?

vishalchangrani: actually keyIndex is the number of key that needs to be used - so I doubt if it will overflow since I would think the signer would have more than 2^32 keys. That being said, JSON doesn't support 64 bit integer data type so we will have to call this out somewhere in the swagger API.

I the protobug Access API also only supports the index as 32 bits - https://github.com/onflow/flow/blob/master/protobuf/flow/entities/transaction.proto#L26.

engine/access/rest/middleware/request_attribute.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"net/http"
+)
+
+type ctxKeyType string
+
+// addRequestAttribute adds the given attribute name and value to the request context
+func addRequestAttribute(req *http.Request, attributeName string, attributeValue interface{}) *http.Request {
+	contextKey := ctxKeyType(attributeName)
```
peterargue, line 12: is this necessary? Isn't this just casting a string to a string?

vishalchangrani: oh so this is necessary to avoid this linter error:

> SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)

engine/access/rest/converter.go
vishalchangrani: @sideninja make sense! may i tackle this separately? - https://github.com/dapperlabs/flow-go/issues/6021. Realized this will need some unit tests of its own.

sideninja: for sure

peterargue (commented):
looking good. added a few comments

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+
+	tx, err := toTransaction(&txBody)
+	if err != nil {
+		h.errorResponse(w, http.StatusBadRequest, err.Error(), h.logger) // todo(sideninja) use refactor err func
+		return
+	}
+
+	err = h.backend.SendTransaction(r.Context(), &tx)
+	if err != nil {
+		h.errorResponse(w, http.StatusBadRequest, err.Error(), h.logger)
```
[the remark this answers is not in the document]
vishalchangrani: @sideninja actually has a follow up PR which addresses error handling (translating GRPC errors to HTTP errors more accurately). https://github.com/onflow/flow-go/pull/1640/files#diff-32628f0bc7adc671dce50f45fe5e5ad51e5c0b2c0f3c06dcbae6d866bbaf5a52R76.
That should address the error reporting (or if not then I will fix it over there since a lot of this will be changing on the next PR)

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+	vars := mux.Vars(r)
+	idFromRequest := vars["id"]
+	id, err := toID(idFromRequest)
+	if err != nil {
+		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("invalid ID: %s", idFromRequest), errorLogger)
+		return
+	}
+
+	tx, err := h.backend.GetTransaction(r.Context(), id)
+	if err != nil {
```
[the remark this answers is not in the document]
vishalchangrani: yup - this is upcoming - https://github.com/onflow/flow-go/pull/1636/files#diff-578ff8bf2fbd6786c506958a949d92963f31bdcd1e68407b247a1c4a75c75496R112-R161

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// GetTransactionByID gets a transaction by requested ID.
+func (h *Handlers) GetTransactionByID(w http.ResponseWriter, r *http.Request) {
+	errorLogger := h.logger.With().Str("request_url", r.URL.String()).Logger() // todo(sideninja) refactor this to be initialized for us
+
+	vars := mux.Vars(r)
+	idFromRequest := vars["id"]
+	id, err := toID(idFromRequest)
+	if err != nil {
```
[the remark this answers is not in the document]
vishalchangrani: fixed

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+func toProposalKey(key *generated.ProposalKey) (flow.ProposalKey, error) {
+	address, err := toAddress(key.Address)
+	if err != nil {
+		return flow.ProposalKey{}, err
+	}
+
+	return flow.ProposalKey{
+		Address:        address,
+		KeyIndex:       uint64(key.KeyIndex),
+		SequenceNumber: uint64(key.SequenceNumber),
```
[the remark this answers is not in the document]
vishalchangrani: `KeyIndex` and `SequenceNumber` are zero-indexed and should take any possible uint64

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+		return flow.TransactionBody{}, err
+	}
+
+	payer, err := toAddress(tx.Payer)
+	if err != nil {
+		return flow.TransactionBody{}, err
+	}
+
+	auths := make([]flow.Address, len(tx.Authorizers))
+	for _, auth := range tx.Authorizers {
```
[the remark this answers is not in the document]
vishalchangrani: I actually found no limit in the existing Flow code that limits the number of authorizer :(.

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+	address, err := toAddress(signature.Address)
+	if err != nil {
+		return flow.TransactionSignature{}, err
+	}
+
+	return flow.TransactionSignature{
+		Address:     address,
+		SignerIndex: int(signature.SignerIndex),
+		KeyIndex:    uint64(signature.KeyIndex),
+		Signature:   []byte(signature.Signature),
```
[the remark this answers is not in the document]
vishalchangrani: added a check to make sure signature is not ridiculously large

vishalchangrani:
> looking good. added a few comments

thanks! replied to your comments

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	vars := mux.Vars(r)
+	idParam := vars["id"]
+
+	// gorilla mux retains opening and ending square brackets for ids
+	idParam = strings.TrimSuffix(idParam, "]")
+	idParam = strings.TrimPrefix(idParam, "[")
+
+	ids := strings.Fields(idParam)
```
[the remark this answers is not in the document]
vishalchangrani: looking into this now - The generated Rest client does a `fmt.Sprintf("%v", id)` for the id which is of type `[]string` resulting in a space separated list of ids instead of a comma separated one.

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+// Converter provides functionality to convert from request models generated using
+// open api spec and flow models.
 
-func toBlock(flowBlock *flow.Block) *generated.Block {
+// Flow section - converting request data to flow models with validation.
+
+func toID(id string) (flow.Identifier, error) {
+	valid, _ := regexp.MatchString(`^[0-9a-fA-F]{64}$`, id)
+	if !valid {
+		return flow.Identifier{}, fmt.Errorf("invalid ID")
```
smnzhu, line 20: nit: Do we want to add some more details here?

Also, `fmt.Errorf` is intended for use when the error string needs to be a format string. Here it is just a static string, so we can use `errors.New`

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+	}
+
+	return flow.HexStringToIdentifier(id)
+}
+
+func toAddress(address string) (flow.Address, error) {
+	valid, _ := regexp.MatchString(`^[0-9a-fA-F]{16}$`, address)
+	if !valid {
+		return flow.Address{}, fmt.Errorf("invalid address")
```
smnzhu, line 29: same comment as above

engine/access/rest/converter.go
smnzhu: Also, `HexStringToIdentifier` already returns an error, perhaps we can use that?

```
id, err := flow.HexStringToIdentifier(idStr)
if err != nil {
    return flow.Identifier{}, fmt.Errorf("invalid ID: %w", err)
}
return id, nil
```

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+		SequenceNumber: uint64(key.SequenceNumber),
+	}, nil
+}
+
+func toSignature(signature string) ([]byte, error) {
+	signatureBytes := []byte(signature)
+	// the maximum signature length is arbitrarily chosen to be 2048. This check is to ensure that the signature is not
+	// absurdly long
+	if len(signatureBytes) > 2048 {
+		return nil, fmt.Errorf("invalid signature")
```
smnzhu, line 53: Again, maybe adding detail would be nice.

Also, define a constant for 2048?

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+	}
+
+	envelopeSigs, err := toSignatures(tx.EnvelopeSignatures)
+	if err != nil {
+		return flow.TransactionBody{}, err
+	}
+
+	return flow.TransactionBody{
+		ReferenceBlockID:   flow.Identifier{},
+		Script:             []byte(tx.Script),
```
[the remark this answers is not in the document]
smnzhu: Should we? Is there actually a limit for this in the protocol?

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+
+	return flow.ProposalKey{
+		Address:        address,
+		KeyIndex:       uint64(key.KeyIndex),
+		SequenceNumber: uint64(key.SequenceNumber),
+	}, nil
+}
+
+func toSignature(signature string) ([]byte, error) {
+	signatureBytes := []byte(signature)
```
smnzhu, line 49: Is this actually correct? in the spec, the signature should be a base64 encoded string. But here, we are directly converting it to a byte array, which won't actually match the data the string is supposed to represent

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+	}
+
+	return signatures, nil
+}
+
+func toTransaction(tx *generated.TransactionsBody) (flow.TransactionBody, error) {
+	args := make([][]byte, len(tx.Arguments))
+	for _, arg := range tx.Arguments {
+		// todo validate
+		args = append(args, []byte(arg))
```
smnzhu, line 95: This is also not correct, since in the spec, arguments are base64 encoded

engine/access/rest/converter.go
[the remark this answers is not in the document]
smnzhu: Also, this is not correct since Script is base64 encoded in the spec

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	vars := mux.Vars(r)
+	idParam := vars["id"]
+
+	// gorilla mux retains opening and ending square brackets for ids
+	idParam = strings.TrimSuffix(idParam, "]")
+	idParam = strings.TrimPrefix(idParam, "[")
+
+	ids := strings.Split(idParam, ",")
```
vishalchangrani, line 51: @peterargue - switched this back to expect a comma separated list. There is an [issue](https://github.com/onflow/flow/blob/master/openapi/go-client-generated/api_blocks.go#L196) in the generated Go client code which makes it send a space separated list instead of a comma separated one.
I checked the generated Java client code and that doesn't have this issue :( . I have modified the test to instead sent comma separated value

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+		)
+	}
+
+	return sigs
+}
+
+func transactionResponse(tx *flow.TransactionBody) *generated.Transaction {
+	var args []string
+	for _, arg := range tx.Arguments {
+		args = append(args, string(arg))
```
smnzhu, line 170: we need to convert to base64 encoded string.

smnzhu (changes requested):

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+	}
+
+	var auths []string
+	for _, auth := range tx.Authorizers {
+		auths = append(auths, auth.String())
+	}
+
+	return &generated.Transaction{
+		Id:                 tx.ID().String(),
+		Script:             string(tx.Script),
```
smnzhu, line 192: base64

engine/access/rest/converter.go
```diff
@@ -3,21 +3,201 @@ package rest
...
+		Arguments:          args,
+		ReferenceBlockId:   tx.ReferenceBlockID.String(),
+		GasLimit:           int32(tx.GasLimit), // todo(sideninja) make sure this is ok
+		Payer:              tx.Payer.String(),
+		ProposalKey:        proposalKeyResponse(&tx.ProposalKey),
+		Authorizers:        auths,
+		PayloadSignatures:  transactionSignatureResponse(tx.PayloadSignatures),
+		EnvelopeSignatures: transactionSignatureResponse(tx.EnvelopeSignatures),
+		Result:             nil, // todo(sideninja) should we provide result, maybe have a wait for result http long pulling system would be super helpful (with reasonable timeout) but careful about resources and dos
```
smnzhu, line 201: Yes, also the result should only actually be included if it is `expand`ed, because I think according to the spec, this field is not `expand`ed by default.

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/onflow/flow-go/access"
+	"github.com/onflow/flow-go/engine/access/rest/generated"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+const BlockIDCntLimit = 50
+
+var MaxAllowedBlockIDsCnt = BlockIDCntLimit
```
smnzhu, line 22: Why do we define this twice?

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+	errorLogger := h.logger.With().Str("request_url", r.URL.String()).Logger()
+
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	vars := mux.Vars(r)
+	idParam := vars["id"]
+
+	// gorilla mux retains opening and ending square brackets for ids
+	idParam = strings.TrimSuffix(idParam, "]")
+	idParam = strings.TrimPrefix(idParam, "[")
```
smnzhu, line 48: Why are there opening and ending square brackets in the first place? They should not have been present in the query string...

Take a look at the table here: https://swagger.io/docs/specification/serialization/

The request should just look something like "v1/blocks/id1,id2,id3"

But maybe I'm misunderstanding how this works.

engine/access/rest/handlers.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if len(ids) > MaxAllowedBlockIDsCnt {
+		h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("at most %d Block IDs can be requested at a time", MaxAllowedBlockIDsCnt), errorLogger)
+		return
+	}
+
+	for i, id := range ids {
+		flowID, err := flow.HexStringToIdentifier(id)
+		if err != nil {
+			h.errorResponse(w, http.StatusBadRequest, fmt.Sprintf("invalid ID %s", id), errorLogger)
```
smnzhu, line 62: This should include the details from the err returned from `toID`?

engine/access/rest/converter.go
vishalchangrani: The transaction api is not yet completely implemented. It doesn't has unit tests either..it is WIP.

engine/access/rest/converter.go
vishalchangrani: added const for 2048

engine/access/rest/converter.go
vishalchangrani: Created this issue to tackle the base64 encoding - https://github.com/dapperlabs/flow-go/issues/6022

engine/access/rest/converter.go
```diff
@@ -3,21 +3,186 @@ package rest
...
+
+		signatures = append(signatures, signature)
+	}
+
+	return signatures, nil
+}
+
+func toTransaction(tx *generated.TransactionsBody) (flow.TransactionBody, error) {
+	args := make([][]byte, len(tx.Arguments))
+	for _, arg := range tx.Arguments {
```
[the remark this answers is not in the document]
vishalchangrani: added 100 as the limit for now. (again don't see any limit in the rest of the code..so just putting in an arbitrary max for now)

engine/access/rest/converter.go
vishalchangrani: can i tackle it as part of https://github.com/dapperlabs/flow-go/issues/6022?

engine/access/rest/handlers.go
vishalchangrani: I do that to make the code a little easier to test - in my unittest, I can lower the limit by adjusting this variable (but not change the constant)

engine/access/rest/handlers.go
vishalchangrani: this is because of how the Generated Go Client is sending the request. It does a `fmt.Sprintf("%v", id)` [here](https://github.com/onflow/flow/blob/master/openapi/go-client-generated/api_blocks.go#L196) which causes the ids to be written out as `[id1 id2 id3]` etc. given how `%v` [works](https://play.golang.org/p/NMNcaMeUdNM) for `[]string`.

Alternatively, I could write my own client or use a different tool to generate the client but to make progress on the implementation I added line 47 and 48  to just remove the extra chars from the request if present (if not then its no-op)

engine/access/rest/converter.go
vishalchangrani: The expanded and select is being worked upon in this PR - https://github.com/onflow/flow-go/pull/1645/files. It is based off a branch which is based off of this branch :)

engine/access/rest/converter.go
peterargue: I believe binary signatures are supposed to be 64bytes, so len could be the base 64 equivalent of that

engine/access/rest/converter.go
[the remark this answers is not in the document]
peterargue: ok, we should enforce some max, even if it's large. just don't want someone to be able to send a very large list that runs up memory usage

engine/access/rest/handlers.go
peterargue: makes sense and I think the workaround is ok for now. can you just add a comment explaining that to the code so we remember later.

if you haven't already, can you create an issue to track fixing the client?

engine/access/rest/converter.go
vishalchangrani: yup ur right ..i just checked [here](https://github.com/onflow/flow-go/blob/7054d07329903c999e6cdceb226292deebfa705d/crypto/types.go)

engine/access/rest/converter.go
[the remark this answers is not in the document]
vishalchangrani: ok will throw in some reasonable limit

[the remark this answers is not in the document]
vishalchangrani: added 100 as a limit to number of authorizers

engine/access/rest/handlers.go
vishalchangrani: https://github.com/onflow/flow/issues/698 - on the flow repo to fix the client (assigned to me)

vishalchangrani: and added comment to code explaining the same.

engine/access/rest/converter.go
```diff
@@ -2,22 +2,219 @@ package rest
...
+	for _, auth := range tx.Authorizers {
+		a, err := toAddress(auth)
+		if err != nil {
+			return flow.TransactionBody{}, err
+		}
+
+		auths = append(auths, a)
+	}
+
+	payloadSigs, err := toTransactionSignatures(tx.PayloadSignatures)
```
peterargue, line 135: what about limits on payload/envelope signature lists? should the max be len(authorizers)+2 (payer+proposer) or something like that?

engine/access/rest/converter.go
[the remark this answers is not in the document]
peterargue: we do if we want to avoid someone uploading an extremely large script. ANs have finite amounts of memory so we need to protect them from easy DoS attacks

engine/access/rest/converter.go
vishalchangrani: can't the payer, propose and authorizer all do a multi sig singing with multiple keys?

vishalchangrani: also - we have this issue to add a validation layer - https://github.com/dapperlabs/flow-go/issues/6006

peterargue: I"m not sure. I'm ok to punt on this for now and circle back later if we don't have a clear picture.

I'm taking the opportunity to hash some of this out now, but the gRPC interface has the same lack of limits so we'll eventually need to address it globally

vishalchangrani: tbh - for many of these, I don't know what the right limits are. Will a separate issue to tackle these limits overall (grpc, web proxy and REST) help?

vishalchangrani: https://github.com/dapperlabs/flow-go/issues/6028

peterargue (approved):
💯

engine/access/rest/converter.go
peterargue: > tbh - for many of these, I don't know what the right limits are. Will a separate issue to tackle these limits overall (grpc, web proxy and REST) help?

yea, that's a good idea. we should at least get the conversation started about how to manage it

PR #1620, opened 2021-11-15
REST: models converters structure

sideninja:
Define converters for flow models and generated open API model stubs. 
Validation is included in converters from request data to flow models.


engine/access/rest/models/transactions.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"regexp"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+type transactionSignatureRequest struct {
```
vishalchangrani, line 12: isn't this the same as the [auto generated one](https://github.com/onflow/flow-go/blob/de532996003fc6623ba4b4334b950d899a759d7a/engine/access/rest/generated/model_transaction_signature.go)?

sideninja: It's similar, it doesn't have that extendable fields which are weirdly generated, but yes, we can reuse those models, but I feel the autogenerated code is pretty spammy, the only code we can use are model stubs (we can take a look if we can only limit generation to stubs so we don't include code that is client specific in this codebase). My thinking creating that by hand was just that it's not much overhead to create them and anyhow we gonna do that pretty much only once, but I'm fine either way as long as we can eliminate the code that won't be used from the codebase.

vishalchangrani (approved):

PR #1622, opened 2021-11-15
Enable instantiating DKG when joining the network in `EpochSetup` phase

jordanschalm:
This PR updates the DKG Reactor engine to start a DKG instance when booting up within the bounds of the `EpochSetup` phase. Nodes joining at an epoch boundary will boot using a state snapshot which it at the beginning of the `EpochSetup` phase. Previously the DKG would not have started for these nodes, because the trigger for starting the DKG is the transition from `EpochStaking->EpochSetup`.

## Changes
* Add `DKGState` storage to track the state of in-progress or completed DKG instances. Currently this only tracks whether a DKG has been started for a given epoch, but this could be extended to persist everything needed to recover an in-progress DKG in the future
* Refactor reactor engine tests to use the `Suite` pattern, as this PR adds a few tests re-using much of the existing set up code


engine/consensus/dkg/reactor_engine.go
```diff
@@ -79,7 +82,38 @@ func NewReactorEngine(
 // that will close when the engine has successfully
 // started.
 func (e *ReactorEngine) Ready() <-chan struct{} {
-	return e.unit.Ready()
+	return e.unit.Ready(func() {
+		// If we are starting up in the EpochSetup phase, try to start the DKG.
+		// If the DKG for this epoch has been started previously, we will exit
+		// and fail this epoch's DKG.
```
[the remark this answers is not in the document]
jordanschalm: Either of those 2 cases are possible here. 

1. We started the DKG previously and restarted before completing it. 
We don't have a way to safely recover an in-progress DKG currently, so the only safe course of action here is to exit the DKG process and consider this a locally failed DKG.

2. We started the DKG previously and completed it, then restarted.
This is an uncommon case, because there is typically very little time between local completion of the DKG and the end of the `EpochSetup` phase. But in this case we want to take the same action as Case 1 - exit the DKG process, since we already completed it.

storage/badger/operation/dkg.go
```diff
@@ -0,0 +1,52 @@
...
+// used in the context of the secrets database. This is enforced in the above
+// layer (see storage.DKGKeys).
+func RetrieveMyDKGPrivateInfo(epochCounter uint64, info *dkg.DKGParticipantPriv) func(*badger.Txn) error {
+	return retrieve(makePrefix(codeDKGPrivateInfo, epochCounter), info)
+}
+
+// InsertDKGStartedForEpoch stores a flag indicating that the DKG has been started for
+// the given epoch.
+func InsertDKGStartedForEpoch(epochCounter uint64) func(*badger.Txn) error {
+	return insert(makePrefix(codeDKGStarted, epochCounter), true)
```
zhangchiqing, line 33: wonder if it's useful to store a block ID of when we started the DKG instead of just a `bool`, since the bool value is always `true`.

when we joining the network, we will retrieve the block ID back, would it be useful to do a comparison with the snapshot's block ID and the stored block ID when DKG was started?

engine/consensus/dkg/reactor_engine.go
```diff
@@ -88,49 +122,85 @@ func (e *ReactorEngine) Done() <-chan struct{} {
...
 		Logger()
 
+	// if we have started the dkg for this epoch already, exit
+	started, err := e.dkgState.GetDKGStarted(nextEpochCounter)
+	if err != nil {
+		// unexpected storage-level error
+		log.Fatal().Err(err).Msg("could not check whether DKG is started")
+	}
+	if started {
+		log.Warn().Msg("DKG already started, skipping starting the DKG for this epoch")
```
zhangchiqing, line 161: This is the case where DKG was started in previous startup and didn't get completed before the node stopped, right? 

In that case, I changed the message a bit, because the `already` sounds like it was started by other process in the current startup.
```suggestion
		log.Warn().Msg("DKG started before, skipping starting the DKG for this epoch")
```

zhangchiqing: Another comment is that, should we check whether we have the private key for this epoch, and verify it if we have, and insert a `nil` if we don't have? Otherwise, we will always hit a not found when retrieving the private key, which would be an exception when this node is to sign a block. 

With consensus V2, we will allow the node to sign block with staking key if DKG key is not generated, but only if a `nil` record is saved in the database. [If the key is not found, it is treated as exception](https://github.com/onflow/flow-go/pull/1194/files#diff-4ba3e824602cc8ebd30e84272cef36403ca1806f7286f34a88dc2045ada665d8R91-R92). 

So when DKG was interrupted after start, the key store gets "dirty" as no record is saved to reflect the failure, and we need to figure out when to clean it up. IMO, we should clean it here. 

Thoughts?

kc1116 (commented):
Would it be useful to add a metric that will allow us to see what view the DKG started on for the current epoch ?
`consensus_compliance_dkg_started_for_epoch_on_view`

kc1116 (commented):
We should store dkgInprogress & dkgFinished (with code for success or failure) as opposed to just dkgStarted so that we may skip starting if dkg is in progress or start (essentially restart) the dkg if it finished and failed. Thoughts ?

engine/consensus/dkg/reactor_engine.go
jordanschalm: > This is the case where DKG was started in previous startup and didn't get completed before the node stopped, right?

Yep. Thank you, this message is clearer

> So when DKG was interrupted after start, the key store gets "dirty" as no record is saved to reflect the failure, and we need to figure out when to clean it up

Good point, we could do it either here or when we check our result at the end of the `EpochCommitted` phase. I agree here (earlier) is a better place to do it, at least for the time being.

storage/badger/operation/dkg.go
jordanschalm: I don't really see what we would use this for to be honest. 

> when we joining the network, we will retrieve the block ID back

When joining the network this flag wouldn't be set, because we would not have already started the DKG. If we store a block ID here, it will always be either:
* the block at which the `EpochSetupPhaseStarted` protocol event is emitted, the majority of the time
* the finalized head of the root snapshot used to bootstrap the node, when joining the network at an epoch boundary

I don't think we have a need to know which of these cases occurred.

engine/consensus/dkg/reactor_engine.go
zhangchiqing: We should do both I think. 

If we crash in setup phase, and start in committed phase, then we should also clean up.

But yeah, for now we can just add it here.

jordanschalm:
> Would it be useful to add a metric that will allow us to see what view the DKG started on for the current epoch ?

It's challenging to know this before-the-fact, because it starts when the `EpochSetup` event is emitted, which isn't known in the protocol state ahead of time.

> We should store dkgInprogress & dkgFinished (with code for success or failure) as opposed to just dkgStarted so that we may skip starting if dkg is in progress or start (essentially restart) the dkg if it finished and failed. Thoughts ?

Currently we cannot recover an in-progress DKG after a restart. We have a [future epic](https://github.com/dapperlabs/flow-go/issues/6013) to enable recovering an in-progress DKG after a restart, but it is low-priority because we get most of the practical benefits it provides from EECC and Crypto V2. When we implement this, we would need to persist more information about the current state of the DKG (in particular the current phase and latest processed broadcast message index), but I think it makes more sense to fully consider and implement those requirements when scoping that feature. 

Regarding storing `dkgFinished`, we partially do this by storing the DKG keys. We store our computed key when we complete the DKG and an empty key when we fail to. We have definitions for how to handle the various failure cases [here](https://www.notion.so/dapperlabs/DKG-Completion-Cases-0fe23d81eff54ad6b82b0df5e8687dde). The missing piece is, because Crypto V2 is not in place, we don't distinguish between a computed invalid key and a missing key (both produce invalid signatures and have the same effect currently). I'll address this in this PR along with Leo's comment https://github.com/onflow/flow-go/pull/1622#discussion_r754736230

engine/consensus/dkg/reactor_engine.go
zhangchiqing: cc @tarakby

jordanschalm:
1 - DKG fails - DB becomes dirty (because no entry for RandomBeaconKey)
2 - 

Can we crash right on edge of epoch
              Setup  Committed
| ----- | ------ | ----------- |


Compliance Engine.             Protocol State.              Reactor Engine
                                                 (1) --- event ----------> async
                                                                                        (2)         

```
// used by DKG engine
type DKGState interface {
  DKGStarted(epoch) bool
  DKGEnded(epoch) success|fail|nil
  InsertRandomBeaconKey(epoch, key)
  RetrieveRandomBeaconKey(epoch) key
}

// used by crypto v2 signer store
type SafeRandomBeaconKeys interface {
  // Returns RandomBeaconKeys.Retrieve(epoch) iff DKGState.DKGEnded(epoch) == success
  // Returns sentinel if DKGState.DKGEnded(epoch) == fail
  // Returns exception if DKGState.DKGEnded(epoch) == nil
  Retrieve(epoch) key
}
```

jordanschalm:
This PR is getting quite big, and there is still some more work to do - closing in favour of https://github.com/onflow/flow-go/pull/1720 and https://github.com/onflow/flow-go/pull/1721 (and 1 future PR implementing https://github.com/dapperlabs/flow-go/issues/6044)

PR #1624, opened 2021-11-16
bumping up the ledger interaction limit to 30K to fix EN panic

vishalchangrani:


Kay-Zee (approved):
This is a temp fix for testnet only

vishalchangrani:
this fix did not work

PR #1625, opened 2021-11-16
Recover ledger interaction error

janezpodhostnik:
Recover Ledger interaction limit panics

### DIFF
--- a/fvm/transaction.go
+++ b/fvm/transaction.go
@@ -1,7 +1,10 @@
 package fvm
 
 import (
+	"fmt"
 	"github.com/opentracing/opentracing-go"
+	"runtime/debug"
+	"strings"
 
 	"github.com/onflow/flow-go/fvm/errors"
 	"github.com/onflow/flow-go/fvm/programs"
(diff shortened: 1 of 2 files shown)


fvm/transaction.go
```diff
@@ -40,6 +43,19 @@ func (proc *TransactionProcedure) SetTraceSpan(traceSpan opentracing.Span) {
...
 func (proc *TransactionProcedure) Run(vm *VirtualMachine, ctx Context, st *state.StateHolder, programs *programs.Programs) error {
 
+	defer func() {
+		if r := recover(); r != nil {
+
+			if strings.Contains(fmt.Sprintf("%v", r), "[Error Code: 1106]") {
+				ctx.Logger.Error().Str("trace", string(debug.Stack())).Msg("VM LedgerIntractionLimitExceeded panic")
+				proc.Err = errors.NewLedgerIntractionLimitExceededError(state.DefaultMaxInteractionSize, state.DefaultMaxInteractionSize)
+				return
+			}
```
ramtinms, line 53: I think we should take the time to fix this properly, the string comparison seems a bit risky to me

janezpodhostnik:
I agree, but this is what is deployed on testnet, so we should probably merge this one and add a new PR for cleanup.

The cleanup would include unwrapping the error as ExternalError, and unwrapping the contents as `LedgerIntractionLimitExceededError`.

The proper fix will probably e made with this issue LedgerIntractionLimitExceededError.https://github.com/onflow/cadence/issues/1255.

ramtinms (approved):

PR #1627, opened 2021-11-16
REST: transactions handlers

sideninja:
Implemented get and create transaction handlers and move them to the handler file.
Safe JSON decoder was added to handlers. 
Refactored:
- Renamed `rest_api` to `handlers` and also rename the struct to `Handlers` accordingly with receiver renamed to `h`, reasons for that: prefixing file names with module prefix (`rest_`) is [not idiomatic in Go](https://golang.org/doc/effective_go#package-names), and handlers are more appropriate name since the struct holds just all the handlers for the API and API in itself seems ambiguous
- Renamed `rest_api_server` to just `server` because of all the points from above, and being just server in the rest models is quite clear that is an API rest server
- Moved middleware files to middleware module so the usage is clear from the code. This is opinionated and if you don't agree we can revert, don't have strong opinion, just seems more organised.  

My suggestion for handler names is to be natural and clear in what they do so to me `GetTransactionByID` and `CreateTransaction` are clear in what they do, I wouldn't use get/post/put methods in the names as they are not as clear in what the function does. 

Also small flag all abbreviations should be written with capital so all `Id` changed to `ID`.


vishalchangrani:
This looks all good to me! 👍

PR #1631, opened 2021-11-16
[FVM] Variable Computation Metering

janezpodhostnik:
Adds calls to MeterComputation On FVM environment methods.

MeterComputation adds the intensity times the weight for the computation kind to the total computation used. If computation used is over the computation (gas) limit, the transaction fails.

The computation kind weights can be set with the service account.

The Meter internally meters computation and memory at a finer granulation then is visible to the rest of the code. This allows for smaller weights to be set.


utils/unittest/execution_state.go
```diff
@@ -24,7 +24,7 @@ const ServiceAccountPrivateKeySignAlgo = crypto.ECDSAP256
 const ServiceAccountPrivateKeyHashAlgo = hash.SHA2_256
 
 // Pre-calculated state commitment with root account with the above private key
-const GenesisStateCommitmentHex = "abdc5bc5456726872d3b3a99f70eedafe7f198736e7f5d44c77a7c12f4b84ba5"
+const GenesisStateCommitmentHex = "9410871291ff9aa6d1a0267fc87fb20db2fb2bc575014aac8df135a42992389a"
```
janezpodhostnik, line 27: This change is due to 1 more read per transaction.

fvm/context.go
```diff
@@ -90,12 +90,14 @@ func defaultContext(logger zerolog.Logger) Context {
...
 		TransactionProcessors: []TransactionProcessor{
 			NewTransactionAccountFrozenChecker(),
+			NewTransactionSetExecutionWeights(logger),
 			NewTransactionSignatureVerifier(AccountKeyWeightThreshold),
 			NewTransactionSequenceNumberChecker(),
 			NewTransactionAccountFrozenEnabler(),
 			NewTransactionInvoker(logger),
 		},
 		ScriptProcessors: []ScriptProcessor{
+			NewScriptSetExecutionWeights(logger),
```
ramtinms, line 100: same here.

fvm/env.go
```diff
@@ -2,12 +2,15 @@ package fvm
...
+
+	"github.com/onflow/flow-go/fvm/state"
 )
 
 // Environment accepts a context and a virtual machine instance and provides
 // cadence runtime interface methods to the runtime.
 type Environment interface {
 	Context() *Context
 	VM() *VirtualMachine
+	StateHolder() *state.StateHolder
```
ramtinms, line 14: StateHolder and state are internal concepts, not sure exposing the environment is that meaningful here, if we endup needing it somewhere, probably some structures need to change.

fvm/transactionContractFunctionInvocator.go
```diff
@@ -44,6 +45,13 @@ func NewTransactionContractFunctionInvoker(
 func (i *TransactionContractFunctionInvoker) Invoke(env Environment, parentTraceSpan opentracing.Span) (cadence.Value, error) {
 	var span opentracing.Span
 
+	if env.StateHolder().EnforceComputationLimits {
+		err := env.StateHolder().State().MeterComputation(meter.ComputationKindContractFunctionInvoke, 1)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
ramtinms, line 54: I think instead of having general concept of contractFunctionInvoke, we should meter things where we actually call this Invoke method and capture meter with a more detailed computation type (e.g. Check Stroage, ...)

fvm/meter/meter.go
```diff
@@ -1,17 +1,78 @@
...
 	MeterComputation(kind uint, intensity uint) error
+	ComputationIntensities() map[uint]uint
 	TotalComputationUsed() uint
 	TotalComputationLimit() uint
 
 	// memory metering
+	SetMemoryWeights(map[uint]uint64)
 	MeterMemory(kind uint, intensity uint) error
+	MemoryIntensities() map[uint]uint
 	TotalMemoryUsed() uint
```
ramtinms, line 76: hmm, I think ideally SetMemoryWeights and MemoryIntensities should not be part of this interface, these are internal properties of a kind of metring which should consider different weights, etc that can be passed at the construction time.

fvm/meter/meter.go
```diff
@@ -1,17 +1,78 @@
...
+var (
+	// DefaultComputationWeights is the default weights for computation intensities
+	// these weighs make the computation metering the same as it was before dynamic execution fees
+	DefaultComputationWeights = map[uint]uint64{
+		uint(common.ComputationKindStatement):          1 << MeterInternalPrecisionBytes,
+		uint(common.ComputationKindLoop):               1 << MeterInternalPrecisionBytes,
+		uint(common.ComputationKindFunctionInvocation): 1 << MeterInternalPrecisionBytes,
+	}
+	// DefaultMemoryWeights are empty as memory metering is not fully defined yet
+	DefaultMemoryWeights = map[uint]uint64{}
```
ramtinms, line 57: these probably should be part of meter implementation that supports weights.

fvm/context.go
```diff
@@ -90,12 +90,14 @@ func defaultContext(logger zerolog.Logger) Context {
 		SignatureVerifier:             crypto.NewDefaultSignatureVerifier(),
 		TransactionProcessors: []TransactionProcessor{
 			NewTransactionAccountFrozenChecker(),
+			NewTransactionSetExecutionWeights(logger),
```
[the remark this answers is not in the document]
janezpodhostnik: I thought this was less complex. I can move it inside the transaction invocator.

fvm/meter/meter.go
janezpodhostnik: change in [3cab02f](https://github.com/onflow/flow-go/pull/1631/commits/3cab02f4f87121ad3117fb85dc926a741161e0c7)

fvm/transactionContractFunctionInvocator.go
janezpodhostnik: Both is actually metered.

The `meter.ComputationKindContractFunctionInvoke` is meant as the meter for the overhead the ContractFunctionInvoke has. 

The contract function invocation itself will still be using and incrementing the meter on the `env` as the `env` is passed into the invocation.

fvm/env.go
janezpodhostnik: Removed with [8cec310](https://github.com/onflow/flow-go/pull/1631/commits/8cec310c368c0e444efb933f2b8e9b576c8512d4).

utils/unittest/execution_state.go
ramtinms: 1 more read or one more right? cause I don't think reads are actually changing anything about the state commitments

fvm/state/state.go
```diff
@@ -56,6 +56,14 @@ func (s *State) View() View {
...
 }
 
+func (s *State) Meter() meter.Meter {
+	return s.meter
+}
+
+func (s *State) SetMeter(m meter.Meter) {
+	s.meter = m
+}
+
```
ramtinms, line 66: this is very risky I think we should do it using the option to NewState

fvm/transactionContractFunctionInvocator.go
```diff
@@ -53,6 +53,25 @@ func (i *TransactionContractFunctionInvoker) Invoke(env Environment, parentTrace
...
+				return nil, err
+			}
+		}
+	case *ScriptEnv:
+		if e.sth.EnforceComputationLimits {
+			err := e.sth.State().MeterComputation(meter.ComputationKindContractFunctionInvoke, 1)
+			if err != nil {
+				return nil, err
+			}
+		}
```
ramtinms, line 70: this should be only transactionContractFunctionInvocator so it shouldn't be any need to disambiguate here, if there is some call from scriptsEnv to this, there should be a copy of this with script prefix.

fvm/meter/basic/meter.go
```diff
@@ -17,21 +17,28 @@ type Meter struct {
 	computationLimit uint
 	memoryUsed       uint
 	memoryLimit      uint
+
+	computationIntensities map[uint]uint
+	memoryIntensities      map[uint]uint
```
ramtinms, line 22: now that we have default weights we might not need these changes to the basic meter and return empty.

ramtinms (approved):
Looks good to me, thanks for the refactors

fvm/meter/meter.go
```diff
@@ -1,17 +1,57 @@
 package meter
 
+type MetringOperationType uint
+
+// continuation of cadence.ComputationKind
+const (
+	// [2_000, 3_000) reserved for the FVM
+	_ = iota + 2_000
```
SupunS, line 8: Should this be typed as `ComputationKind`? e.g:
```suggestion
	_ cadence.ComputationKind = iota + 2_000
```

And then use the key-type of the intensities maps as also `cadence.ComputationKind`? e.g:
```go
type Meter struct {
    ...
    computationIntensities map[cadence.ComputationKind]uint
    ...
}
```

fvm/meter/meter.go
```diff
@@ -1,17 +1,57 @@
...
 
 	// computation metering
 	MeterComputation(kind uint, intensity uint) error
+	ComputationIntensities() map[uint]uint
 	TotalComputationUsed() uint
 	TotalComputationLimit() uint
 
 	// memory metering
 	MeterMemory(kind uint, intensity uint) error
+	MemoryIntensities() map[uint]uint
```
SupunS, line 54: just a suggestion: maybe it's worth defining a type-alias for `map[uint]uint` (rather for `map[cadence.ComputationKind]uint`, when combined with my previous suggestion). seems it is being used in a lot of places

fvm/meter/weighted/meter.go
```diff
@@ -0,0 +1,193 @@
...
+// for any each memory/computation usage call it sums intensity multiplied by the weight of the intensity to the total
+// memory/computation usage metrics and returns error if limits are not met.
+type Meter struct {
+	computationUsed  uint64
+	computationLimit uint64
+	memoryUsed       uint64
+	memoryLimit      uint64
+
+	computationIntensities map[uint]uint
+	memoryIntensities      map[uint]uint
```
SupunS, line 39: are these fields supposed to be a 1:1 mapping from `basic.Meter`? If so, maybe possible to include `basic.Meter` type as a type-inclusion/field?

But having said that, if there are no common logic/code/methods to re-use, then the current approach might be better.

fvm/scriptEnv.go
```diff
@@ -89,9 +91,33 @@ func NewScriptEnvironment(
 		env.seedRNG(ctx.BlockHeader)
 	}
 
+	env.setMeteringWeights()
```
SupunS, line 94: Does this get called for every script/tx?

SupunS (approved):
LGTM! I mostly reviewed the general Go-code.

fvm/scriptEnv.go
janezpodhostnik: in general once per script/transaction yes.

When it comes to transactions, the value being read from storage will already be cached the second time it is needed in a collection.

fvm/meter/meter.go
janezpodhostnik: Nifty! Yes, this is definitely nicer.

fvm/meter/weighted/meter.go
janezpodhostnik: the basic one is intended for testing purposes, so I think its ok to keep them separate for now.

fvm/scriptEnv.go
SupunS: I see! Thanks for the explanation. My concern was on having to encode/decode the entire map every time for each tx. Not a problem for this PR, I'm just thinking out loud whether it's possible to optimize that sometime in future 👌

janezpodhostnik:
bors merge please

PR #1633, opened 2021-11-17
Fix for Mtrie bug

AlexHentschel:
* fixes MTrie bug
* add test 
* cleaned up code so that nodes are truly immutable 

Comment regarding compactification:
* We are planning to optimize the pruning algorithm shortly (in a subsequent PR). Therefore, I have not optimized method `node.Compactify` as this code will anyway change with the optimization.


ledger/complete/mtrie/node/node.go
```diff
@@ -175,24 +169,23 @@ func (n *Node) Compactify() *Node {
...
-	// if childNode is non empty
+	// if childNode is non-empty
 	if rChildEmpty {
-		if !lChildEmpty && n.lChild.IsLeaf() {
+		if n.lChild.IsLeaf() {
 			return n.lChild.copyAndPromoteLeafNode(false)
 		}
-		// this would replace empty nodes with nil
-		n.rChild = nil
+		return NewInterimNode(n.height, n.lChild, nil)
```
AlexHentschel, line 182: This change enforces immutability of `Node`, thereby increasing the safety of the code base at the expense of marginally more GC load. We will optimize the pruning to completely avoid unnecessary garbage generation in a subsequent PR.

ledger/complete/mtrie/trie/trie_test.go
```diff
@@ -300,6 +334,69 @@ func TestSplitByPath(t *testing.T) {
...
+	require.Equal(t, expectedRootHashHex, hashToString(startTrie.RootHash()))
+
+	// Register update:
+	//  * de-allocate the compactified leaf (■), i.e. set its payload to nil.
+	//  * also set a previously already unallocated register to nil as well
+	unallocatedRegister := leftSubTriePath[0]            // copy path to leaf and modify it (next line)
+	unallocatedRegister[len(unallocatedRegister)-1] += 1 // path differs only in the last byte, i.e. register is also in the left Sub-Trie
+	updatedPaths := append(leftSubTriePath, unallocatedRegister)
+	updatedPayloads := []ledger.Payload{*ledger.EmptyPayload(), *ledger.EmptyPayload()}
+	updatedTrie, err := trie.NewTrieWithUpdatedRegisters(startTrie, updatedPaths, updatedPayloads, true)
```
ramtinms, line 391: maybe add err checks here to fix linter

ledger/complete/mtrie/trie/trie.go
```diff
@@ -294,17 +294,21 @@ func update(
...
-	// payload is re-written at a register.
-	if lChild == lchildParent && rChild == rchildParent {
-		return parentNode
+	if !parentNode.IsLeaf() {
+		// mitigate storage exhaustion attack: avoids creating a new node when the exact same
+		// payload is re-written at a register.
+		if lChild == lchildParent && rChild == rchildParent {
+			return parentNode
+		}
 	}
```
ramtinms, line 303: maybe merge them all into one condition statement

ramtinms (approved):
Nice work

ledger/complete/mtrie/node/node.go
tarakby: I totally agree this change enforces code safety and the code will change again _soon_, but I still think avoiding the extra GC cost is a simple change (return the same node, and make `Compactify` private) compared to the gain (I think the number of allocations from pruning is divided by 2). 
We would benefit from this gain till the new pruning logic is implemented.

tarakby (approved):
Very nice work on spotting the bug 👏🏼  Thank you for writing the detailed test 🏄🏼 

I left minor comments below.

ledger/complete/mtrie/node/node.go
AlexHentschel: Its more than just a few lines, because `Compactify` is called by the Trie, which is a different package.  We can't just make `Compactify` private. Therefore, we need to introduce a new constructor for Nodes with compactification:
* I added the new constructor `NewInterimCompactifiedNode `  (-> [code](https://github.com/onflow/flow-go/blob/3afc1e5ded7bb158df49de01be852b29d9891e00/ledger/complete/mtrie/node/node.go#L121-L158)), and conslidated the code that was previously distributed over the `Compactify` and the `copyAndPromoteLeafNode`  methods. 
* The tests also needed refactoring, because we now construct the compactified node in one step and don't mutate a node anymore. 

To me, the code feels more concise now and is probably more performant.

ramtinms (approved):
Looks good to me, but since the latest changes includes refactoring of tests, maybe just run the trie pruning smoke test a few times without cache to make sure we are okey (by default that test is skipped)

AlexHentschel:
> maybe just run the trie pruning smoke test a few times

ran the [smoke test](https://github.com/onflow/flow-go/blob/3afc1e5ded7bb158df49de01be852b29d9891e00/ledger/complete/mtrie/trie/trie_test.go#L507) 10 times, 100% success rate 🎉

ledger/complete/mtrie/node/node_test.go
```diff
@@ -136,125 +136,195 @@ func Test_VerifyCachedHash(t *testing.T) {
...
+
+// requireIsLeafWithHash verifies that `node` is a leaf node, whose hash equals `expectedHash`.
+// We perform the following checks:
+// * both children must be nil
+// * depth is zero
+// * number of registers in the sub-trie is 1
+// * pre-computed hash matches the `expectedHash`
+// * re-computing the hash from the children yields the pre-computed value
+// * node reports itself as a leaf
+func requireIsLeafWithHash(t *testing.T, node *node.Node, expectedHash hash.Hash) {
```
tarakby, line 322: 👌🏼

tarakby (approved):
Sorry for pushing for a "small" change that turned out to be a broader one. I think the update has made the code more readable. Thanks for refactoring the node tests as well 👌🏼

zhangchiqing (approved):
Looks good.

After the fix/refactor, I found the logic is much easier to follow than last time I looked at it. Great work!

ledger/complete/mtrie/trie/trie_test.go
```diff
@@ -300,6 +334,71 @@ func TestSplitByPath(t *testing.T) {
...
+	//          / \
+	//        /    \
+	//       /      \
+	//      ■        o
+	//    Left      / \
+	//  SubTrie     ⋮  ⋮
+	//             Right
+	//            SubTrie
+	// Left Sub-Trie (■) is a single compactified leaf
+	// Right Sub-Trie contains multiple (18) allocated registers
```
turbolent, line 366: 😍 👏

tarakby: Agree, very clean 👏🏼

PR #1636, opened 2021-11-17
REST: custom handlers

sideninja:
A custom handler was implemented which defines a custom handler function that allows us to return errors and responses without worrying about how to handle HTTP responses. It also gives us access to variables, logger and backend through arguments. 
Handler implements the` ServerHTTP` function which complies with the handler interface and that function wraps and handles everything that is common to all the API handlers.

The previous implementation of handlers was removed and refactored out with all already implemented handlers refactored to this structure but in separate files `transactions.go` and `blocks.go`. 

A custom rest errors were also implemented which are then handled by the handler wrapper.

Request data validation is implemented as part of converters.


engine/access/rest/handler.go
```diff
@@ -81,6 +85,17 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
 
+// addToRouter adds handler to provided router
+func (h *Handler) addToRouter(router *mux.Router) {
```
vishalchangrani, line 89: is there a reason you wanna do this? - pass a router to the handler, and then it to add itself to the router. Instead, what we have currently where the server adds the handler to the router instead https://github.com/onflow/flow-go/blob/rest_api_feature_branch/engine/access/rest/server.go#L52-#L56

sideninja: That was done a couple of days ago because of the need for link generator to access routes, which was solved now by having route names exposed as global consts, I feel handler should have access to the route itself, there might be lots of use cases where that will be handy, even for link generator, you could just instantiate it now in handler and pass the route - that way you don't need to pass another argument to the handler, you don't need to expose route names across files as the route already is there etc, but both approaches work.

vishalchangrani: so there are many instances where a handle needs not it's own route but a different route e.g. in the GetBlock request, the block handler needs the Execution result route to get its name. Hence, I thought the best way would be to just completely delegate all of that to the link generator and give it the full route object.

PR #1637, opened 2021-11-17
Add jitter to all QC/DKG requests

jordanschalm:
* Add jitter to backoff on all outstanding QC/DKG requests
* Use capped duration rather than maximum duration backoff option (this was a configuration error, the intention was to cap individual wait intervals, not to terminate the retry)
* Add some documentation and make locking more clear in DKG Broker


module/dkg/broker.go
```diff
@@ -205,18 +216,23 @@ func (b *Broker) GetBroadcastMsgCh() <-chan messages.DKGMessage {
...
 func (b *Broker) Poll(referenceBlock flow.Identifier) error {
-	b.Lock()
-	defer b.Unlock()
 
-	expRetry, err := retry.NewExponential(retryMilliseconds)
+	// We only issue one poll at a time to avoid delivering duplicate broadcast messages.
+	// The messageOffset determines which messages are retrieved by a Poll,
+	// and is not updated until the end of this function.
+	b.pollLock.Lock()
+	defer b.pollLock.Unlock()
```
zhangchiqing, line 224: I think it's better to use an atomic bool here. [See example](https://github.com/onflow/flow-go/blob/jordan/5806-qc-dkg-jitter/module/jobqueue/consumer.go#L155).

The difference is that when using lock, if there are 5 threads blocked waiting to unlock and once unblocked, it will poll again immediately one after another. Instead, we actually want the 5 threads to give up if there is a thread reading from the contract already, which can be done with atomic.Bool

jordanschalm: That's a good idea but I don't think we can always allow the other threads waiting on the lock to exit. 

Each call to `Poll` is linked to a particular block (`referenceBlock` argument). It's important for correctness that we process all messages exactly up to a particular block before starting the next DKG phase. The component issuing the poll requests (Reactor engine) ensures that a DKG phase transition doesn't occur until the final poll request for that phase has returned.

```
                             v- DKG Phase Transition
Polls:      A      B      C  |
Blocks: 1 2 3 4 5 6 7 8 9 10 |
Messages: Y             Z    |
```

For example suppose A, B, C, D are poll request threads waiting on `pollLock` to issue a poll request. If A is currently holding the lock and polling, we can't discard B, C, _and_ D, because otherwise any messages between block 4 and 10 (in this case `Z`) would be processed as messages during the next DKG phase (or in the case of the final DKG phase, wouldn't be processed at all.  

Ideally we would discard B and C, and keep D, but that would be a change in the Reactor engine.

zhangchiqing (approved):

jordanschalm:
Issues from some observations from @zhangchiqing:
* https://github.com/dapperlabs/flow-go/issues/6018
* https://github.com/dapperlabs/flow-go/issues/6019
* https://github.com/dapperlabs/flow-go/issues/6020

module/dkg/broker.go
```diff
@@ -241,30 +257,36 @@ func (b *Broker) Poll(referenceBlock flow.Identifier) error {
 	for _, msg := range msgs {
 		ok, err := b.verifyBroadcastMessage(msg)
 		if err != nil {
+			// CAUTION: potential slashable offense
```
tarakby, line 260: This function seems to only authenticate the message. Failing to authenticate the message can't be slashable.

module/dkg/broker.go
```diff
@@ -241,30 +257,36 @@ func (b *Broker) Poll(referenceBlock flow.Identifier) error {
...
 			b.log.Error().Err(err).Msg("bad broadcast message")
 			continue
 		}
 		if !ok {
-			b.log.Debug().Msg("invalid signature on broadcast dkg message")
+			// CAUTION: potential slashable offense
```
tarakby, line 265: It's same here, if we fail to authenticate, we can't attribute the offence.

tarakby (approved):
👌🏼

module/dkg/broker.go
jordanschalm: Good point

jordanschalm:
bors merge

jordanschalm:
bors merge

PR #1645, opened 2021-11-19
Vishal/gregor/rest api errors

vishalchangrani:


engine/access/rest/request_decorator.go
```diff
@@ -0,0 +1,51 @@
...
+	*http.Request
+	expandFields map[string]bool
+	selectFields map[string]bool
+}
+
+func newRequestDecorator(r *http.Request) *requestDecorator {
+	decoratedReq := &requestDecorator{
+		Request: r,
+	}
+	decoratedReq.expandFields, _ = middleware.GetFieldsToExpand(r)
```
sideninja, line 24: Are these even middlewares anymore? Do you think it would make sense for those functions to be moved in this decorator?

engine/access/rest/request_decorator.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (rd *requestDecorator) expands(field string) bool {
+	return rd.expandFields[field]
+}
+
+func (rd *requestDecorator) selects(field string) bool {
+	return rd.selectFields[field]
+}
+
+func (rd *requestDecorator) ids() ([]flow.Identifier, error) {
```
sideninja, line 37: I like this :)

PR #1646, opened 2021-11-19
Transaction model status bugfix

sideninja:
Flow transaction status model had a bug in string serialization with missing values in the string method. Values were added and tested accordingly.


peterargue (approved):
Great find! thanks for the fix

ramtinms (approved):
Thanks for catching and fixing this.

PR #1648, opened 2021-11-19
[Execution] update ledger interaction limit for bootstrapping

ramtinms:
Bootstrapping needs to be run with a larger ledger interaction limit.


janezpodhostnik (approved):

engine/execution/state/bootstrap/bootstrap.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/onflow/flow-go/storage/badger/operation"
 )
 
+const ledgerIntractionLimitNeeded = 1_000_000_000
```
zhangchiqing, line 20:
```suggestion
// an increased limit for bootstrapping
const ledgerIntractionLimitNeededForBootstrapping = 1_000_000_000
```

zhangchiqing (approved):

PR #1651, opened 2021-11-20
Metrics: fix randomness in execution's example

SaveTheRbtz:
I would suspect that `^` here ment exponentiation.  No functional changes except for the magnitude of random values.


Kay-Zee (approved):
Ironically, it still tested the right things, but at a different magnitude that was expected by the original test writer it seems.

zhangchiqing (approved):

Kay-Zee:
bors merge

PR #1652, opened 2021-11-21
JSON: do not use pointer receiver for MarshalJSON

SaveTheRbtz:
Using a pointer receiver for `MarshalJSON` is error-prone since custom marshal logic would only be used if addressable variable is passed, otherwise the default implementation is used.

Since `Payload`'s `MarshalJSON` relied on tis undefined behavior it was re-written.  While here, expanded tests to cover nil, empty, and non-empty payloads.

See: https://github.com/golang/go/issues/22967


ramtinms (approved):
@SaveTheRbtz nice catch and thanks for cleaning up the code. 🙏  This is really valuable  since most of the MarshalJSON's are used for debugging or 3rd party integrations and might have been future problems without we noticing it in our unit and integration tests

model/flow/payload.go
```diff
@@ -17,25 +17,26 @@ func EmptyPayload() Payload {
...
-		dup.Results = nil
+	if len(p.Results) == 0 {
+		p.Results = nil
 	}
 
-	return json.Marshal(dup)
+	type payloadAlias Payload
+	return json.Marshal(struct{ payloadAlias }{
+		payloadAlias: payloadAlias(p),
+	})
```
zhangchiqing, line 39: I like the refactor. However, it's better to ensure the refactor didn't change the json encoding. 

In order to ensure that, we could add a test case that takes the json output from master and compare with the json output from the refactored method. The json string or bytes should match after the refactor.

zhangchiqing (commented):
Thanks for the refactor, I'd like to request adding a test case.

AlexHentschel (commented):
Looks good to me. 
👍 on adding a test case as [suggested](https://github.com/onflow/flow-go/pull/1652#discussion_r756588191) by @zhangchiqing

SaveTheRbtz:
I've manually verified the equality before and after.  I can add a following test:
```
assert.Equal(t, []byte(`{"Guarantees":null,"Seals":null,"Receipts":null,"Results":null}`), encoded2)
```

The only caveat here is that it will be quite painful to maintain, since every time the struct will change this code will need to be changed.  WDYT, is this worth it?

zhangchiqing:
Thanks for verifying. 

Yes, please add that test case. I don't think it would be that painful to maintain, after all it's human readable json :)

zhangchiqing (approved):
Nice work! Thanks for the refactor 👍

PR #1655, opened 2021-11-22
[Consensus Voting] Update `VoteAggregatorV2` to implement `ComponentManager`

durkmurder:
### Context

Short PR as part of replacing modules saga. `VoteAggregatorV2` was implemented using old design pattern for startable and ready done aware module. I've updated it to use `ComponentManager`. 

Updated both impl and tests.


zhangchiqing (approved):

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
```diff
@@ -63,48 +65,43 @@ func NewVoteAggregatorV2(
...
+	// launch as many workers as we need
+	for i := 0; i < defaultVoteAggregatorWorkers; i++ {
+		componentBuilder.AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+			ready()
+			err := aggregator.queuedVotesProcessingLoop(ctx)
+			if err != nil {
+				ctx.Throw(err)
+			}
+		})
+	}
```
AlexHentschel, line 79: I think we could merge the abstract function `func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) { ... }` into `queuedVotesProcessingLoop`, which essentially is the worker function. Why not make it conform to [`ComponentWorker`](https://github.com/onflow/flow-go/blob/50f0ba0e46898c6dacab9738d2fd946b07f3c99d/module/component/component.go#L152) (?)

```suggestion
	// initialize ComponentManager with workers to process votes
	componentBuilder := component.NewComponentManagerBuilder()
	for i := 0; i < defaultVoteAggregatorWorkers; i++ {
		componentBuilder.AddWorker(aggregator.queuedVotesProcessingLoop)
	}
	aggregator.ComponentManager = componentBuilder.Build()
```

where the method `queuedVotesProcessingLoop` could look something like
```golang
// queuedVotesProcessingLoop implements `component.ComponentWorker`.
// The worker will signal its readiness, and commence processing votes
// immediately when `ready` returns. The worker terminates when ctx.Done() returns
// or when it encounters an unexpected exception while processing votes.
func (va *VoteAggregatorV2) queuedVotesProcessingLoop(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
	notifier := va.queuedVotesNotifier.Channel()
	ready() // signal that this worker is ready
	for {
		select {
		case <-ctx.Done():
			return
		case <-notifier:
			err := va.processQueuedVoteEvents(ctx)
			if err != nil {
				ctx.Throw(
					fmt.Errorf("internal error processing queued vote events: %w", err),
				)
				return
			}
		}
	}
}
```

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -67,11 +71,17 @@ func (s *VoteAggregatorV2TestSuite) SetupTest() {
 	s.aggregator, err = NewVoteAggregatorV2(unittest.Logger(), s.notifier, 0, s.collectors)
 	require.NoError(s.T(), err)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	signalerCtx, _ := irrecoverable.WithSignaler(ctx)
+
 	// startup aggregator
+	s.aggregator.Start(signalerCtx)
 	<-s.aggregator.Ready()
```
AlexHentschel, line 80: In the mindeset of "fail early": 
* this could have a time-out, to fail the test, if component startup hangs?

consensus/hotstuff/voteaggregator/vote_aggregator_v2_test.go
```diff
@@ -67,11 +71,17 @@ func (s *VoteAggregatorV2TestSuite) SetupTest() {
...
 }
 
 func (s *VoteAggregatorV2TestSuite) TearDownTest() {
+	s.cancel()
 	<-s.aggregator.Done()
```
AlexHentschel, line 85: same here: timeout to fail the test?

AlexHentschel (approved):
👍 LGTM

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
```diff
@@ -63,48 +65,43 @@ func NewVoteAggregatorV2(
...
 }
 
-func (va *VoteAggregatorV2) queuedVotesProcessingLoop() {
+func (va *VoteAggregatorV2) queuedVotesProcessingLoop(ctx context.Context) error {
 	notifier := va.queuedVotesNotifier.Channel()
 	for {
 		select {
-		case <-va.unit.Quit():
-			return
+		case <-ctx.Done():
```
AlexHentschel, line 90: ## ⚠️  Caution with substituting an `irrecoverable.SignalerContext` as `context.Context` 

you **cannot use `SignalerContext.Done()`** as a trigger when to terminate the business logic:
* The `SignalerContext` has the method `ShutdownSignal` to retrieve the shutdown signal. In contrast, the channel from `SignalerContext.Done()` is only closed after _all workers_ terminated. 
* If each individual worker keeps going until all worker terminated (i.e. using `ctx.Done()`), then we can't shutdown the workers.

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
```diff
@@ -63,48 +65,43 @@ func NewVoteAggregatorV2(
...
 		}
 	}
 }
 
-func (va *VoteAggregatorV2) processQueuedVoteEvents() error {
+func (va *VoteAggregatorV2) processQueuedVoteEvents(ctx context.Context) error {
 	for {
 		select {
-		case <-va.unit.Quit():
+		case <-ctx.Done():
```
AlexHentschel, line 104: ### :warning: 
same here: we need to listen to  `SignalerContext.ShutdownSignal()`


```golang
func (va *VoteAggregatorV2) processQueuedVoteEvents(ctx irrecoverable.SignalerContext) error {
	for {
		select {
		case <-ctx.ShutdownSignal():
```

consensus/hotstuff/voteaggregator/vote_aggregator_v2.go
durkmurder: My main objection would be to make this function to return error instead of handle it, but it makes initialization code cleaner so let's give it a try.

durkmurder:
bors merge

PR #1656, opened 2021-11-22
[Consensus Voting] Update `EventLoopV2` to implement `ComponentManager`

durkmurder:
### Context

Short PR as part of replacing modules saga. `EventLoopV2` was implemented using old design pattern for startable and ready done aware module. I've updated it to use `ComponentManager`. Also our implementation was missing support for delayed start. I've added it as well.

Updated both impl and tests.


consensus/hotstuff/eventloop/event_loop_v2.go
```diff
@@ -1,53 +1,68 @@
...
 		quorumCertificates: quorumCertificates,
-		unit:               engine.NewUnit(),
+		startTime:          startTime,
 	}
 
+	componentBuilder := component.NewComponentManagerBuilder()
+	componentBuilder.AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+		ready()
+
+		// launch when scheduled by el.startTime
```
zhangchiqing, line 51: better to add some info logs saying, `event loop will start at XXX` and `event loop has started`

zhangchiqing (approved):

consensus/hotstuff/eventloop/event_loop_v2.go
```diff
@@ -61,7 +76,7 @@ func (el *EventLoopV2) loop() {
 	// if hotstuff hits any unknown error, it will exit the loop
 
 	for {
-		quitted := el.unit.Quit()
+		quitted := ctx.Done()
```
AlexHentschel, line 79: "quitted"  sounds a bit funny to me. Maybe we could call it 
```suggestion
		shutdownSignaled := ctx.Done()
```
Furthermore, I believe we should move this out of the loop for performance reasons.

AlexHentschel (approved):
LGTM 👍

consensus/hotstuff/eventloop/event_loop_v2.go
```diff
@@ -1,53 +1,68 @@
...
+		}
+
+	})
+	el.ComponentManager = componentBuilder.Build()
+
 	return el, nil
 }
 
-func (el *EventLoopV2) loop() {
+func (el *EventLoopV2) loop(ctx context.Context) {
```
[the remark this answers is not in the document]
durkmurder: hm, I've propagated error to caller and uses `Throw` once, seems cleaner than writing `Throw` multiple times.

durkmurder:
bors merge

PR #1658, opened 2021-11-22
[Consensus Voting] `EventLoopV2` -> `EventLoop`

durkmurder:
### Context

This PR replaces event loop with new implementation and breaks lots of tests :smile: 
I have tried to keep the changes very minimal, there is some room for further changes, like consolidating `module.Hotstuff` and `module.HotstuffFollower` but not for this PR.

Notable changes:

- Event loop is broken since it doesn't have `EventHandlerV2` which is pending review at this point: https://github.com/onflow/flow-go/pull/1168
- Compliance engine is broken as well since it doesn't push votes, it will depend on vote aggregator which will be replaced as part of other PR. 
- Those are two primary reasons why many tests are disabled since they depend on compliance engine and event handler.

### What was done:

- Deleted old `EventLoop` and replaced with `EventLoopV2`
- Skipped unit tests that depend on event handler
- Skipped all integration tests since compliance engine is broken and almost every test creates a consensus node


integration/Makefile
```diff
@@ -11,8 +11,11 @@ test:
 integration-test: access-tests common-tests execution-tests verification-tests collection-tests consensus-tests epoch-tests
 
 # NOTE: Currently skipping collection and consensus tests due to them not passing properly on CI
+# NOTE: Currently skipping all integration tests while replacing modules for consensus V2
 .PHONY: ci-integration-test
-ci-integration-test: access-tests common-tests execution-tests epoch-tests verification-tests # collection-tests # consensus-tests
+#ci-integration-test: access-tests common-tests execution-tests epoch-tests verification-tests # collection-tests # consensus-tests
+ci-integration-test:
+	exit 0
```
jordanschalm, line 18: Why do we need to disable the tests, given that this isn't being merged to a branch requiring CI, and by the time it is merged to master I assume we will have turned these back on anyway.

engine/collection/compliance/engine.go
```diff
@@ -45,6 +47,7 @@ type Engine struct {
 	pendingVotes   engine.MessageStore
 	messageHandler *engine.MessageHandler
 	con            network.Conduit
+	stopHotstuff   context.CancelFunc
```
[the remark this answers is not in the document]
durkmurder: oops, that's right, didn't do this for collection engine

integration/Makefile
durkmurder: that's eventually merged in `feature/consensus-voting`, I've created it this way to make the diff smaller.

jordanschalm (approved):

consensus/hotstuff/event_loop.go
```diff
@@ -1,216 +1,13 @@
...
-
-// Done implements interface module.ReadyDoneAware
-func (el *EventLoop) Done() <-chan struct{} {
-	el.lm.OnStop(func() {
-		// wait for event loop to exit
-		<-el.unit.Done()
-	})
-	return el.lm.Stopped()
+	// SubmitTrustedQC accepts QC for processing. QC will be dispatched on worker thread.
+	// CAUTION: QC is trusted (_not_ validated again)
```
zhangchiqing, line 11:
```suggestion
	// CAUTION: QC is trusted (_not_ validated again), as it's built by ourselves.
```

engine/consensus/compliance/engine.go
```diff
@@ -179,6 +188,7 @@ func (e *Engine) Ready() <-chan struct{} {
 func (e *Engine) Done() <-chan struct{} {
 	e.lm.OnStop(func() {
 		e.log.Debug().Msg("shutting down hotstuff eventloop")
+		e.stopHotstuff()
 		<-e.core.hotstuff.Done()
 		e.log.Debug().Msg("all components have been shut down")
```
zhangchiqing, line 193:
```suggestion
		e.log.Info().Msg("shutting down hotstuff eventloop")
		e.stopHotstuff()
		<-e.core.hotstuff.Done()
		e.log.Info().Msg("all components have been shut down")
```

engine/collection/compliance/engine.go
```diff
@@ -206,6 +214,7 @@ func (e *Engine) Ready() <-chan struct{} {
 func (e *Engine) Done() <-chan struct{} {
 	e.lm.OnStop(func() {
 		e.log.Debug().Msg("shutting down hotstuff eventloop")
+		e.stopHotstuff()
 		<-e.core.hotstuff.Done()
 		e.log.Debug().Msg("all components have been shut down")
```
zhangchiqing, line 219:
```suggestion
		e.log.Info().Msg("shutting down hotstuff eventloop")
		e.stopHotstuff()
		<-e.core.hotstuff.Done()
		e.log.Info().Msg("all components have been shut down")
```

consensus/hotstuff/eventloop/event_loop.go
```diff
@@ -26,16 +26,16 @@ type EventLoopV2 struct {
 	startTime          time.Time
 }
 
-var _ hotstuff.EventLoopV2 = &EventLoopV2{}
-var _ module.ReadyDoneAware = &EventLoopV2{}
-var _ module.Startable = &EventLoopV2{}
+var _ hotstuff.EventLoop = &EventLoop{}
+var _ module.ReadyDoneAware = &EventLoop{}
+var _ module.Startable = &EventLoop{}
```
zhangchiqing, line 31:
```suggestion
var _ hotstuff.EventLoop = (*EventLoop)(nil)
var _ module.ReadyDoneAware = (*EventLoop)(nil)
var _ module.Startable = (*EventLoop)(nil)
```

zhangchiqing (approved):
Looks good

consensus/hotstuff/event_loop.go
```diff
@@ -1,216 +1,13 @@
 package hotstuff
 
-import (
-	"time"
+import "github.com/onflow/flow-go/model/flow"
 
-	"github.com/rs/zerolog"
+// EventLoop performs buffer and processing of incoming proposals and QCs.
+type EventLoop interface {
```
AlexHentschel, line 6: I am wondering why we need this interface. In the following code, you have replaced the `EventLoop` with `module.HotSuff`, which seems right to me ... 
Maybe the difference is that `EventLoop` has the additional method `SubmitTrustedQC` which HotStuff doesn't have? If that is the case, we could write

```golang
type EventLoop interface {
	module.HotStuff
	
	// SubmitTrustedQC accepts QC for processing. QC will be dispatched on worker thread.
	// CAUTION: QC is trusted (_not_ validated again)
	SubmitTrustedQC(qc *flow.QuorumCertificate)
}
```

consensus/participant.go
```diff
@@ -41,7 +42,7 @@ func NewParticipant(
 	finalized *flow.Header,
 	pending []*flow.Header,
 	options ...Option,
-) (*hotstuff.EventLoop, error) {
+) (module.HotStuff, error) {
```
AlexHentschel, line 45: this is one of those cases, where I thought: "Why do we need the `hotstuff.EventLoop` interfaces, when we are essentially not using it and only have a single implementation?"

engine/collection/compliance/engine.go
```diff
@@ -195,6 +198,11 @@ func (e *Engine) Ready() <-chan struct{} {
 	}
 	e.lm.OnStart(func() {
 		e.unit.Launch(e.loop)
```
AlexHentschel, line 200: Would be great to also update the Collectors' Compliance Engine to the new `ComponentManager` ... but lets do that in a separate PR

engine/consensus/compliance/engine.go
```diff
@@ -168,7 +171,13 @@ func (e *Engine) Ready() <-chan struct{} {
 	}
 	e.lm.OnStart(func() {
 		e.unit.Launch(e.loop)
```
AlexHentschel, line 173: refactoring the code to comply with the new convention for `ReadyDoneAware` would be great; probably as a separate PR; using `ComponentManager` would be great

AlexHentschel (approved):
Thanks for splitting up the code. That really helped clarity. 
LGTM.

consensus/hotstuff/event_loop.go
durkmurder: That's one of the reasons. Originally I have added interface to make it uniform with other components. Agree, let's do it this way.

engine/collection/compliance/engine.go
durkmurder: I wanted to do that but:
1) It increases scope of PR a lot
2) We depend on some unit functionality which is not present in component manager.

PR #1667, opened 2021-11-23
making getBlockByID and getBlockByHeight a little DRY

vishalchangrani:


engine/access/rest/blocks.go
```diff
@@ -194,31 +110,77 @@ func getBlockPayloadByID(req *requestDecorator, backend access.API, _ LinkGenera
...
+		id = blk.ID()
+	} else {
+
+		// else only lookup header and add expandable link for payload
+		header, statusErr := blkProvider.getHeader(req.Context())
+		if statusErr != nil {
+			return nil, statusErr
+		}
+		responseBlock.Header = blockHeaderResponse(header)
+		responseBlock.Payload = nil
```
sideninja, line 143: assigning `nil` is not necessary, this is the default value.

engine/access/rest/blocks.go
```diff
@@ -194,31 +110,77 @@ func getBlockPayloadByID(req *requestDecorator, backend access.API, _ LinkGenera
...
+		id = header.ID()
+
+		payload, err := link.PayloadLink(id)
```
sideninja, line 146: I feel like this links can be moved to responses like here: https://github.com/onflow/flow-go/blob/cd5c42a7045ee8359de8e4db8fa6a3b42bc8db4b/engine/access/rest/converter.go#L326-L338 
Reasons for that is that anywhere that response is built the same functionality applies and we also have all the data, we can check if we have for example execution result and if we do then do nothing if we don't then generate the link. We can also just ignore this for now and then in the end think about how to maybe refactor this a bit, but to me this link generation doesn't yet feels right, because this handlers should really just be minimal an focus on getting resources, the links and how that is shaped has nothing to do with how to get resources. You can treat this for now just as an observation or a note and I would like to hear your opinion about it.

sideninja (approved):

engine/access/rest/blocks.go
vishalchangrani: yeah i did that just so that the reader knows that we sent a nil payload

PR #1670, opened 2021-11-23
unskipping the TestGhostNodeExample tests

vishalchangrani:


integration/tests/common/ghost_node_example_test.go
```diff
@@ -22,7 +22,6 @@ import (
 // TestGhostNodeExample_Subscribe demonstrates how to emulate a node and receive all inbound events for it
 func TestGhostNodeExample_Subscribe(t *testing.T) {
 
-	t.Skip()
```
vishalchangrani, line 25: testing if the unskipping passes CI.

gomisha (approved):

vishalchangrani:
1. The integration test fails so the test is broken.
2. The test does two thing - 1. demonstrate how to use a ghost node and 2. Make sure the ghost node works as expected.
For point 2, I feel the test should be continuously run as a regular test.
cc: @smnzhu

integration/tests/common/ghost_node_example_test.go
```diff
@@ -41,9 +39,11 @@ func TestGhostNodeExample_Subscribe(t *testing.T) {
...
 		verNode = testnet.NewNodeConfig(flow.RoleVerification, testnet.WithLogLevel(zerolog.FatalLevel))
+
+		accessNode = testnet.NewNodeConfig(flow.RoleAccess, testnet.WithLogLevel(zerolog.FatalLevel))
 	)
 
-	nodes := append([]testnet.NodeConfig{collNode, conNode1, conNode2, conNode3, realExeNode, verNode, ghostExeNode})
+	nodes := append([]testnet.NodeConfig{collNode, conNode1, conNode2, conNode3, realExeNode, verNode, ghostExeNode, accessNode})
 	conf := testnet.NewNetworkConfig("ghost_example_subscribe", nodes)
 
 	net := testnet.PrepareFlowNetwork(t, conf)
```
vishalchangrani, line 49: This test needs to call `testnet.PrepareFlowNetwork(t, conf)` to setup all the docker instances. That call need the `*testing.T` object and that is not available if I convert this test to an `ExampleTest` :(

smnzhu: Ahhh I see :/

Okay, in that case it's okay, can we at least update https://github.com/onflow/flow-go/pull/1670/files#diff-6df27b5bc6af729e3267f2404aa0d5e17319d629fbfffd697f1272474f59cfd2L19-L20 to explain that these tests are actually meaningful tests, and don't solely exist to serve as examples? thanks :)

smnzhu (approved):

PR #1674, opened 2021-11-24
adding a select filter to filter out elements from the json response

vishalchangrani:
Have not yet plugged it in the server.

### DIFF
--- a/engine/access/rest/handler.go
+++ b/engine/access/rest/handler.go
@@ -2,6 +2,7 @@ package rest
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/rs/zerolog"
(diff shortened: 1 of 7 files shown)


engine/access/rest/select_filter_test.go
```diff
@@ -0,0 +1,144 @@
...
+func ExampleSelectFilter() {
+
+	block, err := generateBlock()
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	selectKeys := []string{
+		"header.id",
```
sideninja, line 20: Could you add a case for top-level select field such as `payload` and at least one that is non-existing like `foo`?

sideninja (approved):
Nice :) Maybe it's good if you also integrate it within this branch/PR.

engine/access/rest/handler.go
```diff
@@ -69,6 +80,7 @@ func (h *Handler) jsonResponse(w http.ResponseWriter, response interface{}, logg
 		h.errorResponse(w, http.StatusInternalServerError, "error generating response", logger)
 		return
 	}
+	fmt.Println(string(encodedResponse))
```
sideninja, line 83: needs to be removed

engine/access/rest/select_filter_test.go
vishalchangrani: will add - but note, currently the select logic only supports leaf elements not top-level elements.

vishalchangrani: https://github.com/dapperlabs/flow-go/issues/6034 added this issue and assigned it to myself

PR #1679, opened 2021-11-25
[Consensus Voting] Replace `VoteAggregatorV2` -> `VoteAggregator`

durkmurder:
### Context

Replaced `VoteAggregator` with `VoteAggregatorV2`. This appears to be a lot more challenging that originally anticipated. The main reason is that `VoteAggregator` now is a standalone component that needs to be started and stopped. It's like an engine and hotstuff event loop. I've updated a lot initialization of consensus components so they are have less strict dependencies. 

Main logic is here, ready to be reviewed. I am still working on documentation and some refactoring like dealing with worker pool. 

### Important changes

- Consensus modules initialization has changed a lot. We initialize hotstuff in 3 steps. First is dependencies (`consensus.HotstuffModules`), second is compliance engine and final step is hotstuff participant(`hotstuff.EventLoop`). 
- We have a circular dependency between `VoteProcessorFactory` and `EventLoop`, solved that using observer pattern.(`pubsub.QCCreatedDistributor`).
- Updated consensus main to support new initialization
- Updated epoch factory to support new initialization
- Updated epoch manager to support startable components 
- Fixed both compliance engines to use vote aggregator, re-enabled tests.


cmd/consensus/main.go
```diff
@@ -644,6 +592,100 @@ func main() {
...
+				node.Storage.Headers,
+				finalize,
+				hotstuffModules,
+				node.RootBlock.Header,
+				node.RootQC,
+			)
+			if err != nil {
+				return nil, err
+			}
+			hotstuffModules = consensus.InitValidator(mainMetrics, hotstuffModules)
```
zhangchiqing, line 617: It's better to make assignment to hotstuffModules fields explicit, so that it's clear to see when hotstuffModule is initialized, there are 3 `nil` fields, and they are updated with 3 assignments. And after the 3 assignments, all modules are created (no more nil field).
```suggestion
			forks, err := consensus.InitForks(
				finalized,
				node.Storage.Headers,
				finalize,
				hotstuffModules,
				node.RootBlock.Header,
				node.RootQC,
			)
			if err != nil {
				return nil, fmt.Errorf("could not init Forks", err)
			}
			hotstuffModules.Forks = forks
			
			validators := consensus.InitValidator(mainMetrics, hotstuffModules)
			hotstuffModules.Validator = validator
```

cmd/consensus/main.go
```diff
@@ -644,6 +592,100 @@ func main() {
...
+
+			voteProcessorFactory := votecollector.NewCombinedVoteProcessorFactory(committee, hotstuffModules.QCCreatedDistributor.OnQcConstructedFromVotes)
+
+			// TODO: add worker pool as separate component
+			workerPool := workerpool.New(4)
+
+			hotstuffModules.Aggregator, err = consensus.NewVoteAggregator(node.Logger, finalized, pending, hotstuffModules, workerPool, voteProcessorFactory)
+			if err != nil {
+				return nil, fmt.Errorf("could not initialize vote aggregator: %w", err)
+			}
```
zhangchiqing, line 627: same as above comment
```suggestion
			aggregator, err := consensus.NewVoteAggregator(node.Logger, finalized, pending, hotstuffModules, workerPool, voteProcessorFactory)
			if err != nil {
				return nil, fmt.Errorf("could not initialize vote aggregator: %w", err)
			}
			hotstuffModules.Aggregator = aggregator
```

consensus/participant.go
```diff
@@ -136,24 +117,34 @@ func NewParticipant(
...
+
+	return modules, nil
+}
+
+func InitValidator(metrics module.HotstuffMetrics, modules *HotstuffModules) *HotstuffModules {
+	// initialize the Validator
+	modules.Validator = validatorImpl.New(modules.Committee, modules.Forks, modules.Signer)
+	modules.Validator = validatorImpl.NewMetricsWrapper(modules.Validator, metrics) // wrapper for measuring time spent in Validator component
+	return modules
 }
```
zhangchiqing, line 148: better to avoid mutation
```suggestion
func InitValidator(metrics module.HotstuffMetrics, modules *HotstuffModules) hotstuff.Validator {
	// initialize the Validator
	validator := validatorImpl.New(modules.Committee, modules.Forks, modules.Signer)
	return validatorImpl.NewMetricsWrapper(validator, metrics) // wrapper for measuring time spent in Validator component
}
```

engine/collection/epochmgr/factories/hotstuff.go
```diff
@@ -57,59 +59,85 @@ func NewHotStuffFactory(
...
 	}
 
+	hotstuffModules, err = consensus.InitForks(
+		finalized,
+		headers,
+		updater,
+		hotstuffModules,
+		cluster.RootBlock().Header,
+		cluster.RootQC(),
+	)
```
zhangchiqing, line 105: same here

engine/collection/epochmgr/engine.go
```diff
@@ -30,20 +33,56 @@ var ErrUnstakedForEpoch = fmt.Errorf("we are not a staked node in the epoch")
...
+		hotstuff:   hotstuff,
+		aggregator: aggregator,
+	}
+
+	builder := component.NewComponentManagerBuilder()
+	builder.AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+		aggregator.Start(ctx)
+		<-util.AllReady(components.prop, components.sync, components.hotstuff, components.aggregator)
+		ready()
+		<-util.AllDone(components.prop, components.sync, components.hotstuff, components.aggregator)
```
jordanschalm, line 67: Why are we calling `Done` on the epoch components here? `ready()` is non-blocking from [what I can see](https://github.com/onflow/flow-go/blob/8febde2129b92393e96eeca6f9d5d8212cea5f14/module/component/component.go#L241-L244), so we will start (`Ready`) then immediately stop (`Done`) all of these components. I think we should only be calling `Done` on the components when we are tearing them down in `stopEpochComponents`.

durkmurder: That is true, a mistake on my side, I have wrongly assumed new semantics for all components, we should wait only for `Startable` components which is `EventLoop` and `VoteAggregator`

utils/workerpool/workerpool.go
```diff
@@ -0,0 +1,32 @@
...
+// WorkerPool is a utility structure which allows wrapping WorkerPool into module.Startable and module.ReadyDoneAware
+// interfaces and allows to use this structure as component.
+type WorkerPool struct {
+	*component.ComponentManager
+	*workerpool.WorkerPool
+}
+
+// New performs initialization of worker pool and setups component manager
+// which will shutdown workers when requested to quit.
+func New(maxWorkers int) *WorkerPool {
```
jordanschalm, line 19:
```suggestion
func NewManaged(maxWorkers int) *WorkerPool {
```
Could name this differently from "github.com/gammazero/workerpool" so it's easier to distinguish what kind of workerpool is being created at a glance

jordanschalm (approved):
Thanks Yuri, this looks good to me. 

(NOTE: I focussed on the collection node / epochmgr sections)

consensus/recovery/participant.go
```diff
@@ -29,7 +29,10 @@ func Participant(
 		}
 
 		// recovery the proposer's vote
-		_ = voteAggregator.StoreProposerVote(proposal.ProposerVote())
+		err = voteAggregator.AddBlock(proposal)
+		if err != nil {
+			return fmt.Errorf("could not process proposal: %w", err)
```
zhangchiqing, line 34:
```suggestion
			return fmt.Errorf("could not process proposal %v: %w", proposal.Block.BlockID, err)
```

cmd/collection/main.go
```diff
@@ -373,7 +374,13 @@ func main() {
 		}).
 		// Epoch manager encapsulates and manages epoch-dependent engines as we
 		// transition between epochs
-		Component("epoch manager", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+		Component("epoch manager", func(nodeBuilder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+
+			workerPool := workerpool.New(4)
+			nodeBuilder.Component("epoch factory worker pool", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+				return workerPool, nil
+			})
```
zhangchiqing, line 382: Better move this close to [where it's used](https://github.com/onflow/flow-go/pull/1679/files#diff-4688e4f1e6bc5063d6bcb143912d69e213eff1019cf5caed6f75218349b70bc2R456-R465).

```suggestion
			workerPool := workerpool.New(4)
			nodeBuilder.Component("hotstuff vote processer worker pool", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
				return workerPool, nil
			})
```

cmd/consensus/main.go
```diff
@@ -519,62 +522,12 @@ func main() {
...
-			)
-			if err != nil {
-				return nil, fmt.Errorf("could not initialized block builder: %w", err)
-			}
-
-			build = blockproducer.NewMetricsWrapper(build, mainMetrics) // wrapper for measuring time spent building block payload component
+			workerPool := workerpool.New(4)
+			builder.Component("vote aggregator worker pool", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+				return workerPool, nil
+			})
```
zhangchiqing, line 530: same here, better move close to where it's used

zhangchiqing (approved):
Looks good. Nice work!

cmd/consensus/main.go
```diff
@@ -644,6 +597,97 @@ func main() {
...
+			hotstuffModules = &consensus.HotstuffModules{
+				Forks:                nil,
+				Validator:            nil,
+				Notifier:             notifier,
+				Committee:            committee,
+				Signer:               signer,
+				Persist:              persist,
+				Aggregator:           nil,
+				QCCreatedDistributor: pubsub.NewQCCreatedDistributor(),
+			}
```
AlexHentschel, line 609: I like this encapsulation of dependencies in `HotstuffModules`, because it makes the method signatures more compact. Specifically for the `HotStuffFactory` for collectors, it   de-bloats the code. 

The specific aspect that concerns me is that we only partially initialize the fields in `HotstuffModules`, while leaving some `nil`. I personally find this somewhat confusing and error prone, because it unnecessarily creates dependencies between different parts of the code, i.e. it reduces modularity. For example: 
* https://github.com/onflow/flow-go/blob/817feda88c1a57fc6c77251e7705f6ff8feaad7f/consensus/participant.go#L143-L148 here, `modules.Forks` is  not nil
* here (same file) `modules.Forks` _is_ nil: https://github.com/onflow/flow-go/blob/817feda88c1a57fc6c77251e7705f6ff8feaad7f/consensus/participant.go#L127
* Hence, I always have to refer back to `consensus.main` and `collection.main` to determine whether I can use a field in `HotstuffModules` to initialize a dependency or whether the field is still nil. Therefore, I feel this 
   * unnecessarily couples the code
   * circumvents helpful compile-time checks, e.g. here: https://github.com/onflow/flow-go/blob/817feda88c1a57fc6c77251e7705f6ff8feaad7f/cmd/consensus/main.go#L611-L623 If I switched the initialization of `hfForks` and the `validator` the compiler will be happy, but the code will crash at run-time, because the reference to Forks will be `nil`. This is only possible, because `HotstuffModules` contains the field `Forks` but we don't initialize it.  
      When creating a reference only when I actually create the instance, I would get a compile-time error about an unknown variable. I consider the latter much safer.  


While I generally like compact code, I would suggest to prioritize API safety higher (even at the cost of more verbose code). Generally (especially in Go), I would suggest to stick with the rule: "Any struct that crosses API boundaries, should contain no `nil` values, unless `nil` is a valid value." In our case, `'nil` is not a valid value for `Forks`, `Validator`, etc. 
  
### Suggestion
* For most of the factory methods in [package `consensus`](https://github.com/onflow/flow-go/blob/817feda88c1a57fc6c77251e7705f6ff8feaad7f/consensus/participant.go) (like `NewValidator`, `NewForks`, `NewVoteAggregator`, `NewFollower`), I would stick with using basic dependencies. I think the overhead is just one or two extra fields in the method signature, which I think is fine.
* However, for `NewParticipant`, we could use  `HotstuffModules`:
   * `NewParticipant` takes a lot of dependencies, so encapsulating some of them substantially reduces the method signature
   * right before we call `NewParticipant`, all fields in `HotstuffModules` are known. I.e. we can create `HotstuffModules`, without any `nil` values. 

I have implemented this suggestion here: https://github.com/onflow/flow-go/pull/1700 (PR is targeting this branch). Some of the tests still need to be updated.

cmd/collection/main.go
```diff
@@ -373,7 +374,13 @@ func main() {
 		}).
 		// Epoch manager encapsulates and manages epoch-dependent engines as we
 		// transition between epochs
-		Component("epoch manager", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+		Component("epoch manager", func(nodeBuilder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+
+			workerPool := workerpool.New(4)
+			nodeBuilder.Component("epoch factory worker pool", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
```
AlexHentschel, line 380: To me, initializing the `workerPool` as a stand-alone `Component` seems a bit of a hacky approach: 
* Conceptually, I feel that the Worker pool belongs to the `VoteAggregator` and should be managed by it. All other worker threads are usually tied to the component. 

Having said this, I think it might require a noteworthy amount of work to update the code. Not sure it is worth it ... probably not as part of this PR anyway.

AlexHentschel (commented):
Some high-level structural comments. Only got through part of the RP

consensus/hotstuff/notifications/pubsub/qc_created_distributor.go
```diff
@@ -0,0 +1,103 @@
...
+// QCCreatedDistributor subscribes for qc created event from hotstuff and distributes it to subscribers
+// Objects are concurrency safe.
+// NOTE: it can be refactored to work without lock since usually we never subscribe after startup. Mostly
+// list of observers is static.
+type QCCreatedDistributor struct {
+	qcCreatedConsumers []OnQCCreatedConsumer
+	lock               sync.RWMutex
+}
+
+var _ hotstuff.Consumer = (*QCCreatedDistributor)(nil)
```
AlexHentschel, line 22: I would suggest to create a dedicated interface for this (well, specifically for _every_ data producer). Otherwise, it will be really confusing to determine later which component produces what data.

cmd/consensus/main.go
durkmurder: Agree, that's kinda what made me feel uneasy as well.

cmd/collection/main.go
```diff
@@ -373,7 +374,13 @@ func main() {
 		}).
 		// Epoch manager encapsulates and manages epoch-dependent engines as we
 		// transition between epochs
-		Component("epoch manager", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+		Component("epoch manager", func(nodeBuilder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+
+			workerPool := workerpool.New(4)
```
AlexHentschel, line 379: I am wondering if we should abstract this worker pool by an interface. Atm, the worker pool exposes methods to terminate it. It took a bit of time to figure out how the life-cycle management of the worker pool is implemented, specifically because it looks like any component could just terminate the worker pool ... Though, the current implementation would badly break if some component other than main terminated the pool, but we do not represent this requirement in code.

From a conceptual perspective it would make most sense to me, if the worker pools was part of `VoteCollectors` (and `VoteCollectors` would implement the `ReadyDoneAware` and `Startable` interfaces). 

#### suggestion
* we currently only require two distinct methods for the worker pool:
  * a `Submit` method 
  * and a method to terminate it (`StopWait`). 
* I would suggest we introduce two interfaces (maybe put them in `consensus/hotstuff/vote_collectors.go`)
* `VoteCollectors` has a reference to the worker pool so it can terminate it (for this purpose, `VoteCollectors` would itself need to implement the `ReadyDoneAware` and `Startable` interfaces)
* `VoteAggregator` contains `VoteCollectors`

I have implemented this in PR https://github.com/onflow/flow-go/pull/1702 Though, I must admit that it adds quite some complexity to the code. In hindsight, I am not really sure it is worth it.

durkmurder: Thanks for your PR, I think it's well worth it. Originally I was a bit bothered that factory method starts accepting worker pool, not sure if there is a way around it. But let's give it a try, either way it makes start/stop simpler in main code base and tests.

consensus/aggregator.go
```diff
@@ -0,0 +1,66 @@
+package consensus
```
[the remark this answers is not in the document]
durkmurder: our constructors would collide, since `voteaggregator` package has it's own `NewVoteAggregator`

consensus/hotstuff/notifications/pubsub/qc_created_distributor.go
durkmurder: Done

consensus/aggregator.go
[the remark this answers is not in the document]
AlexHentschel: sorry, not rename the package; just the _file_

consensus/aggregator.go
```diff
@@ -0,0 +1,43 @@
...
+	aggregator, err := voteaggregator.NewVoteAggregator(log, notifier, finalized.View, voteCollectors)
+	if err != nil {
+		return nil, fmt.Errorf("could not create vote aggregator: %w", err)
+	}
+
+	// recover the hotstuff state, mainly to recover all pending blocks in Forks
+	err = recovery.Participant(log, forks, aggregator, validator, finalized, pending)
+	if err != nil {
+		return nil, fmt.Errorf("could not recover hotstuff state: %w", err)
+	}
```
AlexHentschel, line 40: To me, it would be more intuitive to leave the recovery in `NewParticipant`, specifically here:  https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/participant.go#L64-L65
Otherwise, we remove the nice symmetry with the Follower: https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/follower.go#L31-L35 and the code becomes more erratic.  

### Suggestion: 
1. We remove the parameters `pendingBlocks` from the `VoteAggregator`'s constructor. Technically, we could just initialize it with view zero and prune later. 
2. Thereby, we avoid the need to have the pending blocks and can replace https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/cmd/consensus/main.go#L588 by
    ```golang
    finalizedBlock, err := state.Final().Head()
    ```
3. our pruning code is super efficient for a empty `VoteCollectors`, so we can just prune here: https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/participant.go#L64
4. We move the recovery code back into `NewParticipant` to line 65 https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/participant.go#L64-L65

### Caution: our pruning convention is inconsistent and needs to be fixed:
* `VoteCollectors` removes elements _below_ the given view number: https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/hotstuff/voteaggregator/vote_collectors.go#L102-L103
* while `VoteAggregator` claims to include the specified view: https://github.com/onflow/flow-go/blob/b4f763879799deb3993e744dbba04941a217934d/consensus/hotstuff/voteaggregator/vote_aggregator.go#L214-L221 (but the implementation doesn't really do so). 

Overall, I would suggest to stick with the existing convention: 
> Prune prunes all blocks UP TO but NOT INCLUDING the given value

Our `VoteAggregator` implementation already does this, but the method documentation is different.

consensus/participant.go
```diff
@@ -60,39 +49,19 @@ func NewParticipant(
...
 
-	// initialize the vote aggregator
-	aggregator := voteaggregator.New(notifier, 0, committee, validator, signer)
-
-	// recover the hotstuff state, mainly to recover all pending blocks
-	// in forks
-	err = recovery.Participant(log, forks, aggregator, validator, finalized, pending)
-	if err != nil {
-		return nil, fmt.Errorf("could not recover hotstuff state: %w", err)
-	}
```
AlexHentschel, line 95: as mentioned above, I think we should:
1. first prune the `VoteAggregator`
2. then recover the internal state of `Forks` and `VoteAggregator` here by feeding them the unfinalized blocks.

AlexHentschel (approved):
This was a massive PR. Good job. Thanks for adding method documentation in various places. 👏

cmd/consensus/main.go
```diff
@@ -639,11 +585,102 @@ func main() {
...
+			if err != nil {
+				return nil, fmt.Errorf("could not initialize compliance core: %w", err)
+			}
+
+			// initialize the compliance engine
+			comp, err = compliance.NewEngine(node.Logger, node.Network, node.Me, prov, core)
+			if err != nil {
+				return nil, fmt.Errorf("could not initialize compliance engine: %w", err)
+			}
+			return comp, nil
```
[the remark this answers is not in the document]
durkmurder: Yeah, we should probably merge it, in the end we only need to start compliance engine

cmd/collection/main.go
```diff
@@ -373,7 +373,7 @@ func main() {
 		}).
 		// Epoch manager encapsulates and manages epoch-dependent engines as we
 		// transition between epochs
-		Component("epoch manager", func(_ cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+		Component("epoch manager", func(nodeBuilder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
```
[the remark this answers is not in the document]
durkmurder: not anymore

consensus/hotstuff/eventhandler/event_handler.go
```diff
@@ -1,3 +1,4 @@
+//nolint
```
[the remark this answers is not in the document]
durkmurder: I would leave this for another PR. Since there will be a lot of changes for tests.

durkmurder:
bors merge

PR #1682, opened 2021-11-25
Remove the replace statement for flow-go/crypto

turbolent:
`replace` statements in `go.mod` files only apply locally, not in downstream dependencies.

This means that when building (or testing, e.g. on CI) `flow-go` itself, the replace statement is effective, and the latest version of `flow-go/crypto` is used.

However, the replace statement has **no effect** when using `flow-go` as a dependency, which e.g. the emulator does.

Once again `flow-go` currently does not build when using it as a dependency, delaying the emulator update:

```
../../../go/pkg/mod/github.com/onflow/flow-go@v0.23.2-0.20211125190211-7b31c986316e/model/flow/identifier.go:115:2: undefined: "github.com/onflow/flow-go/crypto/hash".ComputeSHA3_256
../../../go/pkg/mod/github.com/onflow/flow-go@v0.23.2-0.20211125190211-7b31c986316e/model/flow/identifier.go:124:2: undefined: "github.com/onflow/flow-go/crypto/hash".ComputeSHA3_256
```

This PR removes the replace statements, which should not be committed and only be used for local development purposes, and pins `flow-go/crypto` temporarily to the latest commit. Ideally, `flow-go/crypto` should get tagged with a proper version and `flow-go`'s `go.mod` updated accordingly.

Also see https://axiomzen.slack.com/archives/C015G65HR2P/p1637867663058600


Kay-Zee:
confused where all these lint errors are coming from.

crypto/build_dependency.sh
```diff
@@ -5,6 +5,26 @@ set -euo pipefail
 PKG_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 RELIC_DIR_NAME="relic"
 RELIC_DIR="${PKG_DIR}/${RELIC_DIR_NAME}"
+MOD_DIR="/pkg/mod/"
```
tarakby, line 8: This is maybe still wip, so just an early comment: 
This new part of the script should live outside the crypto package, in our case it would be under the root `flow-go/`. The reason is that `go get` copies the entire crypto package files under `GOPATH/../crypto@v0...`. There needs to be a new script that is run from the project root, and is able to find that specific GOPATH folder.

tarakby: I can continue working on this PR next week if you prefer, since I already did something similar in a local branch.

Kay-Zee: Currently, this script will perform both the install in the mod folder, and for the one it currently lives in, and is my preferred approach. There is no reason that it has to be a separate script in my opinion, but we can chat about it if you like.

tarakby: I didn't fully understand the approach I guess, so I'll wait till the PR is ready.

tarakby:
@Kay-Zee  should I close this PR and keep the new https://github.com/onflow/flow-go/pull/1742 ?

tarakby:
Closing this PR as it is replaced by https://github.com/onflow/flow-go/pull/1742.

PR #1690, opened 2021-11-27
adding support for getblockbyheight with height as 'final' or 'sealed'

vishalchangrani:


engine/access/rest/converter.go
```diff
@@ -24,6 +24,16 @@ const maxAuthorizersCnt = 100
 const MaxAllowedIDs = 50 // todo(sideninja) discuss if we should restrict maximum on all IDs collection or is anywhere required more thant this
 const MaxAllowedHeights = 50
 
+// constants to indicate if final or sealed height was requested
+const (
+	final = iota + 1
```
sideninja, line 29: what if they want to fetch block height 0? this constant equals to that, it might be quite an edge case but something to think about.

sideninja (approved):

engine/access/rest/converter.go
vishalchangrani: changed the implementation to not use these constants at all since they make the code difficult to read.

PR #1691, opened 2021-11-29
Epoch counter reporter

janezpodhostnik:
ref: https://github.com/dapperlabs/flow-go/issues/5998

Epoch counter is reported during state extraction:

image

(tested on canary 11 state)

I added it as the first reporter, so this will be reported first.


Kay-Zee (approved):

ramtinms (approved):

Kay-Zee:
bors merge

PR #1693, opened 2021-11-29
[DKG integration] do not crash for DKG failures

tarakby:
Addresses [#5931 ](https://github.com/dapperlabs/flow-go/issues/5931)

- add a new sentinel error in the crypto package
- use the new sentinel error in the reactor engine (DKG integration)


crypto/dkg.go
```diff
@@ -53,6 +58,33 @@ type DKGState interface {
...
+
+// dkgFailureErrorf constructs a new dkgFailureError
+func dkgFailureErrorf(msg string, args ...interface{}) error {
+	return &dkgFailureError{
+		err: fmt.Errorf(msg, args...),
+	}
+}
+
+func (e dkgFailureError) Error() string {
+	return e.err.Error()
```
jordanschalm, line 77:
```suggestion
	return fmt.Sprintf("dkg failure: %s", e.err.Error())
```
include the context of this error level

jordanschalm (approved):
Thank you, looks good - one small suggestion.

kc1116 (approved):

crypto/dkg.go
tarakby: I think the errors built using `dkgFailureErrorf` are self explanatory. I think wrapping the error message in a context is useful when the error itself is returned by another function, for instance [here](https://github.com/onflow/flow-go/blob/tarak/5931-dkg-failure-bug/engine/consensus/dkg/reactor_engine.go#L278).

On a different note, I think I could embedded the error in the structure `dkgFailureError` and avoid defining `Error()`.

jordanschalm: I think it's useful to have an error type include some context for example for searching logs. 

If I want to search for `dkgFailureError` errors, I can search for the "dkg failure: "` string

tarakby: added in https://github.com/onflow/flow-go/pull/1693/commits/52236feb1ee7f0e17337c5129eb951b6b56e55d8 with missing tests

tarakby:
bors merge

PR #1695, opened 2021-11-29
Add storage V6 migration to v0.23

SupunS:
Adding back the storage migration V6 to `v0.23` that was left out when merging to master.

This simply copies the two files (`storage_v6.go` and `storage_v6_test.go`) from the [v0.23-testnet](https://github.com/onflow/flow-go/tree/v0.23-testnet) branch.


cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -66,11 +66,17 @@ func extractExecutionState(
...
+			Log:       log,
+			OutputDir: outputDir,
+		}
+
 		migrations = []ledger.Migration{
-			storageUsedUpdateMigration.Migrate,
 			mgr.PruneMigration,
+			storageFormatV6Migration.Migrate,
+			storageUsedUpdateMigration.Migrate,
+			mgr.PruneMigration, // we need post migration pruning before deploying trie update pruning
```
SupunS, line 78: I added this post-migration pruning step, similar to what we had in `v0.22-testnet`:
https://github.com/onflow/flow-go/blob/eaf0e5c1e20cde49b94a7ed6bc1768af19bfd3a8/cmd/util/cmd/execution-state-extract/execution_state_extract.go#L72-L76

turbolent (approved):
🎉

cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -64,7 +65,10 @@ func (r *AccountReporter) Report(payload []ledger.Payload) error {
 	defer rwm.Close()
 
 	l := migrations.NewView(payload)
-	st := state.NewState(l)
+	st := state.NewState(
+		l,
+		state.WithMaxInteractionSizeAllowed(math.MaxUint64),
+	)
```
SupunS, line 71: Had to increase the `MaxInteractionSizeAllowed` here and below.
cc: @janezpodhostnik

cmd/util/ledger/migrations/storage_v6.go
```diff
@@ -0,0 +1,3773 @@
...
+	deferredValuePaths := make(map[storagePath]bool)
+	for _, payload := range payloads {
+		m.incrementProgress()
+
+		keyParts := payload.Key.KeyParts
+		rawOwner := keyParts[0].Value
+		rawController := keyParts[1].Value
+		rawKey := keyParts[2].Value
+
+		if state.IsFVMStateKey(
```
janezpodhostnik, line 276: before calling this function the payloads were split: `fvmPayloads, storagePayloads, slabPayloads := splitPayloads(payloads)`

would it make sense here to only use ` storagePayloads` and `slabPayloads` here and skip this if condition?

janezpodhostnik (approved):
Overall looks good to me. I did not review the smart contracts.

cmd/util/ledger/migrations/storage_v6.go
SupunS: yeah, good idea!
We didn't have the split function earlier, and this wasn't cleaned up once it was introduced. good catch!

turbolent (approved):
👍

PR #1697, opened 2021-11-30
[FLIP] Sync engine / core updates

smnzhu:
This FLIP summarizes ideas from https://github.com/dapperlabs/flow-go/issues/5934, review comments from https://github.com/onflow/flow-go/pull/1573, as well as a lengthy discussion / brainstorm with @zhangchiqing.

In addition, it serves as a form of documentation for the sync protocol, hopefully helping others better understand the details of how the sync protocol works.

FLIP: https://github.com/onflow/flow-go/blob/sync-engine-flip/flips/sync-protocol.md


flips/sync-protocol.md
```diff
@@ -0,0 +1,90 @@
...
+
+At any time, there is a single range of heights that the Sync Engine actively requests, which is tracked by the Sync Core. We call this the Active Range. The Active Range is parameterized by two variables, which effectively replace the `Heights` map from the existing implementation:
+* `RangeStart` is the start height of the range, which is greater than or equal to `LocalFinalizedHeight`, the local finalized block height.
+* `RangeEnd` is the end height of the range, which is less than or equal to `TargetFinalizedHeight`, the speculated finalized block height of the overall chain (more details below).
+
+The Active Range has a default size of `DefaultRangeSize`. `RangeStart` is updated according to the following conditions, and `RangeEnd` should always be updated as necessary to satisfy `RangeEnd - RangeStart >= DefaultRangeSize`:
+* If the local finalized height has increased by `h`, increase `RangeStart` by `h`.
+* If every height in the Active Range has been received from at least `MinResponses` different nodes, increase `RangeStart` by `DefaultRangeSize`.
```
jordanschalm, line 76:
```suggestion
* If every height in the Active Range has been received from at least `MinResponses` different nodes, set `RangeStart=RangeEnd+1`.
```
Since we allow for the Active Range to be `>=DefaultRangeSize`

flips/sync-protocol.md
```diff
@@ -0,0 +1,90 @@
...
+
+> **Note:** The second condition is needed to cover the last item in [Potential Problems](#potential-problems). To guarantee liveness, `MinResponses` should be set to a value greater than the maximum tolerable number of Byzantine nodes. Alternatively, we could increase `RangeEnd` by `DefaultRangeSize` instead of increasing `RangeStart`, thereby allowing the Active Range to grow without bound and obviating the need for any lower bound on `MinResponses`. 
+> 
+> A possible optimization here is that instead of waiting for the entire Active Range to satisfy the update condition, we can update as soon as any sub-range starting from `RangeStart` satisfies the condition.
+
+The Active Range can be broken up and requested by the Sync Engine in multiple segments, similar to the existing implementation.
+
+We track the list of responses we've received for each height in a data structure called `Responses`, which we can use to avoid sending duplicate requests to the same node. We can also use it later on to slash nodes for sending faulty responses.
```
jordanschalm, line 84: The `Response` message should explicitly link block height with block ID for "by-height" requests, otherwise slashing would be difficult (since the requester can manufacture any request message with the same nonce).

flips/sync-protocol.md
```diff
@@ -0,0 +1,90 @@
...
+> 
+> A possible optimization here is that instead of waiting for the entire Active Range to satisfy the update condition, we can update as soon as any sub-range starting from `RangeStart` satisfies the condition.
+
+The Active Range can be broken up and requested by the Sync Engine in multiple segments, similar to the existing implementation.
+
+We track the list of responses we've received for each height in a data structure called `Responses`, which we can use to avoid sending duplicate requests to the same node. We can also use it later on to slash nodes for sending faulty responses. 
+
+The value of `TargetFinalizedHeight` should reflect the Sync Height Responses that have been received, yet account for the possibility that some of these responses are malicious. Therefore, the Sync Height Response processing logic should incorporate some sort of expiration mechanism. Some possibilities for determining `TargetFinalizedHeight` are:
+* Calculate a moving average of all the Sync Height Responses received.
```
jordanschalm, line 87: Idea: make `SyncResponse`s commit to a block ID corresponding to their claimed latest finalized height, for slashing.

Also, I think significant outliers would pose a problem to all 3 of these approaches. We should consider discarding responses which are obviously invalid (eg. days or weeks ahead of the mean).

jordanschalm (approved):
Nice detailed writeup -- thank you. 👏

flips/sync-protocol.md
smnzhu: @jordanschalm That's a good idea. As for outliers, I think this can mostly be mitigated by taking the median instead of the mean.

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+* A Range Request response should contain a single chain of blocks which begins at the start height of the requested range and is no longer than the size of the requested range.
+* A Batch Request response should contain a subset of the requested block IDs.
+
+This eliminates any ambiguity about whether a response corresponds to a Batch or Range Request, so we can avoid optimistically setting the statuses of heights as Received for responses to Batch Requests.
+
+At any time, there is a single range of heights that the Sync Engine actively requests, which is tracked by the Sync Core. We call this the Active Range. The Active Range is parameterized by two variables `RangeStart` and `RangeEnd`, which effectively replace the `Heights` map from the existing implementation, but it can be broken up and requested by the Sync Engine in multiple segments. `RangeStart` should be greater than the local finalized block height, and `RangeEnd` should be less than or equal to the target finalized block height (more details below). The logic for updating the Active Range can be abstracted with an interface:
```
AlexHentschel, line 69: > RangeStart should be greater than the local finalized block height

I think to guarantee liveness, we would need

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+type ActiveRange interface {
+    // Update processes a range of blocks received from a Range Request 
+    // response and updates the requestable height range.
+    Update(headers []flow.Header, originID flow.Identifier)
+
+    // LocalFinalizedHeight is called to notify a change in the local finalized height.
+    LocalFinalizedHeight(height uint64)
+
+    // TargetFinalizedHeight is called to notify a change in the target finalized height.
+    TargetFinalizedHeight(height uint64)
```
AlexHentschel, line 81: Sorry, the differentiation between `local finalized height` and `target finalized height` is not quite clear to me.

AlexHentschel (commented):
Looks good. Thanks for the thorough write up.

AlexHentschel (approved):

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+
+At any time, there is a single range of heights that the Sync Engine actively requests, which is tracked by the Sync Core. We call this the Active Range. The Active Range is parameterized by two variables `RangeStart` and `RangeEnd`, which effectively replace the `Heights` map from the existing implementation, but it can be broken up and requested by the Sync Engine in multiple segments. `RangeStart` should be greater than the local finalized block height, and `RangeEnd` should be less than or equal to the target finalized block height (more details below). The logic for updating the Active Range can be abstracted with an interface:
+
+```golang
+type ActiveRange interface {
+    // Update processes a range of blocks received from a Range Request 
+    // response and updates the requestable height range.
+    Update(headers []flow.Header, originID flow.Identifier)
+
+    // LocalFinalizedHeight is called to notify a change in the local finalized height.
```
zhangchiqing, line 77:
```suggestion
    // LocalFinalizedHeight is called to notify a change in the local finalized height.
    // When this value increases, I can prune a range of blocks that hasn't been received.
```

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+    LocalFinalizedHeight(height uint64)
+
+    // TargetFinalizedHeight is called to notify a change in the target finalized height.
```
zhangchiqing, line 80:
```suggestion
    // TargetFinalizedHeight is the finalized height from the peer, when the TargetFinalizedHeight
    // is below the LocalFinalizedHeight, it means the node is falling behind, and needs to sync
    // blocks to catch up. 
```

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+There are many ways to implement this interface, but one possible approach is as follows:
+* Select values for parameters `DefaultRangeSize` and `MinResponses`
+* Let `PendingStart` be the first height greater than `LocalFinalizedHeight` that has been received less than `MinResponses` times
+* Let `RangeStart` be equal to `LocalFinalizedHeight + 1`
+* Let `RangeEnd` be the smaller of `TargetFinalizedHeight` and `PendingStart + DefaultRangeSize`
+
+The reason we keep track of `PendingStart` is to ensure that `RangeEnd` eventually increases even if the local finalized height doesn't. This is needed to address the last item in [Potential Problems](#potential-problems).
```
zhangchiqing, line 94:
```suggestion
The reason we keep track of `PendingStart` is that if we don't track, then the syncing might get stuck in an edge case where there are a range of blocks that doesn't finalize any block, and we kept fetching these heights, and receiving the same range of blocks over and over. `PendingStart` ensures that `RangeEnd` eventually increases even if the local finalized height doesn't change after receiving the range of blocks. This is needed to address the last item in [Potential Problems](#potential-problems).
```

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+    // Update processes a height received from a Sync Height Response 
+    // and updates the finalized height estimate.
+    Update(height uint64, originID flow.Identifier)
+
+    // Get returns the estimated finalized height of the overall chain.
+    Get() uint64
+}
+```
+
+One possible approach is to maintain a sliding window of the most recent Sync Height Responses, and take the median of these values. This implies that the target finalized height will always lag slightly behind the true finalized height, which may or may not be a problem depending on the block finalization rate.
```
zhangchiqing, line 109: So far, we've mentioned about `RangeRequest`, and we'd better also talk about the `BatchRequest`. 

`BatchRequests` are independently calls from compliance engine, there is some concurrency tension between the `RangeRequest` and `BatchRequest`: when compliance engine is asking sync engine to fetch missing blocks, the sync engine might be sending range requests.

To remove the concurrency tension, we maintain a map of missing block IDs, and check the map with a time interval.

When a timer triggers, sync engine will ask sync core for RangeRequest and BatchRequest to be sent out. Now this is where we need to talk about priority. 

Should we give priority to `RangeRequest`? Yes, because if we are falling far behind, the `BatchRequest` will not help increase the local finalized height. So it's better to focus sending `RangeRequest`. For the same reason, if a missing block is far from the local finalized height, we could probably drop it, instead of keeping it in the missing block's map. That's because even if we keep them in that map, their priority is still too low to be included in a BatchRequest. We should give the `RangeRequest` priority to catch up, which is more efficient. And the `RangeRequest` would eventually reach the height of that missing block, and fetch it.

So we will store missing block IDs in a map when they are not too far from local finalized height, so that they will be included in a Batch Request.

For a RangeRequest, when there are a range of heights missing, it's easy to determine which height to fetch first: we should always start with the lowest height. But for BatchRequest, when there are a number of missing blocks with different height, and there are more blocks to be fit in a BatchRequest, how to choose them? Should we give priority to blocks with lower heights? The problem with this is that those blocks might never finalize or not exist, so we might waste effort and stuck if we keep fetching non-existing blocks from malicious nodes. Another solution is to use a _first come last serve_ rule, we will fetch those that requested recently.

flips/sync-protocol.md
```diff
@@ -0,0 +1,109 @@
...
+* Let `PendingStart` be the first height greater than `LocalFinalizedHeight` that has been received less than `MinResponses` times
+* Let `RangeStart` be equal to `LocalFinalizedHeight + 1`
+* Let `RangeEnd` be the smaller of `TargetFinalizedHeight` and `PendingStart + DefaultRangeSize`
+
+The reason we keep track of `PendingStart` is to ensure that `RangeEnd` eventually increases even if the local finalized height doesn't. This is needed to address the last item in [Potential Problems](#potential-problems). 
+
+The target finalized height represents the speculated finalized block height of the overall chain, and should reflect the Sync Height Responses that have been received while accounting for the possibility that some of these responses are malicious. Therefore, the Sync Height Response processing logic should incorporate some sort of expiration / filtering mechanism. The details of this logic can be abstracted with an interface:
+
+```golang
+type TargetFinalizedHeight interface {
```
zhangchiqing, line 99: Does the sync core maintain a range of `TargetFinalizedHeight`, such that if say `ActiveRange.LocalFinalizedHeight` and `ActiveRange.TargetFinalizedHeight` has 100 heights apart, there are `100` `TargetFinalizedHeight` instances?
Does `TargetFinalizedHeight` maintain the `MinResponse`?

flips/sync-protocol.md
smnzhu: This is already explained in Potential Problems, so I don't think there's a need to reiterate here.

flips/sync-protocol.md
smnzhu: It's actually the other way around.

smnzhu: > TargetFinalizedHeight is the finalized height from the peer

Also, this is not entirely accurate.

flips/sync-protocol.md
smnzhu: We are already giving range requests priority, so nothing changes here.

As for the prioritization of batch requests, I would consider this more of a detail, and my intention is not to prescribe specific rules about how this should be implemented. There is more than one solution here, and I don't think we have any strong reason to change from the current method, which is random.

flips/sync-protocol.md
smnzhu: there is only a single target finalized height. It is a single value that is a function of all the data points we've received from Sync Height Responses so far.

> Does TargetFinalizedHeight maintain the MinResponse

I don't really think `TargetFinalizedHeight` and `MinResponse` have that much to do with each other.

smnzhu:
bors merge

PR #1701, opened 2021-11-30
Implemented get account by final and sealed height and modified unit test

vishalchangrani:


engine/access/rest/accounts.go
```diff
@@ -29,5 +46,46 @@ func getAccount(r *requestDecorator, backend access.API, link LinkGenerator) (in
 		}
 	}
 
-	return accountResponse(account), nil
+	expandContracts := r.expands(expandableFieldContracts)
+	expandKeys := r.expands(expandableFieldKeys)
+	return accountResponse(account, link, expandKeys, expandContracts)
+}
+
+func accountResponse(flowAccount *flow.Account, link LinkGenerator, expandKeys bool, expandContracts bool) (generated.Account, error) {
```
vishalchangrani, line 54: @sideninja I moved account response here from converter. I didn't find a good way of not passing the link generator tbh. Passing it actually makes testing easy since its injected.
Also, implemented the `if-then-else` logic for the expandables - `keys` and `contracts`

engine/access/rest/accounts_test.go
```diff
@@ -28,77 +33,76 @@ func accountURL(address string, height string) string {
 func TestGetAccount(t *testing.T) {
 	backend := &mock.API{}
 
-	t.Run("get by address at latest block", func(t *testing.T) {
-		heights := []string{"", "latest"}
+	t.Run("get by address at latest sealed block", func(t *testing.T) {
```
vishalchangrani, line 36: the height can be `final` or `sealed` (not `latest`). Added two unittest cases accordingly

engine/access/rest/blocks_test.go
```diff
@@ -31,8 +31,10 @@ func blockURL(ids []string, start uint64, end uint64, payload bool) string {
 }
 
 func TestGetBlocks(t *testing.T) {
+	t.Skip()
```
vishalchangrani, line 34: yet to do

engine/access/rest/collections_test.go
```diff
@@ -32,6 +33,12 @@ func executeRequest(req *http.Request, backend *mock.API) *httptest.ResponseReco
 	return rr
 }
 
+func assertOKResponse(t *testing.T, req *http.Request, expectedRespBody string, backend *mock.API) {
```
vishalchangrani, line 36: just a convenience wrapper to check for OK results

engine/access/rest/converter.go
```diff
@@ -560,19 +560,3 @@ func accountKeysResponse(keys []flow.AccountPublicKey) []generated.AccountPublic
 
 	return keysResponse
 }
-
-func accountResponse(account *flow.Account) generated.Account {
```
vishalchangrani, line 564: moved to `accounts.go`

engine/access/rest/middleware/common_query_params.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/gorilla/mux"
 )
 
-const expandQueryParam = "expand"
+const ExpandQueryParam = "expand"
```
vishalchangrani, line 10: since this constant is used in the test too

utils/unittest/fixtures.go
```diff
@@ -134,11 +134,15 @@ func AccountFixture() (*flow.Account, error) {
 		return nil, err
 	}
 
+	contracts := make(map[string][]byte, 2)
```
vishalchangrani, line 137: added some dummy contracts

engine/access/rest/accounts.go
sideninja: How come that has been moved here? I feel that it's getting more confusing if we do some response creation here and some in converters. I always prefer to have one place to do a certain functionality so we either have handlers to do also result generation or we keep converters to do that.

engine/access/rest/accounts.go
```diff
@@ -29,5 +46,46 @@ func getAccount(r *requestDecorator, backend access.API, link LinkGenerator) (in
...
+		}
+	}
+
+	if expandContracts {
+		contracts := make(map[string]string, len(flowAccount.Contracts))
+		for name, code := range flowAccount.Contracts {
+			contracts[name] = toBase64(code)
+		}
+		account.Contracts = contracts
+	} else {
```
sideninja, line 75: I think the expandable can always be static (maybe we should check with @smnzhu) no matter what the query is. If that is true then we don't need to dynamically set them and they can just be as part of response generator.

engine/access/rest/accounts.go
sideninja: If that was done due to expanded keys then we can just call `accountResponse` and remove result values if expanded is not passed like so:
```go
account := accountResponse(account, link)

if !r.expands(keys) {
    account.keys = nil
}
```
But I would rather have all of that in converters as that is one place that is generating the response, but I think that passing objects to it is not the best idea so I agree we should change it to support booleans like `expandKeys`, we can even pass the link if you think that is better and generate it in the handler.

engine/access/rest/accounts_test.go
sideninja: Good that you fixed this but in general I also really think we should support `latest` alias, I've pinged Simon about it to open a discussion. We don't have any documentation about statuses anywhere and developers usually know what latest means whereas final and sealed is really technical. Even to me I have to think couple of times which one means latest.

engine/access/rest/accounts_test.go
```diff
@@ -110,3 +114,50 @@ func TestGetAccount(t *testing.T) {
 		}
 	})
 }
+
+func expectedExpandedResponse(account *flow.Account) string {
```
sideninja, line 118: good idea!

engine/access/rest/middleware/common_query_params.go
sideninja: yep good idea

engine/access/rest/accounts_test.go
```diff
@@ -28,77 +33,76 @@ func accountURL(address string, height string) string {
...
-					 "public_key":"%s",
-					 "signing_algorithm":"ECDSA_P256",
-					 "hashing_algorithm":"SHA3_256",
-					 "sequence_number":0,
-					 "weight":1000,
-					 "revoked":false
-				  }
-			   ]
-			}`, account.Address, account.Keys[0].PublicKey.String())
+		assertOKResponse(t, req, expected, backend)
```
sideninja, line 82: I like this utility function

engine/access/rest/converter.go
vishalchangrani: moved back converter.go as discussed.

engine/access/rest/accounts.go
vishalchangrani: I like the idea of having the accountResponse in the converter and passing an `expand` map to let it decide what fields to expand. I have modified the code accordingly.

PR #1703, opened 2021-11-30
REST: Add improvements and tests to handlers

sideninja:
This PR implements handlers for expand fields and adding more tests. Converter tests were added with implementing expendables and links as part of that functionality which allows us to test different combinations of response generation without handlers.

### DIFF
--- a/engine/access/rest/scripts.go
+++ b/engine/access/rest/scripts.go
@@ -11,9 +11,11 @@ func executeScript(r *requestDecorator, backend access.API, _ LinkGenerator) (in
 	blockID := r.getQueryParam("block_id")
 	blockHeight := r.getQueryParam("block_height")
 
+	if blockID == "" && blockHeight == "" {
+		return nil, NewBadRequestError(fmt.Errorf("either block ID or block height must be provided"))
+	}
 	if blockID != "" && blockHeight != "" {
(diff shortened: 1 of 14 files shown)


engine/access/rest/scripts.go
```diff
@@ -32,50 +34,46 @@ func executeScript(r *requestDecorator, backend access.API, _ LinkGenerator) (in
...
-
-		result, err := backend.ExecuteScriptAtBlockID(r.Context(), id, code, args)
+		height = finalBlock.Height
+	} else {
+		height, err = toHeight(blockHeight)
 		if err != nil {
-			return nil, err
+			return nil, NewBadRequestError(err)
 		}
+	}
```
sideninja: @vishalchangrani good catch

engine/access/rest/transactions.go
```diff
@@ -19,16 +19,16 @@ func getTransactionByID(r *requestDecorator, backend access.API, link LinkGenera
...
 		if err != nil {
 			return nil, err
 		}
-
-		return transactionResponse(tx, txr, link, r.expandFields), nil
 	}
 
-	return transactionResponse(tx, nil, link, r.expandFields), nil
+	return transactionResponse(tx, txr, link, r.expandFields), nil
 }
```
sideninja: @vishalchangrani 👍

PR #1706, opened 2021-12-01
[Consensus Voting] Replace Signer Verifier

zhangchiqing:
This PR replaces the old signer and verifier modules with the new ones for consensus voting v2. 

The signer modules are used by the consensus nodes, and the verifiers are used by all node types.

The SignerVerifier modules are replaced by Signer module and Verifier module


consensus/hotstuff/verification/metrics_wrapper.go
```diff
@@ -5,40 +5,41 @@ import (
...
-}
-
-func (w SignerMetricsWrapper) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) (bool, error) {
-	processStart := time.Now()
-	valid, err := w.signer.VerifyQC(signers, sigData, block)
-	w.metrics.SignerProcessingDuration(time.Since(processStart))
-	return valid, err
-}
+// TODO: to be moved to VerifierMetricsWrapper
+// func (w SignerMetricsWrapper) VerifyVote(voter *flow.Identity, sigData []byte, block *model.Block) (bool, error) {
```
durkmurder, line 30: Will it be moved? We have metrics for validation which includes verification.

durkmurder (approved):
Nice, thank you. Almost there!

consensus/hotstuff/signer.go
```diff
@@ -2,23 +2,13 @@ package hotstuff
...
 
-// SignerVerifier can sign and verify HotStuff entities.
-type SignerVerifier interface {
-	Signer
-	Verifier
-}
-
-// Signer is responsible for creating votes, proposals and QC's for a given block.
+// Signer is responsible for creating votes, proposals for a given block.
 type Signer interface {
```
tarakby, line 8: - I'm a bit confused between this `Signer` and `Voter` under consensus/hotstuff/voter.go. Both seem to be producing votes..
- There in another `Signer` under module/signer.go that we can delete too.

consensus/hotstuff/verification/staking_signer.go
```diff
@@ -23,16 +23,14 @@ type StakingSigner struct {
 
 // NewStakingSigner creates a new combined signer with the given dependencies:
 // - the staking signer is used to create and verify aggregatable signatures for Hotstuff
-// - the signer ID is used as the identity when creating signatures;
 func NewStakingSigner(
 	staking module.Local,
```
tarakby, line 27: minor: we could call this field `me` like we do for other `module.Local` fields in other files.

tarakby (approved):
Nice clean up and substitutions 👏🏼

consensus/hotstuff/signer.go
zhangchiqing: `Voter` uses `Signer` to sign vote, and `Voter` also has voting rule in it. 

>There in another Signer under module/signer.go that we can delete too.

Will be done in another clean up PR

cmd/bootstrap/run/qc.go
```diff
@@ -94,48 +122,29 @@ func createValidators(participantData *ParticipantData, identities flow.Identity
 
 	groupSize := uint(len(participantData.Participants))
 	if groupSize < uint(n) {
```
AlexHentschel, line 124: I think this check is legacy code from when we used to handle DKG participants and staking identities completely separate. By now, we compare the same quantity: 
* https://github.com/onflow/flow-go/blob/dd0239af52df8c4d559266e4b9c389592aef36a0/cmd/bootstrap/run/qc.go#L115
* https://github.com/onflow/flow-go/blob/dd0239af52df8c4d559266e4b9c389592aef36a0/cmd/bootstrap/run/qc.go#L123

### suggestion
I would suggest to combine `createSigners` with `GenerateRootBlockVotes`.  This will consolidate the logic and make it apparent that the check here is unnecessary.
**This is addresses in my PR**

AlexHentschel (commented):
started the review. Suggesting some code consolidations mainly around the bootstrapping logic: 
* see my comments for details 
* I have put up a PR (targeting your branch), that implements most of my suggestions

cmd/bootstrap/cmd/finalize.go
```diff
@@ -466,7 +466,7 @@ func readDKGData() dkg.DKGData {
 
 	dkgData := dkg.DKGData{
 		PrivKeyShares: nil,
-		PubGroupKey:   encodableDKG.GroupKey,
+		PubGroupKey:   encodableDKG.GroupKey.PublicKey,
```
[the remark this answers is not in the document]
durkmurder: I've did this and we absolutely need this since there is a cast somewhere in crypto layer to `(*PubKeyBLSBLS12381)` which is not doable if we are dealing with `encodable.RandomBeaconPubKey`

durkmurder:
@AlexHentschel I was working on bootstrapping changes, so Leo is not the one to blame :)
I have merged your suggestions, thanks, looks good to me.

cmd/bootstrap/cmd/finalize.go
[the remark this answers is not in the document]
durkmurder: https://github.com/onflow/flow-go/blob/feature/consensus-voting/crypto/bls_thresholdsign.go#L167

One way of solving this is to change the check, to check the `Algorithm()`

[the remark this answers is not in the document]
durkmurder: @tarakby

AlexHentschel:
thanks @durkmurder

cmd/bootstrap/cmd/finalize.go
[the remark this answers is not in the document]
tarakby: I see 🤔 
I used to make these checks in the crypto library using `Algorithm()` but I changed them to type assertions for a long reason. (`Algorithm()` works well in the case of threshold signatures though). 

I'll create a ticket to fix these issues  in the crypto lib. 

Meanwhile, is there a kind of recursive type assertion in Go? something like `error.As` for errors ? @durkmurder

consensus/hotstuff/signer.go
tarakby: Thanks Leo!
This is low priority: 
I see the difference, the `Signer` still sees some voting context (block, checks for proposal identity..) so it does a bit more than just signing. The main signing function is `genSigData` [here](https://github.com/onflow/flow-go/blob/leo/replace-signer/consensus/hotstuff/verification/combined_signer_v2.go#L98). Btw, there are 2 structures implementing `Signer`: `CombinedSigner` and `StakingSigner`. Both have exactly the same code, but are only different in `genSigData`.

I think a cleaner design would be :
```
type Signer {
    GenSigData
 }
```
- `Signer` is implemented differently by `CombinedSigner` and `StakingSigner`.
- `Voter` needs an instance of `Signer` to blindly sign for votes. All voting logic is inside `Voter`.
- `BlockProducer` needs an instance of `Signer` to blindly sign for proposals. All proposal logic is inside `BlockProducer`.
-  There is no intermediate interface that has methods `CreateProposal` and `CreateVote`.

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -78,6 +82,38 @@ func TestCombinedSignWithDKGKey(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, true, voteValid)
+
+	// vote on different bock should be invalid
+	block.BlockID[0]++
+	_, err = verifier.VerifyVote(nodeID, vote.SigData, block)
+	require.Error(t, err)
+	block.BlockID[0]--
+
+	// vote with changed view should be invalid
```
AlexHentschel, line 100:
```suggestion
	// vote with a wrong view should be invalid
```

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -78,6 +82,38 @@ func TestCombinedSignWithDKGKey(t *testing.T) {
...
+
+	// vote should be valid
+	vote, err = signer.CreateVote(block)
+	require.NoError(t, err)
+
+	voteValid, err := verifier.VerifyVote(nodeID, vote.SigData, block)
+	require.NoError(t, err)
+	require.Equal(t, true, voteValid)
+
+	// vote on different bock should be invalid
```
AlexHentschel, line 94: For the perspective of the signer, it is important to verify that it generates votes that pass validation. Though, the in the following lines, you verify that the verifier rejects invalid votes. I feel these are tests for the  verifier not the singer. 

_Suggestion:_ move the following lines to `combined_verifier_v2_test.go`

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -124,7 +160,7 @@ func TestCombinedSignWithNoDKGKey(t *testing.T) {
 	committee.On("DKG", mock.Anything).Return(dkg, nil)
 
 	packer := signature.NewConsensusSigDataPacker(committee)
-	verifier := NewCombinedVerifierV2(committee, packer)
+	verifier := NewCombinedVerifier(committee, packer)
```
AlexHentschel, line 163: may I request to correct the godoc below: 
https://github.com/onflow/flow-go/blob/8778e8a90c5ade7b9b30afcf123f65ae769a5050/consensus/hotstuff/verification/combined_signer_v2_test.go#L173
Here, we are explicitly testing the case, where the signer does _not_ have a random beacon key. Therefore, I think the comment should be corrected.

I would suggest to replace: 
https://github.com/onflow/flow-go/blob/8778e8a90c5ade7b9b30afcf123f65ae769a5050/consensus/hotstuff/verification/combined_signer_v2_test.go#L173-L179
by 
```golang
// As the proposer does not have a Random Beacon Key, it should sign soley with its staking key.
// In this case, the SigData should be identical to the staking sig.
expectedStakingSig, err := stakingPriv.Sign(
     MakeVoteMessage(block.View, block.BlockID),
     crypto.NewBLSKMAC(encoding.ConsensusVoteTag),
)
require.NoError(t, err)
require.Equal(t, expectedStakingSig, crypto.Signature(proposal.SigData))
```

AlexHentschel:
we still have these structs: 
* https://github.com/onflow/flow-go/blob/8778e8a90c5ade7b9b30afcf123f65ae769a5050/module/signature/aggregation.go#L328-L337
* https://github.com/onflow/flow-go/blob/8778e8a90c5ade7b9b30afcf123f65ae769a5050/module/verifier.go#L7-L11

will these be removed in a follow-up PR?

consensus/hotstuff/verification/staking_signer.go
AlexHentschel: +1 on [Tarak's comment](https://github.com/onflow/flow-go/pull/1706#discussion_r763409179)

consensus/hotstuff/verification/staking_signer.go
```diff
@@ -23,16 +23,14 @@ type StakingSigner struct {
 
 // NewStakingSigner creates a new combined signer with the given dependencies:
 // - the staking signer is used to create and verify aggregatable signatures for Hotstuff
```
AlexHentschel, line 25: I think this could be updated:
```suggestion
// NewStakingSigner instantiates a StakingSigner, which signs votes and 
// proposals with the staking key.  The generated signatures are aggregatable.
```

AlexHentschel (approved):
nice. 🏄‍♂️ 
all comments are of stylistic nature. No bugs or edge cases that I found (except some error handling, which we should probably address in a follow-up PR: https://github.com/dapperlabs/flow-go/issues/6054)

consensus/hotstuff/verification/combined_signer_v2_test.go
zhangchiqing: I'm using the signer tests to test both modules. 

You are right that we can move this specific unhappy case to verifier, but it's just easy to cover all the happy and unhappy path in one file. And the happy path (verifying a valid vote) has to be in this signer tests, the reason is that we need the signer to create a valid vote. 

So it's just easy to maintain all the cases in one file.

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -78,6 +82,38 @@ func TestCombinedSignWithDKGKey(t *testing.T) {
...
+	require.NoError(t, err)
+
+	voteValid, err := verifier.VerifyVote(nodeID, vote.SigData, block)
+	require.NoError(t, err)
+	require.Equal(t, true, voteValid)
+
+	// vote on different bock should be invalid
+	block.BlockID[0]++
+	_, err = verifier.VerifyVote(nodeID, vote.SigData, block)
+	require.Error(t, err)
```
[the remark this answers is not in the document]
zhangchiqing: Thanks for issuing up. I was planning to search for TODO under the consensus package, and clean them up in a separate PR.

consensus/hotstuff/signer.go
zhangchiqing: This is a really good point. I like your idea. I have [issue it up](https://github.com/dapperlabs/flow-go/issues/6055), as the changes for this refactor might not be big.

consensus/hotstuff/verification/metrics_wrapper.go
zhangchiqing: yes, will do in separate PR

zhangchiqing:
bors merge

zhangchiqing:
bors retry

PR #1708, opened 2021-12-01
REST: Test improvements and handlers fixes

sideninja:
This PR fixes some handler tests and introduces some improvements to the tests by using common handy functions. 
The collection was fixed so it returns light collection by default but if expanded transactions field is provided then it loads and returns full transaction payload. The transaction generated model was fixed so it reflects changes discussed not required response fields.

### DIFF
--- a/engine/access/rest/converter.go
+++ b/engine/access/rest/converter.go
@@ -548,12 +548,14 @@ func serviceEventListResponse(eventList flow.ServiceEventList) []generated.Event
 	return events
 }
 
-func executionResultResponse(exeResult *flow.ExecutionResult) *generated.ExecutionResult {
+func executionResultResponse(exeResult *flow.ExecutionResult, link LinkGenerator) *generated.ExecutionResult {
+	self, _ := selfLink(exeResult.ID(), link.ExecutionResultLink)
+
 	return &generated.ExecutionResult{
 		Id:      exeResult.ID().String(),
 		BlockId: exeResult.BlockID.String(),
 		Events:  serviceEventListResponse(exeResult.ServiceEvents),
-		Links:   nil,
+		Links:   self,
 	}
 }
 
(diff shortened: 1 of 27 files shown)


engine/access/rest/converter.go
```diff
@@ -506,14 +506,31 @@ func idsResponse(ids []flow.Identifier) []string {
...
 
-func collectionResponse(flowCollection *flow.LightCollection, link LinkGenerator) generated.Collection {
-	self, _ := selfLink(flowCollection.ID(), link.CollectionLink)
+func collectionResponse(
+	collection *flow.LightCollection,
+	txs []*flow.TransactionBody,
+	link LinkGenerator,
+	expand map[string]bool,
+) generated.Collection {
+	self, _ := selfLink(collection.ID(), link.CollectionLink)
```
vishalchangrani, line 515: why not throw the error back? like in other xxxResponse() functions?

engine/access/rest/converter.go
```diff
@@ -531,12 +548,14 @@ func serviceEventListResponse(eventList flow.ServiceEventList) []generated.Event
...
-func executionResultResponse(exeResult *flow.ExecutionResult) *generated.ExecutionResult {
+func executionResultResponse(exeResult *flow.ExecutionResult, link LinkGenerator) *generated.ExecutionResult {
+	self, _ := selfLink(exeResult.ID(), link.ExecutionResultLink)
+
 	return &generated.ExecutionResult{
 		Id:      exeResult.ID().String(),
 		BlockId: exeResult.BlockID.String(),
 		Events:  serviceEventListResponse(exeResult.ServiceEvents),
-		Links:   nil,
+		Links:   self,
```
vishalchangrani, line 558: ❤️

engine/access/rest/converter.go
sideninja: Yes, that will be a good improvement, sorry

Issue #1623, opened 2021-11-16
[Access Node] Make missing_collection_blk_count threshold configurable


vishalchangrani:
## Problem Definition

What problem is this solving?

Collections were missing on one of the Access node on Flow mainnet. 

> 
> flow blocks get 19960809 --host flow-mainnet-1.stg.blockchain.circle.com:9000
> 
> :warning:  Version warning: a new version of Flow CLI is available (v0.28.4).
>    Read the installation guide for upgrade instructions: https://docs.onflow.org/flow-cli/install
> 
> 
> Block ID                a2698367250990e153d1ce92f484efbd28259eee37376f035db0f6b3efbac80c
> Parent ID               f921a52607f10bb639e7f5e54d51dc4a021b2d4815181acd1a8a566d0d580ff3
> Timestamp               2021-11-03 06:50:09.567052452 +0000 UTC
> Height                  19960809
> Total Seals             0
> Total Collections       2
>     Collection 0:       42314d97c6e0c3bc591856b9e75bed97c8405a41f7276b9c903fdc28c674000e
>     Collection 1:       23e8abfd90fec9be6f81ff1331af2a7e8143c9f04bc3a0868f5af04c8e1e6069
> 
> 
> flow collections get --host flow-mainnet-1.stg.blockchain.circle.com:9000 42314d97c6e0c3bc591856b9e75bed97c8405a41f7276b9c903fdc28c674000e
> 
> :warning:  Version warning: a new version of Flow CLI is available (v0.28.4).
>    Read the installation guide for upgrade instructions: https://docs.onflow.org/flow-cli/install
> 
> :x: Not Found: not found: please retry for collection in finalized block: could not retrieve collection: key not found

Upon investigation, I did not find any errors in the collection node logs related to that particular access node (node id: `e143aa2c14901ae1db4b212034f34d4da1771fc54c18e35ea84e9403672a3474`).
One possible cause of the collection missing could be that no request was made for the collection as the minimum threshold required to trigger a request for missing collection may not have been met. 

The [threshold value](https://github.com/onflow/flow-go/blob/24ac4e94e70b900b56d7bd6e20573733b8ca80a5/engine/access/ingestion/engine.go#L36) is currently fixed to 100 and not configurable. This means that atleast 100 blocks with missing collections need to be present before a request for missing collection is initiated.

This issue is to:
a. Investigate if the threshold can be lowered.
b. Make the threshold value configurable if it can be lowered.

## Proposed Solution

[...]

PR #1711, opened 2021-12-01
Adding a block height threshold to re-request missing collections 

vishalchangrani:
Fixes: https://github.com/onflow/flow-go/issues/1623

### DIFF
--- a/engine/access/ingestion/engine.go
+++ b/engine/access/ingestion/engine.go
@@ -35,10 +35,14 @@ const fullBlockUpdateInterval = 1 * time.Minute
 // a threshold of number of blocks with missing collections beyond which collections should be re-requested
 const missingCollsForBlkThreshold = 100
 
+// a threshold of block height beyond which collections should be re-requested
+const missingCollsForHeightThreshold = 100
+
 var defaultCollectionCatchupTimeout = collectionCatchupTimeout
 var defaultCollectionCatchupDBPollInterval = collectionCatchupDBPollInterval
 var defaultFullBlockUpdateInterval = fullBlockUpdateInterval
 var defaultMissingCollsForBlkThreshold = missingCollsForBlkThreshold
+var defaultMissingCollsForHeightThreshold = missingCollsForHeightThreshold
 
 // Engine represents the ingestion engine, used to funnel data from other nodes
 // to a centralized location that can be queried by a user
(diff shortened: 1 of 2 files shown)


engine/access/ingestion/engine.go
```diff
@@ -33,12 +33,18 @@ const collectionCatchupDBPollInterval = 10 * time.Millisecond
...
 
+// a threshold of block height beyond which collections should be re-requested (regardless of the number of blocks for which collection are missing)
+// this is to ensure that if a collection is missing for a long time (in terms of block height) it is eventually re-requested
+const missingCollsForHeightThreshold = 100
+
 var defaultCollectionCatchupTimeout = collectionCatchupTimeout
 var defaultCollectionCatchupDBPollInterval = collectionCatchupDBPollInterval
 var defaultFullBlockUpdateInterval = fullBlockUpdateInterval
 var defaultMissingCollsForBlkThreshold = missingCollsForBlkThreshold
+var defaultMissingCollsForHeightThreshold = missingCollsForHeightThreshold
```
jordanschalm, line 47: These seem like they should be config fields on the `Engine` instance rather than global variables

Kay-Zee (approved):

engine/access/ingestion/engine.go
vishalchangrani: so these private fields were meant to help white box testing where they can be modified as part of the unit test. Yeah, I could add it to the engine too and make them configurable (but may do that as a different issue).

engine/access/ingestion/engine.go
```diff
@@ -33,12 +33,18 @@ const collectionCatchupDBPollInterval = 10 * time.Millisecond
 const fullBlockUpdateInterval = 1 * time.Minute
 
 // a threshold of number of blocks with missing collections beyond which collections should be re-requested
+// this is to prevent spamming the collection nodes with request
 const missingCollsForBlkThreshold = 100
 
+// a threshold of block height beyond which collections should be re-requested (regardless of the number of blocks for which collection are missing)
+// this is to ensure that if a collection is missing for a long time (in terms of block height) it is eventually re-requested
+const missingCollsForHeightThreshold = 100
```
smnzhu, line 41: a better name would be something like `missingCollAgeThreshold`

engine/access/ingestion/engine.go
```diff
@@ -569,8 +575,8 @@ func (e *Engine) updateLastFullBlockReceivedIndex() {
 		}
 	}
 
-	// additionally, if more than threshold blocks have missing collection, re-request those collections
-	if incompleteBlksCnt >= defaultMissingCollsForBlkThreshold {
+	// additionally, if more than threshold blocks have missing collection OR collections are missing since defaultMissingCollsForHeightThreshold, re-request those collections
```
smnzhu, line 578:
```suggestion
	// additionally, if more than threshold blocks have missing collection OR collections are missing for more than defaultMissingCollAgeThreshold, re-request those collections
```

engine/access/ingestion/engine_test.go
```diff
@@ -540,15 +540,46 @@ func (suite *Suite) TestUpdateLastFullBlockReceivedIndex() {
 		// assert that missing collections are requested
 		suite.request.AssertExpectations(suite.T())
 
+		// last full blk index is not advanced
+		suite.blocks.AssertExpectations(suite.T()) // no new call to UpdateLastFullBlockHeight should be made
+	})
+
+	suite.Run("missing collections are requested when count exceeds defaultMissingCollsForHeightThreshold", func() {
```
smnzhu, line 547:
```suggestion
	suite.Run("missing collections are requested when age exceeds defaultMissingCollAgeThreshold", func() {
```

engine/access/ingestion/engine.go
vishalchangrani: changed to `missingCollAgeThreshold`

PR #1717, opened 2021-12-02
Smnzhu/injectable libp2p

smnzhu:


cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -102,7 +116,7 @@ func (anb *StakedAccessNodeBuilder) Build() AccessNodeBuilder {
 
 				// register the proxy engine with the unstaked network
 				var err error
-				unstakedNetworkConduit, err = node.Network.Register(engine.PublicSyncCommittee, proxyEngine)
+				unstakedNetworkConduit, err = anb.AccessNodeConfig.PublicNetworkConfig.Network.Register(engine.PublicSyncCommittee, proxyEngine)
```
yhassanzadeh13, line 119: 🟥  **Red Flag**: Current implementation of `splitter.Engine` and `RequestHandlerEngine` is breaking a cardinal principle of Flow's software architecture: 

A [`Conduit`](https://github.com/onflow/flow-go/blob/master/network/conduit.go#L59-L64) is the _exclusive_ interface of an engine for interacting with the network, and each engine _must be_ provided a mutually exclusive conduit for networking interactions. From the perspective of the networking layer, each conduit _must be_ in the sole possession of one and _only one_ engine. Unique conduit(s) per engine is similar to the unique port(s) per each process in operating systems. 

However, the current implementation breaks this software architecture principle of Flow by sharing a conduit between two engines, where `splitter.Engine` reads from it, and `synchronization.RequestHandlerEngine` writes on it. 

Please refactor the architecture to make sure that a conduit is in the sole possession of only a single engine. You may refactor the `synchronization.RequestHandlerEngine` to make the conduit mutually exclusive to it. Then for ingress events, the `synchronization.RequestHandlerEngine` utilizes the `ProcessLocal` method of the  `splitter.Engine` to pass the events to it. In this way, the implementation observes the software architecture principles of Flow, and engines and conduits are utilized consistently with what they are meant for. 

A similar architecture has been implemented with [`requester`](https://github.com/onflow/flow-go/blob/master/engine/verification/requester/requester.go) and [`fetcher` ](https://github.com/onflow/flow-go/blob/master/engine/verification/fetcher/engine.go) engines in verification node, where ingress events to the `requester` are dispatched to `fetcher` through the [`ChunkDataPackHandler`](https://github.com/onflow/flow-go/blob/master/engine/verification/fetcher/requester.go#L20) interface. You may define a similar interface-based interaction between  `splitter.Engine` and `synchronization.RequestHandlerEngine` or go with the `ProcessLocal` option.

yhassanzadeh13 (changes requested):
Review is still in progress, but highlighting a _red flag 🟥 _ (please see comments below).

cmd/access/node_builder/access_node_builder.go
```diff
@@ -102,6 +102,16 @@ type AccessNodeConfig struct {
...
 	baseOptions                  []cmd.Option
+
+	PublicNetworkConfig PublicNetworkConfig
+}
+
+type PublicNetworkConfig struct {
+	// NetworkKey crypto.PublicKey // TODO: do we need a different key for the public network?
+	BindAddress string
+	Network     network.Network
+	Middleware  network.Middleware
```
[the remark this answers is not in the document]
smnzhu: cfda6bc

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -89,6 +95,14 @@ func (builder *StakedAccessNodeBuilder) Initialize() error {
...
 
+func (anb *StakedAccessNodeBuilder) enqueueSplitterNetwork() {
+	anb.Component("splitter network", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+		splitterNet := splitternet.NewNetwork(node.Network, node.Logger)
+		node.Network = splitterNet
+		return splitterNet, nil
+	})
+}
+
 func (anb *StakedAccessNodeBuilder) Build() AccessNodeBuilder {
```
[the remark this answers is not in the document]
smnzhu: cfda6bc

cmd/access/node_builder/staked_access_node_builder.go
smnzhu: A `Conduit` is thread-safe as far as I can tell, so it should be fine to share it among multiple engines. As for in coming messages, it already works the way you describe.

cmd/access/node_builder/access_node_builder.go
```diff
@@ -423,6 +436,13 @@ func (builder *FlowAccessNodeBuilder) extraFlags() {
 		flags.StringSliceVar(&builder.bootstrapNodeAddresses, "bootstrap-node-addresses", defaultConfig.bootstrapNodeAddresses, "the network addresses of the bootstrap access node if this is an unstaked access node e.g. access-001.mainnet.flow.org:9653,access-002.mainnet.flow.org:9653")
 		flags.StringSliceVar(&builder.bootstrapNodePublicKeys, "bootstrap-node-public-keys", defaultConfig.bootstrapNodePublicKeys, "the networking public key of the bootstrap access node if this is an unstaked access node (in the same order as the bootstrap node addresses) e.g. \"d57a5e9c5.....\",\"44ded42d....\"")
 		flags.BoolVar(&builder.supportsUnstakedFollower, "supports-unstaked-node", defaultConfig.supportsUnstakedFollower, "true if this staked access node supports unstaked node")
+		flags.StringVar(&builder.PublicNetworkConfig.BindAddress, "public-network-address", defaultConfig.PublicNetworkConfig.BindAddress, "staked access node's public network bind address")
```
vishalchangrani, line 439: maybe we call this "dht-server-address" or somethig. `public-network-address` seems confusing for the unstaked node (as if it is its own public address which it is not)

crypto/Makefile
```diff
@@ -9,9 +9,9 @@ ADX_SUPPORT := $(shell if ([ -f "/proc/cpuinfo" ] && grep -q -e '^flags.*\badx\b
 test:
 	# test all packages with Relic library enabled
 ifeq ($(ADX_SUPPORT), 1)
-	GO111MODULE=on go test -coverprofile=$(COVER_PROFILE) $(if $(TEST_LONG),,-short) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic $(if $(VERBOSE),-v,) ./...
+	GO111MODULE=on go test -coverprofile=$(COVER_PROFILE) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic $(if $(VERBOSE),-v,) ./...
```
vishalchangrani, line 12: is this change related?

cmd/access/node_builder/staked_access_node_builder.go
yhassanzadeh13: @smnzhu I revised my comment, please read it again. I was not concerned about the thread safety of conduit, rather how it is created and utilized.

crypto/Makefile
smnzhu: No, but it is correct and I just thought I'd include it as part of this PR

cmd/access/node_builder/access_node_builder.go
smnzhu: So this is the bind address of the Staked AN on the public network, hence the name. I don't think the name "dht-server-address" is really accurate?

cmd/access/node_builder/staked_access_node_builder.go
AlexHentschel: I think there is a lot of value in uniform design patterns: engineers familiar with other parts of the code base will recognize such patterns ,which makes the new code easier to understand. Also, constrains and api-level contracts are much easier follow if they are identical all over the code base, thereby reducing probability of bugs and uncovered edge-cases. 
To follow established design patterns is a guideline, and guidelines are not absolute. Therefore, following or deviating from an established design pattern is ultimately the decision of the engineers. But, if you decide to deviate from an established design pattern, please make sure you have a solid reasoning. For example: if following the design pattern would entail that you write 500 lines of extra business-logic code plus tests, that sounds like a valid reason to me. In contrast, "my solution also works" is not a good reason IMHO, because it creates extra cost in the future (for anyone having to understand your custom solution), without any evidence of benefits. 

**I'd like to request that  deviating from the established design pattern is thoroughly document _in code_.** Specifically, please include comments that explain: what the downsides are of following the established pattern in this particular case and the benefits of your alternative approach. Also, please clearly document what engineers have to be cautious about when working with your custom solution. (if there are no extra considerations, please also document this).
If this documentation is more work than following the established pattern, you should most probably follow the existing pattern instead of making up your own. 

In this particular case, it seems to me that @yhassanzadeh13's suggestion wouldn't entail noteworthy overhead (?) Wouldn't it just be that there is one dedicated engine associated with the conduit and all outgoing messages for this conduit would be routed through that engine? 
Also from a conceptual perspective, having one dedicated engine for a conduit seems very intuitive to me. As explained in my [Coding Conventions for Flow](https://github.com/onflow/flow-go/blob/alex/coding-guidelines/docs/Codingconventions.md): 
> On a high level, a Flow node consists of various data-processing components that receive information from node-internal and external sources. We model the data flow inside a Flow node as a graph: a data-processing component forms a vertex; two components exchanging messages is represented as an edge. An Engine is the interface for a data-processing component.

In other words, a conduit wrapped into an engine fits very organically in this data flow graph. But if we start moving  data within a node outside of the engine abstraction, it becomes much harder to comprehend. Sooner or later, we will have to draw the internal data flow graph for each node role. For some nodes, like consensus nodes, this data flow graph is going to be pretty complicated when using engines as vertices alone. Now imagine, we have different types of data flow vertices (engine, conduits, ...), which would add even more complexity to the mental mode. 

I guess the bottom line of my argument is this:
* We are dealing with a system here, that has extremely high intrinsic complexity. Engineers are generally able to understand any small piece of code on isolation. But the challenge for an engineer is generally a wholistic picture of the broader system. 
* Therefore, I think it is most important to keep the conceptual design of the overall system as clean as possible. Specifically, I would encourage you to avoid different types of vertices in the data flow graph, such as engines and conduits.  
* I feel it is a good trade of to add a bit more boilerplate code such that the conceptual design and mathematical abstractions are easily recognizable by engineers. The engineers are not going to be confused by the little extra boilerplate code; but they are going to get completely lost in all the details if they can't identify overarching design patterns. 
* Likewise, a some extra boilerplate code is irrelevant for mathematical correctness or liveness proofs. But a non-uniform conceptual design is going to make such proofs much harder.

smnzhu: Okay, I've updated it to use a different approach in e9bafdb and subsequent commits.

AlexHentschel: 🙇 🖖

network/p2p/dns/resolver.go
```diff
@@ -105,7 +92,7 @@ func (r *Resolver) lookupIPAddr(ctx context.Context, domain string) ([]net.IPAdd
 	}
 
 	if !fresh && r.shouldResolveIP(domain) {
-		r.unit.Launch(func() {
```
yhassanzadeh13, line 108: What is the reason behind replacing `unit` with a `goroutine`? existence of `unit` is crucial to perform a bottom-up graceful shutting down of the node's component. Replacing it with a simple `goroutine` makes an incomplete graceful shutting down. This is specifically important when it comes to the networking layer, since it stops dispatching any message in the network the moment stopping the node triggers in: 
https://github.com/onflow/flow-go/blob/bf146a8e63d2626c3d2533a4f0e95284959c2030/engine/unit.go#L46-L57

🚩  more importantly we are planning to refactor this DNS resolver to a [worker-based](https://github.com/dapperlabs/flow-go/issues/5868) one, which will be entirely relying on a lifecycle management component like `unit` to startup and shutdown the workers.

yhassanzadeh13 (changes requested):
Review is still in progress, but highlighting a few more red flags (please find them with the 🚩  icon). 

Also, please keep my comments unresolved when you apply them so that I can track them. 
Finally, the PR is getting gradually bigger (initially touching 52 files and now 59 files), which makes it a bit harder to review and track. Maybe it is more efficient to extend new features in a separate PR.

network/p2p/dns/resolver.go
smnzhu: This is because currently there's no need for `unit` in the resolver, because it doesn't actually need to be a component with `Ready` and `Done` methods as far as I can tell. Whatever component owns the resolver is responsible for stopping calls to the resolver as soon as it's stopped. If you're refactoring to a worker-based design, I think it would be much better to use `ComponentManager` instead of `unit`.

smnzhu:
Thanks @yhassanzadeh13, I am going to address all of those comments in a follow up PR. I'm going to merge these changes first because they're needed for the state sync engine changes for canary spork.

yhassanzadeh13:
> Thanks @yhassanzadeh13, I am going to address all of those comments in a follow up PR. I'm going to merge these changes first because they're needed for the state sync engine changes for canary spork.

Thanks @smnzhu, maybe it is safer to make this PR against `canary` spork branch than the master.

network/p2p/dns/resolver.go
yhassanzadeh13: Then please refactor it to `ComponentManager` instead of bare `goroutines`, thanks!

smnzhu:
> > Thanks @yhassanzadeh13, I am going to address all of those comments in a follow up PR. I'm going to merge these changes first because they're needed for the state sync engine changes for canary spork.
> 
> Thanks @smnzhu, maybe it is safer to make this PR against `canary` spork branch than the `master`.

Actually since we can do a second canary spork next week, I'll probably just try to make the changes before then

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -272,60 +284,88 @@ func (anb *StakedAccessNodeBuilder) Build() (cmd.Node, error) {
...
+				return head.Height, nil
+			},
+			HotstuffViewFun: func() (uint64, error) {
+				return 0, fmt.Errorf("non-consensus nodes do not report hotstuff view in ping")
+			},
+		}
+
+		pingLibP2PProtocolID := unicast.PingProtocolId(node.SporkID)
+		pingService := p2p.NewPingService(node.Middleware.Host(), pingLibP2PProtocolID, pingProvider, node.Logger)
+
```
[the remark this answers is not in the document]
smnzhu: At first I thought this makes sense, but I just went back and looked at the code now, and I realized the reason ping service is initialized outside of ping engine is because it is needed on its own for non-Access nodes.

The ping engine is used by staked ANs to ping all other nodes, and all other nodes use the PingService to respond to ping requests. So all other node types simply use PingService component without the PingEngine wrapping it.

However, I think it probably does make sense to move the ping provider initialization into the pingservice.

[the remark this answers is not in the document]
smnzhu: fc511f6

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -337,69 +377,57 @@ func (builder *StakedAccessNodeBuilder) enqueueUnstakedNetworkInit() {
...
-			SetLogger(builder.Logger).
-			SetResolver(resolver).
+			SetRoutingSystem(func(ctx context.Context, h host.Host) (routing.Routing, error) {
+				return p2p.NewDHT(ctx, h, unicast.FlowPublicDHTProtocolID(anb.SporkID), p2p.AsServer(true))
+			}).
+			SetPubSub(pubsub.NewGossipSub).
 			Build(ctx)
+
 		if err != nil {
 			return nil, err
```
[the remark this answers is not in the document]
smnzhu: 7f73f4d

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -337,69 +377,57 @@ func (builder *StakedAccessNodeBuilder) enqueueUnstakedNetworkInit() {
...
-
+func (anb *StakedAccessNodeBuilder) initLibP2PFactory(networkKey crypto.PrivateKey) p2p.LibP2PFactoryFunc {
 	return func(ctx context.Context) (*p2p.Node, error) {
-		libp2pNode, err := p2p.NewDefaultLibP2PNodeBuilder(nodeID, myAddr, networkKey).
-			SetSporkID(builder.SporkID).
-			// no connection gater
+		connManager := p2p.NewConnManager(anb.Logger, anb.PublicNetworkConfig.Metrics)
+		resolver := dns.NewResolver(anb.PublicNetworkConfig.Metrics, dns.WithTTL(anb.BaseConfig.DNSCacheTTL))
+
+		node, err := p2p.NewNodeBuilder(anb.Logger, anb.PublicNetworkConfig.BindAddress, networkKey, anb.SporkID).
```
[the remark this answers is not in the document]
smnzhu: 137e116

cmd/access/node_builder/unstaked_access_node_builder.go
```diff
@@ -175,60 +176,47 @@ func (anb *UnstakedAccessNodeBuilder) validateParams() error {
 //		No connection gater
 // 		No connection manager
 // 		Default libp2p pubsub options
-func (builder *UnstakedAccessNodeBuilder) initLibP2PFactory(nodeID flow.Identifier, networkKey crypto.PrivateKey) (p2p.LibP2PFactoryFunc, error) {
+func (builder *UnstakedAccessNodeBuilder) initLibP2PFactory(networkKey crypto.PrivateKey) p2p.LibP2PFactoryFunc {
```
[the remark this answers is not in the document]
smnzhu: 977de9b

cmd/access/node_builder/unstaked_access_node_builder.go
```diff
@@ -175,60 +176,47 @@ func (anb *UnstakedAccessNodeBuilder) validateParams() error {
...
-			// for the unstaked side of the network, the  nodes need to discover each other using DHT Discovery.
-			SetDHTOptions(dhtOptions...).
-			SetLogger(builder.Logger).
-			SetResolver(resolver).
-			SetPubsubOptions(psOpts...).
+		node, err := p2p.NewNodeBuilder(builder.Logger, builder.BaseConfig.BindAddr, networkKey, builder.SporkID).
+			SetBasicResolver(resolver).
+			SetSubscriptionFilter(
+				p2p.NewRoleBasedFilter(
+					0, builder.IdentityProvider,
```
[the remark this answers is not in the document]
smnzhu: 3b6f137

engine/access/ping/engine.go
```diff
@@ -85,39 +93,45 @@ func (e *Engine) Done() <-chan struct{} {
 }
 
 func (e *Engine) startPing() {
-	peers, err := e.state.Final().Identities(filter.Not(filter.HasNodeID(e.me.NodeID())))
-	if err != nil {
-		e.log.Err(err).Msg("could not get identity list")
-		return
-	}
+	pingInterval := time.Minute
```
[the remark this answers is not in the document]
smnzhu: 1c758ea

engine/common/splitter/network/network.go
```diff
@@ -19,32 +21,48 @@ import (
...
-		lm:        lifecycle.NewLifecycleManager(),
 	}
+
+	n.ComponentManager = component.NewComponentManagerBuilder().
+		AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+			<-n.net.Ready()
+
+			ready()
+
+			<-n.net.Done()
```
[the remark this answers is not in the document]
smnzhu: 96d0ecf

engine/common/synchronization/request_handler_engine.go
```diff
@@ -0,0 +1,61 @@
...
+
+	con, err := net.Register(engine.PublicSyncCommittee, e)
+	if err != nil {
+		return nil, fmt.Errorf("could not register engine: %w", err)
+	}
+
+	e.requestHandler = NewRequestHandler(
+		logger,
+		metrics,
+		con.Unicast,
```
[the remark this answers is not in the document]
smnzhu: 77ea5ed

engine/common/synchronization/request_handler.go
```diff
@@ -40,7 +40,7 @@ type RequestHandlerEngine struct {
 	blocks          storage.Blocks
 	core            module.SyncCore
 	finalizedHeader *FinalizedHeaderCache
-	con             network.Conduit // used for sending responses to requesters
+	messageSender   func(interface{}, flow.Identifier) error
```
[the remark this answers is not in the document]
smnzhu: 77ea5ed

network/p2p/dns/resolver.go
smnzhu: f38a0a4

cmd/scaffold.go
```diff
@@ -134,67 +134,44 @@ func (fnb *FlowNodeBuilder) BaseFlags() {
...
-			},
-			HotstuffViewFun: nil, // set in next code block, depending on role
-		}
-
-		// only consensus roles will need to report hotstuff view
-		if fnb.BaseConfig.NodeRole == flow.RoleConsensus.String() {
-			// initialize the persister
-			persist := persister.New(node.DB, node.RootChainID)
-
-			pingProvider.HotstuffViewFun = func() (uint64, error) {
```
peterargue, line 167: is this behavior replicated somewhere?

peterargue (commented):
looks pretty good. I'm not very familiar with this area, but added some comments below.

cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -80,12 +79,16 @@ func (fnb *StakedAccessNodeBuilder) InitIDProviders() {
...
 	builder.InitIDProviders()
 
+	builder.EnqueueResolver()
+
+	// enqueue the regular network
+	builder.EnqueueNetworkInit()
+
 	// if this is an access node that supports unstaked followers, enqueue the unstaked network
 	if builder.supportsUnstakedFollower {
+		builder.enqueueUnstakedResolver()
```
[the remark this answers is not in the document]
smnzhu: different metrics. That's the only reason really, so if we don't care about having separate DNS metrics for private vs public network, we could use the same resolver. 

@yhassanzadeh13

cmd/access/node_builder/access_node_builder.go
```diff
@@ -423,6 +457,13 @@ func (builder *FlowAccessNodeBuilder) extraFlags() {
...
+	}).ValidateFlags(func() error {
+		if builder.supportsUnstakedFollower && (builder.PublicNetworkConfig.BindAddress == cmd.NotSet || builder.PublicNetworkConfig.BindAddress == "") {
```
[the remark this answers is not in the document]
smnzhu: For the staked network the default is NotSet, so i've just done the same here.

cmd/scaffold.go
smnzhu: Yes, it is in `EnqueuePingService`. I just added the consensus node logic now.

network/middleware.go
```diff
@@ -48,17 +48,14 @@ type Middleware interface {
...
-	UpdateAllowList()
-
 	// UpdateNodeAddresses fetches and updates the addresses of all the staked participants
 	// in the Flow protocol.
 	UpdateNodeAddresses()
 
+	RoutingSystem() routing.Routing
+
+	Host() host.Host
+
```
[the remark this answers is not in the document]
yhassanzadeh13: Highlighting this comment

module/metrics/network.go
```diff
@@ -16,128 +16,148 @@ const (
...
+	for _, opt := range opts {
+		opt(nc)
+	}
+
+	nc.outboundMessageSize = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
 			Namespace: namespaceNetwork,
 			Subsystem: subsystemGossip,
-			Name:      "outbound_message_size_bytes",
+			Name:      nc.prefix + "outbound_message_size_bytes",
```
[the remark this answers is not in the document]
yhassanzadeh13: Highlighting this comment.

consensus/integration/network_test.go
```diff
@@ -66,7 +66,7 @@ type Network struct {
 
 // Register registers an Engine of the attached node to the channel via a Conduit, and returns the
 // Conduit instance.
-func (n *Network) Register(channel network.Channel, engine network.Engine) (network.Conduit, error) {
+func (n *Network) Register(channel network.Channel, engine network.MessageProcessor) (network.Conduit, error) {
```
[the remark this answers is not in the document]
yhassanzadeh13: Highlighting this comment.

cmd/access/node_builder/staked_access_node_builder.go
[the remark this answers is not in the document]
yhassanzadeh13: It is fine having the same metrics for public and private networks hence reducing complexity.

network/p2p/dht_test.go
```diff
@@ -28,10 +28,10 @@ func TestFindPeerWithDHT(t *testing.T) {
 	golog.SetAllLoggers(golog.LevelFatal) // change this to Debug if libp2p logs are needed
 
 	sporkId := unittest.IdentifierFixture()
-	dhtServerNodes, _ := nodesFixture(t, ctx, sporkId, 2, withDHTNodeEnabled(true))
+	dhtServerNodes, _ := nodesFixture(t, ctx, sporkId, "dht_test", 2, withDHTOptions(AsServer(true)))
```
yhassanzadeh13, line 31: We may have separate `AsServer()` and `AsClient()` methods, hence no longer need to pass a boolean parameter.

network/p2p/dns/resolver.go
```diff
@@ -104,18 +173,12 @@ func (r *Resolver) lookupIPAddr(ctx context.Context, domain string) ([]net.IPAdd
...
-				}
-			}
-			r.doneResolvingIP(domain)
-		})
+	if !fresh && r.shouldResolveIP(domain) && !util.CheckClosed(r.cm.ShutdownSignal()) {
+		select {
+		case r.ipRequests <- &lookupIPRequest{domain}:
+		default:
+			r.logger.Warn().Str("domain", domain).Msg("IP lookup request queue is full, dropping request")
+		}
```
yhassanzadeh13, line 181: 🚩  This part is needed to get refactored to make sure updating an expired DNS request is not consigned to starvation, i.e., Each time the DNS cache is queried for a domain that is expired, the DNS cache is out of available workers, and hence keeps only returning that expired entry. This can simply happen if the DNS query rate outpaces the number of workers' resolution rate. The previous approach was taking care of this edge case. You may revert to the previous version (i.e., `engine.Unit`-based one) if resolving this edge case is not trivial.

network/p2p/dns/resolver.go
```diff
@@ -158,19 +221,12 @@ func (r *Resolver) lookupTXT(ctx context.Context, txt string) ([]string, error)
...
-				}
-			}
-		})
-
+	if !fresh && r.shouldResolveTXT(txt) && !util.CheckClosed(r.cm.ShutdownSignal()) {
+		select {
+		case r.txtRequests <- &lookupTXTRequest{txt}:
+		default:
+			r.logger.Warn().Str("txt", txt).Msg("TXT lookup request queue is full, dropping request")
+		}
```
yhassanzadeh13, line 229: 🚩  This part is needed to get refactored to make sure updating an expired DNS request is not consigned to starvation, i.e., Each time the DNS cache is queried for a domain that is expired, the DNS cache is out of available workers, and hence keeps only returning that expired entry. This can simply happen if the DNS query rate outpaces the number of workers' resolution rate. The previous approach was taking care of this edge case. You may revert to the previous version (i.e., `engine.Unit`-based one) if resolving this edge case is not trivial.

network/p2p/dns/resolver_test.go
```diff
@@ -23,9 +25,13 @@ const happyPath = true
 // instead of going through the underlying basic resolver, and hence through the network.
 func TestResolver_HappyPath(t *testing.T) {
 	basicResolver := mocknetwork.BasicResolver{}
-	resolver := NewResolver(metrics.NewNoopCollector(), WithBasicResolver(&basicResolver))
+	resolver := NewResolver(zerolog.Nop(), metrics.NewNoopCollector(), WithBasicResolver(&basicResolver))
```
yhassanzadeh13, line 28: Please use `unittest.Logger`. We use logs to investigate failure/flakiness on CI.

yhassanzadeh13 (changes requested):
Thanks for applying the changes. The PR is shaping out nicely. I resolved comments that have been addressed. Though there are still major comments that need to be addressed. Please look at the unresolved comments, especially those with a 🚩  flag, and the ones I highlighted again. Thanks!

network/p2p/dns/resolver.go
smnzhu: The previous approach takes care of this edge case because it was not worker-based, so in theory there could be an infinite number of go-routines spawned. In a worker-based design, it is impossible to avoid this issue, because you have a finite number of workers and you cannot have a queue of infinite length. Dropping requests when a queue is full is the standard approach in scenarios like this. The alternative solution is to block callers when the queue is full (ie backpressure), but in this scenario I don't think that's a better approach.

consensus/integration/network_test.go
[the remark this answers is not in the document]
smnzhu: Nothing, the inter-engine communication will work as before. The changes are only in the network, and the network never needs to call an engine's ProcessLocal.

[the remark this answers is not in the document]
smnzhu: Currently, we are passing `network.Engine` interface when we register an engine, but this is not necessary. The only method that actually gets called by network is `Process`.

network/p2p/dns/resolver.go
yhassanzadeh13: Correct, then I would suggest reverting it to the previous version that was using goroutines through `unit.Launch`. Running a `goroutine` per _expired_ (and not all) DNS requests is still an efficient approach considering our system scale. The benchmarking results also do not show any `goroutine` overheads, hence the infinite number of spawned `goroutines` is still not a concern.

The current worker-based version poses a starvation concern and such issues would be extremely hard to fix in production. To be correctly functional, we need an intermediate mempool that sits between the resolver and workers. So that DNS requests are written to the mempool, then the resolver takes requests from mempools and spins up workers. Hence, no DNS request is ever facing starvation. This is a missing functionality in the current approach and breaks the liveness of the code, i.e., there is a likelihood that an expired DNS request remains expired for a long while. We have this worker-based DNS cache issued up [here](https://github.com/dapperlabs/flow-go/issues/5868) to be architected and implemented later.

smnzhu: I want to avoid using `unit` since this is something being deprecated, so new code should not use it.

For now, using a large-enough channel buffer size and blocking callers in case the channel gets full should be good enough for the current system scale, and this can be easily updated later by replacing the channel with a proper mempool implementation. This way we can do a chunk of the worker-based implementation for https://github.com/dapperlabs/flow-go/issues/5868, and leave the remaining work for later.

smnzhu: Or instead of blocking callers, we can add a new metric to track how often the request queue really gets full and requests get dropped, and we can track this on production. My guess is that with a sufficiently large buffer size it will never happen

yhassanzadeh13: > Or instead of blocking callers, we can add a new metric to track how often the request queue really gets full and requests get dropped, and we can track this on production. My guess is that with a sufficiently large buffer size it will never happen

Please go with this one, and I will expedite https://github.com/dapperlabs/flow-go/issues/5868 to implement the buffer. Also, please make sure that this new metric is scrapped on all of our networks.

cmd/access/node_builder/staked_access_node_builder.go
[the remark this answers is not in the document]
smnzhu: c1f6f01

integration/testnet/network.go
```diff
@@ -782,9 +782,11 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
 
 			if nodeConf.SupportsUnstakedNodes {
 				hostExternalNetworkPort := testingdock.RandomPort(t)
-				nodeContainer.bindPort(hostExternalNetworkPort, fmt.Sprintf("%s/tcp", strconv.Itoa(DefaultFlowPort)))
+				containerExternalNetworkPort := "9876/tcp"
```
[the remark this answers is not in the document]
smnzhu: c7bdcd1

module/metrics/network.go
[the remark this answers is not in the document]
smnzhu: Thanks, this is good catch. 

I think since staked / unstaked network are using two completely separate libp2p hosts, it does make sense to have them as completely separate metrics. Furthermore, when @vishalchangrani and I originally talked about adding these unstaked metrics a while ago, we considered using tags, but in the end we decided against it because there were some downsides to using tags which I don't exactly remember now. @vishalchangrani?

2d60610

network/p2p/dht_test.go
smnzhu: 946c4ae

network/p2p/dns/resolver_test.go
smnzhu: 946c4ae

smnzhu: 67f581e

network/p2p/dns/resolver.go
smnzhu: 166aef6

network/middleware.go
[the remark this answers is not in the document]
smnzhu: 867d63c

fdc39c7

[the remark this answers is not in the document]
smnzhu: I did not move the `network.RegisterPingService` call into the Ping Engine, because as mentioned before:
* Every node needs to register a PingService to handle incoming pings. In other words, every node needs to call `network.RegisterPingService`
* Only Access nodes need to start a PingEngine to send outgoing pings. 

Thus, I think it's simpler to initialize the PingService in a single place for all node types, and then Access Nodes can pass this to the PingEngine.

yhassanzadeh13 (approved):
LGTM, thanks for applying comments.

vishalchangrani (approved):
lgtm

PR #1718, opened 2021-12-02
wip

tarakby:
In DKG fails locally, the output private key is `nil`. There are two issues with accessing `a nil` pointer in that case:
- [Checking key consistency](https://github.com/onflow/flow-go/blob/792ef25bb27d24cbc112459d35f1ebb31cd9d5d4/engine/consensus/dkg/reactor_engine.go#L187) by computing the public key.  
- `nil` is [wrapped](https://github.com/onflow/flow-go/blob/792ef25bb27d24cbc112459d35f1ebb31cd9d5d4/engine/consensus/dkg/reactor_engine.go#L287) and stored in the key stored, and used to later to sign as it's not caught by this [check](https://github.com/onflow/flow-go/blob/792ef25bb27d24cbc112459d35f1ebb31cd9d5d4/module/signature/threshold.go#L80).

This PR solves these 2 issues:
 - if the DKG output is `nil`, `nil` is stored in the key store instead of a wrapped `nil` (this also solves an issue in V2)
 - when checking the key consistency, do not compute the public key of a `nil` private key.

### DIFF
--- a/engine/consensus/dkg/reactor_engine.go
+++ b/engine/consensus/dkg/reactor_engine.go
@@ -172,9 +172,14 @@ func (e *ReactorEngine) EpochCommittedPhaseStarted(currentEpochCounter uint64, f
 		return
 	}
 
-	dkgPrivInfo, err := e.keyStorage.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
+	beaconPrivateKey, err := e.keyStorage.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
 	if err != nil {
 		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG private info for next epoch")
+		return
+	}
+
+	if beaconPrivateKey == nil {
+		e.log.Warn().Msg("locally computed private key for next epoch is nil")
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


engine/consensus/dkg/reactor_engine.go
```diff
@@ -168,26 +168,31 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
-		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG private info for next epoch")
+		e.log.Err(err).Msg("checking random beacon key consistency: could not retrieve beacon private key for next epoch")
 		return
 	}
 
-	nextDKGPubKey, err := nextDKG.KeyShare(e.me.NodeID())
+	if beaconPrivateKey == nil {
+		e.log.Warn().Msg("locally computed private key for next epoch is nil")
+		return
+	}
```
jordanschalm, line 184: For this, disallowed storing a nil key (instead using DKG end state)

engine/consensus/dkg/reactor_engine.go
```diff
@@ -283,10 +288,16 @@ func (e *ReactorEngine) end(epochCounter uint64) func() error {
...
-		privKeyInfo := encodable.RandomBeaconPrivKey{
-			PrivateKey: privateShare,
+		var beaconPrivateKey encodable.RandomBeaconPrivKey
+		// if privateShare (DKG output) is nil, store nil in the key storage,
+		// otherwise, wrap it as a RandomBeaconPrivKey instance and store it in the key storage.
+		if privateShare != nil {
+			beaconPrivateKey = encodable.RandomBeaconPrivKey{
+				PrivateKey: privateShare,
+			}
 		}
```
jordanschalm, line 298: Changed this to not store a nil beacon key at all (instead use dkg end state)

tarakby:
Closing this PR as the points are being addressed by https://github.com/onflow/flow-go/pull/1732

PR #1720, opened 2021-12-02
DKG Storage Pt. 1: Update storage layer

jordanschalm:
This PR is Part I of an update to DKG State storage.
* refactors DKG key storage to store start and end states
* adds the `SafeBeaconPrivateKeys` interface which provides a safe way for signing components to access beacon keys
* updates storage-layer tests

Pt II (https://github.com/onflow/flow-go/pull/1721) updates all components depending on the modified storage components.
Pt III (https://github.com/onflow/flow-go/pull/1732) updates the beacon key checking to occur atomically with the transition to `EpochCommitted` phase.

PR #1721, opened 2021-12-02
DKG Storage Pt II: Update components using `DKGState` storage

jordanschalm:
This PR is Pt. II of an update to DKG state storage.
* updates components using DKG state storage to use `DKGState` and `SafeBeaconKeys`

⚠️ Depends on #1720

Pt III (https://github.com/onflow/flow-go/pull/1732) updates the beacon key checking to occur atomically with the transition to `EpochCommitted` phase.

### DIFF
--- a/cmd/consensus/main.go
+++ b/cmd/consensus/main.go
@@ -274,7 +274,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-			err = dkgState.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey)
+			err = dkgState.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey.PrivateKey)
 			if err != nil && !errors.Is(err, storage.ErrAlreadyExists) {
 				return err
 			}
--- a/cmd/consensus/main.go
+++ b/cmd/consensus/main.go
@@ -254,8 +254,8 @@ func main() {
 
 			// if the node is not part of the current epoch identities, we do
 			// not need to load the key
-			epoch := node.State.AtBlockID(node.RootBlock.ID()).Epochs().Current()
-			initialIdentities, err := epoch.InitialIdentities()
+			rootEpoch := node.State.AtBlockID(node.RootBlock.ID()).Epochs().Current()
+			initialIdentities, err := rootEpoch.InitialIdentities()
 			if err != nil {
 				return err
 			}
@@ -270,7 +270,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-			epochCounter, err := epoch.Counter()
+			epochCounter, err := rootEpoch.Counter()
 			if err != nil {
 				return err
 			}
@@ -288,12 +288,12 @@ func main() {
 			// participant in the epoch and we don't have the corresponding DKG
 			// key in the database.
 			checkEpochKey := func(protocol.Epoch) error {
-				identities, err := epoch.InitialIdentities()
+				identities, err := rootEpoch.InitialIdentities()
 				if err != nil {
 					return err
 				}
 				if _, ok := identities.ByNodeID(node.NodeID); ok {
-					counter, err := epoch.Counter()
+					counter, err := rootEpoch.Counter()
 					if err != nil {
 						return err
 					}
--- a/cmd/consensus/main.go
+++ b/cmd/consensus/main.go
@@ -115,7 +115,8 @@ func main() {
 		dkgBrokerTunnel         *dkgmodule.BrokerTunnel
 		blockTimer              protocol.BlockTimer
 		finalizedHeader         *synceng.FinalizedHeaderCache
-		dkgKeyStore             *bstorage.BeaconPrivateKeys
+		dkgState                *bstorage.DKGState
+		safeBeaconKeys          *bstorage.SafeBeaconPrivateKeys
 	)
 
 	nodeBuilder := cmd.FlowNode(flow.RoleConsensus.String())
@@ -169,8 +170,12 @@ func main() {
 			conMetrics = metrics.NewConsensusCollector(node.Tracer, node.MetricsRegisterer)
 			return nil
 		}).
-		Module("dkg key storage", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
-			dkgKeyStore, err = bstorage.NewBeaconPrivateKeys(node.Metrics.Cache, node.SecretsDB)
+		Module("dkg state", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
+			dkgState, err = bstorage.NewDKGState(node.Metrics.Cache, node.SecretsDB)
+			return err
+		}).
+		Module("beacon keys", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
+			safeBeaconKeys, err = bstorage.NewSafeBeaconPrivateKeys(dkgState)
 			return err
 		}).
 		Module("mutable follower state", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
@@ -269,7 +274,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-			err = dkgKeyStore.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey)
+			err = dkgState.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey)
 			if err != nil && !errors.Is(err, storage.ErrAlreadyExists) {
 				return err
 			}
@@ -287,7 +292,7 @@ func main() {
 					if err != nil {
 						return err
 					}
-					_, err = dkgKeyStore.RetrieveMyBeaconPrivateKey(counter)
+					_, err = dkgState.RetrieveMyBeaconPrivateKey(counter)
 					if err != nil {
 						return err
 					}
@@ -604,7 +609,7 @@ func main() {
 
 			epochLookup := epochs.NewEpochLookup(node.State)
 
-			thresholdSignerStore := signature.NewEpochAwareSignerStore(epochLookup, dkgKeyStore)
+			thresholdSignerStore := signature.NewEpochAwareSignerStore(epochLookup, safeBeaconKeys)
 
 			// initialize the combined signer for hotstuff
 			var signer hotstuff.SignerVerifier
@@ -743,7 +748,7 @@ func main() {
 				node.Logger,
 				node.Me,
 				node.State,
-				dkgKeyStore,
+				dkgState,
 				dkgmodule.NewControllerFactory(
 					node.Logger,
 					node.Me,
--- a/cmd/consensus/main.go
+++ b/cmd/consensus/main.go
@@ -278,6 +278,11 @@ func main() {
 			if err != nil && !errors.Is(err, storage.ErrAlreadyExists) {
 				return err
 			}
+			// mark the root DKG as successful, so it is considered safe
+			err = dkgState.SetDKGEndState(epochCounter, flow.DKGEndStateSuccess)
+			if err != nil && !errors.Is(err, storage.ErrAlreadyExists) {
+				return err
+			}
 
 			// Given an epoch, checkEpochKey returns an error if we are a
 			// participant in the epoch and we don't have the corresponding DKG
--- a/engine/consensus/dkg/reactor_engine.go
+++ b/engine/consensus/dkg/reactor_engine.go
@@ -2,6 +2,7 @@ package dkg
 
 import (
 	"crypto/rand"
+	"errors"
 	"fmt"
 
 	"github.com/rs/zerolog"
@@ -41,7 +42,7 @@ type ReactorEngine struct {
 	log               zerolog.Logger
 	me                module.Local
 	State             protocol.State
-	keyStorage        storage.BeaconPrivateKeys
+	dkgState          storage.DKGState
 	controller        module.DKGController
 	controllerFactory module.DKGControllerFactory
 	viewEvents        events.Views
@@ -53,7 +54,7 @@ func NewReactorEngine(
 	log zerolog.Logger,
 	me module.Local,
 	state protocol.State,
-	keyStorage storage.BeaconPrivateKeys,
+	dkgState storage.DKGState,
 	controllerFactory module.DKGControllerFactory,
 	viewEvents events.Views,
 ) *ReactorEngine {
@@ -67,7 +68,7 @@ func NewReactorEngine(
 		log:               logger,
 		me:                me,
 		State:             state,
-		keyStorage:        keyStorage,
+		dkgState:          dkgState,
 		controllerFactory: controllerFactory,
 		viewEvents:        viewEvents,
 		pollStep:          DefaultPollStep,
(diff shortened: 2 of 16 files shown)


engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
+		e.log.Fatal().Err(err).Msg("checking DKG key consistency: could not retrieve my DKG public key for next epoch")
 		return
 	}
 
-	localPubKey := dkgPrivInfo.PublicKey()
+	localPubKey := myBeaconPrivKey.PublicKey()
 
 	if !nextDKGPubKey.Equals(localPubKey) {
 		e.log.Warn().Msg("checking DKG key consistency: locally computed dkg public key does not match dkg public key for next epoch")
+		err := e.dkgState.SetDKGEndState(dkgEpochCounter, flow.DKGEndStateInconsistentKey)
```
zhangchiqing, line 278: what ensures this is called only once? 

because otherwise it will hit already exist error, and crash

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
+		if err != nil {
+			e.log.Fatal().Err(err).Msg("failed to set dkg end state")
```
zhangchiqing, line 280: Better to add current epoch in the error message, same for other errors in this function

```suggestion
			e.log.Fatal().Uint64("cur_epoch", currentEpochCounter).Err(err).Msg("failed to set dkg end state")
```

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
-		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG public key for next epoch")
+		e.log.Fatal().Err(err).Msg("checking DKG key consistency: could not retrieve my DKG public key for next epoch")
 		return
 	}
 
-	localPubKey := dkgPrivInfo.PublicKey()
+	localPubKey := myBeaconPrivKey.PublicKey()
 
 	if !nextDKGPubKey.Equals(localPubKey) {
 		e.log.Warn().Msg("checking DKG key consistency: locally computed dkg public key does not match dkg public key for next epoch")
```
zhangchiqing, line 277: better to print both public key for comparison

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
+// that our locally computed key share is consistent with the canonical key vector.
+//
+// If our key is inconsistent, we log a warning message.
+// CAUTION: we will still sign using the inconsistent key, to be fixed with Crypto
+// v2 by falling back to staking signatures in this case.
+//
+func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uint64) {
+
+	// the epoch counter the DKG just completed is for
+	dkgEpochCounter := currentEpochCounter + 1
```
zhangchiqing, line 242: better to create a logger, include both epoch counter value in the fields, and use the logger to log success case or failure case

engine/consensus/dkg/reactor_engine.go
```diff
@@ -87,49 +119,85 @@ func (e *ReactorEngine) Done() <-chan struct{} {
...
+//
+// It starts a new controller for the epoch and registers the triggers to regularly
+// query the DKG smart-contract and transition between phases at the specified views.
+//
+func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow.Header) {
+
+	firstID := first.ID()
+	nextEpochCounter := currentEpochCounter + 1
+	log := e.log.With().
 		Uint64("current_epoch", currentEpochCounter).
```
zhangchiqing, line 146: better to include the `nextEpochCounter` in the log

engine/consensus/dkg/reactor_engine.go
```diff
@@ -273,20 +364,22 @@ func (e *ReactorEngine) registerPhaseTransition(view uint64, fromState dkgmodule
 func (e *ReactorEngine) end(epochCounter uint64) func() error {
 	return func() error {
 		err := e.controller.End()
-		if err != nil {
-			if crypto.IsDKGFailureError(err) {
-				e.log.Warn().Msgf("node %s with index %d failed DKG locally: %s", e.me.NodeID(), e.controller.GetIndex(), err.Error())
-			} else {
-				return err
+		if crypto.IsDKGFailureError(err) {
```
tarakby, line 367: Just making sure that `crypto.IsDKGFailureError(nil)` returns `false` ?

engine/consensus/dkg/reactor_engine.go
```diff
@@ -273,20 +364,22 @@ func (e *ReactorEngine) registerPhaseTransition(view uint64, fromState dkgmodule
...
 			}
+		} else if err != nil {
+			return fmt.Errorf("unknown error ending the dkg: %w", err)
 		}
 
 		privateShare, _, _ := e.controller.GetArtifacts()
 
 		privKeyInfo := encodable.RandomBeaconPrivKey{
 			PrivateKey: privateShare,
 		}
```
tarakby, line 381: I think we don't have to wrap the key here as it's done already inside `InsertMyBeaconPrivateKey`

module/signature/signer_store.go
```diff
@@ -40,16 +39,24 @@ func (s *EpochAwareSignerStore) GetThresholdSigner(view uint64) (module.Threshol
...
 	}
 
-	privDKGData, err := s.keys.RetrieveMyBeaconPrivateKey(epoch)
-	if errors.Is(err, storage.ErrNotFound) {
-		signer = NewThresholdProvider(encoding.RandomBeaconTag, nil)
-	} else if err != nil {
+	privDKGData, safe, err := s.keys.RetrieveMyBeaconPrivateKey(epoch)
+	if err != nil {
+		// there are no expected errors here
 		return nil, fmt.Errorf("could not retrieve DKG private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
```
tarakby, line 45:
```suggestion
		return nil, fmt.Errorf("could not retrieve beacon private key for epoch counter: %v, at view: %v, err: %w", epoch, view, err)
```

module/signature/signer_store.go
```diff
@@ -40,16 +39,24 @@ func (s *EpochAwareSignerStore) GetThresholdSigner(view uint64) (module.Threshol
...
+	}
+
+	if !safe {
+		// we do not have a consistent beacon key which is safe for signing
+		// CAUTION: for now, we produce explicitly invalid signatures
+		// TODO: in Crypto V2, we will fallback to using staking signatures
+		signer = NewThresholdProvider(encoding.RandomBeaconTag, nil)
 	} else {
+		// we have a beacon key which has been explicitly marked safe for use
+		// by the DKG process
```
tarakby, line 55: The way `RetrieveMyBeaconPrivateKey` is written, `safe` is `false` iff `privDKGData` is `nil`. This means we can just call `signer = NewThresholdProvider(encoding.RandomBeaconTag, privDKGData)` in all cases.

tarakby: This also means the boolean return of `RetrieveMyBeaconPrivateKey` is not very useful. The safety check is done already within `RetrieveMyBeaconPrivateKey`.

tarakby (approved):
🏄🏼  
This looks fine but there are still a few points I would like to check in the future PR.

engine/consensus/dkg/reactor_engine.go
jordanschalm: 👍  it does -> https://go.dev/play/p/6nHREXi5R0O

cmd/consensus/main.go
```diff
@@ -269,7 +274,12 @@ func main() {
 			if err != nil {
 				return err
 			}
-			err = dkgKeyStore.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey)
+			err = dkgState.InsertMyBeaconPrivateKey(epochCounter, beaconPrivateKey)
```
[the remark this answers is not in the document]
jordanschalm: We don't use the current epoch, we use the epoch associated with the root block:
https://github.com/onflow/flow-go/blob/aa1fc973a9f6f6f0b5a3167c20b7f9dc326c8238/cmd/consensus/main.go#L257

I'll rename the variable to `rootEpoch` to hopefully that make that clearer: https://github.com/onflow/flow-go/pull/1721/commits/6931854d7ca00e67b034fc0bebcdfe1693551d68

engine/consensus/dkg/reactor_engine.go
jordanschalm: I was planning to address that in the next PR with these two TODOs: https://github.com/onflow/flow-go/blob/3c2f3f35b4970973d128cc95beed66981afee3d4/engine/consensus/dkg/reactor_engine.go#L244-L248

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/b5e69b4507024026c63471c93ecd41229b4f5365

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/b5e69b4507024026c63471c93ecd41229b4f5365

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
+		e.log.Fatal().Err(err).Msg("checking DKG key consistency: could not retrieve next DKG info")
 		return
 	}
 
-	dkgPrivInfo, err := e.keyStorage.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
-	if err != nil {
-		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG private info for next epoch")
+	myBeaconPrivKey, err := e.dkgState.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
+	if errors.Is(err, storage.ErrNotFound) {
+		e.log.Warn().Msg("checking DKG key consistency: no key found")
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/b5e69b4507024026c63471c93ecd41229b4f5365

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/b5e69b4507024026c63471c93ecd41229b4f5365

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/7c52c9c3618d2f890edae4cb9125dcb203201150

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
+		err := e.dkgState.SetDKGEndState(dkgEpochCounter, flow.DKGEndStateNoKey)
```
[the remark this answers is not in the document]
jordanschalm: Definitely, was planning to address that in the next PR, since that storage read would also need to make use of the badger transaction context we're passing in.

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,33 +229,56 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
 
-	localPubKey := dkgPrivInfo.PublicKey()
+	localPubKey := myBeaconPrivKey.PublicKey()
 
 	if !nextDKGPubKey.Equals(localPubKey) {
 		e.log.Warn().Msg("checking DKG key consistency: locally computed dkg public key does not match dkg public key for next epoch")
+		err := e.dkgState.SetDKGEndState(dkgEpochCounter, flow.DKGEndStateInconsistentKey)
+		if err != nil {
+			e.log.Fatal().Err(err).Msg("failed to set dkg end state")
+		}
```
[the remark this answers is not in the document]
jordanschalm: Good idea, added here https://github.com/onflow/flow-go/pull/1721/commits/b4d0830ebe8bd1e93a0f51ba4cbc4354949c76f3

module/metrics/labels.go
```diff
@@ -78,7 +78,7 @@ const (
 
 	ResourceClusterBlockProposalQueue = "cluster_compliance_proposal_queue" // collection node, compliance engine
 	ResourceClusterBlockVoteQueue     = "cluster_compliance_vote_queue"     // collection node, compliance engine
-	ResourceDKGKey                    = "dkg-key"                           // consensus node, DKG engine
+	ResourceBeaconKey                 = "dkg-key"                           // consensus node, DKG engine
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/1ba670bf92ad9fd63b642ce1e4eeaacd3b057a8a

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/5f62cb3b0f49c59b2118a4387a621b74a5516f03

module/signature/signer_store.go
jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/868204233bf21d7c0c324ef1650d0262ec5eb9a2

module/signature/signer_store.go
jordanschalm: That happens to be true here (we could remove the conditional and behaviour would be the same), but that won't be true when this is incorporated into Crypto V2.

It would be possible to remove `safe` and say that `(nil, nil)` means we don't have a safe key, but this breaks with common convention -- in particular, if a function returns `(value, error)` and the caller observes `error==nil`, 99% of the time that means the function returned a valid `value` that can be safely used.

engine/consensus/dkg/reactor_engine.go
```diff
@@ -161,34 +229,67 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
...
-		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve next DKG info")
+		// CAUTION: this should never happen, indicates a storage failure or corruption
+		log.Fatal().Err(err).Msg("checking beacon key consistency: could not retrieve next DKG info")
 		return
 	}
 
-	dkgPrivInfo, err := e.keyStorage.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
-	if err != nil {
-		e.log.Err(err).Msg("checking DKG key consistency: could not retrieve DKG private info for next epoch")
+	myBeaconPrivKey, err := e.dkgState.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
```
zhangchiqing, line 261: Is it possible this method would return `(nil, nil)`? 

If DKG was completed, but a node has no key generated, would it store `nil` as beacon private key, if so, [myBeaconPrivKey.PublicKey()](https://github.com/onflow/flow-go/pull/1721/files#diff-97d28bb0b296d39ca2f3c1a1c2cf3c34db65aa0e992efbadc14f0148dd764fe7R279) would panic.

jordanschalm: Good point - @tarakby is it possible for this line to return `nil`?

> I'll update `DKGState` to explicitly disallow that case anyway.

tarakby: Yes that's possible, I noticed that case but I'm waiting for @jordanschalm's third DKG PR to look at everything together. 

It is the point I addressed in this PR https://github.com/onflow/flow-go/pull/1718 and Jordan mentioned he will pick it up into his upcoming PR.

jordanschalm: https://github.com/onflow/flow-go/pull/1721/commits/a87622afa1e7f96c0a4cedbdd79e13e4392a5480

engine/consensus/dkg/reactor_engine.go
```diff
@@ -386,10 +387,21 @@ func (e *ReactorEngine) end(epochCounter uint64) func() error {
...
-		err = e.dkgState.InsertMyBeaconPrivateKey(epochCounter, privateShare)
-		if err != nil {
-			return fmt.Errorf("couldn't save DKG private key in db: %w", err)
+		if privateShare == nil {
+			// we failed to compute a private key - set the corresponding DKG failure state
+			e.log.Warn().Msg("computed nil private key share")
+			err = e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateNoKey)
+			if err != nil {
+				return fmt.Errorf("with nil key could not set dkg end state to %s: %w", flow.DKGEndStateNoKey, err)
+			}
```
tarakby, line 396: We get to this case only when there was a `crypto.IsDKGFailureError(err)` in line 378. This means there is no need to update the end state with `DKGEndStateNoKey`.

jordanschalm: Addressing this in https://github.com/onflow/flow-go/pull/1732

jordanschalm:
@zhangchiqing @kc1116 @tarakby this PR is ready for another look, and the follow-up is ready as well https://github.com/onflow/flow-go/pull/1732

zhangchiqing (approved):
Looks good

PR #1724, opened 2021-12-02
Remove metering computation twice

janezpodhostnik:
Found a bug while doing some measurements. Computation was being added twice. This is also probably why some transactions were now costing more.

I also added a test for it. Previous test did not detect it as you had to have a non 0 computation to start with, to see this bug.

It looks like we did not hit this bug yet, as cadence only calls `AddUsed` only once per transaction.


fvm/handler/computation.go
```diff
@@ -101,7 +101,6 @@ func (c *computationMeteringHandler) Limit() uint64 {
 }
 
 func (c *computationMeteringHandler) AddUsed(used uint64) error {
-	used = c.computation.Used() + used
 	return c.computation.AddUsed(used)
 }
```
ramtinms, line 105: nice catch but not a big issue, since this is mostly for reporting to users, cadence tracks and enforces it own computation used and calls this one once at the end of transaction execution

ramtinms (approved):

turbolent (approved):
Good catch!

janezpodhostnik:
bors merge

PR #1725, opened 2021-12-02
Fixed pruning for VoteAggregator and misc clean ups 

AlexHentschel:
* fixed VoteAggregator.PruneUpToView to be consistent with existing convention
* removed a "V2" designators in tests for components for which we already removed the V1 version


durkmurder (approved):
Nice and clean, thank you!

consensus/hotstuff/voteaggregator/vote_aggregator.go
```diff
@@ -211,11 +211,11 @@ func (va *VoteAggregator) InvalidBlock(proposal *model.Proposal) error {
...
 
-// PruneUpToView will delete all votes below to the given view, as well as related indexes.
+// PruneUpToView deletes all votes _below_ to the given view, as well as related
+// indices. If `view` is smaller than the previous value, the previous value is
+// kept and the method call is a NoOp.
 func (va *VoteAggregator) PruneUpToView(view uint64) {
-	// if someone else has updated view in parallel don't bother doing extra work for cleaning, whoever
-	// is able to advance counter will perform the cleanup
-	if va.highestPrunedView.Set(view) {
+	if va.lowestView.Set(view) {
```
zhangchiqing, line 218: If we prune up to view `100`, then any votes for view 100 should be pruned (removed), therefore, the lowest view would become `101`, rather than `100`, because 100 has been pruned. In that case `highestPrunedView` sounds more correct

AlexHentschel: For a given `view` number, there are two possible conventions:
1.  prune all views `v` < `view`, i.e. I _exclude_ `view` in the pruning 
2. prune all views `v` ≤ `view`, i.e. I _include_ `view` in the pruning

From a theoretical algorithmic perspective, it doesn't make much of a difference (they are equally powerful). But from an implementation perspective, option 1 creates less edge cases (details below). 

So far, we use convention 1 throughout the code base (except in this particular case). I think there is value in having one uniform convention; hence I am suggesting to change the implementation here to be consistent with the all other pruning implementations. 

#### Motivation for choosing convention 1 over convention 2

Lets look at the levelled forest, which we use for storing blocks. In this particular case (and in multiple other applications in the code base), we actually need to keep the latest finalized block.   In other words, when we prune, we only want to remove the blocks with height _strictly lower_ than the finalized block. 

1. The current code implements convention 1: https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/module/forest/leveled_forrest.go#L51-L52 for the following reason:
   * I can always just write 
      ```golang
      PruneUpToLevel(latestFinalizedBlock.Height)
      ```
       which will remove all blocks _below_ the `latestFinalizedHeight` but keep the blocks at this height 
2. Now lets look at how the code would look like, if we were to use convention 2:
    ```golang
    PruneUpToLevel(latestFinalizedBlock.Height-1)
    ```
    but wait: 
    * the height is a `uint64` and if the node boots up with the genesis block at height `0` as `latestFinalizedBlock`, the operation  `latestFinalizedBlock.Height-1` would underflow. I have to worry about this every time I prune ... which makes the API quite error prone.

More generally, the threshold value  of `0` is a common case for blockchains (i.e. `view=0` in the case of `VoteAggregator` or `height=0` in the case of `Forks`). In contrast, getting close to the max value of `uint64` is generally not anything we worry about. Convention 1 is very safe in case you want to _keep_ the entities at the threshold; while convention 2 is error prone is this scenario. 

We _could_ have different pruning conventions, depending on the application. In some implementations we prune _including_ the threshold, in other cases we prune _excluding_ the threshold. But wouldn't that be super confusing? Therefore, I am recommending using the safe convention 1 everywhere and be clear about the convention in the documentation. The current implementation states: https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/consensus/hotstuff/vote_collectors.go#L25-L28

zhangchiqing: Thanks for the detail of the reasoning. 

I agree that it's better to use consistent convention everywhere, but I think it's a bit different here. 

The difference is that: 
For levelled forrest, we would like to keep the last finalized block, whereas in vote aggregator, we don't need to keep any vote at the finalized height. 

However, we would like to Prune both by the finalized height, meaning:
- for levelled forrest, we would like to call `PruneUpToLevel(finalizedBlock.Height)`, and prune all blocks whose view is `below` the finalized height, because we want to keep the finalized block.
- for vote aggregator, we would like to call `PruneUpToView(finalizedBlock.View)`, and prune all blocks whose view is `below and equal` the finalized height, because we `don't` want to keep the votes for finalized block. 

Now the question is: should we keep the votes for the finalized block? 
The benefit of keeping the votes for finalized block is to keep both implementation consistent, and becomes a convention. However, the drawback is that it allows the possibility of vote aggregator creating a different QC for the finalized block. Due to this drawback, I would prefer breaking the convention and eliminating that edge case. In other words, keep the original implementation.

AlexHentschel: Good discussion. Thanks Leo for the detailed reply. 

> Now the question is: should we keep the votes for the finalized block?

That is a valid point. I agree that we probably _don't_ want to keep the votes for the finalized block: there is no point to keeping them. Now, the question is how do we remove the votes for the finalized view. We could
1. Have a different convention for the `VoteAggregator`; as you are suggesting  `VoteAggregator.PruneUpToView(finalizedBlock.View)` would remove also the votes for `finalizedBlock.View`.
2. The alternative approach, which I have taken in this PR is:
   * `VoteAggregator.PruneUpToView(view)` follows the established convention (_not_ removing the votes for `view`)
   * Here, `view` is essentially the lowest view for which we retain votes. This is reflected by the variable naming: https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/consensus/hotstuff/voteaggregator/vote_aggregator.go#L32
   * So when we have `finalizedBlock`, the lowest view for which we still want to process votes is `finalizedBlock.View+1`:
      * when instantiating the `VoteAggregator`, is already does this https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/consensus/vote_aggregator.go#L27
      * we haven't implemented the pruning logic yet (I think ?), but we would need also the `+1` there

Let me know what you think. Thanks

zhangchiqing: Thanks for the reasoning. 

For option 2, I think asking the caller to do the following, implies the caller knows how vote aggregator prune logic. I would suggest to remove the `+1` by letting the vote aggregator take finalized view, rather than lowest view, because `lowest` is only meaningful to the vote aggregator internal implementation. Even if we would like to compute `lowest = finalizedView + 1`, we could compute it in the constructor function. 
```
 aggregator, err := voteaggregator.NewVoteAggregator(log, notifier, finalized.View+1, voteCollectors) 
```

However, I still think it would be a bit confusing: what is lowest? is it a finalized view or pruned view, is finalized view pruned? 
But IMO we don't have this confusion with option 1, because the constructor would do:
```
va. highestPrunedView.Set(finalizedView)
```
And the `PruneUpToView` function would do:
```
func (va *VoteAggregator) PruneUpToView(finalizedView uint64) {
	if va.highestPrunedView.Set(finalizedView) {
             va.collectors.PruneUpToView(finalizedView)
        }
```

It's clear that the `highestPrunedView` is the view where votes have been pruned, and by renaming the argument of `PruneUpToView` to `finalizedView`, it's clear, votes up to the finalized view will be pruned.

AlexHentschel: I feel the pruning policy is an _external_ aspect to vote aggregator. Technically we could prune already up (and including) the view of the highest known QC. My intrinsic preference is to keep the `VoteAggregator` as generic as possible. Therefore, the `VoteAggregator` doesn't make any specific assumption about the pruning policy. `VoteAggregator` says: give me some view and I throw everting away below that view. We _externally_ decide to tell it to prune at the finalized view. But we could make other (more performant but also more complex to implement) decisions. 

You are right that the [utility logic to initialize the `VoteAggregator` and its dependencies](https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/consensus/vote_aggregator.go#L18) is inconsistent with what I just said, in that is _directly_ uses the finalized block: https://github.com/onflow/flow-go/blob/ab5038a35fe6e841b60afa2a90e8209866198415/consensus/vote_aggregator.go#L18

Suggestions for avoiding confusions (quickly implemented in my most recent commit https://github.com/onflow/flow-go/pull/1725/commits/b09fad53d9140816f8574c7cff33935469375ea9): 
* In `VoteCollectors` as well as `VoteAggregator`, use the variable naming `lowestRetainedView`
* clarify in goDoc https://github.com/onflow/flow-go/blob/b09fad53d9140816f8574c7cff33935469375ea9/consensus/hotstuff/voteaggregator/vote_aggregator.go#L214-L218
* In the initialization utility for Vote Aggregator, I now also use `lowestRetainedView` https://github.com/onflow/flow-go/blob/e778aba54e93a53b89e0c47e2e5b0e29d5c0ef56/consensus/vote_aggregator.go#L15-L20

Thereby, the VoteAggregator becomes completely agnostic to the pruning policy (and consistent with the prevalent `PruneUpToView` convention 😅 )

If you still find this confusion, let me know and I'll change to convention 1. Though, if we go with convention 1., my preference would be to use a different method name, e.g. `PruneToFinalizedView(finalized view)` to clarify that this method uses a different convention than all the other `PruneUpToView` methods.

engine/collection/epochmgr/factories/hotstuff.go
```diff
@@ -110,7 +110,7 @@ func (f *HotStuffFactory) CreateModules(epoch protocol.Epoch,
 	voteProcessorFactory := votecollector.NewStakingVoteProcessorFactory(committee, qcDistributor.OnQcConstructedFromVotes)
 	aggregator, err := consensus.NewVoteAggregator(
 		f.log,
-		finalizedBlock,
+		finalizedBlock.View+1,
```
zhangchiqing, line 113:
```suggestion
		// since we don't want to aggregate votes for finalized view, 
		// the lowest retained view starts with the next view of the last finalized view.
		finalizedBlock.View+1,
```

zhangchiqing (approved):
Looks good. The new variable name is also clear.

AlexHentschel:
thanks @zhangchiqing for consistently advocating for clear implementations and the insightful discussions on this PR. Your insights are much appreciated.

PR #1731, opened 2021-12-06
Log epoch number on execution node startup

janezpodhostnik:
closes: https://github.com/dapperlabs/flow-go/issues/5999

This change logs the epoch number from the FlowEpoch smart contract and from the protocol state on execution node startup.

Excerpt from the logs:

```json
[
  {
    "level": "info",
    "node_role": "execution",
    "node_id": "385341c8b20cd5ad80c84092693279185cb0144465a4af56b2ea43990a780a8a",
    "component": "provider engine",
    "contractEpochCounter": 0,
    "protocolStateEpochCounter": 0,
    "flowEpochAddress": "0xf8d6e0586b0a20c7",
    "time": "2021-12-06T15:57:10Z",
    "message": "Fetched epoch counter from the FlowEpoch smart contract"
  },
  {
    "level": "info",
    "node_role": "execution",
    "node_id": "385341c8b20cd5ad80c84092693279185cb0144465a4af56b2ea43990a780a8a",
    "component": "provider engine",
    "time": "2021-12-06T15:57:10Z",
    "message": "component initialization complete"
  },
  {
    "level": "info",
    "node_role": "execution",
    "node_id": "385341c8b20cd5ad80c84092693279185cb0144465a4af56b2ea43990a780a8a",
    "component": "provider engine",
    "time": "2021-12-06T15:57:10Z",
    "message": "component startup complete"
  }
]
```

Since it is only meant for logging, it does not fail if an error is encountered.

I am not sure this is the right location for this code, but I could not find a better one.

### DIFF
--- a/cmd/execution/main.go
+++ b/cmd/execution/main.go
@@ -578,7 +578,7 @@ func logEpochNumber(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState st
 
 	// Get the epoch counter form the protocol state
 	protocolStateEpoch, err := node.State.
-		AtBlockID(node.RootBlock.ID()).
+		Final().
 		Epochs().
 		Current().
 		Counter()
--- a/cmd/execution/main.go
+++ b/cmd/execution/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/onflow/flow-go/engine/execution/state/delta"
 	"io"
 	"os"
 	"path"
@@ -391,38 +392,50 @@ func main() {
 				return nil, err
 			}
 
-			// Get the epoch counter from the smart contract.
-			contractEpochCounter, err := getContractEpochCounter(vm, vmCtx, executionState)
+			// Get latest executed block and a view at that block
+			ctx := context.Background()
+			_, blockID, err := executionState.GetHighestExecutedBlockID(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("cannot get the latest executed block id: %w", err)
+			}
+			stateCommit, err := executionState.StateCommitmentByBlockID(ctx, blockID)
+			if err != nil {
+				return nil, fmt.Errorf("cannot get the state comitment at latest executed block id %s: %w", blockID.String(), err)
+			}
+			blockView := executionState.NewView(stateCommit)
+
+			// Get the epoch counter from the smart contract at the last executed block.
+			contractEpochCounter, err := getContractEpochCounter(vm, vmCtx, blockView)
 			// Failing to fetch the epoch counter from the smart contract is a fatal error.
 			if err != nil {
-				return nil, fmt.Errorf("cannot get epoch counter from the smart contract: %w", err)
+				return nil, fmt.Errorf("cannot get epoch counter from the smart contract at block %s: %w", blockID.String(), err)
 			}
-			// Get the epoch counter form the protocol state.
+
+			// Get the epoch counter form the protocol state, at the same block.
 			protocolStateEpochCounter, err := node.State.
-				Final().
+				AtBlockID(blockID).
 				Epochs().
 				Current().
 				Counter()
 			// Failing to fetch the epoch counter from the protocol state is a fatal error.
 			if err != nil {
-				return nil, fmt.Errorf("cannot get epoch counter from the protocol state: %w", err)
+				return nil, fmt.Errorf("cannot get epoch counter from the protocol state at block %s: %w", blockID.String(), err)
 			}
+
+			l := node.Logger.With().
+				Str("component", "provider engine").
+				Uint64("contractEpochCounter", contractEpochCounter).
+				Uint64("protocolStateEpochCounter", protocolStateEpochCounter).
+				Str("blockID", blockID.String()).
+				Logger()
 
 			if contractEpochCounter != protocolStateEpochCounter {
 				// Do not error, because immediately following a spork they will be mismatching,
 				// until the resetEpoch transaction is submitted.
-				node.Logger.
-					Warn().
-					Str("component", "provider engine").
-					Uint64("contractEpochCounter", contractEpochCounter).
-					Uint64("protocolStateEpochCounter", protocolStateEpochCounter).
+				l.Warn().
 					Msg("Epoch counter from the FlowEpoch smart contract and from the protocol state mismatch!")
 			} else {
-				node.Logger.
-					Log().
-					Str("component", "provider engine").
-					Uint64("contractEpochCounter", contractEpochCounter).
-					Uint64("protocolStateEpochCounter", protocolStateEpochCounter).
+				l.Info().
 					Msg("Epoch counter from the FlowEpoch smart contract and from the protocol state match.")
 			}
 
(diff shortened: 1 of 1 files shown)


cmd/execution/main.go
```diff
@@ -564,6 +570,81 @@ func main() {
...
+func logEpochNumber(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState state.ExecutionState, node *cmd.NodeConfig, log zerolog.Logger) {
+	logFail := func(err error) {
+		log.Error().Err(err).Msg("could not log epoch number")
+	}
+
+	protocolStateEpoch, err := node.State.
+		AtBlockID(node.RootBlock.ID()).
+		Epochs().
+		Current().
+		Counter()
```
janezpodhostnik, line 582: I am not 100% on this part. Can someone confirm that this is the correct way to get the protocol state epoch counter.

cmd/execution/main.go
```diff
@@ -564,6 +570,83 @@ func main() {
...
 
+// Get and log the epoch counters from the protocol state and from the FlowEpoch smart contract.
+func logEpochNumber(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState state.ExecutionState, node *cmd.NodeConfig, log zerolog.Logger) {
+	logFail := func(err error) {
+		log.Error().Err(err).Msg("could not log epoch number")
+	}
+
+	// Get the epoch counter form the protocol state
+	protocolStateEpoch, err := node.State.
+		AtBlockID(node.RootBlock.ID()).
```
jordanschalm, line 581:
```suggestion
		Final().
```

`AtBlockID(node.RootBlock.ID())` will reference the root block of the snapshot, `Final()` will reference the latest finalized block (staying up to date as the execution node incorporates blocks). I'm pretty sure we want the up-to-date version here, since we're using the latest executed block for reading from the smart contract.

jordanschalm (approved):
Thanks! Just one note

cmd/execution/main.go
```diff
@@ -564,6 +570,83 @@ func main() {
 		}).Run()
 }
 
+// Get and log the epoch counters from the protocol state and from the FlowEpoch smart contract.
+func logEpochNumber(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState state.ExecutionState, node *cmd.NodeConfig, log zerolog.Logger) {
```
zhangchiqing, line 574: It's better to make a function that simply reads the epoch counter from the smart contract.
it should either return
- (epochContractAddress, epochCounter, nil)  // if was able to read the epoch counter successfully
- ("", 0, err) // or run into an exception
```suggestion
func readEpochCounterFromSmartContract(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState state.ExecutionState) (string, uint64, error) {
```

cmd/execution/main.go
```diff
@@ -385,6 +389,8 @@ func main() {
 				chdpDeliveryTimeout,
 			)
 
+			logEpochNumber(vm, vmCtx, executionState, node, node.Logger.With().Str("component", "provider engine").Logger())
```
zhangchiqing, line 392: I would suggest to organize the logic a bit, and we should crash the node if we ran into any error, or in the end the epoch counter in protocol state and smart contract is mismatching.

```golang
epochCounterFromProtocolState, err := node.State.Final().Epochs().Current().Counter()
if err != nil {
  return fmt.Errorf(...) // we would like to stop the node if we fail to query protocol state
}
epochCounterFromSmartContract, err := readEpochCounterFromSmartContract(...)
if err != nil {
  return fmt.Errorf(...) // we would like to stop the node if we fail to query smart contract
}
// we would like to stop the node if the epoch counter is mismatching
if epochCounterFromProtocolState != epochCounterFromSmartContract {
  return fmt.Errorf("mismatching epoch counter between protocol state %v and smart contract %v", 
  epochCounterFromProtocolState, epochCounterFromSmartContract)
} 

log.Info().Uint64("epochCounter", epochCounterFromProtocolState).Msg("epoch counter is matching")
```

zhangchiqing (changes requested):
Thanks for adding this check! 

Changes looks good, just added some suggestion to improve the error handling.

cmd/execution/main.go
jordanschalm: > we should crash the node if we ran into any error, or in the end the epoch counter in protocol state and smart contract is mismatching

@zhangchiqing We can't do this because immediately following a spork they will be mismatching until the `resetEpoch` transaction is submitted.

zhangchiqing: I see. Right. Let's simply logging them then.

cmd/execution/main.go
janezpodhostnik: refactored a bit

cmd/execution/main.go
```diff
@@ -384,8 +387,46 @@ func main() {
...
+				// until the resetEpoch transaction is submitted.
+				node.Logger.
+					Warn().
+					Str("component", "provider engine").
+					Uint64("contractEpochCounter", contractEpochCounter).
+					Uint64("protocolStateEpochCounter", protocolStateEpochCounter).
+					Msg("Epoch counter from the FlowEpoch smart contract and from the protocol state mismatch!")
+			} else {
+				node.Logger.
+					Log().
```
zhangchiqing, line 422:
```suggestion
					Info().
```

cmd/execution/main.go
```diff
@@ -564,6 +605,55 @@ func main() {
 		}).Run()
 }
 
+// getContractEpochCounter Gets the epoch counters from the FlowEpoch smart contract from the latest executed block.
+func getContractEpochCounter(vm *fvm.VirtualMachine, vmCtx fvm.Context, executionState state.ExecutionState) (uint64, error) {
+	// Get the address of the FlowEpoch smart contract
+	sc, err := systemcontracts.SystemContractsForChain(vmCtx.Chain.ChainID())
+	if err != nil {
+		return 0, err
```
zhangchiqing, line 613:
```suggestion
		return 0, fmt.Errorf("could not get system contracts: %w", err)
```

cmd/execution/main.go
```diff
@@ -564,6 +605,55 @@ func main() {
...
+
+	// Generate the script to get the epoch counter from the FlowEpoch smart contract
+	scriptCode := templates.GenerateGetCurrentEpochCounterScript(templates.Environment{
+		EpochAddress: address.Hex(),
+	})
+	script := fvm.Script(scriptCode)
+
+	// Get latest executed block and a view at that block
+	ctx := context.Background()
+	_, blockID, err := executionState.GetHighestExecutedBlockID(ctx)
```
zhangchiqing, line 625: should we compare the epoch counter for the same block? it's possible that we are reading epoch counter from smart contract for block A , but reading epoch counter from protocol state for block B, and I'm not sure comparing them is accurate.

cc @jordanschalm 

However, should we compare the same block for the last finalized block? no, because a finalized block might not be executed yet, in which case we don't have the execution state to read from.

Maybe we should read the epoch counter for the last executed block, which guarantees we can read from both protocol state and smart contract. 

Also we are reading epoch counter for the same block, we'd better log the block view and block ID, along with the epoch counter.

cmd/execution/main.go
```diff
@@ -564,6 +605,55 @@ func main() {
...
+	if script.Err == nil && script.Value != nil {
+		epochCounter := script.Value.ToGoValue().(uint64)
+		return epochCounter, nil
+	}
+
+	if script.Err != nil {
+		return 0, script.Err
+	}
+
+	return 0, fmt.Errorf("script returned no value")
```
zhangchiqing, line 654:
```suggestion
	if script.Err != nil {
		return 0, fmt.Errorf("could not read epoch counter, script error: %w", script.Err)
	}
	
	if script.Value == nil {
		return 0, fmt.Errorf("script to read epoch counter returned no value at block view %v", blockView)
	}

	epochCounter := script.Value.ToGoValue().(uint64)
	return epochCounter, nil
	
```

cmd/execution/main.go
jordanschalm: > Maybe we should read the epoch counter for the last executed block, which guarantees we can read from both protocol state and smart contract.

Yes, great point. We should definitely do this.

janezpodhostnik: I changed it so they should use the same block now.

cmd/execution/main.go
janezpodhostnik: I added the blockId to the errors outside of this function, and refactored a bit.

zhangchiqing (approved):
Nice work! Looks good to me.

janezpodhostnik:
bors merge

PR #1732, opened 2021-12-06
DKG Storage Pt III: Check beacon key consistency atomically with epoch phase transition

jordanschalm:
This PR updates the consistency check of beacon keys to occur atomically with the corresponding protocol state mutation. It also modifies protocol events in general to occur within the scope of the corresponding protocol state mutation, so that they generally provide guarantee at-least-once semantics. Also: refactor and extend tests for beacon key checking in reactor engine.

⚠️ Depends on #1720, #1721


engine/consensus/dkg/reactor_engine.go
```diff
@@ -229,42 +229,59 @@ func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow
...
+// program will crash). Since this function is invoked synchronously before the end
+// of the current epoch, this guarantees that when we reach the end of the current epoch
+// we will either have a usable beacon key (successful DKG) or a DKG failure end state
+// stored, so we can safely fall back to using our staking key.
 //
-func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uint64) {
+// CAUTION: This function is not safe for concurrent use. This is not enforced within
+// the ReactorEngine - instead we rely on the protocol event emission being single-threaded
+//
+func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uint64, firstBlock *flow.Header) {
```
jordanschalm, line 246: During the last discussion we considered the option of passing in the Badger transaction context through the `EpochCommittedPhaseStarted` protocol event callback. I decided against doing this for two reasons:
* Since epoch-related protocol events are emitted when the relevant block is **finalized**, the block (and corresponding snapshot) are already accessible in the protocol state, and are immutable so access them here is safe.
* Many of the storage operations we need to perform here are targeting `DKGState`, which uses a separate physical database backend (`SecretsDB`), so we would be unable to have ACID semantics across all the storage operations here anyway.

state/protocol/badger/mutator.go
```diff
@@ -871,31 +879,36 @@ func (m *FollowerState) MarkValid(blockID flow.Identifier) error {
...
+
+		// root blocks and blocks below the root block are considered as "processed",
+		// so we don't want to trigger `BlockProcessable` event for them.
+		var rootHeight uint64
+		err = m.db.View(operation.RetrieveRootHeight(&rootHeight))
+		if err != nil {
+			return fmt.Errorf("could not retrieve root block's height: %w", err)
+		}
+		// trigger BlockProcessable for parent blocks above root height
+		if parent.Height > rootHeight {
```
zhangchiqing, line 902: is root block the first block of an epoch? 
Why do we need this check?

state/protocol/badger/mutator.go
```diff
@@ -871,31 +879,36 @@ func (m *FollowerState) MarkValid(blockID flow.Identifier) error {
...
+		}
+
+		// root blocks and blocks below the root block are considered as "processed",
+		// so we don't want to trigger `BlockProcessable` event for them.
+		var rootHeight uint64
+		err = m.db.View(operation.RetrieveRootHeight(&rootHeight))
+		if err != nil {
+			return fmt.Errorf("could not retrieve root block's height: %w", err)
+		}
+		// trigger BlockProcessable for parent blocks above root height
```
zhangchiqing, line 901:
```suggestion
		// a root block is considered as processed. Since we are triggering BlockProcessable for 
		// parent blocks, it should not be triggered if parent block is a root block. Therefore, 
		// trigger BlockProcessable for parent blocks above root height
```

state/protocol/badger/mutator.go
```diff
@@ -871,31 +879,36 @@ func (m *FollowerState) MarkValid(blockID flow.Identifier) error {
...
+			return fmt.Errorf("could not insert validity for block (id=%x, height=%d): %w", blockID, header.Height, err)
+		}
+
+		// root blocks and blocks below the root block are considered as "processed",
+		// so we don't want to trigger `BlockProcessable` event for them.
+		var rootHeight uint64
+		err = m.db.View(operation.RetrieveRootHeight(&rootHeight))
+		if err != nil {
+			return fmt.Errorf("could not retrieve root block's height: %w", err)
+		}
```
zhangchiqing, line 900: Since root block never change, we could move this query out of the transaction, in order to reduce the lock duration

zhangchiqing (approved):
Looks good. Just one optimization suggestion

state/protocol/badger/mutator.go
jordanschalm: The root block is the first block of a spork. I think we don't want to trigger `BlockProcessable` for it because it's considered to have already been processed at startup. 

The logic here should not be different with this PR, it's just rearranged a bit.

state/protocol/badger/mutator.go
jordanschalm: https://github.com/onflow/flow-go/pull/1732/commits/5588c1cd247eb7bc358a4ecf8cd08f98d1568645

engine/consensus/dkg/reactor_engine.go
```diff
@@ -126,8 +126,8 @@ func (e *ReactorEngine) EpochSetupPhaseStarted(currentEpochCounter uint64, first
 
 // EpochCommittedPhaseStarted handles the EpochCommittedPhaseStarted protocol
 // event by checking the consistency of our locally computed key share.
-func (e *ReactorEngine) EpochCommittedPhaseStarted(currentEpochCounter uint64, _ *flow.Header) {
-	e.handleEpochCommittedPhaseStarted(currentEpochCounter)
+func (e *ReactorEngine) EpochCommittedPhaseStarted(currentEpochCounter uint64, first *flow.Header) {
+	e.handleEpochCommittedPhaseStarted(currentEpochCounter, first)
```
tarakby, line 130: Is there a reason for introducing `handleEpochCommittedPhaseStarted` ?

engine/consensus/dkg/reactor_engine.go
```diff
@@ -229,42 +229,59 @@ func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow
...
 		log.Fatal().Err(err).Msg("checking beacon key consistency: could not retrieve next DKG info")
 		return
 	}
 
-	myBeaconPrivKey, err := e.dkgState.RetrieveMyBeaconPrivateKey(currentEpochCounter + 1)
+	myBeaconPrivKey, err := e.dkgState.RetrieveMyBeaconPrivateKey(nextEpochCounter)
 	if errors.Is(err, storage.ErrNotFound) {
 		log.Warn().Msg("checking beacon key consistency: no key found")
-		err := e.dkgState.SetDKGEndState(dkgEpochCounter, flow.DKGEndStateNoKey)
+		err := e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateNoKey)
```
tarakby, line 280: Aren't we overwriting the DKG end state here? For instance, in the case where there was a local failure, DKG end state is set to  `DKGEndStateDKGFailure` and no key is stored. 

We could : 
- avoid setting the DKG end state here. The state might stay in its default state `DKGEndStateUnknown` but that should be fine.
- OR read the state first and overwrite only if the state is `DKGEndStateUnknown`.
- OR merge `DKGEndStateDKGFailure` and `DKGEndStateNoKey` cases but then we would lose the detailed state type.

engine/consensus/dkg/reactor_engine.go
```diff
@@ -386,15 +409,7 @@ func (e *ReactorEngine) end(nextEpochCounter uint64) func() error {
...
-
-		if privateShare == nil {
-			// we failed to compute a private key - set the corresponding DKG failure state
-			e.log.Warn().Msg("computed nil private key share")
-			err = e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateNoKey)
-			if err != nil {
-				return fmt.Errorf("with nil key could not set dkg end state to %s: %w", flow.DKGEndStateNoKey, err)
-			}
-		} else {
+		if privateShare != nil {
```
tarakby, line 412: 👌🏼

tarakby (commented):
Nice work on the comments 🏄🏼

engine/consensus/dkg/reactor_engine.go
jordanschalm: > OR read the state first and overwrite only if the state is DKGEndStateUnknown.

This is what we do:
https://github.com/onflow/flow-go/blob/bdfe4b7ded8fd198c5cc821079c4ae4c499a5d8c/engine/consensus/dkg/reactor_engine.go#L256-L264

If we have a local DKG failure, we will store (`key=nil,end_state=dkg_failure`). If we restart during the DKG, we will store (`key=nil,end_state=no_key`).

This test case captures this scenario:
https://github.com/onflow/flow-go/blob/562dc57f75b674272289cd0b68d69df001142a11/engine/consensus/dkg/reactor_engine_test.go#L418

engine/consensus/dkg/reactor_engine.go
jordanschalm: Previously it spawned a goroutine, which was the original reason. Even without the goroutine though, I think there's a case for separating the public function implementing the interface from the internal handling logic.

engine/consensus/dkg/reactor_engine.go
tarakby: You're right, I missed the check for the nil error

engine/consensus/dkg/reactor_engine.go
```diff
@@ -229,42 +229,59 @@ func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow
...
 
-	// TODO - first check whether we've already stored the dkg end state, since
-	// we need to handle multiple calls to this method
+	// Check whether we have already set the end state for this DKG.
+	// This can happen if the DKG failed locally, if we failed to generate
+	// a local private beacon key, or if we crashed while performing this
+	// check previously.
+	endState, err := e.dkgState.GetDKGEndState(nextEpochCounter)
+	if err == nil {
+		log.Warn().Msgf("already set dkg end state when checking beacon key consistency: %s", endState.String())
```
tarakby, line 262: I think the end state could be set by `(e *ReactorEngine) end` to `DKGEndStateDKGFailure`. I mean the state was not necessarily set by a previous consistency check right? 
```suggestion
		log.Warn().Msgf("already set dkg end state: %s", endState.String())
```

tarakby (approved):

engine/consensus/dkg/reactor_engine.go
jordanschalm: https://github.com/onflow/flow-go/pull/1732/commits/44d144ccff93dd705c1e0e075f8d3375a047accc

PR #1734, opened 2021-12-08
reduce unnecessary convertion

zhiqiangxu:
Just stumble upon, here seems no convertion is actually needed.


Kay-Zee (approved):

AlexHentschel:
thanks for the PR; good catch. I think this optimization already made it into our code base through some other way. This is the corresponding code on master:
https://github.com/onflow/flow-go/blob/495fa47681fa500a0a4156ab0b0f897273487329/model/flow/sealing_segment.go#L198-L202

AlexHentschel:
Closing this PR, as it is practically is now a no-op now 😅.

PR #1742, opened 2021-12-08
Remove the crypto replace and add a crypto setup script

tarakby:
tldr;
 - remove the `replace` statement for the crypto package in flow-go. 
 - add a script to install the crypto package properly (by flow-go or any other project importing the package).
 - update the CI, docker scripts to follow the new process.
 - update the Crypto README about the process + [unrelated] mention the crypto security audit 
 

In order to remove the `replace` statement from the `go.mod` file in flow-go, we would need to manage importing the crypto package in Go properly. This is done by installing the crypto dependencies in the crypto package folder with the right version under `GOPATH/pkg/mod`. 
The crypto package needs to be importable by flow-go, but also by any other Go project outside Flow. The update should  be generic and not specific to the flow-go case.

This PR updates the crypto package `README` to include a script showing how Flow crypto package can be imported by any project. The PR also applies the example for `onflow/flow-go` project which imports the crypto package. 

Unlike other projects importing the crypto package, `onflow/flow-go` requires  2 versions of the crypto package:
 1. the source used by all other packages. It is located under `GOPATH/pkg/mod` and is set up by the example script above 
 2. the source under `onflow/flow-go` where the `flow-go` was cloned. This version is needed to run the CI tests, called by `make ci` through `make unittest-main`. A new call to set up the crypto package was added to `flow-go/crypto/Makefile` at every general unit test call.
  
[possible optimization]:
 `1` and `2` were separated for a clearer distinction. However, for `flow-go` we don't have to use  the example script. step `2` is usually done after `1` is done, so `2` could be achieved by making the script copy the resulting binary to  `flow-go/crypto`. This could avoid building Relic twice.


Kay-Zee:
The main difference i see between this, and the approach i took in https://github.com/onflow/flow-go/pull/1682 is that this now requires a script call instead of a go generate call.

The rest, e.g. the example script you're adding to the readme, etc, could still be added in the other PR, for example.

The reason i put everything into one script is it then allows us to stay a bit more loyal to go tooling, and continue to make use of go generate, although the difference is not significant, as go generate then goes and calls a script file anyways

Kay-Zee (approved):

zhangchiqing (approved):

tarakby:
bors merge

PR #1741, opened 2021-12-08
[Consensus Voting] refactor VerifyVote

zhangchiqing:
Before VerifyVote returns a bool to indicate whether the vote is valid, however, when the vote is invalid, the returned `bool` value can't carry details of why the vote is invalid. 

After changing to use sentinal error to indicate invalid vote, we could include details like the vote is invalid because the staking sig is invalid.

PR #1743, opened 2021-12-09
[Consensus Voting] refactor verify QC

zhangchiqing:
Review https://github.com/onflow/flow-go/pull/1741 first

This is a similar idea as  [1741](https://github.com/onflow/flow-go/pull/1741). 

After changing to use sentinal error to indicate invalid QC, we could include details like the QC is invalid because the staking sig is invalid.


durkmurder (approved):
LGTM

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -43,112 +43,118 @@ func NewCombinedVerifierV3(committee hotstuff.Committee, packer hotstuff.Packer)
...
 	}
 
 	// first fetch all beacon signers public keys
 	beaconPubKeys := make([]crypto.PublicKey, 0, len(blockSigData.RandomBeaconSigners))
 	for _, signerID := range blockSigData.RandomBeaconSigners {
 		keyShare, err := dkg.KeyShare(signerID)
 		if err != nil {
-			return false, fmt.Errorf("could not find key share for signer %v: %w", signerID, err)
+			// the packer should have found the signer, otherwise the packer will return an ErrInvalidFormat
+			// and we won't reach it. So reaching here would be an exception
```
AlexHentschel, line 148: the packer only checks that the ID corresponds to a valid consensus participant. Not sure what happens, if the consensus participant failed the DKG (?) Will we retrieve a dummy key?

Your comment is correct, as long as we (implicitly) assume that every consensus participant is also a random beacon participant. While we have this convention at the moment, it is theoretically also possible to only have a _sub-set_ of the consensus committee run the Random Beacon. I think it is worth-while to document this assumption here: 

 ```suggestion
// We assume that every consensus member is also a random beacon participant. The 
// Packer checks that only valid consensus participants are in the signers list. Otherwise,
// the packer should return an ErrInvalidFormat. Consequently, we expect for every 
// signer a DKG record. Reaching this failure case indicates a bug in the internal logic.
```

File: consensus/hotstuff/verification/combined_verifier_v3.go [hunk 2]
```diff
@@ -158,14 +164,14 @@ func (c *CombinedVerifierV3) VerifyQC(signers flow.IdentityList, sigData []byte,
 	for _, signerID := range blockSigData.StakingSigners {
 		identity, ok := signerIdentities[signerID]
 		if !ok {
-			return false, fmt.Errorf("invalid signer identity %v", signerID)
+			return fmt.Errorf("invalid signer identity %v: %w", signerID, model.ErrInvalidSignature)
```
AlexHentschel, line 167: ## ⚠️ 

You correctly argued above that this case should not occur, and reaching this failure condition indicates an internal bug: https://github.com/onflow/flow-go/blob/d2788f297c099edf033310d70fc66cee3a63ce8b/consensus/hotstuff/verification/combined_verifier_v3.go#L147-L148
I think the same argument applies here.  Hence, I think we should _not_ return a `model.ErrInvalidSignature` here, but an exception instead.

AlexHentschel (approved):
Looks good. Mostly stylistic comments. Only critical comment is [this one](https://github.com/onflow/flow-go/pull/1743#discussion_r768328880). In addition, a couple comments on a few aspects that I missed in the tests:


#### Requesting to add another test to `VoteSuite`: 
* in [this test](https://github.com/onflow/flow-go/blob/d2788f297c099edf033310d70fc66cee3a63ce8b/consensus/hotstuff/validator/validator_test.go#L316-L325) we verify that the Validator recognizes an `ErrInvalidSignature` as invalid vote, i.e.  wraps it as `InvalidVoteError`. 
* However, I think we are not verifying the correct handling of `signature.ErrInvalidFormat` 

####  `StakingVerifier` tests: 
* We test [here](https://github.com/onflow/flow-go/blob/d2788f297c099edf033310d70fc66cee3a63ce8b/consensus/hotstuff/verification/staking_signer_test.go#L83-L101) the happy path.
* Do we have tests for the unhappy path (i.e. invalid votes or QCs)? Would be great to cover in tests that the expected sentinel errors are returned. 

(we could do that in a subsequent PR)

consensus/hotstuff/verification/combined_verifier_v3.go
zhangchiqing: >Not sure what happens, if the consensus participant failed the DKG (?) Will we retrieve a dummy key?

@AlexHentschel Let me confirm your question:
Say a consensus participant node B failed the DKG, and we are node A , who successfully completed the DKG. The question is: will node A retrieve a dummy public key for node B? 

My understanding is yes, if DKG has completed, we will have a key for every consensus participant, regardless whether each participant actually succeeded or failed DKG. cc @tarakby to confirm

And your question also reminds me one more thing that I would like to double check. 
There are two events: 
1. DKG completed
2. DKG ended

The difference between these two events are:
- event 1 happens _near_ the end of the current Epoch, whereas event 2 happens _at_ the end of the current Epoch. 
- event 1 will produce locally the private key for myself for the next epoch, whereas event 2 will receive from the block payload the DKG keys for all DKG participants for the next epoch. 
- The DKG keys from event 2 is the canonical DKG keys for all DKG participants, which can be used verify the local genereated DKG private key generated from event 1. 

The question is would event 1 also produce locally the DKG keys for all consensus participants? 
If yes, then is there a possibility that it will be different from the canonical DKG keys from event 2? 
If yes, then we need to ensure that the `dkg.KeyShare` will read the key share from event 2, not event 1. Right? Is this done already? cc @tarakby @jordanschalm

consensus/hotstuff/verification/combined_verifier_v3.go
zhangchiqing: Good catch. 

I'm also thinking that maybe we should let the packer return Identity instead, then we can eliminate this case

consensus/hotstuff/verification/combined_verifier_v3.go
jordanschalm: > event 1 happens near the end of the current Epoch, whereas event 2 happens at the end of the current Epoch.

Both happen strictly before the end of the current epoch. Event 2 corresponds with the transition to the `EpochCommitted` phase. Then we have the entirety of that phase before the end of the current epoch (typically at least an hour).

Clearer descriptions might be:
Event 1 - local DKG completion (I have my key, and submitted my results)
Event 2 - global DKB completion (I observe the results committed to by a majority of participants)

> The question is would event 1 also produce locally the DKG keys for all consensus participants?

Yes

> If yes, then is there a possibility that it will be different from the canonical DKG keys from event 2?

Yes

> If yes, then we need to ensure that the dkg.KeyShare will read the key share from event 2, not event 1

Yes this is what we do. Only the canonical DKG keys from the service event are incorporated into the protocol state.

zhangchiqing:
bors merge

consensus/hotstuff/verifier.go
```diff
@@ -41,5 +41,5 @@ type Verifier interface {
 	// * signature.ErrInvalidFormat if the signature has an incompatible format.
 	// * unexpected errors should be treated as symptoms of bugs or uncovered
 	//   edge cases in the logic (i.e. as fatal)
-	VerifyQC(voters flow.IdentityList, sigData []byte, block *model.Block) (bool, error)
+	VerifyQC(voters flow.IdentityList, sigData []byte, block *model.Block) error
```
tarakby, line 44: It would be good tp update the go doc too to mirror the new API

tarakby (commented):
Looks good to me!
I just added a comment to update the go doc to reflect the API change.

consensus/hotstuff/verifier.go
zhangchiqing: Good point. Updated here
https://github.com/onflow/flow-go/pull/1779/files#diff-808ac5bc91604a4ce38e9e740a6efe7ef6701b1e935d4bd1499d6ae1bf9ad406R35

PR #1753, opened 2021-12-09
Relaxing ledger interaction limit for service and system transactions

ramtinms:
This PR:
 - disables ledger interaction limits when a transaction is signed by the service account. this enables the network to support operations like end-epoch to be applicable even when node counts are growing. 
 - let the service chunk uses more ledger interaction and available space on chunk data pack size since it would be the only transaction inside that chunk.

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -26,9 +26,9 @@ import (
 	"github.com/onflow/flow-go/utils/logging"
 )
 
-const SystemChunkEventCollectionMaxSize = 256_000_000 // ~256MB
-const SystemChunkLedgerIntractionLimit = 500_000_000  // ~500MB
-const MaxTransactionErrorStringSize = 1000            // 1000 chars
+const SystemChunkEventCollectionMaxSize = 256_000_000  // ~256MB
+const SystemChunkLedgerIntractionLimit = 1_000_000_000 // ~1GB
+const MaxTransactionErrorStringSize = 1000             // 1000 chars
 
 // VirtualMachine runs procedures
 type VirtualMachine interface {
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -27,6 +27,7 @@ import (
 )
 
 const SystemChunkEventCollectionMaxSize = 256_000_000 // ~256MB
+const SystemChunkLedgerIntractionLimit = 500_000_000  // ~500MB
 const MaxTransactionErrorStringSize = 1000            // 1000 chars
 
 // VirtualMachine runs procedures
@@ -62,6 +63,7 @@ func SystemChunkContext(vmCtx fvm.Context, logger zerolog.Logger) fvm.Context {
 		fvm.WithTransactionFeesEnabled(false),
 		fvm.WithServiceEventCollectionEnabled(),
 		fvm.WithTransactionProcessors(fvm.NewTransactionInvoker(logger)),
+		fvm.WithMaxStateInteractionSize(SystemChunkLedgerIntractionLimit),
 		fvm.WithEventCollectionSizeLimit(SystemChunkEventCollectionMaxSize),
 	)
 }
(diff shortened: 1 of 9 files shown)


engine/execution/computation/computer/computer.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 const SystemChunkEventCollectionMaxSize = 256_000_000 // ~256MB
+const SystemChunkLedgerIntractionLimit = 500_000_000  // ~500MB
```
ramtinms, line 30: @jordanschalm is this enough in comparison to the current 20MB one?

fvm/fvm_test.go
```diff
@@ -950,24 +986,22 @@ func TestBlockContext_ExecuteTransaction_InteractionLimitReached(t *testing.T) {
...
 				txBody.SetProposalKey(chain.ServiceAddress(), 0, 0)
-				txBody.SetPayer(chain.ServiceAddress())
+				txBody.SetPayer(accounts[0])
 
-				err = testutil.SignPayload(txBody, accounts[0], privateKeys[0])
+				err = testutil.SignPayload(txBody, chain.ServiceAddress(), unittest.ServiceAccountPrivateKey)
 				require.NoError(t, err)
 
-				err = testutil.SignEnvelope(txBody, chain.ServiceAddress(), unittest.ServiceAccountPrivateKey)
+				err = testutil.SignEnvelope(txBody, accounts[0], privateKeys[0])
```
ramtinms, line 997: I changed the order so its still authorized by service account but payer is not service account so it won't skip the limit

engine/execution/computation/computer/computer.go
jordanschalm: I'm pretty sure that would be enough, but we don't know what the actual usage is (since the tx reverts once it crosses the current limit).

fvm/transactionInvoker.go
```diff
@@ -158,7 +164,9 @@ func (i *TransactionInvoker) Process(
 	if feesError != nil {
 		txError = feesError
 	}
-	sth.EnableLimitEnforcement()
+	if !payerIsServiceAccount {
+		sth.EnableLimitEnforcement()
```
janezpodhostnik, line 168: there are more `EnableLimitEnforcement` calls in `AddEncodedAccountKey` and `CreateAccount` if those would be called during a transaction, then the limit would be reenabled...

It might be better to put `payerIsServiceAccount` inside the `StateHolder` and use it in `EnforceInteractionLimits`.

Would it make sense to increase the limits instead of disabling them?

Is this meant as a very temporary fix? If that is the case then it is fine, as the transactions that are causing problems are not calling  `AddEncodedAccountKey` or `CreateAccount`.

jordanschalm (approved):

fvm/transactionInvoker.go
ramtinms: good idea, fixed it.

engine/execution/computation/computer/computer.go
ramtinms: how much bigger is mainnet in comparison to testnet? (in terms of input size (probably node counts) 10X ?

jordanschalm: Testnet has ~100 nodes and Mainnet has ~400. But there are also many delegators on Mainnet (and none or close to none on Testnet). I don't know exactly how many delegators there are, but I think there are tens of thousands. I remember seeing just one node (an outlier, but still) with 13,000 delegators.

ramtinms: and the system chunk transaction would touch the delegators list?

jordanschalm: It will when we `moveTokens`, which happens both when we pay rewards (service account) and when we transition epochs (system chunk)

janezpodhostnik (approved):

engine/execution/computation/computer/computer.go
Kay-Zee: do we remember what this use to be before we brought it down to 20mb?

ramtinms: 2_000_000_000 (~2GB) initially but I don't think we ever reached it with the old version

ramtinms: Since this is the only transaction in the system chunk, I'm going to change it to 1GB (and leave 1GB for proofs)

Kay-Zee: sounds good

Kay-Zee (approved):

PR #1755, opened 2021-12-09
Extend Sealing Segment Structure

jordanschalm:
Extends sealing segment structure and construction to handle additional edge cases when the sealing segment is used to bootstrap a new state.

In particular:
* ensure the sealing segment contains the latest seal for all block in the segment
* ensure the sealing segment contains any execution results referenced by receipts (already implemented) and seals (new ehre)

More details in https://github.com/dapperlabs/flow-go/issues/6040


kc1116 (approved):
good work :rocket:

model/flow/sealing_segment.go
```diff
@@ -108,49 +251,39 @@ func (builder *SealingSegmentBuilder) isValidHeight(block *Block) bool {
...
-	// highest contains seals check if seal for lowest exists
-	for _, seal := range highest.Payload.Seals {
-		if seal.BlockID == lowestID {
-			return true
+		// check if block seals lowest
+		for _, seal := range block.Payload.Seals {
+			if seal.BlockID == lowestID {
+				return true
+			}
 		}
```
zhangchiqing, line 273: ⚠️  Here we are assuming a seal that seals the lowest block is the last seal. It might not always be the case. For example:

The following is not a valid SealingSegment, though the last block contains a seal to the first block, but that seal (Seal_A) is not the last seal. The last seal (Seal_B) actually seals `B`.

`A <- B <- C <- D(Result_A, Result_B) <- E (Seal_A, Seal_B)`.

So we need to first find the last seal from the payload. Usually, the last seal is the last one in the seals payload. So should we _just_ check the last one: `block.Payload.Seals[len(block.Payload.Seals) - 1]` ? This would assume there is such ordering rule and has been checked by the protocol when adding this block.

model/flow/sealing_segment.go
```diff
@@ -16,22 +16,51 @@ const (
...
+// Example 1 - E seals A:
+//   A <- B <- C <- D <- E(SA)
 // The above sealing segment's last block (E) has a seal for block A, which is
 // the first block of the sealing segment.
 //
-// It is guaranteed there are at least 2 blocks in a SealingSegment
+// Example 2 - E contains no seals, but latest seal prior to E seals A:
+//   A <- B <- C <- D(SA) <- E
+//
+// It is guaranteed there are at least 2 blocks in a SealingSegment, unless
```
durkmurder, line 32: I am not sure this statement is correct, our seal validator and builder use new rules, to be more specific we require a gap between incorporated result and seal, we can't construct something like:
`A <- B[ER_A] <- C[S_A]` we actually need to have a chain: `A <- B[ER_A] <- C <- D[S_A]`

model/flow/sealing_segment.go
```diff
@@ -42,45 +71,153 @@ func (segment *SealingSegment) Lowest() *Block {
...
+// GetSealByBlockIDFunc is a getter function for seals by block ID, returning
+// the latest seals incorporated as of the given block.
+type GetSealByBlockIDFunc func(blockID Identifier) (*Seal, error)
+
+// SealingSegmentBuilder is a utility for incrementally building a sealing segment.
 type SealingSegmentBuilder struct {
-	resultLookup    func(resultID Identifier) (*ExecutionResult, error)
+	// access to storage to read referenced by not included resources
+	resultLookup        GetResultFunc
+	sealByBLockIDLookup GetSealByBlockIDFunc
```
durkmurder, line 151:
```suggestion
	sealByBlockIDLookup GetSealByBlockIDFunc
```

model/flow/sealing_segment.go
```diff
@@ -42,45 +71,153 @@ func (segment *SealingSegment) Lowest() *Block {
...
 	if !builder.isValidHeight(block) {
 		return fmt.Errorf("invalid block height (%d): %w", block.Header.Height, ErrSegmentInvalidBlockHeight)
 	}
+	blockID := block.ID()
+
+	// construct a list of missing result IDs referenced by this block
+	missingResultIDs := make(map[Identifier]struct{})
 
-	// cache results in included results
+	// for the first (lowest) block, if it contains no seal, store the latest
```
durkmurder, line 172: here we are assuming that we will add blocks in ascending order, maybe we shouldn't do this? 
We can wait for first seal calculation right before returning `SealingSegment`

model/flow/sealing_segment.go
jordanschalm: Good catch. Since all the blocks are already added to the segment at this point, I think we can just use `LatestSeals` to check the latest seal of the highest block here.

jordanschalm: https://github.com/onflow/flow-go/pull/1755/commits/0b4997831571d628bc033fa947618c353f667447

model/flow/sealing_segment.go
jordanschalm: https://github.com/onflow/flow-go/pull/1755/commits/09f68ac85e5b3b55f5a5d2e6b19da32a3c57add0

model/flow/sealing_segment.go
durkmurder: It's a technical requirement for having correct SoR, more here: https://github.com/onflow/flow-go/blob/d69930a2472c99c7e51e58155a0de46022b267e7/module/validation/seal_validator.go#L162

durkmurder (approved):
I didn't find any obvious problems, looks good to me. I really liked clean approach with implementation and testing, thank you!

model/flow/sealing_segment.go
```diff
@@ -108,49 +256,35 @@ func (builder *SealingSegmentBuilder) isValidHeight(block *Block) bool {
...
-
-			return false
-		}
-	}
-
-	// highest contains seals check if seal for lowest exists
-	for _, seal := range highest.Payload.Seals {
-		if seal.BlockID == lowestID {
-			return true
 		}
```
zhangchiqing, line 277:
```suggestion
		}
		
		// the latest seal must be found in the a block that has Seal when traversing blocks 
		// backwards from higher height to lower height. 
		if len(block.Payload.Seals) > 0 {
			return false
		}		
```

zhangchiqing (approved):
Looks good. Just added one more sanity check here
https://github.com/onflow/flow-go/pull/1755/files#r771628952

PR #1759, opened 2021-12-10
rm duplicate code in getNextAncestryLevel

zhiqiangxu:
The same check is already done at L257.


AlexHentschel (approved):
thanks. good catch

AlexHentschel:
bors merge

PR #1763, opened 2021-12-10
1619/Epochs Integration testing: Verification node

kc1116:
This PR splits current epochs test into separate test runs that use the same core logic, and adds a new SkipReason for resource intensive tests called TEST_RESOURCE_INTENSIVE . 

- Add test run for VN 
- add VN specific network assertion func

This test starts the flow network locally, adds a new VN node, and removes a VN from the networking effectively swapping the nodes. This test requires more resources than dev laptops and current CI machines can provide, because of this the test was developed on a devbox and run multiple times before publishing this PR. This test utilizes the TEST_RESOURCE_INTENSIVE skip reason. 

[Devbox](https://console.cloud.google.com/compute/instancesDetail/zones/us-west1-b/instances/dev-builders-khalil-001?q=search&referrer=search&project=flow-multi-region) and command used to run the test . 

```
TEST_RESOURCE_INTENSIVE=true GO111MODULE=on go test -tags relic -json -count=1 ./tests/epochs
```


integration/tests/epochs/suite.go
```diff
@@ -509,10 +514,22 @@ func (s *Suite) assertNodeApprovedAndProposed(ctx context.Context, env templates
 	require.Containsf(s.T(), proposedTable.(cadence.Array).Values, cadence.String(info.NodeID.String()), "expected new node to be in proposed table: %x", info.NodeID)
 }
 
+// assertNodeNotApprovedOrProposed executes the read approved nodes list and get proposed table scripts
+// and checks that the info.NodeID is in
```
jordanschalm, line 518:
```suggestion
// and checks that the info.NodeID is not included in either - this means the node would be excluded from future epochs
```

integration/tests/epochs/suite.go
```diff
@@ -417,8 +411,19 @@ func (s *Suite) SubmitStakingCollectionCloseStakeTx(
...
-func (s *Suite) SubmitAdminRemoveNodeTx(ctx context.Context,
+
+func (s *Suite) removeNodeFromProtocol(ctx context.Context, env templates.Environment, nodeID flow.Identifier)  {
+	result, err := s.submitAdminRemoveNodeTx(ctx, env, nodeID)
+	require.NoError(s.T(), err)
+	require.NoError(s.T(), result.Error)
+
+	// ensure we submit transaction while in epoch setup phase
+	currView := s.assertInPhase(ctx, flow.EpochPhaseStaking)
+	require.True(s.T(), stakingAuctionViews > currView, "expected remove node transaction to be completed in the epoch staking phase")
```
jordanschalm, line 422: Is the second assertion (line 422) to make sure we are in the _first_ epoch staking phase? Could add a comment if so

integration/tests/epochs/suite.go
```diff
@@ -586,3 +617,20 @@ func (s *Suite) assertNetworkHealthyAfterANChange(ctx context.Context, env templ
...
 }
+
+// assertNetworkHealthyAfterVNChange after an verification node is removed or added to the network
+// this func can be used to perform sanity.
+// 1. Ensure sealing continues by comparing latest sealed block from the root snapshot to the current latest sealed block
+func (s *Suite) assertNetworkHealthyAfterVNChange(ctx context.Context, _ templates.Environment, rootSnapshot *inmem.Snapshot, _ *StakedNodeOperationInfo)  {
+	bootstrapHead, err := rootSnapshot.Head()
+	require.NoError(s.T(), err)
+
+	snapshot, err := s.client.GetLatestProtocolSnapshot(ctx)
```
jordanschalm, line 628: This is returning the [latest finalized state](https://github.com/onflow/flow-go/blob/master/engine/access/rpc/backend/backend.go#L229), so by checking the height of `Head` (the latest finalized block), we aren't guaranteeing anything about the latest sealed block.

We could use [GetLatestBlock](https://github.com/onflow/flow-go/blob/3ec98807fbfa081e5f8613b01d6ed32f30dc4c65/access/api.go#L18) instead, with `isSealed` set to true.

integration/tests/epochs/suite.go
kc1116: I am removing since its only verifying if we are in first epoch staking auction

smnzhu:
Is this ready for review? Seems like a couple of checks are still failing so I wasn't sure

integration/tests/epochs/suite.go
kc1116: https://github.com/onflow/flow-go/pull/1763/commits/d7f322b3587614443f114e7c6bdf40be867acbf5

kc1116:
> Is this ready for review? Seems like a couple of checks are still failing so I wasn't sure

Yes ready for review. The failures are flaky integration test unrelated to this PR

integration/testnet/network.go
```diff
@@ -681,6 +706,13 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
 
 			nodeContainer.bindPort(hostPort, containerPort)
 
+			hostMetricsPort := testingdock.RandomPort(t)
+			containerMetricsPort := "8080/tcp"
+
+			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			nodeContainer.Ports[ColNodeMetricsPort] = hostMetricsPort
+			net.AccessPorts[ColNodeMetricsPort] = hostMetricsPort
```
jordanschalm, line 714: Great addition to make metrics accessible here. 

At the moment this will allocate and bind a host port for metrics for each container, but we will only store/print the ports of one container / node role, and we don't necessarily know _which_ container this is, if we have multiple with the same role). It would be great to be able to see metrics for all containers. 

What do you think about storing the metrics ports separately, keyed by container name (we do this with the `AccessPortsByContainerName` field to have distinguishable GRPC ports for each AN). Then we store a port for every container and can print out the corresponding container name alongside each metrics port. 
```
type FlowNetwork struct {
    // ...
    MetricPortsByContainerName map[string]string
}
```

jordanschalm (approved):
The VN test updates look good now, thank you. Added some suggestions for how we handle exposing the metrics ports.

integration/testnet/network.go
kc1116: Good catch https://github.com/onflow/flow-go/pull/1763/commits/e5f55d1eb315704b2954d82d3800c6695bbb1d8e

integration/testnet/container.go
```diff
@@ -179,8 +179,8 @@ func (c *Container) bindPort(hostPort, containerPort string) {
 
 }
 
-// addFlag adds a command line flag to the container's startup command.
-func (c *Container) addFlag(flag, val string) {
+// AddFlag adds a command line flag to the container's startup command.
+func (c *Container) AddFlag(flag, val string) {
```
smnzhu, line 183: Question: Why did we make this public?

integration/testnet/network.go
```diff
@@ -546,6 +569,8 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
 		flowNetwork.addConsensusFollower(t, rootProtocolSnapshotPath, followerConf, confs)
 	}
 
+	fmt.Print(flowNetwork.MetricsPorts())
```
smnzhu, line 572: See comment above.

smnzhu (approved):

integration/testnet/container.go
kc1116: This snuck in from some testing I was doing updating now

integration/testnet/network.go
kc1116: https://github.com/onflow/flow-go/pull/1763/commits/156ded703dcef22ed689085bad5a019152edb406

integration/testnet/container.go
kc1116: https://github.com/onflow/flow-go/pull/1763/commits/156ded703dcef22ed689085bad5a019152edb406

integration/testnet/network.go
```diff
@@ -243,6 +255,17 @@ func (net *FlowNetwork) ContainerByName(name string) *Container {
 	return container
 }
 
+func (net *FlowNetwork) MetricsPorts() string {
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1763/commits/156ded703dcef22ed689085bad5a019152edb406

Issue #1454, opened 2021-10-06
[State sync engine] Implement prioritization in bitswap


smnzhu:
This issue is tracked in the Bitswap repo here: https://github.com/ipfs/go-bitswap/issues/532

We would like to preserve the bandwidth of Execution Nodes, which are the only creators of new data. One way to do this is to implement some mechanism in bitswap for how we want to prioritize responding to requests for blocks.

Execution nodes should prioritize responding to requests for more recent blocks, to make sure that enough AN's download these blocks

Issue #1451, opened 2021-10-06
[State sync engine] Add new "ExecutionDataCID" field to ExecutionResult


smnzhu:
Add a ExecutionDataCID field to execution result, so that consensus followers can know how to request the state diff for a given block over bitswap.

Each time an execution node computes a new ExecutionData for a block, it can forward it to the `ExecutionDataStorer` to be stored into the blockstore, and include the returned CID in the execution result.

> Note: This may introduce additional latency / CPU usage on execution nodes, we should benchmark to see how much the performance impact will be.

smnzhu:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Flow-Protocol-5a2ad022cee40e25b0885f82/issues/onflow/flow-go/1451?planning-poker) with ZenHub @franklywatson @m4ksio @peterargue @vishalchangrani

PR #1775, opened 2021-12-15
[State Sync] Add execution data cid

smnzhu:
This PR add code to the Execution node to compute the execution data CID

closes https://github.com/onflow/flow-go/issues/1454
closes https://github.com/onflow/flow-go/issues/1451

TODO: 

- [x] Test ExecutionDataCIDCache
- [x] Flow CID to ID conversion functions tests
- [ ] Address minor comments from @Kay-Zee


cmd/execution/main.go
```diff
@@ -123,6 +129,7 @@ func main() {
 			flags.StringVarP(&rpcConf.ListenAddr, "rpc-addr", "i", "localhost:9000", "the address the gRPC server listens on")
 			flags.BoolVar(&rpcConf.RpcMetricsEnabled, "rpc-metrics-enabled", false, "whether to enable the rpc metrics")
 			flags.StringVar(&triedir, "triedir", datadir, "directory to store the execution State")
+			flags.StringVar(&executionDataDir, "execution-data-dir", filepath.Join(homedir, ".flow", "execution_data"), "directory to use for Execution Data blobstore")
```
Kay-Zee, line 132: Somewhat concerned this might get confusing, since we already have a `/execution` folder that holds the mtries. Although maybe `execution` and `execution_data` is sufficiently distinct?

engine/execution/computation/manager.go
```diff
@@ -227,28 +231,50 @@ func (e *Manager) ComputeBlock(
 
 	e.programsCache.Set(block.ID(), toInsert)
 
-	if len(e.uploaders) > 0 {
-		var g errgroup.Group
+	g, uploadCtx := errgroup.WithContext(ctx)
```
Kay-Zee, line 234: I know this is from before, but wish this was more descriptive than `g`

model/flow/execution_result.go
```diff
@@ -14,6 +14,7 @@ type ExecutionResult struct {
 	BlockID          Identifier // commit of the current block
 	Chunks           ChunkList
 	ServiceEvents    ServiceEventList
+	ExecutionDataID  Identifier
```
Kay-Zee, line 17: As this will permanently change the shape of `ExecutionResults` and therefore how their IDs are calculated, i wanted to make sure we've properly though through this change, and that it makes sense.

engine/execution/convert.go
```diff
@@ -70,6 +70,7 @@ func GenerateExecutionResultForBlock(
 	block *flow.Block,
 	chunks []*flow.Chunk,
 	serviceEvents []flow.Event,
+	executionDataId flow.Identifier,
```
Kay-Zee, line 73:
```suggestion
	executionDataID flow.Identifier,
```

nit pick

engine/execution/convert.go
```diff
@@ -87,6 +88,7 @@ func GenerateExecutionResultForBlock(
 		BlockID:          block.ID(),
 		Chunks:           chunks,
 		ServiceEvents:    convertedServiceEvents,
+		ExecutionDataID:  executionDataId,
```
Kay-Zee, line 91:
```suggestion
		ExecutionDataID:  executionDataID,
```

nitpick

module/state_synchronization/execution_data.go
```diff
@@ -9,7 +9,7 @@ import (
 type ExecutionData struct {
 	BlockID            flow.Identifier
 	Collections        []*flow.Collection
-	Events             []*flow.Event
+	Events             []flow.EventsList
```
Kay-Zee, line 12: we're updating this to a 2d list, is that intentional? does it affect the ingestion side from DPS?

engine/execution/computation/manager_test.go
```diff
@@ -111,10 +113,14 @@ func TestComputeBlockWithStorage(t *testing.T) {
 	programsCache, err := NewProgramsCache(10)
 	require.NoError(t, err)
 
+	eds := new(state_synchronization.ExecutionDataService)
+	eds.On("Add", mock.Anything, mock.Anything).Return(flow.ZeroID, nil)
```
Kay-Zee, line 117: For all these mocked functions, i'm kind of tempted to use a non-zero id, and save the flow.ZeroID for error scenarios

model/flow/identifier.go
```diff
@@ -188,3 +191,26 @@ func Sample(size uint, ids ...Identifier) []Identifier {
...
+	}
+
+	if decoded.Length != IdentifierLen {
+		return ZeroID, fmt.Errorf("invalid CID length: %d", decoded.Length)
+	}
+
+	return HashToID(decoded.Digest), nil
+}
+
+func FlowIDToCid(f Identifier) cid.Cid {
```
Kay-Zee, line 213: any easy unit tests for these two new functions?

module/state_synchronization/execution_data.go
smnzhu: Yes, previously the events for a block were flattened, here we've grouped them by Chunk (i believe?). If anything, I'd assume it would make ingestion simpler for DPS since the events would be pre-grouped.

engine/execution/computation/manager_test.go
smnzhu: @Kay-Zee Are you suggesting that we actually compute and return the correct ID? I agree that this would be more ideal, but it means that we'd have to implement the ID computation logic (non-trivial) even though it never actually gets used for anything.

model/flow/identifier.go
smnzhu: Sure, will add.

peterargue (approved):
looks great. thanks @smnzhu

model/flow/execution_result.go
smnzhu: Thanks, the change makes sense, but the thing remaining that I need to check is if this will be populated properly for the root block of a spork.

cmd/execution/main.go
```diff
@@ -318,6 +329,71 @@ func main() {
...
+							if err != nil {
+								return true
+							}
+
+							if ra.BlobTreeRecord.BlockHeight > rb.BlobTreeRecord.BlockHeight {
+								// more recent block has higher priority
+								return true
+							} else if ra.BlobTreeRecord.BlockID == rb.BlobTreeRecord.BlockID {
+								// deeper node in the same blob tree has higher priority
+								return ra.BlobTreeLocation.Height < rb.BlobTreeLocation.Height
```
Kay-Zee, line 373: is this in the right direction? Does deeper mean larger Height or smaller height?

engine/execution/computation/manager_test.go
Kay-Zee: maybe not the correct ID, just not the ZeroID

integration/testnet/network.go
```diff
@@ -61,6 +61,8 @@ const (
 	// DefaultExecutionRootDir is the default directory for the execution node
 	// state database.
 	DefaultExecutionRootDir = "/exedb"
+	// DefaultExecutionDataServiceDir for the execution data service blobstore.
+	DefaultExecutionDataServiceDir = "/data/execution_data"
```
Kay-Zee, line 65: I remember seeing `execution_data_blobstore` above, are the two values suppose to be the same? might make sense to name them the same if they have the same purpose, to make debugging easier

model/flow/identifier_test.go
```diff
@@ -116,3 +118,19 @@ func referenceMerkleRoot(t *testing.T, ids ...flow.Identifier) flow.Identifier {
 	copy(root[:], hash)
 	return root
 }
+
+func TestCIDConversion(t *testing.T) {
```
Kay-Zee, line 122: this is just converting a fixture back and forth right? might make sense just to add a few quick lines to explain

module/state_synchronization/execution_data_cid_cache.go
```diff
@@ -0,0 +1,109 @@
...
+		BlockHeight: header.Height,
+		BlobTree:    blobTree,
+	}
+
+	if len(e.blobTrees) < int(e.size) {
+		e.blobTrees = append(e.blobTrees, blobTreeRecord)
+	} else {
+		evictee := e.blobTrees[e.blobTreeIndex]
+
+		for _, cids := range evictee.BlobTree {
```
Kay-Zee, line 74: this seems like a lot of work to evict one item

module/state_synchronization/execution_data_cid_cache_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestCacheHit(t *testing.T) {
+	cache := state_synchronization.NewExecutionDataCIDCache(10)
+
+	header := unittest.BlockHeaderFixture()
+	var blobTree state_synchronization.BlobTree
+	blobTree = append(blobTree, []cid.Cid{unittest.CidFixture(), unittest.CidFixture()})
+	blobTree = append(blobTree, []cid.Cid{unittest.CidFixture()})
```
Kay-Zee, line 20: any reason the above two lines are separate, and not just one line?

module/state_synchronization/execution_data_cid_cache_test.go
```diff
@@ -0,0 +1,120 @@
...
+			assert.Equal(t, record.BlobTreeLocation.Index, uint(index))
+		}
+	}
+}
+
+func TestCacheMiss(t *testing.T) {
+	cache := state_synchronization.NewExecutionDataCIDCache(10)
+
+	_, err := cache.Get(unittest.CidFixture())
+	assert.Error(t, err)
```
Kay-Zee, line 39: since we have an error to compare to, might be worth making a comparison

module/state_synchronization/execution_data_cid_cache_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	cache.Insert(&header1, [][]cid.Cid{{c}})
+	cache.Insert(&header2, [][]cid.Cid{{c}})
+
+	record, err := cache.Get(c)
+	assert.NoError(t, err)
+	assert.Equal(t, record.BlobTreeRecord.BlockHeight, header2.Height)
+	assert.Equal(t, record.BlobTreeRecord.BlockID, header2.ID())
+	assert.Equal(t, record.BlobTreeLocation.Height, uint(0))
+	assert.Equal(t, record.BlobTreeLocation.Index, uint(0))
```
Kay-Zee, line 90: i guess theres no way to check size of the cache?

Kay-Zee (approved):
A few minor comments, but otherwise LGTM

cmd/execution/main.go
smnzhu: larger index in the array means smaller height

So the root CID would be at the last index in the array

module/state_synchronization/execution_data_cid_cache.go
smnzhu: each CID is stored as a separate entry, and when we evict an entire blob tree we need to evict all CID's in the tree

module/state_synchronization/execution_data_cid_cache_test.go
smnzhu: I can add a method for that

cmd/execution/main.go
```diff
@@ -318,6 +334,71 @@ func main() {
...
 			return compactor, nil
 		}).
+		Component("execution data service", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+			err := os.MkdirAll(executionDataDir, 0700)
+
+			if err != nil {
+				return nil, err
+			}
+
+			ds, err := badger.NewDatastore(executionDataDir, &badger.DefaultOptions)
```
peterargue, line 344: we need to add some logic to close the datastore after the execution data service has stopped. 

One idea I had was to add a chainable method to scaffold that lets you register a function to be called when `Node` calls the `postShutdown` callback.

smnzhu: 23d4538

peterargue: I put a PR together that implements this: https://github.com/onflow/flow-go/pull/1907

PR #1777, opened 2021-12-15
Cleanup: removing legacy `Identities` interface 

AlexHentschel:
**removing [`Identities` interface](https://github.com/onflow/flow-go/blob/4dc8c451a635c5a8b45bf0302a852edb004d6d67/storage/identities.go)** as it seems to be legacy code:
* we don’t have an implementation of the interface (aside from the auto-generated mock);
* the methods are not called anywhere in our code
* we have an `Identities` field in our  [`storage.All` struct](https://github.com/onflow/flow-go/blob/4dc8c451a635c5a8b45bf0302a852edb004d6d67/storage/all.go#L22), but it is never accessed.


jordanschalm (approved):
👌 🧹

zhangchiqing (approved):

PR #1781, opened 2021-12-16
Better printing for keys in errors

m4ksio:
Make printable error properly use slab number
Use hex-encoded key for other cases


fvm/state/state.go
```diff
@@ -320,11 +318,8 @@ func IsFVMStateKey(owner, controller, key string) bool {
...
 	if key[0] == '$' && len(key) == 9 {
-		i := uint64(binary.LittleEndian.Uint64([]byte(key)))
+		i := uint64(binary.BigEndian.Uint64([]byte(key[1:])))
 		return fmt.Sprintf("$%d", i)
 	}
-	if !utf8.ValidString(key) {
-		return strings.ToValidUTF8(key, "?")
-	}
-	return key
+	return fmt.Sprintf("#%x", []byte(key))
```
janezpodhostnik, line 324: What is the reasoning behind a `#` prefix? could we use no prefix or a `0x` prefix?

janezpodhostnik (approved):
LGTM, just had one question.

ramtinms (approved):

fvm/state/state.go
m4ksio: No real reason, just seems like a nice counterpart for `$` and using `0x` would make it look like a number

PR #1782, opened 2021-12-17
[Cryptography] Chacha20-based PRG

tarakby:
This PR mainly replaces the current PRG implementation (Xorshift128+) by a new PRG construction based on Chacha20 stream cipher. It takes over a work done by François a few months ago. 

 - the new PRG is crypto-secure, which is a property needed for future protocol components (safe randomness for block execution).
 - the new PRG offers a better randomness quality.
 - Xorshift is not crypto secure and it is therefore better not to keep it inside the crypto library. 
 
 More updates :
 - remove the multi-PRG construction allowing arbitrary long seeds. The new implementation only accepts seeds of 32 bytes. 
 - add a new interface method `Rand()`. Different PRGs only need to implement the `Rand()` function, which is the core to all other PRG methods. Refactor the implementation accordingly.
 - Improve the uniformity of the outputs from `UintN` by implementing the sample and reject method. 
 - improve tests.
 
 For Chacha 20: 
  - the stream cipher implementation follows https://datatracker.ietf.org/doc/html/rfc7539.
  - the Chacha20-based PRG uses seeds as keys and customizers as nonces (stream IDs). 
  
  ( side unrelated change: replace `.` by `${SRCDIR}` in cgo directives for a better robustness )


crypto/random/chacha20.go
```diff
@@ -0,0 +1,190 @@
...
+// It is recommended to sample the seed uniformly at random.
+// The function errors if the the seed is different than 32 bytes,
+// or if the customizer is larger than 12 bytes.
+// Shorter customizers than 12 bytes are padded by zero bytes.
+func NewChacha20PRG(seed []byte, customizer []byte) (*chachaPRG, error) {
+
+	// check the key size
+	if len(seed) != Chacha20SeedLen {
+		return nil, fmt.Errorf("chacha20 seed length should be %d, got %d", Chacha20SeedLen, len(seed))
+	}
```
AlexHentschel, line 79: I think this check is already implemented by `NewUnauthenticatedCipher` (-> [code](https://github.com/golang/crypto/blob/e495a2d5b3d3be43468d0ebb413f46eeaedf7eb3/chacha20/chacha_generic.go#L81-L83))(?) What is the reason for checking here?

crypto/random/chacha20.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// check the nonce size
+	if len(customizer) > Chacha20CustomizerMaxLen {
+		return nil, fmt.Errorf("new Rand streamID should be %d bytes", Chacha20CustomizerMaxLen)
+	}
```
AlexHentschel, line 84: I think `NewUnauthenticatedCipher` only accepts nonce values of _exactly_ 12 bytes (less are not allowed either) (-> [code](https://github.com/golang/crypto/blob/e495a2d5b3d3be43468d0ebb413f46eeaedf7eb3/chacha20/chacha_generic.go#L92). In contrast, the documentations here seems to indicate that less than 12 bytes would work too (?)

crypto/random/chacha20.go
```diff
@@ -0,0 +1,190 @@
...
+	c.cipher.XORKeyStream(buffer, message)
+	// increase the counter
+	c.bytesCounter += uint64(len(buffer))
+}
+
+// Store returns the internal state of the concatenated Chacha20s
+// This is used for serialization/deserialization purposes.
+func (c *chachaPRG) Store() []byte {
+	bytes := append(c.core.seed[:], c.core.customizer[:]...)
```
AlexHentschel, line 139: I would suggest to pre-allocate a slice of the desired length, so we avoid repeated memory allocation:
```suggestion
	bytes := make([]byte, 0, keySize+nonceSize+8)
	bytes := append(bytes, c.core.seed[:]...)
	bytes := append(bytes, c.core.customizer[:]...)
```

crypto/random/chacha20.go
```diff
@@ -0,0 +1,190 @@
...
+	binary.LittleEndian.PutUint64(counter, c.core.bytesCounter)
+	bytes = append(bytes, counter...)
+	// output is seed || streamID || counter
+	return bytes
+}
+
+// RestoreChacha20PRG creates a chacha20 base PRG based on a previously stored state.
+// The created PRG is restored at the same state where the previous PRG was stored.
+func RestoreChacha20PRG(stateBytes []byte) (*chachaPRG, error) {
+	// inpout should be seed (32 bytes) || streamID (12 bytes) || bytesCounter (8 bytes)
```
AlexHentschel, line 150:
```suggestion
	// input should be seed (32 bytes) || streamID (12 bytes) || bytesCounter (8 bytes)
```

ledger/k_test.go
```diff
@@ -0,0 +1,36 @@
...
+	kp2 := utils.KeyPartFixture(22, "key part 2")
+	kp3 := utils.KeyPartFixture(25, "key part 3")
+	k := ledger.NewKey([]ledger.KeyPart{kp1, kp2, kp3})
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = k.CanonicalForm()
+	}
+	b.StopTimer()
+}*/
```
AlexHentschel, line 36: still needed or can we remove this commented-out code?

AlexHentschel (approved):
LGTM. Clean code with solid tests 👏 
I like your `randCore` interface and `genericPRG` struct.

ledger/k_test.go
tarakby: Yep, all the file was added by mistake

crypto/random/chacha20.go
tarakby: Another good catch, as I [mentioned in the doc](https://github.com/onflow/flow-go/blob/tarak/4067-chacha20/crypto/random/chacha20.go#L73), short customizers are padded to 12 bytes. However the code didn't include the feature! I will add it along with a test.

crypto/random/chacha20.go
tarakby: I added it to get a specific error message about the seed length, rather than the "key" size, which might confuse callers who are using a PRG (seed) and do not know there is a cipher(key) underneath.

crypto/random/rand.go
```diff
@@ -31,11 +42,136 @@ type Rand interface {
...
+// The function panics if input `n` is zero.
+func (p *genericPRG) UintN(n uint64) uint64 {
+	if n == 0 {
+		panic("input to UintN can't be 0")
+	}
+	// count the size of n in bytes
+	size := 0
+	for tmp := n; tmp != 0; tmp >>= 8 {
+		size++
+		p.uintnBuffer[8-size] = 0
```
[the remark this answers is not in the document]
tarakby: That's a left over from another heuristic I tried. You are right it is not needed now that I have the mask operation 👌🏼

crypto/random/rand.go
```diff
@@ -31,11 +42,136 @@ type Rand interface {
...
+type genericPRG struct {
+	randCore
+	// buffer used by UintN function to avoid extra memory allocation
+	uintnBuffer [8]byte
+}
+
+// UintN returns an uint64 pseudo-random number in [0,n-1],
+// using `p` as an entropy source.
+// The function panics if input `n` is zero.
+func (p *genericPRG) UintN(n uint64) uint64 {
```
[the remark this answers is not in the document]
tarakby: Good catch! yes we can stretch further to `n-1` instead of `n` 👌🏼

module/chunks/chunk_assigner.go
```diff
@@ -102,7 +102,7 @@ func (p *ChunkAssigner) rngByBlockID(stateSnapshot protocol.Snapshot) (random.Ra
 		return nil, fmt.Errorf("failed to retrieve source of randomness: %w", err)
 	}
 
-	rng, err := random.NewRand(seed)
+	rng, err := random.NewChacha20PRG(seed, []byte("chunk_assign"))
```
ramtinms, line 105: maybe we define "chunk_assign", "net_topology" as global constants somewhere in model files

ramtinms (approved):
Looks good to me, just some a minor question and a minor suggestions

crypto/random/chacha20.go
```diff
@@ -0,0 +1,194 @@
...
+// It is recommended to sample the seed uniformly at random.
+// The function errors if the seed is different than 32 bytes,
+// or if the customizer is larger than 12 bytes.
+// Shorter customizers than 12 bytes are padded by zero bytes.
+func NewChacha20PRG(seed []byte, customizer []byte) (*chachaPRG, error) {
+
+	// check the key size
+	if len(seed) != Chacha20SeedLen {
+		return nil, fmt.Errorf("chacha20 seed length should be %d, got %d", Chacha20SeedLen, len(seed))
+	}
```
zhangchiqing, line 79: Can we restrict the type and eliminates the error case here?
```suggestion
func NewChacha20PRG(seed [Chacha20SeedLen]byte, customizer []byte) (*chachaPRG, error) {
```


or 
```
func NewChacha20PRG(seed [32]byte, customizer []byte) (*chachaPRG, error) {
```

crypto/random/rand_test.go
```diff
@@ -1,263 +1,405 @@
...
+		distribution := make([]float64, listSize)
+		testElement := mrand.Intn(listSize)
+		// Slice to shuffle
+		list := make([]int, 0, listSize)
+		for i := 0; i < listSize; i++ {
+			list = append(list, i)
+		}
+
+		for i := 0; i < sampleSize; i++ {
+			err = rng.Shuffle(listSize, func(i, j int) {
```
zhangchiqing, line 238: The test here run for 100000 times, and each time it will shuffle the list, and pick an item at the same index, to check if the picked 100000 items are distributed evenly across all elements in the list.

My question is why do we reuse the shuffled list and shuffle again on top of the shuffled list, rather than reset the list to start from `1` to `100`, and shuffle again. 

Technically, when user uses the shuffle API, they would only call Shuffle once, rather than shuffle the list for X times. So they would expect the shuffled list is random after one time shuffling. So I think the test should also test the property under the same scenario. 

It also applies to the testing to other methods, such as `Permutation`, `SubPermutation`

zhangchiqing (approved):
Looks good. Just some minor comments.

crypto/random/rand.go
```diff
@@ -31,11 +42,136 @@ type Rand interface {
...
+// O(n) space and O(n) time.
+func (p *genericPRG) Permutation(n int) ([]int, error) {
+	if n < 0 {
+		return nil, fmt.Errorf("population size cannot be negative")
+	}
+	items := make([]int, n)
+	for i := 0; i < n; i++ {
+		j := p.UintN(uint64(i + 1))
+		items[i] = items[j]
+		items[j] = i
```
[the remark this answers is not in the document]
tarakby: This is an optimized variant of Fisher-Yates when initializing and shuffling is happening at the same time, you can check it [here](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_%22inside-out%22_algorithm). 
The "classic" way would be to initialize all the array from `0` to `n-1` and then to shuffle as you said, but that would take more steps.

module/chunks/chunk_assigner.go
tarakby: updated in https://github.com/onflow/flow-go/pull/1782/commits/a82b88514fab186d229fa133efa70ee46eb974f0

crypto/random/chacha20.go
tarakby: That is possible, though it means we will have the copy the seed slice into an array to pass it to `NewChacha20PRG` each time in the protocol code.
I'm planning to make changes to the protocol in a follow-up issue, so I added an item for this: https://github.com/dapperlabs/flow-go/issues/6107

crypto/random/rand_test.go
tarakby: Both experiments (the implemented one or the suggested one) should give the same result if the outcome is uniformly distributed. The experiment should be: 
 1. start from any permutation of the array
 2. shuffle using the shuffle to test
 3. check what's in the some constant index
 What you're suggesting is to always start from the same specific permutation in step 1 (which can be [1,2,3..,100]). While what I implemented is to start from a different permutation each time. Instead of choosing a random permutation, I just re-use the latest permutation as an optimization. 
 Both ways must give the a uniform result for the implementation to be valid. I can add another the suggested way too.

tarakby: added in https://github.com/onflow/flow-go/pull/1782/commits/60d5d19df1351645725a780c4c1f4c1ee536f17b

tarakby:
bors merge

tarakby:
bors merge

PR #1792, opened 2021-12-21
Make hex address parsing consistent with SDK

jordanschalm:
Previously `flow-go` would not correctly handle un-padded hex addresses. This format is rarely used, but was used until recently by the [default string representation of addresses](https://github.com/onflow/cadence/issues/1131) in Cadence, which was [directly used](https://github.com/dapperlabs/flow-mgmt/blob/master/sporking/machine-accounts/main.go#L145) when creating internal machine account configs, resulting in some configs having the un-padded address representation. 

This change avoids misleading error messages for these particular machine account configs, and makes the implementations consistent with the SDK.


model/flow/address_test.go
```diff
@@ -15,6 +15,27 @@ type addressWrapper struct {
...
+	type testCase struct {
+		literal string
+		value   []byte
+	}
+
+	for _, test := range []testCase{
+		{"123", []byte{0x1, 0x23}},
+		{"1", []byte{0x1}},
+		// leading zero
+		{"01", []byte{0x1}},
```
Kay-Zee, line 29: Should we add `0001`? I do think i've seen an address that started with three zeros though https://flowscan.org/account/0x0001dfd237612b1e

Kay-Zee (approved):
One small comment, although the code does match the SDK.

tarakby (approved):

PR #1793, opened 2021-12-21
Flaky Test Monitor - Level 3 Summary

gomisha:
Processes a level 2 summary JSON and produces level 3 summary JSON which summarizes:

1. tests with no-results (ordered by most no-results)
2. tests with failures (ordered by most failures)
3. tests with durations > 0 (ordered by longest durations)

We can also further control what data is returned so we can set minimum thresholds for failure rate and duration. This will allow us to focus on the tests with the most failures and the longest duration:

1) minimum failure rate (e.g. give me top 10 tests with failure rate >= 50%)
2) minimum duration (e.g. give me top 5 tests with duration >= 5 seconds)

We would still get the TOTAL # of tests that satisfy the min failure rate or duration but it will be reported as a number

Sample parameters we can set:

1) return top 5 longest tests with duration of at least 0.25 seconds and total count (can be larger than 5) of tests with duration above 0.25 seconds

2) return top 10 failed tests (with highest failure rate) with failure rate at least 25% and get total count (can be larger than 10) of tests with failure rate at least 25%


tools/flaky_test_monitor/level3/process_summary3_results_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	for durationIndex := range expectedTestSummary3.LongestRunning {
+		common.AssertTestSummariesEqual(t, expectedTestSummary3.LongestRunning[durationIndex], actualTestSummary3.LongestRunning[durationIndex])
+	}
+
+	require.Equal(t, expectedTestSummary3, actualTestSummary3)
+}
+
+var expectedPanicFunc_WrongPath = func() {
```
Kay-Zee, line 71: I think i usually prefer the contents to be defined in the test itself, especially if it's only used in the test.

this can be moved into the context of the test, e.g. 
```
func TestProcessSummary3TestRun_Panic_WrongPath(t *testing.T) {
        unknownPath := "foobar"
	require.PanicsWithValue(t, "error reading level 2 json: open foobar: no such file or directory",
		func() {processSummary3TestRun(unknownPath)})
}
```

or if you prefer:

```
func TestProcessSummary3TestRun_Panic_WrongPath(t *testing.T) {
        expectedPanicFuncWrongPath := func() {processSummary3TestRun("foobar")}
	require.PanicsWithValue(t, "error reading level 2 json: open foobar: no such file or directory",
		expectedPanicFuncWrongPath)
}
```
or something similar

Kay-Zee (commented):
A few minor comments.

Just one concern around the actual summary format itself.

tools/flaky_test_monitor/level3/process_summary3_results.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(testSummary2.TestResults) == 0 {
+		panic("invalid summary 2 file - no test results found")
+	}
+
+	// create lists to keep track of 3 main things
+	// 1. tests with no-results (ordered by most no-results)
+	// 2. tests with failures (ordered by most failures)
+	// 3. tests with durations > 0 (ordered by longest durations)
+
+	var noResultsTRS []common.TestResultSummary = []common.TestResultSummary{}
```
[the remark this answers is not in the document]
gomisha: Updated https://github.com/onflow/flow-go/pull/1793/commits/4d9793c47d209e0de2a222e61df92e35bc51e001

tools/flaky_test_monitor/level3/process_summary3_results_test.go
gomisha: Updated https://github.com/onflow/flow-go/pull/1793/commits/b0f77ac8aa498e958bff318758fa9deda024ba4c

tools/flaky_test_monitor/level3/process_summary3_results.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		if trs.AverageDuration > 0 {
+			durationTRS = insertSortedByDuration(durationTRS, *trs)
+		}
+	}
+
+	var testSummary3 common.TestSummary3
+	testSummary3.NoResults = noResultsTRS
+	testSummary3.MostFailures = failuresTRS
+	testSummary3.LongestRunning = durationTRS
```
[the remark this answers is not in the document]
gomisha: Aggregation happens at level 2 (all results for each test over many test runs are aggregated from level 1). Level 3 is to sort and group all the aggregated results from level 2. As for capping the number of results, yes, we can have that. A property file can dictate max results for most failures and longest running.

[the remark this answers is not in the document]
Kay-Zee: Yeah, i meant more aggregation of the level 2 data, i.e. now many are in each array, and if there is any more info we can glean from combining data within those arrays, but the total "number" of items in each group is only an interesting aggregation if we don't return everything, as other wise we can just use the length of the slice.

> A property file can dictate max results for most failures and longest running.
Is this something that's planned for a future PR I guess? that works for me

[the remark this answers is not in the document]
gomisha: Below is a proposed property file and level 3 JSON changes that will allow us to specify:

1) minimum thresholds for failure rate and duration (e.g. only find tests with duration longer than 1.0 second)
2) max slice size for failures and duration (e.g. get top 10 most failed tests)

This allows us to do scenarios like:

a) find all the tests with durations longer than 1.0 seconds and return the top 10 tests - we would know the top 10 longest tests that are longer than a second and also know that there are n tests in total (e.g. 50) that are longer than 1.0 seconds

b) find all the tests with failure rate of at least 40% and return the top 5 tests - we would know the top 5 tests with failure rates of at least 40% and also know that there are x tests in total (e.g. 25) that have a failure rate of at least 40%

Level 3 JSON will have 2 new fields to add additional aggregation data that isn't viewable from just the slices:

`"longest_running_total": 50`
`"most_failures_total": 25,`

flaky-test-monitor.properties:

```
# includes tests that have a failure rate of at least the specified percentage
# e.g. 0.40 == 40%, 0.75 == 75%
failures_threshold_percent=0.40

# max number of test failures to return in slice that's at least `failures_threshold_percent`
# does not impact total number of tests with failure rate at least `failures_threshold_percent` which will be reported separately
failures_slice_max=10

# include tests that are at least this many seconds long
# this is meant to filter out the tests that aren't as long running
duration_threshold_seconds=1.0

# max number of test durations to return in slice that's at least `duration_threshold_seconds` long
# does not impact total durations at least `duration_threshold_seconds` long which will be reported separately
duration_slice_max=10
```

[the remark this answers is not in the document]
gomisha: This is now implemented - we can now get data like this:

1) return top 5 longest tests with duration of at least 0.25 seconds and total count of tests with duration above 0.25 seconds

2) return top 5 failed tests (with highest failure rate) with failure rate at least 25% and get total count of tests with failure rate at least 25%

tools/flaky_test_monitor/common/config.go
```diff
@@ -0,0 +1,76 @@
...
+	FailureThresholdPercent float32
+	FailuresSliceMax        int
+
+	DurationThresholdSeconds float32
+	DurationSliceMax         int
+}
+
+func ReadProperties(directory string) Config {
+	// looks for properties file with specific name
+	propertyFile := AddTrailingSlash(directory) + "flaky-test-monitor.properties"
```
Kay-Zee, line 29: Could probably use https://pkg.go.dev/path/filepath#Join

tools/flaky_test_monitor/common/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+	file, err := os.Open(propertyFile)
+
+	AssertNoError(err, "error opening property file")
+	defer file.Close()
+
+	properties := map[string]string{}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
Kay-Zee, line 38: a little surprised we decided to write our own parser here, thought there'd be some lib we'd be able to use for this.

tools/flaky_test_monitor/common/config.go
```diff
@@ -0,0 +1,76 @@
...
+		line := scanner.Text()
+
+		// if line starts with # or //, treat that as a comment and skip it
+		if strings.HasPrefix(line, "#") || strings.HasPrefix(line, "//") {
+			continue
+		}
+
+		// check that line has "="
+		equalIndex := strings.Index(line, "=")
+		if equalIndex > 0 {
```
Kay-Zee, line 49: Doesn't handle the case where there is multiple `=` signs, but for now, since we're mostly doing numerical values for config, this is probably fine.

Could consider using `strings.Split`

tools/flaky_test_monitor/common/level3.go
```diff
@@ -0,0 +1,24 @@
...
+	MostFailuresTotal int `json:"most_failures_total"`
+
+	// ordered list of tests (from level 2 summary) that:
+	// a) met minimum duration threshold as specified in property file property `duration_threshold_seconds`
+	// b) is up to maximum slice size as specified in `duration_slice_max` in property file
+	LongestRunning []TestResultSummary `json:"longest_running"`
+
+	// total # of tests (from level 2 summary) that:
+	// a) met minimum duration threshold as specified in property file property `duration_threshold_seconds`
+	LongestRunningTotal int `json:"longest_running_total"`
```
Kay-Zee, line 23: :+1:

tools/flaky_test_monitor/level3/flaky-test-monitor.properties
```diff
@@ -0,0 +1,21 @@
+# property file for Flaky Test Monitor
+
+##############################
+#####   LEVEL 3  #############
+##############################
+
+# includes tests that have a failure rate of at least the specified percentage
+# e.g. 0.40 == 40%, 0.75 == 75%
+failures_threshold_percent=0
```
Kay-Zee, line 9: do we want 0 as a default here?

tools/flaky_test_monitor/level3/flaky-test-monitor.properties
```diff
@@ -0,0 +1,21 @@
...
+
+# max number of test failures to return in slice that's at least `failures_threshold_percent`
+# does not impact total number of tests with failure rate at least `failures_threshold_percent` which will be reported separately
+failures_slice_max=10
+
+# include tests that are at least this many seconds long
+# this is meant to filter out the tests that aren't as long running
+duration_threshold_seconds=0
```
Kay-Zee, line 17: do we want 0 as a default here?

Kay-Zee (approved):

tools/flaky_test_monitor/level3/flaky-test-monitor.properties
gomisha: Good point. I increased it to 0.05 seconds

https://github.com/onflow/flow-go/pull/1793/commits/7ca42b8a930036d6e038e08ad342650a254e482a

tools/flaky_test_monitor/level3/flaky-test-monitor.properties
gomisha: Good catch. We want to filter out tests with 0% failure rate. I increased this to 0.01 (1%)

https://github.com/onflow/flow-go/pull/1793/commits/7ca42b8a930036d6e038e08ad342650a254e482a

tools/flaky_test_monitor/common/config.go
gomisha: Yeah, I thought about using a lib, but I didn't want the overhead. This property file parser is only about 15 lines if you just take into account the "generic parsing" it's doing. It's also doing a bunch of Flaky Test Monitor specific parsing which we'd need to do with an off-the-shelf library, too.

tools/flaky_test_monitor/common/config.go
gomisha: Good idea. I changed all occurrences of file path building to use it.

https://github.com/onflow/flow-go/pull/1793/commits/d59bce00cfd268deebe47b3dcb7d0e4776ef6abe

tools/flaky_test_monitor/common/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+		// if line is empty or doesn't follow the `property=value` format, skip it
+		property := strings.Split(line, "=")
+		if len(property) != 2 {
+			continue
+		}
+
+		// add property and trim any whitespace from property name and value
+		properties[strings.TrimSpace(property[0])] = strings.TrimSpace(property[1])
+	}
```
smnzhu, line 54: .properties is a java file format, why are we using it here? Let's use json or yaml instead, there are built in parsers for these formats

tools/flaky_test_monitor/level3/process_summary3_results.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// total # of long tests that satisfy min duration threshold
+	testSummary3.LongestRunningTotal = len(durationTRS)
+
+	// check if # of durations exceeded max durations to return
+	if len(durationTRS) > config.DurationSliceMax {
+		// truncate slice to return only the first config.DurationSliceMax durations
+		durationTRS = durationTRS[:config.DurationSliceMax]
+	}
+	testSummary3.LongestRunning = durationTRS
```
smnzhu, line 75: I feel like we can make this logic way simpler with https://pkg.go.dev/sort#Slice.

We can just sort the level 2 results multiple different ways, and output the JSON each time to a separate file. This also makes it a lot easier to add different sorting criteria imo.

smnzhu: And we won't need any of these helper functions below

tools/flaky_test_monitor/common/level3.go
```diff
@@ -0,0 +1,24 @@
+package common
+
+// models full level 3 summary of a test run from 1 (single) level 2 test run
+type TestSummary3 struct {
+	NoResults []TestResultSummary `json:"no_results"`
+
+	// ordered list of tests (from level 2 summary) that:
+	// a) met minimum failure threshold as specified in property file property `failures_threshold_percent`
+	// b) is up to maximum slice size as specified in `failures_slice_max` in property file
```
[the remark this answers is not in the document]
gomisha: You're saying that if want the "top x failed tests" that would automatically gives us the tests with the highest failure rate, making the min failure threshold redundant?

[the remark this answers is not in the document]
gomisha: It might be useful to know the total # of tests that have a certain failure threshold but not actually return all of them (e.g. we might have 50 tests with failure rate >= 30% but we want to return just the top 10). This was something I discussed with @Kay-Zee 

It might be useful to know how many tests in total we have above a certain failure rate. Otherwise, we're always returning the "top 10" failed tests but we don't know how many tests we have in total above a certain failure threshold.

tools/flaky_test_monitor/level3/flaky-test-monitor.properties
```diff
@@ -0,0 +1,21 @@
+# property file for Flaky Test Monitor
```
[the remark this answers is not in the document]
gomisha: Yes, good idea. This simplified the code.

https://github.com/onflow/flow-go/pull/1793/commits/d60e920103e573d4bb6efe5ec4a4582a7795248b

tools/flaky_test_monitor/common/config.go
gomisha: Changed to use json.

https://github.com/onflow/flow-go/pull/1793/commits/d60e920103e573d4bb6efe5ec4a4582a7795248b

tools/flaky_test_monitor/level3/process_summary3_results.go
gomisha: Agreed. Updated to use `sort.Slice()`

https://github.com/onflow/flow-go/pull/1793/commits/45f2adf7f0396d0624d35112d28b5598179eb0b1

tools/flaky_test_monitor/common/level3.go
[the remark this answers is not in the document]
Kay-Zee: yeah, I specifically requested this, since I felt like i only wanted details about the top 10 or so of each category, but i also wanted to know if there was something like 50 tests that matched the condition.

we do get two specific pieces of data, (e.g. top 10 worst offenders, and total number of offenders) so i don't think the two are redundant

smnzhu (approved):

PR #1794, opened 2021-12-21
Avoid double-instantiating hotstuff cluster metrics

jordanschalm:
Uses the same hotstuff metrics instance for all components of a given epoch. Previously we would instantiate two of the same instances separately, which isn't allowed by `promauto`


durkmurder (approved):
Thank you, looks great

kc1116 (approved):

PR #1795, opened 2021-12-22
Remove more V1 deprecated modules

tarakby:
This PR addresses https://github.com/dapperlabs/flow-go/issues/6068.

It removes:
- module/Verifier interface
- module/signature/AggregationVerifier struct
- module/signature/AggregationProvider struct
- crypto/thresholdSigner struct

It replaces the `Verifier.Verify` interface by a call to `PublicKey.Verify`.  `Verifier` is a thin layer with only one method `Verify` that is equivalent to the `crypto.PublicKey.Verify` method. 
The `Verifier` struct used in protocol modules is replaced by a signature hasher or simply by the signature tag used by the hasher.


Makefile
```diff
@@ -118,6 +118,7 @@ generate-mocks:
 	GO111MODULE=on mockery -name 'ComputationManager' -dir=engine/execution/computation -case=underscore -output="engine/execution/computation/mock" -outpkg="mock"
 	GO111MODULE=on mockery -name 'EpochComponentsFactory' -dir=engine/collection/epochmgr -case=underscore -output="engine/collection/epochmgr/mock" -outpkg="mock"
 	GO111MODULE=on mockery -name 'ProviderEngine' -dir=engine/execution/provider -case=underscore -output="engine/execution/provider/mock" -outpkg="mock"
+	(cd ./crypto && GO111MODULE=on mockery -name 'PublicKey' -case=underscore -tags="relic" -output="../module/mock" -outpkg="mock")
```
tarakby, line 121: For some reason, running this command from the root causes issues related to cgo and the `C` package. Getting into `./crypto` solved the issue.

durkmurder (commented):
Looks good I am just not sure about passing tag in places where we have previously passed `module.Verifier`. I am tempted to pass the `hash.Hasher` instead but not really sure if that's the best option. Would like to hear your thoughts `string` vs `hash.Hasher`. My main motivation is that hasher provides more context than string.

tarakby:
> Would like to hear your thoughts string vs hash.Hasher. My main motivation is that hasher provides more context than string.

The reason I chose the tag and not the hasher is that the the hasher choice depends on the signature algorithm (here BLS requires the KMAC hasher, but other signature algorithms require other hashers) while the domain tag doesn't. 
Since the modules are high level and should be agnostic of the signature algorithm, I chose to make them only dependent on a domain tag.
Happy to hear your thoughts and adjust. @durkmurder

module/validation/receipt_validator_test.go
```diff
@@ -22,41 +22,54 @@ type ReceiptValidationSuite struct {
 	unittest.BaseChainSuite
 
 	receiptValidator module.ReceiptValidator
-	verifier         *mock2.Verifier
+	sigTag           string
+	publicKey        *mock2.PublicKey
```
durkmurder, line 26: let's avoid `mock2` and provide a meaningful alias

tarakby:
>But that's kinda what I am confused about. If the module is agnostic to signing algorithm then it should inject the hasher but instead it creates it which kinda bounds it to concrete algorithm.

The module implementation itself is not agnostic. If we use `Hasher` in the interface, the caller who uses the modules will need to understand what signatures the module implementation is using, and create a hasher that is compatible with the signature. Using the domain tag in the interface means we abstract this cryptographic complexity from the caller.

durkmurder:
> >But that's kinda what I am confused about. If the module is agnostic to signing algorithm then it should inject the hasher but instead it creates it which kinda bounds it to concrete algorithm.
> 
> 
> 
> The module implementation itself is not agnostic. If we use `Hasher` in the interface, the caller who uses the modules will need to understand what signatures the module implementation is using, and create a hasher that is compatible with the signature. Using the domain tag in the interface means we abstract this cryptographic complexity from the caller. 

Agree, makes sense, but if we go this far maybe we should just embed domain separation tag into the implementation ? For instance seal validator or receipt validator always use the same values.

zhangchiqing (approved):
Nice clean up 👍

tarakby:
>but if we go this far maybe we should just embed domain separation tag into the implementation ?

Yes this is possible too 👌🏼

durkmurder:
> > but if we go this far maybe we should just embed domain separation tag into the implementation ?
> 
> Yes this is possible too 👌🏼

Then lets do this, no need to pass it as parameter

durkmurder (approved):
Looks great, thank you for this cleanup.

tarakby:
bors merge

PR #1799, opened 2022-01-03
[Consensus Voting] CI Integration tests

durkmurder:
### What was done

This PR enables back previously deactivated integration tests that were used on CI.
Most of the issues were related to incorrect initialization which is not captured by other tests. 
I have left comments to explain changes. 

Also as part of this PR I have removed `Combiner` which is not used anymore


cmd/consensus/main.go
```diff
@@ -614,36 +614,6 @@ func main() {
 
 			return aggregator, nil
 		}).
-		Module("compliance engine", func(builder cmd.NodeBuilder, node *cmd.NodeConfig) error {
```
durkmurder, line 617: modules are initialized before components, but compliance engine initialized depends on `hotstuffModules` which can be initialized only as component(because it uses network which is a component)

consensus/hotstuff/packer/packer.go
```diff
@@ -0,0 +1,54 @@
+package packer
```
durkmurder, line 1: I had to extract this logic into separate package to allow using it from other places. There is a hard circular dependency if we try to use `consensus/hotstuff/signature` in `state/protocol/bader/seed`. Extracting just RLP encoding solves the problem.

zhangchiqing (commented):
Looks good. Thanks for the fixes.

I made a PR to add an integration tests to verify the usage of pack and unpack. Also added some comments to remind us some work to be done in v3

zhangchiqing (approved):

tarakby (approved):

zhangchiqing:
bors merge

Issue #1786, opened 2021-12-17
[Merkle'ized Event Proofs] Revise Payload trie to allow for safe inclusion proofs


AlexHentschel:
At the moment, our [`merkle.Tree`](https://github.com/onflow/flow-go/blob/d69930a2472c99c7e51e58155a0de46022b267e7/storage/merkle/tree.go) is geared to provide a set commitment for an _entire_ set. However, it is not safe for inclusion / exclusion proofs of _individual_ elements. Furthermore, the trie is not safe for storing elements with different key lengths, but the API does not reflect this limitation and the implementation does not check.

### Goal: 
* update the trie's graph structure to support safe inclusion and exclusion proofs
* add safety checks to enforce consistent key length

AlexHentschel:
implemented by PR https://github.com/onflow/flow-go/pull/1800

tarakby:
I guess we can close this issue? @AlexHentschel

PR #1800, opened 2022-01-03
Payload Patricia Trie Security fixes 

AlexHentschel:
implements https://github.com/onflow/flow-go/issues/1786

## This PR addresses a couple security problems with the Patricia Tree implementation:

Fundamentally, the `Tree` implements cryptographic commitment scheme to a _set_ of key-value pairs. 

1. Currently, the values are not included in the cryptographic hash of the tree: https://github.com/onflow/flow-go/blob/4bf22f6bb2f6568005a6fd965a5b4a5dbd4a920a/storage/merkle/tree.go#L414-L416
2. The current implementation is vulnerable to a hash-collision attacks: It is possible to replace a full node with a short node. Therefore, the current implementation is not safe for providing inclusion or non-inclusion proofs. The current `Tree` is safe to verify cryptographic integrity when obtaining the _full set_ of key-value pairs, but this defeats the purpose of using a merkle trie. For the full _set of key-value pairs_, a simple hash would be sufficient as a commitment. 
3. The current trie only works correctly with keys that have all identical length. However, this requirement was not checked in the code and not reflected by the API (accepting byte _slices_ as keys).  
4. The current trie only works correctly for key with _at most_ 256 bits. However, this requirement was not checked in the code and not reflected by the API (accepting byte _slices_ as keys).  

### This PR addresses some code-clarity issues:

* refactored node structures to be objects each providing their own hash functions, thereby removing unnecessary `switch` statement for hash computation
* to simplify enforcing cryptographic security, we introduce the convention that hashing a `nil` node is an illegal operation, which panics

# Overview of structural changes
* The Payload trie now works with a larger key space of up to 8192 bytes (-> [code](https://github.com/onflow/flow-go/blob/c781aaf61b5a136c29bbf7546d46a93d82b96843/storage/merkle/tree.go#L16-L25))
   * Larger key space is useful for event proves, to provide pre-filtered event sets to the client plus correctness proves. This reduces bandwidth (client doesn't have to download _all_ events for the block anymore), but requires merkle trees with a suitable internal structure using a larger key space. 
* Before adding a new key-value pair to the trie, the trie now checks whether the key has the same length as the already-stored pairs and returns dedicated sentinel errors otherwise. 
* I have replaced the `jrick/bitset` implementation for bit-vectors by our [`ledger/common/bitutils`](https://github.com/onflow/flow-go/blob/c781aaf61b5a136c29bbf7546d46a93d82b96843/ledger/common/bitutils/utils.go) implementation (also used by the mTrie):
  * `bitutils` uses a more intuitive and simpler convention for indexing bits in a `byte` slice. Theoretically, either indexing convention works, but for constructing test cases, `bitutils` is way easier to work with 
  * `bitutils` works directly on byte slices (the native representation of keys), thereby removing the need for repetitive type conversion
*  I added a [python reference implementation](https://github.com/dapperlabs/flow-internal/blob/56102dbdd5238e7f2bd19c54e093831f25196fb4/reference_implementations/patricia_tree.py) for some basic consistency checks. 

### Relation to earlier PRs

_This PR encompasses the improvements from https://github.com/onflow/flow-go/pull/1580_. I made an earlier attempt of amending PR https://github.com/onflow/flow-go/pull/1580 via ~PR https://github.com/onflow/flow-go/pull/1678~, which I accidentally merged while still being incomplete. To keep a cleaner commit history for master, I have now consolidated all changes into this PR targeting master directly. 

@tarakby provided many review comments on my failed ~PR https://github.com/onflow/flow-go/pull/1678~. They are all addressed in this PR. I have also provided comments in ~PR https://github.com/onflow/flow-go/pull/1678~ what changes I made in response to the review comments.


ledger/common/bitutils/utils.go
```diff
@@ -12,8 +12,18 @@ func Bit(b []byte, idx int) int {
...
+func SetBit(b []byte, idx int, value int) {
+	byteIndex := idx >> 3
+	idx &= 7
+	mask := byte(1 << (7 - idx))
+	if value == 0 {
+		b[byteIndex] &= ^mask
+	} else {
+		b[byteIndex] |= mask
+	}
+}
```
ramtinms, line 24: value as input might not be trival on what it means, It might be better to keep both function SetBit and SetBitWithValue even if the first one uses the second one, the use of input of 1 for value param in the ledger might be not easy to read or understand.

ramtinms: also user might pass 2 and it might be cause missusage

storage/merkle/node.go
```diff
@@ -3,23 +3,110 @@
 package merkle
 
 import (
-	"github.com/onflow/flow-go/crypto/hash"
+	"golang.org/x/crypto/blake2b"
```
ramtinms, line 6: hmm, this seems like one of the changes that needs @tarakby's approval.

storage/merkle/node.go
```diff
@@ -3,23 +3,110 @@
...
+// The node represents a generic vertex in the sparse Merkle trie.
+// CONVENTION:
+//  * This implementation guarantees that within the trie there are
+//    _never_ nil-references to descendants. We completely avoid this by
+//    utilizing `short nodes` that represent any path-segment without branching.
+type node interface {
+	// Hash computes recursively the hash of this respective sub trie.
+	// To simplify enforcing cryptographic security, we introduce the convention
+	// that hashing a nil node is an illegal operation, which panics.
+	Hash() []byte
```
ramtinms, line 18: thats a wise idea

storage/merkle/node.go
```diff
@@ -3,23 +3,110 @@
...
-	left  node // holds the left path node (bit 0)
-	right node // holds the right path node (bit 1)
+	left  node // holds the left path node (bit 0); never nil
+	right node // holds the right path node (bit 1); never nil
+}
+
+var _ node = &full{}
+
+func (n *full) Hash() []byte {
+	h, _ := blake2b.New256(fullNodeTag) // blake2b.New256(..) error for given MAC (verified in tests)
```
ramtinms, line 76: Even if the lenght is fixed for full nodes It might be valuable to include `serializedPathSegmentLength` to be consistent with the short nodes.

ramtinms (approved):
Looks good to me, just some minor comments

ledger/common/bitutils/utils_test.go
```diff
@@ -46,16 +58,20 @@ func TestBitTools(t *testing.T) {
...
-		for i := 0; i < maxBits; i++ {
+		for idx := 0; idx < maxBits; idx++ {
 			bit := rand.Intn(2)
 			// b = 2*b + bit
 			b.Lsh(&b, 1)
 			b.Add(&b, big.NewInt(int64(bit)))
 			// sets the bit at index i
 			if bit == 1 {
-				SetBit(bytes, i)
+				SetBit(bytes, idx, 1)
```
Kay-Zee, line 72: any reasion this isn't just `SetBit(bytes, idx, bit)`?

ledger/common/bitutils/utils.go
```diff
@@ -12,8 +12,18 @@ func Bit(b []byte, idx int) int {
 // SetBit sets the bit at position i in the byte array b
```
Kay-Zee, line 12:
```suggestion
// SetBit sets the bit at position i in the byte array b to 0 if the value is 0, otherwise, sets it to 1
```

ledger/common/bitutils/utils.go
Kay-Zee: or add the described behaviour to the function comment

storage/merkle/tree.go
```diff
@@ -131,31 +192,27 @@ PutLoop:
...
 			// if we have reached the end of the key, insert the new value
-			totalCount := uint(len(key)) * 8
+			totalCount := len(key) * 8
 			if index == totalCount {
+				// Instantiate a new leaf holding a _copy_ of the provided key-value pair,
+				// to protect the slices from external modification.
 				*cur = &leaf{
-					key: key,
-					val: val,
+					val: append(make([]byte, 0, len(val)), val...),
```
Kay-Zee, line 206: :+1:

storage/merkle/tree.go
```diff
@@ -131,31 +192,27 @@ PutLoop:
 
 		// if we have a leaf node, we reached a non-empty leaf
 		case *leaf:
-
-			// replace the current value with the new one
-			n.val = val
-
-			// return true to indicate that we overwrote
-			return true
+			n.val = append(make([]byte, 0, len(val)), val...)
```
Kay-Zee, line 195: :+1:

Kay-Zee (approved):
I'm approving the addition of the value into the trie, and the removal of storing the key in the leaf nodes, rendering the original issue from https://github.com/onflow/flow-go/pull/1580 no longer relevant.

Also added a few comments, plus took a detailed look over the tests which look good. Have not run the reference implementation myself, so assuming the values copied over are accurate.

storage/merkle/tree.go
```diff
@@ -348,76 +407,41 @@ DelLoop:
...
 	return true
 }
 
-// Hash will return the root hash of this patricia merkle tree.
+// Hash returns the root hash of this patricia merkle tree.
 func (t *Tree) Hash() []byte {
-	hash := t.nodeHash(t.root)
-	return hash
+	if t.size == 0 {
+		return []byte{14, 87, 81, 192, 38, 229, 67, 178, 232, 171, 46, 176, 96, 153, 218, 161, 209, 229, 223, 71, 119, 143, 119, 135, 250, 171, 69, 205, 241, 47, 227, 168}
```
tarakby, line 428: I think it would be cleaner to have the empty root hash as a constant.

storage/merkle/tree.go
```diff
@@ -348,76 +407,41 @@ DelLoop:
...
-	for i := uint(0); i < p.count; i++ {
-		totalPath.SetBool(int(i), parentPath.Get(int(i)))
+	totalPath := bitutils.MakeBitVector(totalCount)
+	for i := 0; i < p.count; i++ {
+		bitutils.SetBit(totalPath, i, bitutils.Bit(p.path, i))
 	}
-	childPath := bitset.Bytes(c.path)
-	for i := uint(0); i < c.count; i++ {
-		totalPath.SetBool(int(i+p.count), childPath.Get(int(i)))
+	//childPath := bitset.Bytes(c.path)
```
tarakby, line 440: I guess we do not need the commented line anymore?
```suggestion
```

storage/merkle/tree.go
```diff
@@ -3,39 +3,103 @@
...
+// Returns:
+//  * (false, nil): key-value pair is stored; key did _not_ yet exist prior to update
+//  * (true, nil):  key-value pair is stored; key existed prior to update and the old
+//                  value was overwritten
+//  * (false, error): with possible error returns
+//    - ErrorIncompatibleKeyLength if `key` is empty, or exceeds the max supported length of 8192 bytes.
+//    - ErrorVariableKeyLengths if the provided key length is _different_ than previously stored elements.
+//    No other errors are returned.
+func (t *Tree) Put(key []byte, val []byte) (bool, error) {
+	if t.size == 0 { // empty key-value store
```
tarakby, line 69: I like how the interface allows the same merkle tree to be re-used for keys of different lengths: once the tree is empty of all elements, the key length is set to the length of the first put element. This is a flexible interface though it comes with the cost of keeping track of `size`. 
I'm wondering if it's better to set the key length at the constructor level and not change it later. A new tree needs to be instantiated for a new key length. The advantage is to remove `size` and avoid updating it. 
I let you decide what is the better interface (maybe you have a use case in mind for the method `Size()` )

storage/merkle/node_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+var payload1, _ = hex.DecodeString("62b0326507ebce9d4a242908d20559ceca965c5e9848646bd0c05047c8487aadfcb3d851e77e5a055d306e48c376f8")
+var payload2, _ = hex.DecodeString("bab02e6213dfad3546aa473922bba0")
+
+// TestBlakeMac verifies that constructor blake2b.New256(m) never errors for any of the used MAC
+// values `m`. This is assumed by the Node's Hash() implementations. We test this assumption holds
+// at build time, but avoid runtime-checks for performance reasons.
+// Context: The blake2b MAC key should be less than 64 bytes. Constructor blake2b.New256(m) checks
+// this condition and errors, but we ignore its error return at runtime.
+func TestBlakeMac(t *testing.T) {
```
tarakby, line 21: 👌🏼

tarakby: Very minor: just to follow Go guidelines for acronyms in function names. 
```suggestion
func TestBlakeMAC(t *testing.T) {
```

storage/merkle/node.go
```diff
@@ -3,23 +3,110 @@
...
 }
 
+/* ******************************** Leaf Node ******************************* */
+// Leaf represents a key-value pair. We only store the value, because the
+// key is implicitly stored as the merkle path through the tree.
+
 type leaf struct {
-	key hash.Hash   // redundant copy of hash (key/path) for root hash
-	val interface{} // the concrete data we actually stored
+	val []byte
```
tarakby, line 87: Good point removing the path!

storage/merkle/node.go
tarakby: Blake2 was already used by this package, there is nothing new. The import update is a result of a file refactor, so all good

tarakby (approved):
Clean optimizations and security improvements, all with complete tests! Nice work 👏🏼

ledger/common/bitutils/utils.go
AlexHentschel: 👍 refactored code to have dedicated `SetBit` and `ClearBit` methods as well as general `WriteBit` method
https://github.com/onflow/flow-go/pull/1800/commits/3426d41a814a6f658e23363e6281e0702fa02a9a

storage/merkle/node.go
AlexHentschel: Sorry, not quite sure what you mean. Full nodes don't have a path segment at all (neither a path segment of variable length nor a path segment with fixed length). 

Each of the `fullNodeTag`, `leafNodeTag`, `shortNodeTag` are _unique and constant_ 1-byte identifiers to include the node type in the cryptographic hash: https://github.com/onflow/flow-go/blob/aa285f72ff6cbb8c601178e909eb9f86bccfe9a9/storage/merkle/node.go#L25-L29
Their purpose is to make collision attacks impossible, where an attacker pretends that a node is of a different type than it actually is. 

I probably misunderstood your suggestion @ramtinms. Some clarification / further explanation would be very helpful. Thanks.

ledger/common/bitutils/utils_test.go
AlexHentschel: Good observation! This is unnecessarily complex: 
https://github.com/onflow/flow-go/blob/aa285f72ff6cbb8c601178e909eb9f86bccfe9a9/ledger/common/bitutils/utils_test.go#L71-L75
Replaced it by the following code, which is much more compact 
https://github.com/onflow/flow-go/blob/3426d41a814a6f658e23363e6281e0702fa02a9a/ledger/common/bitutils/utils_test.go#L70

ledger/common/bitutils/utils.go
AlexHentschel: Following [Ramtin's and Tarak's suggestion](https://github.com/onflow/flow-go/pull/1800#discussion_r778338959), I reverted `SetBit` back to its previous definition https://github.com/onflow/flow-go/blob/3426d41a814a6f658e23363e6281e0702fa02a9a/ledger/common/bitutils/utils.go#L24-L26
and added the corresponding method for setting a bit to `0`
https://github.com/onflow/flow-go/blob/3426d41a814a6f658e23363e6281e0702fa02a9a/ledger/common/bitutils/utils.go#L33-L35


The generalized method, where the user specifies the bit value is now called `WriteBit` to avoid confusion:  https://github.com/onflow/flow-go/blob/3426d41a814a6f658e23363e6281e0702fa02a9a/ledger/common/bitutils/utils.go#L11-L16

ledger/common/bitutils/utils_test.go
```diff
@@ -11,6 +11,18 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestBitVectorAllocation(t *testing.T) {
```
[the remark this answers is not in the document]
AlexHentschel: The mathematical identity is probably trivially apparent for you, but it wasn't for me at fist sight 😬. As we are not the only engineers that are ever going to touch the code, I personally think there is value in keeping the test. Following the mind-set rather one test too much than one test too little 😅

storage/merkle/tree.go
```diff
@@ -3,39 +3,103 @@
...
+		l := len(key)
+		if l < 1 || maxKeyLength < l {
```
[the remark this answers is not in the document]
AlexHentschel: thanks. Included in https://github.com/onflow/flow-go/pull/1800/commits/15cac741858e8a9869629945c8397305f3e33fb3

storage/merkle/tree.go
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1800/commits/050a98873836fb6f63e566a9668927232741a7ac

storage/merkle/tree.go
AlexHentschel: I like your suggestion of pre-defining the key length at construction time, as it reduces the complexity of the trie implementation.

storage/merkle/node.go
ramtinms: all good, everything looks fine now

storage/merkle/tree.go
tarakby: Thanks for the refactor! Yes ignoring the error and adding the test is fine. This is similar to what you did for the Blake API 👌🏼

storage/merkle/tree.go
```diff
@@ -169,16 +214,20 @@ PutLoop:
...
 // if the key was found and false otherwise.
-func (t *Tree) Get(key []byte) (interface{}, bool) {
-
-	// we start at the root again
-	cur := &t.root
-
-	// we use the given key as path again
-	path := bitset.Bytes(key[:])
+func (t *Tree) Get(key []byte) ([]byte, bool) {
+	if t.root == nil || t.keyLength != len(key) {
```
tarakby, line 218: I think `unsafeGet` can handle the empty tree case if I'm not wrong.
```suggestion
	if t.keyLength != len(key) {
```

storage/merkle/tree.go
```diff
@@ -218,78 +264,79 @@ GetLoop:
...
 
-// Del will remove the value associated with the given key from the patricia
-// merkle trie. It will return true if they key was found and false otherwise.
+// Del removes the value associated with the given key from the patricia
+// merkle trie. It returns true if they key was found and false otherwise.
 // Internally, any parent nodes between the leaf up to the closest shared path
 // will be deleted or merged, which keeps the trie deterministic regardless of
 // insertion and deletion orders.
 func (t *Tree) Del(key []byte) bool {
+	if t.root == nil || t.keyLength != len(key) {
```
tarakby, line 282: and I guess it's the same here for `unsafeDel`, but it's better to double check :)
```suggestion
	if t.keyLength != len(key) {
```

tarakby (approved):
Nice refactor! Thank you for going through all the comments 🙇🏼

storage/merkle/tree.go
AlexHentschel: great observation 👍

storage/merkle/tree.go
AlexHentschel: 👍

AlexHentschel:
bors merge

PR #1803, opened 2022-01-04
[Consensus Voting] Test QC pack and unpack

zhangchiqing:
Adding a test to check the integration between the protocol state and combined vote processor. 

The combined vote processor packs the block sig data type into QC's sigData field, and protocol state's `Seed` method unpacks QC's sigData. This test case verifies the pair of Pack and Unpack methods used in the two difference places are integrated well.

Also added some comments for future clean up and moving to V3.

### DIFF
--- a/consensus/hotstuff/packer/packer.go
+++ b/consensus/hotstuff/packer/packer.go
@@ -42,11 +42,9 @@ func (p *SigDataPacker) Decode(data []byte) (*SignatureData, error) {
 	return &sigData, err
 }
 
-// UnpackRandomBeaconSigForV2 takes sigData previously packed by packer,
+// UnpackRandomBeaconSig takes sigData previously packed by packer,
 // decodes it and extracts random beacon signature
-// TODO: to be removed in V3. Remove it when the combined vote processor
-// uses ConsensusSigDataPacker to pack the the signature
-func UnpackRandomBeaconSigForV2(sigData []byte) (crypto.Signature, error) {
+func UnpackRandomBeaconSig(sigData []byte) (crypto.Signature, error) {
 	// decode into typed data
 	packer := SigDataPacker{}
 	sig, err := packer.Decode(sigData)
--- a/consensus/hotstuff/packer/packer.go
+++ b/consensus/hotstuff/packer/packer.go
@@ -42,8 +42,11 @@ func (p *SigDataPacker) Decode(data []byte) (*SignatureData, error) {
 	return &sigData, err
 }
 
-// UnpackRandomBeaconSig takes sigData previously packed by packer, decodes it and extracts random beacon signature
-func UnpackRandomBeaconSig(sigData []byte) (crypto.Signature, error) {
+// UnpackRandomBeaconSigForV2 takes sigData previously packed by packer,
+// decodes it and extracts random beacon signature
+// TODO: to be removed in V3. Remove it when the combined vote processor
+// uses ConsensusSigDataPacker to pack the the signature
+func UnpackRandomBeaconSigForV2(sigData []byte) (crypto.Signature, error) {
 	// decode into typed data
 	packer := SigDataPacker{}
 	sig, err := packer.Decode(sigData)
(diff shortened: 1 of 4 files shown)


consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -259,20 +259,43 @@ func (p *CombinedVoteProcessorV2) buildQC() (*flow.QuorumCertificate, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return qc, nil
+}
+
+// buildBlockSignatureDataForV2 build a block sig data for V2
+// It reuses the hotstuff.BlockSignatureData type to create the sig data without filling the RandomBeaconSigners field and
+// the AggregatedRandomBeaconSig field, so that the packer can be reused by both V2 and V3 to pack QC's sig data.
```
zhangchiqing, line 274: Adding this to explain why V2 and V3 can both use the `ConsensusSigDataPacker` to pack the sig data

consensus/hotstuff/packer/packer.go
```diff
@@ -42,8 +42,11 @@ func (p *SigDataPacker) Decode(data []byte) (*SignatureData, error) {
 	return &sigData, err
 }
 
-// UnpackRandomBeaconSig takes sigData previously packed by packer, decodes it and extracts random beacon signature
-func UnpackRandomBeaconSig(sigData []byte) (crypto.Signature, error) {
+// UnpackRandomBeaconSigForV2 takes sigData previously packed by packer,
+// decodes it and extracts random beacon signature
+// TODO: to be removed in V3. Remove it when the combined vote processor
```
durkmurder, line 47: I don't understand this, in V3 we will use exactly the same code. Difference between V2 and V3 is in `AggregatedRandomBeaconSig`. But this unpacking is for `ReconstructedRandomBeaconSig`. We can't use packer here since it needs a committee, we only need to extract the random beacon sig so no point in doing full decoding of QC sig.

state/protocol/seed/seed.go
```diff
@@ -11,9 +11,11 @@ import (
 // FromParentSignature reads the raw random seed from a QC sigData.
 // The sigData is an RLP encoded structure that is part of QuorumCertificate.
 // The indices can be used to  generate task-specific seeds from the same signature.
+// TODO: this only works for consensus V2 where a V2 packer is used to unpack the sig data
```
durkmurder, line 14: same thing, it will work for V3 exactly the same, unless I am missing something very obvious.

consensus/hotstuff/packer/packer.go
zhangchiqing: Yes, you are right. We only need to read the random beacon sig. Good catch

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -259,20 +259,43 @@ func (p *CombinedVoteProcessorV2) buildQC() (*flow.QuorumCertificate, error) {
 		return nil, fmt.Errorf("could not reconstruct random beacon group signature: %w", err)
 	}
 
-	blockSigData := &hotstuff.BlockSignatureData{
+	blockSigData := buildBlockSignatureDataForV2(stakingSigners, aggregatedStakingSig, reconstructedBeaconSig)
+
```
durkmurder, line 263:
```suggestion
```

consensus/hotstuff/votecollector/combined_vote_processor_v2_test.go
```diff
@@ -884,3 +886,40 @@ func VoteWithDoubleSig() func(*model.Vote) {
 		vote.SigData = unittest.RandomBytes(96)
 	}
 }
+
+// Test that a constructed QC can be unpacked and from which the random source can be
```
durkmurder, line 890:
```suggestion
// TestReadRandomSourceFromPackedQCV2 tests that a constructed QC can be unpacked and from which the random source can be
```

durkmurder (approved):
LGTM. Thank you.

PR #1808, opened 2022-01-06
Flaky Test Monitor - CI fix - updated level 1 script to use level 1 sub-directory

gomisha:
Fixing [Flakiness Summary](https://github.com/onflow/flow-go/actions/workflows/flakiness-summary.yml) workflow to use level 1 sub-directory to run level 1 flaky test monitor script.

In the process also removed Flaky Test Monitor module because there were issues running Flaky Test Monitor in CI (was getting error `package flaky-test-monitor/common is not in GOROOT`) and it was easier to just remove the module and import the common package as `github.com/onflow/flow-go/tools/flaky_test_monitor/common` since it's under the main `flow-go` repo


.github/workflows/flakiness-summary.yml
```diff
@@ -1,18 +1,19 @@
-name: Flakiness Summary
+name: Flaky Test Summary
 
 on:
   push:
     branches:
       - master
+      - 'misha/flaky-test-monitor-ci-fix'
```
smnzhu, line 7: remove

tools/flaky_test_monitor/go.mod
```diff
@@ -1,5 +0,0 @@
-module flaky-test-monitor
```
smnzhu, line 1: Why did we remove these files?

.github/workflows/flakiness-summary.yml
gomisha: removed

https://github.com/onflow/flow-go/pull/1808/commits/cc6ea3f46f9a28856019da445b54a4ce48b49e98

smnzhu (approved):

tools/flaky_test_monitor/go.mod
Kay-Zee: Also wondering

gomisha: I had issues with running flaky tests summary with the module there - go had a hard time loading the module so I realized I can just remove it and import the helper functions the way other code in flow-go imports reusable packages - that simplified things. I updated the PR description to explain this

Kay-Zee (commented):
The removal of the go.mod and go.sum means that we no longer treat this as a separate module, are we sure we want that?

it would make it much harder to separate it out if we ever wanted to do that.

gomisha:
> 

If I put back the module, it works locally, but I keep getting this error in CI:

```
no required module provides package github.com/onflow/flow-go/tools/flaky_test_monitor/common; to add it:
	go get github.com/onflow/flow-go/tools/flaky_test_monitor/common
```

https://github.com/onflow/flow-go/runs/4741750476?check_suite_focus=true

I think we should leave it without the module so it can run properly in CI. In the future, we can possibly move it out of the main `flow-go` repo if we need to run it separately.

Kay-Zee (approved):
While I do think we should have it as a separate module, the way it's currently setup, e.g. having multiple `main`s seems to be making the submodule compatible a bit difficult

Therefore, for now, we will merge this PR, and get the flakyness detection going, before figuring out the submodule setup

> we can possibly move it out of the main flow-go repo if we need to run it separately.

I don't think we want necessarily want to move it out separately, we just want it to be properly modularized, like the integration module, but it's not really a blocker, so we can merge and clean up the submodule after.

PR #1813, opened 2022-01-06
Consensus Voting V2

zhangchiqing:


network/p2p/peerstore_provider.go
```diff
@@ -0,0 +1,43 @@
+package p2p
```
zhangchiqing: to be reverted here

https://github.com/onflow/flow-go/pull/1814/files#diff-b55ca931ab95dac962bcdc52809f490fc06abfc9bb20aa7c1eea8fffe5dd7ccf

cmd/util/cmd/execution-state-extract/cmd.go
```diff
@@ -108,6 +108,42 @@ func run(*cobra.Command, []string) {
 		if err != nil {
 			log.Fatal().Err(err).Msg("invalid state commitment length")
 		}
+	} else if !flagNoMigration {
```
zhangchiqing, line 111: to be reverted https://github.com/onflow/flow-go/pull/1814/files#r779809581

PR #1816, opened 2022-01-07
Simonhf/6072 fix localnet loader for transaction execution success

simonhf:
Low touch / quick and dirty `-feedback-enabled` implementation, plus trace log enhancements.
Note: If `-feedback-enabled` is not given then loader functionality should be unchanged!
Note: Issue [1] has been created as tech debt to later optimize how feedback has been implemented.

[1] https://github.com/dapperlabs/flow-go/issues/6089


integration/utils/contLoadGenerator.go
```diff
@@ -386,81 +401,152 @@ func (lg *ContLoadGenerator) sendAddKeyTx(workerID int) {
...
 
-	lg.sendTx(addKeysTx)
+	lg.sendTx(workerID, addKeysTx)
+}
+
+func (lg *ContLoadGenerator) pushAccountsHappening(workerID int, acc *flowAccount, tx_id string, timeNow time.Time) {
+	lg.log.Trace().Msgf("workerID=%d pushing happening address", workerID)
+	lg.happeningAccounts <- (func() (*flowAccount, string, time.Time) { return acc, tx_id, timeNow })
+}
+
```
jwinkler2083233, line 429: A comment here and for the above function would be helpful

jwinkler2083233 (approved):

integration/utils/contLoadGenerator.go
simonhf: It's just a Golang-ism for encapsulating pushing 3 items onto the channel atomically by wrapping them in a function... definitely not the nicest thing to look at... :-)

integration/utils/contLoadGenerator.go
```diff
@@ -386,81 +401,152 @@ func (lg *ContLoadGenerator) sendAddKeyTx(workerID int) {
...
+func (lg *ContLoadGenerator) probeAccountsHappening(workerID int) {
+	channelElements := len(lg.happeningAccounts)
+	lg.log.Trace().Msgf("workerID=%d probing happening address from channel with %d elements", workerID, channelElements)
+	for i := 0; i < channelElements; i++ {
+		acc, tx_id, timeAtSendTx := (<-lg.happeningAccounts)()
+		txIdFile := fmt.Sprintf("/dev/shm/flow-transaction-feedback/%s/%s/%s.txt", tx_id[0:2], tx_id[2:4], tx_id[4:]) // e.g. /dev/shm/flow-transaction-feedback/70/b9/4146a01bc7843dc9547837e19ba0da6899fdd83ae61e2cf2e3fb4555da6a.txt
+		lg.log.Trace().Msgf("workerID=%d probing happening address %x account %d tx_id %s AKA %s", workerID, acc.address.Bytes(), acc.i, tx_id, txIdFile)
+		if _, err:= os.Stat(txIdFile); errors.Is(err, fs.ErrNotExist) {
```
Kay-Zee, line 437: Could you replace the above three lines with a function, so that we could easily replace it with a flowClient call at a later date? Something like

```go
func isAccountReady(acc *flowAccount, tx_id string) bool {
	txIdFile := fmt.Sprintf("/dev/shm/flow-transaction-feedback/%s/%s/%s.txt", tx_id[0:2], tx_id[2:4], tx_id[4:]) // e.g. /dev/shm/flow-transaction-feedback/70/b9/4146a01bc7843dc9547837e19ba0da6899fdd83ae61e2cf2e3fb4555da6a.txt
	lg.log.Trace().Msgf("workerID=%d probing happening address %x account %d tx_id %s AKA %s", workerID, acc.address.Bytes(), acc.i, tx_id, txIdFile)
	_, err:= os.Stat(txIdFile)
	return  errors.Is(err, fs.ErrNotExist)
}
```

and then we could just call replace `if _, err:= os.Stat(txIdFile); errors.Is(err, fs.ErrNotExist) {` with `if isAccountReady(acc, tx_id) {

integration/utils/contLoadGenerator.go
```diff
@@ -386,81 +401,152 @@ func (lg *ContLoadGenerator) sendAddKeyTx(workerID int) {
...
+func (lg *ContLoadGenerator) pushAccountsHappening(workerID int, acc *flowAccount, tx_id string, timeNow time.Time) {
+	lg.log.Trace().Msgf("workerID=%d pushing happening address", workerID)
+	lg.happeningAccounts <- (func() (*flowAccount, string, time.Time) { return acc, tx_id, timeNow })
+}
+
+func (lg *ContLoadGenerator) probeAccountsHappening(workerID int) {
+	channelElements := len(lg.happeningAccounts)
+	lg.log.Trace().Msgf("workerID=%d probing happening address from channel with %d elements", workerID, channelElements)
+	for i := 0; i < channelElements; i++ {
+		acc, tx_id, timeAtSendTx := (<-lg.happeningAccounts)()
```
Kay-Zee, line 434: more idomatic to not use underscore for this var name. Will have to change in all locations it's used as well

```suggestion
		acc, txID, timeAtSendTx := (<-lg.happeningAccounts)()
```

integration/utils/contLoadGenerator.go
```diff
@@ -386,81 +401,152 @@ func (lg *ContLoadGenerator) sendAddKeyTx(workerID int) {
...
+		txIdFile := fmt.Sprintf("/dev/shm/flow-transaction-feedback/%s/%s/%s.txt", tx_id[0:2], tx_id[2:4], tx_id[4:]) // e.g. /dev/shm/flow-transaction-feedback/70/b9/4146a01bc7843dc9547837e19ba0da6899fdd83ae61e2cf2e3fb4555da6a.txt
+		lg.log.Trace().Msgf("workerID=%d probing happening address %x account %d tx_id %s AKA %s", workerID, acc.address.Bytes(), acc.i, tx_id, txIdFile)
+		if _, err:= os.Stat(txIdFile); errors.Is(err, fs.ErrNotExist) {
+			// come here if file does not exist; means account is still happening...
+			lg.pushAccountsHappening(workerID, acc, tx_id, timeAtSendTx)
+
+			elapsed := time.Since(timeAtSendTx).Seconds()
+			elapsedSuspicious := 30.0
+			if (elapsed >= elapsedSuspicious) && (elapsed < (1+elapsedSuspicious)) {
```
Kay-Zee, line 443: so it only goes here if its between 30.0 and 31.0? Is the upper bound just so it doesn't spam?

integration/utils/contLoadGenerator.go
```diff
@@ -386,81 +401,152 @@ func (lg *ContLoadGenerator) sendAddKeyTx(workerID int) {
...
+				// come here to capture transactions which mysteriously have had no feedback after 30 seconds
+				// note: this should only be output once as its age reaches and passes through a certain amount of seconds...
+				lg.log.Warn().Msgf("workerID=%d address %x account %d tx_id=%s has been happening for %f seconds and counting", workerID, acc.address.Bytes(), acc.i, tx_id, elapsed)
+			}
+		} else {
+			// come here if file exists; means account is good to be used again
+			lg.availableAccounts <- acc
+			e := os.Remove(txIdFile)
```
Kay-Zee, line 451: If you take my suggestion above, for `isAccountReady`, will will either have to remove this (e.g. just leave garbage there, since the txIDs should always be unique) or add a `cleanup` function that takes the tx id.

integration/utils/contLoadGenerator.go
```diff
@@ -45,14 +48,17 @@ type ContLoadGenerator struct {
...
+	happeningAccounts    chan func() (*flowAccount, string, time.Time) // queue with accounts happening after worker processing
 	txTracker            *TxTracker
 	txStatsTracker       *TxStatsTracker
 	workerStatsTracker   *WorkerStatsTracker
 	workers              []*Worker
 	blockRef             BlockRef
 	stopped              bool
 	loadType             LoadType
+	feedbackEnabled      bool
+	availableAccountsLo  int
```
Kay-Zee, line 61: the Lo is lower bound?

integration/utils/contLoadGenerator.go
```diff
@@ -45,14 +48,17 @@ type ContLoadGenerator struct {
 	fungibleTokenAddress *flowsdk.Address
 	favContractAddress   *flowsdk.Address
 	accounts             []*flowAccount
-	availableAccounts    chan *flowAccount // queue with accounts that are available for workers
+	availableAccounts    chan         *flowAccount                     // queue with accounts available for   workers
+	happeningAccounts    chan func() (*flowAccount, string, time.Time) // queue with accounts happening after worker processing
```
Kay-Zee, line 52: maybe `inUse` instead of `happening` could be more clear?

Kay-Zee (commented):
A few small style based comments that's not a huge deal, but did want to make a small refactor to add the `isAccountReady` func

integration/utils/contLoadGenerator.go
simonhf: Yes! :-)

integration/utils/contLoadGenerator.go
simonhf: Yes! :-)

integration/utils/contLoadGenerator.go
simonhf: I thought about something like `inUseAccounts` but it seemed a bit vague to me because it didn't seem to differentiate itself enough from the existing `availableAccounts` and I was reluctant to refactor the `availableAccounts` name :-) But I don't mind `happeningAccounts` changing... it seemed at the time like the best of a bad bunch of names :-) Maybe something like `wait2useAccounts` or `wait4blockAccounts`?

integration/utils/contLoadGenerator.go
simonhf: Sorry, good idea. I look at the logs a lot and they like to use `tx_id`... and that's how it crept in there :-)

integration/loader/main.go
```diff
@@ -38,6 +38,7 @@ func main() {
 	logLvl := flag.String("log-level", "info", "set log level")
 	metricport := flag.Uint("metricport", 8080, "port for /metrics endpoint")
 	profilerEnabled := flag.Bool("profiler-enabled", false, "whether to enable the auto-profiler")
+	feedbackEnabled := flag.Bool("feedback-enabled", false, "whether to enable feedback / transaction tracking before account reuse (to avoid sequence number mismatch errors during transaction execution)")
```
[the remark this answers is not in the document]
simonhf: I'm not married to `feedback-enabled` but please note that there maybe an existing half baked / not on by default transaction tracking mechanism in the loader already. And so the `transaction-tracking` name might cause some confusion?

integration/utils/contLoadGenerator.go
simonhf: > I think with the way it's laid out, it'll actually be very easy to incorporate using the flow client in the loader, so i might just go ahead and do that, instead of waiting for an intern

> so, let me know if you don't have the cycles to make those changes, i can always incorporate them into my own change when i have time

Thanks! Yep, if you're going to make the changes anyway then we might just wait for that :-) Thanks again!

integration/utils/contLoadGenerator.go
Kay-Zee: yeah, maybe something like waitingAccounts

Kay-Zee (approved):
Since it's a non-breaking change, and mostly for local development with additional changes incoming, we can merge and make the updates in the following PR

PR #1818, opened 2022-01-07
[Consensus Voting] Fix consensus integration tests

zhangchiqing:
This PR fixes the disabled integration tests for consensus and collection. See comments on why they were failing.

However, the collection test were still flakey: they are passing locally but failing on CI. The flakiness will be addressed in a separate PR.


integration/tests/collection/suite_test.go
```diff
@@ -66,11 +66,14 @@ func TestCollectorSuite(t *testing.T) {
 func (suite *CollectorSuite) SetupTest(name string, nNodes, nClusters uint) {
 	var (
 		conNode = testnet.NewNodeConfig(flow.RoleConsensus, testnet.WithLogLevel(zerolog.ErrorLevel), testnet.AsGhost())
-		exeNode = testnet.NewNodeConfig(flow.RoleExecution, testnet.WithLogLevel(zerolog.FatalLevel), testnet.AsGhost())
-		verNode = testnet.NewNodeConfig(flow.RoleVerification, testnet.WithLogLevel(zerolog.FatalLevel), testnet.AsGhost())
+		// DKG require at least 2 consensus nodes
+		conNode2 = testnet.NewNodeConfig(flow.RoleConsensus, testnet.WithLogLevel(zerolog.ErrorLevel), testnet.AsGhost())
```
zhangchiqing, line 70: The collection integration tests was failing because the bootstrap step needs the random source to determine the cluster assignment, but to get the random source, it needs to get the DKG size, however, previously there was only 1 consensus node, which won't be able to return a correct DKG size. 

Adding another consensus node resolves the problem.

integration/tests/consensus/sealing_test.go
```diff
@@ -235,11 +254,29 @@ SearchLoop:
...
+
+	valid, err := ss.exe2SK.PublicKey().Verify(receipt2.ExecutorSignature, id[:], exeUtils.NewExecutionReceiptHasher())
+	require.NoError(ss.T(), err)
+	require.True(ss.T(), valid)
+
 ReceiptLoop:
 	for time.Now().Before(deadline) {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 		err := ss.Execution().Send(ctx, engine.PushReceipts, &receipt, ss.conIDs...)
+		err = ss.Execution2().Send(ctx, engine.PushReceipts, &receipt2, ss.conIDs...)
```
zhangchiqing, line 279: the consensus integration was failing because the sealing requires two matching receipts, but previously the integration tests only create one EN. 

Adding another EN solves the problem

integration/Makefile
```diff
@@ -8,13 +8,10 @@ test:
...
 .PHONY: integration-test
-integration-test: access-tests common-tests execution-tests verification-tests collection-tests epochs-tests network-tests consensus-tests 
+integration-test: access-tests common-tests execution-tests verification-tests collection-tests epochs-tests network-tests consensus-tests
 
-# NOTE: Currently skipping collection and consensus tests due to them not passing properly on CI
-# NOTE: Currently skipping all integration tests while replacing modules for consensus V2
 .PHONY: ci-integration-test
-
-ci-integration-test: access-tests common-tests execution-tests epochs-tests verification-tests network-tests # collection-tests # consensus-tests
+ci-integration-test: access-tests common-tests execution-tests epochs-tests verification-tests network-tests # collection-tests consensus-tests
```
zhangchiqing, line 14: Keep the collection-tests and consensus tests disabled. 

They are passing on local but failing on CI. Will address the flakiness in separate PR

durkmurder (approved):
Very nice, seems those issues were here for a while and not really related to changes that we have made in feature branch.

integration/README.md
```diff
@@ -1,14 +1,75 @@
-## Flow Integration Tests
+## How does Flow Integration Tests work
 
-A small testing utility which uses local Docker to run a network of nodes.
+The integration tests creates a network in local with different node types running as docker instances. The docker instances are created and started in the test case by a testing utility called `testingdock`.
```
tarakby, line 3:
```suggestion
The integration tests create a network locally with different node types running as docker instances. The docker instances are created and started in the test case by a testing utility called `testingdock`.
```

integration/README.md
```diff
@@ -1,14 +1,75 @@
...
+Why not just launching a full execution node instance in the consensus integration tests instead of using Ghost node?
+
+Because launching a full execution node in the consensus integration tests will be resource heavy. It also slows down the test case and might introduce other noises. Using a ghost node as a mock simplifes the test cases, and allows us to focus more on the specfic consensus related test cases to cover.
+
+### Rebuild image when debugging
+During test cases debugging, you might want to update some code. However, if you run `make integration-test` after updating the code, the new change will not be included, because the integration tests still use the old code from the docker image, which was built before adding the changes.
+
+So you need to rebuild all the images by running `make docker-build-flow` again before re-running the integration tests.
+
+Rebuilding all images take quite some time, here is a shortcut:
```
tarakby, line 70:
```suggestion
Rebuilding all images takes quite some time, here is a shortcut:
```

integration/tests/collection/suite_test.go
tarakby: It's possible to simulate the random beacon with 1 node only (DKG + threshold signature). I remember this was done in some earlier integration tests. 
Let me know if that's still needed and we can fix it, otherwise working with at least 2 nodes is fine.

tarakby (approved):

integration/tests/collection/suite_test.go
zhangchiqing: It was [failing here](https://github.com/onflow/flow-go/blob/0ab49815b6fae8a890f1c287f35604ab5bec2bf6/crypto/bls_thresholdsign.go#L150-L154), where the [`ThresholdSignMinSize` is `2`](https://github.com/onflow/flow-go/blob/0ab49815b6fae8a890f1c287f35604ab5bec2bf6/crypto/types.go#L95), and dkg.Size() was returning `1` with only 1 node.

zhangchiqing:
bors merge

PR #1823, opened 2022-01-10
Improve efficiency of `Views` and `Heights` gadgets

jordanschalm:


state/protocol/events/gadgets/views.go
```diff
@@ -48,22 +71,28 @@ func (v *Views) BlockFinalized(block *flow.Header) {
...
-
-	for _, view := range finalizedViews {
-		for _, callback := range v.callbacks[uint64(view)] {
-			callback(block)
+	// we have no other queued view callbacks
+	if cutoff >= len(v.orderedViews) {
+		if len(v.orderedViews) > 0 {
+			v.orderedViews = []uint64{}
+			return
 		}
```
zhangchiqing, line 94:
```suggestion
		if len(v.orderedViews) > 0 {
			v.orderedViews = []uint64{}
		}
		return
```

Should we do this instead?
otherwise, when `len(v.orderedViews) == 0 `, executing L97 will crash due to out of indices

jordanschalm: I think they are equivalent. L97 would be safe (though unnecessary) because `cutoff` must be `0` if `len(v.orderedViews) == 0`. `someSlice[0:]` is safe when `someSlice` is empty or `nil`. 

Committing the suggestion though, because it is clearer 👍

zhangchiqing (approved):
Looks good

kc1116 (approved):

PR #1824, opened 2022-01-10
Conditional skipping of normal tests

gomisha:


gomisha:
After a lot of discussion and feedback from the team, we will proceed with conditional test skipping as follows:
- no changes to normal tests - no need to update them with a guard
- special tests (e.g. flaky, long running, WIP, etc) will have a suffix added to the function name (e.g. `func TestAbc_Flaky(t *testing.T) {)` and continue using the conditional skip logic they currently use
- running special tests will be done with regex based on the function name suffix (e.g. `go test -run _Flaky`)

PR #1826, opened 2022-01-11
Check and Use Contract Audit Vouchers

tehranifar:
Added additional check on contract deployment to the FVM: Contracts can be deployed if there is a corresponding voucher on the `FlowContractAudits` contract on the service account.

Related Flow core-contracts PR: https://github.com/onflow/flow-core-contracts/pull/271

### Discussion
Since the code path is only triggered on mainnet, is there a way to test this feature on canary or testnet before going live?

I tried moving the restricted environment boolean check out of `isAuthorized` but a lot of test cases depend on the flag. In particular the test suite for frozen accounts fails because it's not using the bootstrap procedure to init the environment.


fvm/blueprints/contracts.go
```diff
@@ -59,3 +59,49 @@ func DeployContractTransaction(address flow.Address, contract []byte, contractNa
 		SetScript([]byte(fmt.Sprintf(deployContractTransactionTemplate, contractName, hex.EncodeToString(contract)))).
 		AddAuthorizer(address)
 }
+
+const auditContractForDeploymentTransactionTemplate = `
```
janezpodhostnik, line 63: This is used only in tests, so I would prefer this being in a test module instead of in blueprints directly.

fvm/transactionEnv.go
```diff
@@ -179,6 +180,11 @@ func (e *TransactionEnv) GetAuthorizedAccountsForContractUpdates() []common.Addr
 	return addresses
 }
 
+func (e *TransactionEnv) UseContractAuditVoucher(address runtime.Address, code []byte) (bool, error) {
```
janezpodhostnik, line 183: Could be private.

janezpodhostnik (commented):
Looks good to me.

I left some small comments for some code cleanup.

fvm/transactionEnv.go
tehranifar: Good idea, addressed https://github.com/onflow/flow-go/pull/1826/commits/07ac4a829bfc31538a0825507fcecbf262f9ff91
Also updated the core-contracts branch now that it's merged

janezpodhostnik (approved):

fvm/bootstrap.go
```diff
@@ -181,6 +181,7 @@ func (b *BootstrapProcedure) Run(vm *VirtualMachine, ctx Context, sth *state.Sta
 
 	service := b.createServiceAccount(b.serviceAccountPublicKey)
 
+	b.deployContractAuditVouchers(service)
```
ramtinms, line 184: does the order of deployment matter? shouldn't it be after deploying fungible token and flow tokens contracts?

fvm/handler/contract.go
```diff
@@ -55,8 +59,15 @@ func (h *ContractHandler) GetContract(address runtime.Address, name string) (cod
...
 	// check if authorized
 	if !h.isAuthorized(signingAccounts) {
-		err = errors.NewOperationAuthorizationErrorf("SetContract", "setting contracts requires authorization from specific accounts")
-		return fmt.Errorf("setting contract failed: %w", err)
+		// check if there's an audit voucher for the contract
+		voucherUsed, err := h.useContractAuditVoucher(address, code)
+		if err != nil {
+			return fmt.Errorf("setting contract failed: (useContractAuditVoucher) %w", err)
+		}
+		if !voucherUsed {
```
ramtinms, line 67: should it this be if voucher is used we return an error ?

fvm/handler/contract.go
```diff
@@ -55,8 +59,15 @@ func (h *ContractHandler) GetContract(address runtime.Address, name string) (cod
 func (h *ContractHandler) SetContract(address runtime.Address, name string, code []byte, signingAccounts []runtime.Address) (err error) {
 	// check if authorized
 	if !h.isAuthorized(signingAccounts) {
-		err = errors.NewOperationAuthorizationErrorf("SetContract", "setting contracts requires authorization from specific accounts")
-		return fmt.Errorf("setting contract failed: %w", err)
+		// check if there's an audit voucher for the contract
+		voucherUsed, err := h.useContractAuditVoucher(address, code)
+		if err != nil {
+			return fmt.Errorf("setting contract failed: (useContractAuditVoucher) %w", err)
```
ramtinms, line 65: @janezpodhostnik did we check that errors that are passed to this method are not considered fatal?

ramtinms (approved):
Looks good to me, I just put some comments and questions

fvm/handler/contract.go
ramtinms: maybe renaming the return value to voucherAvailable

fvm/bootstrap.go
tehranifar: The order changes the hardcoded root hash. The `FlowContractAudits` contract doesn't import any other contracts and is used for deployments so I chose to deploy in this order.

fvm/handler/contract.go
tehranifar: When a voucher is used, it means that an auditor has approved the contract so we only reject the deployment if no vouchers are used (i.e. no appropriate vouchers exist)

tehranifar: Makes sense, addressed here https://github.com/onflow/flow-go/pull/1826/commits/7a550c6655690ba3864d97d61723e258412c3965

tehranifar:
bors merge

fvm/handler/contract.go
tehranifar: Fixed here https://github.com/onflow/flow-go/pull/1826/commits/47b3651096243fe7ddd6cf76fa9688e19734576e

PR #1828, opened 2022-01-11
Sealing Core: Allow unverifiable block references below root height

jordanschalm:
In general the sealing segment may contain execution results which reference blocks before the sealing segment. For nodes using the sealing segment as an initial state, these blocks would be unknown. It is safe to ignore these ERs, because they reference blocks prior to the latest sealed block, by definition of the sealing segment. 

This PR detects and ignores `UnverifiableInput` errors for these blocks.


engine/consensus/sealing/core.go
```diff
@@ -241,16 +247,25 @@ func (c *Core) RepopulateAssignmentCollectorTree(payloads storage.Payloads) erro
...
 			incRes.IncorporatedBlockID, err)
 	}
 	incorporatedAtHeight := incorporatedBlock.Height
+	belowRootHeight := incorporatedAtHeight <= c.rootHeader.Height
+
+	err = c.checkBlockOutdated(incRes.Result.BlockID)
+	// we can safely ignore results incorporated below the root height which
+	// are unverifiable (reference unknown blocks) - these are blocks which
+	// are incorporated prior to this node's initial snapshot
+	if belowRootHeight && engine.IsUnverifiableInputError(err) {
```
durkmurder, line 263: It seems to me it's safe to check only if it's belowRootHeight and just drop it, we don't even need to check if block is outdated since it's already below root snapshot height == outdated

engine/consensus/sealing/core.go
```diff
@@ -75,6 +75,7 @@ type Core struct {
 	sealingTracker             consensus.SealingTracker           // logic-aware component for tracking sealing progress.
 	tracer                     module.Tracer                      // used to trace execution
 	config                     Config
+	rootHeader                 *flow.Header // the root header of this node's root state snapshot
```
durkmurder, line 78: let's reoder with config to keep comments aligned

engine/consensus/sealing/core_test.go
```diff
@@ -722,3 +732,68 @@ func (s *ApprovalProcessingCoreTestSuite) TestRepopulateAssignmentCollectorTree(
...
 }
+
+// TestRepopulateAssignmentCollectorTree_RootSealingSegment tests that the sealing
+// engine will be initialized correctly when bootstrapping with a root sealing
+// segment with multiple blocks, as is the case when joining the network at an epoch
+// boundary.
+//
+// In particular, the assignment collector tree population step should ignore
+// unknown block references below the root height.
+//
```
durkmurder, line 743:
```suggestion
```

durkmurder (commented):
After thinking about it more I would strongly suggest incorporate the check for `belowRootHeight` inside of `checkBlockOutdated` and treat this case as outdated error. This will cover both cases in `processIncorporatedResult` and `processApproval` with minimal changes.

jordanschalm:
> After thinking about it more I would strongly suggest incorporate the check for belowRootHeight inside of checkBlockOutdated and treat this case as outdated error. This will cover both cases in processIncorporatedResult and processApproval with minimal changes.

@durkmurder I don't think this is possible for the other case (`processApproval`), since there we only have access to the [block ID](https://github.com/onflow/flow-go/blob/2e18250ef7bd38c4d15cd8457866fc27d3814b81/engine/consensus/sealing/core.go#L416). Without the block height, we can't distinguish between blocks which are unknown because they are below our root height, or blocks which are unknown for another reason.

engine/consensus/sealing/core.go
jordanschalm: I wasn't sure about this point. 

When we initialize a new node with a sealing segment, we have an initial state like:
```
A<-B<-C<-D
         ^- root height / finalized height
^-sealed height
```

The node only knows that `A` is sealed, so information about sealing `B`, `C`, and `D` is still (at least theoretically) relevant to us. Practically, with the joining process we've defined, the node would not actively participate in sealing these blocks because the node would have 0 weight at this time (they would observe the sealing by incorporating blocks proposed by others, but would not propose or vote on these blocks). 

I feel it is safer to be strict about this here, and make sure we initialize the sealing engine with ERs for unsealed blocks in this case as well.

engine/consensus/sealing/core_test.go
jordanschalm: This is intentional -- whitespace in longer comments for easier reading ;)

engine/consensus/sealing/core.go
zhangchiqing: Does sealing engine need to process those result at all if the block of the result is unknown?

To me it's an overkill to adjust the logic in `processIncorporatedResult`. `processIncorporatedResult` should just assume the block of the input result must be found. Adding the logic here makes this function more complex.

For such a sealing segment, we can create a list of unknown blocks referred by all results from the blocks in the sealing segment. And not to process them [when initializing the sealing core here](https://github.com/onflow/flow-go/blob/139cf03b1c9971af34f39e22a5e1c45a90724a9e/engine/consensus/sealing/core.go#L178-L179)
```
A<-B(Result_X)<-C<-D
```

Once the sealing core has been initialized, all future incorporated results must refer to a known block. 

So I think we should update the `RepopulateAssignmentCollectorTree` function, rather than the `processIncorporatedResult`

zhangchiqing (changes requested):
The solution could work, but I have different opinion and I think the alternative could work better.

engine/consensus/sealing/core.go
durkmurder: > I feel it is safer to be strict about this here, and make sure we initialize the sealing engine with ERs for unsealed blocks in this case as well. But let me know what you think.

That is actually right, I've thought about root height a little bit differently. Yeah, better to be strict.

@zhangchiqing 's suggestion is also worth thinking about. Indeed this happens only when initialization, so maybe we deal with that explicitly in `RepopulateAssignmentCollectorTree` ? Later we can't even receive such result since it must be incorporated into some block.

durkmurder:
> > After thinking about it more I would strongly suggest incorporate the check for belowRootHeight inside of checkBlockOutdated and treat this case as outdated error. This will cover both cases in processIncorporatedResult and processApproval with minimal changes.
> 
> @durkmurder I don't think this is possible for the other case (`processApproval`), since there we only have access to the [block ID](https://github.com/onflow/flow-go/blob/2e18250ef7bd38c4d15cd8457866fc27d3814b81/engine/consensus/sealing/core.go#L416). Without the block height, we can't distinguish between blocks which are unknown because they are below our root height, or blocks which are unknown for another reason.

Yeah, you are right.

engine/consensus/sealing/core.go
jordanschalm: 👍  Going to go ahead with Leo's suggestion

engine/consensus/sealing/core.go
```diff
@@ -169,6 +183,11 @@ func (c *Core) RepopulateAssignmentCollectorTree(payloads storage.Payloads) erro
 		}
 
 		for _, result := range payload.Results {
+			// skip results referencing blocks before the root sealing segment
+			_, isOutdated := outdatedBlockIDs[result.BlockID]
+			if isOutdated {
+				continue
```
zhangchiqing, line 189: If this result is referring to an old block, better to add a WARN log with both the result ID and the blockID

zhangchiqing (approved):
Looks good 👍

engine/consensus/sealing/core.go
jordanschalm: https://github.com/onflow/flow-go/pull/1828/commits/443ed762c03f5924a8d8afeb1575d82b17ddd799

durkmurder (approved):
Nice. Thank you. Looks good.

PR #1830, opened 2022-01-12
Broad consolidation and cleanup of error handling

AlexHentschel:
Addresses issue https://github.com/dapperlabs/flow-go/issues/5916 and https://github.com/dapperlabs/flow-go/issues/6054

## Goal

The main goal of this PR is to consolidate and clean up error handling for the Consensus Voting V2 code base. 
### package `hotstuff` (and all its sub-packages) now _doesn't_ use `engine` errors anymore
* All components in the `hotstuff` package (and all its sub-packages) now emit consensus-specific errors (from package `hotstuff/model`). Thereby, we reduce inter-dependency of the code and reduce the possebility of mixing error paths (hotstuff errors accidentally being mistaken for engine errors). 
* Originally, we though of HotStuff as somewhat stand-alone. Using its own error increases modularity of the code. 

Details:
* some of the existing sentinel errors (e.g. `model.ErrInvalidSigner`) were previously implemented using `errors.New("..")`. This males it impossible to wrap errors from lower level components. I have replaced some of these errors with custom types, that allow to wrap other errors (e.g. [`model.InvalidSignerError`](https://github.com/onflow/flow-go/blob/5ade9984e9b359e3d71ce4213b2d525cd0e6a465/consensus/hotstuff/model/errors.go#L237-L257))

### package [`module/signature`](https://github.com/onflow/flow-go/tree/9ce1b2c7dfff916416c6e919b4e597b56c7b108a/module/signature) now _doesn't_ use `engine` errors anymore
Previously we re-used engine errors here too. I added package-specific errors and wrapped lower-level errors from the crypto package, which decouples the internal implementation from the external-facing API. 

### unified and documented error returns for implementations of `hotstuff.Committee`
previously, different implementations were returning inconsistent sentinel errors

### Removed internal [`collectedIDs`](https://github.com/onflow/flow-go/blob/9ce1b2c7dfff916416c6e919b4e597b56c7b108a/consensus/hotstuff/signature/weighted_signature_aggregator.go#L26) from `WeightedSignatureAggregator` 
The `WeightedSignatureAggregator` internally uses `signature.SignatureAggregatorSameMessage`, which already rejects duplicates. There is no need to have the de-duplication logic twice. The revised implementation just checks for the respective sentinel error of the underlying code. 

### QC validity

Aggregation of public keys or signatures is not defined for empty key set or empty signature set. However, for QCs, our verifier code didn't check whether the respective lists are actually empty and just tried to aggregate. For byzantine QC, this would have led to a crash instead of recognizing the QC as invalid. 

Possible edge cases are different for [`CombinedVerifier`](https://github.com/onflow/flow-go/blob/5ade9984e9b359e3d71ce4213b2d525cd0e6a465/consensus/hotstuff/verification/combined_verifier_v2.go), [`CombinedVerifierV3`](https://github.com/onflow/flow-go/blob/5ade9984e9b359e3d71ce4213b2d525cd0e6a465/consensus/hotstuff/verification/combined_verifier_v3.go) and [`StakingVerifier`](https://github.com/onflow/flow-go/blob/5ade9984e9b359e3d71ce4213b2d525cd0e6a465/consensus/hotstuff/verification/staking_verifier.go). I fixed the edge cases such that QCs with empty signers are now recognized as invalid, added in-code documentation and unit tests. 

### separating error paths for [`SignatureAggregatorSameMessage`](https://github.com/onflow/flow-go/blob/5ade9984e9b359e3d71ce4213b2d525cd0e6a465/module/signature/aggregation.go) (@tarakby)
* I explicitly now check if no signatures are stored so far and return a dedicated sentinel error (as opposed to re-using the sentinel error from the underlying `crypto` package). This allows to decouple error paths (no signatures added vs some of the signatures being invalid)
* added a unit test to verify that the implementation properly distinguishes both error cases and returns different sentinel errors
* As an optimization, `SignatureAggregatorSameMessage` now also caches the `indices` in addition to the aggregated signature for later recall


consensus/hotstuff/signature/packer.go
```diff
@@ -36,8 +36,8 @@ func (p *ConsensusSigDataPacker) Pack(blockID flow.Identifier, sig *hotstuff.Blo
 	signerIDs := make([]flow.Identifier, 0, count)
 	sigTypes := make([]hotstuff.SigType, 0, count)
 
-	// read all the possible signer IDs at the given block
-	consensus, err := p.committees.Identities(blockID, filter.HasRole(flow.RoleConsensus))
+	// retrieve all authorized consensus participants at the given block
+	consensus, err := p.committees.Identities(blockID, filter.Any)
```
durkmurder, line 40: Why we don't filter anymore?

durkmurder: because it already filters inside of implementation?

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -59,74 +60,54 @@ func NewWeightedSignatureAggregator(
...
-//
-// The function errors:
-//  - engine.InvalidInputError if signerID is invalid (not a consensus participant)
-//  - module/signature.ErrInvalidFormat if signerID is valid but signature is cryptographically invalid
-//  - random error if the execution failed
+// Verify verifies the signature under the stored public keys and message.
+// Error returns:
+//  - model.InvalidSignerError if signerID is invalid (not a consensus participant)
+//  - model.ErrInvalidSignature if signerID is valid but signature is cryptographically invalid
+//  - generic error in case of unexpected runtime failures
```
durkmurder, line 80: shall we use exception instead of generic error? I believe we use exception in most of the cases

consensus/hotstuff/validator.go
```diff
@@ -9,11 +9,18 @@ import (
 type Validator interface {
 
 	// ValidateQC checks the validity of a QC for a given block.
+	// During normal operations, the following error returns are expected:
+	//	* model.InvalidBlockError if the QC is invalid
```
durkmurder, line 13: too much space?

consensus/hotstuff/validator.go
```diff
@@ -9,11 +9,18 @@ import (
...
 	ValidateQC(qc *flow.QuorumCertificate, block *model.Block) error
 
 	// ValidateProposal checks the validity of a proposal.
+	// During normal operations, the following error returns are expected:
+	//	* model.InvalidBlockError if the block is invalid
```
durkmurder, line 18: same here

consensus/hotstuff/validator/validator_test.go
```diff
@@ -314,13 +355,27 @@ func (vs *VoteSuite) TestVoteSignatureError() {
 	assert.False(vs.T(), model.IsInvalidVoteError(err), "internal exception should not be interpreted as invalid vote")
 }
 
+func (vs *VoteSuite) TestVoteInvalidSignerID() {
```
durkmurder, line 358: can we add a minimal godoc?

consensus/hotstuff/verification/combined_signer_v2_test.go
```diff
@@ -3,6 +3,8 @@ package verification
 import (
 	"testing"
 
+	"github.com/onflow/flow-go/state/protocol"
```
durkmurder, line 6: I think we should move it to other flow-go imports

consensus/hotstuff/verification/combined_signer_v3_test.go
```diff
@@ -3,6 +3,12 @@ package verification
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
durkmurder, line 6: Shall we group it as usual?

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
```diff
@@ -141,7 +141,20 @@ func (p *CombinedVoteProcessorV3) Status() hotstuff.VoteCollectorStatus {
...
 // * model.InvalidVoteError - submitted vote with invalid signature
+// * model.DuplicatedSignerError - vote from a signer whose vote was previously already processed
 // All other errors should be treated as exceptions.
+//
+// CAUTION: implementation is NOT (yet) BFT
+// Explanation: for correctness, we require that no voter can be counted repeatedly. However,
+// CombinedVoteProcessorV3 relies on the `VoteCollector`'s `votesCache` filter out all votes but the first for
+// every signerID. However, we have the edge case, where we still feed the proposers vote twice into the
+// `VerifyingVoteProcessor` (once as part of a cached vote, once as an individual vote). This can be exploited
+// by a byzantine proposer to be erroneously counted twice, which would lead to a safety fault.
```
durkmurder, line 152: Maybe I am missing something but we aren't counting votes twice, when `TrustedAdd` will be called then it will return an error for duplicated entry if that `signerID` was already accounted.

consensus/hotstuff/committees/static.go
```diff
@@ -66,18 +68,22 @@ func (s staticDKG) GroupKey() crypto.PublicKey {
 	return s.dkgGroupKey
 }
 
+// Index returns the index for the given node. Error Returns:
+// protocol.IdentityNotFoundError if nodeID is not a valid DKG participant.
 func (s staticDKG) Index(nodeID flow.Identifier) (uint, error) {
 	participant, ok := s.dkgParticipants[nodeID]
 	if !ok {
-		return 0, fmt.Errorf("could not get participant")
+		return 0, protocol.IdentityNotFoundError{NodeID: nodeID}
```
durkmurder, line 76: how about we add a constructor to follow same design as for other errors?

durkmurder (commented):
Very solid work, big refactoring. Took some to review it. In general seems very good, made a few comments about stylistic stuff and a few points which need clarification. Otherwise, GREAT job.

module/signature/aggregation.go
```diff
@@ -242,12 +261,12 @@ type PublicKeyAggregator struct {
...
 }
 
-// creates a new public key aggregator from all possible public keys
+// NewPublicKeyAggregator creates an index-based key aggregator, for the given list of authorized signers
 func NewPublicKeyAggregator(publicKeys []crypto.PublicKey) (*PublicKeyAggregator, error) {
 	// check for BLS keys
 	for i, key := range publicKeys {
 		if key == nil || key.Algorithm() != crypto.BLSBLS12381 {
-			return nil, engine.NewInvalidInputErrorf("key at index %d is not a BLS key", i)
+			return nil, fmt.Errorf("key at index %d is not a BLS key", i)
```
zhangchiqing, line 269: Is this an exception because the caller is always ourselves who is providing the correct pub keys. 

Isn’t this also being used by the QC validator where the pub key might be a wrong one?

zhangchiqing (commented):
Looks good. Just to double check my last comment.

consensus/hotstuff/signature/weighted_signature_aggregator.go
AlexHentschel: I often write:
```golang
//  - model.InvalidSignerError if signerID is invalid (not a consensus participant)
//  - model.ErrInvalidSignature if signerID is valid but signature is cryptographically invalid
```
(removes one extra line; thereby avoiding the overhead of always explicitly mentioning generic errors)

consensus/hotstuff/validator/validator_test.go
AlexHentschel: added documentations for test functions https://github.com/onflow/flow-go/pull/1830/commits/3a5ee936f6d86c451c674dd66d6e33f5671cdd42

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
```diff
@@ -242,10 +267,18 @@ func (p *CombinedVoteProcessorV3) Process(vote *model.Vote) error {
...
 // Any error should be treated as exception.
 func (p *CombinedVoteProcessorV3) buildQC() (*flow.QuorumCertificate, error) {
+	// STEP 1: aggregate staking signatures (if there are any)
+	// Note: it is possible that all replicas signed with their random beacon keys. In this case, the Aggregator
+	// returns an `model.InsufficientSignaturesError`. If no replicas signed with their staking key,
+	// `BlockSignatureData.StakingSigners` and `BlockSignatureData.AggregatedStakingSig` should be set to nil.
 	stakingSigners, aggregatedStakingSig, err := p.stakingSigAggtor.Aggregate()
-	if err != nil {
-		return nil, fmt.Errorf("could not aggregate staking signature: %w", err)
+	if err != nil && !model.IsInsufficientSignaturesError(err) {
```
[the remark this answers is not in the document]
AlexHentschel: Context:
* This is consensus voting V3, i.e. consensus participants vote with _either_ their staking signature _or_ their random beacon signature. 
* Currently, all consensus nodes are also random beacon members. Ideally, all consensus nodes would successfully participate in the DKG and then only sign their vote with their random beacon keys. 

From the perspective of the `CombinedVoteProcessorV3`:
* all votes are processed by the [`rbSigAggtor` and  `rbRector`](https://github.com/onflow/flow-go/blob/c6fd13ec8bf7adf51f631ae12a2cfcef7780a840/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L202-L227)
* There is not a single signature added to the `stakingSigAggtor`. 
* While the  `stakingSigAggtor`'s total weight remains `0`, the `rbSigAggtor` can easily reach the required weight for consensus supermajority. (-> [code](https://github.com/onflow/flow-go/blob/c6fd13ec8bf7adf51f631ae12a2cfcef7780a840/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L234))
* As all the voters sign with their random beacon keys, the required threshold for reconstructing the group signature is also easily reached (->[code](https://github.com/onflow/flow-go/blob/c6fd13ec8bf7adf51f631ae12a2cfcef7780a840/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L237-L239))

At this point, we have collected enough votes to construct a valid QC. We proceed to call `buildQC`: https://github.com/onflow/flow-go/blob/c6fd13ec8bf7adf51f631ae12a2cfcef7780a840/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L269-L274
where `stakingSigAggtor.Aggregate()` returns with an `InsufficientSignaturesError`
* The `stakingSigAggtor` is essentially telling us, that it can't build an aggregated signature, because it doesn't have any signatures at all.
* At the point where we want to build the QC, we have to decide how we represent the empty set of staking signers.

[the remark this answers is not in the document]
AlexHentschel: I have updated the logic for better readability (commit https://github.com/onflow/flow-go/pull/1830/commits/40321dfb4df551354015f098e18490c5a094b811 ) as follows: https://github.com/onflow/flow-go/blob/40321dfb4df551354015f098e18490c5a094b811/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L269-L287

I have this is clearer and better documented.

module/signature/aggregation.go
AlexHentschel: In all cases, public keys (should) have undergone verification before they are committed into the protocol state:
* As part of staking, nodes (should) be submitting a 'proof of possession' (essentially a signature of some small piece of information, as evidence that they actually have the private key for the public one they are submitting).  As part of their verifying their 'proof of possession', we use their public key. If their public key was invalid, their staking transaction should already fail, in which they submit the public key + proof of possession. 
* For the random beacon private keys, those are generated during the DKG, where a majority of consensus nodes must consistently locally arrive at the same public key vector. I was assuming that the DKG protocol guarnatees that each public key share is valid (no-one might have a corresponding private key, if the respective participant failed the DKG ... nevertheless, it is a properly formatted public BLS key). 

At the point where we pull the public keys out of our storage, and find that these keys are wrong, the networks protocol state is already compromised. 


@tarakby is my assumption 👆 about DKG always producing valid keys correct? Would appreciate your thoughts on whether or not we could be encountering invalid BLS keys (not signatures) here during normal operation.

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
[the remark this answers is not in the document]
AlexHentschel: updated the property test to also cover this case (-> [revised code](https://github.com/onflow/flow-go/blob/adb2de0ec9114e2d44f8fb8b5ceb04ee34e37080/consensus/hotstuff/votecollector/combined_vote_processor_v3_test.go#L519-L557))

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
AlexHentschel: Lets consider the following scenario: 
* node `Eve` is byzantine, it is her view to propose the next block
* she constructs her block, but before broadcasting it, she creates two different votes:
   * for `vote_s`  she uses her private staking key to sign her block
       * Eve sends `vote_s` her vote _before_ releasing her block. 
   * for `vote_r`,  she signs with her private random beacon key
       * Eve waits a 100ms or so after sending our her first vote `vote_s`. So she can be pretty sure that the next primary after her, lets call him `Bob`, has already received her vote but not her block. 
       * Then, Eve bundles her `vote_r`  together with her block and broadcasts it as her proposal.  

Lets now look how Bob's vote processing logic behaves:
1. When Eve's `vote_s` arrives, the block proposal itself is still unknown. Hence, Bob is still just caching votes in the [`VoteCollector`](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/statemachine.go#L31). At this point, Eve's `vote_s` is successfully [stored in the votes Cache](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/statemachine.go#L89). 
2. Now, Eve's [block proposal arrives](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/statemachine.go#L159)
    * a [`CombinedVoteProcessorV3` instance is created](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/statemachine.go#L173) for the block proposal, during which we [right away ingest the vote included in the proposal](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/factory.go#L44-L49)
    * Note that the proposer vote never passes through the cache and is therefore not de-duplicated w.r.t. all the cached votes. This is the reason, we also have the edge case of a [`DuplicatedSignerError` here](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/statemachine.go#L122-L127).
    * Byzantine Eve exploits this edge case to have her vote double-counted: We have just seen that the cache does not catch her duplicated vote. The `stakingSigAggtor` only sees `vote_s`. Her vote `vote_r` is only processed by the `rbSigAggtor` and  `rbRector`. Individually, each of these components only ever see a single vote (either `vote_s` or `vote_r`, never both). **In summary, both of Eve's votes `vote_s` and `vote_r` are successfully ingested, [each is counted towards the `minRequiredStake` threshold](https://github.com/onflow/flow-go/blob/8e5cfac395a8a9de1b54c2b801eb084ee6cd5b70/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L234), i.e. Bob has double-counted Eve**  ☣️🔥☠️

durkmurder: Yeap, you are right, I've missed the fact that you can provide sigs of both types, we definitely need an extra cache on vote processor level to avoid such situation.

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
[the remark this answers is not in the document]
durkmurder: Totally makes sense, thanks for explanation.

durkmurder (approved):
Great job. Very solid. :guitar: :guitar: :guitar:

zhangchiqing (approved):
Nice cleanup 👏

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
```diff
@@ -242,10 +267,28 @@ func (p *CombinedVoteProcessorV3) Process(vote *model.Vote) error {
...
+	//   and only call `Aggregate`, if the `stakingSigAggtor` has collected signatures
+	//   with non-zero weight (i.e. at least one signature was collected).
+	var stakingSigners []flow.Identifier // nil (zero value) represents empty set of staking signers
+	var aggregatedStakingSig []byte      // nil (zero value) for empty set of staking signers
+	if p.stakingSigAggtor.TotalWeight() > 0 {
+		var err error
+		stakingSigners, aggregatedStakingSig, err = p.stakingSigAggtor.Aggregate()
+		if err != nil {
+			return nil, fmt.Errorf("unexpected error aggregating staking signatures: %w", err)
+		}
```
tarakby, line 286: That's a very good catch 👌🏼

crypto/bls_multisig.go
```diff
@@ -72,6 +73,9 @@ func BLSVerifyPOP(pk PublicKey, s Signature) (bool, error) {
 // The order of the signatures in the slice does not matter since the aggregation
 // is commutative. The slice should not be empty.
 // No subgroup membership check is performed on the input signatures.
+// Expected error returns during normal operations:
+//  - invalidInputsError if no signatures are provided (sigs is empty) or
+//    some signatures are invalid
```
tarakby, line 78: Just clarifying what invalid means.
```suggestion
// Expected error returns during normal operations:
//  - invalidInputsError if no signatures are provided (sigs is empty) or
//    at least one signature fails to deserialize.
```

crypto/bls_multisig.go
```diff
@@ -114,6 +118,9 @@ func AggregateBLSSignatures(sigs []Signature) (Signature, error) {
 // The order of the keys in the slice does not matter since the aggregation
 // is commutative. The slice should not be empty.
 // No check is performed on the input private keys.
+// Error returns:
+//  - invalidInputsError if no signatures are provided (sigs is empty) or
+//    some keys are not BLS12 381 keys
```
tarakby, line 123: fixing the signature typo and adding that the only check done is about the Go type (no cryptographic check done as stated above).
```suggestion
// Expected error returns:
//  - invalidInputsError if keys is empty or at least one key type is not BLS12-381.
```

crypto/bls_multisig.go
```diff
@@ -144,6 +151,9 @@ func AggregateBLSPrivateKeys(keys []PrivateKey) (PrivateKey, error) {
 // The order of the keys in the slice does not matter since the aggregation
 // is commutative. The slice should not be empty.
 // No check is performed on the input public keys.
+// Error returns:
+//  - invalidInputsError if no signatures are provided (sigs is empty) or
+//    some keys are not BLS12 381 keys
```
tarakby, line 156:
```suggestion
// Expected error returns:
//  - invalidInputsError if keys is empty or at least one key type is not BLS12-381.
```

crypto/bls_multisig.go
```diff
@@ -186,6 +196,8 @@ func NeutralBLSPublicKey() PublicKey {
 // The order of the keys to be removed in the slice does not matter since the removal
 // is commutative. The slice of keys to be removed can be empty.
 // No check is performed on the input public keys.
+// Error returns:
+//  - invalidInputsError if some provided keys are not BLS12 381 keys
```
tarakby, line 200:
```suggestion
// Expected error returns:
//  - invalidInputsError if at least one key type is not BLS12-381.
```

module/signature/errors.go
```diff
@@ -5,21 +5,92 @@ import (
...
 }
 
-// newErrDuplicatedSigner constructs a new ErrDuplicatedSigner
-func newErrDuplicatedSigner(msg string, args ...interface{}) error {
-	return &ErrDuplicatedSigner{message: fmt.Sprintf(msg, args...)}
+/* ************************* InvalidSignerIdxError ************************* */
+
+// InvalidSignerIdxError indicates that the signer index is invalid
+type InvalidSignerIdxError struct {
+	err error
```
tarakby, line 36: Same comment for the embedded error and deleting `Error()` , although this is very minor and there isn't a need to update

tarakby (approved):
Very clean updates! Thanks for consolidating the errors and catching a few edge cases (the empty staking signatures in v3 in particular).

I left a few comments and questions below.

consensus/hotstuff/signature/randombeacon_inspector.go
```diff
@@ -32,7 +34,7 @@ func NewRandomBeaconInspector(
 		message,
 		encoding.RandomBeaconTag)
 	if err != nil {
-		return nil, engine.NewInvalidInputErrorf("create a new Random Beacon inspector failed: %w", err)
+		return nil, fmt.Errorf("creating BSL Threshold Signature Inspector failed: %w", err)
```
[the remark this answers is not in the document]
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

consensus/hotstuff/verification/combined_signer_v3_test.go
```diff
@@ -139,3 +153,165 @@ func TestCombinedSignWithNoDKGKeyV3(t *testing.T) {
...
+
+	// generate some BLS key as a stub of the random beacon group key and use it to generate a reconstructed beacon sig
+	privGroupKey, beaconSig := generateSignature(t, msg, encoding.RandomBeaconTag)
+	dkg := &mocks.DKG{}
+	dkg.On("GroupKey").Return(privGroupKey.PublicKey(), nil)
+	dkg.On("Size").Return(uint(20))
+	committee := &mocks.Committee{}
+	committee.On("DKG", mock.Anything).Return(dkg, nil)
+
+	// generate 17 BSL keys as stubs for staking keys and use them to generate an aggregated staking sig
```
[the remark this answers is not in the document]
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

consensus/hotstuff/verification/combined_signer_v3_test.go
```diff
@@ -139,3 +153,165 @@ func TestCombinedSignWithNoDKGKeyV3(t *testing.T) {
...
+	privStakingKeys, aggStakingSig := generateAggregatedSignature(t, 17, msg, encoding.ConsensusVoteTag)
+	// generate 11 BSL keys as stubs for individual random beacon key shares and use them to generate an aggregated rand beacon sig
```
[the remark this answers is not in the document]
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

consensus/hotstuff/signature/randombeacon_inspector.go
```diff
@@ -106,11 +108,21 @@ func (r *randomBeaconInspector) EnoughShares() bool {
...
-	return r.inspector.ThresholdSignature()
+	sig, err := r.inspector.ThresholdSignature()
+	if err != nil {
+		if crypto.IsInvalidInputsError(err) {
+			return nil, model.NewInvalidSignatureIncludedError(err)
+		}
+		if crypto.IsNotEnoughSharesError(err) {
+			return nil, model.NewInsufficientSignaturesError(err)
+		}
+		return nil, fmt.Errorf("unexpected error random beacon sig reconstruction: %w", err)
```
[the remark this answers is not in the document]
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

module/signature/aggregation.go
```diff
@@ -158,69 +157,89 @@ func (s *SignatureAggregatorSameMessage) HasSignature(signer int) (bool, error)
...
 	aggregatedSignature, err := crypto.AggregateBLSSignatures(signatures)
 	if err != nil {
-		return nil, nil, fmt.Errorf("BLS signature aggregtion failed %w", err)
+		// invalidInputsError for:
+		//  * empty `signatures` slice, i.e. sharesNum == 0, which we exclude by earlier check
+		//  * if some signature(s), included via TrustedAdd, could not be decoded
+		if crypto.IsInvalidInputsError(err) {
+			return nil, nil, NewInvalidSignatureIncludedErrorf("signatures with invalid structure were included via TrustedAdd: %w", err)
+		}
+		return nil, nil, fmt.Errorf("BLS signature aggregtion failed: %w", err)
```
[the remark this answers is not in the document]
AlexHentschel: included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

module/signature/aggregation.go
```diff
@@ -158,69 +157,89 @@ func (s *SignatureAggregatorSameMessage) HasSignature(signer int) (bool, error)
...
 	}
-	ok, err := s.VerifyAggregate(indices, aggregatedSignature)
+	ok, err := s.VerifyAggregate(indices, aggregatedSignature) // no errors expected (unless some public BLS keys are invalid)
 	if err != nil {
-		return nil, nil, fmt.Errorf("BLS signature aggregtion failed %w", err)
+		return nil, nil, fmt.Errorf("unexpected error during signature aggregation: %w", err)
 	}
 	if !ok {
-		return nil, nil, errors.New("resulting BLS aggregated signatutre is invalid")
+		return nil, nil, NewInvalidSignatureIncludedErrorf("signatures that do not match their public key were included via TrustedAdd")
```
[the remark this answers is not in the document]
AlexHentschel: I indeed overlooked the cases you describe. Personally, I feel that it is worse to have a misleading error message than a less precise one. thanks for catching this. I appreciate you thoroughly analyzing my rational here. 🙇 

included you suggesting in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

module/signature/aggregation_test.go
```diff
@@ -173,47 +164,60 @@ func TestAggregatorSameMessage(t *testing.T) {
...
+			ok, err = aggregator.VerifyAndAdd(i, sigs[(i+1)%signersNum]) // different signature for same index
+			assert.False(t, ok)
+			assert.True(t, IsDuplicatedSignerIdxError(err))
 		}
 	})
 
+	// Generally, `Aggregate()` can fail in two places, when invalid signatures were added via `TrustedAdd`:
+	//  1. The signature itself has an invalid structure, i.e. it can't be decoded to a curve point. In this
+	//     case, already the aggregation step fails.
+	//  2. The signature can be decoded to a curve point, but the signature doesn't match the public key. In
```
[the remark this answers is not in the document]
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

module/signature/aggregation_test.go
```diff
@@ -173,47 +164,60 @@ func TestAggregatorSameMessage(t *testing.T) {
...
-			assert.IsType(t, duplicate, err)
+			assert.True(t, IsDuplicatedSignerIdxError(err))
+			ok, err = aggregator.VerifyAndAdd(i, sigs[(i+1)%signersNum]) // different signature for same index
+			assert.False(t, ok)
+			assert.True(t, IsDuplicatedSignerIdxError(err))
 		}
 	})
 
+	// Generally, `Aggregate()` can fail in two places, when invalid signatures were added via `TrustedAdd`:
+	//  1. The signature itself has an invalid structure, i.e. it can't be decoded to a curve point. In this
```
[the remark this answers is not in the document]
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

crypto/bls_multisig.go
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

crypto/bls_multisig.go
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

crypto/bls_multisig.go
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

crypto/bls_multisig.go
AlexHentschel: suggestion included in https://github.com/onflow/flow-go/pull/1830/commits/42a51d24cee91f35b355670231e72f7f9aec52c5

consensus/hotstuff/model/errors.go
```diff
@@ -7,6 +7,14 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 )
 
+var (
+	ErrUnverifiableBlock = errors.New("block proposal can't be verified, because its view is above the finalized view, but its QC is below the finalized view")
+	ErrInvalidFormat     = errors.New("invalid signature format")
+	ErrInvalidSignature  = errors.New("invalid signature")
+)
+
+/* ****************************** NoVoteError ****************************** */
```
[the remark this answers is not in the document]
AlexHentschel: Thanks for the comment. Removed these separators

consensus/hotstuff/model/errors.go
```diff
@@ -143,3 +159,99 @@ func NewDoubleVoteErrorf(firstVote, conflictingVote *Vote, msg string, args ...i
...
+func IsInvalidSignatureIncludedError(err error) bool {
+	var e InvalidSignatureIncludedError
+	return errors.As(err, &e)
+}
+
+/* ********************** InsufficientSignaturesError ********************** */
+
+// InsufficientSignaturesError indicates that not enough signatures have been stored to complete the operation.
+type InsufficientSignaturesError struct {
+	err error
```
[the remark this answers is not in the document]
AlexHentschel: I think there is one potentially significant difference between embedding and error vs wrapping it:
* For a wrapped error, I can define the `Unwrap` method
    ```golang
    // WrappedError wraps the error
    type WrappedError struct {
    	err error
    }

    func (e WrappedError) Unwrap() error {
    	return e.err
    }
    ```
* In contrast, for an embedded error, I am not sure how to implement `Unwrap`. Is this possible at all?  
 
Without defining `Unwrap` we essentially break `errors.Is` from the runtime:
* once a low-level error is embedded into a higher-level error, we can't user `errors.Is` to determine the embedded error 
* while with wrapping, we still can use `errors.Is`

I made some code to illustrate this: https://go.dev/play/p/3hJ5JKNTqkA

module/signature/errors.go
AlexHentschel: see my earlier comment: https://github.com/onflow/flow-go/pull/1830#discussion_r787279755

consensus/hotstuff/signature/randombeacon_inspector.go
```diff
@@ -16,8 +15,11 @@ type randomBeaconInspector struct {
 }
 
 // NewRandomBeaconInspector instantiates a new randomBeaconInspector.
-//
-// It errors with engine.InvalidInputError if any input is not valid.
+// The constructor errors if:
```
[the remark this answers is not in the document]
AlexHentschel: I got the impression that the constructor only errors with `crypto.invalidInputsError` (?) I tried to only expose package-specific errors, which this error is not. So I would like to:
1. Either wrap it into a package-specific error if it is an expected error during normal operations. 
2. Or view a potential `crypto.invalidInputsError`  as in unexpected error. In this case, we would not need to expose the underlying error type, because we usually don't document the types of unexpected errors (only errors expected during normal operations).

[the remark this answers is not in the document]
AlexHentschel: Maybe, I'll wrap it as a `model.ConfigurationError` 🤷

module/signature/aggregation.go
```diff
@@ -30,15 +28,19 @@ type SignatureAggregatorSameMessage struct {
 	n                int                // number of participants indexed from 0 to n-1
 	publicKeys       []crypto.PublicKey // keys indexed from 0 to n-1, signer i is assigned to public key i
 	indexToSignature map[int]string     // signatures indexed by the signer index
-	cachedSignature  crypto.Signature   // cached aggregated signature
+
+	// To remove overhead from repeated Aggregate() calls, we cache the aggregation result.
+	// Whenever a new signature is added, we reset `cachedSignature` to nil.
+	cachedSignature     crypto.Signature // cached raw aggregated signature
+	cachedSignerIndices []int            // cached indices of signers that contributed to `cachedSignature`
```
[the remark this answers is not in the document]
AlexHentschel: 😅

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -97,9 +104,17 @@ func (c *CombinedVerifier) VerifyVote(signer *flow.Identity, sigData []byte, blo
...
+// VerifyQC checks the cryptographic validity of the QC's `sigData` for the
+// given block. It is the responsibility of the calling code to ensure
+// that all `voters` are authorized, without duplicates. Return values:
+//  - nil if `sigData` is cryptographically valid
+//  - model.ErrInvalidFormat if `sigData` has an incompatible format
+//  - model.ErrInvalidSignature if a signature is invalid
+//  - error if running into any unexpected exception (i.e. fatal error)
 func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
-
+	if len(signers) == 0 {
```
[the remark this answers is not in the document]
AlexHentschel: Good question. Thanks. I'll document the answer in the code as well. 
Reasoning:
* The function call to [`AggregateBLSPublicKeys`](https://github.com/onflow/flow-go/blob/e61f2934f815e1877cd616965d36753df26b249b/crypto/bls_multisig.go#L159-L171), which we do in line 142, https://github.com/onflow/flow-go/blob/e61f2934f815e1877cd616965d36753df26b249b/consensus/hotstuff/verification/combined_verifier_v2.go#L142-L145 returns a `crypto.invalidInputsError` in two distinct cases:
   1. In case no keys are provided, i.e.  `len(signers) == 0`. 
      This scenario _is expected_ during normal operations, because a byzantine proposer might construct an (invalid) QC with an empty list of signers. 
   2. In case some of the provided public keys are invalid. 
       This scenario is _not expected_ during normal operations, because all keys are supposed to be checked _before_ they enter the protocol state bases on a proof of possession.

In case of (i), we have an invalid QC. In case of (ii), we have a bug (or uncovered edge case) in our implementation. At the moment, we don't have a proof of possession, i.e. there is the possibility of  having invalid keys in the protocol state. 

The check `len(signers) == 0` is for robustness of the code: 
* By checking `len(signers) == 0` upfront, we can rule out case (i) as a source of error. Hence, if we encounter an error in line 142, we know it is case (ii). 
* Thereby, we can clearly distinguish a faulty _external_ input from an _internal_ bug.

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -119,60 +139,84 @@ func (c *CombinedVerifierV3) VerifyQC(signers flow.IdentityList, sigData []byte,
...
-	// first fetch all beacon signers public keys
-	beaconPubKeys := make([]crypto.PublicKey, 0, len(blockSigData.RandomBeaconSigners))
+	// STEP 2: verify aggregated random beacon key shares
+	// Step 2a: fetch all beacon signers public keys.
+	// Note: A valid random beacon group sig is required for QC validity. To reconstruct
+	// the group sig, _strictly more_ than `threshold` sig shares are required.
+	threshold := msig.RandomBeaconThreshold(int(dkg.Size()))
+	numRbSigners := len(blockSigData.RandomBeaconSigners)
+	if numRbSigners <= threshold {
+		return fmt.Errorf("require at least %d random beacon sig shares but only got %d: %w", threshold+1, numRbSigners, model.ErrInvalidFormat)
```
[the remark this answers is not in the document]
AlexHentschel: This check is to mitigate grieving attacks, i.e. depriving honest participants of their earned rewards. Some more context:
* Participating in the DKG is extra work. We encourage consensus participants to sign with their random beacon key, so that they contribute to HotStuff as well as the random beacon. While we allow consensus participants to only sign  with their staking key, it is discouraged because then, the participant does not contribute to the random beacon. 
* Likely, we will "encourage" nodes via their rewards: if they sign with their random beacon keys, they get a bit more rewards vs. only providing a staking sig. This also compensates the nodes for their extra effort participating in the DKG as opposed to being lazy and skipping the DKG. 

Lets consider the case where we see a QC:
1. a supermajority of consensus participants have signed either with their staking or random beacon key
2. the QC has a valid random beacon group signature 
3. _but_ the random beacon signers are insufficient 

At this point, we know something fishy is going on:
* The node that constructed the QC must have collected enough random beacon key shares. Otherwise, it would not be able to reconstruct the group signature. 
* BUT, the node is not giving credit to some of these random beacon signers for their contribution. Why not?
   * In case the network gives extra reward to random beacon signers, this is clearly a grieving attack. 
   * But even if we don't pay extra rewards, an honest node is supposed to credit the random beacon signers in its QC. If the node reports fewer random beacon signers than the minimally needed number, it is purposefully miss-representing node contributions. Not for rewards, but probably for some other off-chain benefits (e.g. attracting more delegators by under-reporting contributions of other nodes).

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -119,60 +139,84 @@ func (c *CombinedVerifierV3) VerifyQC(signers flow.IdentityList, sigData []byte,
...
-	stakingPubKeys := make([]crypto.PublicKey, 0, len(blockSigData.StakingSigners))
+	// STEP 3: validating the aggregated staking signatures
+	// Note: it is possible that all replicas signed with their random beacon keys, i.e.
+	// `blockSigData.StakingSigners` could be empty. In this case, the
+	// `blockSigData.AggregatedStakingSig` should also be empty.
+	numStakingSigners := len(blockSigData.StakingSigners)
+	if numStakingSigners == 0 {
+		if len(blockSigData.AggregatedStakingSig) > 0 {
+			return fmt.Errorf("all replicas signed with random beacon keys, but QC has aggregated staking sig for block %v: %w", block.BlockID, model.ErrInvalidFormat)
+		}
```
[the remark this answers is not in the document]
AlexHentschel: 🙇

consensus/hotstuff/verification/staking_verifier.go
```diff
@@ -51,20 +54,24 @@ func (v *StakingVerifier) VerifyVote(signer *flow.Identity, sigData []byte, bloc
...
 // In the single verification case, `sigData` represents a single signature (`crypto.Signature`).
 func (v *StakingVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
-	// verify the aggregated staking signature
-	msg := MakeVoteMessage(block.View, block.BlockID)
-
-	pks := make([]crypto.PublicKey, 0, len(signers))
-	for _, identity := range signers {
-		pks = append(pks, identity.StakingPubKey)
+	if len(signers) == 0 {
+		return fmt.Errorf("empty list of signers: %w", model.ErrInvalidFormat)
```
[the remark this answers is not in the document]
AlexHentschel: Whether or not we enter this check is dependent on the global order of all the checks for the QC. Essentially, there are two different checks that both work with the list of `signers`:
1. `signers` should represent a supermajority of consensus participants 
2. the aggregated staking sig should match the aggregated public keys  

This component here only implements check 2. If we check first 1. and then 2., the list of `signers` here should never be empty. However, if we first check 2. and then 1, this component has to handle the case of a malicious QC with empty signer list. 
The order in which we check 1. and 2. is implemented elsewhere. I generally trie as much as possible to reduce pre-conditions on other part of the code base, because it makes the code less brittle:
* adding the check here for `len(signers) == 0` means that the code here works without assuming any order. You can first check 1. and 2. or vice versa. Invalid QCs are detected and rejected. 
* However, without the check for `len(signers) == 0`, this line errors for empty signer list: https://github.com/onflow/flow-go/blob/e61f2934f815e1877cd616965d36753df26b249b/consensus/hotstuff/verification/staking_verifier.go#L74
   * we could require that 1. is checked first, but it would make the code more brittle (as changing the order of checks in some other part of the code would break this code here)
   * alternatively, we could check for an `crypto.invalidInputsError` in line 75: https://github.com/onflow/flow-go/blob/e61f2934f815e1877cd616965d36753df26b249b/consensus/hotstuff/verification/staking_verifier.go#L74-L77 However, then we are unable to distinguish a malformed external input from an internal uncovered edge-case (see [my earlier comment](https://github.com/onflow/flow-go/pull/1830#discussion_r787337554))

consensus/hotstuff/verification/combined_verifier_v2.go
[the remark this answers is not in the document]
AlexHentschel: I added the following documentation to the code: https://github.com/onflow/flow-go/blob/44d665742ac284822af7233fc564a766a6837d33/consensus/hotstuff/verification/combined_verifier_v2.go#L142-L157

consensus/hotstuff/verification/combined_verifier_v3.go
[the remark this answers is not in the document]
AlexHentschel: added the following documentation to the code: https://github.com/onflow/flow-go/blob/44d665742ac284822af7233fc564a766a6837d33/consensus/hotstuff/verification/combined_verifier_v3.go#L165-L172

consensus/hotstuff/verification/staking_verifier.go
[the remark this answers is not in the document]
AlexHentschel: added the following documentation: https://github.com/onflow/flow-go/blob/44d665742ac284822af7233fc564a766a6837d33/consensus/hotstuff/verification/staking_verifier.go#L75-L89

consensus/hotstuff/signature/weighted_signature_aggregator.go
```diff
@@ -17,37 +17,38 @@ type signerInfo struct {
...
+// WeightedSignatureAggregator implements consensus/hotstuff.WeightedSignatureAggregator.
+// It is a wrapper around signature.SignatureAggregatorSameMessage, which implements a
+// mapping from node IDs (as used by HotStuff) to index-based addressing of authorized
+// signers (as used by SignatureAggregatorSameMessage).
 type WeightedSignatureAggregator struct {
-	aggregator   *signature.SignatureAggregatorSameMessage // low level crypto BLS aggregator, agnostic of weights and flow IDs
-	ids          flow.IdentityList                         // all possible ids (only gets updated by constructor)
-	idToInfo     map[flow.Identifier]signerInfo            // auxiliary map to lookup signer weight and index by ID (only gets updated by constructor)
-	totalWeight  uint64                                    // weight collected (gets updated)
-	collectedIDs map[flow.Identifier]struct{}              // map of collected IDs (gets updated)
```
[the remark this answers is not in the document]
AlexHentschel: Thanks for the explanation. I did not anticipate extending signature aggregation to multiplicities larger than 1. I understand your reasoning now. Added `collectedIDs` and associated checks again plus documentation of the reasoning:

consensus/hotstuff/model/errors.go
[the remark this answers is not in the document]
tarakby: Thanks for the code. You can refer to the embedded attribute with the type name, here is how to define `Unwrap`:
```
func (e EmbeddedError) Unwrap() error {
	return e.error
}
```

I added the function to your example code and it seems to have the desired behaviour https://go.dev/play/p/ZRLH8wySjnj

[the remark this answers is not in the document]
tarakby: There is no need to update though, the purpose is just to avoid the `Error()` function which is not a huge advantage.

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -141,6 +141,18 @@ func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, b
...
 	aggregatedKey, err := crypto.AggregateBLSPublicKeys(signers.PublicStakingKeys()) // caution: requires non-empty slice of keys!
 	if err != nil {
+		// `AggregateBLSPublicKeys` returns a `crypto.invalidInputsError` in two distinct cases:
+		//  (i) In case no keys are provided, i.e.  `len(signers) == 0`.
+		//      This scenario _is expected_ during normal operations, because a byzantine
+		//      proposer might construct an (invalid) QC with an empty list of signers.
+		// (ii) In case some provided public keys are invalid.
+		//      This scenario is _not expected_ during normal operations, because all keys are
+		//      supposed to be checked _before_ they enter the protocol state bases on a proof
+		//      of possession. BUT, we don't have proof of possessions implemented at the moment.
```
tarakby, line 151: This not quite correct. PoP is not related to the possible error here. 

I'll try to explain:
The API of `AggregateBLSPublicKeys` accepts `PublicKey` elements and each has a signature algorithm attached to it. The function errors only if the algorithm is not `BLS_BLS12-381` (that's BLS signature on a curve called BLS12-381). There is no any other cryptographic check done. The public APIs available only allow generating valid BLS `PublicKey` structs and there is no way to play with the inner algorithm information. The error on this function can only happen if a developer mistakenly used let's say an `ECDSA` key. 

Now I wanted to clarify how the staking keys are validated. The staking contract receives bytes and validate that they represent a BLS key for the BLS12-381 group (a specific subgroup of the entire curve). The validation steps are:
 1. the bytes need to deserialize to a point on the elliptic curve BLS12-381 
 2. the point obtained is on the correct subgroup of BLS12-381 
 3. The PoP is valid
 Today, the contract only implements 1 and 2. Once the 2 steps are valid, the node software generates a `PublicKey` struct to be used by the rest of the protocol software (including hotstuff). The key struct has the correct assigned algorithm. In particular, it passes the check of `AggregateBLSPublicKeys`.
 The non-implemented step 3 of the validation does not impact this process at all. 
 
 Why is step 3 needed? Once we verify aggregated signatures using aggregated keys, we do need an extra measure to protect against Rogue attacks. Many measures are available and PoP is one of them and the most convenient for our case.

tarakby: For the comments above, we should not mention PoP, I suggest:
```
//      This scenario is _not expected_ during normal operations, because all staking keys are guaranteed by the protocol to be BLS keys.
```

tarakby: Updated the comments here : https://github.com/onflow/flow-go/pull/1831/commits/865febd177a21af2e5af1dc551f2d02ac92f75f5

PR #1831, opened 2022-01-12
Minor Key aggregation optimization

tarakby:
Update the greedy key aggregation algorithm that adjusts the latest aggregated key based on the signers delta.

In this PR, the aggregation is adjusting the existing aggregated key only if that's cheaper than computing the key from scratch. 
In Hotstuff, adjusting the existing key is always faster as long as all signer's weights are equal (because the 2/3 Hotstuff threshold) . This PR is useful in the case:
 - Hotstuff when weights are not equal
 - other non-Hotstuff usage of the module when aggregating keys


durkmurder (approved):
LGTM

AlexHentschel (approved):
Nice optimization. 

As part of my Error Handling PR, I didn't properly go over `PublicKeyAggregator`  because you mentioned that you were planning some work on it. Now that you are taking a pass over the `PublicKeyAggregator`, it would be great if you could address the following aspects:
* document expected error returns for each of it's public methods. 
* In case some of the signer indices are out of bound, your code will panic here https://github.com/onflow/flow-go/blob/46eaa014aa4712d9d81026b2d82d84ec9dd2fad8/module/signature/aggregation.go#L326 https://github.com/onflow/flow-go/blob/46eaa014aa4712d9d81026b2d82d84ec9dd2fad8/module/signature/aggregation.go#L334
   which is undesired for production code. I would recommend using the sentinel https://github.com/onflow/flow-go/blob/37be391b7f0924d1328525b1207b935cb7abc23a/module/signature/errors.go#L35 
 
P.S. it looks like your `aggregation.go` is an older version than what we have in the feature branch. I have worked on this file, so there will be merge conflicts.

tarakby:
Thanks @AlexHentschel for the review! 
- I added missing check for indices in https://github.com/onflow/flow-go/pull/1831/commits/6c9cd0c21a95c1d3c916f17fa6023c4373bdcd47
- I added checks for empty signers and added error-related comments in https://github.com/onflow/flow-go/pull/1831/commits/9a16e85b5fc5f4adb48777c4ec752a89b5504325
- left over from your errors PR : I updated the PoP related comments in https://github.com/onflow/flow-go/pull/1831/commits/865febd177a21af2e5af1dc551f2d02ac92f75f5

PR #1833, opened 2022-01-13
[Protocol State] Adding proof to the merkle trie

ramtinms:
This PR adds inclusion proof and verify functionality to the Merkle trie used by the protocol state. 
Aside from intermediate hash values, the proof includes some extra info for detecting the structure and for computing short node hashes.


storage/merkle/tree_test.go
```diff
@@ -338,6 +338,14 @@ func TestRandomOrder(t *testing.T) {
 	rand.Shuffle(len(keys), func(i int, j int) {
 		keys[i], keys[j] = keys[j], keys[i]
 	})
+
```
fxamacker, line 341: Not in this spot, but what do you think of adding a separate test for proofs with non-existent keys, including proofs for empty tree?

fxamacker (commented):
Nice work! 👍 Just a few minor comments and a question I'm not sure about.

storage/merkle/tree.go
```diff
@@ -273,6 +273,94 @@ GetLoop:
...
+	// we start at the root again
+	cur := &t.root
+
+	// we use the given key as path again
+	path := bitutils.MakeBitVector(t.keyLength)
+	for i := 0; i < t.keyLength; i++ {
+		if bitutils.ReadBit(key, i) == 1 {
+			bitutils.SetBit(path, i)
+		}
+	}
```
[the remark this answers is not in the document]
ramtinms: nice find, removed it.

storage/merkle/tree_test.go
ramtinms: yup, I'm going to add several tests as soon as the proof structure is approved.

AlexHentschel (commented):
Conceptually, the proof generation and verification looks good. 

A few high-level suggestions
* Please keep in mind that the proof is potentially constructed by a malicious prover; hence, we have to worry about the prover trying to cause index-out-of-bound exceptions (or overflows) on the recipient's side. (see my comment [here](url))
* To me, the different conventions of `hashIndex` vs `pathIndex` took me quite a bit to wrap my head around. I think we could improve clarity of the code by using a uniform convention (see comment [here](url)).
* The naming of `shortCounts` is counter-intuitive to me, as it primarily encodes the node type (short or full). Providing auxiliary information regarding the short node's count is in my view just for memory-efficient encoding of the proof. Would suggest to rename this (comment [here](url))

**To simplify further discussion, I have implemented my suggestions in PR https://github.com/onflow/flow-go/pull/1855 (targeting this branch).**

storage/merkle/proof.go
```diff
@@ -0,0 +1,85 @@
+package merkle
+
+import (
+	"bytes"
+
+	"github.com/onflow/flow-go/ledger/common/bitutils"
+)
+
+// Proof caputres all data needed for inclusion proof of a single value inserted under key `Key` into the merkle trie
```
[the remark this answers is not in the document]
ramtinms: fixed

storage/merkle/proof.go
```diff
@@ -0,0 +1,85 @@
...
+	// iterate backward and verify the proof
+	currentHash := computeLeafHash(p.Value)
+	hashIndex := len(p.InterimHashes) - 1
+
+	// compute last path index
+	pathIndex := len(p.InterimHashes)
+	for _, sc := range p.ShortCounts {
+		pathIndex += sc
+	}
+
```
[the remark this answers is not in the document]
ramtinms: added check for the first case, for the second case I've changed the format of the proof, so I don't think that's possible anymore.

storage/merkle/tree.go
```diff
@@ -273,6 +273,86 @@ GetLoop:
 	}
 }
 
+// Prove constructs inclusion proof for a given key if the key exists in the trie.
+// It traverses the trie from top to down and collects data for proof as follows:
+//  - if full node, capture the sibling node hash value and append zero to short counts
+//  - if short node, appends the node.shortCount to the short count list
+//  - if leaf, would capture the leaf value
+func (t *Tree) Prove(key []byte) (*Proof, bool) {
+
```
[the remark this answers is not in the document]
ramtinms: added

storage/merkle/tree.go
```diff
@@ -273,6 +273,86 @@ GetLoop:
...
+		// if we have a leaf, we found the key, return value and true
+		case *leaf:
+			return &Proof{
+				Key:           key[:],
+				Value:         n.val,
+				ShortCounts:   shortCounts,
+				InterimHashes: hashValues,
+			}, true
+
+		// if we have a nil node, key doesn't exist, return nil and false
```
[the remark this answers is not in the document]
ramtinms: updated

storage/merkle/tree.go
```diff
@@ -273,6 +273,94 @@ GetLoop:
...
+			cur = &n.child
+			index += n.count
+
+			continue ProveLoop
+
+		// if we have a leaf, we found the key, return value and true
+		case *leaf:
+			return &Proof{
+				Key:           key[:],
+				Value:         n.val,
```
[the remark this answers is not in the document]
ramtinms: removed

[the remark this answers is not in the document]
ramtinms: maybe, I was not sure

storage/merkle/proof.go
```diff
@@ -0,0 +1,85 @@
...
+	Key []byte
+	// value stored in the trie for the given key
+	Value []byte
+	// size of this is equal to the steps needed for verification,
+	// constructed by traversing top to down, if set to zero means we had a full node and a value from a hash value from InterimHashes should be read,
+	// any other value than 0 means we hit a short node and we need the shortcounts for computing the hash
+	//
+	// TODO: as this would include many zeros and would be sparse if the keys are computed as output of hashes, we might optimize
+	// this by using an sparse representation in the future (index and counts for non-zeros only)
+	ShortCounts []int
```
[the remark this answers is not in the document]
ramtinms: done

storage/merkle/tree.go
```diff
@@ -273,6 +273,86 @@ GetLoop:
...
+			// forward pointer and index to child
+			cur = &n.child
+			index += n.count
+
+			continue ProveLoop
+
+		// if we have a leaf, we found the key, return value and true
+		case *leaf:
+			return &Proof{
+				Key:           key[:],
```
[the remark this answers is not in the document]
ramtinms: updated

storage/merkle/tree.go
```diff
@@ -273,6 +273,86 @@ GetLoop:
...
+
+	// we start at the root again
+	cur := &t.root
+
+	// and we start at a zero index in the path
+	index := 0
+
+	// init proof params
+	hashValues := make([][]byte, 0)
+	shortCounts := make([]int, 0)
```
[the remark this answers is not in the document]
ramtinms: I did a refactor and separated out the shortCounts from the node types

storage/merkle/proof.go
```diff
@@ -0,0 +1,85 @@
...
+func (p *Proof) Verify(expectedRootHash []byte) bool {
+	// iterate backward and verify the proof
+	currentHash := computeLeafHash(p.Value)
+	hashIndex := len(p.InterimHashes) - 1
+
+	// compute last path index
+	pathIndex := len(p.InterimHashes)
+	for _, sc := range p.ShortCounts {
+		pathIndex += sc
+	}
```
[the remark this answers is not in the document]
ramtinms: I tried to fix this and update it to be readable, let me know if its still not clear

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	SkipBits []uint16
+	// InterimHashes is a slice of hash values, every value is read when we reach a full node (hash value of the siblings)
+	InterimHashes [][]byte
+}
+
+// Verify verifies the proof by constructing the hash values bottom up and cross check
+// the constructed root hash with the given one.
+// if the proof is valid it returns true and false otherwise
+func (p *Proof) Verify(expectedRootHash []byte) (bool, error) {
+
```
AlexHentschel, line 30: I would suggest to add detailed documentation here what conditions are checked and where. It took me literally half an hour to convince myself that these checks are sufficient (except for an edge case, which I'll comment on below).

The way I would document the security conditions is as follows
```suggestion

	// A valid proof as to satisfy the following consistency conditions:
	// 1. A valid inclusion proof represents a full path through the merkle tree.
	//    We separate the path into a sequence of interim vertices and a tailing leaf.
	//    For interim vertex (with index i, counted from the root node) along the path,
	//    the proof as to contain the following information:
	//      (i) whether the vertex is a short node (InterimNodeTypes[i] == 1) or
	//          a full node (InterimNodeTypes[i] == 0)
	//     (ii) for each short node, we need the number of bits in the node's key segment
	//          (entry in ShortPathLengths)
	//    (iii) for a full node, we need the hash of the sibling that is _not_ on the path
	//          (entry in SiblingHashes)
	//    Hence, len(ShortPathLengths) + len(SiblingHashes) specifies how many _interim_
	//    vertices are on the merkle path. Consequently, we require the same number of _bits_
	//    in InterimNodeTypes. Therefore, we know that InterimNodeTypes should have a length
	//    of `(numberBits+7)>>3` _bytes_.
	// 2. The key length (measured in bytes) has to be in the interval [1, 8192].
	//    Furthermore, each interim vertex on the merkle path represents:
	//    * either a single bit in case of a full node:
	//      we expect InterimNodeTypes[i] == 0
	//    * a positive number of bits in case of a short node:
	//      we expect InterimNodeTypes[i] == 1 
	//      and the number of bits is encoded in the respective element of ShortPathLengths
	//    Hence, the total key length _in bits_ should be: len(SiblingHashes) + sum(ShortPathLengths)
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	// first check the key size to be smaller than max allowed
+	if len(p.Key) > maxKeyLength {
+		return false, NewErrorMalformedProoff("key length is larger than max key lenght allowed (%d > %d)", len(p.Key), maxKeyLength)
+	}
+
+	// number of steps
+	steps := len(p.SkipBits) + len(p.InterimHashes)
+	// number of steps should be smaller than max key length
```
AlexHentschel, line 38: it can also be _equal_ 
```suggestion
	// number of steps should be smaller or equal to the max key length
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	if steps > maxKeyLength {
+		return false, NewErrorMalformedProoff("length of SkipBits pluse length of InterimHashes is larger than max key lenght allowed (%d > %d)", steps, maxKeyLength)
+	}
+	// number of steps should be smaller than number of bits in the IsAShortNode
+	if steps > len(p.IsAShortNode)*8 {
```
AlexHentschel, line 43: you check here, that `IsAShortNode` is not too short. But you don't check that it is not excessively too long. I could literally send you a `IsAShortNode` that is 2GB in size, where only the leading bytes are used and the rest is filled with garbage. Your algorithm wouldn't detect anything wrong with this proof. 

* we know _precisely_ what length we expect: 
```suggestion
	if len(p.IsAShortNode) != (steps+7)>>3 {
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+		return false, NewErrorMalformedProoff("the length of IsAShortNode doesn't match the length of SkipBits and InterimHashes")
+	}
+
```
AlexHentschel, line 46: Furthermore, there is ambiguity in the proof, because `IsAShortNode` as a byte slice might be slightly bigger than the minimally required number of bits. Generally, I think there is value in keeping things deterministic., which we could enforce by requiring that the tailing auxiliary bits are zero:
```suggestion
	// For deterministic proof: require that tailing auxiliary bits (to make a complete full byte) are all zero
	for i := len(p.IsAShortNode) - 1; i >= steps; i-- {
		if bitutils.ReadBit(p.IsAShortNode, i) != 0 {
			return false, NewErrorMalformedProoff("tailing auxiliary bits in IsAShortNode should all be zero")
		}
	}
	
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+		if bitutils.ReadBit(p.IsAShortNode, isAShortNodeIndex) == 0 {
+
+			// read and pop the sibling hash value from InterimHashes
+			if interimHashIndex < 0 {
+				return false, NewErrorMalformedProoff("no more InterimHashes available to read")
+			}
+			sibling := p.InterimHashes[interimHashIndex]
+			interimHashIndex--
+
+			// based on the bit at pathIndex of the key compute the hash
```
AlexHentschel, line 84:
```suggestion
			// based on the bit at keyIndex, compute the hash
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+			if bitutils.ReadBit(p.Key, keyIndex) == 0 { // left branching
+				currentHash = computeFullHash(currentHash, sibling)
+			} else {
+				currentHash = computeFullHash(sibling, currentHash) // right branching
+			}
+
+			// decrement the keyindex by 1
+			keyIndex--
+
+			continue
```
AlexHentschel, line 94: You are describing _what_ is happening,  i.e. `keyIndex` is decremented, which is trivially apparent from the code. But you are not describing why (which might deserve a comment)

```suggestion
			// move to the parent vertex along the path
			keyIndex--
			continue
```

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+			if bitutils.ReadBit(p.Key, startIndexOfCommonPath+j) == 1 {
+				bitutils.SetBit(commonPath, j)
+			}
+		}
+
+		// compute the hash for the short node
+		currentHash = computeShortHash(skipBits, commonPath, currentHash)
+
+		// decrement the keyIndex by the number of bits that were skiped in the short node
+		keyIndex -= skipBits
```
AlexHentschel, line 119: Personally, I find it easier to understand, if the code walks consistently backwards over the keys: 

```suggestion
		// construct the common path
		commonPath := bitutils.MakeBitVector(skipBits)
		for c := skipBits - 1; c >= 0; c-- {
			if bitutils.ReadBit(p.Key, keyIndex) == 1 {
				bitutils.SetBit(commonPath, c)
			}
			keyIndex--
		}

		// compute the hash for the short node
		currentHash = computeShortHash(skipBits, commonPath, currentHash)
```

AlexHentschel (changes requested):
First set of comments. It took me quite some time to  understand the code to the level where I could convince myself that it is correct. I think we have to take another pass to make it easier to comprehend. Added lots of suggestions.

Would like to take another pass over the code after the revisions. Also need to finish the review.

storage/merkle/errors.go
```diff
@@ -0,0 +1,27 @@
...
+
+var ErrorIncompatibleKeyLength = errors.New("key has incompatible size")
+
+// MalformedProofError is returned when the proof format has an issue
+type MalformedProofError struct {
+	err error
+}
+
+// NewErrorMalformedProoff constructs a new MalformedProofError
+func NewErrorMalformedProoff(msg string, args ...interface{}) *MalformedProofError {
```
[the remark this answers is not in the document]
ramtinms: nice catch, updated

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/onflow/flow-go/ledger/common/bitutils"
+)
+
+// Proof captures all data needed for proving inclusion of a single value inserted under key `Key` into the merkle trie
+type Proof struct {
+	// Key used to insert and look up the value
+	Key []byte
+	// Value stored in the trie for the given key
+	Value []byte
+	// IsAShortNode holds bits of data to determine short nodes versus full nodes while tranversing the
```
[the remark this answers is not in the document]
ramtinms: fixed

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	// trie downward. if the bit is set to 1, it means that we have reached to a short node, and
+	// if is set to 0 means we have reached a full node.
+	IsAShortNode []byte
```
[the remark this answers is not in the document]
ramtinms: updated the name

storage/merkle/proof.go
```diff
@@ -0,0 +1,133 @@
...
+	// SkipBits is read when we reach a short node, and the value represents number of bits that were skipped
+	// by the short node (shortNode.count)
+	SkipBits []uint16
+	// InterimHashes is a slice of hash values, every value is read when we reach a full node (hash value of the siblings)
+	InterimHashes [][]byte
```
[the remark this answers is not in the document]
ramtinms: updated the name

storage/merkle/tree.go
```diff
@@ -273,6 +268,102 @@ GetLoop:
...
+				if bitutils.ReadBit(key, i+index) != bitutils.ReadBit(n.path, i) {
+					return nil, false
+				}
+			}
+
+			cur = &n.child
+			index += n.count
+			if n.count > 65535 {
+				panic("short node count not fitting a uint16")
+			}
```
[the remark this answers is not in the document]
ramtinms: I would say it might be better to take the other approach limit tree max depth to 8191, do we really need that special case for an extra depth?

storage/merkle/proof.go
ramtinms: I'll add it

storage/merkle/proof.go
ramtinms: ok would add this, though its a computation cost we pay for the uniqueness of the proof.

storage/merkle/proof.go
ramtinms: fixed

storage/merkle/proof.go
ramtinms: fixed

storage/merkle/proof.go
ramtinms: updated

storage/merkle/proof.go
ramtinms: added

storage/merkle/proof.go
ramtinms: added

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	// fail for non-existing key
+	key2, _ := randomKeyValuePair(32, 128)
+
+	proof, existed = tree1.Prove(key2)
+	require.False(t, existed)
+	require.Nil(t, proof)
+}
+
+// TestValididateFormat tests cases a proof can not be valid
+func TestValididateFormat(t *testing.T) {
```
AlexHentschel, line 43:
```suggestion
// TestValidateFormat tests cases a proof can not be valid
func TestValidateFormat(t *testing.T) {
```

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestProofWithASingleKey tests proof generation and verification
+// when trie includes only a single value
+func TestProofWithASingleKey(t *testing.T) {
```
AlexHentschel, line 14: We should be testing boundary conditions here, i.e. keys with the **smallest possible size** and keys with the **largest possible size**. In addition, I would also test with some key in the middle (like case you already have).

storage/merkle/errors.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrorIncompatibleKeyLength = errors.New("key has incompatible size")
+
+// MalformedProofError is returned when a proof format
+// has some issues (syntax checks).
+type MalformedProofError struct {
```
AlexHentschel, line 12: would suggest to also include a method to test the error type:
```golang
// IsMalformedProofError checks if err is a IsMalformedProofError
func IsMalformedProofError(err error) bool {
	var target *MalformedProofError
	return errors.As(err, &target)
}
```

storage/merkle/errors.go
```diff
@@ -0,0 +1,50 @@
...
+// Unwrap unwraps the error
+func (e MalformedProofError) Unwrap() error {
+	return e.err
+}
+
+// InvalidProofError is returned when proof
+// verification has failed (semantic check).
+// The most common case for this error is when the computed root hash
+// doesn't match the root hash provided to the Verify method.
+type InvalidProofError struct {
```
AlexHentschel, line 34: same here:
```golang
// IsInvalidProofError checks if err is a InvalidProofError
func IsInvalidProofError(err error) bool {
	var target *InvalidProofError
	return errors.As(err, &target)
}
```

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	replaced, err = tree1.Put(key2, value2)
+	assert.NoError(t, err)
+	require.False(t, replaced)
+	proof, existed := tree1.Prove(key)
+	require.True(t, existed)
+
+	// invalid key size
+	proof.Key = make([]byte, 0)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 70: the primary criterium here to test for is that the expected error _type_ is returned, because calling code should work with the type and not rely on parsing the error message:
```suggestion
	assert.True(t, IsMalformedProofError(err))
```

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, key is empty")
+
+	// invalid key size (too large)
+	proof.Key = make([]byte, maxKeyLength+1)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 76: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, key length is larger than max key length allowed (8193 > 8192)")
+
+	// issue with the key size not matching the rest of the proof
+	proof.Key = make([]byte, 64)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 82: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, key length doesn't match the length of ShortPathLengths and SiblingHashes")
+
+	// reset the key back to its original value
+	proof.Key = key
+
+	// empty InterimNodeTypes
+	InterimNodeTypesBackup := proof.InterimNodeTypes
+	proof.InterimNodeTypes = make([]byte, 0)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 92: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, InterimNodeTypes is empty")
+
+	// empty InterimNodeTypes
+	proof.InterimNodeTypes = make([]byte, maxKeyLength+1)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 98: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, InterimNodeTypes is larger than max key length allowed (8193 > 8192)")
+
+	// issue with the size of InterimNodeTypes
+	proof.InterimNodeTypes = append(InterimNodeTypesBackup, byte(0))
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 104: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, the length of InterimNodeTypes doesn't match the length of ShortPathLengths and SiblingHashes")
+
+	proof.InterimNodeTypes = InterimNodeTypesBackup
+
+	// issue with a short count
+	backupShortPathLengths := proof.ShortPathLengths
+	proof.ShortPathLengths[0] = uint16(10)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 113: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.Equal(t, err.Error(), "malformed proof, key length doesn't match the length of ShortPathLengths and SiblingHashes")
+
+	// drop a shortpathlength - index out of bound
+	proof.ShortPathLengths = proof.ShortPathLengths[:1]
+	proof.ShortPathLengths[0] = uint16(255)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
```
AlexHentschel, line 120: same here: verify that returned error has type `MalformedProofError`

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	proof.Key = key
+
+	// empty InterimNodeTypes
+	InterimNodeTypesBackup := proof.InterimNodeTypes
+	proof.InterimNodeTypes = make([]byte, 0)
+	err = proof.valididateFormat()
+	assert.Error(t, err)
+	require.Equal(t, err.Error(), "malformed proof, InterimNodeTypes is empty")
+
+	// empty InterimNodeTypes
```
AlexHentschel, line 95:
```suggestion
	// too many interim nodes
```

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+	assert.Error(t, err)
+	require.Equal(t, err.Error(), "malformed proof, InterimNodeTypes is larger than max key length allowed (8193 > 8192)")
+
+	// issue with the size of InterimNodeTypes
+	proof.InterimNodeTypes = append(InterimNodeTypesBackup, byte(0))
+	err = proof.valididateFormat()
+	assert.Error(t, err)
+	require.Equal(t, err.Error(), "malformed proof, the length of InterimNodeTypes doesn't match the length of ShortPathLengths and SiblingHashes")
+
+	proof.InterimNodeTypes = InterimNodeTypesBackup
```
AlexHentschel, line 107: Do we have tests for the following failure conditions (?)
1. auxiliary bits at the end of `InterimNodeTypes` (to pad it to full bytes) not set to zero 
2. number of bits set to `1` in `InterimNodeTypes` different than `ShortPathLengths`
3. number of bits set to `0` in `InterimNodeTypes` (ignoring auxiliary bits at the end) different than `SiblingHashes`
   - comment: we test that a proof is rejected where `InterimNodeTypes` is too long here: https://github.com/onflow/flow-go/blob/d36176260dbe89bce7cb2fe8bf391487feb4044b/storage/merkle/proof_test.go#L102 But I specifically means to test the case, where it has the right size, but the counts of `0`s and `1`s don't match up

storage/merkle/proof_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// get proofs for keys and verify for a subset of keys
+	for _, key := range keys[:numberOfProofsToVerify] {
+		proof, existed := tree1.Prove(key)
+		require.True(t, existed)
+		isValid, err := proof.Verify(tree1.Hash())
+		assert.NoError(t, err)
+		require.True(t, isValid)
+	}
+}
```
AlexHentschel, line 166: Do we have a test that checks correct handling of a syntactically properly formatted proof, which does not match the `expectedRootHash`?

storage/merkle/tree.go
```diff
@@ -273,6 +268,99 @@ GetLoop:
 	}
 }
 
+// Prove constructs inclusion proof for a given key if the key exists in the trie.
```
AlexHentschel, line 271:
```suggestion
// Prove constructs an inclusion proof for the given key, provided the key exists in the trie.
```

storage/merkle/tree.go
```diff
@@ -273,6 +268,99 @@ GetLoop:
...
+// It traverses the trie from top to down and collects data for proof as follows:
+//  - if full node, capture the sibling node hash value and append zero to short counts
+//  - if short node, appends the node.shortCount to the short count list
+//  - if leaf, would capture the leaf value
+func (t *Tree) Prove(key []byte) (*Proof, bool) {
```
AlexHentschel, line 276: A brief documentation of the return values would be great. Thanks

storage/merkle/tree.go
```diff
@@ -273,6 +268,99 @@ GetLoop:
...
+
+			index++
+			siblingHashes = append(siblingHashes, sibling.Hash())
+			shortNodeVisited = append(shortNodeVisited, false)
+			steps++
+
+			continue ProveLoop
+
+		// if we have a short path, we can only follow the path if we have all
+		// of the short node path in common with the key
```
AlexHentschel, line 321:
```suggestion
		// if we have a short node, we can only follow the path if the key's subsequent 
		// bits match the entire path segment of the short node. 
```

storage/merkle/tree.go
```diff
@@ -273,6 +268,99 @@ GetLoop:
...
+
+			cur = &n.child
+			index += n.count
+			shortPathLengths = append(shortPathLengths, n.CountAsUint16Encoding())
+			shortNodeVisited = append(shortNodeVisited, true)
+			steps++
+
+			continue ProveLoop
+
+		// if we have a leaf, we found the key, return value and true
```
AlexHentschel, line 339:
```suggestion
		// if we have a leaf, we found the key, return proof and true
```

AlexHentschel (commented):
Code is much clearer now with the refactoring. Thank you. Great documentation. 
* I think the test coverage might still be incomplete 
* there is a bug in the verification logic

storage/merkle/proof.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// validate number of bits that is going to be checked matches the size of the given key
+	keyBitCount := len(p.SiblingHashes)
+	for _, sc := range p.ShortPathLengths {
+		keyBitCount += int(sc)
+	}
+
+	if len(p.Key)*8 != keyBitCount {
+		return NewMalformedProofErrorf("key length doesn't match the length of ShortPathLengths and SiblingHashes")
+	}
```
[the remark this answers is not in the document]
ramtinms: Nice catch, fixed the bug and applied suggestion, going to add a test for this special case

storage/merkle/errors.go
ramtinms: added

storage/merkle/tree.go
ramtinms: added

storage/merkle/errors.go
ramtinms: added

storage/merkle/proof_test.go
ramtinms: good suggestion, updated the test to cover all cases

storage/merkle/proof_test.go
ramtinms: updated

storage/merkle/proof_test.go
ramtinms: 1. I'll add one 
2. we do actually have it, the error message was not informative, so I updated 
3. I tried to add one but it was always exiting another path (step size check or shortPathLengths), I'll try again

storage/merkle/proof_test.go
ramtinms: added an extra case to fail to make sure we are calling validate

storage/merkle/proof.go
```diff
@@ -0,0 +1,210 @@
...
+	// zero should be mapped to 65536 (all other values are the same)
+	ShortPathLengths []uint16
+	// SiblingHashes is read when we reach a full node. The corresponding element represents
+	// the hash of the non-visited sibling node for each full node on the path. Elements are ordered from root to leaf.
+	SiblingHashes [][]byte
+}
+
+const maxStepsForProofVerification = maxKeyLength * 8
+
+// valididateFormat validates the format and size of elements of the proof (syntax check)
```
AlexHentschel, line 38:
```suggestion
// validateFormat validates the format and size of elements of the proof (syntax check)
```

AlexHentschel (approved):
Thanks Ramtin. Code is very clean now, easy to follow, and well document ✨

storage/merkle/tree.go
```diff
@@ -273,6 +268,100 @@ GetLoop:
 	}
 }
 
+// Prove constructs an inclusion proof for the given key, provided the key exists in the trie.
+// It returns a proof and a boolean (true if key exist and false if key not found)
+// Proof is constructed by traversing the trie from top to down and collects data for proof as follows:
+//  - if full node, capture the sibling node hash value and append zero to short counts
```
fxamacker, line 274: Maybe something like this to sync comment with new code

```suggestion
//  - if full node, append the sibling node hash value to sibling hash list
```

fxamacker (approved):
LGTM! 👍 Just a minor note to update a comment to match newer code.

storage/merkle/tree.go
```diff
@@ -24,6 +20,13 @@ var (
 // efficient. In summary, we are able to represent key length of up to 65536 bits, i.e. 8192 bytes.
 const maxKeyLength = 8192
 
+var EmptyTreeRootHash []byte
+
+func init() {
+	h, _ := blake2b.New256([]byte{})
+	EmptyTreeRootHash = h.Sum(nil)
+}
+
```
AlexHentschel, line 29: * for the test we use a static value (fast)
* but for the run time, we are working with an initialization logic (slower)

for performance reasons, how about we switch the logic:
```suggestion
var EmptyTreeRootHash []byte = []byte{
	14, 87, 81, 192, 38, 229, 67,
	178, 232, 171, 46, 176, 96, 153,
	218, 161, 209, 229, 223, 71, 119,
	143, 119, 135, 250, 171, 69, 205,
	241, 47, 227, 168}
```

and place the following computation in the test (?)
```golang
	h, _ := blake2b.New256([]byte{})
	EmptyTreeRootHash = h.Sum(nil)
```

AlexHentschel (approved):

storage/merkle/tree.go
ramtinms: we could but init would only run once during life time of EN

ramtinms:
bors merge

ramtinms:
bors merge

storage/merkle/tree.go
```diff
@@ -273,6 +277,100 @@ GetLoop:
...
+	// and we start at a zero index in the path
+	index := 0
+
+	// init proof params
+	siblingHashes := make([][]byte, 0)
+	shortPathLengths := make([]uint16, 0)
+
+	steps := 0
+	shortNodeVisited := make([]bool, 0, len(key))
```
tarakby, line 304: Not sure what is the reason of allocating `len(key)` for the short node visited slice but not for the others? 
If there is no specific reason, it would make sense to make the approximative allocation for all the 3 slices or for nothing, just for consistency. 

`len(key)` is also not the maximum length the slice could have, the maximum would be the bit size of the key I think, `len(key)>>3`.

tarakby (commented):
Nicely written and organized code 🏄🏼 

I left a few comments. Since the PR is merged, we can address these in a future comment if you are working on one, or I can follow up with another PR if you're done with the trie.

tarakby:
@ramtinms FYI I'm addressing my own comments in a separate branch, but feel free to comment on my own comments if there is something I misunderstood.

storage/merkle/tree.go
ramtinms: I think this is a leftover from the past, we could definitely update it.

storage/merkle/proof.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+// Verify verifies the proof by constructing the hash values bottom up and cross-check
+// the constructed root hash with the given one. For valid proofs, `nil` is returned.
+// During normal operations, the following error returns are expected:
+//  * MalformedProofError if the proof has a syntactically invalid structure
+//  * InvalidProofError if the proof is syntactically valid, but the reconstructed
+//    root hash does not match the expected value.
+func (p *Proof) Verify(expectedRootHash []byte) error {
```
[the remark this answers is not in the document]
ramtinms: Yeah I thought about it, but I also didn't find a way anyone can attack this considering that every hash at every level is not reversible, so if you could construct a valid proof with a different key size, you probably could construct another valid proof with the same size but different subtrees. but happy to change if we want to change it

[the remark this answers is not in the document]
AlexHentschel: Sorry for the late reply. Here are my thoughts:

Lets start with a Lemma:
```
Every Merkle path from the root to a leaf contains the full key, bit by bit.
```
reasoning: 
* consider the node instances that form the merkle-path (go struct instances in memory)
* from these structs, you can reconstruct the exact key (hence also the key length)
* the entire information of these structs in successively hashed, one node at a time, using a bijective binary representation


Starting from this premiss, a safety proof could look something like
* let `r` be a root hash for the set of key-value pairs `S = {(k,v) | where all k have identical length}`
* now, consider a proof `P` for a different key-value pair `(k', v')` where `k'` has a different length than `k`
* As part of verifying proof `P`, the verifier re-constructs the merkle path, from which is then locally computes the root hash (in the implementation, both steps happens on the fly, but conceptually the path reconstructions is there). As the reconstructed merkle path exactly contains `k'`, forging a proof requires a hash collision.

PR #1839, opened 2022-01-14
Migration for Ordered Map storage change

dsainati1:
Closes https://github.com/dapperlabs/flow-go/issues/6027

https://github.com/onflow/cadence/pull/1248 changed the way that data is stored in accounts on Flow, so existing data needs to be migrated to the new format. Luckily the change is fairly simple; this migration sorts the storage entries by owner and domain, and then inserts them into the new atree OrderedMap format.


cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,473 @@
...
+			for _, pair := range keyValuePairs {
+				storageMap := m.newStorage.GetStorageMap(common.MustBytesToAddress([]byte(owner)), domain)
+				// in the interest of not having to update Cadence and break its abstractions to allow
+				// this one-time migration, just use reflection to grab the unexported field
+				unsafeOrderedMap := reflect.ValueOf(storageMap).Elem().FieldByName("orderedMap")
+				orderedMap := reflect.NewAt(
+					unsafeOrderedMap.Type(),
+					unsafe.Pointer(unsafeOrderedMap.UnsafeAddr()),
+				).Elem().Interface().(*atree.OrderedMap)
+				orderedMap.Set(
```
fxamacker, line 456: It would be safer not to use `RawStorable` as `atree.Value`.  `RawStorable.Storable()` returns the entire raw bytes.

For example, if `RawStorable` represents a Cadence array, `Storable()` needs to return array's storage ID instead because atree stores storage ID for mutable and very large immutable elements.

cmd/util/ledger/migrations/ordered_map_migration_test.go
```diff
@@ -0,0 +1,278 @@
...
+		six := interpreter.NewIntValueFromInt64(1)
+
+		payload := []ledger.Payload{
+			{Key: createAccountPayloadKey(address1, "storage\x1fFoo"), Value: encodeValue(one, address1)},
+			{Key: createAccountPayloadKey(address2, "storage\x1fFoo"), Value: encodeValue(two, address2)},
+			{Key: createAccountPayloadKey(address3, "storage\x1fFoo"), Value: encodeValue(three, address3)},
+			{Key: createAccountPayloadKey(address1, "storage\x1fBar"), Value: encodeValue(four, address1)},
+			{Key: createAccountPayloadKey(address1, "storage\x1fBaz"), Value: encodeValue(five, address1)},
+			{Key: createAccountPayloadKey(address3, "public\x1fFoo"), Value: encodeValue(six, address3)},
+		}
```
turbolent, line 155: This example data here looks like it correctly represents the current encoding scheme: Paths are encoded like this, and the values are encoded storables.

However, the test cases here only cover the fixed-size `Int64` type. As the encoding of the values to storables differs depending on type and size, could you please also add test cases / example payloads where the storable representation is not the inlined encoded data?

For example, a small integer or string is directly stored "inline" in the value. However, large immutable values (like large strings) and especially mutable Cadence values / containers, like Cadence arrays, dictionaries, and composites, are actually stored in a separate register/slab, and the "top-level"/"entry" registers here are just `StorageIDStorable`s, effectively "pointing" to the registers holding the actual data.

turbolent (commented):
Great work!

Have you tried this migration already on a recent Testnet and Mainnet snapshot?

cmd/util/ledger/migrations/ordered_map_migration_test.go
dsainati1: I thought the previous test checked this case, but it's possible I am wrong. Does creating and storing an array/dictionary not exercise this case?

fvm/state/state.go
```diff
@@ -306,6 +306,9 @@ func IsFVMStateKey(owner, controller, key string) bool {
 		if key == KeyStorageUsed {
 			return true
 		}
+		if key == KeyStorageIndex {
+			return true
+		}
```
ramtinms, line 311: ✨

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,272 @@
...
+		owner, entry :=
+			string(p.Key.KeyParts[0].Value),
+			string(p.Key.KeyParts[2].Value)
+		// if the entry doesn't contain a separator, we just ignore it
+		// since it is storage metadata and does not need migration
+		if !strings.Contains(entry, "storage\x1f") &&
+			!strings.Contains(entry, "public\x1f") &&
+			!strings.Contains(entry, "contract\x1f") &&
+			!strings.Contains(entry, "private\x1f") {
+			m.Log.Info().Msgf("Ignoring key in storage payloads: %s", entry)
```
ramtinms, line 229: maybe using warning here instead of Info

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,272 @@
...
+func (m *OrderedMapMigration) migrate(storagePayloads []ledger.Payload) ([]ledger.Payload, error) {
+	groupedByOwnerAndDomain := make(map[string](map[string][]Pair))
+
+	for _, p := range storagePayloads {
+		owner, entry :=
+			string(p.Key.KeyParts[0].Value),
+			string(p.Key.KeyParts[2].Value)
+		// if the entry doesn't contain a separator, we just ignore it
+		// since it is storage metadata and does not need migration
+		if !strings.Contains(entry, "storage\x1f") &&
```
ramtinms, line 225: shouldn't we use HasPrefix instead of Contains ?

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,272 @@
...
+			continue
+		}
+		splitEntry := strings.Split(entry, "\x1f")
+		domain, key := splitEntry[0], splitEntry[1]
+		value := p.Value
+
+		domainMap, domainOk := groupedByOwnerAndDomain[owner]
+		if !domainOk {
+			domainMap = make(map[string][]Pair)
+		}
```
ramtinms, line 239: This assumes if we don't have any value on a specific domain for an account we won't create an empty map for it, just checking this won't cause any issue adding a new value later to that domain? or cadence would expect the map exists for all domains for all accounts

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,272 @@
...
+		if !orderedOk {
+			keyValuePairs = make([]Pair, 0)
+		}
+		domainMap[domain] = append(keyValuePairs, Pair{Key: key, Value: value})
+		groupedByOwnerAndDomain[owner] = domainMap
+	}
+
+	for owner, domainMaps := range groupedByOwnerAndDomain {
+		for domain, keyValuePairs := range domainMaps {
+			address, err := common.BytesToAddress([]byte(owner))
```
ramtinms, line 250: we had some issues in the past with addresses with leading zeros, I don't remember the details on what was the issue (maybe using Bytes() method), just want to make sure we double check if this one is safe cc: @SupunS

go.mod
```diff
@@ -105,3 +105,5 @@ require (
 )
 
 replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
+
+replace github.com/onflow/cadence => ./cadence
```
ramtinms, line 109: we probably don't want to commit this, we need to change it to use cadence at specific tag

fvm/transactionEnv.go
```diff
@@ -953,6 +953,22 @@ func (e *TransactionEnv) ValueDecoded(duration time.Duration) {
 	e.metrics.ValueDecoded(duration)
 }
 
+func (m *TransactionEnv) BLSVerifyPOP(_ *runtime.PublicKey, _ []byte) (bool, error) {
+	panic("unexpected BLSVerifyPOP call")
```
ramtinms, line 957: I think a better way might be returning this error instead of panic (then we don't have to go through recovery, etc, ...)
e.g. return errors.NewOperationNotSupportedError("BLSVerifyPOP")

fvm/transactionEnv.go
```diff
@@ -953,6 +953,22 @@ func (e *TransactionEnv) ValueDecoded(duration time.Duration) {
...
+	panic("unexpected BLSAggregateSignatures call")
+}
+
+func (m *TransactionEnv) AggregateBLSPublicKeys(_ []*runtime.PublicKey) (*runtime.PublicKey, error) {
+	panic("unexpected BLSAggregatePublicKeys call")
+}
+
+func (m *TransactionEnv) ResourceOwnerChanged(*interpreter.CompositeValue, common.Address, common.Address) {
+	panic("unexpected ResourceOwnerChanged call")
+}
```
ramtinms, line 970: same for these

fvm/transactionEnv.go
```diff
@@ -953,6 +953,22 @@ func (e *TransactionEnv) ValueDecoded(duration time.Duration) {
...
+func (m *TransactionEnv) AggregateBLSSignatures(_ [][]byte) ([]byte, error) {
+	panic("unexpected BLSAggregateSignatures call")
+}
+
+func (m *TransactionEnv) AggregateBLSPublicKeys(_ []*runtime.PublicKey) (*runtime.PublicKey, error) {
+	panic("unexpected BLSAggregatePublicKeys call")
+}
+
+func (m *TransactionEnv) ResourceOwnerChanged(*interpreter.CompositeValue, common.Address, common.Address) {
+	panic("unexpected ResourceOwnerChanged call")
```
ramtinms, line 969: this actually should be no-op otherwise the cadence runtime won't work

fvm/scriptEnv.go
```diff
@@ -673,6 +673,22 @@ func (e *ScriptEnv) ValueDecoded(duration time.Duration) {
 	e.metrics.ValueDecoded(duration)
 }
 
+func (m *ScriptEnv) BLSVerifyPOP(_ *runtime.PublicKey, _ []byte) (bool, error) {
+	panic("unexpected BLSVerifyPOP call")
```
ramtinms, line 677: same as TxEnv

fvm/fvm.go
```diff
@@ -56,8 +56,8 @@ func (vm *VirtualMachine) Run(ctx Context, proc Procedure, v state.View, program
 			// Cadence may fail to encode certain values.
 			// Return an error for now, which will cause transactions to revert.
 			//
-			if encodingErr, ok := r.(interpreter.EncodingUnsupportedValueError); ok {
-				err = errors.NewEncodingUnsupportedValueError(encodingErr.Value, encodingErr.Path)
+			if _, ok := r.(interpreter.UnsupportedTagDecodingError); ok {
+				//err = errors.NewEncodingUnsupportedValueError(encodingErr.Value, encodingErr.Path)
```
ramtinms, line 60: not sure why we commented this out, @janezpodhostnik is this safe ?

cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -365,40 +371,33 @@ func (c *balanceProcessor) storageUsed(address flow.Address) (uint64, error) {
...
-func resourceId(address flow.Address, path interpreter.PathValue) flow.RegisterID {
-	// Copied logic from interpreter.storageKey(path)
-	key := fmt.Sprintf("%s\x1F%s", path.Domain.Identifier(), path.Identifier)
-
-	return flow.RegisterID{
-		Owner:      string(address.Bytes()),
-		Controller: "",
-		Key:        key,
+func (c *balanceProcessor) ReadStored(address flow.Address, domain common.PathDomain, id string) (cadence.Value, error) {
+	addr, err := common.HexToAddress(address.Hex())
```
ramtinms, line 391: I think we can do this more efficient using bytes than conversion to hex and decode from hex,

cmd/util/ledger/migrations/ordered_map_migration.go
dsainati1: cc @turbolent this is a good question, if an account has nothing in `storage` for example, is that represented by an empty map, or should there be nothing associated with the `"storage"` key? If something is later added to that account's `storage`, is the map created when its first element is added?

fvm/transactionEnv.go
dsainati1: Ah this was left over from what I was basing this on Cadence master, it shouldn't be necessary anymore.

fvm/transactionEnv.go
dsainati1: same as above

fvm/fvm.go
dsainati1: Reverting this change

janezpodhostnik: EncodingUnsupportedValueError was actually removed from cadence, so this whole part of the code can be removed, But that should probably be a different PR (like this one: https://github.com/onflow/flow-go/pull/1975)

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (r RawStorable) ChildStorables() []atree.Storable {
+	return nil
+}
+
+func (r RawStorable) Storable(_ atree.SlabStorage, _ atree.Address, _ uint64) (atree.Storable, error) {
+	return r, nil
+}
+
+func splitPayloads(inp []ledger.Payload) (fvmPayloads []ledger.Payload, storagePayloads []ledger.Payload, slabPayloads []ledger.Payload) {
```
janezpodhostnik, line 187: This looks useful! It could go into `utils.go`.

janezpodhostnik (commented):
One minor comment from my side. Looking good.

turbolent (commented):
Looks good, just needs to pass CI

turbolent (commented):
I mainly only reviewed the migration and its tests. I don't know much about the reporters

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,251 @@
...
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/onflow/atree"
+	"github.com/onflow/cadence/runtime"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/onflow/flow-go/fvm/state"
```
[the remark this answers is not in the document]
ramtinms: I'm going to fix it.

ramtinms (approved):

cmd/util/ledger/migrations/ordered_map_migration_test.go
```diff
@@ -0,0 +1,240 @@
...
+	cadenceAddress, _ := common.HexToAddress("0x1")
+
+	_, _ = mig.initialize([]ledger.Payload{
+		{Key: createAccountPayloadKey(address1, state2.KeyExists), Value: []byte{1}},
+		{Key: createAccountPayloadKey(address1, state2.KeyStorageUsed), Value: utils.Uint64ToBinary(1)},
+	})
+
+	encodeValue := func(v interpreter.Value) ledger.Value {
+		storable, _ := v.Storable(mig.NewStorage, atree.Address(address1), math.MaxUint64)
+		encodedInt, _ := atree.Encode(storable, interpreter.CBOREncMode)
```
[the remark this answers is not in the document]
ramtinms: Updated those

cmd/util/ledger/migrations/ordered_map_migration.go
```diff
@@ -0,0 +1,251 @@
...
+	fvmPayloads, storagePayloads, slabPayloads := splitPayloads(payload)
+
+	m.ledgerView = NewView(append(fvmPayloads, slabPayloads...))
+	m.initPersistentSlabStorage(m.ledgerView)
+	m.initIntepreter()
+	return storagePayloads, nil
+}
+
+func (m *OrderedMapMigration) migrate(storagePayloads []ledger.Payload) ([]ledger.Payload, error) {
+	groupedByOwnerAndDomain := make(map[string](map[string][]Pair))
```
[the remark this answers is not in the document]
ramtinms: I think it should be fine, maybe we could do a follow up PR to fix it if it was time consuming to migrate the state on the canary.

ramtinms:
bors merge

turbolent (approved):
Agreed! Let's get it in for Canary and then maybe can address points in a follow-up PR 👍

PR #1847, opened 2022-01-17
Leo/v0.23.6 fail tx

zhangchiqing:


Kay-Zee (approved):
Trying on testnet only

PR #1849, opened 2022-01-17
skip starting dkg reactor engine if node not part of dkg committee

kc1116:
This PR refactors the logic to check if node is part of the DKG committee, and moves this check to the reactor_engine where we can log a warning and return rather than shut down the node with a fatal log.


module/dkg/utils.go
```diff
@@ -0,0 +1,22 @@
+package dkg
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+func GetDKGCommitteeIndex(nodeID flow.Identifier, participants flow.IdentityList) (int, error) {
```
zhangchiqing, line 9: Can we move this function to `model/flow/identity.go` to be something like `GetIndex(flow.Identifier) (uint, bool)`?

module/dkg/utils.go
```diff
@@ -0,0 +1,22 @@
...
+	myIndex := -1
+	for i, id := range participants.NodeIDs() {
+		if id == nodeID {
+			myIndex = i
+			break
+		}
+	}
+	if myIndex < 0 {
+		return myIndex, fmt.Errorf("node does not belong to dkg committee")
```
zhangchiqing, line 18: if we use an error, better to include the nodeID in it

engine/consensus/dkg/reactor_engine.go
```diff
@@ -181,6 +174,13 @@ func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow
 		Interface("members", committee.NodeIDs()).
 		Msg("epoch info")
 
+	_, err = dkgmodule.GetDKGCommitteeIndex(e.me.NodeID(), committee)
+	if err != nil {
+		// node not found in DKG committee bypass starting the DKG
+		log.Warn().Err(err).Msg("node index not found in dkg committee")
```
zhangchiqing, line 180: better to be explicit to say something like: node (%v) is not part of DKG committee, skip starting DKG engine

zhangchiqing (approved):
Looks good, some minor comments

module/dkg/utils.go
kc1116: Thanks for this

engine/consensus/dkg/reactor_engine.go
```diff
@@ -201,6 +201,13 @@ func (e *ReactorEngine) startDKGForEpoch(currentEpochCounter uint64, first *flow
 		}
 	})
 
+	// flag that we are starting the dkg for this epoch
+	err = e.dkgState.SetDKGStarted(nextEpochCounter)
```
jordanschalm, line 204: I think we should still set `DKGStarted` before starting the DKG process. The purpose of this flag currently is to avoid restarting a DKG which we have already started, since we are unable to recover an in-progress DKG state. We can be sure of this if we set the flag before starting the DKG. 

Though it's unlikely that the DKG controller would eg. send a message and the node restart before setting the flag, it's less easy to reason about if we asynchronously start the DKG before setting the flag.

engine/consensus/dkg/reactor_engine.go
jordanschalm: In particular, we can log that we were unable to find _ourself_ in the DGK committee. We can add in comments that this indicates the node is un-staking from the network and will not be a participant beginning in the subsequent epoch.

module/dkg/utils.go
```diff
@@ -0,0 +1,22 @@
+package dkg
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+func GetDKGCommitteeIndex(nodeID flow.Identifier, participants flow.IdentityList) (int, error) {
+	myIndex := -1
```
jordanschalm: Nit: the function here is now more general (find _any_ node's index), but the variable name still reflects the more specific use the code had previously (find the local node's index).
```suggestion
	nodeIndex := -1
```

engine/consensus/dkg/reactor_engine.go
kc1116: https://github.com/onflow/flow-go/pull/1849/commits/aa369fc7718a37db5db1e76cb77481bc0c1cca33

module/dkg/utils.go
kc1116: https://github.com/onflow/flow-go/pull/1849/commits/aa369fc7718a37db5db1e76cb77481bc0c1cca33

module/dkg/utils.go
kc1116: https://github.com/onflow/flow-go/pull/1849/commits/aa369fc7718a37db5db1e76cb77481bc0c1cca33

module/dkg/utils.go
kc1116: https://github.com/onflow/flow-go/pull/1849/commits/aa369fc7718a37db5db1e76cb77481bc0c1cca33

engine/consensus/dkg/reactor_engine.go
kc1116: https://github.com/onflow/flow-go/pull/1849/commits/ae869441320976a92be677daf76299ea178572c5

jordanschalm (approved):
🎸

module/dkg/controller_factory.go
```diff
@@ -49,30 +49,24 @@ func (f *ControllerFactory) Create(
...
-		if id == f.me.NodeID() {
-			myIndex = i
-			break
-		}
-	}
-	if myIndex < 0 {
-		return nil, fmt.Errorf("node does not belong to dkg committee")
+	myIndex, ok := participants.GetIndex(f.me.NodeID())
+	if !ok {
+		return nil, fmt.Errorf("failed to create controller factory, node %s is not part of DKG committee", f.me.NodeID().String())
```
tarakby, line 54: Do we need to error in this case? Shouldn't it be a "normal" case that a consensus node is not part of the beacon committee? 
cc @jordanschalm

jordanschalm: It is an expected case that's true. A sentinel would be better so the caller can distinguish between "expected, but we can't initialize the DKG" and "unexpected error"

PR #1778, opened 2021-12-15
[Optimization] Implements GC-optimized cache

yhassanzadeh13:
This PR implements a GC-optimized `BackData` that is used as the model layer of `Backend` in our `mempools`. The existing `BackData` is `map`-based, which creates a linear number of objects on the heap. This causes garbage collection off-heap to ramp up as the mempool size grows. As the solution, this PR implements an array-based implementation that creates a fixed number of heap-objects _regardless_ of the total objects it stores. Our benchmarking results shows that the GC-optimized one is around **3-4 times faster on garbage collection** than the existing `hashicorp` and `map`-based mempools we currently use. It is also **15%** faster on reads and writes, and always keeps around 3-4 objects on the heap (even when we maintain 100M objects, which is in contrast to the existing ones that take 300M-400M objects on heap for storing 100M ones).

PR #1852, opened 2022-01-17
[Optimizations] Optimizes All method of BackData

yhassanzadeh13:
**Refactoring BackData Interface**: The mempool's `BackData` interface has an `All() map[flow.Identifier]flow.Entity` method, which adapts the backward compatibility introduced by the map-based implementation. However, for some use cases, returning a map by the `All` method is a heavy lift since a slice would suffice. Creating a slice is usually going to be faster and use less memory than creating a map. Because appending to a slice appends at the next cache line, whereas appending to a map results in random access across the map. This PR refactors `BackData` interface with two other methods `Entities` and `Identifiers` that return the slice of all stored entities and their identifiers, respectively. It also refactors use cases of `All` with the corresponding `Entities` and `Identifiers` where it fits.

**Optimizing All Method for HeroCache**:  The array-based `heroCache` implementation of `BackData` does heavy lifting on each `All` method invocation to turn its array-based cache into a map (see [this comment](https://github.com/onflow/flow-go/pull/1778#discussion_r782649163)). With a limit of `l` and bucket size of `s`, calling all would cause `O(l * s)` iterations regardless of the actual number of stored entities. This PR, also optimizes the implementation of `All` from `O(l * s)` to `O(n)` where `n < l` is the total number of stored entities.


integration/utils/contLoadGenerator.go
```diff
@@ -30,6 +30,7 @@ const (
 )
 
 var accountCreationBatchSize int = 100
+
```
yhassanzadeh13, line 33: May skip reviewing this file, it is just a `go fmt` change.

module/mempool/stdmap/backdata/herocache/cache_test.go
```diff
@@ -249,24 +249,24 @@ func TestArrayBackData_All(t *testing.T) {
...
+			limit:        1000,
+			items:        10_000,
 			ejectionMode: heropool.LRUEjection,
 		},
-		{ // mempool has the limit of 100K, and we put 1M items with random ejection.
-			limit:        100_000,
-			items:        1_000_000,
+		{ // mempool has the limit of 1000, and we put 10K items with random ejection.
+			limit:        1000,
+			items:        10_000,
```
yhassanzadeh13, line 269: Lightening the tests to save up CI time.

module/mempool/stdmap/eject_test.go
```diff
@@ -228,14 +228,3 @@ func TestLRUEjector_EjectAll(t *testing.T) {
 		require.Equal(t, id, items[i])
 	}
 }
-
-// MockEntity is a mocked entity type used in internal testing of ejectors.
```
yhassanzadeh13, line 232: replaced with `unittest.MockEntity`.

module/mempool/stdmap/eject.go
```diff
@@ -165,7 +165,7 @@ func (q *LRUEjector) Eject(b *Backend) (flow.Identifier, flow.Entity, bool) {
 	// finds the oldest entity
 	oldestSQ := uint64(math.MaxUint64)
 	var oldestID flow.Identifier
-	for id := range b.backData.All() {
```
simonhf, line 168: Am a bit confused about this change. In `eject.go` then there are two other functions -- `EjectTrueRandom()` and `EjectTrueRandomFast()` -- which also do `range b.backData.All()` but those functions have not been refactored, only this one?

Also, presumably none of `Eject()`, `EjectTrueRandom()`, or `EjectTrueRandomFast()` are used anyway with the new `heroCache` efficient eject mechanisms, or? So why refactor these at all?

Also, how to to ensure that they are not accidentally used in addition to the new `heroCache` efficient eject mechanisms?

simonhf (commented):
Made some general questions, comments, and suggestions :-)

module/mempool/stdmap/backDataHeapBenchmark_test.go
```diff
@@ -193,14 +193,47 @@ func (b baselineLRU) All() map[flow.Identifier]flow.Entity {
...
 		all[id] = entity
 	}
 
 	return all
 }
 
+func (b baselineLRU) Identifiers() flow.IdentifierList {
+	ids := make(flow.IdentifierList, b.c.Len())
+	i := 0
+	for _, entityID := range b.c.Keys() {
```
jwinkler2083233, line 207: since the 'range' keyword is causing the compiler to generate a for loop, and since 'Keys()' is an array type, it would be preferable to use 'i' like this:  'for i:=0;i < keyArrayLen;i++' and avoid using the 'range' keyword entirely.  That cuts the computational work in half.

module/mempool/stdmap/backDataHeapBenchmark_test.go
```diff
@@ -193,14 +193,47 @@ func (b baselineLRU) All() map[flow.Identifier]flow.Entity {
...
+		ids[i] = id
+		i++
+	}
+	return ids
+}
+
+func (b baselineLRU) Entities() []flow.Entity {
+	entities := make([]flow.Entity, b.c.Len())
+	i := 0
+	for _, entityID := range b.c.Keys() {
```
jwinkler2083233, line 221: See above comment.  Let's replace 'range' with a 'for' loop.

module/mempool/stdmap/backdata/herocache/cache.go
```diff
@@ -177,21 +177,31 @@ func (c Cache) All() map[flow.Identifier]flow.Entity {
 	defer c.logTelemetry()
 
 	all := make(map[flow.Identifier]flow.Entity)
```
jwinkler2083233, line 179: maybe you can pre-allocate space in the map, by adding a parameter to make with the length of 'entities.All()' ?

module/mempool/stdmap/backdata/mapBackData.go
```diff
@@ -82,6 +82,26 @@ func (b MapBackData) All() map[flow.Identifier]flow.Entity {
...
+		ids[i] = entityID
+		i++
+	}
+	return ids
+}
+
+func (b MapBackData) Entities() []flow.Entity {
+	entities := make([]flow.Entity, len(b.entities))
+	i := 0
+	for _, entity := range b.entities {
```
jwinkler2083233, line 98: maybe replace 'range' with 'for' loop

module/mempool/stdmap/backdata/mapBackData.go
```diff
@@ -82,6 +82,26 @@ func (b MapBackData) All() map[flow.Identifier]flow.Entity {
 	return entities
 }
 
+func (b MapBackData) Identifiers() flow.IdentifierList {
+	ids := make(flow.IdentifierList, len(b.entities))
+	i := 0
+	for entityID := range b.entities {
```
jwinkler2083233, line 88: maybe replace 'range' with 'for' loop, or maybe use the 'for i, entityID' syntax.

jwinkler2083233 (changes requested):

module/mempool/stdmap/backDataHeapBenchmark_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1852/commits/f2768318602b479b6a3ad0f9c96db6370433310b

module/mempool/stdmap/backDataHeapBenchmark_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1852/commits/f2768318602b479b6a3ad0f9c96db6370433310b

module/mempool/stdmap/backdata/herocache/cache.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1852/commits/d62f7c79438a16f96d13781557064140c6e154a5

module/mempool/stdmap/backdata/mapBackData.go
yhassanzadeh13: `b.entities` is a map and we may not be able to iterate on it using a for loop unless through reflection. Not sure if it is the desired case?

module/mempool/stdmap/backdata/mapBackData.go
yhassanzadeh13: `b.entities` is a map and we may not be able to iterate on it using a for loop unless through reflection. Not sure if it is the desired case?

module/mempool/backData.go
```diff
@@ -32,6 +32,12 @@ type BackData interface {
 	// All returns all entities from the pool.
 	All() map[flow.Identifier]flow.Entity
 
+	// Identifiers returns the list of identifiers (keys) of entities stored in the backdata.
+	Identifiers() flow.IdentifierList
```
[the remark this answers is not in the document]
yhassanzadeh13: > Why is this the long form Identifiers() and not the short form IDs()... which would be inline with the above function ByID()? Or vica-versa, why not refactor ByID() to ByIdentifier() to match the long form of this function?

`ByID` and `Identifiers` are typical naming conventions across our codebase for methods that are serving the same purposes: 
`ByID` refers to retrieving an `Entity` type interface using its `flow.Identifier`, [example](https://github.com/onflow/flow-go/blob/608f40d0d57afee287f9a5608abc5f8a3688ec5d/network/p2p/protocol_state_provider.go#L104).
`Identifers` refers to returning the list of identifiers maintained by an identifier provider, e.g., a cache or database, [example](https://github.com/onflow/flow-go/blob/df1c6d59f1ece197411b37da30573d5c143e0e5c/module/mempool/stdmap/transactions.go#L32).

Following such naming conventions makes the code more understandable and coherent. Unless exceptional circumstances, let's not go with ad-hoc namings for such widely used methods that already possess an established naming convention. 

> Just out of interest, why is this flow.IdentifierList and not []flow.Identifier? Or vice-versa, why is it []flow.Entity below and not flow.EntityList? :-)

`flow.IdentifierList` is a [type](https://github.com/onflow/flow-go/blob/c007c2f3b3548cd125c555fd5f1becbbed8deae9/model/flow/identifierList.go#L11). We defined it to furnish it with a set of methods that are widely used across our codebase, e.g., [`Union`](https://github.com/onflow/flow-go/blob/c007c2f3b3548cd125c555fd5f1becbbed8deae9/model/flow/identifierList.go#L76). Moreover, we made the type sortable for a wide variety of use cases. Hence, it is more preferable than the bare `[]flow.Identifier`. On the other hand, the need for `[]flow.EntityList` has not been yet raised, hence we stick to the bare `[]flow.Entity`.

module/mempool/backData.go
```diff
@@ -32,6 +32,12 @@ type BackData interface {
 	// All returns all entities from the pool.
 	All() map[flow.Identifier]flow.Entity
 
+	// Identifiers returns the list of identifiers (keys) of entities stored in the backdata.
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1852/commits/dc62370e81bea2bcffda030cd05c6a7c29b5fee8

module/mempool/stdmap/eject.go
yhassanzadeh13: > Also, presumably none of Eject(), EjectTrueRandom(), or EjectTrueRandomFast() are used anyway with the new heroCache efficient eject mechanisms, or? So why refactor these at all?

Depending on their functionality some ejections need only the entities, some only the identifiers, and some both. When both entities and identifiers are needed, the `All` is utilized, and in other cases only the relevant methods. Note that the entities may be stored with a different identifier than their `entity.ID()`. Hence, for those applications that require both identifiers and entities, we cannot derive the former from the latter. 

Correct, `eject` methods are idempotent on `heroCache`, however, since `MapBackData` is utilizing them, it can benefit from this refactor, i.e., to have a slice instead of the map as much as places as possible. 

> Also, how to to ensure that they are not accidentally used in addition to the new heroCache efficient eject mechanisms?

The `Backend` which is the controller layer of `mempools` and runs on top of a `backdata` only kicks in the `eject` [if after adding an entity to the backdata, its _size_ goes beyond its _capacity_](https://github.com/onflow/flow-go/blob/c9eb546e78bf3f62af8b00798d9d7ed11b460bc5/module/mempool/stdmap/backend.go#L209). This is only applicable to the `MapBackData` that lets the size of map grows and then scrapes any excess entities if it goes beyond its `guaranteedCapacity` parameter. Please note that this will _never happen_ in `heroCache` (which is also an implementation of `backdata`. Why? because `heroCache` always conducts the size adjustment as part of the `Add` method. So, it guarantees that by the time its `Add` method returns, the size of `heroCache` is always less than or equal to its `sizeLimit`. Hence [this if statement](https://github.com/onflow/flow-go/blob/c9eb546e78bf3f62af8b00798d9d7ed11b460bc5/module/mempool/stdmap/backend.go#L209) is always false for `heroCache`.

simonhf: Ah, that makes sense. The old `Add()` method first makes the `backdata` larger and then prunes it when it gets too big. But `heroCache` never gets too big. Very good! :-)

simonhf (approved):
Thanks! All my comments, questions, and suggestions have been answered and/or implemented :-)

module/mempool/stdmap/backdata/mapBackData.go
jwinkler2083233: looks like Go prevents getting access to a pure key collection.  You're right.

module/mempool/stdmap/backdata/mapBackData.go
jwinkler2083233: ok

jwinkler2083233 (approved):

PR #1859, opened 2022-01-18
[Sporking] Improve state extraction

zhangchiqing:
Address https://github.com/dapperlabs/flow-go/issues/6062

In order to print the state commitment earlier, this PR

- defer writing the checkpoint file to disk
- defer the report generations, like account reports, storage reports, etc.

### DIFF
--- a/cmd/util/cmd/execution-state-extract/execution_state_extract.go
+++ b/cmd/util/cmd/execution-state-extract/execution_state_extract.go
@@ -72,7 +72,11 @@ func extractExecutionState(
 		}
 
 	}
+
+	// generating reports at the end, so that the checkpoint file can be used
+	// for sporking as soon as it's generated.
 	if report {
+		log.Info().Msgf("preparing reporter files")
 		reportFileWriterFactory := reporters.NewReportFileWriterFactory(outputDir, log)
 
 		rs = []ledger.Reporter{
@@ -92,6 +96,7 @@ func extractExecutionState(
 			},
 		}
 	}
+
 	newState, err := led.ExportCheckpointAt(
 		ledger.State(targetHash),
 		migrations,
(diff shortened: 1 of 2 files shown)


cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -72,7 +72,28 @@ func extractExecutionState(
 		}
 
 	}
+	newState, err := led.ExportCheckpointAt(
```
zhangchiqing, line 75: move up so that we will generate checkpoint file before generating reports

ledger/complete/ledger.go
```diff
@@ -342,6 +342,10 @@ func (l *Ledger) ExportCheckpointAt(
 		return ledger.State(hash.DummyHash), fmt.Errorf("constructing updated trie failed: %w", err)
 	}
 
+	statecommitment := ledger.State(newTrie.RootHash())
+
+	l.logger.Info().Msgf("successfully built new trie. New statecommitment: %v", statecommitment.String())
```
zhangchiqing, line 347: log the statecommitment as soon as it's generated, so that we can continue generating root block earlier

durkmurder (approved):

janezpodhostnik (approved):
Looks good, but I do have one concern:

I tried this before, and when I was testing it I noticed that the memory consumption goes way up, which consumed all memory on the machine I was testing on and killed the process.

Have you noticed anything similar?

koko1123 (approved):

zhangchiqing:
>Have you noticed anything similar?


[I tested the changes with mainnet14](https://github.com/dapperlabs/flow-go/issues/6062#issuecomment-1021794615), I didn't run into any memory issue.

zhangchiqing:
bors merge

PR #1865, opened 2022-01-19
[Crypto] add legacy Keccak 256 implementation

tarakby:
- Add legacy Keccak implementation using the local implementation of Keccak optimized for memory allocations. 
- Only Keccak with 256-bits output is implemented. 
- unrelated change: SHA length constants are renamed to follow Go abbreviation guidelines

[ Some files were renamed which makes the global review a bit complicated. Reviewing the commits one by one would be easier in this case, starting from commit 8c5cbf9 - for some reason, all local "git pulls" are shown as commits on the PR  ]


crypto/hash/types.go
```diff
@@ -8,27 +8,40 @@ type HashingAlgorithm int
...
 	// Lengths of hash outputs in bytes
-	HashLenSha2_256 = 32
-	HashLenSha2_384 = 48
-	HashLenSha3_256 = 32
-	HashLenSha3_384 = 48
+	HashLenSHA2_256   = 32
+	HashLenSHA2_384   = 48
+	HashLenSHA3_256   = 32
+	HashLenSHA3_384   = 48
+	HashLenKeccak_256 = 32
```
ramtinms, line 45: it seems there are some dependency on this in flow-go, is it possible that there are some dependencies out there as well?

crypto/hash/sha3.go
```diff
@@ -1,250 +1,51 @@
...
-// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
-// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
-// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-const (
-	// maxRate is the maximum size of the internal buffer. SHA3-256
-	// currently needs the largest buffer.
-	maxRate = 1088 / 8
```
ramtinms, line 119: I guess we don't need this one anymore? do we have similar logic somewhere else ?
??

ramtinms (approved):
Looks good to me, I think you just need to make sure flow-go code dependent on hash size are updated with the proper names and we need to report this as breaking change

crypto/hash/sha3.go
tarakby: The same constant just moved here : https://github.com/onflow/flow-go/blob/tarak/6106-keccak-256/crypto/hash/keccak.go#L70

crypto/hash/types.go
tarakby: You're right, updated in https://github.com/onflow/flow-go/pull/1865/commits/7f6a5f63123ea11d4c9a54c2e535a734532a3ffa

tarakby:
> we need to report this as breaking change

Did you mean that external projects have been using flow-go/crypto and will need to be notified? I don't think we have treating flow-go public packages like we're treating Cadence yet, but ideally we should

crypto/hash/types.go
ramtinms: there might be some more dependencies in FCL and other client sides

crypto/hash/hash_test.go
```diff
@@ -188,6 +199,29 @@ func TestSha3(t *testing.T) {
...
+	rand.Seed(r)
+	t.Logf("math rand seed is %d", r)
+	var expected [HashLenKeccak_256]byte
+
+	for i := 0; i < 5000; i++ {
+		value := make([]byte, i)
+		rand.Read(value)
+		k := sha3.NewLegacyKeccak256()
+		k.Write(value)
+		k.Sum(expected[:0])
```
AlexHentschel, line 216: The documentation of `Sum`:
> Sum appends the current hash to b and returns the resulting slice.
> It does not change the underlying hash state.

To me, it looks like you are pre-allocating the slice based on your knowledge of how long you are expecting the result to be. To me, this is a bit confusing. This is a test and therefore probably not performance critical. I would suggest favouring readability over runtime speed:

```suggestion
		expectedHash := k.Sum(nil)
```

crypto/hash/keccak.go
```diff
@@ -1,14 +1,203 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+const (
+	// maxRate is the maximum size of the internal buffer. SHA3-256
+	// currently needs the largest buffer among supported sponge-based
+	// algorithms.
+	maxRate = rateSHA3_256
+)
+
+type spongeState struct {
```
AlexHentschel, line 73: my understanding is that `spongeState` is shared between different hashing algorithms (?) If that is the case, I would suggest putting it in its own file (e.g. `spongestate.go`). 

Maybe I am confused because `Keccak` is an entire family, to which sha3 also belongs?

crypto/hash/keccak.go
```diff
@@ -1,14 +1,203 @@
...
+}
+
+// setBuf assigns `buf` (sub-slice of `storage`) to a sub-slice of `storage`
+// defined by a starting index and size.
+func (d *spongeState) setBuf(start, size int) {
+	d.bufIndex = start
+	d.bufSize = size
+}
+
+const bufNilValue = -1
```
AlexHentschel, line 111: can we collect all the const values in one statement?

AlexHentschel (approved):
Looks good. Don't really have the crypto background to verify that the actual algorithm is correct. But I am sure you have it covered 😄

crypto/hash/keccak.go
tarakby: Yes, Keccak is the core base of the entire family. From that family, the crypto lib implements SHA3-256, SHA3-384, and Legacy Keccak-256. The sponge state is a common structure for all the Keccak-based family. That's why it is defined in the common file `keccak.go`.

tarakby: Calling the file `sponge.go` or `keccak.go` is basically equivalent. Sha3 uses Keccak (and therefore the sponge), and same for Legacy-Keccak, it uses Keccak (and the sponge).

tarakby:
bors merge

PR #1868, opened 2022-01-19
[FVM] Support legacy Keccak 256 hashing algorithm

ramtinms:
This PR adds Keccak256 hashing support to the FVM, so can be used for ECDSA account signature verification, and also would be available to the Cadence Runtime which would unlock many applications such as on-chain proof verification.


fvm/fvm_test.go
```diff
@@ -1948,6 +1948,25 @@ func TestHashing(t *testing.T) {
...
+			},
+		},
+		{
+			Algo:    runtime.HashAlgorithmKECCAK_256,
+			WithTag: true,
+			Tag:     "some_tag",
+			Check: func(t *testing.T, result string, scriptErr errors.Error, executionErr error) {
+				require.NoError(t, scriptErr)
+				require.NoError(t, executionErr)
+				require.Equal(t, "8454ec77f76b229a473770c91e3ea6e7e852416d747805215d15d53bdc56ce5f", result)
```
ramtinms, line 1967: I was not 100% sure how to compute Keccak 256 reference hash with some tags, @tarakby could you please check the correctness of this reference.

fvm/crypto/crypto.go
```diff
@@ -135,7 +139,7 @@ func VerifySignatureFromRuntime(
 	// check ECDSA compatibilites
 	if sigAlgo == crypto.ECDSAP256 || sigAlgo == crypto.ECDSASecp256k1 {
 		// hashing compatibility
-		if hashAlgo != hash.SHA2_256 && hashAlgo != hash.SHA3_256 {
+		if hashAlgo != hash.SHA2_256 && hashAlgo != hash.SHA3_256 && hashAlgo != hash.Keccak_256 {
```
ramtinms, line 142: is this something acceptable for ECDSA signatures? I guessed there might be many ethereum signatures that can be verified by this

tarakby: We need to distinguish two different types of ECDSA signatures in the FVM:
 - the transaction signatures : for these, only SHA2-256 and SHA3-256 are supported for now. Keccak-256 can be added as well, though this will need changes in the account format in many places in our code/ecosystem. I think we can avoid that for now.
 - the contract signatures: signatures to be verified on chain by Cadence contracts. There is no problem for these signatures to support Keccak-256.  
 
The issue is that the FVM uses the same code to verify both these signatures above. I think it would be great to separate these 2 verifications as they have different requirements (a new issue to be created), and then we can support Keccak-256 for the second type. For now I suggest not to add Keccak-256

ramtinms: I just update the code at d56bb0b to limit the transaction signatures hashing to specific ones, let me know if this addresses your concern, otherwise I can drop this for now.

fvm/fvm_test.go
tarakby: You should be able to use this online tool: https://emn178.github.io/online-tools/keccak_256.html

fvm/transactionVerifier.go
```diff
@@ -184,6 +185,14 @@ func (v *TransactionSignatureVerifier) verifyAccountSignature(
 		return nil, errors.NewInvalidPayloadSignatureError(txSig.Address, txSig.KeyIndex, err)
 	}
 
+	if !v.accountSignatureSchemeIsSupported(accountKey) {
```
tarakby, line 188: So I looked further at the code and I noticed that my previous comment wasn't totally valid: when we are in the transaction signature case, we fetch the account public key from the state, that key is guaranteed to have the correct hashing algorithm, otherwise it wouldn't have been added to the state. You can see the check done [here](https://github.com/onflow/flow-go/blob/master/fvm/state/accounts.go#L274-L282). 
This means that we do not need to add this extra check and the original code is fine.

fvm/crypto/crypto.go
tarakby: My previous comment confirms that this current file is fine. We can revert the commit https://github.com/onflow/flow-go/commit/d56bb0b58b73857e1d7eeac0066cf7d48b0759d1, my bad!

tarakby (commented):
The code looks good to me. 

There are just a few missing test cases where we need to add Keccak:
 - https://github.com/onflow/flow-go/blob/master/fvm/crypto/crypto_test.go#L60
 - https://github.com/onflow/flow-go/blob/master/fvm/crypto/crypto_test.go#L64
 - https://github.com/onflow/flow-go/blob/master/fvm/crypto/crypto_test.go#L220
 - https://github.com/onflow/flow-go/blob/master/fvm/fvm_test.go#L1548

fvm/fvm_test.go
ramtinms: Yeah but how could you pass the tag, is it just prepending the tag to the data?

fvm/transactionVerifier.go
ramtinms: great, going to revert it then

fvm/fvm_test.go
tarakby: Yes, for Keccak it is just prepending the message with the tag

janezpodhostnik (approved):

fvm/fvm_test.go
```diff
@@ -1558,6 +1558,12 @@ func TestSignatureVerification(t *testing.T) {
 				return hash.NewSHA2_256()
 			},
 		},
+		{
+			"KECCAK_256",
+			func() hash.Hasher {
+				return hash.NewKeccak_256()
+			},
```
tarakby, line 1565: This could be written as:
```suggestion
			"KECCAK_256",
			hash.NewKeccak_256,
```
and same for lines 1551 and 1557 (even though they are not part of this PR)

tarakby (approved):
👌🏼

go.mod
```diff
@@ -56,7 +56,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.4.1
 	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/onflow/atree v0.1.1
-	github.com/onflow/cadence v0.20.3
+	github.com/onflow/cadence v0.20.4-0.20220124195858-d9317cf57e25
```
ramtinms, line 59: this has to be updated to a released on, it has been added so tests can run earlier

fvm/fvm_test.go
ramtinms: updated

ramtinms:
bors merge

PR #1869, opened 2022-01-19
[Execution] Improve transaction execution log

zhangchiqing:
This PR updates the logs for tracing transaction execution, such that:
- We log one entry when start executing a tx, and log another entry after executing the tx with the result indicating succeeded or not. It did not increase the number of logs, since it was logging twice before after executing a tx.
- Ensure tx ID and block height are included in the log (including failed tx), so that it's easier to fetch them by the next executed block height for debugging purpose.
- Indicate whether a tx is a system chunk tx.

The log before the change for a tx:
```
{"level":"info","node_role":"execution","node_id":"b462c11e771b1ce51b808d3efdc57a75f9a1785213f0d8fdabe297423607bd66","engine":"computation","component":"block_computer","txHash":"33f84536f30f37c3df1f5ce9b5fbd7ec526b43852e3218790ca61f14b0769be1","blockHeight":1261,"ledgerInteractionUsed":655244,"retried":0,"time":"2022-01-19T21:33:22Z","message":"transaction executed successfully"}

{"level":"info","node_role":"execution","node_id":"b462c11e771b1ce51b808d3efdc57a75f9a1785213f0d8fdabe297423607bd66","engine":"computation","component":"block_computer","txHash":"33f84536f30f37c3df1f5ce9b5fbd7ec526b43852e3218790ca61f14b0769be1","traceID":"","timeSpentInMS":2845,"time":"2022-01-19T21:33:22Z","message":"transaction executed"}
```

After the change:
```
{"level":"info","node_role":"execution","node_id":"b462c11e771b1ce51b808d3efdc57a75f9a1785213f0d8fdabe297423607bd66","engine":"computation","component":"block_computer","tx_id":"33f84536f30f37c3df1f5ce9b5fbd7ec526b43852e3218790ca61f14b0769be1","tx_index":0,"block_id":"7f021f157b637c1404e234f5c1ea9f5d70f905abcbc316ae742dc0400eaee630","height":93,"system_chunk":true,"time":"2022-01-19T20:59:47Z","message":"executing transaction in fvm"}

{"level":"info","node_role":"execution","node_id":"b462c11e771b1ce51b808d3efdc57a75f9a1785213f0d8fdabe297423607bd66","engine":"computation","component":"block_computer","tx_id":"33f84536f30f37c3df1f5ce9b5fbd7ec526b43852e3218790ca61f14b0769be1","block_id":"7f021f157b637c1404e234f5c1ea9f5d70f905abcbc316ae742dc0400eaee630","traceID":"","computation_used":13,"timeSpentInMS":6,"time":"2022-01-19T20:59:47Z","message":"transaction executed successfully"}
```

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -247,7 +247,7 @@ func (e *blockComputer) executeSystemCollection(
 		return txIndex, fmt.Errorf("could not get system chunk transaction: %w", err)
 	}
 
-	err = e.executeTransaction(tx, colSpan, collectionView, programs, systemChunkCtx, collectionIndex, txIndex, res)
+	err = e.executeTransaction(tx, colSpan, collectionView, programs, systemChunkCtx, collectionIndex, txIndex, res, true)
 	txIndex++
 
 	if err != nil {
@@ -302,7 +302,7 @@ func (e *blockComputer) executeCollection(
 
 	txCtx := fvm.NewContextFromParent(blockCtx, fvm.WithMetricsReporter(e.metrics), fvm.WithTracer(e.tracer))
 	for _, txBody := range collection.Transactions {
-		err := e.executeTransaction(txBody, colSpan, collectionView, programs, txCtx, collectionIndex, txIndex, res)
+		err := e.executeTransaction(txBody, colSpan, collectionView, programs, txCtx, collectionIndex, txIndex, res, false)
 		txIndex++
 		if err != nil {
 			return txIndex, err
@@ -330,6 +330,7 @@ func (e *blockComputer) executeTransaction(
 	collectionIndex int,
 	txIndex uint32,
 	res *execution.ComputationResult,
+	isSystemChunk bool,
 ) error {
 	startedAt := time.Now()
 	txID := txBody.ID()
@@ -351,9 +352,13 @@ func (e *blockComputer) executeTransaction(
 	}
 	defer txInternalSpan.Finish()
 
-	e.log.Debug().
-		Hex("tx_id", logging.Entity(txBody)).
-		Msg("executing transaction")
+	e.log.Info().
+		Str("tx_id", txID.String()).
+		Uint32("tx_index", txIndex).
+		Str("block_id", res.ExecutableBlock.ID().String()).
+		Uint64("height", res.ExecutableBlock.Block.Header.Height).
+		Bool("system_chunk", isSystemChunk).
+		Msg("executing transaction in fvm")
 
 	tx := fvm.Transaction(txBody, txIndex)
 	if isSampled {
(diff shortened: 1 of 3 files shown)


engine/execution/computation/computer/computer.go
```diff
@@ -330,6 +330,7 @@ func (e *blockComputer) executeTransaction(
 	collectionIndex int,
 	txIndex uint32,
 	res *execution.ComputationResult,
+	isSystemChunk bool,
```
Kay-Zee, line 333: This is purely for logging?

engine/execution/computation/computer/computer.go
```diff
@@ -401,19 +400,31 @@ func (e *blockComputer) executeTransaction(
 	// of failed transaction invocation
 	err = collectionView.MergeView(txView)
 	if err != nil {
-		return fmt.Errorf("merging tx view to collection view failed: %w", err)
+		return fmt.Errorf("merging tx view to collection view failed for tx %v: %w",
```
Kay-Zee, line 403: this path would have zero logs, does it make sense to still log something when we fail here?

Kay-Zee (commented):
Nice! liking the clean-up! just one small comment

engine/execution/computation/computer/computer.go
zhangchiqing: We do have log for this error case.

[The error will popup, and eventually logged with other context, including the block id.](https://github.com/onflow/flow-go/blob/0ffe937b9d0077f6fc4889d334d2dedd6a4b834b/engine/execution/ingestion/engine.go#L575)

I don't think it's necessary to log again.

engine/execution/computation/computer/computer.go
zhangchiqing: yes

Kay-Zee (approved):
we may have to consider putting these back onto debug eventually, but for now, this should be fine.

fvm/transactionInvoker.go
```diff
@@ -205,22 +205,14 @@ func (i *TransactionInvoker) Process(
 			// drop delta
 			childState.View().DropDelta()
 			programs.Cleanup(nil)
-			i.logger.Info().
+			i.logger.Warn().
```
janezpodhostnik, line 208: This is an expected error in the case where the funds of the payer approach 0. I am not sure a warn level is appropriate here.

zhangchiqing: Both the error message and comments sound like the error was unexpected.

If the error is expected, is there a specific error type that we can compare and print a info log saying something like "payer has insufficient transaction fee"

I will revert this change for now.

janezpodhostnik (approved):

fvm/transactionInvoker.go
janezpodhostnik: Created a bug report https://github.com/dapperlabs/flow-go/issues/6123

zhangchiqing:
bors merge

fvm/transactionInvoker.go
```diff
@@ -220,14 +220,6 @@ func (i *TransactionInvoker) Process(
 
 			return feesError
 		}
-	} else {
-		// transaction is ok, log as successful
-		i.logger.Info().
-			Str("txHash", txIDStr).
-			Uint64("blockHeight", blockHeight).
-			Uint64("ledgerInteractionUsed", sth.State().InteractionUsed()).
-			Int("retried", proc.Retried).
```
ramtinms, line 229: I saw this PR late, unfortunately removal of ledgerInteractionUsed broke some metrics we were interested in, anyway we can put them back @zhangchiqing

PR #1877, opened 2022-01-21
Disable machine account balance checking on test networks

jordanschalm:
Transient test networks do not have transaction fees enabled, and do not fund machine accounts. This resulted in misleading error logs in these cases. This PR disables the unneeded balance checks when running on a transient test network.


kc1116 (approved):

module/epochs/machine_account.go
```diff
@@ -37,28 +37,58 @@ const (
...
+
+// WithoutBalanceChecks sets minimum balances to 0 to effectively disable minimum
+// balance checks. This is useful for test networks where transaction fees are
+// disabled.
+func WithoutBalanceChecks(conf *MachineAccountValidatorConfig) {
+	conf.SoftMinBalanceLN = 0
+	conf.HardMinBalanceLN = 0
+	conf.SoftMinBalanceSN = 0
+	conf.HardMinBalanceSN = 0
+}
```
zhangchiqing, line 83: maybe expose this method only in tests? Since it's only for local network.

zhangchiqing (approved):
Looks good. One minor suggestion

module/epochs/machine_account.go
jordanschalm: Well, it's for any transient network, so local integration tests, Localnet, and Benchnet. We use the same binary for Localnet/Benchnet as real networks, so I think it needs to be exposed for all of them.

PR #1880, opened 2022-01-21
[Integration] Topology Factory

yhassanzadeh13:
This PR addresses the `ghost` node topology impairment by decoupling the ghost node topology from the real node ones:

Proposed solution is to decouple the topology protocol of `ghost` nodes from other nodes: 
- A `ghost` node is running on a fully-connected topology, meaning it picks all nodes of the test as its fanout, regardless of its role and subscribed channels. This saves the ghost nodes from dealing with constraints governed by the `topic-based` topology regarding the valid roles. 
- All other nodes than `ghost` node are running on the default `topic-based` topology. 

# Context:

At the networking layer of Flow, we use a topology protocol to construct a P2P overlay network graph. Flow nodes use this overlay graph to disseminate epidemic messages through `publish` and `multicast`. 

The default topology of a Flow node is a `topic-based` topology that constructs a separate overlay graph per subscription channel on the pubsub overlay. On the other hand, solely for sake of testing, we have a `ghost` node role that subscribes to every pubsub channel so that it can send and receive message through arbitrary channels. 

This poses an issue with the topology construction:
(1) Some channels are known as cluster-based channels, where only certain roles can be part of, i.e., consensus and collection. 
(2) As part of its mission, a ghost node tries subscribing to _all channels_, including those cluster-based channels. 
(3) When a ghost node is running in any role than consensus or collection, it hits an error on subscribing to cluster-based channels. The reason is a node can only subscribe to the channel of its corresponding cluster. However, only consensus and collection nodes belong to a cluster. Hence, when a `ghost` node is running other than these two roles, it is not part of any cluster, and `topic-based` topology fails to construct its part of overlay graph (i.e., fanout) for cluster-based topics. Fanout generation is an all-or-nothing operation. Hence, if it fails for one channel, 
it fails for all channels of the node. In this situation the `ghost` node remains with no fanout. 

When a node does not have a fanout it cannot send or receive any message through the pubsub protocol. The connectivity of the networking overlay graph may also be damaged affected by zero-fanout nodes. Hence, `ghost` nodes running any role other than consensus or collection node end up with no fanout at all. Running multiple instances of such nodes may even disconnect the entire networking graph with several partitions, hence imparing the message delivery. 

# Evidence: 

Following log entry is taken from a `ghost` node running as an `execution` node. As the log shows, the node fails to generate any fanout for itself due an error while generating a cluster-based fanout subset. 

```
··· DOCK: (clogs ) execution_1               (3d7a5ccfd207cc8c0adc124454aa7b13add0dc2788761de535c0996cda2ebc4b) - �����={"level":"error","node_role":"ghost","node_id":"e416b0622dc4ec439a1c3b2054c643973b6ee3835b62c849d0d4561d3e48b4fc","error":"could not generate topology: could not generate fanout for topic consensus-cluster-cluster-0-1832762ff87fae977450ccbce106f00afee32fadfac40a2f7928ea5cd633dc7d: failed to find cluster peers for node e416b0622dc4ec439a1c3b2054c643973b6ee3835b62c849d0d4561d3e48b4fc: failed to find the cluster for node ID e416b0622dc4ec439a1c3b2054c643973b6ee3835b62c849d0d4561d3e48b4fc, subscribed: 12","time":"2022-01-21T00:46:11Z","message":"failed to update peers"}
```

This error message means that the ghost node is unable to generate fanout topology for itself. The fanout is the set of nodes that a node "talks" over the pub/sub overlay. Since the topology generation hits an error, the ghost node is left with "no fanout", hence cannot _talk_ with other nodes on pubsub and cannot receive messages. Further investigations show that this error _happens all the time even when the tests are passing!_. However, in those cases that the tests are passing, although the ghost node does not have any fanout, it is selected by other nodes of the system as part of their fanout. We found three cases for this situation that lead to different behavior: 
1. Ghost node does not have fanout, but it is selected by other nodes on _every_ channel it is subscribing to.
    - Tests are passing. 
    - This happens rarely in tests that have a super majority of ghost nodes. 
2. Ghost node does not have fanout, and it is not selected by _any_ node on channels it is subscribing to.
    - Happens frequently. 
    - Tests are failing, hence flaky behavior. 
3. Ghost node does not have fanout, and it is not selected by nodes on _some_ channels it is subscribing to.
    - Happens very often. 
    - Ghost node cannot receive messages from those channels it missed subscription.

### DIFF
--- a/integration/testnet/network.go
+++ b/integration/testnet/network.go
@@ -466,7 +466,7 @@ func WithDebugImage(debug bool) func(config *NodeConfig) {
 func AsGhost() func(config *NodeConfig) {
 	return func(config *NodeConfig) {
 		config.Ghost = true
-                 // using the fully-connectred topology to ensure a ghost node is always connected to all other nodes in the network, 
+		// using the fully-connectred topology to ensure a ghost node is always connected to all other nodes in the network,
 		config.AdditionalFlags = append(config.AdditionalFlags, "--topology=fully-connected")
 	}
 }
--- a/integration/testnet/network.go
+++ b/integration/testnet/network.go
@@ -466,6 +466,7 @@ func WithDebugImage(debug bool) func(config *NodeConfig) {
 func AsGhost() func(config *NodeConfig) {
 	return func(config *NodeConfig) {
 		config.Ghost = true
+		config.AdditionalFlags = append(config.AdditionalFlags, "--topology=fully-connected")
 	}
 }
 
--- a/network/topology/factory.go
+++ b/network/topology/factory.go
@@ -1,6 +1,8 @@
 package topology
 
 import (
+	"fmt"
+
 	"github.com/rs/zerolog"
 
 	"github.com/onflow/flow-go/model/flow"
@@ -22,5 +24,7 @@ func Factory(name Name) (FactoryFunction, error) {
 		return RandomizedTopologyFactory(), nil
 	case TopicBased:
 		return TopicBasedTopologyFactory(), nil
+	default:
+		return nil, fmt.Errorf("unknown topology name: %s", name)
 	}
 }
--- a/network/topology/factory.go
+++ b/network/topology/factory.go
@@ -10,6 +10,8 @@ import (
 	"github.com/onflow/flow-go/state/protocol"
 )
 
+// MaximumEdgeProbability defines the maximum value of probability for existing an edge between any arbitrary pair of nodes.
+// Edge probability is utilized selectively in some topology protocols, e.g., RandomizedTopology.
 const MaximumEdgeProbability = float64(1)
 
 type Name string
--- a/network/topology/factory.go
+++ b/network/topology/factory.go
@@ -10,6 +10,8 @@ import (
 	"github.com/onflow/flow-go/state/protocol"
 )
 
+const MaximumEdgeProbability = float64(1)
+
 type Name string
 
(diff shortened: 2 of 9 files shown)


network/topology/factory.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/network"
+	"github.com/onflow/flow-go/state/protocol"
+)
+
+const MaximumEdgeProbability = float64(1)
```
vishalchangrani, line 13: can you throw in a comment on what this signifies?

integration/testnet/network.go
```diff
@@ -466,6 +466,7 @@ func WithDebugImage(debug bool) func(config *NodeConfig) {
 func AsGhost() func(config *NodeConfig) {
 	return func(config *NodeConfig) {
 		config.Ghost = true
+		config.AdditionalFlags = append(config.AdditionalFlags, "--topology=fully-connected")
```
vishalchangrani, line 469:
```suggestion
                 // using the fully-connectred topology to ensure a ghost node is always connected to all other nodes in the network, 
		config.AdditionalFlags = append(config.AdditionalFlags, "--topology=fully-connected")
```

vishalchangrani (approved):
lgtm

zhangchiqing (approved):

PR #1887, opened 2022-01-24
modifying json response to take into account errors that occur when w…

vishalchangrani:
…riting the http response and removing recursive call to error response

PR to address this comment: https://github.com/onflow/flow-go/pull/1809#discussion_r788106557

### DIFF
--- a/engine/access/rest/handler.go
+++ b/engine/access/rest/handler.go
@@ -131,15 +131,16 @@ func (h *Handler) errorHandler(w http.ResponseWriter, err error, errorLogger zer
 // jsonResponse builds a JSON response and send it to the client
 func (h *Handler) jsonResponse(w http.ResponseWriter, code int, response interface{}, errLogger zerolog.Logger) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
-	w.WriteHeader(code)
 
 	// serialize response to JSON and handler errors
 	encodedResponse, err := json.MarshalIndent(response, "", "\t")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


engine/access/rest/handler.go
```diff
@@ -131,20 +131,19 @@ func (h *Handler) errorHandler(w http.ResponseWriter, err error, errorLogger zer
...
 func (h *Handler) jsonResponse(w http.ResponseWriter, code int, response interface{}, errLogger zerolog.Logger) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(code)
 
-	// serialise response to JSON and handler errors
+	// serialize response to JSON and handler errors
 	encodedResponse, err := json.MarshalIndent(response, "", "\t")
 	if err != nil {
-		h.errorHandler(w, err, errLogger)
+		errLogger.Error().Err(err).Str("response", string(encodedResponse)).Msg("failed to indent response")
```
peterargue, line 139: we should update the code here, and after the other error to 500

vishalchangrani: so that _may_ not work since the HTTP response code has already been written out at this point.

vishalchangrani: So calling `Write` ends up also calling `WriteHeader` (https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/net/http/server.go;l=120-121). Hence, if the header is not set before the call to write, then it returns 200 OK.

peterargue (approved):

PR #1888, opened 2022-01-24
[Networking] Refactors DNS cache with HeroCache

yhassanzadeh13:
This PR implements an alternative `DNSCache` mempool based on `HeroCache`. It then removes the existing map-based DNS cache and reactors the code to utilize the `DNSCache`.  

Note: `HeroCache` is our homebrewed cache which is around 4 times faster than `stdmap` version and incures `O(1)` heap objects.


module/mempool/herocache/dns_cache.go
```diff
@@ -0,0 +1,143 @@
...
+	return &DNSCache{
+		txtCache: stdmap.NewBackendWithBackData(
+			herocache.NewCache(sizeLimit, 8, heropool.LRUEjection, logger)),
+		ipCache: stdmap.NewBackendWithBackData(
+			herocache.NewCache(sizeLimit, 8, heropool.LRUEjection, logger)),
+	}
+}
+
+// PutIpDomain adds the given ip domain into the cache.
+func (d *DNSCache) PutIpDomain(domain string, timestamp int64, addresses []net.IPAddr) bool {
```
simonhf, line 29: Any reason why this is called `PutIpDomain()` and not `PutDomainIp()`?
Also the argument order is 1st domain, 2nd timestamp, 3rd addresses.
If the function does end up getting renamed to `PutDomainIp()`, why not use the argument order 1st domain, 2nd addresses, 3rd timestamp... to sort of match the name?

module/mempool/herocache/dns_cache.go
```diff
@@ -0,0 +1,143 @@
...
+		domain:    domain,
+		addresses: addresses,
+		timestamp: timestamp,
+	}
+
+	return d.ipCache.Add(i)
+}
+
+// PutTxtDomain adds the given txt domain into the cache.
+func (d *DNSCache) PutTxtDomain(domain string, timestamp int64, addresses []string) bool {
```
simonhf, line 41: Consider renaming (potentially misleading) `addresses` to `TxtRecords` or `TextRecords` as per [1].

[1] https://en.wikipedia.org/wiki/TXT_record

simonhf (approved):
LGTM except for a couple of renaming suggestions :-)

module/mempool/herocache/dns_cache.go
```diff
@@ -0,0 +1,143 @@
...
+			herocache.NewCache(sizeLimit, 8, heropool.LRUEjection, logger)),
+		ipCache: stdmap.NewBackendWithBackData(
+			herocache.NewCache(sizeLimit, 8, heropool.LRUEjection, logger)),
+	}
+}
+
+// PutIpDomain adds the given ip domain into the cache.
+func (d *DNSCache) PutIpDomain(domain string, timestamp int64, addresses []net.IPAddr) bool {
+	i := ipEntity{
+		id:        domainToIdentifier(domain),
```
jwinkler2083233, line 31: This is using 'MakeID'.
That seems like overkill.  What does it gain for you, since the lookups have to use MakeID(), too ??

jwinkler2083233 (changes requested):

module/mempool/herocache/dns_cache.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/1888/commits/a0728490ee1136453bf6e1102f2db006d8d73e6c

jwinkler2083233 (approved):

module/mempool/herocache/dns_cache.go
yhassanzadeh13: Hashing our offline conversations around this comment:

> At this stage, we needed a collision-resistant identifier for DNS records. We want to make sure that no two distinct domains collide upon storing in the cache. Otherwise, it would cause safety concerns on the code that DNS queries for the domain `A` are resolved to domain `B` only because `A` and `B` collide on the same identifier in the cache. We examined several options but reached the conclusion that we should not compromise on the collision-resistance feature, since it enables DNS poisoning attacks. Hence we confined utilization of hash literally once per record upon inserting it into the cache. The `Rem` method is barely used only when we face an error in resolving a domain, which is an extremely unlikely event in the current configurations. Also, this translation from `string -> flow.Identifier ` is done only once when we insert a domain into the cache.

PR #1893, opened 2022-01-25
1651/ Dynamic Node Startup

kc1116:
This PR adds a pre-init function that will allow users to specify a target epoch and phase to wait for before initializing their nodes. This allows node operators to start their node any time before it is accepted as a network participant, and the node will retrieve and bootstrap an appropriate root snapshot, and begin normal operation once it is able. 

Nodes which join at epoch N will stake during epoch N-1, and must start up during the `EpochSetup` phase of epoch N-1. 

The typical usage of this feature is:
* Node operator stakes (staking phase of epoch N-1)
* Node operators starts their node (with default configs `epoch=current phase=EpochSetup`) (staking phase of epoch N-1)
* The node will wait until the network enters the setup phase of epoch N-1, then retrieve a snapshot, bootstrap the state, and begins following the state
* If necessary, the node will participate in QC voting or the DKG during the setup phase of epoch N-1
* The node begins regular operation as a fully qualified network participant once epoch N begins


cmd/dynamic_startup.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/onflow/flow-go/cmd/util/cmd/common"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/state/protocol/inmem"
+	"github.com/rs/zerolog"
+)
+
+const sleepInterval = time.Minute
+
+func DynamicStartup(logger zerolog.Logger, initANAddress, initANPubkey string, startupEpoch uint64, startupEpochPhase flow.EpochPhase, rootSnapshotOutDir string) error {
```
peterargue, line 21: can you add a comment about what this is doing?

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,106 @@
...
+
+	ctx := context.Background()
+	start := time.Now()
+	var snapshot *inmem.Snapshot
+
+	// wait until we reach the desired Epoch and EpochPhase
+	for {
+		b, err := flowClient.GetLatestProtocolStateSnapshot(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get latest finalized protocol state snapshot during pre-initialization: %w", err)
```
peterargue, line 42: are any of these errors transient/might recover without a restart?

peterargue (commented):
looks good. added a few small comments. Any unit tests we could add?

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get the current epoch counter: %w", err)
+		}
+
+		currEpochPhase, err := snapshot.Phase()
+		if err != nil {
+			return fmt.Errorf("failed to get the current epoch phase: %w", err)
+		}
+
+		if currEpochCounter == startupEpoch && currEpochPhase == startupEpochPhase {
```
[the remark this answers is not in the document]
jordanschalm: What happens if the node operators uses Dynamic Join, their node starts successfully at the appropriate time, but later on the node restarts? Currently I believe the node will perpetually fail this check and be unable to restart. 

In general, Dynamic Join should only successfully run once and should have as little footprint as possible on subsequent restarts, or when it is disabled. A couple of ideas for this:

1. We could also treat the existence of a `root-protocol-state-snapshot.json` file as an indication that Dynamic-Join has completed successfully. If the file exists, then we can use it to start up. 
2. We do initialize the Badger database and storage before the `PreInit` function ([code](https://github.com/onflow/flow-go/blob/d1e29bf8d1a90182748768da9063b889af85dbcc/cmd/scaffold.go#L1047-L1053)), perhaps we can add a persistent flag indicating whether Dynamic-Join has completed successfully, and skip it altogether if this flag is set. 

I'm favouring option 1.

cmd/collection/main.go
```diff
@@ -147,6 +155,14 @@ func main() {
...
+
+		// dynamic node startup flags
+		flags.BoolVar(&dynamicStartup, "dynamic-startup", false, "when set to true the node will attempt to wait for a specified epoch counter & phase before initializing")
+		flags.StringVar(&initANPubkey, "init-access-publickey", "", "the public key of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&initANAddress, "init-access-address", "", "the access address of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&startupEpochPhase, "startup-epoch-phase", "EpochPhaseSetup", "the target epoch phase in integer form for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
+		flags.IntVar(&startupEpoch, "startup-epoch", 0, "the epoch the node will wait for before initializing modules when using dynamic-startup")
```
jordanschalm, line 164: I find it helpful to prefix flag names with the component they are for, especially when there are several related flags.  They show up together in an ordered list of flags (eg. in documentation or CLI help message) and it's easy to see at a glance that a group of flags are configuring the same thing.

```suggestion
		flags.BoolVar(&dynamicStartup, "dynamic-startup", false, "when set to true the node will attempt to wait for a specified epoch counter & phase before initializing")
		flags.StringVar(&initANPubkey, "dynamic-join-access-publickey", "", "the public key of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
		flags.StringVar(&initANAddress, "dynamic-join-access-address", "", "the access address of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
		flags.StringVar(&startupEpochPhase, "dynamic-join-epoch-phase", "EpochPhaseSetup", "the target epoch phase in integer form for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
		flags.IntVar(&startupEpoch, "dynamic-join-epoch-counter", 0, "the epoch the node will wait for before initializing modules when using dynamic-startup")
```

cmd/collection/main.go
```diff
@@ -147,6 +155,14 @@ func main() {
 		// epoch qc contract flags
 		flags.BoolVar(&insecureAccessAPI, "insecure-access-api", false, "required if insecure GRPC connection should be used")
 		flags.StringSliceVar(&accessNodeIDS, "access-node-ids", []string{}, fmt.Sprintf("array of access node IDs sorted in priority order where the first ID in this array will get the first connection attempt and each subsequent ID after serves as a fallback. Minimum length %d. Use '*' for all IDs in protocol state.", common.DefaultAccessNodeIDSMinimum))
+
+		// dynamic node startup flags
+		flags.BoolVar(&dynamicStartup, "dynamic-startup", false, "when set to true the node will attempt to wait for a specified epoch counter & phase before initializing")
```
jordanschalm, line 160: I actually think we can safely remove this flag in tandem with some other suggested changes. 

In the PreInit, we first check for a root snapshot. 
* If there is no snapshot, and `dynamic-join-access-pubkey` and `dynamic-join-access-address` are set, we can initiate Dynamic-Join. 
* If there is a snapshot, we skip Dynamic-Join start up using the snapshot.

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,107 @@
...
+	"time"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/cmd/util/cmd/common"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/state/protocol/inmem"
+)
+
+const sleepInterval = time.Minute
```
jordanschalm, line 19: We could make this a flag as well, though I think most people would be fine with a 1-minute default.

jordanschalm (commented):
Added some suggestions for tailoring the configuration more to the typical case.

kc1116:
> looks good. added a few small comments. Any unit tests we could add?

@peterargue do you have any ideas on how you would unittest this func? One option I am leaning towards because we can't mock the flow client , I will add the logic that fetches the root snapshot as a func argument then in unittest we can mock this argument. What do you think about that?

peterargue:
> we can't mock the flow client
is that a limitation of the client? Could we do the setup in the `PreInit` closure and pass the client into `DynamicStartup`?

If not, that approach sounds reasonable.

cmd/collection/main.go
```diff
@@ -147,6 +154,14 @@ func main() {
...
+		flags.StringVar(&initANPubkey, "init-access-publickey", "", "the public key of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&initANAddress, "init-access-address", "", "the access address of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&startupEpochPhase, "startup-epoch-phase", "EpochPhaseSetup", "the target epoch phase in integer form for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/dynamic_startup.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/collection/main.go
```diff
@@ -147,6 +155,14 @@ func main() {
...
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/collection/main.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/collection/main.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/dynamic_startup.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/collection/main.go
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/8ea656ac507e9969c635a600e908891394d7ac2c

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,153 @@
...
+			return fmt.Errorf("failed to get the current epoch counter: %w", err)
+		}
+
+		currEpochPhase, err := snapshot.Phase()
+		if err != nil {
+			return fmt.Errorf("failed to get the current epoch phase: %w", err)
+		}
+
+		// check if we are in or past the target epoch and phase
+		if currEpochCounter >= uint64(startupEpoch) && currEpochPhase >= startupEpochPhase {
```
jordanschalm, line 84: If we configure the node to startup at `counter=5,phase=setup`, and we observe a snapshot with `counter=6,phase=staking`, this will evaluate to false. But we would like to start up in this case, because we have passed the target startup time. (Sorry, I oversimplified this condition in a previous comment.)

We would need to change the condition here to:
```
currEpochCounter > startupEpoch || (currEpochCounter == startupEpoch && currEpochPhase >= startupEpochPhase)
```

But, changing this condition has a side effect on the desired property that operators taking the standard path should need to change configuration as little as possible (ideally not at all). These operators joining at epoch N will stake in the staking phase of epoch N-1, then start their node anytime afterward (typically during the staking phase of epoch N-1). With the above change, these operators would need to manually specify their startup epoch (since the default is 0). 

To address this, I suggest we change the default value for `startupEpoch` from `0` to something differentiable from a user input (since 0 is also a valid value to specify manually). If `startupEpoch` is unspecified, we will set it to the current epoch value of the first snapshot we observe. That gives the desired behaviour for the standard usage without needing to manually set any extra flags (to join the network at epoch N, we begin waiting in epoch N-1, and we would like to start up in the setup phase of epoch N-1).

For example, we could add a sentinel value for the flag like `--dynamic-startup-epoch-counter=current` which would have this behaviour of using the epoch counter for the first snapshot we see, and set this to be the default.

kc1116: This `--dynamic-startup-epoch-counter=current` would be nice but there is no way to support this without adding an addtional string var flag, or turning the `--dynamic-startup-epoch-counter` flag into a string and doing some uint64 parsing which is not needed. I like the idea of using 0 to mean "current" epoch . Update the flag usage to reflect this 

`use 0 to default to starting node in the current epoch`

kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/collection/main.go
```diff
@@ -147,6 +158,14 @@ func main() {
...
 		flags.StringSliceVar(&accessNodeIDS, "access-node-ids", []string{}, fmt.Sprintf("array of access node IDs sorted in priority order where the first ID in this array will get the first connection attempt and each subsequent ID after serves as a fallback. Minimum length %d. Use '*' for all IDs in protocol state.", common.DefaultAccessNodeIDSMinimum))
+
+		// dynamic node startup flags
+		flags.StringVar(&dynamicStartupANPubkey, "dynamic-startup-access-publickey", "", "the public key of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&dynamicStartupANAddress, "dynamic-startup-access-address", "", "the access address of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&dynamicStartupEpochPhase, "dynamic-startup-startup-epoch-phase", "EpochPhaseSetup", "the target epoch phase for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
+		flags.IntVar(&dynamicStartupEpoch, "dynamic-startup-startup-epoch", 0, "the epoch the node will wait for before initializing modules when using dynamic-startup")
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,153 @@
...
+			logger.Info().
+				Str("time-waiting", time.Since(start).String()).
+				Str("current-epoch", fmt.Sprintf("%d", currEpochCounter)).
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,153 @@
...
+				Msg("reached desired epoch and phase in dynamic startup pre-init")
+
+			return nil
+		}
+
+		logger.Warn().
+			Str("time-waiting", time.Since(start).String()).
+			Str("current-epoch", fmt.Sprintf("%d", currEpochCounter)).
+			Str("current-epoch-phase", currEpochPhase.String()).
+			Msg(fmt.Sprintf("waiting for epoch %d and phase %s", startupEpoch, startupEpochPhase.String()))
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,153 @@
...
+	config, err := common.NewFlowClientConfig(anAddress, anPub, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create flow client config for node dynamic startup pre-init: %w", err)
+	}
+
+	flowClient, err := common.FlowClient(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create flow client for node dynamic startup pre-init: %w", err)
+	}
+
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/collection/main.go
```diff
@@ -147,6 +158,14 @@ func main() {
 		// epoch qc contract flags
 		flags.BoolVar(&insecureAccessAPI, "insecure-access-api", false, "required if insecure GRPC connection should be used")
 		flags.StringSliceVar(&accessNodeIDS, "access-node-ids", []string{}, fmt.Sprintf("array of access node IDs sorted in priority order where the first ID in this array will get the first connection attempt and each subsequent ID after serves as a fallback. Minimum length %d. Use '*' for all IDs in protocol state.", common.DefaultAccessNodeIDSMinimum))
+
+		// dynamic node startup flags
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/collection/main.go
```diff
@@ -147,6 +158,14 @@ func main() {
...
+		flags.StringVar(&dynamicStartupANPubkey, "dynamic-startup-access-publickey", "", "the public key of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&dynamicStartupANAddress, "dynamic-startup-access-address", "", "the access address of the trusted secure access node to connect to when using dynamic-startup, this access node must be staked")
+		flags.StringVar(&dynamicStartupEpochPhase, "dynamic-startup-startup-epoch-phase", "EpochPhaseSetup", "the target epoch phase for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
+		flags.IntVar(&dynamicStartupEpoch, "dynamic-startup-startup-epoch", 0, "the epoch the node will wait for before initializing modules when using dynamic-startup")
+		flags.DurationVar(&dynamicStartupSleepInterval, "dynamic-startup-sleep-interval", time.Minute, "the interval in milliseconds in which the node will check if it can start")
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/62ca33e40cb6e5a2182429d6f647dd7eaca0562d

cmd/access/main.go
```diff
@@ -24,6 +25,8 @@ func main() {
 		anb.Logger.Fatal().Err(err).Send()
 	}
 
+	nodeBuilder.PreInit(cmd.DynamicStartPreInit)
```
jordanschalm, line 28: It's a bit confusing to have `PreInit` _after_ `Initialize`, but this seems like correct usage.

cmd/dynamic_startup.go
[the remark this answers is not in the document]
jordanschalm: I don't think this was changed actually.

cmd/dynamic_startup.go
```diff
@@ -130,20 +123,70 @@ func ValidateDynamicStartupFlags(accessPublicKey, accessAddress string, startPha
 		return fmt.Errorf("invalid flag --dynamic-startup-startup-epoch-phase unknow epoch phase")
 	}
 
-	if startEpoch <= 0 {
-		return fmt.Errorf("invalid flag --dynamic-startup-startup-epoch must be integer > 0")
+	return nil
+}
+
+// DynamicStartPreInit will attempt to save a the latest finalizes protocol snapshot
+// if we have reached the target epoch and phase are
```
jordanschalm, line 130: We can probably workshop this a bit 😛

cmd/dynamic_startup_test.go
```diff
@@ -159,50 +153,4 @@ func TestGetSnapshotAtEpochAndPhase(t *testing.T) {
 		require.Equalf(t, 1, hookInfoCalls, "expected 1 info log got %d", hookInfoCalls)
 		require.Equal(t, expectedSnapshot, actualSnapshot)
 	})
-	t.Run("should wait for target phase if target epoch has passed but not in target phase", func(t *testing.T) {
```
jordanschalm: It would be good to test the typical case. If a node is joining at epoch 5 and they run with the default configs (epoch_counter=0, epoch_phase=setup), they should not start up during staking phase of epoch 5, but should start up during the setup phase of epoch 5.

cmd/dynamic_startup.go
jordanschalm: That is a good idea. I'm just worried about one case which will not have the expected behaviour:

I want my node to start up at epoch 0, setup phase. I configure my node with `--dynamic-startup-epoch-counter=0 --dynamic-startup-epoch-phase=EpochPhaseSetup`. 

When I start the process, the network is at epoch 1, staking phase. Because my configuration of the startup epoch counter being set to 0 is interpreted as "current" epoch rather than "epoch with counter 0", my node now won't start up until epoch 1, setup phase (it should start immediately).

This wouldn't be a practical problem on Mainnet, since we're well past epoch 0 and we won't be going back. So maybe it is OK. It would complicate testing a bit though. 

Another thing we could do which would address this edge case is to incorporate an identity table check into the snapshot polling. So, if the default configs were used (epoch=0, phase=setup), and we observe a snapshot where our node ID appears, then also consider that as sufficient reason to start up.

cmd/dynamic_startup_test.go
kc1116: Should a node start up immediately if the target epoch has passed ?

cmd/dynamic_startup.go
kc1116: With the current implementation your node should start immediately if you 
1. specify an epoch in the past 
2. if you don't specify anything (defaults to 0) 

So during tests if the network is at epoch 1 staking phase and you configure your node to start up with epoch 0 setup phase, your node should start immediately because that epoch has past.

cmd/dynamic_startup_test.go
jordanschalm: Yes, definitely. It should always start up as soon as possible after it is able to.

jordanschalm (commented):
Adding a couple more comments around logging and time bounds on requests.

cmd/dynamic_startup_test.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/a24cc2a5241a68045645e4d1203580e1f759b6ae

cmd/dynamic_startup.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/a24cc2a5241a68045645e4d1203580e1f759b6ae

cmd/access/main.go
kc1116: Changed it up a bit https://github.com/onflow/flow-go/pull/1893/commits/a24cc2a5241a68045645e4d1203580e1f759b6ae

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,196 @@
...
+	start := time.Now()
+	var snapshot protocol.Snapshot
+
+	constRetry, err := retry.NewConstant(retryInterval)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create retry mechanism: %w", err)
+	}
+
+	err = retry.Do(context.Background(), constRetry, func(ctx context.Context) error {
+		snapshot, err = getSnapshot()
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/54b265ce0911c65f61bb7abaa6550a66e5ee6b4f

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// DynamicStartPreInit will attempt to save a the latest finalizes protocol snapshot
+// if we have reached the target epoch and phase are
+func DynamicStartPreInit(nodeConfig *NodeConfig) error {
+	rootSnapshotPath := filepath.Join(nodeConfig.BootstrapDir, bootstrap.PathRootProtocolStateSnapshot)
+
+	// root snapshot exists skip dynamic start up
+	if rootSnapshotExists(rootSnapshotPath) {
+		return nil
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/f4ce007e151811653c7822d996e4ccb0765c5f32

cmd/access/node_builder/unstaked_access_node_builder.go
```diff
@@ -126,6 +126,7 @@ func (anb *UnstakedAccessNodeBuilder) Initialize() error {
 	anb.enqueueConnectWithStakedAN()
 
 	anb.PreInit(anb.initUnstakedLocal())
+	anb.PreInit(cmd.DynamicStartPreInit)
```
peterargue, line 129: This is synchronizing startup for newly staked nodes right? This is an unstaked node, so I don't think it's needed here.

cmd/dynamic_startup.go
```diff
@@ -0,0 +1,247 @@
...
+	log := nodeConfig.Logger.With().Str("component", "dynamic-startup").Logger()
+
+	rootSnapshotPath := filepath.Join(nodeConfig.BootstrapDir, bootstrap.PathRootProtocolStateSnapshot)
+
+	// root snapshot exists skip dynamic start up
+	if rootSnapshotExists(rootSnapshotPath) {
+		return nil
+	}
+
+	// get flow clinet with secure client connection to download protocol snapshot from access node
```
peterargue, line 149:
```suggestion
	// get flow client with secure client connection to download protocol snapshot from access node
```

peterargue (commented):
looks great. added a couple small comments

cmd/access/node_builder/unstaked_access_node_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1893/commits/ce2914b1954138e89a40c9b84d93e56014876c00

jordanschalm (approved):

jordanschalm:
Should be ready for another pass @peterargue

peterargue (approved):
looks great!

yhassanzadeh13 (approved):

kc1116:
bors merge

jordanschalm:
bors merge

jordanschalm:
bors merge

Issue #1884, opened 2022-01-23
[Execution State] EN startup was sometimes delayed because older checkpoint file was loaded, causing 40 extra WAL segments to replay


fxamacker:
## 🐞 Bug Report

Intermittently, EN used an older-than-expected checkpoint file during startup which caused 40 extra WAL segments to be replayed, causing a long delay before execution.

The perception was 16-20 minute startup times (for this activity) and fixing this bug can bring that down to 9-10 minutes.

On Jan 14, EN startup duration included:

- &nbsp; 5m20s - to load checkpoint.1912
- 10m34s - to replay 67 WAL segments (because old checkpoint file was loaded instead of latest)
\----------
**15m54s**

Expected:

- 5m20s (approx) - to load checkpoint.1953
- 4m15s (instead of 10m34s) - to replay 40 fewer WAL segments after loading checkpoint.1953.
\---------
**9m35s** (approx)

Logs show creation of checkpoint.1953 began on Jan 13 but it was not used on Jan 14.

Recent logs indicate the loading of older-than-expected checkpoint file happened on Dec 22, Jan 7, and Jan 14.

## Additional context

See https://github.com/dapperlabs/flow-go/issues/6114.

I think some possible non-startup causes can include scenarios like:
- latest checkpoint file was deleted before startup begins looking for latest checkpoint file, or
- latest checkpoint file didn't successfully finish being written prior to startup

fxamacker:
If the checkpoint files were not deleted, there's overlap in some upfront tasks for this issue and #1750.  On Friday, Jan 21, I [suggested](https://axiomzen.slack.com/archives/C015G65HR2P/p1642798138150700?thread_ts=1642181542.092500&cid=C015G65HR2P) some ways to reduce number of WAL segments and some near-term optimizations to checkpoint creation that can allow us to further reduce WAL segments by allowing more frequent checkpoint creation.

fxamacker:
Root cause of this problem is checkpoint creation taking 12-15+ hours, which can accumulate enough WAL segments to trigger another checkpoint creation immediately after the current checkpoint finishes.

Given the long duration and constant checkpoint creation, shutdowns would likely interrupt checkpoint creation which causes more than 40 WAL segments to accumulate and they need to be replayed during EN startup.

Additionally, there wasn't sufficient logging so the root cause wasn't as easy to identify as it should've been.

This is closed by PR #1944 #2126 #2165 #2119 #2187

Issue #1750, opened 2021-12-09
[Execution state] - Checkpointer and Flattener speed and memory improvements 


ramtinms:
currently the checkpoint and trie flattener reconstructs the nodes as the storable version which would duplicate the memory usage when checkpointing, by reusing the same structure the speed and memory usage of the checkpointer can be improved a lot.

Ullaakut:
@ramtinms Currently, checkpoints contain clear text ledger payloads, which seems very inefficient to me both in terms of encoding/decoding speed and storage space usage to a lesser extent.

Using a zstandard compression dictionary, for example, could both improve the speed (by having less data to read/write) of processing checkpoints and reduce their size.

This could be done without breaking changes by still supporting the uncompressed format when reading old checkpoints, of course. WDYT?

bluesign:
Payload compression can be further extended by custom zstd dictionary for each cadence type (I think limit there is 2^32 dictionaries) and uncompress on first access ( or better LRU cache )

fxamacker:
## Ideas for Improving Speed and Memory

@ramtinms These are just some ideas that came up while getting familiar with mtrie code in flow-go.  Thoughts?

### Use stream encoding

The current implementation builds the entire nodes in memory in `FlattenForest()` before writing to checkpoint file.  Because `NodeIterator` returns nodes in the sequence of "DESCENDANTS-FIRST-RELATIONSHIP", I think we can stream encode each  node returned by the iterator and filtered for uniqueness.  Conversion from `Node` to `StorableNode` can also be eliminated if we stream encode each node directly.

Checkpoint file format would need to be modified to enable stream mode.  For example, instead of encoding number of nodes in the header, we can encode a marker after all nodes are serialized.  Same idea for stream encode `StorableTries`.

### Optimize `NodeIterator`

Only return nodes that weren't visited while traversing other tries.

### Reduce buffer allocation

Interim nodes are encoded to checkpoint with the same size, so we can reuse a scratch buffer.  Same for `StorableTrie`.

E.g. `wal.Storecheckpoint()` can pass a reusable scratch buffer to `flattener.EncodeStorableNode()` and `flattener.EncodeStorableTrie()`.

fxamacker:
## Ideas for Reducing Checkpoint File Size 

@ramtinms Maybe we can save over 500MB by not repeating version number in the checkpoint.  Thoughts?

[...]

PR #1894, opened 2022-01-25
Add logs for MTrie checkpoint file creation

fxamacker:
Updates #1884 #1750

Changes include:

Add more logs during checkpoint file creation.  Use INFO instead of DEBUG given the limited frequency and quantity of logs for checkpoint file creation.

These logs can be useful for detecting issues related to checkpoint creation.  E.g. the checkpoint duration log can be used by log monitoring programs to alert us if it exceeds a threshold.

### DIFF
--- a/ledger/complete/wal/checkpointer.go
+++ b/ledger/complete/wal/checkpointer.go
@@ -185,10 +185,14 @@ func (c *Checkpointer) Checkpoint(to int, targetWriter func() (io.WriteCloser, e
 		return fmt.Errorf("cannot replay WAL: %w", err)
 	}
 
+	c.wal.log.Info().Msgf("flattening forest for checkpoint %d\n", to)
+
 	forestSequencing, err := flattener.FlattenForest(forest)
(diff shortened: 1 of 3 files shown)


ramtinms (approved):

ledger/complete/wal/checkpointer.go
```diff
@@ -185,11 +185,15 @@ func (c *Checkpointer) Checkpoint(to int, targetWriter func() (io.WriteCloser, e
 		return fmt.Errorf("cannot replay WAL: %w", err)
 	}
 
+	c.wal.log.Info().Msgf("flattening forest for checkpoint %d\n", to)
```
m4ksio, line 188: Do we want `\n` in logs?

m4ksio (approved):

ledger/complete/wal/checkpointer.go
fxamacker: Good catch!  Fixed in commit 0859bfa.

PR #1899, opened 2022-01-26
Update to Cadence v0.21.0

turbolent:
Update to [Cadence v0.21.0](https://github.com/onflow/cadence/releases/tag/v0.21.0).

Note that this should not be used on production (Testnet and Mainnet), as it is missing security fixes in the internal repo that will need to be ported.


ramtinms (approved):

fvm/fvm_test.go
```diff
@@ -1303,11 +1302,8 @@ func TestBlockContext_GetBlockInfo(t *testing.T) {
 		ledger := testutil.RootBootstrappedLedger(vm, ctx)
 		require.NoError(t, err)
 
-		assert.PanicsWithValue(t, interpreter.ExternalError{
-			Recovered: logPanic{},
-		}, func() {
-			_ = vm.Run(blockCtx, fvm.Transaction(tx, 0), ledger, programs.NewEmptyPrograms())
-		})
+		err = vm.Run(blockCtx, fvm.Transaction(tx, 0), ledger, programs.NewEmptyPrograms())
+		require.Error(t, err)
```
ramtinms, line 1306: I updated these two testcases since cadence now handles these cases gracefully (according to https://github.com/onflow/cadence/pull/1364/files) 

@turbolent @janezpodhostnik just checking if this is the expected behaviour

janezpodhostnik: The error will be classified as a failure error in `SplitErrorTypes` and will crash the node so I think this is ok.

janezpodhostnik (approved):

vishalchangrani:
bors merge

fvm/fvm_test.go
turbolent: 👍

PR #1903, opened 2022-01-27
Fix undefined: sha3State errors when building on arm64

peterargue:
After https://github.com/onflow/flow-go/pull/1865 was merged, we started seeing the following error while building flow on arm64 machines:
```
 => ERROR [build-production 2/3] RUN --mount=type=cache,target=/go/pkg/mod     --mount=type=cache,target=/root/.cache/go-build     GO111MODULE=on CGO_ENABLED=1 GOOS=linux go build --tags "relic,netg  0.6s
------
 > [build-production 2/3] RUN --mount=type=cache,target=/go/pkg/mod     --mount=type=cache,target=/root/.cache/go-build     GO111MODULE=on CGO_ENABLED=1 GOOS=linux go build --tags "relic,netgo" -ldflags "-extldflags -static     -X 'github.com/onflow/flow-go/cmd/build.commit=cedf6df37c81c5c82a199cd640aa0b503c5b96b7' -X  'github.com/onflow/flow-go/cmd/build.semver=v0.23.1-1557-gcedf6'"     -o ./app ./cmd/collection:
#21 0.593 # github.com/onflow/flow-go/crypto/hash
#21 0.593 crypto/hash/xor_generic.go:48:15: undefined: sha3State
#21 0.593 crypto/hash/xor_generic.go:59:27: undefined: sha3State
------
executor failed running [/bin/sh -c GO111MODULE=on CGO_ENABLED=1 GOOS=linux go build --tags "relic,netgo" -ldflags "-extldflags -static     -X 'github.com/onflow/flow-go/cmd/build.commit=${COMMIT}' -X  'github.com/onflow/flow-go/cmd/build.semver=${VERSION}'"     -o ./app ./cmd/${TARGET}]: exit code: 2
make: *** [docker-build-collection] Error 1
```

This appears to be because the `sha3State` struct was replaced by `spongeState`, but the refactor missed this one file.

PR #1904, opened 2022-01-27
Support arm64 when building flow docker containers and localnet

peterargue:
This PR updates the `make docker-build-*` commands to override the `GOARCH` when running on an `arm64` CPU (Apple M1). It also makes this change when building localnet containers.

This should provide transparent support for building and testing on `arm64` base machines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ GOPRIVATE=github.com/dapperlabs/*
 # OS
 UNAME := $(shell uname)
 
+# Used when building within docker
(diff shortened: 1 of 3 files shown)


sjonpaulbrown:
With these changes as well as those on https://github.com/onflow/flow-go/pull/1903, I was able to successfully execute a `make docker-build-flow` and have the containers built on my M1 Mac. Thank you for this work @peterargue!

sjonpaulbrown (approved):

peterargue:
bors merge

PR #1905, opened 2022-01-27
[Execution] [Access] Report execution errors for cadence scripts as OK in Access Node RPC calls

koko1123:
This review makes the following changes:
- Reports script execution errors in EN as with the `InvalidArgument` error code
- if this error code is detected in AN, response is set to code `OK`, and value is still `nil`

testing:
unit tests for new error code included in execution node suite
 
Fixes https://github.com/dapperlabs/flow-go/issues/5385


engine/execution/rpc/engine.go
```diff
@@ -164,7 +164,8 @@ func (h *handler) ExecuteScriptAtBlockID(
 
 	value, err := h.engine.ExecuteScriptAtBlockID(ctx, req.GetScript(), req.GetArguments(), blockID)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to execute script: %v", err)
+		// return code 3 as this passes the litmus test in our context
+		return nil, status.Errorf(codes.InvalidArgument, "failed to execute script: %v", err)
```
peterargue, line 168: maybe use `status.Code(err)` instead of hardcoding to `codes.InvalidArgument` in case the underlying implementation of `ExecuteScriptAtBlockID` changes

m4ksio: But wouldn't that bound error type returned from engine to implement gRPC specific interface therefore bounding it to types?

m4ksio (approved):

engine/execution/rpc/engine.go
peterargue: yea, after looking further, this doesn't make sense.

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -104,11 +104,20 @@ func (b *backendScripts) executeScriptOnExecutionNode(
 				Msg("Successfully executed script")
 			return result, nil
 		}
+		// return OK status if it's just a script failure as opposed to an EN failure
+		if status.Code(err) == codes.InvalidArgument {
+			b.log.Debug().
```
vishalchangrani, line 109:
```suggestion
			b.log.Debug().Err(err).
```

engine/access/rpc/backend/backend_test.go
```diff
@@ -1616,6 +1616,74 @@ func (suite *Suite) TestExecutionNodesForBlockID() {
 	})
 }
 
+// TestExecuteScriptOnExecutionNode tests the method backend.scripts.executeScriptOnExecutionNode for script execution
+func (suite *Suite) TestExecuteScriptOnExecutionNode() {
```
vishalchangrani, line 1620: ❤️

vishalchangrani (approved):
left a nit but otherwise lgtm! thanks for putting this together.

PR #1907, opened 2022-01-27
Allow registering custom node shutdown handlers

peterargue:
Add a new method to scaffold that allows registering custom shutdown handlers to run after all components have stopped.

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -1091,17 +1091,15 @@ func (fnb *FlowNodeBuilder) onStart() error {
 // postShutdown is called by the node before exiting
 // put any cleanup code here that should be run after all components have stopped
 func (fnb *FlowNodeBuilder) postShutdown() error {
+	// close the DBs last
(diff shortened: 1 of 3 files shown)


cmd/scaffold.go
```diff
@@ -1083,11 +1091,20 @@ func (fnb *FlowNodeBuilder) onStart() error {
...
+	var errs *multierror.Error
+	for _, fn := range fnb.postShutdownFns {
+		err := fn()
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+
 	err := fnb.closeDatabase()
 	if err != nil {
```
smnzhu, line 1103: nit: We can probably implement this closeDatabase() as a special ShutdownFunc inside scaffold.

smnzhu (approved):

cmd/scaffold.go
peterargue: updated fde01f5

turbolent (approved):

peterargue:
bors merge

PR #1909, opened 2022-01-28
Merkle trie updates

tarakby:
- address some late comments in the previous Merkle trie [PR](https://github.com/onflow/flow-go/pull/1833#pullrequestreview-863237021)
- limit the maximum bit size to 65535 (more precisely to the largest multiple of 8 smaller than 65535). The size 65536 is not mapped to 0 anymore. 0 is just a not an allowed key size, checked at the tree constructor level
- the above change simplifies conversions of the key lengths in bits since all path lengths are guaranteed to fit in `uint16`.
- add a test for security requirements on hashing tags, used by the security proof.


storage/merkle/proof.go
```diff
@@ -76,51 +75,46 @@ func (p *Proof) validateFormat() error {
...
 
 	// size checks
 	if len(p.InterimNodeTypes) == 0 {
-		return NewMalformedProofErrorf("InterimNodeTypes is empty")
+		return newMalformedProofErrorf("InterimNodeTypes is empty")
 	}
 	if len(p.InterimNodeTypes) > maxKeyLength {
-		return NewMalformedProofErrorf("InterimNodeTypes is larger than max key length allowed (%d > %d)", len(p.InterimNodeTypes), maxKeyLength)
+		return newMalformedProofErrorf("InterimNodeTypes is larger than max key length allowed (%d > %d)", len(p.InterimNodeTypes), maxKeyLength)
 	}
```
AlexHentschel, line 95: ## ⚠️ 

I am not sure, but after taking a second look, I think these checks might be problematic:
* https://github.com/onflow/flow-go/blob/8f41c5c459f4be76e3afb64f4b11bbfd5b2d905d/storage/merkle/proof.go#L93-L95 `maxKeyLength` gives the key lengths in _bytes_. But theoretically, there could be a an interim node for every _bit_.
* furthermore, I think _both_ checks are redundant. We already compute precisely the expected key length: https://github.com/onflow/flow-go/blob/8f41c5c459f4be76e3afb64f4b11bbfd5b2d905d/storage/merkle/proof.go#L97-L98 Why add additional checks that are less precise and only add redundant computation? 

#### Suggestion
I think both checks above can be removed.

storage/merkle/proof.go
```diff
@@ -76,51 +75,46 @@ func (p *Proof) validateFormat() error {
...
 		numberOfShortNodes += bits.OnesCount8(d)
 	}
 
 	if numberOfShortNodes != len(p.ShortPathLengths) {
-		return NewMalformedProofErrorf("len(ShortPathLengths) (%d) does not match number of set bits in InterimNodeTypes (%d)", len(p.ShortPathLengths), numberOfShortNodes)
-	}
-
-	numberOfFullNodes := steps - numberOfShortNodes
-	if numberOfFullNodes != len(p.SiblingHashes) {
-		return NewMalformedProofErrorf("len(SiblingHashes) (%d) does not match number of set bits in InterimNodeTypes (%d)", numberOfFullNodes, len(p.SiblingHashes))
```
AlexHentschel, line 117: 💡 nice insight. Yes, this check is redundant.

storage/merkle/tree.go
```diff
@@ -13,13 +13,10 @@ import (
...
-// we use the convention:
-//  * for path length l with 1 ≤ l ≤ 65535: we represent l as unsigned int with big-endian encoding
-//  * for l = 65536: we represent l as binary 00000000 00000000
-// This convention organically utilizes the natural occurring overflow and is therefore extremely
-// efficient. In summary, we are able to represent key length of up to 65536 bits, i.e. 8192 bytes.
-const maxKeyLength = 8192
+// Note that a short node with zero path length is not part of our storage model.
+// In summary, we are able to represent key lengths multiple of 8 of up to 65535 bits, i.e. less than 8191 bytes.
+const maxKeyLenBits = 65535
+const maxKeyLength = maxKeyLenBits / 8
```
AlexHentschel, line 19: To me, this is confusing because we state essentially conflicting information about the max possible key length:
* we say that a key can at most be `8191` _bytes_ long (which is `65528` bits)
* but we also state that a key can be `65535` bits long

My suggestion would be to just use 

```suggestion
// Hence, the theoretically possible value range is [0,65535]. However, a short node with
// zero path length is not part of our storage model. Furthermore, we always represent
// keys as _byte_ slices, i.e. their number of bits must be an integer-multiple of 8.
// Therefore, the longest possible key is 8191 bytes long. In summary, the range of valid
// values for a short node's length in bits is [1, 65528].
const maxKeyLength = 8191
```

and adjust the [test](https://github.com/onflow/flow-go/blob/8f41c5c459f4be76e3afb64f4b11bbfd5b2d905d/storage/merkle/node_test.go#L87-L104) to use a `8191`-byte key

storage/merkle/node_test.go
```diff
@@ -69,19 +84,20 @@ func TestShortHash(t *testing.T) {
...
 		}
+		path = append(path, k[:maxKeyLength%len(k)]...)
 		// expected value from python reference implementation
-		ref := "b49541ce56e5cd207cac3a08821ddfc61fc5aaf8886907d226581a7c53a48827"
+		ref := "22eb614e20aae939ff38a0dced42fdebb95199e725f1c898e2c5141e7c9d6848"
 
 		s := short{
 			path:  path,
-			count: 65536,
+			count: maxKeyLenBits,
```
AlexHentschel, line 100: ⚠️ 
To me, it looks like we are testing a case here that is illegal in our storage model:
* the longest possible key in our storage model is 65528 bits or 8191 bytes. 
* While you construct a key with 655**28** bytes: https://github.com/onflow/flow-go/blob/8f41c5c459f4be76e3afb64f4b11bbfd5b2d905d/storage/merkle/node_test.go#L90-L94 you set the bit length of the key to 655**35**: https://github.com/onflow/flow-go/blob/8f41c5c459f4be76e3afb64f4b11bbfd5b2d905d/storage/merkle/node_test.go#L100

```suggestion
			count: len(path),
```

AlexHentschel (approved):
Looks good. Mostly stylistic comments. The only problematic aspect is this check in the proof, which looks broken to me (?). And a slight inconsistency in the the test.

AlexHentschel:
P.S. it looks like you have generated new values for testing with the Python reference code. Did you commit that code? Don't see anything (feel free to just push to master).

ramtinms (approved):

storage/merkle/errors.go
```diff
@@ -13,8 +13,8 @@ type MalformedProofError struct {
 	err error
 }
 
-// NewMalformedProofErrorf constructs a new MalformedProofError
-func NewMalformedProofErrorf(msg string, args ...interface{}) *MalformedProofError {
+// newMalformedProofErrorf constructs a new MalformedProofError
+func newMalformedProofErrorf(msg string, args ...interface{}) *MalformedProofError {
```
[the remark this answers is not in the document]
tarakby: My bad, my intention was to make the type not exported either.
Having the type (and constructor) not exported makes sure only the package is able to declare a certain error, this allows high level error assertion to identify exactly where the error is coming from.  If the error is exported, other package can mistakenly re-use the same error in a different situation. I enforced that in the crypto package, but maybe it's not needed in all packages. 

I'll revert to exported in this case.

storage/merkle/tree.go
tarakby: I think there is still an advantage of keeping `maxKeyLenBits`. 
The sentence says that we support "key length multiple of 8 bits, up to 65535". Since 65535 in not a multiple of 8, the largest possible length is the multiple of 8 right before, which is 65528. 

This is clearly confusing so I'll try to rephrase it and remove mentions of 65535.

storage/merkle/node_test.go
tarakby: Good catch, I think this is solved now by fixing `maxKeyLenBits` to 65528 in [a21444e](https://github.com/onflow/flow-go/pull/1909/commits/a21444e0faada936ce5994b66642e5fa7c8ad414)

storage/merkle/tree.go
tarakby: [a21444e](https://github.com/onflow/flow-go/pull/1909/commits/a21444e0faada936ce5994b66642e5fa7c8ad414)

storage/merkle/proof.go
AlexHentschel: > That's true, but `len(p.InterimNodeTypes)` also represents a length in bytes (the interim node type is a bit vector)

You are right. I obviously had some wires crossed in my brain. 

> I think the check `len(p.InterimNodeTypes) == 0` is redundant and is covered by the other checks.

storage/merkle/tree.go
```diff
@@ -13,13 +13,13 @@ import (
...
-// This convention organically utilizes the natural occurring overflow and is therefore extremely
-// efficient. In summary, we are able to represent key length of up to 65536 bits, i.e. 8192 bytes.
-const maxKeyLength = 8192
+//
+// Hence, the theoretically possible value range is [0,65535]. However, a short node with
+// zero path length is not part of our storage model. Furthermore, we always represent
+// keys as _byte_ slices, i.e. their number of bits must be an integer-multiple of 8.
+// Therefore, the range of valid key length in bits is [1, 65528].
+const maxKeyLenBits = 65528
+const maxKeyLength = maxKeyLenBits / 8
```
AlexHentschel, line 22:
```suggestion
const maxKeyLength = 8191
const maxKeyLenBits = maxKeyLength * 8
```

AlexHentschel (approved):
Looks great. Thanks for the great cleanup.

PR #1913, opened 2022-01-28
Non-Relic build fix for the Emulator

janezpodhostnik:
This change is needed to upgrade flow-go in the emulator. The emulator uses a non-relic build of flow-go and on v0.24 the non-relic build of flow-go for the emulator is failing.

This creates stubs in place of actual crypto methods when using non-relic build.

Once merged into master it will need to be ported to v0.24. And afterwards both flow-go and flow-go/crypto will need a tag. flow-go will also need to reference the new tag of flow-go/crypto.


tarakby:
I think we need add CI tests for the FVM package (and maybe all "direct-only" dependencies of the emulator from flow-go) that they can compile without the relic flag. I think that would make sure the emulator isn't broken in the future. 

One way to do this is to run `go test -run=NoTestHasThisPrefix ./...` in the FVM (and other packages). It makes sure the code compiles but not necessarily run correctly.

tarakby:
PR https://github.com/onflow/flow-go/pull/1915 is about to be merged to master and has many conflicts with this PR (new crypto API changes), It would be better to wait for that PR to be merged to master first.

tarakby:
I've made another PR to replace this one after we merge Consensus V2 branch into master (many changes and conflicts). 
We can close this one in favour of https://github.com/onflow/flow-go/pull/1922

janezpodhostnik:
Closed in favour of #1922

PR #1917, opened 2022-01-28
Improve handling of post-spork beacon key initialization

jordanschalm:
The current handling of initializing beacon key storage with the beacon key generated during the spork makes some over-broad assumptions about when it is appropriate to assume a beacon key file must exist. These assumptions fail when a node joins the network mid-spork (say, epoch 4), then starts up using a root snapshot already in the next epoch (say, epoch 5) where the node is already a full member of the identity table. This situation is atypical (usually the node would use a root snapshot within the `EpochSetup` phase of epoch 4), but it is still valid. 

Instead, we directly check that the root snapshot is a spork root snapshot, and only then require that the beacon key file is present.

Also:
* change some terminology (DKG key -> beacon key)
* remove checks that the node has a beacon key for future epochs (if a node fails the DKG, they can now fallback to using the staking key)
* add sanity check that the beacon key is consistent with the protocol state


cmd/consensus/main.go
```diff
@@ -241,91 +242,68 @@ func main() {
...
 
-			// if we participated in the DKG protocol for the next epoch, and we
-			// are in EpochCommitted phase, make sure we have saved the
-			// resulting DKG key
-			phase, err := node.State.Final().Phase()
-			if err != nil {
+			if !myBeaconPublicKeyShare.Equals(beaconPrivateKey.PrivateKey.PublicKey()) {
+				return fmt.Errorf("configured beacon key is inconsistent with this node's canonical public beacon key (%s!=%s)",
+					beaconPrivateKey.PrivateKey.PublicKey().String(),
+					myBeaconPublicKeyShare.String())
```
tarakby, line 295: I guess this is equivalent
```suggestion
					beaconPrivateKey.PrivateKey.PublicKey(),
					myBeaconPublicKeyShare)
```

tarakby (approved):
Thanks for making the process more robust!

cmd/consensus/main.go
```diff
@@ -238,92 +239,68 @@ func main() {
...
+			rootSnapshot := node.State.AtBlockID(node.RootBlock.ID())
+			isSporkRoot, err := protocol.IsSporkRootSnapshot(rootSnapshot)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not check whether root snapshot is spork root: %w", err)
 			}
-			if _, ok := initialIdentities.ByNodeID(node.NodeID); !ok {
-				node.Logger.Info().Msg("node joined at epoch boundary, not reading DKG file")
+			if !isSporkRoot {
+				node.Logger.Info().Msg("node starting from mid-spork snapshot, will not read spork random beacon key file")
```
zhangchiqing, line 259: is it possible that node joining from mid-spork, but still have random beacon key, is it worth to check?

cmd/consensus/main.go
```diff
@@ -238,92 +239,68 @@ func main() {
...
 				return nil
 			}
 
-			// otherwise, load and save the key in DB for the current epoch (wrt
-			// root block)
+			// If the node has a beacon key file, then save it to the secrets database
+			// as the beacon key for the first epoch of the root snapshot.
```
zhangchiqing, line 264:
```suggestion
			// as the beacon key for the epoch of the root snapshot.
```

I found the `first` here is a bit confusing, since a snapshot can only belong to one epoch.

cmd/consensus/main.go
jordanschalm: > is it possible that node joining from mid-spork, but still have random beacon key

It's possible the node operator has the beacon key file, but it cannot be valid for the root epoch (the node is not a random beacon committee member in that epoch). 

We could check for the file and warn or crash if we see it, as a likely misconfiguration, but we can't make use of it.

zhangchiqing: OK, BTW, do we log the block height of the snapshot, if it's bootstrapped from a segment? 

Maybe we can log here.

zhangchiqing (approved):

cmd/consensus/main.go
jordanschalm: I actually just added this in a separate PR: https://github.com/onflow/flow-go/pull/1893/files#diff-cae89c2f7daa439631988089cdd51c551e4c25db2452885527785d0774235041R690-R694. With that, we will log the root and latest finalized height on startup.

jordanschalm:
bors merge

PR #1929, opened 2022-02-01
removing signer_index from signrature in the REST API implementation

vishalchangrani:
1. Removing `signer_index` from the REST API implementation.
2. Adding call to `convert.MessageToTransaction` to set the signer_index appropriately during a `create_transaction` call.

closes: https://github.com/dapperlabs/flow-go/issues/6125


engine/access/rest/handler.go
```diff
@@ -36,19 +37,22 @@ type Handler struct {
 	backend        access.API
 	linkGenerator  models.LinkGenerator
 	apiHandlerFunc ApiHandlerFunc
+	chain          flow.Chain
```
vishalchangrani, line 40: The chain ID is needed when converting the incoming JSON transaction to a flow.Transaction. It is used to validate if the address belongs to the chain. This is similar to the gRPC implementation which also needs the chain id to be passed in.

go.mod
```diff
@@ -59,7 +59,7 @@ require (
 	github.com/multiformats/go-multihash v0.1.0
 	github.com/onflow/atree v0.1.1
 	github.com/onflow/cadence v0.21.0
-	github.com/onflow/flow v0.2.3-0.20220124225841-6dd3bed23faf
+	github.com/onflow/flow v0.2.3-0.20220131193101-d4e2ca43a621
```
vishalchangrani, line 62: [the change to open API spec](https://github.com/onflow/flow/pull/790) which removes the `signer_index` has already been merged in.

engine/access/rest/request/transaction.go
```diff
@@ -114,7 +115,17 @@ func (t *Transaction) Parse(raw io.Reader) error {
...
-	})
+	}
+
+	// we use the gRPC method of converting the incoming transaction to a Flow transaction since
+	// it sets the signer_index appropriately.
+	entityTransaction := convert.TransactionToMessage(flowTransaction)
+	flowTx, err := convert.MessageToTransaction(entityTransaction, chain)
+	if err != nil {
+		return err
+	}
```
sideninja, line 126: This is good to really make sure we are using correctly formatted tx

engine/access/rest/handler.go
```diff
@@ -67,7 +71,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// create request decorator with parsed values
-	decoratedRequest := request.Decorate(r)
+	decoratedRequest := request.Decorate(r, h.chain)
```
sideninja, line 74: I'm afraid this is a sneaky but dirty way of solving this, it's nice because it doesn't require change of all handler functions but at the same time it's not great because chain ID has nothing to do with request object itself. For the sake of saving some time maybe we can leave it like that, but I would like to, once we have time, refactor this along with the refactor of handlers to be contained inside a struct that has chain, link generator, backend etc. This pattern is obviously starting to emerge

vishalchangrani: i agree...there was no easy way to pass-in the chain ID other than completely changing the handler interface to accept additional arguments.
I have created this issue to take another stab at it https://github.com/dapperlabs/flow-go/issues/6132 and to pass chain id to CreateTransaction handler during server initialization.

sideninja: I feel we should have a conversation later about how we can improve some parts of the API, I'm sure you have bunch of ideas too and I want to share mine. So we can create some smaller PRs with refactors, but I would do that after we launch v1 and we are sure it's stable.

engine/access/rest/transactions_test.go
```diff
@@ -62,13 +62,16 @@ func validCreateBody(tx flow.TransactionBody) map[string]interface{} {
 			"key_index":       fmt.Sprintf("%d", tx.ProposalKey.KeyIndex),
```
sideninja, line 62: Generally speaking this bug make me think we don't have enough tests on complex transactions with potentially bad data. It should be done as part of new PR in the future for sure.

engine/access/rest/handler.go
vishalchangrani: my thoughts exactly! there is whole notion of session scoped, application scoped, request scope variables that is missing and that makes it difficult to change the implementation to factor in new pieces of information from different sources. But, like you said, I would really like to get this out first and have real world testing against it.

peterargue (approved):

vishalchangrani:
bors merge

PR #1934, opened 2022-02-02
Log unique cadence scripts in AN execution engine

koko1123:
Adding a simple map with small memory and compute footprint to keep track of cadence scripts submitted for execution, no need for LRU until we get in the order of 10e7 unique cadence scripts

fixes https://github.com/dapperlabs/flow-go/issues/5383


engine/access/rpc/backend/backend_scripts.go
```diff
@@ -20,6 +22,7 @@ type backendScripts struct {
 	state             protocol.State
 	connFactory       ConnectionFactory
 	log               zerolog.Logger
+	seenScripts       map[[16]byte]time.Time
```
vishalchangrani, line 25:
```suggestion
	seenScripts       map[[16]byte]time.Time // a map of script hash to time to keep track of unique scripts sent by clients
```

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +93,27 @@ func (b *backendScripts) executeScriptOnExecutionNode(
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to find execution nodes at blockId %v: %v", blockID.String(), err)
 	}
+	// encode to MD5 as low compute/memory lookup key
+	encodedScript := md5.Sum(script) //nolint:gosec
+	tenMinutes := time.Duration(10) * time.Minute
```
vishalchangrani, line 98: can't this be a constant outside somewhere instead of being calculated here?

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +93,27 @@ func (b *backendScripts) executeScriptOnExecutionNode(
...
-				Msg("Successfully executed script")
+			timestamp, seen := b.seenScripts[encodedScript]
+			// log unique script seen in the last 10 minutes
+			if !seen || executionTime.Sub(timestamp) >= tenMinutes {
+				b.log.Debug().
+					Str("execution_node", execNode.String()).
+					Hex("block_id", blockID[:]).
+					Str("script", string(script)).
+					Msg("Successfully executed script")
+				b.seenScripts[encodedScript] = executionTime
```
vishalchangrani, line 115: so if we don't have an LRU, wouldn't this map grow indefinitely? There can be a large number of different type of scripts

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +93,27 @@ func (b *backendScripts) executeScriptOnExecutionNode(
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to find execution nodes at blockId %v: %v", blockID.String(), err)
 	}
+	// encode to MD5 as low compute/memory lookup key
+	encodedScript := md5.Sum(script) //nolint:gosec
```
vishalchangrani, line 97: curious why the `//nolint:gosec` was needed. What was the linter complaining about?

koko1123: `G401: Use of weak cryptographic primitive (gosec)` was the warning, but we're not using it for cryptographic purposes, so it's safe to ignore

engine/access/rpc/backend/backend_scripts.go
koko1123: yes! but since the script is compressed to a 16 byte identifier, and I assumed there aren't an order of ~1,000,000 unique scripts, so this wouldn't be an issue. I can switch to LRU, what's a reasonable cache size?

engine/access/rpc/backend/backend_scripts.go
vishalchangrani: got it ..thanks

engine/access/rpc/backend/backend_scripts.go
vishalchangrani: tbh - i don't know what the cache size should be but we can start with ~10K maybe?

vishalchangrani: btw - is there a way put things in the cache only if we are logging at debug level? Then we could still run with just a map since in production we run with log level `info`. Only when we want to see the scripts we swtich to log level debug.

vishalchangrani: Also - @yhassanzadeh13 has a new HeroCache which is an optimized LRU. So lets use that if we are switching.

koko1123: a way to cache only when we're logging at debug level is to just put in a check like `if log.level == log.DebugLevel`. Looked up HeroCache, it's meant for `flow.Identifier` -> `flow.Entity` key/vals for which `script string -> flow.Entity` casts have to be done. I'm leaning on using the `if` branch for space/computation simplicity.

vishalchangrani: thanks for looking into HeroCache. 
Ok yeah lets go with `if` branch.

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +96,28 @@ func (b *backendScripts) executeScriptOnExecutionNode(
...
 
 	// try each of the execution nodes found
 	var errors *multierror.Error
 	// try to execute the script on one of the execution nodes
 	for _, execNode := range execNodes {
+		executionTime := time.Now()
```
vishalchangrani, line 106: actually, this too can be inside the `if b.log.GetLevel` block since we don't really care about the exact time when the script was executed.

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -14,12 +16,16 @@ import (
 	"github.com/onflow/flow-go/storage"
 )
 
+// uniqueErrorLogTimeWindow is the duration for checking the uniqueness of scripts sent for execution
+const uniqueErrorLogTimeWindow = time.Duration(10) * time.Minute
```
vishalchangrani, line 20: nit: since this has nothing to do with error could we rename it to `unqiueScriptLoggingTimeWindow` or something to that effect?

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -14,12 +16,16 @@ import (
 	"github.com/onflow/flow-go/storage"
 )
 
+// uniqueScriptLoggingTimeWindow is the duration for checking the uniqueness of scripts sent for execution
+const uniqueScriptLoggingTimeWindow = time.Duration(10) * time.Minute
```
peterargue, line 20: Nit: the compiler will figure out the type in this case
```suggestion
const uniqueScriptLoggingTimeWindow = 10 * time.Minute
```

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -14,12 +16,16 @@ import (
...
+
 type backendScripts struct {
 	headers           storage.Headers
 	executionReceipts storage.ExecutionReceipts
 	state             protocol.State
 	connFactory       ConnectionFactory
 	log               zerolog.Logger
+	seenScripts       map[[16]byte]time.Time // a map of script hash to time to keep track of unique scripts sent by clients
```
peterargue, line 28: Nit:
```suggestion
	seenScripts       map[[md5.Size]byte]time.Time // a map of script hash to time to keep track of unique scripts sent by clients
```

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +96,28 @@ func (b *backendScripts) executeScriptOnExecutionNode(
...
-				Msg("Successfully executed script")
+			if b.log.GetLevel() == zerolog.DebugLevel {
+				executionTime := time.Now()
+				timestamp, seen := b.seenScripts[encodedScript]
+				// log if the script is unique in the time window
+				if !seen || executionTime.Sub(timestamp) >= uniqueScriptLoggingTimeWindow {
+					b.log.Debug().
+						Str("execution_node", execNode.String()).
+						Hex("block_id", blockID[:]).
+						Str("script", string(script)).
```
peterargue, line 116: maybe include the hash in the message as well so it's easier to match up later

peterargue (commented):
Added a few nits. It might be worth adding some logic to periodically cleanup old entries from the `seenScripts` map, otherwise it will grow unbounded. it could just be a small goroutine that runs periodically and removes entries that are older than `uniqueScriptLoggingTimeWindow`

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -14,12 +16,16 @@ import (
...
+// uniqueScriptLoggingTimeWindow is the duration for checking the uniqueness of scripts sent for execution
+const uniqueScriptLoggingTimeWindow = 10 * time.Minute
+
 type backendScripts struct {
 	headers           storage.Headers
 	executionReceipts storage.ExecutionReceipts
 	state             protocol.State
 	connFactory       ConnectionFactory
 	log               zerolog.Logger
+	seenScripts       map[[md5.Size]byte]time.Time // to keep track of unique scripts sent by clients
```
peterargue, line 28: maybe add a comment here that the max memory size of this slice is bounded to `1MB (2^16*2*8)`

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +96,29 @@ func (b *backendScripts) executeScriptOnExecutionNode(
...
+				// log if the script is unique in the time window
+				if !seen || executionTime.Sub(timestamp) >= uniqueScriptLoggingTimeWindow {
+					b.log.Debug().
+						Str("execution_node", execNode.String()).
+						Hex("block_id", blockID[:]).
+						Hex("script_hash", encodedScript[:]).
+						Str("script", string(script)).
+						Msg("Successfully executed script")
+					b.seenScripts[encodedScript] = executionTime
+				}
```
peterargue, line 120: Think it's worthwhile to log just the script_hash in the else case?

engine/access/rpc/backend/backend.go
```diff
@@ -96,6 +98,7 @@ func New(
 			connFactory:       connFactory,
 			state:             state,
 			log:               log,
+			seenScripts:       seenScripts,
```
peterargue, line 101: Nit: no need for the extra variable.
```suggestion
			seenScripts:       make(map[[md5.Size]byte]time.Time),
```

peterargue (approved):
Looks good. added a few small comments

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -90,18 +96,29 @@ func (b *backendScripts) executeScriptOnExecutionNode(
...
 	// try to execute the script on one of the execution nodes
 	for _, execNode := range execNodes {
 		result, err := b.tryExecuteScript(ctx, execNode, execReq)
 		if err == nil {
-			b.log.Debug().
-				Str("execution_node", execNode.String()).
-				Hex("block_id", blockID[:]).
-				Str("script", string(script)).
-				Msg("Successfully executed script")
+			if b.log.GetLevel() == zerolog.DebugLevel {
```
vishalchangrani, line 108: this is **optional** but i personally prefer less nesting and we could combine the `if err == nil` with this condition with an `&&`.

vishalchangrani (approved):

PR #1937, opened 2022-02-02
Update to Cadence v0.21.1-patch.1

turbolent:
Diff: https://github.com/dapperlabs/cadence-internal/compare/v0.21.0...v0.21.1-patch.1

v0.21.1-patch.1 is v0.21.1, which is v0.21.0 + https://github.com/onflow/cadence/pull/1395, and the security fixes from v0.20

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,4 +108,4 @@ replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
 
 replace github.com/onflow/flow-go/crypto => ./crypto
 
-replace github.com/onflow/cadence => github.com/dapperlabs/cadence-internal v0.21.0-patch.1
+replace github.com/onflow/cadence => github.com/dapperlabs/cadence-internal v0.21.1-patch.1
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -107,3 +107,5 @@ require (
 replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
 
 replace github.com/onflow/flow-go/crypto => ./crypto
+
+replace github.com/onflow/cadence => github.com/dapperlabs/cadence-internal v0.21.0-patch.1
```
vishalchangrani, line 111: is this something that will be switched back to the non-internal public facing repo before the mainnet spork on Feb 9th? Some operators compile code from scratch and have complained about reference to internal repo in the past :(

vishalchangrani (approved):

go.mod
turbolent: Yes, once the security patch is deployed to the networks, we will port them to the public repository. We are aware of the problem of being unable to compiling the code from source, but we cannot make security fixes public before they are not deployed.

We try to get the fixes in as part of the regularly scheduled sporks to avoid additional ad-hoc / unplanned software updates that need to be performed in coordination. If it would be easier to do deploy the fix outside of the spork schedule, then we should consider changing the current approach.

SupunS (approved):
🎉

bluesign:
Now spork is more interesting for me, another one bites the dust or not

PR #1938, opened 2022-02-02
Fix non-x86 linter issue

tarakby:
- a declared constant is only used in a x86 file
- update the code to use the constant for both x86 and non-x86 files. 
- this solves a linter issue reported on ARM


koko1123 (approved):
Thanks for this!

AlexHentschel (approved):

PR #1939, opened 2022-02-03
Fix flow-go failing to cross compile

janezpodhostnik:
The cli build failed when building for different architectures that the current one.

I am not sure why this fixes it, but it does.


tarakby (approved):

sideninja (approved):
Nice catch!

PR #1940, opened 2022-02-03
add func that will check for snapshot validity

kc1116:
This PR updates the **_GetLatestProtocolStateSnapshot_** RPC endpoint so that it gurantee's the following 2 conditions about the SealingSegment of the Snapshot it returns

1. The TAIL and HEAD blocks are within the same epoch
2. The TAIL and HEAD blocks are within the same epoch phase

The protocol bootstrapping currently explicitly [checks for the first condition](https://github.com/onflow/flow-go/blob/13cf439cfed8f049c0fd22815fdf07ed43495613/state/protocol/badger/validity.go#L188-L194), but not the second.


engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +226,65 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
 
-	return data, nil
+	phaseAtLowest, err := snapshotAtLowest.Phase()
+	if err != nil {
+		return nil, err
+	}
+
+	if counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest {
+		for i := len(segment.Blocks) - 1; i >= 0; i-- {
+			snapshotAtBlock := b.state.AtBlockID(segment.Blocks[i].ID())
```
jordanschalm, line 270: Since we are only dealing with finalized blocks here, we can query `AtHeight` to save computing these hashes. (if we do this please note this constraint in `getValidSnapshot`)

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +226,65 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+
+	validSnapshot, err := b.getValidSnapshot(snapshot)
+	if err != nil {
+		return nil, err
+	}
+
+	return convert.SnapshotToBytes(validSnapshot)
+}
+
+func (b *Backend) getValidSnapshot(snapshot protocol.Snapshot) (protocol.Snapshot, error) {
```
jordanschalm, line 239: We should limit our iteration here to avoid both (1) expending a lot of time iterating in this endpoint and (2) sending snapshots with unusually large sealing segments over the wire. I think a reasonable limit (`SnapshotHistoryLimit`) would be ~50 blocks, since they typical sealing segment length is <10. 

We can define this limit more concretely as the maximum difference between the global latest finalized block height, and the block height of the lowest block in `snapshot.SealingSegment`, so it covers both cases.

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +226,65 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
 	}
 
-	return data, nil
+	phaseAtLowest, err := snapshotAtLowest.Phase()
+	if err != nil {
+		return nil, err
+	}
+
+	if counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest {
+		for i := len(segment.Blocks) - 1; i >= 0; i-- {
```
jordanschalm, line 269: Could also use [binary search](https://pkg.go.dev/sort#Search) here - the code would be quite similar. Though if we have a hard lookback limit of 50 and skip the hashing, it's probably unnecessary.

jordanschalm (commented):
Draft generally looks good to me! Added a few performance suggestions.

engine/access/rpc/backend/backend.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/bef03a1f8a838f5ee4a723f070c81b7e85c2cd62

engine/access/rpc/backend/backend.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/fe38afdec394fd3b9ac6caaf99023882dc95b68a

engine/access/rpc/backend/backend.go
kc1116: Querying state by height, and have 50 block limit , so this is not needed.

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,80 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+// getValidSnapshot will return a valid snapshot that has a sealing segment which
+// 1. does not contain any blocks that span an epoch transition
+// 2. does not contain any blocks that span an epoch phase transition
+// If a snapshot does contain an invalid sealing segment query the state
+// by height of each block in the segment and return a snapshot at the point
+// where the transition happens.
+func (b *Backend) getValidSnapshot(snapshot protocol.Snapshot, blocksVisited int) (protocol.Snapshot, error) {
+	if blocksVisited > snapshotHistoryLimit {
+		return nil, fmt.Errorf("failed to get valid snapshot reached look back limit ")
+	}
```
zhangchiqing, line 252: Could you explain why adding this limit in the comment? What's the concern if there is no much limit?
also print the blocksVisited and the limit

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,80 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+				return nil, err
+			}
+
+			phaseAtBlock, err := snapshotAtBlock.Phase()
+			if err != nil {
+				return nil, err
+			}
+
+			if counterAtHighest != counterAtBlock || phaseAtHighest != phaseAtBlock {
+				fmt.Println(blocksVisited)
```
zhangchiqing, line 300: better to remove debug log

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,80 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+	phaseAtLowest, err := snapshotAtLowest.Phase()
+	if err != nil {
+		return nil, err
+	}
+
+	// check if any epoch or phase transition occurs, visit each node
+	// in strict order of decreasing height starting at head
+	if counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest {
+		for i := len(segment.Blocks) - 1; i >= 0; i-- {
+			blocksVisited++
```
zhangchiqing, line 287: better not to update the input value, instead, do:

```
return b.getValidSnapshot(snapshotAtBlock, blocksVisited + 1)
```

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,80 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+
+	phaseAtLowest, err := snapshotAtLowest.Phase()
+	if err != nil {
+		return nil, err
+	}
+
+	// check if any epoch or phase transition occurs, visit each node
+	// in strict order of decreasing height starting at head
+	if counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest {
+		for i := len(segment.Blocks) - 1; i >= 0; i-- {
```
zhangchiqing, line 286: I don't think we need query the segment for every height. Instead we could do:
```
if a segment has crossed epoch
  if no, then return the snapshot
  if yes, then iterate through heights until finding the first height that has different phase than the head's phase
    recur
```

The logic would be like this
```
func getValidSnapshot(snapshot) {
  segment := snapshot.Segment()
  lowestEpoch, lowestPhase, lowestHeight := getLowest(segment)
  highestEpoch, highestPhase, highestHeight := getHighest(segment)
  if lowestEpoch == highestEpoch && lowestPhase == highestPhase {
    return snapshot
  }

  // find the first height that is in a different epoch/phase
  loop from highest to lowest {
    epoch, phase := getEpochPhase(height)
    if epoch == highestEpoch && phase == highestPhase {
      continue
    }

    // recur again to check the segment no longer cross epoch/phase
    //  if still crossing, then it will get the snapshot of previous 
    return getValidSnapshot(getSnapshotAtHeight(height))
  } 
}

```

Also I wonder if possible we can get the height range for epoch or phase if so we can further simplify to be:
```
func getValidSnapshot(snapshot) {
  segment := snapshot.Segment()
  lowestEpoch, lowestPhase, lowestHeight := getLowest(segment)
  highestEpoch, highestPhase, highestHeight := getHighest(segment)
  if lowestEpoch == highestEpoch && lowestPhase == highestPhase {
    return snapshot
  }

  epoch, phase := getPreviousEpochPhase(lowestEpoch, lowestPhase)
  startHeight, endHeight := getHeightRange(epoch, phase)
  // recur again to check the segment no longer cross epoch/phase
  //  if still crossing, then it will get the snapshot of previous 

  return getValidSnapshot(getSnapshotAtHeight(endHeight))
}
```

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,80 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+// by height of each block in the segment and return a snapshot at the point
+// where the transition happens.
+func (b *Backend) getValidSnapshot(snapshot protocol.Snapshot, blocksVisited int) (protocol.Snapshot, error) {
+	if blocksVisited > snapshotHistoryLimit {
+		return nil, fmt.Errorf("failed to get valid snapshot reached look back limit ")
+	}
+
+	segment, err := snapshot.SealingSegment()
+	if err != nil {
+		return nil, err
```
zhangchiqing, line 256: A good practice is always wrap the error with some messages, so that when the error occurs, we could see the call stacks along with some useful parameters.

zhangchiqing (changes requested):

engine/access/rpc/backend/backend.go
kc1116: How does this differ from the current implementation aside from moving the logic to get phase and counter at a specific block into helper funcs **_getLowest(segment)_** _**getHighest(segment)**_ ? Currently we check if there is a epoch or phase transition (in the unhappy path we will recur until we get a valid snapshot ) , if not we simply return the snapshot requested.

engine/access/rpc/backend/backend.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/a24a5f1342adf1523fafcdbf08216a1465ec00b0

engine/access/rpc/backend/backend.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/a24a5f1342adf1523fafcdbf08216a1465ec00b0

engine/access/rpc/backend/backend.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/a24a5f1342adf1523fafcdbf08216a1465ec00b0

engine/access/rpc/backend/backend.go
kc1116: Also should note we are not querying the state for every height, current implementation checks if there is a difference in epoch or epoch phase at each height in the segment starting at the highest. In most sealing segments (5-10) blocks this check will only be about 3-4 blocks before finding the block on the boundary of the transition. Only in the worst case scenario (segment with a lot of blocks) will we iterate deep into the segment blocks.

zhangchiqing: Let's take an example:
Epoch1's three phases are ranging from height: `[1001-1500][1501-1800][1801-1840]`
Epoch2's three phases are ranging from height: `[1841-2500][2501-3000][3001-3080]`

Now, if `GetLatestProtocolStateSnapshot` is called at block 2530, which is at the 2nd phase of Epoch2, then you only do the following querys:
1. find the segment, and it returns the block of the segment ranges from `[2460-2530]`, which goes from Epoch2's 1st phase to Epoch2's 2nd phase.
2. Instead of going 1 height at a time from `2530,2529,2528,...2501,2500` until reaching 2500 to find the first block that is in a different phase, which will loop 30 times, we could jump to 2500 right away, because we know it has crossed the phase, and we the previous phase is phase 1, and it's block range is `[1841-2500]` so I can get the snapshot of the end height (2500) of previous phase, and recursively run the check again.

If we want to keep track of the blockVisited, we could add the range that we jumped: 
```
func getValidSnapshot(snapshot, blockVisited) (snapshot, error) {
  if blockVisited > limit {
    return nil, err
  }

  segment := snapshot.Segment()
  lowestEpoch, lowestPhase, lowestHeight := getLowest(segment)
  highestEpoch, highestPhase, highestHeight := getHighest(segment)
  if lowestEpoch == highestEpoch && lowestPhase == highestPhase {
    return snapshot
  }

  epoch, phase := getPreviousEpochPhase(lowestEpoch, lowestPhase)
  startHeight, endHeight := getHeightRange(epoch, phase)
  // recur again to check the segment no longer cross epoch/phase
  //  if still crossing, then it will get the snapshot of previous 

  return getValidSnapshot(getSnapshotAtHeight(endHeight), blockVisited + (highestHeight - endHeight))
}
```

kc1116: Correct me if i'm wrong but at the moment the endpoint has no information about the height ranges of phases or when a phase starts or stops. We only know the current phase we are in for the given snapshot, this is not possible without that information.

engine/access/rpc/backend/backend.go
jordanschalm: The `blocksVisited` parameter variable is a primitive so I agree it is fine to modify it directly here, rather than creating a new variable.

engine/access/rpc/backend/backend.go
jordanschalm: Unfortunately we don't index the heights of epoch/phase transitions, so we do need to search the segment to find the transition point. 

In general, I don't think we need to over-engineer this too much for a few reasons:
* The vast majority of sealing segments do not transit epoch/phase transitions (we will very, very rarely enter the linear search)
* The vast majority of sealing segments are <10 blocks (when we do enter the linear search, it is inexpensive)
* In practice, if a sealing segment transits an epoch/phase transition, the next segment we check will not, and we will be able to return without a second linear search
* If any of the above points are not true, then we want to fail fast without returning any snapshot at all, because it will be either a huge amount of data, or very outdated (ie. we are explicitly not trying to handle the expensive edge cases of this function)

zhangchiqing: If we can't get the height range for a given epoch/phase, then what do you think about my first suggestion?

```
func getValidSnapshot(snapshot) {
  segment := snapshot.Segment()
  lowestEpoch, lowestPhase, lowestHeight := getLowest(segment)
  highestEpoch, highestPhase, highestHeight := getHighest(segment)
  if lowestEpoch == highestEpoch && lowestPhase == highestPhase {
    return snapshot
  }

  // find the first height that is in a different epoch/phase
  loop from highest to lowest {
    epoch, phase := getEpochPhase(height)
    if epoch == highestEpoch && phase == highestPhase {
      continue
    }

    // recur again to check the segment no longer cross epoch/phase
    //  if still crossing, then it will get the snapshot of previous 
    return getValidSnapshot(getSnapshotAtHeight(height))
  } 
}
```
What this improves is that, we don't need to query `Segment()` when iterate through each block, because iterating through blocks is just to find the first block that's in a different phase.

jordanschalm: > what do you think about my first suggestion?

I think it's good but it seems functionally equivalent to the current implementation. (Though it is clearer to read, so for that reason I think it's worth adopting anyway.)

For example, instead of:
```
if lowestEpoch == highestEpoch && lowestPhase == highestPhase {
    return snapshot
}
```
we have the inverse conditional for entering the segment iteration:
```
if counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest { ... }
```

> What this improves is that, we don't need to query Segment() when iterate through each block

I don't see where this is happening currently 😅

kc1116: We can discuss after PE standup, I still don't see any difference in either of your suggestions from what's currently implemented. Maybe another set of eyes would help cc: @jordanschalm

zhangchiqing: I see. Reading through the logic again makes sense now. I was a bit confused by this check
```
counterAtHighest != counterAtLowest || phaseAtHighest != phaseAtLowest
```

Maybe we could extract into a function:
```
func isEpochOrPhaseDifferent(counter1, counter2, phase1, phase2) bool
```

kc1116: @jordanschalm : _we have the inverse conditional for entering the segment iteration:_

This is because of the semantics of the recursion. In @zhangchiqing suggestion there is no return statement at the end of the func . Thus it's actually much clearer to read currently. 

- get counter and phases 
- check them 
  - if transition, check for block that straddles boundary and do recursion
 - return

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,82 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get epoch counter for snapshot at block %s: %w", segment.Blocks[i].ID(), err)
+			}
+
+			phaseAtBlock, err := snapshotAtBlock.Phase()
+			if err != nil {
+				return nil, fmt.Errorf("failed to get phase for snapshot at block %s: %w", segment.Blocks[i].ID(), err)
+			}
+
+			if counterAtHighest != counterAtBlock || phaseAtHighest != phaseAtBlock {
```
zhangchiqing, line 302:
```suggestion
			// iterate through blocks until we find a block that is either in a different epoch or in a different phase,
			// if found, recursively checking if snapshot of that block is valid, which should eventually find a 
			// valid snapshot, or reach the limit on the number of visited blocks.
			if counterAtHighest != counterAtBlock || phaseAtHighest != phaseAtBlock {
```

zhangchiqing (approved):

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,82 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+// 1. does not contain any blocks that span an epoch transition
+// 2. does not contain any blocks that span an epoch phase transition
+// If a snapshot does contain an invalid sealing segment query the state
+// by height of each block in the segment and return a snapshot at the point
+// where the transition happens.
+func (b *Backend) getValidSnapshot(snapshot protocol.Snapshot, blocksVisited int) (protocol.Snapshot, error) {
+	// NOTE: check if we have reached our history limit, in edge cases
+	// where the sealing segment is abnormally long we want to short circuit
+	// the recursive calls and return an error. The API caller can retry.
+	if blocksVisited > snapshotHistoryLimit {
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/9d2783e1e189d5409391fd720d82533466033f01

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,88 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+		// Visit each node in strict order of decreasing height starting at head
+		// to find the block that straddles the transition boundary.
+		for i := len(segment.Blocks) - 1; i >= 0; i-- {
+			blocksVisited++
+
+			// NOTE: Check if we have reached our history limit, in edge cases
+			// where the sealing segment is abnormally long we want to short circuit
+			// the recursive calls and return an error. The API caller can retry.
+			if blocksVisited > snapshotHistoryLimit {
+				return nil, fmt.Errorf("failed to get valid snapshot reached snapshot history limit")
```
jordanschalm, line 277:
```suggestion
				return nil, fmt.Errorf("failed to get valid snapshot reached snapshot history limit (%d)", snapshotHistoryLimit)
```

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,88 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+	return snapshot, nil
+}
+
+// getCounterAndPhase will return the epoch counter and phase at the specified height in state
+func (b *Backend) getCounterAndPhase(height uint64) (uint64, flow.EpochPhase, error) {
+	snapshot := b.state.AtHeight(height)
+
+	counter, err := snapshot.Epochs().Current().Counter()
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get counter at highest block in the segment: %w", err)
```
jordanschalm, line 302:
```suggestion
		return 0, 0, fmt.Errorf("failed to get counter for block (height=%d): %w", height, err)
```

engine/access/rpc/backend/backend.go
```diff
@@ -226,12 +230,88 @@ func (b *Backend) GetNetworkParameters(_ context.Context) access.NetworkParamete
...
+	}
+
+	phase, err := snapshot.Phase()
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to get phase at highest block in the segment: %w", err)
```
jordanschalm, line 307:
```suggestion
		return 0, 0, fmt.Errorf("failed to get phase for block (height=%d): %w", height, err)
```

engine/access/rpc/backend/backend_test.go
```diff
@@ -134,33 +139,248 @@ func (suite *Suite) TestGetLatestFinalizedBlockHeader() {
...
+			BuildEpoch().
+			CompleteEpoch()
+
+		// get heights of each phase in built epochs
+		epoch1, ok := epochBuilder.EpochHeights(1)
+		require.True(suite.T(), ok)
+		epoch2, ok := epochBuilder.EpochHeights(2)
+		require.True(suite.T(), ok)
+
+		// setup AtBlockID mock returns for state
```
jordanschalm, line 222:
```suggestion
		// setup AtHeight mock returns for state
```

engine/access/rpc/backend/backend_test.go
```diff
@@ -134,33 +139,248 @@ func (suite *Suite) TestGetLatestFinalizedBlockHeader() {
...
```
jordanschalm, line 349:
```suggestion
		// setup AtHeight mock returns for state
```

engine/access/rpc/backend/backend_test.go
```diff
@@ -134,33 +139,248 @@ func (suite *Suite) TestGetLatestFinalizedBlockHeader() {
...
 }
 
-func (suite *Suite) TestGetLatestProtocolStateSnapshot() {
-	// setup the snapshot mock
-	snap := unittest.RootSnapshotFixture(unittest.CompleteIdentitySet())
-	suite.state.On("Final").Return(snap).Once()
+// TestGetLatestProtocolStateSnapshot_NoTransitionSpan tests our GetLatestProtocolStateSnapshot RPC endpoint
+// where the sealing segment for the state requested at latest finalized  block does not contain any blocks that
+// spans an epoch or epoch phase transition.
+func (suite *Suite) TestGetLatestProtocolStateSnapshot_NoTransitionSpan() {
```
jordanschalm, line 145: 👏  Great test coverage and comments, thank you

engine/access/rpc/backend/backend_test.go
```diff
@@ -134,33 +139,248 @@ func (suite *Suite) TestGetLatestFinalizedBlockHeader() {
...
+		// blocks in current state
+		// P <- A(S_P-1) <- B(S_P) <- C(S_A) <- D(S_B) |setup| <- E(S_C) <- F(S_D) |commit| <- G(S_E)
+		epochBuilder.BuildEpoch()
+
+		// add more blocks to our state in the commit phase, this will allow
+		// us to take a snapshot at the height where the epoch1 -> epoch2 transition
+		// and no block spans an epoch phase transition. The third block added will
+		// have a seal for the first block in the commit phase allowing us to avoid
+		// spanning an epoch phase transition.
+		epochBuilder.AddBlocksWithSeals(3, 1)
```
jordanschalm, line 335: Could we add one additional test case which covers the 50-block history limit.

utils/unittest/epoch_builder.go
```diff
@@ -50,15 +51,20 @@ func (epoch EpochHeights) SetupRange() []uint64 {
 
 // CommittedRange returns the range of all heights in the committed phase.
 func (epoch EpochHeights) CommittedRange() []uint64 {
-	return []uint64{epoch.Committed}
+	var heights []uint64
+	for height := epoch.Committed; height < epoch.CommittedFinal; height++ {
```
jordanschalm, line 55:
```suggestion
	for height := epoch.Committed; height <= epoch.CommittedFinal; height++ {
```

If `CommittedFinal` is the final height of the committed phase, this should include it

utils/unittest/epoch_builder.go
```diff
@@ -280,7 +287,7 @@ func (builder *EpochBuilder) BuildEpoch() *EpochBuilder {
 	receiptF := ReceiptForBlockFixture(F)
 
 	// build block G
-	// G contains a seal for block E and a receipt for block F
+	// seal for block N , receipt for block n-1
```
jordanschalm, line 290: seems like the removed comment is the one we want here

utils/unittest/epoch_builder.go
```diff
@@ -374,5 +381,48 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
+// will add a seal for the second highest block in the state and a
+// receipt for the highest block in state to the given block before adding it to the state.
+// NOTE: This func should only be used after BuildEpoch to extend the commit phase
+func (builder *EpochBuilder) AddBlocksWithSeals(n int, counter uint64) *EpochBuilder {
+	for i := 0; i < n; i++ {
+		// given the last 3 blocks in state a <- b <- c
+		// the new block will contain
+		// - seal for c
+		// - execution receipt for b
+		// - execution result for a
```
jordanschalm, line 398: This looks outdated - we only need to look back two blocks, and we should be including the receipt and result for the same block (no need to separate them).

utils/unittest/epoch_builder.go
```diff
@@ -374,5 +381,48 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
+		a := builder.blocks[len(builder.blocks)-1]
+		b := builder.blocks[len(builder.blocks)-2]
+
+		// get execution result for third highest block
+		//builder.blocks[len(builder.blocks)-3].Payload.Receipts
```
jordanschalm, line 403: can remove

utils/unittest/epoch_builder.go
```diff
@@ -374,5 +381,48 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
+
+		// get receipt for the highest block
+		receiptF := ReceiptForBlockFixture(a)
```
jordanschalm, line 406:
```suggestion
		receiptA := ReceiptForBlockFixture(a)
```

utils/unittest/epoch_builder.go
```diff
@@ -374,5 +381,48 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
+		// build a new block
+		// this new block contains a seal for previous highest block and a receipt for the highest block
+		block := BlockWithParentFixture(a.Header)
+		seal := Seal.Fixture(
+			Seal.WithResult(b.Payload.Results[0]),
+		)
+		block.SetPayload(flow.Payload{
+			Receipts: []*flow.ExecutionReceiptMeta{receiptF.Meta()},
+			Results:  []*flow.ExecutionResult{&receiptF.ExecutionResult},
+			Seals:    []*flow.Seal{seal},
```
jordanschalm, line 417: Can use the helper functions `WithReceipts` and `WithSeals` here: https://github.com/onflow/flow-go/blob/8f5f429ff969444c527e6a73b1474a37f6c7f18f/utils/unittest/fixtures.go#L259

kc1116: https://github.com/onflow/flow-go/pull/1940/commits/6e64634d3c34f3e167f62d74e1c1cf5e49f5d740

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/6e64634d3c34f3e167f62d74e1c1cf5e49f5d740

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/c6a183c6f9e27cdf6e2e7625f129aaffa5932c16

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/6e64634d3c34f3e167f62d74e1c1cf5e49f5d740

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/6e64634d3c34f3e167f62d74e1c1cf5e49f5d740

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/6e64634d3c34f3e167f62d74e1c1cf5e49f5d740

utils/unittest/epoch_builder.go
```diff
@@ -391,31 +392,25 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
-
-		// get receipt for the highest block
-		receiptF := ReceiptForBlockFixture(a)
-
-		// build a new block
-		// this new block contains a seal for previous highest block and a receipt for the highest block
-		block := BlockWithParentFixture(a.Header)
+		// Given the last 2 blocks in state A <- B when we add block C it will contain the following.
+		// - seal for B
+		// - execution result for A
```
jordanschalm, line 397:
```suggestion
		// - seal for A
		// - execution result for B
```

utils/unittest/epoch_builder.go
```diff
@@ -391,31 +392,25 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
...
+
+		block := BlockWithParentFixture(b.Header)
 		seal := Seal.Fixture(
-			Seal.WithResult(b.Payload.Results[0]),
+			Seal.WithResult(a.Payload.Results[0]),
+		)
+
+		payload := PayloadFixture(
+			WithReceipts(receiptB),
+			WithExecutionResults(&receiptB.ExecutionResult),
```
jordanschalm, line 410:
```suggestion
			WithReceipts(receiptB),
```
When we add a receipt, by default it will include the result (since the result is an embedded field of a receipt)

kc1116: https://github.com/onflow/flow-go/pull/1940/commits/0149bd876a34803334e4b10e3873b68e344a46e3

utils/unittest/epoch_builder.go
kc1116: https://github.com/onflow/flow-go/pull/1940/commits/0149bd876a34803334e4b10e3873b68e344a46e3

jordanschalm:
One outstanding comment about adding a test case for the iteration limit: https://github.com/onflow/flow-go/pull/1940/files#r801981270

kc1116:
> One outstanding comment about adding a test case for the iteration limit: https://github.com/onflow/flow-go/pull/1940/files#r801981270

Missed this one 
https://github.com/onflow/flow-go/pull/1940/commits/3efe3a9375de4c61ff7568e3047ba640da835c3d#diff-6ace96d2d825afe51ad51c2055cbdb47c44e0deb5a86d172898a588fa3f2d593R431

jordanschalm (approved):
🎸

kc1116:
bors merge

Issue #1746, opened 2021-12-09
[Execution state] complete ledger - Separate storage of payloads 


ramtinms:
## Problem Definition

Currently, we do keep payloads as part of the leaf nodes and in memory, if we store them on a disk-based data store we can separate the index (trie) from the actual data. and that should reduce memory usage drastically. should also improve garbage collection.  

This should be done in a way that doesn't impact the time spent on operations like read and update and be parallelizable as much as possible.

Ullaakut:
> [...] if we store them on a disk-based data store [...]
> This should be done in a way that doesn't impact the time spent on operations like read and update and be parallelizable as much as possible.

Do you have suggestions on how to achieve this? It seems to me like the only way to effectively do this is to have both a persistent store on the filesystem and a cache in memory, but then it would not consistently have no impact on those operations. It would have no impact only in ideal scenarios where the cache (LRU would be best) contains all of the values that we need to read, and that those values never need to be retrieved from the filesystem. If we ever need to retrieve a value from the filesystem in order to satisfy a read call, how could it have no impact? We would have to block the read call until we successfully fetched the value from disk, which is probably an operation that is a few orders of magnitude more costly in performance terms than a read on memory is.

The most optimal way I've found is to have [the LRU cache write on disk upon evicting values](https://github.com/optakt/flow-dps/blob/trie-improvements/ledger/store/store.go#L63-L117), and to [regularly evict (and therefore, persist) its oldest entries](https://github.com/optakt/flow-dps/blob/trie-improvements/ledger/store/store.go#L187-L209), to hopefully never reach a full cache (which would mean blocking disk operations) and allow the disk writes to be done concurrently without interfering with new read/write operations, but that solution is limited by how much memory is devoted to the cache. The bigger the cache and the more often its oldest entries are purged, the better performance is to be expected, but even that does not solve the problem where a read call could come for any old key that is now on disk, and that would require a disk read which would inevitably be slow.

PR #1944, opened 2022-02-03
Optimize MTrie checkpoint for speed, memory, and file size

### LINKED ISSUE #1750: [Execution state] - Checkpointer and Flattener speed and memory improvements

fxamacker:
## Description

Optimize checkpoint creating (includes loading):
- 47x speedup (11.7 hours to 15 mins), avoid 431 GB alloc/op, avoid 7.6 billion allocs/op
  - 171x speedup (11.4 hours to 4 mins) in MTrie traversal+flattening+writing phase
- Reduce long-held data in RAM by 116+ GB (lowers hardware requirements)
- Reduce checkpoint file size by 6.9+ GB (another 4.4+ GB reduction planned in separate PR) without using compression

Most of the optimizations were proposed in [comments](https://github.com/onflow/flow-go/issues/1750#issuecomment-1004870851) to issue #1750.  I'm moving all remaining optimizations like concurrency and/or compression, etc. to separate PRs so this is ready for review as-is.

Increased interim + leaf node counts are causing checkpoint creation to take hours.   This PR sacrifices some readability and simplicity as tradeoffs and gains speed, memory efficiency, and storage efficiency.

I limited scope of PR to optimizations that don't require performance tradeoffs or overhead (like adding processes, IPC).

Big thanks to @ramtinms for opening #1750 to point out that trie flattening can have big optimizations. 👍

Closes #1750, #1884
Updates #1746, https://github.com/dapperlabs/flow-go/issues/6114

## Impact on Execution Nodes

:warning: Unoptimized checkpoint creation reaches 248+GB RAM within the first 30 minutes and can run for about 15-17+ hours on EN3.  This duration during heavy load was long enough on EN3 to accumulate enough WAL files to trigger another checkpoint immediately after the current one finishes.  So the 248 GB RAM is held again with 590 GB alloc/op and 9.8 billion allocs/op.

### EN Startup Time

This PR speeds up EN startup time in several ways:

- Checkpoint loading and WAL replaying will be optimized for speed (see benchmarks).
- Checkpoint creation will be fast enough to run multiple times per day, which will reduce WAL segments that need to be replayed during startup.
- Checkpoint creation finishing in minutes rather than 15+ hours reduces risk of being interrupted by shutdown, etc. which can cause extra WAL segments to be replayed on next EN startup.

See issue #1884 for more info about extra WAL segments causing EN startup delays.

### EN Memory Use and System Requirements

This PR can reduce long-held data in RAM (15+ hours on EN3) by up to 116 GB.  Additionally, eliminating 431 GB alloc/op and 7.6 billion allocs/op will reduce load on the Go garbage collector.

#### Unoptimized Checkpoint Creation  
After the first 30 minutes and for next 11+ hours (15-17+ hours on EN3):
image

#### Optimized Checkpoint Creation  
Finishes in 15+ minutes and peaks in the last minute at:
image

#### MTrie Checkpoint Load+Create v3 (old) vs v4 (WIP)

```
Input: checkpoint.00003443 + 41 WAL files
Platform: Go 1.16, benchnet (the big one)
name                        old time/op       new time/op       delta
NewCheckpoint-48            42052s ± 0%        886s ± 0%       -97.89%

name                        old alloc/op      new alloc/op      delta
NewCheckpoint-48             590GB ± 0%       159GB ± 0%       -73.04%

name                        old allocs/op     new allocs/op     delta
NewCheckpoint-48             9.80G ± 0%       2.19G ± 0%       -77.67%

DISCLAIMERS: not done yet, didn't add concurrency yet, n=1 due to duration, 
not reviewed yet.
```

UPDATE: on March 1, optimized checkpoint creation speed (v4 -> v4 with 41 WALs) varied by 63 seconds between the first 2 runs (all 3 used same input files to create same output):
* 926 secs (first run right after OS booted, maybe didn't wait long enough)
* 863 secs (second run without rebooting OS, maybe file system cache helped)
* 879 secs (third run after other activities without rebooting OS)

#### Load Checkpoint File + replay 41 WALs v3 (old) vs v4 (WIP)
```
Input: checkpoint.00003443 + 41 WAL files
Platform: Go 1.16, benchnet (the big one)
name                        old time/op       new time/op       delta
LoadCheckpointAndWALs-48     989s ± 0%         676s ± 0%       -31.64%

name                        old alloc/op      new alloc/op      delta
LoadCheckpointAndWALs-48    297GB ± 0%        136GB ± 0%       -54.35%

name                        old allocs/op     new allocs/op     delta
LoadCheckpointAndWALs-48    5.98G ± 0%        2.17G ± 0%       -63.67%

DISCLAIMERS: not done yet, didn't add concurrency yet, n=1, 
filesystem cache affects speed so delta can be -28% to -32%,
not reviewed yet.
```

### Changes include:
- [x] Create checkpoint file v4 and replace v3, while retaining ability to load older versions. (v4 is not yet finalized).  First, [Reduce checkpoint file size by 5.8+GB](https://github.com/onflow/flow-go/commit/422b75b1c56fdcb968569fde82bfdfd5f726f2c3).  Next, [reduce checkpoint file size by 1.1+GB](https://github.com/onflow/flow-go/pull/1944/commits/dcca74efe8e79b27f4e1ac3919e118fac481156d) by removing encoded hash size and path size.  Further reduction of 4.4+GB is planned for 10.2 GB combined reduction compared to v3.  These file size reductions don't use compression.
- [x] [Use stream encoding and writing](https://github.com/onflow/flow-go/commit/8b03a22c12f51cb1bcb41b2d6e97554b7025be5d) for checkpoint file creation. This reduces RAM use by avoiding the creation of a ~400 million element slice containing all nodes and creation of 400 million objects.  Savings will be about 43.2+ GB plus more from other changes in this PR.
- [x] [Add NewUniqueNodeIterator() to skip shared nodes](https://github.com/onflow/flow-go/commit/04e5c08b0e6fe35d31646addd5a8c30c0baeae20). NewUniqueNodeIterator() can be used to optimize node iteration for forest.  It skips shared sub-tries that were visited and only iterates unique nodes.
- [x] [Optimize reading checkpoint file by reusing buffer](https://github.com/onflow/flow-go/pull/1944/commits/6c5ad37be8b14e7d5aefa6265270a9d7dd44c781).  Reduce allocs by using a 4096 byte scratch buffer to reduce another 400+ million allocs during checkpoint reading.  Since checkpoint creation requires reading checkpoint, this optimization benefits both.
- [x] [Optimize creating checkpoint by reusing buffer](https://github.com/onflow/flow-go/pull/1944/commits/c9a8f145835d7addbc4fbacb98ba091c6892aca6).  Reduce allocs by using a 4096 byte scratch buffer to reduce another 400+ million allocs during checkpoint writing.
- [x] [Skip StorableNode/StorableTrie when creating checkpoint](https://github.com/onflow/flow-go/commit/f185fd93e24bf1323a2feb8661039cca03d8a629) 
  - Merge FlattenForest() with StoreCheckpoint() to iterate and serialize nodes without creating intermediate StorableNode/StorableTrie objects.
  - Stream encode nodes to avoid creating 400+ million element slice holding 400 million StorableNode objects.
  - Change checkpoint file format (v4) to store node count and trie count at the footer (instead of header) required for stream encoding.
  - Support previous checkpoint formats (v1, v3).
- [x] [Skip StorableNode/Trie when reading checkpoint](https://github.com/onflow/flow-go/commit/8b03a22c12f51cb1bcb41b2d6e97554b7025be5d) 
  - Merge RebuildTries() with LoadCheckpoint() to deserialize data to nodes without creating intermediate StorableNode/StorableTrie objects.
  - Avoid creating 400+ million element slice holding all StorableNodes read from checkpoint file.
  - DiskWal.Replay*() APIs are changed.  checkpointFn receives []*trie.MTrie instead of FlattenedForest.
- [x] Add [flattening encoding tests](https://github.com/onflow/flow-go/pull/1944/commits/802347fc627b345cf5b9150549dc91d5be04cc2e), add [checkpoint v3 decoding tests](https://github.com/onflow/flow-go/pull/1944/commits/2b9d1aeca4a45cadfefaf32ee26a3b94f2b56ecf), add [more validation](https://github.com/onflow/flow-go/pull/1944/commits/aa1549eb65ca1e74d740ba17e55b0da42b9b2539), add comments, [refactor code for readability](https://github.com/onflow/flow-go/pull/1944/commits/6618948bb1d290087d306cedd6e8d8fdbe0ab9ef), and etc.

### TODO

- [ ] update benchmark comparisons using latest results

#### Additional TODOs that will probably be wrapped up in a separate PR

- maybe add zeroCopy flag and more tests for these functions: `DecodeKey()`, `DecodeKeyPart()`, and `DecodePayload()`.  Not high priority because these functions appear to be unused.
- further reduce data written to checkpoint part 2 (e.g. encoded payload value size can be uint32 instead of uint64 but changing this affects code outside checkpoint creation)
- micro optimizations (depends on speedup vs readability tradeoff)
- add concurrency
- maybe add file compression or payload compression (only if concurrency is added)
- maybe replace CRC32 with BLAKE3 or BLAKE2 since checkpoint file is >60GB
- maybe encode integers using variable length to reduce space (possibly not needed if/when we use file compression)
- ~~maybe split checkpoint file into 3 files (metadata, nodes, and payload file).~~ I synced with Ramtin and his preference is to keep the checkpoint as one file for this PR.


bluesign:
Hey @fxamacker, first of all great work. 

Does this changes allow using checkpoint file without loading it totally to memory ? If not is it possible to add an index to the end ? It would be great to have ability to query checkpoint file with a CLI tool for lookup. But with recent changes using atree on storage, I have to import all checkpoint file to Badger ( or similar DB ) or load fully in to the memory.

fxamacker:
Hello @bluesign,

> Does this changes allow using checkpoint file without loading it totally to memory ? 

The entire checkpoint file will be stream decoded in order to create in-memory tries.  Currently, the in-memory tries will have all the data from checkpoint file.  Issue #1746 will be handled by a separate PR after this one is merged.

> It would be great to have ability to query checkpoint file with a CLI tool for lookup. But with recent changes using atree on storage, I have to import all checkpoint file to Badger ( or similar DB ) or load fully in to the memory.

I understand. 

> If not is it possible to add an index to the end ? 

Maybe (depends on details/tradeoffs/etc), but adding index is outside the scope of this PR.

There's a chance this PR might store the payload in a separate file.  It is a tiny intermediate step towards #1746. Not sure, but maybe this step could be useful to projects like https://github.com/bluesign/checkpointState

Please consider opening an issue to add index to checkpoint file, so discussions can continue there if needed.  This PR is focused on optimizing checkpoint (creation and loading) for speed, memory, and file size.

ledger/complete/wal/checkpointer.go
```diff
@@ -242,49 +246,129 @@ func CreateCheckpointWriterForFile(dir, filename string) (io.WriteCloser, error)
...
+	// assign unique index to every node
+	allNodes := make(map[*node.Node]uint64)
+	allNodes[nil] = 0 // 0th element is nil
+
+	allRootNodes := make([]*node.Node, len(tries))
+
+	// Scratch buffer is used as temporary buffer that node can encode into.
+	// Data in scratch buffer should be copied or used before scratch buffer is used again.
+	// If the scratch buffer isn't large enough, a new buffer will be allocated.
+	// However, 4096 bytes will be large enough to handle almost all payloads
```
ramtinms, line 296: hmm, not sure if all payloads are expected to fit in 4096 bytes, what if it goes beyond that ?

fxamacker: If the 4096 byte scratch buffer isn't large enough, a new buffer will be allocated.  I think 99% of payloads in the mainnet snapshot were smaller than 2048 bytes and I used 4096 as default.

tarakby (commented):
Very clean 👏🏼

ledger/common/encoding/encoding.go
```diff
@@ -168,21 +184,36 @@ func EncodeKey(k *ledger.Key) []byte {
...
 	return buffer
 }
 
+func encodedKeyLength(k *ledger.Key) int {
+	// Key is encoded as: number of key parts (2 bytes) and for each key part,
+	// the key part size (4 bytes) + encoded key part (n bytes).
+	size := 2
+	for _, kp := range k.KeyParts {
+		size += 4 + encodedKeyPartLength(&kp)
+	}
```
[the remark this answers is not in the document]
fxamacker: Nice suggestion!  :+1:   It reduces 2 cpu instructions to 1 in the loop on linux_amd64 with Go 1.16 and 1.17.  
Added in commit d12c3b3dd1572b167c6ee15cdf12edaa241350a6

ledger/common/encoding/encoding.go
```diff
@@ -327,30 +372,52 @@ func EncodePayload(p *ledger.Payload) []byte {
...
+	buffer = utils.AppendUint32(buffer, uint32(encodedKeyLength(&p.Key)))
 
-	// encode value
-	encV := encodeValue(p.Value)
+	// encode key
+	buffer = encodeAndAppendKey(buffer, &p.Key)
 
 	// encode encoded value size
-	buffer = utils.AppendUint64(buffer, uint64(len(encV)))
+	buffer = utils.AppendUint64(buffer, uint64(encodedValueLength(p.Value)))
```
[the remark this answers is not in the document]
fxamacker: I agree 100%.  This one has been in my notes and was thinking of addressing this in a separate PR.

Currently, the entire payload length is encoded in 4 bytes while the payload value length is encoded in 8 bytes.  So reducing to 4 bytes makes sense.  This encoding is used in checkpoint, TrieUpdate (WAL), and TrieProof.

ledger/complete/mtrie/flattener/encoding.go
```diff
@@ -1,187 +1,422 @@
...
 
-	// 2-bytes Big Endian uint16 encoded path length and n-bytes encoded path
-	buf = utils.AppendShortData(buf, storableNode.Path)
+	pos += copy(scratch[pos:], path[:])
 
-	// 4-bytes Big Endian uint32 encoded payload length and n-bytes encoded payload
-	buf = utils.AppendLongData(buf, storableNode.EncPayload)
+	// Encode payload (4-bytes Big Endian for encoded payload length and n-bytes encoded payload)
+	binary.BigEndian.PutUint32(scratch[pos:], uint32(encPayloadSize))
+	pos += 4
```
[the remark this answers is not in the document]
fxamacker: Right.  It seems Go 1.16 and 1.17 compiler on amd64 can optimize to use final value instead of incrementing `pos`.

ledger/complete/mtrie/flattener/iterator.go
```diff
@@ -75,6 +84,30 @@ func NewNodeIterator(mTrie *trie.MTrie) *NodeIterator {
 	return i
 }
 
+// NewUniqueNodeIterator returns a node NodeIterator, which iterates through all unique nodes
```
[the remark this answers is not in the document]
fxamacker: Interesting point!  I think recursive DFS can work and recursive calls would increase as trie depth grows.  Currently, we allocate array (stack) once with 256 elements (I can optimize it to be fewer).

My hunch is recursion would probably be slower (based past experience with old compilers and function call overhead).  When time allows, I may do benchmark comparisons with recursion to see how it goes with Go 1.17+.

ledger/common/encoding/encoding.go
```diff
@@ -168,21 +184,36 @@ func EncodeKey(k *ledger.Key) []byte {
...
+	// the key part size (4 bytes) + encoded key part (n bytes).
+	size := 2 + 4*len(k.KeyParts)
+	for _, kp := range k.KeyParts {
+		size += encodedKeyPartLength(&kp)
+	}
+	return size
+}
+
 // DecodeKey constructs a key from an encoded key part
 func DecodeKey(encodedKey []byte) (*ledger.Key, error) {
```
ramtinms, line 218: do we need to add zeroCopy flag to the root method like this one and pass it to the underlying methods ?

ledger/complete/mtrie/flattener/encoding.go
```diff
@@ -1,187 +1,456 @@
...
+// so any extra capacity will not be utilized.
+// WARNING: The returned buffer is likely to share the same underlying array as
+// the scratch buffer. Caller is responsible for copying or using returned buffer
+// before scratch buffer is used again.
+// TODO: reduce hash size from 2 bytes to 1 byte.
+func encodeLeafNode(n *node.Node, scratch []byte) []byte {
+
+	encPayloadSize := encoding.EncodedPayloadLengthWithoutPrefix(n.Payload())
+
+	encodedNodeSize := 1 + 2 + 2 + 8 + 2 + hash.HashLen + 2 + ledger.PathLen + 4 + encPayloadSize
```
ramtinms, line 43: moving hash to before payload is very smart.

ledger/complete/mtrie/flattener/encoding.go
```diff
@@ -1,187 +1,456 @@
...
 	}
 
-	storableNode := &StorableNode{}
+	// Read interim node
+
+	pos = 0
 
-	storableNode.Height, buf, err = utils.ReadUint16(buf)
+	// Read left and right child index (8 bytes each)
+	_, err = io.ReadFull(reader, scratch[:16])
```
ramtinms, line 249: isn't it we already have read 13 bytes of the scratch up to this point ?

ledger/complete/mtrie/flattener/encoding.go
```diff
@@ -1,187 +1,456 @@
...
-	storableNode.HashValue, err = utils.ReadShortDataFromReader(reader)
+	mtrie, err := trie.NewMTrie(rootNode)
 	if err != nil {
-		return nil, fmt.Errorf("cannot read hashValue data: %w", err)
+		return nil, fmt.Errorf("failed to restore serialized trie: %w", err)
+	}
+
+	rootHash := mtrie.RootHash()
+	if !rootHash.Equals(ledger.RootHash(readRootHash)) {
+		return nil, fmt.Errorf("failed to restore serialized trie: roothash doesn't match")
```
ramtinms, line 360: +1

ledger/common/encoding/encoding.go
fxamacker: Not sure. I don't think `DecodeKey()` is currently used, but if we add zeroCopy flag to it, then:
- we need to add more tests for `DecodeKey()`
- and for consistency, also add zeroCopy flag to related functions, like `DecodeKeyPart()` and `DecodePayload()`, then add tests for them too

I'll add these to the TODO list of this PR.

ledger/complete/mtrie/flattener/encoding.go
ramtinms: hmm, but don't we need to start reading a full value at index 13 instead of 0
?

ledger/complete/mtrie/forest.go
```diff
@@ -50,7 +50,7 @@ func NewForest(forestCapacity int, metrics module.LedgerMetrics, onTreeEvicted f
 			if !ok {
 				panic(fmt.Sprintf("cache contains item of type %T", value))
 			}
-			//TODO Log error
+			// TODO Log error
```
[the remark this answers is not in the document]
ramtinms: lets remove it

ledger/complete/wal/checkpointer.go
```diff
@@ -30,6 +32,29 @@ const VersionV1 uint16 = 0x01
 // Version 3 contains a file checksum for detecting corrupted checkpoint files.
 const VersionV3 uint16 = 0x03
 
+// Version 4 contains a footer with node count and trie count (previously in the header).
```
ramtinms, line 35: We need to flag that this is a breaking change when we want to deploy

ledger/common/encoding/encoding.go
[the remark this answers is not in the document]
ramtinms: yeah, this seems like a bug in the first place, I can confirm that reducing to 4 should be more than enough.

ramtinms (approved):
👩‍💻 🚀 🌔  - looks great to me, well done @fxamacker

ledger/complete/mtrie/flattener/encoding.go
fxamacker: > don't we need to start reading a full value at index 13 instead of 0?

No because we can overwrite the scratch buffer content after decoding its previous contents.  For example,

1. we read 13 bytes from reader to `scratch[0:13]`
2. we decode data in `scratch[0:13]` (such as node type height, reg count, max depth), so at this point we don't need anything else from `scratch[0:13]`
3. we read 16 bytes from reader to `scratch[0:16]`
4. we decode two uint64 from `scratch[0:16]`

ledger/common/encoding/encoding.go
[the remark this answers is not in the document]
fxamacker: Great!  Opened issue #2070 to use 4 bytes to encode payload value length.

ledger/complete/mtrie/forest.go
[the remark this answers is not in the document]
fxamacker: Removed in 0ee96753c82d52d49152860d19ff44484a98ddec.  

I'll open separate issue/pr to improve logging.  It will add logging of this error and some other warnings/errors/etc.

ledger/complete/wal/checkpointer.go
fxamacker: Great catch. For now, I marked issue #1750 and this PR with "breaking change" label to help us remember.

ledger/common/encoding/encoding.go
```diff
@@ -505,6 +589,9 @@ func decodeTrieUpdate(inp []byte) (*ledger.TrieUpdate, error) {
 		return nil, fmt.Errorf("error decoding trie update: %w", err)
 	}
 
+	paths := make([]ledger.Path, numOfPaths)
+	payloads := make([]*ledger.Payload, numOfPaths)
```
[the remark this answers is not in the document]
AlexHentschel: I think there is a miss-conception involved here. Up to this moment, I also had this (apparently incorrect) assumption:
* When allocating the underlying array for a slice, the compiler needs to write the default values to each slice element.  
* We were under the impression that if you initialize a slice with zero-length but non-zero capacity, the compiler would skip this initialization step.

But your example shows that this assumption is false. Thanks, I learned something 🎉 : 
* https://go.dev/blog/slices-intro explicitly states 
   > Arrays do not need to be initialized explicitly; the zero value of an array is a ready-to-use array whose elements are themselves zeroed:

ledger/complete/mtrie/flattener/iterator.go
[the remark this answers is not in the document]
AlexHentschel: I implemented this algorithm originally. You are right that it is a DFS. The non-recursive stack version is more efficient because it avoids the usage of the stack. (at least assuming that the compiler is not able to turn the recursive into the non-recursive version; at least back then, the recursive version was noticeably slower).

The stack size of 256 is given by the max depth of the tree. We now have the `maxDepth` field (the max number of edges). I think we could use this instead of the full length. Not sure that this would result in any noteworthy performance gain, because for a realistic trie, we need probably anyway around 60 - 100 elements.

ledger/complete/mtrie/flattener/iterator.go
```diff
@@ -53,6 +53,15 @@ type NodeIterator struct {
...
 	stack           []*node.Node
+	// visitedNodes are nodes that were visited and can be skipped during
+	// traversal through dig(). visitedNodes is used to optimize node traveral
+	// IN FOREST by skipping nodes in shared sub-tries after they are visited,
+	// because sub-tries are shared between tries (original MTrie before register updates
+	// and updated MTrie after register writes).
+	// NodeIterator only uses visitedNodes for read operation.
+	// No special handling is needed if visitedNodes is nil.
+	// WARNING: visitedNodes is not safe for concurrent use.
+	visitedNodes map[*node.Node]uint64
```
AlexHentschel, line 64: nice

ledger/complete/mtrie/flattener/iterator.go
```diff
@@ -53,6 +53,15 @@ type NodeIterator struct {
...
 	unprocessedRoot *node.Node
 	stack           []*node.Node
+	// visitedNodes are nodes that were visited and can be skipped during
+	// traversal through dig(). visitedNodes is used to optimize node traveral
+	// IN FOREST by skipping nodes in shared sub-tries after they are visited,
+	// because sub-tries are shared between tries (original MTrie before register updates
+	// and updated MTrie after register writes).
+	// NodeIterator only uses visitedNodes for read operation.
+	// No special handling is needed if visitedNodes is nil.
+	// WARNING: visitedNodes is not safe for concurrent use.
```
AlexHentschel, line 63: could you add this line also to the existing `NewNodeIterator` constructor please
https://github.com/onflow/flow-go/blob/1ebde088e9e978de4fc3a522168f349d9eb6eb15/ledger/complete/mtrie/flattener/iterator.go#L77

ledger/complete/mtrie/forest.go
```diff
@@ -50,7 +50,6 @@ func NewForest(forestCapacity int, metrics module.LedgerMetrics, onTreeEvicted f
 			if !ok {
 				panic(fmt.Sprintf("cache contains item of type %T", value))
 			}
-			//TODO Log error
 			_ = onTreeEvicted(trie)
```
AlexHentschel, line 53: in my opinion, `onTreeEvicted` should not throw an error for the following reasons:
* What happens if we receive an error here? What can we do? I think nothing. 
* I think only `onTreeEvicted` has the context to decide what to do. It could log the error itself or crash. Either way, for the forrest there is nothing to do. 

In fact, this is an instance of the anti-pattern of "continue on a best effort basis, when we receive an error". It is very important to me that we avoid this anti-pattern, because it is fundamentally incompatible with BFT. Would you be able to clean this up (i.e. remove the error return) as part of this PR? That would be great. Thanks.

AlexHentschel (approved):
I just took high-level look at the PR. I mainly focused my review on any changes to the core logic of the trie. Didn't go deep into the checkpointing logic itself, as I am not familiar with this part of the code base.

ledger/complete/mtrie/forest.go
fxamacker: > Would you be able to clean this up (i.e. remove the error return) as part of this PR? That would be great. Thanks.

[Done](https://github.com/onflow/flow-go/pull/1944/commits/c99bc8c9128a56bc0569c38f91964432c7cbe4d3) (cleaned up and improved old error logging as part of this PR).  It was going to be part of issue #2074.

Great suggestion about removing the returned error.  👍  Thanks for detailed and thoughtful explanation.  💎

ledger/complete/mtrie/flattener/iterator.go
fxamacker: I'm not sure we should add the same concurrency warning here because `visitedNodes` will always be `nil` which appears to make it safe for concurrent use (unless I'm mistaken).

I [added this comment](https://github.com/onflow/flow-go/pull/1944/commits/d3a32c6eae1a1606f90a837876f7699765c3e550) to `NewNodeIterator`:

```Go
// NodeIterator created by NewNodeIterator is safe for concurrent use
// because visitedNodes is always nil in this case.
```

Please let me know if I need to update the comment.  🙏

ledger/common/encoding/encoding.go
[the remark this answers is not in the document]
tarakby: Thanks @fxamacker for clarifying this mistake! I mistakenly thought that the pattern :
```
t := make([]int, 0, 10)
for i:=0; i<10; i++ {
  t = append(t, val)
}
```
is faster than :
```
t := make([]int, 10)
for i:=0; i<10; i++ {
  t[i] = val
}
```
because it avoids the initialization of t. It turns out this is not true! 
I've learned something today!

ledger/common/encoding/encoding.go
```diff
@@ -671,7 +760,8 @@ func decodeTrieProof(inp []byte) (*ledger.TrieProof, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error decoding proof: %w", err)
 	}
-	interims := make([]hash.Hash, 0)
+
+	interims := make([]hash.Hash, interimsLen)
```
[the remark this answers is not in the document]
tarakby: You're right! I made the same error here!

ledger/complete/mtrie/flattener/iterator.go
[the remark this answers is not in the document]
tarakby: Thank you both for the details! 
Interesting, I didn't expect the recursive version to be much slower than the non-recursive version. Plus I thought the simplicity of the recursive code is worth the time overhead. It looks like no :) I guess the call stack turns out to be much more expensive to manage than the "manual" stack.

 > When time allows, I may do benchmark comparisons with recursion to see how it goes with Go 1.17+
 No rush for this @fxamacker, I guess you already optimized this process significantly so there is no reason to spend more time on it for now.

PR #1947, opened 2022-02-03
Limit range queries for sync engine

m4ksio:
Limits requesting blocks rages to 2*max range size from latest finalized - this prevents spamming large amounts of requests for very distant blocks. Limit grows naturally with finalization spreading the load.

### DIFF
--- a/engine/common/synchronization/engine.go
+++ b/engine/common/synchronization/engine.go
@@ -4,7 +4,6 @@ package synchronization
 
 import (
 	"fmt"
-	"math"
 	"math/rand"
 	"time"
 
@@ -293,17 +292,11 @@ func (e *Engine) onSyncResponse(originID flow.Identifier, res *messages.SyncResp
 // onBlockResponse processes a response containing a specifically requested block.
 func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockResponse) {
 	// process the blocks one by one
+	min := res.Blocks[0].Header.Height
+	max := res.Blocks[len(res.Blocks)-1].Header.Height
+	e.log.Debug().Uint64("min", min).Uint64("max", max).Msg("received block response")
 
-	min, max := uint64(math.MaxUint64), uint64(0)
 	for _, block := range res.Blocks {
-		if e.log.GetLevel() <= zerolog.DebugLevel {
-			if block.Header.Height > max {
-				max = block.Header.Height
-			}
-			if block.Header.Height < min {
-				min = block.Header.Height
-			}
-		}
 		if !e.core.HandleBlock(block.Header) {
 			e.log.Debug().Uint64("height", block.Header.Height).Msg("block handler rejected")
 			continue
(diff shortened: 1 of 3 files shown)


engine/common/synchronization/engine.go
```diff
@@ -303,6 +314,7 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockRe
 		}
 		e.comp.SubmitLocal(synced)
 	}
+	e.log.Debug().Uint64("min", min).Uint64("max", max).Msg("received block response")
```
zhangchiqing, line 317: We would only print this log _after_ all blocks have been handled, which makes the log ordering a bit confusing:
```
processing block at height 100
processing block at height 101
processing block at height 102
...
processing block at height 110
received block response min: 100, max: 110
```

Usually, the first block is the min block, and the last block is the max block, we could just log with that, if someone is cheating that's not a big deal, they only mess up our Debug log.

So I suggest to just print this log at the beginning of this function.

zhangchiqing (approved):
Looks good. minor suggestion to the logs

engine/common/synchronization/engine.go
m4ksio: hmmm, ok, this makes more sense. I was under an impression that order is not really guaranteed, but even if, it doesn't matter that much

engine/common/synchronization/engine.go
```diff
@@ -291,10 +291,14 @@ func (e *Engine) onSyncResponse(originID flow.Identifier, res *messages.SyncResp
 
 // onBlockResponse processes a response containing a specifically requested block.
 func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockResponse) {
-	e.log.Debug().Str("origin_id", originID.String()).Msg("received block response")
 	// process the blocks one by one
+	min := res.Blocks[0].Header.Height
```
zhangchiqing, line 295: Better to check if res.Blocks has at least one block, otherwise it's vulnerable to attack.

If there is no block at all, we could simply drop the message

m4ksio: Great spot once again!

smnzhu (dismissed):

smnzhu (approved):

engine/common/synchronization/engine.go
```diff
@@ -291,10 +291,19 @@ func (e *Engine) onSyncResponse(originID flow.Identifier, res *messages.SyncResp
...
 	// process the blocks one by one
+	if len(res.Blocks) == 0 {
+		e.log.Debug().Msg("received empty block response")
+		return
+	}
+
+	min := res.Blocks[0].Header.Height
+	max := res.Blocks[len(res.Blocks)-1].Header.Height
+	e.log.Debug().Uint64("min", min).Uint64("max", max).Msg("received block response")
+
```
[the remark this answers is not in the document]
m4ksio: As we discussed, renamed fields to `first` and `last` to better reflect that. Should be a good tradeoff for iterating over all blocks

Kay-Zee (approved):

PR #1948, opened 2022-02-03
Fix all integration tests

zhangchiqing:
This PR resolved a few flakiness in the integration tests.

1. Sometimes the tests fail to start the container due to fail to bind the port, I found bind a lot of ports to expose metrics, which isn't needed for integration tests, but only for debugging purpose. Removing these ports binding, reduced the chance of running into this issue
2. Consensus nodes need to start up the same time, if a consensus node starts up late, and missed the proposal from the leader who started earlier, then the tests will wait until timeout. I added hotstuff-startup-time to allow all consensus nodes to wait for a few seconds after startup, so that we could ensure the cluster are all ready, because running the test case. This helped resolve some flakiness
3. When test case fails the log was reordered, and also included logs for tests that were succeeded, very hard to find the root cause. I refactored all the integration tests to split them into different folders. One integration per folder, so that if one fails, it will stop and print all logs about the failed test.
4. Logs are re-organized so that it's clear to see the start and end of preparing network, running test case, as well as stopping containers.


consensus/hotstuff/integration/integration_test.go
```diff
@@ -40,6 +40,7 @@ func TestSingleInstance(t *testing.T) {
 }
 
 func TestThreeInstances(t *testing.T) {
+	t.Skip("flakey")
```
jordanschalm, line 43: We should use the flakey test skipping helper: https://github.com/onflow/flow-go/blob/cf38eb7aaf3a313ebe9337bd85b4a6582e852210/utils/unittest/unittest.go#L50

integration/testnet/network.go
```diff
@@ -174,9 +176,39 @@ func (net *FlowNetwork) Result() *flow.ExecutionResult {
...
+	net.log.Info().Msgf("starting flow network %v with docker containers",
+		net.config.Name)
+
+	// print all existing containers before starting our containers, Useful to debug the issue
+	// that the tests fail due to "port is already allocated"
+
+	cli, err := dockerclient.NewClientWithOpts(dockerclient.FromEnv, dockerclient.WithAPIVersionNegotiation())
+	if err != nil {
+		panic(err)
+	}
```
jordanschalm, line 188: Any reason not to use `require.NoError` here?

utils/unittest/logging.go
```diff
@@ -19,13 +19,13 @@ func LogVerbose() {
 // Logger returns a zerolog
 // use -vv flag to print debugging logs for tests
 func Logger() zerolog.Logger {
-	writer := ioutil.Discard
+	return LoggerWithLevel(zerolog.DebugLevel)
```
jordanschalm, line 22: ⚠️  This is pretty significantly changing the behaviour of `Logger` for all other usages. Previously it would only log when you set the `verbose` flag, now it will log to the test output (even if the test is passing - which isn't what we want for more cases). 

I think we should keep the default behaviour, and add a method which allows specifying the output directly for the integration tests, if that's the behaviour we want there.

jordanschalm (commented):
Thank you for tackling this. Generally good with these changes, just a couple suggestions.

integration/testnet/network.go
```diff
@@ -174,9 +176,39 @@ func (net *FlowNetwork) Result() *flow.ExecutionResult {
...
+
+	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{})
+	if err != nil {
+		panic(err)
+	}
+
+	t := net.t
+	t.Logf("%v (%v) before starting flow network, found %v docker container", time.Now().UTC(), t.Name(), len(containers))
```
AlexHentschel, line 196:
```suggestion
	t.Logf("%v (%v) before starting flow network, found %d docker containers", time.Now().UTC(), t.Name(), len(containers))
```

integration/testnet/network.go
```diff
@@ -174,9 +176,39 @@ func (net *FlowNetwork) Result() *flow.ExecutionResult {
...
+
+	for _, container := range containers {
+		t.Logf("%v (%v) before starting flow network, found docker container: %v %v", time.Now().UTC(), t.Name(), container.Names, container.Ports)
```
AlexHentschel, line 199: I think we generally use `:` to indicate the start of wrapped error message. I personally prefer to avoid other usages 
```suggestion
		t.Logf("%v (%v) before starting flow network, found docker container %v with ports %v", time.Now().UTC(), t.Name(), container.Names, container.Ports)
```

integration/testnet/network.go
```diff
@@ -174,9 +176,39 @@ func (net *FlowNetwork) Result() *flow.ExecutionResult {
...
+	}
+
 	net.suite.Start(ctx)
+
+	containers, err = cli.ContainerList(ctx, types.ContainerListOptions{})
+	if err != nil {
+		panic(err)
+	}
+	t.Logf("%v (%v after starting flow network, found %v docker container", time.Now().UTC(), t.Name(), len(containers))
```
AlexHentschel, line 208: missing bracket (?)
```suggestion
	t.Logf("%v (%v) after starting flow network, found %d docker containers", time.Now().UTC(), t.Name(), len(containers))
```

integration/testnet/network.go
```diff
@@ -557,7 +595,7 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
 		}
 	}
 	require.GreaterOrEqualf(t, len(accessNodeIDS), DefaultMinimumNumOfAccessNodeIDS,
-		"at-least %d access node that is not a ghost must be configured for test suite", DefaultMinimumNumOfAccessNodeIDS)
+		fmt.Sprintf("at-least %d access node that is not a ghost must be configured for test suite", DefaultMinimumNumOfAccessNodeIDS))
```
AlexHentschel, line 598:
```suggestion
		fmt.Sprintf("at least %d access nodes that are not a ghost must be configured for test suite", DefaultMinimumNumOfAccessNodeIDS))
```

integration/testnet/network.go
```diff
@@ -591,9 +629,9 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
 		flowNetwork.addConsensusFollower(t, rootProtocolSnapshotPath, followerConf, confs)
 	}
 
-	flowNetwork.PrintMetricsPorts()
+	// flowNetwork.PrintMetricsPorts()
```
AlexHentschel, line 633: can we remove this?
```suggestion
```

integration/testnet/network.go
```diff
@@ -731,16 +771,18 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
+			// containerMetricsPort := "8080/tcp"
 
-			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
-			nodeContainer.Ports[ColNodeMetricsPort] = hostMetricsPort
-			net.AccessPorts[ColNodeMetricsPort] = hostMetricsPort
-			net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
+			// nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			// nodeContainer.Ports[ColNodeMetricsPort] = hostMetricsPort
+			// net.AccessPorts[ColNodeMetricsPort] = hostMetricsPort
+			// net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
```
AlexHentschel, line 780: is the commented-out code still needed? If so, a comment would be great indicating when, e.g. 

```golang
\\ use the following lines for testing <something>
```

integration/testnet/network.go
```diff
@@ -754,20 +796,20 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
 
-			hostMetricsPort := testingdock.RandomPort(t)
-			containerMetricsPort := "8080/tcp"
+			// hostMetricsPort := testingdock.RandomPort(t)
+			// containerMetricsPort := "8080/tcp"
 
-			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
-			net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
+			// nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			// net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
```
AlexHentschel, line 803: still needed?

integration/testnet/network.go
```diff
@@ -754,20 +796,20 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
 			nodeContainer.AddFlag("rpc-addr", fmt.Sprintf("%s:9000", nodeContainer.Name()))
 
 			nodeContainer.Ports[ExeNodeAPIPort] = hostPort
 			nodeContainer.opts.HealthCheck = testingdock.HealthCheckCustom(healthcheckExecutionGRPC(hostPort))
 			net.AccessPorts[ExeNodeAPIPort] = hostPort
 
-			nodeContainer.Ports[ExeNodeMetricsPort] = hostMetricsPort
-			net.AccessPorts[ExeNodeMetricsPort] = hostMetricsPort
+			// nodeContainer.Ports[ExeNodeMetricsPort] = hostMetricsPort
+			// net.AccessPorts[ExeNodeMetricsPort] = hostMetricsPort
```
AlexHentschel, line 812: still needed?

integration/testnet/network.go
```diff
@@ -823,36 +865,31 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
-			containerMetricsPort := "8080/tcp"
-
-			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
-			nodeContainer.Ports[AccessNodeMetricsPort] = hostMetricsPort
-			net.AccessPorts[AccessNodeMetricsPort] = hostMetricsPort
-			net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
+			// nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			// nodeContainer.Ports[AccessNodeMetricsPort] = hostMetricsPort
+			// net.AccessPorts[AccessNodeMetricsPort] = hostMetricsPort
+			// net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
```
AlexHentschel, line 871: still needed?

integration/testnet/network.go
```diff
@@ -823,36 +865,31 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
+			nodeContainer.AddFlag("hotstuff-startup-time", hotstuffStartupTime)
 
-			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
-			nodeContainer.Ports[ConNodeMetricsPort] = hostMetricsPort
-			net.AccessPorts[ConNodeMetricsPort] = hostMetricsPort
-			net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
+			// nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			// nodeContainer.Ports[ConNodeMetricsPort] = hostMetricsPort
+			// net.AccessPorts[ConNodeMetricsPort] = hostMetricsPort
+			// net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
```
AlexHentschel, line 883: still needed?

integration/testnet/network.go
```diff
@@ -823,36 +865,31 @@ func (net *FlowNetwork) AddNode(t *testing.T, bootstrapDir string, nodeConf Cont
...
-			containerMetricsPort := "8080/tcp"
 
-			nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
-			nodeContainer.Ports[VerNodeMetricsPort] = hostMetricsPort
-			net.AccessPorts[VerNodeMetricsPort] = hostMetricsPort
-			net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
+			// nodeContainer.bindPort(hostMetricsPort, containerMetricsPort)
+			// nodeContainer.Ports[VerNodeMetricsPort] = hostMetricsPort
+			// net.AccessPorts[VerNodeMetricsPort] = hostMetricsPort
+			// net.MetricsPortsByContainerName[nodeContainer.Name()] = hostMetricsPort
```
AlexHentschel, line 892: still needed?

AlexHentschel (approved):
LGTM 
(just keep in mind that I am not very familiar with the integrations test frameworks ... edge-cases might easily slip through in my review unnoticed 😅 ).

integration/testnet/network.go
zhangchiqing: yeah, @kc1116 still need this for debugging purpose

integration/tests/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+How do we know which test case is failing if the error isn't clear about it?
+
+Using log.
+
+Usually when tests are passing, no log will be printed. But when the tests is failing, _all_ logs for the failed tests in that package will be printed.
+
+In order to use log, they must be printed in a way that is easy to reason about.
+
+If logs are printed in pairs, one at the entrance and the other at the exit of a test case, then if we find both logs, it means the test case has passed; if we only find the entrance log line, but no log for the exit, it means the test case must have failed.
```
kc1116, line 15:
```suggestion
It is useful if logs are printed in pairs at the start and at the end of a test case.  Then when debugging a test case you can search for the start and end log for that test case, any logs in between will be for that test case only.  If we only find the entrance log line, but no log for the exit, it means the test case must have failed.
```

integration/testnet/network.go
kc1116: It is useful when you need to view metrics for a particular server. Can you refactor this into a separate func that can be used to expose metrics on a single container in the network? It is a nice to have.

kc1116 (approved):
Some small suggestions, overall looks good.

PR #1949, opened 2022-02-04
[Integration Testing] Set default value for peer update interval

yhassanzadeh13:
This PR identifies and fixes one of the root causes of message delivery failure on integration tests. The reason is the scale of the system in integration tests is much smaller than the production system. Hence, the mathematics behind the topology is not progressing at the same speed as the production networks. 

To give some contexts: 
The `pubsub` topology is shaped through `PeerManager` component of each node manually trying to connect to the fanout set determined by its `Topology` component. This operation is known as _peer update_. 

In the current codebase, this peer update is happening only as the result of following event:
a. At the PeerManager upon the [start-up of the node](https://github.com/onflow/flow-go/blob/master/network/p2p/peerManager.go#L90).
b. At the Middleware upon a [change in the identity list of the system](https://github.com/onflow/flow-go/blob/d57e3e383258a0eafbfe01d65f2cbce75ce8e18e/network/p2p/middleware.go#L548).
c. At the Middleware upon un-[subscribing to a topic](https://github.com/onflow/flow-go/blob/master/network/p2p/middleware.go#L446).
d. At the Middleware upon [subscribing to a topic](https://github.com/onflow/flow-go/blob/master/network/p2p/middleware.go#L432).
e. At the Middleware on [regular intervals](https://github.com/onflow/flow-go/blob/master/network/p2p/peerManager.go#L97) determined by the peerUpdateInterval which is currently 10 minutes.

In production, we rely on the events (a) and (b) that are collectively executed by a large scale of the nodes, i.e., at the sparking time when we run a large number of nodes almost at once, each node connects to its topology peers and running _enough (i.e., > 100 nodes)_ instance of nodes simultaneously results in a connected topology. This event is resonated once more for the startup of each engine of the node (even The connected topology then keeps updated every 10 minutes and also upon each epoch transition. 

However, at the scale of our integration tests with normally fewer than 10 nodes, events (a) and (b) are not executed at the scale that yields a connected topology. This results in most of the nodes having a fanout set of empty on their channel topology (sample evidence log): 

```
{"level":"info","node_role":"ghost","node_id":"a3d80bb53792afa71a39b8252573222754303012eb18b2480ef2755a01d309a3","channel":"push-guarantees","topic_fanout":"[]","time":"2022-02-02T00:17:23Z","message":"channel fanout updated"}
```

Also, our tests are much shorter than 10 minutes hence event (e) is not getting any chance of helping in. Which results in a poorly connected topology, and causes flakey integration testing behavior due to an unreliable level of message delivery. 

As a short-term solution, we set the default flag of peer update intervals for _integration tests_ to 1 second hence making sure that the connection to fanouts is established very quickly. Below is a sample evidence log that shows the connections in the same `push-guarantees` channel are established very quickly when switching to one second update intervals:
```
cc92","channel":"push-guarantees","topic_fanout":"[QmUWRdiYfuwuNP7hKgAm7jzAs8wJ585YFnRFgLfjTraUph QmcAfZCNeVW4EpGJTNEYMSdMLETyWNeK9DV5R4Uv369Rsz QmbTZdQNTuTVmSYZtRcQXXa6VdzV17TcxU2R7L88wzoD1x QmVbAeG8qtFubgeS9EAjx8RTqDpKhK7U3qnCf7Q5mpUuV7 QmYJQKPCdNAYhZZH4k3JmnvrirDCw8WJWGLEcs3Hspv5NK]","time":"2022-02-02T01:03:03Z","message":"channel fanout updated"}

```

Moreover, as a longer-term fix, we are working on https://github.com/dapperlabs/flow-go/issues/6134 to decouple the topology updating procedure into different components hence letting distinct components of `pubsub` be updated at distinct rates. In that way, we can have the updating procedure of `pubsub` connections being done very frequently every few seconds even in the production environment without interfering with other components.


vishalchangrani (approved):

integration/testnet/network.go
```diff
@@ -576,11 +576,16 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig) *FlowNetwork {
 
 		// if node is of LN/SN role type add additional flags to node container for secure GRPC connection
 		// this is required otherwise collection node will fail to startup
+		nodeContainer := flowNetwork.Containers[nodeConf.ContainerName]
+
+		// testnet is running at a significantly smaller scale than a production network, hence to
+		// hold the mathematics behind topology connectivity valid, we need a faster connection update rate than the default value:
+		nodeContainer.AddFlag("peerupdate-interval", "1s")
```
jordanschalm, line 583: @yhassanzadeh13 Could you add the flag here instead: https://github.com/onflow/flow-go/blob/8f5f429ff969444c527e6a73b1474a37f6c7f18f/integration/testnet/network.go#L671.

In `AddNode` is the standard place where we add all config flags. The only reason `access-node-ids` is added here instead, is because we need two know all AN addresses to set the flag, and this isn't guaranteed to be known in `AddNode`.

yhassanzadeh13: Addressed in https://github.com/onflow/flow-go/pull/1949/commits/994d7b001a57b6e20598431e2c30c66b1f7db1bc

jordanschalm (approved):

PR #1950, opened 2022-02-04
DKG Messaging Improvements

jordanschalm:
This PR addresses several problems with DKG private and broadcast messages from [Benchnet Testing Consensus Node Dynamic Join](https://www.notion.so/dapperlabs/Test-18-Dynamic-Startup-Single-Consensus-Node-2794ba8584124f8db415469ade46fdfa).

## Changes
* Adjust backoff configuration for DKG private messages
  * Previously, we would retry DKG private messages for about 30s total. If they are not delivered in this time, they would never be delivered. Nodes joining the DKG >30s after the start would not receive these messages and would need to broadcast a complaint to recover in-protocol.
  * In the absence of a way to re-attempt delivery for the duration of phase 1, or allow late nodes to request messages, we extend the duration of attempted submission (to ~8m)
* Add backoff config to `Broker` to avoid sleeping in tests
* Log number of attempts and transaction IDs
* Consistent naming for backoff middlewares (don't rename intermediary variables, use the same name as the type `Backoff`)
* Use the `unit` context rather than background, where possible


module/dkg/broker.go
```diff
@@ -120,44 +141,48 @@ func (b *Broker) PrivateSend(dest int, data []byte) {
...
+		// NOTE: We're counting the number of times the underlying DKG requested
+		// a broadcast so that we can detect an unhappy path (any time there is
+		// more than 1 broadcast message per DKG) Thus incrementing broadcasts
+		// before we perform the broadcasts is okay.
 		b.broadcastLock.Lock()
 		if b.broadcasts > 0 {
-			// The Warn log is used by the integration tests to check if this method
-			// is called more than once within one epoch.
+			// The warn-level log is used by the integration tests to check if this
+			// method is called more than once within one epoch (unhappy path).
```
kc1116, line 152:
```suggestion
			// func is called more than once within one epoch (unhappy path).
```

module/dkg/client.go
```diff
@@ -50,6 +50,43 @@ func NewClient(
 	}
 }
 
+// ReadBroadcast reads the broadcast messages from the smart contract.
+// Messages are returned in the order in which they were broadcast (received
+// and stored in the smart contract)
```
kc1116, line 55:
```suggestion
// Messages are returned in the order in which they were received
// and stored in the smart contract
```

kc1116 (approved):

module/dkg/broker.go
```diff
@@ -166,27 +191,78 @@ func (b *Broker) Broadcast(data []byte) {
...
+	// the nil slice to a slice of nil keys before submission.
+	//
+	// In general, if pubKeys does not have one key per participant, we cannot submit
+	// a valid result - therefore we submit a nil vector (indicating that we have
+	// completed the process, but we know that we don't have a valid result).
+	if len(pubKeys) != len(b.committee) {
+		b.log.Warn().Msgf("submitting dkg result with incomplete key vector (len=%d, expected=%d)", len(pubKeys), len(b.committee))
+		// create a key vector with one nil entry for each committee member
+		pubKeys = make([]crypto.PublicKey, len(b.committee))
+	}
```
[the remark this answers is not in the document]
jordanschalm: The smart contract [enforces that submissions are the expected length](https://github.com/onflow/flow-core-contracts/blob/master/contracts/epochs/FlowDKG.cdc#L154), so the transaction would revert.

The desired format for the DKG smart contract to submit a known invalid submission is a key vector with all nil keys. 

The specific case this new condition was added to deal with, is `SubmitResult` is called with `nil` values for both `groupKey` and `publicKeys`, resulting a key vector with a single `nil` key in it. I generalized that to any time the given key vector length isn't the expected value.

jordanschalm:
@tarakby @kc1116 Could I get one more pass on this? Added one more commit: https://github.com/onflow/flow-go/pull/1950/commits/e172b94d9890adca5da95c7dae9e6300fab7068b, and would like to confirm this line of thinking is OK: https://github.com/onflow/flow-go/pull/1950#issuecomment-1033102014

module/dkg/broker.go
```diff
@@ -166,27 +191,78 @@ func (b *Broker) Broadcast(data []byte) {
 			b.updateLastSuccessfulClient(clientIndex)
 			return nil
 		})
+		b.broadcastLock.Unlock()
 
-		// Various network can conditions can result in errors while broadcasting DKG messages,
-		// because failure to send an individual DKG message doesn't necessarily result in local or global DKG failure
+		// Various network conditions can result in errors while broadcasting DKG messages.
+		// Because the overall DKG is resilient to individual message failures,
```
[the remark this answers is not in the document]
tarakby: I think this still says that DKG is resilient to message failures while it's not.

tarakby:
> No. It is possible that that private message message would arrive after the end of phase 1 currently. It depends on the DKG phase length configurations.

What I meant is that it would be great to have a test for the configuration: `max_private_message_delay < duration phase1` (since private messages are only sent in phase 1). I guess it's not possible because one duration is defined in views and the other in seconds? 
The condition is not sufficient to say everything will go fine, but it is necessary. 

> But I don't think we can't guarantee this regardless, because different nodes may have different views of the current state. I might think it's phase 1 when I send a message, but another node might have just ended phase 1 when it receives that message. My understanding is that messages intended for delivery in a previous phase would be ignored by the recipient.

Yes true, and we're mitigating that to happen at the end of the phase by making phase1 long enough so that private messages are sent way before the end of it. 
However there is the same issue at the beginning of phase1 (one node sees DKG started and sends a message, the other node didn't start DKG yet and drops the message). Did you see that happening before? if yes, it might me better to insert a local delay between starting phase1 and sending the private message.

module/dkg/broker.go
```diff
@@ -166,27 +192,78 @@ func (b *Broker) Broadcast(data []byte) {
...
+	b.log.Info().Msgf("dkg result submitted successfully on attempt %d", attempts)
+	return nil
 }
 
 // Disqualify flags that a node is misbehaving and got disqualified
 func (b *Broker) Disqualify(node int, log string) {
-	// The Warn log is used by the integration tests to check if this method is
-	// called.
+	// The Warn log is used by the integration tests to check if this method is called.
 	b.log.Warn().Msgf("participant %d is disqualifying participant %d because: %s", b.myIndex, node, log)
```
zhangchiqing, line 261: how to know the node id by the index? is it worth to include the node ID in the log?

zhangchiqing (approved):

module/dkg/broker.go
jordanschalm: Changed to log both index and node ID: https://github.com/onflow/flow-go/pull/1950/commits/9c084fec9bfedfd7cc05df8ae06f7dca86bb8b5a

jordanschalm:
@tarakby Added the config test we discussed in https://github.com/onflow/flow-go/pull/1950/commits/1870a10c478ba72f107f6eb8c0504580b5fb3637

module/dkg/broker_test.go
```diff
@@ -37,10 +38,28 @@ func initCommittee(n int) (identities flow.IdentityList, locals []module.Local)
...
+	phase1Views := 2000       // present configuration for all networks
+	viewsPerSecMainnet := 0.8 // observation from Feb 16 2022
+	phase1LenMainnet := time.Duration(float64(phase1Views)/viewsPerSecMainnet) * time.Second
+
+	conf := DefaultBrokerConfig()
+	// cumulative max delay is sum of delays
+	// 1+2+4+8...+2^n = (2^{n+1}-1)
+	maxDelay := conf.RetryInitialWait<<(conf.PublishMaxRetries+1) - time.Second
+	t.Run("all retries occur within phase 1", func(t *testing.T) {
+		assert.Less(t, maxDelay, phase1LenMainnet)
```
tarakby, line 57: 👌🏼

module/dkg/broker_test.go
```diff
@@ -37,10 +38,28 @@ func initCommittee(n int) (identities flow.IdentityList, locals []module.Local)
...
+	})
+
+	t.Run("last possible retry is after mid-point of phase 1", func(t *testing.T) {
+		assert.Greater(t, maxDelay, phase1LenMainnet/2)
```
tarakby, line 61: 👌🏼 👌🏼

tarakby (approved):

PR #1952, opened 2022-02-04
[Execution] Adding a pre and post script execution log

ramtinms:
Adding a pre/post-script execution logs helps to debug in case a bad script is sent.

### DIFF
--- a/engine/execution/computation/manager.go
+++ b/engine/execution/computation/manager.go
@@ -133,6 +133,7 @@ func (e *Manager) ExecuteScript(code []byte, arguments [][]byte, blockHeader *fl
 
 	programs := e.getChildProgramsOrEmpty(blockHeader.ID())
 
+	e.log.Info().Hex("script_hex", code).Msg("script is sent for execution")
 	err := func() (err error) {
 
 		start := time.Now()
--- a/engine/execution/computation/manager.go
+++ b/engine/execution/computation/manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"math/rand"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


engine/execution/computation/manager.go
```diff
@@ -127,6 +128,13 @@ func (e *Manager) ExecuteScript(code []byte, arguments [][]byte, blockHeader *fl
 
 	startedAt := time.Now()
 
+	trackerID := rand.Uint32()
```
jordanschalm, line 131: Could you add a comment here with some context on why we're doing this, and either a TODO to remove this when it is no longer needed, or maybe make it configurable by flag later on.

ramtinms: added.

engine/execution/computation/manager.go
```diff
@@ -127,6 +128,16 @@ func (e *Manager) ExecuteScript(code []byte, arguments [][]byte, blockHeader *fl
 
 	startedAt := time.Now()
 
+	// allocate a random ID to be able to track this script when its done,
+	// scripts might not be unique so we use this extra tracker to follow their logs
+	// TODO: this is a temporary measure, we could remove this in the future
+	trackerID := rand.Uint32()
+	e.log.Info().Hex("script_hex", code).Uint32("trackerID", trackerID).Msg("script is sent for execution")
```
janezpodhostnik, line 135: Info level seems a bit high for this as this might produce a huge amount of logs in a short amount of time.

janezpodhostnik (approved):
Left a comment/question, otherwise it lgtm.

jordanschalm (approved):
🎸

engine/execution/computation/manager.go
ramtinms: yeah but thats the goal for now

ramtinms:
bors merge

PR #1956, opened 2022-02-06
Crash Access Node if disk is full

koko1123:
Fixes https://github.com/dapperlabs/flow-go/issues/5340
- Simple panic when we see the Storage return an error
- unit test included


smnzhu:
I've commented on the attached issue, I'm not sure if this is the right approach.

smnzhu:
This is good, I really like that we've moved the disk error check to storage level

storage/badger/operation/modifiers.go
```diff
@@ -42,3 +44,15 @@ func RetryOnConflictTx(db *badger.DB, action func(*badger.DB, func(*transaction.
 		return err
 	}
 }
+
+// TerminateOnFullDisk helper function to crash node if write failed because disk is full
+func TerminateOnFullDisk(err error) error {
+	// using panic so any deferred functions can still execute
+	// string check on Badger's or other Wraps on syscall disk error, instead of Unwrap or Is.
```
smnzhu, line 51: `Is` should match any error in the chain, so I don't think you need to compare strings here: https://pkg.go.dev/errors#Is

smnzhu:
This might be okay for now, but I wonder if we may need some more flexibility than hardcoding this behavior for every call.

The ideal solution in my mind is that we define a sentinel error for the "full disk" situation, which would get returned from each method. This error can then be propagated up to parent components via `SignalerContext`, until it reaches the top level handler which can decide how to handle it.

storage/badger/blocks.go
```diff
@@ -65,7 +65,7 @@ func (b *Blocks) retrieveTx(blockID flow.Identifier) func(*badger.Txn) (*flow.Bl
 
 // Store ...
 func (b *Blocks) Store(block *flow.Block) error {
-	return operation.RetryOnConflictTx(b.db, transaction.Update, b.StoreTx(block))
+	return operation.TerminateOnFullDisk(operation.RetryOnConflictTx(b.db, transaction.Update, b.StoreTx(block)))
```
jordanschalm, line 68: Given we want all node roles to have the same behaviour in this case, I'd suggest we move the check for the out-of-disk error to a lower level. Currently, if we add a new storage method or resource, we need to remember to wrap insertion operations in `TerminateOnFullDisk`. If we put the logic in the common storage insertion methods, all current and future insertions will have the same behaviour. 

`batchInsert`, `insert`, and `update` in [this file](https://github.com/onflow/flow-go/blob/add2b9d79ace88740e9a4e3c1d19f4d2af36aa19/storage/badger/operation/common.go#L20) are the three base-level methods which can insert data to the Badger DB.

jordanschalm: > this is the implementation of the "light touch" idea

The intention there was to use the "light touch" approach only if we want different behaviour in the case of out-of-disk, depending on the caller. But to always have the same behaviour to handle out-of-disk, it's quite a bit simpler to put the check in those three lower-level storage methods. 

> also note that some disk writes for transactions also seem to happen in [this other file](https://github.com/onflow/flow-go/blob/710e4bf266d53882fa407374925ba23d9846655c/storage/badger/transaction/tx.go)

I think this file should be OK without changes. This file is creating and committing Badger transactions, but not directly inserting data itself. That would happen within the [functional argument](https://github.com/onflow/flow-go/blob/710e4bf266d53882fa407374925ba23d9846655c/storage/badger/transaction/tx.go#L28), and ultimately would be using those 3 insertion methods mentioned.

storage/badger/operation/common.go
```diff
@@ -386,3 +387,11 @@ func Fail(err error) func(*badger.Txn) error {
 		return err
 	}
 }
+
+// terminateOnFullDisk panics to crash the node if an insert / batchInsert / update fail due to a full disk
+func terminateOnFullDisk(err error) error {
+	if err != nil && errors.Is(err, syscall.ENOSPC) {
+		panic("disk full, terminating node...")
```
jordanschalm, line 394:
```suggestion
		panic(fmt.Sprintf("disk full, terminating node: %s", err.Error())
```
include full error text

jordanschalm (approved):
Had one final suggestion about including the error text in the panic, otherwise it looks good to me. Thank you!

storage/badger/operation/common.go
```diff
@@ -76,7 +77,7 @@ func insert(key []byte, entity interface{}) func(*badger.Txn) error {
 		}
 
 		// persist the entity data into the DB
-		err = tx.Set(key, val)
+		err = terminateOnFullDisk(tx.Set(key, val))
```
peterargue, line 80: I don't think `tx.Set` actually does the write to disk. That's done by `tx.Commit`. It's worth testing this out to make sure

koko1123: Absolutely right! I did more digging and we have a dual-use problem:
- some parts of the code like [StoreLight in collection](https://github.com/onflow/flow-go/blob/710e4bf266d53882fa407374925ba23d9846655c/storage/badger/collections.go#L28-L35) actually use badger's inbuilt `Update` that has a similar flow to our own [tx wrapper's update](https://github.com/onflow/flow-go/blob/710e4bf266d53882fa407374925ba23d9846655c/storage/badger/transaction/tx.go#L26-L45) 
- so the question is: do we go back to the original model of applying this to `operation` and wrapping that call everywhere (not a fan of this personally) or do we make this disk crash logic only apply to `transaction.Update` calls?
- one approach is to do both, keep `terminateOnFullDisk ` in `operation` and wrap where `db.Update` is used, and wrap the `Commit()` in `transaction` with it as well

koko1123: Implemented the above approach without wrapping `db.Update` calls, minimum viable refactor

utils/io/write_test.go
```diff
@@ -0,0 +1,34 @@
...
+	})
+	t.Run("benign non disk related error should return the error", func(t *testing.T) {
+		benignError := errors.New("benign error")
+		result := TerminateOnFullDisk(benignError)
+		require.ErrorIs(t, result, benignError)
+	})
+	// sad path
+	t.Run("panic on full disk", func(t *testing.T) {
+		// imitate badgerDB error wrapping
+		badgerDiskFullError := syscall.ENOSPC
```
jordanschalm, line 25: I'm assuming Badger wraps the error - in which case it's better to mimic this in the test case. Though, we should confirm exactly what error Badger is throwing in this case, like Peter mentioned. 
```suggestion
		badgerDiskFullError := fmt.Errorf("this is intended to mimic an internal Badger error wrapping an out-of-disk system error: %w", syscall.ENOSPC)
```

utils/io/write.go
```diff
@@ -43,3 +45,10 @@ func WriteJSON(path string, data interface{}) error {
 
 	return WriteFile(path, bz)
 }
+
+func TerminateOnFullDisk(err error) error {
+	if err != nil && errors.Is(err, syscall.ENOSPC) {
```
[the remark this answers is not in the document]
koko1123: yes! https://github.com/dgraph-io/badger/blob/156819ccb106bbeb207e985f561780e2929344bc/value.go#L1431
follows the UNIX syscall `write` here: https://github.com/dgraph-io/badger/blob/156819ccb106bbeb207e985f561780e2929344bc/value.go#L1459 that returns `ENOSPC          = Errno(0x1c)` when the disk is full

jordanschalm, line 50:
```suggestion
// TerminateOnFullDisk panics if the input error is (or wraps) the system "out of disk
// space" error. It's a no-op for any other error, or nil. 
func TerminateOnFullDisk(err error) error {
	if err != nil && errors.Is(err, syscall.ENOSPC) {
```

peterargue (approved):

PR #1966, opened 2022-02-08
fix flag validation for access nodes

smnzhu:
Because access node code is different from main scaffold code, it's missing the flag validation. Hence why the validation code added to check for public network bind address didn't run.


cmd/node_builder.go
```diff
@@ -40,7 +40,7 @@ type NodeBuilder interface {
 	ExtraFlags(f func(*pflag.FlagSet)) NodeBuilder
 
 	// ParseAndPrintFlags parses all the command line arguments
```
peterargue, line 42: we should update this comment to indicate it also does validation

peterargue (approved):

vishalchangrani (approved):

smnzhu:
bors merge

PR #1968, opened 2022-02-08
Convert `fmv.VirtualMachine` to an interface

SupunS:
This is some initial refactoring required to implement the fvm process isolation.

The summary of the changes are:
- Converted `fvm.VirtuslMachine` to an interface, and renamed the implementation.
  - Hide the cadence `Runtime` field, and expose its operations through interface methods of the `fvm.VirtuslMachine`
- Similarly, converted `programs.Programs` to an interface, and renamed the implementation.
- This was done to make it easy to split from those two interfaces, by hiding the implementation.
  - Can easily plug-in a proxy in between, without affecting the usages.
  - i.e: `blockComputer <--> fvm.VM` can be replaced by `blockComputer <--> proxy over IPC <--> fvm.VM`


janezpodhostnik (approved):
lgtm!

PR #1969, opened 2022-02-08
perf: Reorganize TransactionBody code to avoid unnecessary heap usage

jwinkler2083233:
This change moves the code for 'payloadCanonicalForm' into the methods where it's used.
Results show a significant improvement in heap usage statistics.

With a good compiler that has a copy elision feature, like g++, this would not be necessary.

see https://github.com/dapperlabs/flow-go/issues/6184


model/flow/transaction.go
```diff
@@ -53,14 +53,83 @@ func NewTransactionBody() *TransactionBody {
...
+	authorizers := make([][]byte, len(tb.Authorizers))
+	for i, auth := range tb.Authorizers {
+		authorizers[i] = auth.Bytes()
+	}
+
+	// this comment below is to enforce that this struct
+	// definition is exactly sync'd with any others.  The
+	// constraint is enforced with a test.
+	///%[[ EQUALITY-PERF-CONSTRAINT {transaction-body-payload}
+	var payload = struct {
```
Kay-Zee, line 65: since we're now reusing this struct in multiple places, could we just not use it as an anonymous struct, and just use it as a private struct in this file? Does that still give us the same gains?

jwinkler2083233:
Here's an updated memory profile:

```
(pprof) Showing nodes accounting for 2062890282, 85.03% of 2426044532 total
Dropped 1775 nodes (cum <= 12130222)
Showing top 100 nodes out of 170
      flat  flat%   sum%        cum   cum%
 943985866 38.91% 38.91% 1738103850 71.64%  github.com/onflow/flow-go/model/flow.TransactionBody.Fingerprint
 698624358 28.80% 67.71%  698624358 28.80%  github.com/ethereum/go-ethereum/rlp.(*encbuf).list (inline)
 101422192  4.18% 71.89%  101422192  4.18%  github.com/ethereum/go-ethereum/rlp.(*encbuf).toBytes
  98286480  4.05% 75.94% 1831917090 75.51%  github.com/onflow/flow-go/model/flow.TransactionBody.ID (partial-inline)
```

The number of heap allocations dropped from approximately 6 billion, to approximately 4.3 billion.

jwinkler2083233:
This is the result I was looking for.
```
(pprof) Showing nodes accounting for 2475768298, 86.55% of 2860374824 total
Dropped 1707 nodes (cum <= 14301874)
Showing top 100 nodes out of 166
      flat  flat%   sum%        cum   cum%
1149939605 40.20% 40.20% 2229329748 77.94%  github.com/onflow/flow-go/model/flow.TransactionBody.Fingerprint
 948201532 33.15% 73.35%  948201532 33.15%  github.com/ethereum/go-ethereum/rlp.(*encbuf).list (inline)
 137764809  4.82% 78.17%  137764809  4.82%  github.com/ethereum/go-ethereum/rlp.(*encbuf).toBytes
 ```

File: model/flow/transaction.go (hunk 1 above)
Comment (jwinkler2083233) [reply model/flow/transaction.go]: done

File: model/flow/identifier.go [hunk 2]
```diff
@@ -120,6 +120,14 @@ func MakeID(entity interface{}) Identifier {
 	return id
 }
 
+func MakeIDFast(data *[]byte) Identifier {
+	//bs2 := binstat.EnterTimeVal(binstat.BinMakeID+".??lock.Hash", int64(len(data)))
+	var id Identifier
+	hash.ComputeSHA3_256((*[32]byte)(&id), *data)
+	//binstat.Leave(bs2)
```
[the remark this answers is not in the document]
jwinkler2083233: These comments are for enabling the binstat profiler..  I should leave them alone.

model/flow/identifier.go
```diff
@@ -121,6 +121,14 @@ func MakeID(entity interface{}) Identifier {
 	return id
 }
 
+func MakeIDFromBytes(data *[]byte) Identifier {
+	//bs2 := binstat.EnterTimeVal(binstat.BinMakeID+".??lock.Hash", int64(len(data)))
+	var id Identifier
+	hash.ComputeSHA3_256((*[32]byte)(&id), *data)
```
AlexHentschel, line 127: The signature of `ComputeSHA3_256` is
```golang
func ComputeSHA3_256(result *[HashLenSHA3_256]byte, data []byte) {
``` 

We could `HashLenSHA3_256` also here:
```suggestion
	hash.ComputeSHA3_256((*[HashLenSHA3_256]byte)(&id), *data)
```

model/flow/identifier.go
```diff
@@ -121,6 +121,14 @@ func MakeID(entity interface{}) Identifier {
 	return id
 }
 
+func MakeIDFromBytes(data *[]byte) Identifier {
```
AlexHentschel, line 124: Why do we have a pointer to the slice here? I would consider it better Go-style to use the slice directly:

```suggestion
func MakeIDFromBytes(data []byte) Identifier {
```

AlexHentschel (approved):
👍

PR #1971, opened 2022-02-08
[Execution State] Full search json dump

m4ksio:
Adds extra flag for full scan and in consequence ability to export any trie


cmd/util/cmd/export-json-execution-state/cmd.go
```diff
@@ -98,6 +102,92 @@ func ExportLedger(ledgerPath string, targetstate string, outputPath string) erro
...
+		if err != nil {
+			if !errors.Is(err, sentinel) {
+				return fmt.Errorf("cannot restore LedgerWAL: %w", err)
+			}
+		}
+
+		write = func(writer io.Writer) error {
+			mTrie, err := forest.GetTrie(rootState)
+			if err != nil {
+				return fmt.Errorf("cannot get trie")
```
zhangchiqing, line 163:
```suggestion
				return fmt.Errorf("cannot get trie: %w", err)
```

cmd/util/cmd/export-json-execution-state/cmd.go
```diff
@@ -98,6 +102,92 @@ func ExportLedger(ledgerPath string, targetstate string, outputPath string) erro
...
+				if err != nil {
+					return fmt.Errorf("adding rebuilt tries to forest failed: %w", err)
+				}
+
+				for _, trie := range rebuiltTries {
+					rootHash := trie.RootHash()
+					if rootState.Equals(rootHash) {
+						return sentinel
+					}
+				}
```
zhangchiqing, line 138: I wonder if rebuilding tries could take in the `rootState` for comparison, so that we could bail as soon as we found a trie that matches the root hash.

zhangchiqing (approved):
Looks good. 

Can we also add some logs in the whole process? 

Would be great to have logs like this along with timestamp
```
Looking for trie with hash <> in full search mode
Start loading the forest
Loaded forest
Start replaying WAL logs
Loading WAL Log(1/10)
Loading WAL Log(2/10)
Loading WAL Log(3/10)
...
Loading WAL Log(10/10)
Finish loading WAL Log
found trie 
Start dumping trie as json
```

ramtinms (approved):

PR #1972, opened 2022-02-08
Fix account reporter

zhangchiqing:
Fix https://github.com/dapperlabs/flow-go/issues/6140

Was running into a crash.

### DIFF
--- a/cmd/util/ledger/reporters/account_reporter.go
+++ b/cmd/util/ledger/reporters/account_reporter.go
@@ -78,31 +78,32 @@ func (r *AccountReporter) Report(payload []ledger.Payload) error {
 	addressIndexes := make(chan uint64, workerCount)
 	defer close(addressIndexes)
 
+	// create multiple workers to generate account data report concurrently
 	wg := &sync.WaitGroup{}
 	for i := 0; i < workerCount; i++ {
 		go func() {
 			adp := newAccountDataProcessor(r.Log, rwa, rwc, rwm, r.Chain, l)
-			for {
-				select {
-				case indx := <-addressIndexes:
-					adp.reportAccountData(indx)
-					wg.Done()
-				}
+			for indx := range addressIndexes {
+				adp.reportAccountData(indx)
+				wg.Done()
 			}
 		}()
 	}
 
 	addressCount := gen.AddressCount()
+	// produce jobs for workers to process
 	for i := uint64(1); i <= addressCount; i++ {
 		addressIndexes <- i
 
 		wg.Add(1)
+
 		err := progress.Add(1)
 		if err != nil {
 			panic(fmt.Errorf("progress.Add(1): %w", err))
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -92,7 +94,6 @@ func (r *AccountReporter) Report(payload []ledger.Payload) error {
 			panic(err)
 		}
 	}
-	close(addressIndexes)
```
zhangchiqing, line 95: I think L86 `go adp.reportAccountData(addressIndexes)` was still sending data to the channel, but we are closing the channel too early. 

It should only be closed after the `Wait()`.

I used `defer` to ensure it's closed in the end

janezpodhostnik: reportAccountData  gets addressIndexes as a receive only channel so I think that channel is not the problem.

This close must be before `wg.Wait()` because this close causes all the `go adp.reportAccountData` to finish processing and exit (calling `wg.Done()` as they do). with this change `wg.Wait()` will never finish.

I suspect the problem is coming from the `ReportWriter` channel closing too early. However going through the code  I am still not sure how, since the report writer channels are the last to be closed: 
```
defer rwa.Close()
defer rwc.Close()
defer rwm.Close()
```

turbolent:
cc @dsainati1

cmd/util/ledger/reporters/account_reporter.go
zhangchiqing: The Waitgroup management is a bit confusing to me, because we are waiting for the worker to stop working, but it might run into concurrency issue where all worker finished their work but immediately pick up new work, however the wait group will consider as finished. And close all the channels including the writers channel, however, there might be some process still running `reportAccountData` in a goroutine, and when it writes to the writer channel, it will crash.

A more reliable way is to add all indexes to the wait group, and wait until the workers have finished the same number of jobs as the total number of address.

cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -192,113 +200,109 @@ func newAccountDataProcessor(wg *sync.WaitGroup, logger zerolog.Logger, rwa Repo
 	}
 }
 
-func (c *balanceProcessor) reportAccountData(addressIndexes <-chan uint64) {
```
zhangchiqing, line 195: index of processing the whole channel, just processing one index

cmd/util/ledger/reporters/account_reporter.go
janezpodhostnik: That makes sense and looks much nicer!

janezpodhostnik (approved):

zhangchiqing:
Just tested, the report can be generated without error

dsainati1 (approved):
Thanks for fixing this!

zhangchiqing:
bors merge

PR #1975, opened 2022-02-09
Cadence upgrade

janezpodhostnik:
This is an upgrade of cadence, to this cadence version: https://github.com/onflow/flow-go/pull/1954

This uses flow-core-contracts from [here](https://github.com/onflow/flow-core-contracts/pull/278) due to some cadence breaking changes.

I added three wrapper methods around the crypto package, that the cadence runtime interface implementation needs.
- BLSVerifyPOP
- AggregateBLSSignatures
- AggregateBLSPublicKeys

There was a tests regarding restricted contract deployment that started failing. I suspect that the test needs different assertions, but I do not know yet why this did not show up earlier.


janezpodhostnik:
@tarakby @turbolent  could you check the logic in the three wrapper methods

- BLSVerifyPOP
- AggregateBLSSignatures
- AggregateBLSPublicKeys

I am not sure how to handle user errors here. Especially in `AggregateBLSPublicKeys` it is possible for each public key to have an input error... does there need to be a way to signal this up the stack?

engine/execution/computation/execution_verification_test.go
```diff
@@ -161,7 +161,7 @@ func Test_ExecutionMatchesVerification(t *testing.T) {
 		err = testutil.SignTransaction(addKeyTx, accountAddress, accountPrivKey, 0)
 		require.NoError(t, err)
 
-		minimumStorage, err := cadence.NewUFix64("0.00008021")
+		minimumStorage, err := cadence.NewUFix64("0.00008164")
```
janezpodhostnik, line 164: The storage for accounts has inflated a bit. This is good to keep in mind! We will need to run the storage sync migration.

fvm/transactionEnv.go
```diff
@@ -172,7 +187,7 @@ func (e *TransactionEnv) GetAuthorizedAccountsForContractUpdates() []common.Addr
 		e.ctx.Logger.Warn().Msg("failed to read contract deployment authorized accounts from service account. using default behaviour instead.")
 		return defaultAccounts
 	}
-	addresses, ok := utils.OptionalCadenceValueToAddressSlice(value)
+	addresses, ok := utils.CadenceValueToAddressSlice(value)
```
janezpodhostnik, line 190: @turbolent ReadStored now returns T instead of optional of T, is this expected?

tarakby:
I think your PR is competing with this one: https://github.com/onflow/flow-go/pull/1605. 
Let's merge #1605 first (working on it now) and then add your changes (less conflicts this way than the other way around).

tarakby:
Well, I think we need to merge your branch and mine eventually. There are Cadence tests on my branch that would only work with the latest Cadence version, while your branch already updates Cadence.

fvm/transactionEnv.go
turbolent: Good catch! Yes, the behaviour changed, instead of `OptionalValue` with `NilValue`/`SomeValue`, the function now just returns `nil` and the value as-is if it exists. 

I forgot to update the comments during the refactor, I fixed it in https://github.com/onflow/cadence/pull/1439

turbolent:
Nice work! As there likely won't be a Cadence release in time for the next spork, master should stay on the latest released version of Cadence. Of course we still want to update flow-go early to the next Cadence release, which this PR does, but I think it should go into a feature branch, e.g. `feature-secure-cadence`. There is also a plan to update Cadence's flow-go update GitHub action to target the feature branch for now (https://github.com/onflow/cadence/issues/1415).

fvm/bootstrap.go
```diff
@@ -678,8 +678,8 @@ transaction(clusterWeights: [{String: UInt64}]) {
 		randomSource: %s,
 		collectorClusters: clusters,
         // NOTE: clusterQCs and dkgPubKeys are empty because these initial values are not used
-		clusterQCs: [],
-		dkgPubKeys: [],
+		clusterQCs: [] as [FlowClusterQC.ClusterQC],
+		dkgKeys: [] as [String],
```
turbolent, line 682: Did type checking fail when just using `[]`?

fvm/fvm.go
```diff
@@ -50,21 +49,6 @@ func (vm *VirtualMachine) Run(ctx Context, proc Procedure, v state.View, program
 		state.WithMaxInteractionSizeAllowed(ctx.MaxStateInteractionSize))
 	sth := state.NewStateHolder(st)
 
-	defer func() {
```
turbolent, line 53: I guess this was removed because there is similar logic on the Cadence side now?

tarakby:
@janez is there any further update to make on this PR following Bastian's comments? 
If no, I'll merge this PR into another [branch](https://github.com/onflow/flow-go/pull/1605).

fvm/bootstrap.go
janezpodhostnik: Yes. I had to do a lot of similar changes to core contracts as well.

fvm/fvm.go
janezpodhostnik: This error no loner exists from the cadence side.

PR #1976, opened 2022-02-09
Fix bootstrap cluster qc generation

jordanschalm:
Previously we passed in only node infos for which we had private keys (internal nodes) to generate the cluster QCs. The previous logic would make use of all nodes to produce a QC.

With changes in Consensus V2, this logic now only uses 2/3 of these node votes (assuming the passed in list is the full cluster committee), which causes QC generation failure in the presence of partners.

This commit separately provides the cluster signers and the full cluster committee.


zhangchiqing (approved):

AlexHentschel (approved):
🎉

PR #1980, opened 2022-02-10
Fix race in epoch switchover handling

jordanschalm:
There is a race in `epochmgr` when responding to epoch transition events, because the event handler used `state.Final` rather than `state.AtBlockID` with the block ID in the event. More details in [testing doc](https://www.notion.so/dapperlabs/Test-21-Dynamic-Startup-Collection-Node-d0bf044632fe4ab78e5c814b78d853b1#8fa9538d5c094ea089ac57dd99814d66)

### DIFF
--- a/engine/collection/epochmgr/engine.go
+++ b/engine/collection/epochmgr/engine.go
@@ -192,7 +192,7 @@ func (e *Engine) Ready() <-chan struct{} {
 		finalSnapshot := e.state.Final()
 		phase, err := finalSnapshot.Phase()
 		if err != nil {
-			e.log.Error().Err(err).Msg("could not check phase")
+			e.log.Fatal().Err(err).Msg("could not check phase")
 			return
 		}
 		if phase == flow.EpochPhaseSetup {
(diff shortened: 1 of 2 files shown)


kc1116 (approved):

engine/collection/epochmgr/engine.go
```diff
@@ -189,13 +189,16 @@ func (e *Engine) Ready() <-chan struct{} {
 	}, func() {
 		// check the current phase on startup, in case we are in setup phase
 		// and haven't yet voted for the next root QC
-		phase, err := e.state.Final().Phase()
+		finalSnapshot := e.state.Final()
+		phase, err := finalSnapshot.Phase()
 		if err != nil {
 			e.log.Error().Err(err).Msg("could not check phase")
```
zhangchiqing, line 195: I think this is an unexpected error,  should we fatal instead?
```suggestion
			e.log.Fatal().Err(err).Msg("could not check phase")
```

zhangchiqing (approved):

jordanschalm:
bors merge

PR #1982, opened 2022-02-10
log vote's signer and block when vote is invalid

zhangchiqing:
Adding logs so that when we receive invalid vote, we know the signer from the error log


jordanschalm (approved):

durkmurder (approved):

zhangchiqing:
bors merge

PR #1983, opened 2022-02-10
reverting the execute script from AN to EN to throw a gRPC error on s…

vishalchangrani:
…cript failure

Slack [convo](https://axiomzen.slack.com/archives/CEXKUH74G/p1644515348776389)

Reverting back some part of these [changes](https://github.com/onflow/flow-go/pull/1905)

### DIFF
--- a/engine/access/rpc/backend/backend_scripts.go
+++ b/engine/access/rpc/backend/backend_scripts.go
@@ -104,14 +104,14 @@ func (b *backendScripts) executeScriptOnExecutionNode(
 				Msg("Successfully executed script")
 			return result, nil
 		}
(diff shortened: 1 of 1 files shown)


Kay-Zee:
We're not skipping the multi error, i'm guessing we're okay with this?

Kay-Zee (approved):

koko1123:
> We're not skipping the multi error, i'm guessing we're okay with this?

yeah, we don't need to try other ENs if it fails for this reason, so just short-circuit

vishalchangrani:
> Thanks for this!

sorry for not catching this in ur PR - i didn't realize it will break clients.

PR #1992, opened 2022-02-11
[FVM] relax tag conditions for ECDSA runtime signature verification 

ramtinms:
In order to facilitate verifying external signatures with generic formats (e.g. not having a domain separation tag at all or not-fixed-length tags) we are relaxing some of the conditions on the runtime ECDSA signature verification method with regards to domain separation tags (specifically the `PublicKey.verify` in Cadence)

Note this change is only related to runtime (contract) ECDSA signatures, and there is no change to ECDSA transaction signatures. 

A domain tag now can be empty or any byte slice smaller or equal to 32 bytes. 
- If an empty tag is provided, it won't be included in hashing the message for the signature verification.
- if any non-empty tag shorter than 32 bytes is provided, it would be padded with zero bytes to 32 bytes and will prepend the message to be verified. 
- if you still need to use larger size domain tags or shorter tags without padding, you could use an empty tag and prepend your tag to the message directly. 

⚠️ We highly recommend dapp developers to add tags to protect their users and separate their signatures from other dapp domains, although we are opening up this condition for supporting generic 3rd party networks and applications. Using a non-empty tag in the `Verify` function is preferred, as it guarantees a fixed length tag and enforces injectivity of messages.

A common practice is for the dapp developer to use a hash function to digest contextual and domain data (such as dapp account address, contract name, version, etc) and compute a tag. Allowing a tag of 32-bytes allow using a hash function of 32-bytes output.


fvm/crypto/crypto_test.go
```diff
@@ -205,13 +220,6 @@ func TestVerifySignatureFromRuntime(t *testing.T) {
...
 					require.False(t, sigOk)
 				},
-			}, {
-				signTag:   "random_tag",
-				verifyTag: "random_tag",
-				require: func(t *testing.T, sigOk bool, err error) {
-					require.Error(t, err)
-					require.False(t, sigOk)
-				},
 			},
```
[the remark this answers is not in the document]
ramtinms: added at e308874

janezpodhostnik (approved):

fvm/crypto/crypto_test.go
```diff
@@ -132,6 +132,21 @@ func TestVerifySignatureFromRuntime(t *testing.T) {
...
+			[]byte("public kye"),
+			runtime.SignatureAlgorithmECDSA_P256,
+			runtime.HashAlgorithmKECCAK_256,
+		)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), "[Error Code: 1051] invalid value (a very large tag with more than thirty two bytes): tag length (48) is larger than max length allowed (32 bytes).")
+		require.False(t, ok)
+	})
+
+	t.Run("ECDSA verification tag size > 32 bytes shouldn't pass", func(t *testing.T) {
```
tarakby, line 149: I think there is a typo mix up in the test title, the title here got swapped with the one in line 134

fvm/crypto/crypto_test.go
```diff
@@ -132,6 +132,21 @@ func TestVerifySignatureFromRuntime(t *testing.T) {
...
+			string("a very large tag with more than thirty two bytes"),
+			[]byte("some random message"),
+			[]byte("public kye"),
+			runtime.SignatureAlgorithmECDSA_P256,
+			runtime.HashAlgorithmKECCAK_256,
+		)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), "[Error Code: 1051] invalid value (a very large tag with more than thirty two bytes): tag length (48) is larger than max length allowed (32 bytes).")
+		require.False(t, ok)
+	})
```
tarakby, line 147: We could add this to the tag combinations at line 181, and we can delete the extra test:
```
				signTag:   "valid_tag",
				verifyTag: "a very large tag with more than thirty two bytes",
				require: func(t *testing.T, sigOk bool, err error) {
					require.Error(t, err)
                                         require.Equal(t, err.Error(), "[Error Code: 1051] invalid value (a very large tag with more than thirty two bytes): tag length (48) is larger than max length allowed (32 bytes).")
					require.False(t, sigOk)
				},
```

fvm/crypto/crypto_test.go
```diff
@@ -205,13 +228,6 @@ func TestVerifySignatureFromRuntime(t *testing.T) {
 					require.NoError(t, err)
 					require.False(t, sigOk)
 				},
-			}, {
```
tarakby, line 208: Maybe we could keep this test since it tests a random tag?

tarakby (commented):
Thanks for the update 👌🏼 

I left some comments to improve the tests.

fvm/crypto/crypto_test.go
ramtinms: I put it back but had to adjust to NoError and returning sigOK, which is similar to "user" case now

fvm/crypto/crypto_test.go
ramtinms: yup, fixed it

fvm/crypto/crypto_test.go
ramtinms: done

fvm/crypto/crypto_test.go
tarakby: yep it's similar to "user". 
"user" used to be another special tag for something (can't remember) and this was a specific test for that special tag to make things worked for it.

tarakby (approved):
Looks good 🏄🏼

PR #1993, opened 2022-02-11
Target westmere processor when compiling relic for better compatibility and performance

yagop:
Current `-march=native` causes some compilations falls in a AVX512 machine which produces AVX512 code.

There is no point of targeting AVX512 instructions as it may even produce worst performance.

Read Linus Torvalds words (yes, Linux creator).
https://www.extremetech.com/computing/312673-linus-torvalds-i-hope-avx512-dies-a-painful-death 

> in additional to the known performance drop due to decreased frequency, there’s also a small additional penalty of about three percent when switching into and out of 512-bit execution mode

Only because your nodes are running in AVX512 CPU's doesn't mean you have to target your machines, that leaves the community in the dust if they don't use Skylake.

This MR targets `westmere` architecture as is well advanced and have support for AES and other set of instructions. See https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

Either way vector instructions shouldn't be your bottleneck, so you can lower the target to `core2` or even to `x86-64` (most binaries are distributed targeting this version because of compatibility). 

In most scenarios the bottleneck would be the GO-C bindings rather than the C code itself.


tarakby:
- I took some measures on different machines (with AVX512 or AVX2 support) to compare the crypto library bench when compiling for `westmere` and `native`. 
I noticed `westmere` provides similar timings to `native` (some functions are slightly slower, some are slightly faster - btw the crypto library isn't using any AES). I'll keep monitoring the numbers (this depends on the Relic version we're using). For now I'll switch the flag to `westmere` for Intel targets since this seems to cause a friction for node operators not supporting AVX. ARM targets won't be able to support `westmere` though, so we need to make a small change to the build script. 

- Ideally, each node operators should build its own binary from the code. In that case, everyone can use the `native` flag and it won't cause any issue because there will be no cross-compiling involved. I know the Flow team is distributing docker images for now, but my preference would be distribute a build script instead.  

> In most scenarios the bottleneck would be the GO-C bindings rather than the C code itself.
That's correct when crossing Go to C takes a large percentage of the computation. The Go function we implemented make 1 or at most 2 calls to Cgo to minimize that overhead (by batching all C operations in one call). I made some bench to show that cgo calls take a negligible time compared to the expensive crypto computations: 

Call to an empty function using Cgo and then using Go.
```
BenchmarkCGO-8       18769291                64.22 ns/op
BenchmarkGo-8        1000000000               0.3105 ns/op
```
Our crypto functions using cgo:
```
cpu: Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
BenchmarkScalarMult/G1-8                            5659            188836 ns/op
BenchmarkScalarMult/G2-8                            2551            459235 ns/op
BenchmarkHashToG1-8                                 2364            485478 ns/op
BenchmarkCheckG1-8                                  4063            291434 ns/op
BenchmarkBLSBLS12381Sign-8                          1288            922254 ns/op
BenchmarkBLSBLS12381Verify-8                         278           4239508 ns/op
BenchmarkBatchVerify/happy_path-8                     10         107589535 ns/op
BenchmarkBatchVerify/unhappy_path-8                   10         160911913 ns/op
BenchmarkVerifySignatureManyMessages-8               237           5033540 ns/op
BenchmarkAggregate/PrivateKeys-8                   11539            104057 ns/op
BenchmarkAggregate/PublicKeys-8                      294           4044159 ns/op
BenchmarkAggregate/Signatures-8                       16          66326376 ns/op
BenchmarkSimpleKeyGen-8                               38          28224543 ns/op
```
You can notice that 64ns is negligible here.

tarakby:
Thanks for suggesting to work with `westmere`, I'll make the change as mentioned and should reconsider this again later when the Flow team stops distributing compiled images. 

Would you like to make the change to fix the ARM case? or do you prefer I make a commit into your branch instead?

yagop:
Hey @tarakby sorry for the delay I missed it a little.

Many thanks for the review, good analysis. 👍 

Very happy you target a architecture our hardware can run. AVX512 was giving us lot of headaches.

> Would you like to make the change to fix the ARM case? or do you prefer I make a commit into your branch instead?

I did this change, I think should work in any of your cases. Let me know if doesn't.

```
ARCH=$(uname -m 2>/dev/null || true)

# Compile x86_64 as westmere by default.
# See https://github.com/onflow/flow-go/pull/1993#issuecomment-1056048192
if [[ "$ARCH" = "x86_64" ]]; then
    MARCH="-march=westmere"
else
    # Use defaults for ARM
    MARCH=""
fi
```

crypto/relic_build.sh
```diff
@@ -36,13 +36,15 @@ CC_VAL=${CC_VAL:-"$(which cc)"}
...
-ARCH=$(uname -m 2>/dev/null ||true)
+ARCH=$(uname -m 2>/dev/null || true)
 
-if [[ "$ARCH" =~ ^(arm64|armv7|armv7s)$ && "${CC_VERSION_STR[0]}" =~ (clang)  ]]; then
-    #  the "-march=native" option is not supported with clang on ARM
-    MARCH=""
+# Compile x86_64 as westmere by default.
+# See https://github.com/onflow/flow-go/pull/1993#issuecomment-1056048192
+if [[ "$ARCH" = "x86_64" ]]; then
+    MARCH="-march=westmere"
```
tarakby, line 44: Let's keep "native" with ARM-gcc as it is in the original script, because I didn't investigate the 'native" impact in that case yet.

```suggestion
if [[ "$ARCH" =~ "x86_64" ]]; then
# Compile as westmere arch to avoid cross-compilation issues on machines not supporting AVX extensions.
# Relic performance as used in flow crypto library is not impacted by whether it is compiled with "native" or "westmere", as proven by benchmark results.  
    MARCH="-march=westmere"
elif [[ "$ARCH" =~ ^(arm64|armv7|armv7s)$ && "${CC_VERSION_STR[0]}" =~ (clang)  ]]; then
    #  the "-march=native" option is not supported with clang on ARM
    MARCH=""
else
    MARCH="-march=native"
```

tarakby (approved):
Nice, thanks for the update! That should cover the ARM case. 

I left a suggestion below. Feel free to apply it (it should also fix the linter error), and I'll look for a second approval from the team

crypto/relic_build.sh
yagop: I think the default is `native` for gcc under ARM, which is ok.

I will commit your suggestion either way.

crypto/relic_build.sh
```diff
@@ -36,13 +36,20 @@ CC_VAL=${CC_VAL:-"$(which cc)"}
...
+    # Relic performance as used in flow crypto library is not impacted by whether it is compiled with "native" or "westmere", as proven by benchmark results.  
+    MARCH="-march=westmere"
+elif [[ "$ARCH" =~ ^(arm64|armv7|armv7s)$ && "${CC_VERSION_STR[0]}" =~ (clang)  ]]; then
     #  the "-march=native" option is not supported with clang on ARM
     MARCH=""
 else
     MARCH="-march=native"
+else
+    # Use defaults for ARM
+    MARCH=""
```
tarakby, line 52: I think we forgot to remove this bit. 
```suggestion
else
    # Use defaults for ARM
    MARCH=""
```

AlexHentschel (approved):

tarakby:
bors merge

PR #1994, opened 2022-02-11
[Execution] avoiding double encoding of event key before insertion into the event merkle trie

ramtinms:
Currently, due to the bug, the key of every event is double RLP encoded before adding to the event trie. 
This PR changes event encoding before adding it to the event collection by only fingerprinting once. 

Why this is a breaking change, 3rd party software who uses their own implementation of verifying the root hash of the event trie, needs to update to do the same as this PR is doing.


model/flow/event.go
```diff
@@ -111,7 +112,8 @@ func EventsMerkleRootHash(el EventsList) (Identifier, error) {
 		// event fingerprint is the rlp encoding of the wrapperevent
 		// eventID is the standard sha3 hash of the event fingerprint
 		fingerPrint := event.Fingerprint()
-		eventID := MakeID(fingerPrint)
```
ramtinms, line 114: MakeID calls fingerprint again on the fingerprint, so we end up double sha3 hashing the event

ramtinms:
This work has been decided to stay open until the new way of encoding events is ready, so we have a breaking change only once.

model/flow/event.go
m4ksio: won't this invalidate already existing hashes?

tarakby: Looking at the code again, it seems that the old code isn't double hashing. It just calls `FingerPrint` twice, but there is only one hashing happening inside `MakeID`. Did I miss something? 

To avoid calling `FingerPrint` twice, we could replace line 113 and 114 by:
```
eventID := event.ID()
```
That includes only one call to `MakeID` which is one `FingerPrint` + one `ComputeSHA3_256`.

model/flow/event_test.go
```diff
@@ -94,3 +94,27 @@ func TestEventsList(t *testing.T) {
...
+
+	eventB := flow.Event{
+		Type:             "eventTypeString",
+		TransactionIndex: 1,
+		EventIndex:       3,
+		Payload:          []byte("cadence-json encoded data"),
+		TransactionID:    [flow.IdentifierLen]byte{1, 2, 3},
+	}
+
+	expectedRootHashHex := "355446d7b2b9653403abe28ccc405f46c059d2059cb7863f4964c401ee1aa83b"
```
tarakby, line 115: Is this hash computed by some reference implementation?

model/flow/event.go
AlexHentschel: Yep, agree with Tarak: 
* `event.Fingerprint()` only returns a collision-resistant serialization of the data structure; no hashing is included: https://github.com/onflow/flow-go/blob/06b03d95388c6149639dca6e989e8c833737365e/model/fingerprint/fingerprint.go#L13-L27
* BUT, we are fingerprinting the finger-print, i.e. we unnecessarily encode the data structure twice. Don't know how RLP works under the hood, but I would assume that the second step of encoding might add some auxiliary bytes (?). Either way, I would expect that some overhead is there.

AlexHentschel: @tarakby the reason why Ramtin is not using `event.ID()` is because we later need the serialized byte blob for the entity to store it into the trie:
https://github.com/onflow/flow-go/blob/06b03d95388c6149639dca6e989e8c833737365e/model/flow/event.go#L114-L116
In other words, if we compute the ID first (which requires a serialization) and then serialize the entity again in order to store it in the trie, we are doing the same overhead as the existing code (i.e. one unnecessary serialization).

model/flow/event.go
```diff
@@ -111,7 +112,8 @@ func EventsMerkleRootHash(el EventsList) (Identifier, error) {
...
+		var eventID Identifier
+		hash.ComputeSHA3_256((*[32]byte)(&eventID), fingerPrint)
```
AlexHentschel, line 116:
```suggestion
		hash.ComputeSHA3_256((*[hash.HashLenSHA3_256]byte)(&eventID), fingerPrint)
```

AlexHentschel (approved):
While I don't think we hash twice, we encode unnecessarily twice. This would be very confusing for somebody who is re-implementing the trie somewhere else (maybe a different programming language). Hence, I am in favour of cleaning this up, even if the overhead of encoding twice is potentially small.

model/flow/event.go
tarakby: - I agree with Alex, we should not define the event ID as `Hash(fingerPrint(fingerPrint(event))`, we should update it to `Hash(fingerPrint(event))`. 
- If we would like the implementation to avoid computing `fingerPrint(event)` twice, I suggest to add a new function `MakeIDFromFingerPrint(fingerPrint []data)` right under `MakeID(entity interface{})`:
   - `MakeIDFromFingerPrint` implements the hashing logic. 
   -  `MakeID` uses `MakeIDFromFingerPrint`. 
   - If we don't do this, we would implicitly rely on the merke tree keys being equal to the correct event IDs (for event look up and proofs for instance). While this is not enforced by the implementation.

ramtinms: good points, updated at [f9ee25b](https://github.com/onflow/flow-go/pull/1994/commits/f9ee25ba539ef0d6ec5f255ce33868c93e1fc53e)

model/flow/event_test.go
ramtinms: not through a reference implementation but looking at the trie constructed this and verified the details

tarakby (approved):

PR #1995, opened 2022-02-12
[Consensus] SignerIndices Optimization Interface changes

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/6147

The base branch is the latest master.

This PR made changes to the interfaces in order to present as a proposal for the signer indices optimization.

The main change is to change QuorumCertificate's SignerID field into SignerIndices.

See details and reasonings in comments.

Note: The test cases are still broken, and linting errors are everywhere. They will be addressed once the interface changes are confirmed.

### DIFF
--- a/model/flow/quorum_certificate.go
+++ b/model/flow/quorum_certificate.go
@@ -14,7 +14,7 @@ type QuorumCertificate struct {
 	// In order to distinguish the signature types, the SigData has to be deserialized. Specifically,
 	// the field `SigData.SigType` (bit vector) indicates for each signer which sig type they provided.
 	// For collection cluster, the SignerIDs includes all the staking sig signers.
-	SignerIDs []Identifier
+	SignerIndices []byte
 
 	// For consensus cluster, the SigData is a serialization of the following fields
 	// - SigType []byte, bit-vector indicating the type of sig produced by the signer.
--- a/consensus/hotstuff/blockproducer/block_producer.go
+++ b/consensus/hotstuff/blockproducer/block_producer.go
@@ -33,7 +33,7 @@ func (bp *BlockProducer) MakeBlockProposal(qc *flow.QuorumCertificate, view uint
 	// in hotstuff, we use this for view number and signature-related fields
 	setHotstuffFields := func(header *flow.Header) error {
 		header.View = view
-		header.ParentVoterIDs = qc.SignerIDs
+		header.ParentVoterIndices = qc.SignerIndices
 		header.ParentVoterSigData = qc.SigData
 		header.ProposerID = bp.committee.Self()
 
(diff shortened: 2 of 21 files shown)


model/flow/quorum_certificate.go
```diff
@@ -14,7 +14,7 @@ type QuorumCertificate struct {
 	// In order to distinguish the signature types, the SigData has to be deserialized. Specifically,
 	// the field `SigData.SigType` (bit vector) indicates for each signer which sig type they provided.
 	// For collection cluster, the SignerIDs includes all the staking sig signers.
-	SignerIDs []Identifier
+	SignerIndices []byte
```
zhangchiqing, line 17: Q: Why changing the field here?

A: I was evaluating two options:
1. Keeping the QC's `SignerIDs` field unchanged, and decode the signer indices from the header and query the signer IDs when creating the QC.
2. Change the field here

Option 1 seems better since it can keep the QC's structure unchanged, however it makes a lot of things harder, for instance, the [`BlockFromFlow`](https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/model/block.go#L21) function is a pure function that can convert a block header into a hotstuff Block. Since header uses signerIndices but QC uses signer IDs, then we need to query signerID by signerIndices first. And that query is stateful, and would have to be mocked in tests, it would be more complex to do. 

Instead, I picked Option 2. Basically, we are delaying the query to convert signer indices into signer IDs until its needed by certain modules. And since some modules can work directly with signer indices, keeping it as is could skip some back-and-force conversions.

model/flow/quorum_certificate.go
```diff
@@ -14,7 +14,7 @@ type QuorumCertificate struct {
 	// In order to distinguish the signature types, the SigData has to be deserialized. Specifically,
 	// the field `SigData.SigType` (bit vector) indicates for each signer which sig type they provided.
 	// For collection cluster, the SignerIDs includes all the staking sig signers.
-	SignerIDs []Identifier
```
zhangchiqing, line 17: Q: Why changing the field here?

A: I was evaluating two options:
1. Keeping the QC's `SignerIDs` field unchanged, and decode the signer indices from the header and query the signer IDs when creating the QC.
2. Change the field here

Option 1 seems better since it can keep the QC's structure unchanged, however it makes a lot of things harder, for instance, the [`BlockFromFlow`](https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/model/block.go#L21) function is a pure function that can convert a block header into a hotstuff Block. Since header uses signerIndices but QC uses signer IDs, then we need to query signerID by signerIndices first. And that query is stateful, and would have to be mocked in tests, it would be more complex to do. 

Instead, I picked Option 2. Basically, we are delaying the query to convert signer indices into signer IDs until its needed by certain modules. And since some modules can work directly with signer indices, keeping it as is could skip some back-and-force conversions.

zhangchiqing:
Close in favor of https://github.com/onflow/flow-go/pull/1995

PR #1998, opened 2022-02-14
implement read execution data admin command

smnzhu:
Add a new admin command so we can read execution data from an execution node. This was added to help debug the execution fork


cmd/util/cmd/execution-data-blobstore/cmd/get.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	edID := flow.HashToID(b)
+
+	ed, err := eds.Get(context.Background(), edID)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("failed to get execution data")
+	}
+
+	bytes, err := json.MarshalIndent(ed, "", "  ")
```
zhangchiqing, line 62: The JSON output is well formatted for printing, but the admin tool doesn't. Could we do the same for admin tool as well?

ledger/common/hash/hash.go
```diff
@@ -15,6 +16,27 @@ func (h Hash) String() string {
 	return hex.EncodeToString(h[:])
 }
 
+func (h Hash) MarshalJSON() ([]byte, error) {
```
zhangchiqing, line 19: what needs this method?

cmd/util/cmd/execution-data-blobstore/cmd/get.go
smnzhu: I don't actually think it's necessary for either of them tbh, because it's very easy to take the output and format it in a text editor or pipe to command line tools like `jq`. The reason I format it for the util is because this is what I've seen is the convention for all the util tools, so I decided to just follow this convention. There is no such convention for the admin commands.

cmd/execution/main.go
```diff
@@ -170,6 +172,9 @@ func main() {
 	}
 
 	nodeBuilder.
+		AdminCommand("read-execution-data", func(config *cmd.NodeConfig) commands.AdminCommand {
```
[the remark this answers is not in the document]
smnzhu: This command is specific to Execution nodes, so we add it only in the execution node builder (and later for access nodes as well). This call is just to register the admin command. The admin server is started up later when the entire node is started up.

zhangchiqing (approved):

cmd/util/cmd/execution-data-blobstore/cmd/root.go
```diff
@@ -0,0 +1,72 @@
...
+		log.Fatal().Err(err).Msg("could not init badger datastore")
+	}
+
+	bstore := blockstore.NewBlockstore(ds)
+
+	blockService := blockservice.New(bstore, nil)
+
+	bs := new(mocknetwork.BlobService)
+
+	bs.
```
Kay-Zee, line 54: nit: This looks super out of place, can we just follow the regular convention of having the first func on the same line?

nit^2: all three things here are "bs", could we just spell this one out too? `blockService`

admin/commands/state_synchronization/read_execution_data.go
```diff
@@ -0,0 +1,76 @@
...
+	errInvalidIDValue := fmt.Errorf("invalid value for \"execution_data_id\": %v", id)
+	data := &requestData{}
+
+	idStr, ok := id.(string)
+
+	if !ok {
+		return errInvalidIDValue
+	}
+
+	if len(idStr) == 2*flow.IdentifierLen {
```
Kay-Zee, line 57: a bit surprised this is the way to validate the input here, not that i have a problem with it.

Kay-Zee (approved):

smnzhu:
bors merge

PR #2001, opened 2022-02-14
Protocol randomness use cleanup 

tarakby:
There are 2 sources of randomness being used in the protocol, the beacon output (48-bytes signature), and the epoch setup seed (16-bytes). This PR consolidates how the 2 sources are used. 

Eventually, the epoch setup seed would use the random beacon output, though this is not done in this PR. 

- consolidate the logic of extracting a PRG from the beacon source under the same package `state/protocol/seed`:
    - hashing the source to uniformize the entropy
    - use a customizer to extract different PRGs from the same seed
    - instantiate the ChaCha20 PRGs
    - this allows the possibility to abstract this logic from the rest of the protocol. 
- use a single string customizer of 12 bytes, instead of the list of indices `[]byte`. Add a tool function to convert indices into a string if needed. 
- `RandomSource` of snapshot and epoch return the unique random source, regardless of customizers. For snapshot, the source is the cryptographic signature (48 bytes), for epoch the source is the setup seed (16 bytes). The snapshot source is a verifiable signature. 
- The seed package now contains `PRGFromRandomSource` to extract the PRG needed for the protocol logic. The `Seed()` method is removed (although could be replaced by `PRG()` )
- Tidy up the seed lengths for bootstrapping.


cmd/bootstrap/cmd/key.go
```diff
@@ -43,23 +43,23 @@ func init() {
...
+	keyCmd.Flags().BytesHexVar(&flagMachineSeed, "machine-seed", []byte{}, fmt.Sprintf("hex encoded machine account seed (min %d bytes)", crypto.KeyGenSeedMinLenECDSAP256))
 }
 
 // keyCmdRun generate the node staking key, networking key and node information
 func keyCmdRun(_ *cobra.Command, _ []string) {
 
 	// generate private key seeds if not specified via flag
 	if len(flagNetworkSeed) == 0 {
-		flagNetworkSeed = GenerateRandomSeed()
+		flagNetworkSeed = GenerateRandomSeed(crypto.KeyGenSeedMinLenECDSAP256)
```
zhangchiqing, line 56: 👍 

Nice catch!

cmd/bootstrap/cmd/key.go
```diff
@@ -138,8 +136,7 @@ func generateKeys() (crypto.PrivateKey, crypto.PrivateKey, []byte, error) {
 func generateMachineAccountKey() (crypto.PrivateKey, error) {
 
 	log.Debug().Msg("will generate machine account key")
-	machineSeed := validateSeed(flagMachineSeed)
```
zhangchiqing, line 141: why do we remove this validation?

cmd/bootstrap/cmd/seal.go
```diff
@@ -40,7 +40,7 @@ func constructRootResultAndSeal(
 		DKGPhase3FinalView: firstView + flagNumViewsInStakingAuction + flagNumViewsInDKGPhase*3 - 1,
 		Participants:       participants.Sort(order.Canonical),
 		Assignments:        assignments,
-		RandomSource:       getRandomSource(flagBootstrapRandomSeed),
```
zhangchiqing, line 43: `getRandomSource` validates the seed length, why do we remove the validation?

cmd/bootstrap/cmd/dkg.go
```diff
@@ -21,7 +22,7 @@ func runDKG(nodes []model.NodeInfo) dkg.DKGData {
 	if flagFastKG {
 		dkgData, err = bootstrapDKG.RunFastKG(n, flagBootstrapRandomSeed)
 	} else {
-		dkgData, err = bootstrapDKG.RunDKG(n, GenerateRandomSeeds(n))
+		dkgData, err = bootstrapDKG.RunDKG(n, GenerateRandomSeeds(n, crypto.SeedMinLenDKG))
```
zhangchiqing, line 25: 👍

zhangchiqing (commented):
Nice refactor 👏  Had some questions

state/protocol/seed/seed.go
```diff
@@ -1,44 +1,44 @@
...
-func FromParentSignature(indices []uint32, sigData []byte) ([]byte, error) {
-	// unpack sig data to extract random beacon sig
-	randomBeaconSig, err := packer.UnpackRandomBeaconSig(sigData)
+// PRGFromRandomSource returns a PRG seeded by the source of randomness of the protocol.
+// The customizer is used to generate a task-specific PRG (customizer in this implementation
+// is up to 12-bytes long).
+//
+// The function extracts first the source of randomness from sigData and then hashes
+// it to obtain a seed. Hashing is required to uniformize the entropy over the output,
+// since the source of randmoness is a cryptographic signature.
```
jordanschalm, line 18: Now the random source is passed into this function, not sure if this part of the comment is still relevant.

jordanschalm (approved):

cmd/bootstrap/cmd/seal.go
tarakby: `getRandomSource` just returns a copy of the input seed with a different length: crops it if it is long, and padds with zeros if it's short. 
The padding part is dangerous: it makes the seed look long enough, but doesn't use a proper entropy for it. I prefer to use the originally generated seed and not "correct" it.

cmd/bootstrap/cmd/machine_account_key.go
```diff
@@ -23,15 +24,15 @@ var machineAccountKeyCmd = &cobra.Command{
...
+	machineAccountKeyCmd.Flags().BytesHexVar(&flagMachineSeed, "seed", []byte{}, fmt.Sprintf("hex encoded machine account seed (min %d bytes)", crypto.KeyGenSeedMinLenECDSAP256))
 }
 
 // machineAccountKeyRun generate a machine account key and writes it to a default file path.
 func machineAccountKeyRun(_ *cobra.Command, _ []string) {
 
 	// generate seed if not specified via flag
 	if len(flagMachineSeed) == 0 {
-		flagMachineSeed = GenerateRandomSeed()
+		flagMachineSeed = GenerateRandomSeed(crypto.KeyGenSeedMinLenECDSAP256)
```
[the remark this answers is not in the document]
tarakby: The default flag looked fine in this case 👌🏼  https://github.com/onflow/flow-go/pull/2001/commits/ddb4b26cc1316feb50540bfbd6b8f5282d9dfe44

tarakby:
bors merge

PR #2003, opened 2022-02-14
Make ContractHandler Commit produce deterministic errors

janezpodhostnik:
`h.draftUpdates` is a dictionary.

It was possible that in the for loop an error was hit, which topped the for loop. This could produce a non-deterministic error.

this PR changes the iteration to an ordered list.

### DIFF
--- a/fvm/handler/contract.go
+++ b/fvm/handler/contract.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"bytes"
 	"fmt"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


fvm/handler/contract.go
```diff
@@ -98,14 +99,31 @@ func (h *ContractHandler) RemoveContract(address runtime.Address, name string, s
...
 }
 
+type contractUpdateList []programs.ContractUpdate
+
+func (a contractUpdateList) Len() int      { return len(a) }
+func (a contractUpdateList) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a contractUpdateList) Less(i, j int) bool {
+	return a[i].Address.String() < a[j].Address.String() ||
+		(a[i].Address.String() == a[j].Address.String() && a[i].Name < a[j].Name)
+}
```
turbolent, line 109: Maybe avoid (repeated) string allocation by comparing address arrays directly, and improve naming of receiver:

```suggestion
func (l contractUpdateList) Len() int      { return len(l) }
func (l contractUpdateList) Swap(i, j int) { l[i], l[j] = l[j], l[i] }
func (l contractUpdateList) Less(i, j int) bool {
	switch bytes.Compare(l[i].Address[:], l[j].Address[:]) {
	case -1:
		return true
	case 0:
		return l[i].Name < l[j].Name
	default:
		return false
	}
}

```

turbolent (approved):
👍

fvm/handler/contract.go
```diff
@@ -98,14 +100,37 @@ func (h *ContractHandler) RemoveContract(address runtime.Address, name string, s
...
 	defer h.lock.Unlock()
 
 	updatedKeys := h.UpdateKeys()
+	updateList := make(contractUpdateList, 0)
+
+	for _, uk := range h.draftUpdates {
+		updateList = append(updateList, uk)
+	}
+	// sort does not need to be stable as the contract update key is unique
+	sort.Sort(updateList)
```
ramtinms, line 130: maybe just adding a test to make sure it works as expected.

ramtinms (approved):

fvm/handler/contract.go
janezpodhostnik: added test [62d68d8](https://github.com/onflow/flow-go/pull/2003/commits/62d68d8bfd7d90ec21c257ccc60a0bd722db57ec)

janezpodhostnik:
bors merge

PR #2006, opened 2022-02-15
Create unit test for gRPC ExecutionResult response structure

koko1123:
Adds end to end test for the new protobuf changes to match the structure of `flow.ExecutionResult`


engine/access/access_test.go
```diff
@@ -424,12 +424,53 @@ func (suite *Suite) TestGetExecutionResultByBlockID() {
 		blockID := unittest.IdentifierFixture()
 
 		er := unittest.ExecutionResultFixture(
-			unittest.WithExecutionResultBlockID(blockID),
-			unittest.WithServiceEvents(2))
```
smnzhu, line 428: why did we remove this line?

engine/access/access_test.go
```diff
@@ -424,12 +424,53 @@ func (suite *Suite) TestGetExecutionResultByBlockID() {
...
+			parsedExecResp := &flow.ExecutionResult{
+				PreviousResultID: convert.MessageToIdentifier(er.PreviousResultId),
+				BlockID:          convert.MessageToIdentifier(er.BlockId),
+				Chunks:           parsedChunks,
+				ServiceEvents:    parsedServiceEvents,
+				ExecutionDataID:  convert.MessageToIdentifier(er.ExecutionDataId),
+			}
+			assert.Equal(suite.T(), executionResult.ID(), parsedExecResp.ID())
+		}
+
```
smnzhu, line 473: Originally, I was thinking a slightly different approach:

We define a conversion function that converts a `*accessproto.ExecutionResultForBlockIDResponse` to a `*flow.ExecutionResult`.

Then, we use the conversion function to convert the response, and compare the two using something like [`assert.EqualValues`](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualValues)

engine/access/access_test.go
koko1123: `ServiceEvent` fixture generation is non-deterministic, making parsing hard, but I can do what you suggested below and be more comprehenesive

smnzhu (approved):

engine/common/rpc/convert/convert.go
```diff
@@ -393,3 +394,70 @@ func BytesToInmemSnapshot(bytes []byte) (*inmem.Snapshot, error) {
 
 	return inmem.SnapshotFromEncodable(encodable), nil
 }
+
+func ExecResultProtoToFlowExecResult(proto *entities.ExecutionResult) (*flow.ExecutionResult, error) {
```
sideninja, line 398: a bit nit but maybe to follow the naming convention for converters this could be named `protoToExecutionResult`

engine/common/rpc/convert/convert.go
```diff
@@ -393,3 +394,70 @@ func BytesToInmemSnapshot(bytes []byte) (*inmem.Snapshot, error) {
...
+	// convert Chunks
+	parsedChunks := make(flow.ChunkList, len(proto.Chunks))
+	for i := 0; i < len(proto.Chunks); i++ {
```
sideninja, line 401: nit again
```suggestion
	for i, chunk := range proto.Chunks {
```

engine/common/rpc/convert/convert.go
```diff
@@ -393,3 +394,70 @@ func BytesToInmemSnapshot(bytes []byte) (*inmem.Snapshot, error) {
...
+		}
+	}
+	// convert ServiceEvents
+	parsedServiceEvents := make(flow.ServiceEventList, len(proto.ServiceEvents))
+	for i := 0; i < len(proto.ServiceEvents); i++ {
+		var event interface{}
+		rawEvent := proto.ServiceEvents[i].Payload
+		// map keys correctly
+		tp := proto.ServiceEvents[i].Type
+		switch tp {
```
sideninja, line 431:
```suggestion
		switch event.Type {
```

engine/access/access_test.go
```diff
@@ -461,6 +462,11 @@ func (suite *Suite) TestGetExecutionResultByBlockID() {
 
 				assert.Equal(suite.T(), marshalledEvent, er.ServiceEvents[i].Payload)
 			}
+			parsedExecResult, err := convert.ExecResultProtoToFlowExecResult(resp.ExecutionResult)
+			require.NoError(suite.T(), err)
+			assert.Equal(suite.T(), parsedExecResult, executionResult)
+			assert.EqualValues(suite.T(), parsedExecResult, executionResult)
```
sideninja, line 468: isn't `EqualValues` redundant in this case since `Equal` asserts follow more strict rules

sideninja (approved):
Small comments, else it looks good to me

engine/common/rpc/convert/convert.go
```diff
@@ -393,3 +394,68 @@ func BytesToInmemSnapshot(bytes []byte) (*inmem.Snapshot, error) {
 
 	return inmem.SnapshotFromEncodable(encodable), nil
 }
+
+func ProtoToExecutionResult(proto *entities.ExecutionResult) (*flow.ExecutionResult, error) {
+	// convert Chunks
+	parsedChunks := make(flow.ChunkList, len(proto.Chunks))
```
kc1116, line 400: nit: Add parsedChunks and parsedServiceEvents parsing logic into their own funcs

kc1116 (approved):
Small nit comment, otherwise LGTM

PR #2016, opened 2022-02-16
[Consensus] Logs for `VoteCollectors` cleanup

durkmurder:
### Context

Added extra logs for narrowing down possible memory leak.

### DIFF
--- a/consensus/hotstuff/voteaggregator/vote_collectors.go
+++ b/consensus/hotstuff/voteaggregator/vote_collectors.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/rs/zerolog"
+
 	"github.com/onflow/flow-go/consensus/hotstuff"
 	"github.com/onflow/flow-go/module/component"
 	"github.com/onflow/flow-go/module/irrecoverable"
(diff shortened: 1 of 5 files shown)


consensus/hotstuff/voteaggregator/vote_collectors.go
```diff
@@ -106,6 +113,11 @@ func (v *VoteCollectors) getCollector(view uint64) (hotstuff.VoteCollector, bool
 func (v *VoteCollectors) PruneUpToView(lowestRetainedView uint64) {
 	v.lock.Lock()
 	defer v.lock.Unlock()
+	v.log.Info().
+		Uint64("from", v.lowestRetainedView).
+		Uint64("to", lowestRetainedView).
+		Msgf("pruning vote collectors, total: %d", len(v.collectors))
```
zhangchiqing, line 119: can we remove this log, and calculate the number of pruned, and add it to the log [`pruned vote collectors, total` here ](https://github.com/onflow/flow-go/pull/2016/files#diff-cf8d024c393d09fb8178e6a6bed7da12e7ff00ab7d8ced31645b47d02ebd7290R148)

durkmurder: Of course, but why? I think for logging purposes it's very straightforward and easy to understand. I would like to keep it as simple as possible for this purpose.

PR #2020, opened 2022-02-16
hotfix for consensus memory leak

AlexHentschel:
we forgot to prune the VoteAggregator (`voteAggregator.PruneUpToView` is never called by the business logic; except at startup)


AlexHentschel:
hotfix; mature version implemented by PR https://github.com/onflow/flow-go/pull/2040

PR #2028, opened 2022-02-17
FVM error detection

sideninja:
This is a small refactor of the logic for error detection, instead of using a literal string value it references the error definition.


janezpodhostnik (approved):
Nice find

ramtinms (approved):

PR #2031, opened 2022-02-17
Fix TransactionContractFunctionInvoker error log

janezpodhostnik:
Reduced the severity of the log, as it is sometimes expected that this might happen.

Also added some extra information to the log.


ramtinms (approved):

jordanschalm (approved):
🙇

PR #2033, opened 2022-02-18
[BFT Testing] Implements Corruptible Conduits

yhassanzadeh13:
Corruptible Conduit Framework is an integration testing framework for Byzantine Fault Tolerant (BFT) setups.
As shown by the figure below, this framework is composed of Corruptible Conduits and Corruptible Conduit Factory.
A Corruptible Conduit Factory (CCF) is configured on each node that is meant to play _malicious_ during the test scenario. 
The CCF utilizes Corruptible Conduits (CC)s to connect the engines of its node to the networking adaptor.
In contrast to the normal conduits, the CCs do not relay the events from their engine to the network adaptor. 
Instead, they relay the events from their engine to the CCF. On receiving an event from a CC, the CCF forwards 
the message to a remote attacker. The attacker is in charge of orchestrating specific attacks through its 
programmable attack vectors. The attacker controls the set of malicious nodes. 
All messages from engines of attacker-controlled corrupted nodes are directly wired to the attacker. 
The attacker runs its registered attack vectors on each received message from its corrupted nodes' engines. 
As the result of running the attack vectors, the attacker may ask CCF of some corrupted nodes to send some certain messages on its behalf. In this way, the Corruptible Conduit Framework empowers testing attack vectors by just replacing the default conduit factory of malicious nodes with a CCF, and without any further modification to the application-layer implementation of the node. 
corruptible


go.mod
```diff
@@ -90,13 +90,13 @@ require (
...
+	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	golang.org/x/sys v0.0.0-20220114195835-da31bd327af9 // indirect
+	golang.org/x/sys v0.0.0-20220209214540-3681064d5158 // indirect
 	golang.org/x/time v0.0.0-20210220033141-f8bda1e9f3ba
 	google.golang.org/api v0.63.0
-	google.golang.org/genproto v0.0.0-20211208223120-3a66f561d7aa
-	google.golang.org/grpc v1.43.0
+	google.golang.org/genproto v0.0.0-20220211171837-173942840c17
+	google.golang.org/grpc v1.44.0
```
yhassanzadeh13, line 99: @vishalchangrani not sure why this version gets upgraded. I tried a downgrade manually but realized it is applied by `go mod tidy`.

insecure/corruptible/factory_test.go
```diff
@@ -0,0 +1,225 @@
...
+	require.NotNil(t, c)
+}
+
+// TestNewConduit_MissingAdapter checks when factory does not have an adapter registered,
+// any attempts on creating a conduit fails with an error.
+func TestNewConduit_MissingAdapter(t *testing.T) {
+	f := NewCorruptibleConduitFactory(flow.BftTestnet, unittest.IdentifierFixture(), cbor.NewCodec())
+	channel := network.Channel("test-channel")
+
+	c, err := f.NewConduit(context.Background(), channel)
```
gomisha, line 55: Why can't the adapter registration be done when creating a new conduit? That would reduce the chance of forgetting to do this.

insecure/corruptible/factory_test.go
```diff
@@ -0,0 +1,225 @@
...
+	require.Error(t, err)
+	require.Nil(t, c)
+}
+
+// TestFactoryHandleIncomingEvent_AttackerObserve evaluates that the incoming messages to the conduit factory are routed to the
+// registered attacker if one exists.
+func TestFactoryHandleIncomingEvent_AttackerObserve(t *testing.T) {
```
gomisha, line 62: I don’t see where `Observe()` is called for the attacker. Can there be more explanation how we set up simulating routing messages to the attacker?

insecure/corruptible/factory_test.go
yhassanzadeh13: For the default (i.e., correct) conduits, the `adapter` is registered at [the creation time](https://github.com/onflow/flow-go/blob/9cb2a74ae2cc062efb7faeb817fa3d1831c7c80b/network/p2p/conduit/conduit.go#L48).

The corrupted conduits, however, do not interact directly with `adapter`. Rather all interactions are done through their `master` (i.e., `CorruptibleConduitFactory`). However, I think your question is "Why can't the adapter registration be done when creating a _new conduit factory_ (and not a conduit itself)?" The answer is: 

The instantiation of the [`adapter` (i.e., `network`)](https://github.com/onflow/flow-go/blob/master/network/p2p/network.go#L134) happens after the creation of the `ConduitFactory`. Indeed, the `ConduitFactory` is passed as an instantiation parameter to the `network` factory. 

That is the reason we may not pass `adapter` when creating a new `ConduitFactory`, since it basically does not exist at that time.

insecure/corruptible/factory_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/onflow/flow-go/insecure"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/model/libp2p/message"
+	"github.com/onflow/flow-go/network"
+	"github.com/onflow/flow-go/network/codec/cbor"
+	"github.com/onflow/flow-go/network/mocknetwork"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestRegisterAdapter(t *testing.T) {
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2033/commits/8d738198013eeb3010c480d3fe4bbfa6eb285585

go.mod
vishalchangrani: i think thats fine...it is just a minor version upgrade.

insecure/attacker.proto
```diff
@@ -0,0 +1,44 @@
...
+// protoc insecure/attacker.proto --go_out=plugins=grpc:.
+
+// CorruptibleConduitFactory allows an attacker to register itself to the factory, and send messages on behalf of its conduits.
+service CorruptibleConduitFactory {
+  rpc RegisterAttacker(AttackerRegisterMessage) returns (google.protobuf.Empty) {}
+  rpc ProcessAttackerMessage(Message) returns (google.protobuf.Empty) {}
+}
+
+service Attacker {
+  rpc Observe(Message) returns (google.protobuf.Empty) {}
```
vishalchangrani, line 19: shouldn't this return a [stream](https://grpc.io/docs/languages/go/basics/#server-side-streaming-rpc) instead? That way the Attacker would keep getting all the messages coming on the conduit?

insecure/attacker.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Message is represents the message exchanged between the CorruptibleConduitFactory and Attacker services.
+message Message {
+  string ChannelID = 1;
+  bytes OriginID = 2;
+  uint32 Targets = 3;
+  repeated bytes TargetIDs = 4;
+  bytes Payload = 5;
+  string Type = 6;
```
vishalchangrani, line 29: what is type?

insecure/corruptible/factory_test.go
yhassanzadeh13: Added comment in https://github.com/onflow/flow-go/pull/2033/commits/55f291604dbfde4ad3ec9fd9a53f397a2926573a.

insecure/attacker.proto
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2033/commits/b4f979db4706d4e020341841de24712364a6596a

insecure/attacker.proto
yhassanzadeh13: removed it in https://github.com/onflow/flow-go/pull/2033/commits/45719f9d5c61f8fb4b612a3dd7715928271434b0, as we no longer use type for encoding and decoding (it is embedded into the message).

insecure/corruptible/factory.go
```diff
@@ -0,0 +1,239 @@
...
+	for {
+		select {
+		case <-c.ComponentManager.ShutdownSignal():
+			if c.attackerObserveClient != nil {
+				_, err := c.attackerObserveClient.CloseAndRecv()
+				if err != nil {
+					c.logger.Fatal().Err(err).Msg("could not close processing stream from attacker")
+					return err
+				}
+			}
```
vishalchangrani, line 87:
```suggestion
			}
			returtn nil
```

insecure/readme.md
```diff
@@ -0,0 +1,6 @@
+# Insecure Package
+
+The insecure package encapsulates tools and technologies related to testing codebase against
```
vishalchangrani, line 3:
```suggestion
The insecure package encapsulates tools and technologies for testing codebase against
```

insecure/readme.md
```diff
@@ -0,0 +1,6 @@
...
+attack vectors of malicious nodes. The package and sub-packages namings are chosen in a way that reflect
+this purpose. This package should not be utilized for any production-grade development that affects
```
vishalchangrani, line 5:
```suggestion
this purpose. This package if fore **testing only** and should **not**  be utilized for any production-grade development that affects
```

model/flow/identifier.go
```diff
@@ -196,7 +197,7 @@ func Sample(size uint, ids ...Identifier) []Identifier {
 	return dup[:size]
 }
 
-func CidToFlowID(c cid.Cid) (Identifier, error) {
+func CidToId(c cid.Cid) (Identifier, error) {
```
vishalchangrani, line 200: is there a reason for this rename? `CidToFlowID` seems less confusing, no?

network/codec/cbor/envelope.go
```diff
@@ -52,5 +52,7 @@ const (
 	// DKG
 	CodeDKGMessage
 
+	// BFT Testing
+	CodeInsecureMessage
```
vishalchangrani, line 56:
```suggestion
	CodeInsecureMessage
	
```

model/flow/identifier.go
yhassanzadeh13: It is externally used as `flow.CiDToID(..)`, and since it has the package name, I found the `Flow` in the function name redundant. But I can take it back if you think it is confusing.

insecure/corruptible/factory.go
yhassanzadeh13: Thanks for catching this, applied in https://github.com/onflow/flow-go/pull/2033/commits/e55c51e2c7529904ab0abdd7ade71b60577d96a0.

model/flow/identifier.go
vishalchangrani: no that makes sense.

vishalchangrani (approved):
lgtm

insecure/corruptible/factory_test.go
```diff
@@ -0,0 +1,217 @@
...
+	// corruptible conduit factory with no attacker registered.
+	f := NewCorruptibleConduitFactory(unittest.Logger(), flow.BftTestnet, unittest.IdentifierFixture(), codec)
+
+	adapter := &mocknetwork.Adapter{}
+	err := f.RegisterAdapter(adapter)
+	require.NoError(t, err)
+
+	channel := network.Channel("test-channel")
+
+	adapter.On("UnRegisterChannel", channel).Return(nil).Once()
```
gomisha, line 211: Can you add a comment here to explain what we're setting up for the adapter?

insecure/corruptible/factory_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	err = f.EngineClosingChannel(channel)
+	require.NoError(t, err)
+
+	testifymock.AssertExpectationsForObjects(t, adapter)
```
gomisha, line 216: Can you add a comment here to explain what we're asserting?

gomisha (approved):

insecure/corruptible/factory_test.go
yhassanzadeh13: added here: https://github.com/onflow/flow-go/pull/2033/commits/f3128ce727f942ddbe9726023f661c89e2054ed1

insecure/corruptible/factory_test.go
yhassanzadeh13: added here: https://github.com/onflow/flow-go/pull/2033/commits/c5cdd5a3e6e849bbc890629be3bca430dceadc12

PR #2035, opened 2022-02-19
Add message queue to transaction `ingest` engine

jordanschalm:
This PR adds a message queue to the `ingest` engine to avoid blocking network layer threads when processing transactions. 

* Uses `ComponentManager` in `ingest` engine
* Replace `ProcessLocal` with typed `ProcessTransaction` for accepting transactions from local source
* Update `ingress` RPC server component
  * Move to `engine/collection/rpc`, for consistency with the Access node RPC engine
  * Define typed `Backend` for transaction processor dependency (the `ingest` engine) rather than a generic `network.Engine`)


zhangchiqing (approved):
Looks good

engine/collection/ingest/engine.go
```diff
@@ -169,60 +254,88 @@ func (e *Engine) onTransaction(originID flow.Identifier, tx *flow.TransactionBod
...
 
-	counter, err := refEpoch.Counter()
+	epochCounter, err := refEpoch.Counter()
 	if err != nil {
+		log.Debug().Msg("trace - cannot retrieve epoch counter for transaction")
 		return fmt.Errorf("could not get counter for reference epoch: %w", err)
 	}
 	clusters, err := refEpoch.Clustering()
 	if err != nil {
+		log.Debug().Msg("trace - cannot retrieve clusters for transaction")
```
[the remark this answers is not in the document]
jordanschalm: I added that to logs I think could be removed once we determine the underlying cause of the blockage/slow processing from https://github.com/dapperlabs/flow-go/issues/6164, to differentiate from the existing logs. We're going a bit overkill on the logs here to get some more data on this specific issue.

engine/collection/ingest/engine.go
```diff
@@ -23,13 +27,16 @@ import (
 // transactions are delegated to the correct collection cluster, and prepared
 // to be included in a collection.
 type Engine struct {
-	unit                 *engine.Unit
+	lm                   *lifecycle.LifecycleManager
```
smnzhu, line 30: I think you could consider looking into using `component.ComponentManager` instead? It's pretty well documented, and if you grep the codebase there are already many examples of it in use.

It's a better fit for the `Component` interface, because the idea is supposed to be that cancelling the context passed into `Start` should stop the component. You shouldn't need to explicitly call `Done()` to stop it.

engine/collection/ingest/engine.go
```diff
@@ -65,13 +73,44 @@ func New(
 		},
 	)
 
+	// FIFO queue for transactions
+	queue, err := fifoqueue.NewFifoQueue(
+		fifoqueue.WithCapacity(int(config.MaxMessageQueueSize)),
+		fifoqueue.WithLengthObserver(func(len int) {
+			mempoolMetrics.MempoolEntries(metrics.ResourceTransactionIngestQueue, uint(len))
+		}),
+	)
```
smnzhu, line 82: general question: Any reason we seem to prefer using `fifoqueue` in the codebase for this sort of thing? Aside from maybe the fact that it allows you to register a length observer?

I think in most cases buffered channels serve this purpose just as well, if not better (allows multiple consumers to consume from it at once), and it is simpler because you don't have to manually trigger the Notifier.

engine/collection/ingest/engine.go
smnzhu: I would maybe look at `p2p/network.go` as a specific example, because it demonstrates how you would implement public methods (e.g `Register` in the case of the network, but for engines an example of a public method would be `Process`) in a way that takes the context into account.

So that once the component has been shutdown by cancelling the context, calls made to public methods will return an error.

engine/collection/ingest/engine.go
```diff
@@ -87,75 +126,122 @@ func New(
...
 func (e *Engine) Submit(channel network.Channel, originID flow.Identifier, event interface{}) {
-	e.unit.Launch(func() {
-		err := e.process(originID, event)
-		if err != nil {
-			engine.LogError(e.log, err)
-		}
-	})
+	err := e.Process(channel, originID, event)
+	if err != nil {
+		e.log.Fatal().Err(err).Msg("internal error processing event")
```
smnzhu, line 189: Referring back to the comments above, this is an example where ideally you would throw an irrecoverable error using `ctx.Throw`, rather than logging fatal.

engine/collection/ingest/engine.go
```diff
@@ -65,13 +73,44 @@ func New(
...
+				return ok
+			},
+			Store: pendingTransactions,
+		},
+	)
+
 	e := &Engine{
-		unit:                 engine.NewUnit(),
+		lm:                   lifecycle.NewLifecycleManager(),
+		wg:                   new(sync.WaitGroup),
```
peterargue, line 106: rather than using a `LifecycleManager` and `WaitGroup`, maybe this could use a `ComponentManager`. Right now this mixes the interfaces and it may get confusing.

e.g. after engine is instantiated, add something like
```
	cm := component.NewComponentManagerBuilder().
		AddWorker(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
			ready()
			e.loop(ctx)
		}).
		Build()
	e.Component = cm
```

then you can ditch the explicit ready/done/start methods since the `Component` interface is provided by `cm`

smnzhu:
Also, just tracking https://github.com/onflow/flow-go/pull/2046#discussion_r813342555 here since it also needs to updated on this PR

engine/collection/ingest/engine.go
```diff
@@ -87,75 +126,122 @@ func New(
...
-// a blocking manner. It returns the potential processing error when done.
-func (e *Engine) Process(channel network.Channel, originID flow.Identifier, event interface{}) error {
-	return e.unit.Do(func() error {
-		return e.process(originID, event)
-	})
-}
+// processAvailableMessages is called when the message queue is non-empty. It
+// will process transactions while the queue is non-empty, then return.
+func (e *Engine) processAvailableMessages() error {
+	for {
```
peterargue, line 216: this should probably also check if the context is cancelled

engine/collection/ingest/engine.go
smnzhu: The only main benefit I see of using `fifoqueue` is that it could allow for a potentially infinitely growing queue, meaning we don't need to drop requests because of reaching capacity.

However, we eliminate this benefit if we restrict the capacity using `WithCapacity`.

engine/collection/ingest/engine.go
```diff
@@ -87,75 +126,122 @@ func New(
...
+	err := e.messageHandler.Process(originID, event)
+	if err != nil {
+		if engine.IsIncompatibleInputTypeError(err) {
+			e.log.Warn().Msgf("%v delivered unsupported message %T through %v", originID, event, channel)
+			return nil
+		}
+		return fmt.Errorf("unexpected error while processing engine message: %w", err)
+	}
+	return nil
+}
```
smnzhu, line 160: As an example, ideally if the context has been cancelled, we would fail this call immediately and not even bother storing the message into the messageHandler.

https://github.com/onflow/flow-go/blob/913abdd0df521bd9c0aba4d0d2a76154716e721d/network/p2p/network.go#L256-L274

shows example of how you might set that up with ComponentManager

engine/collection/ingest/engine.go
jordanschalm: 👍  Thanks, I was working off the compliance engine as an [example](https://github.com/onflow/flow-go/blob/37781b8649ff483942163bbca14b17140c8997f4/engine/collection/compliance/engine.go#L36), `Network` looks like a better example.

engine/collection/ingest/engine.go
jordanschalm: 👍  Will do this, makes much more sense. Thanks!

engine/collection/ingest/engine.go
jordanschalm: The length observer is one, and also it is more flexible in terms of using different message stores (eg. a priority queue). We initially used buffered channels when first adding a per-engine message queue, and ultimately decided against it. Another reason was that the implementation we had initially had a lot of channel management boilerplate duplicated in each engine which was difficult to reason about if you're not already quite familiar with channels. Another alternative which we talked about (maybe worth revisiting) is to have the `MessageStore` expose a channel for getting the next message to process, rather than using the notifier. Then the `MessageStore` needs to maintain a goroutine to shovel messages into that channel, and to be startable/stoppable.

engine/collection/ingest/engine.go
```diff
@@ -65,97 +72,141 @@ func New(
...
+// loop is the main message processing loop for transaction messages.
+func (e *Engine) processTransactions(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+	ready()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-e.ComponentManager.ShutdownSignal():
+			return
```
peterargue, line 158: under the covers these are the same signal, so you only really need to check the context

engine/collection/ingest/engine.go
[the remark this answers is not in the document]
jordanschalm: Removed these as we have confirmed this engine is not the root cause of the network thread blocking.

engine/collection/ingest/engine.go
smnzhu: I remember a while back we discovered that the `Notifier` had a subtle bug. I don't remember the exact details, but basically if multiple messages are put into the queue concurrently, there's a race condition that could cause only a single notification to be emitted from the notifier.

So if we want to keep using this, we should either fix the Notifier or expose the channel as you suggested.

engine/collection/ingest/engine.go
```diff
@@ -65,97 +72,147 @@ func New(
...
+		return nil
 	}
 }
 
 // onTransaction handles receipt of a new transaction. This can be submitted
 // from outside the system or routed from another collection node.
 func (e *Engine) onTransaction(originID flow.Identifier, tx *flow.TransactionBody) error {
 
-	txID := tx.ID()
+	defer e.engMetrics.MessageHandled(metrics.EngineCollectionIngest, metrics.MessageTransaction)
```
smnzhu, line 213: This function is enormous, would it be possible to split it up a bit?

smnzhu (approved):

engine/collection/ingest/engine.go
```diff
@@ -65,97 +72,147 @@ func New(
...
-func (e *Engine) Ready() <-chan struct{} {
-	return e.unit.Ready()
-}
+// Process processes a transaction message from the network and enqueues the
+// message. Validation and ingestion is performed in the processQueuedTransactions
+// worker.
+func (e *Engine) Process(channel network.Channel, originID flow.Identifier, event interface{}) error {
+	select {
+	case <-e.ComponentManager.ShutdownSignal():
+		return component.ErrComponentShutdown
```
durkmurder, line 135: do we actually handle this sentinel somewhere? How about we just return `nil` in such case?

engine/collection/ingest/engine.go
```diff
@@ -65,97 +72,147 @@ func New(
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-e.messageHandler.GetNotifier():
+			err := e.processAvailableMessages(ctx)
+			if err != nil {
+				// if an error reaches this point, it is unexpected
+				ctx.Throw(err)
```
durkmurder, line 169: Generally speaking `Throw` should terminate current goroutine but I would still add a `return` after to be safe

module/component/component.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/onflow/flow-go/module/util"
 )
 
+// ErrComponentShutdown is returned by a component which has already been shut down.
+var ErrComponentShutdown = fmt.Errorf("component has already shut down")
```
durkmurder, line 16: again, don't believe we handle it somewhere and don't see a practical usage for it, I mean even if we handle it we would most likely just ignore it either way?

durkmurder (approved):
LGTM. Clean code, thank you. Added a few minor comments. 
You did some refactoring of RPC ingestion engine, maybe we should move it to `ComponentManager` as well? Could be done in separate PR as well.

engine/collection/ingest/engine.go
jordanschalm: I would rather return an error than `nil`, since the message is not being processed. We shouldn't assume the caller knows the component is shut down - even if the error isn't handled, it at least gives visibility that the component is not in a healthy state (eg. in network layer logs, which is where this would currently surface).

engine/collection/ingest/engine.go
jordanschalm: > if multiple messages are put into the queue concurrently, there's a race condition that could cause only a single notification to be emitted from the notifier

This sounds like expected behaviour of the `Notifier` to me. If routine A and B attempt to `Notify` before routine `C` wakes to read the notification, one of the `Notify` calls from A or B will be a no-op and `C` will only observe one notification.

engine/collection/ingest/engine.go
durkmurder: I can't see where it will be useful, I mean in future iterations based on [FLIP](https://github.com/onflow/flow/blob/alex/networking-api-flip/flips/20210201-consuming-messages-from-network-layer.md) we won't return any error to network layer since it doesn't know what to do with it.

jordanschalm: What about `ProcessTransaction` for another example? If another component calls `ProcessTransaction` and the component has been shut down, should we return an error or `nil`? I think we should return an error, for the same reasons mentioned in the last comment.

(Currently we don't handle this at all and just process the transaction normally, which I'll fix.)

durkmurder: I would say in case of `ProcessTransaction` it makes sense to return error since it's called from RPC and it can be reported to user.
Based on FLIP, I am mainly concerned about `Process` which is called directly from network layer and as I said, network layer have no clue what to do with error so it's better to omit it.

jordanschalm: Updated in https://github.com/onflow/flow-go/pull/2035/commits/2b6cd4f0b2dc296b6fa7d943d9ef2e057b456cec

engine/collection/ingest/engine.go
smnzhu: This is not the problem, the problem is when there are multiple consumers, e.g `C1`, `C2`, and `C3`. They could all be idle and waiting for a notification, but with the existing `Notifier` implementation, it is possible that a producer could come and insert 20 items into the queue and call `Notify` 20 times, yet only one of the consumers would be woken up. This is undesirable, since if there are 3 workers available, we would obviously want to awaken all of them.

smnzhu: Basically, the root of the issue with the existing `Notifier` implementation lies in the fact that all messages sent to a channel are delivered asynchronously, even when a consumer is already blocked on a read at the point when a send is initiated.

Therefore, in the scenario above, it's possible that when the producer sends notifications 2 to 20, notification 1 has still not been received by any receiver, even though all 3 receivers had already been blocked on a channel read prior to the first notification send.

jordanschalm: You're right that one `Notifier` does not work to notify multiple consumers. It is intended to send a "wake" signal to a worker when "some work" is ready. Not to send one signal per unit of work. That is why the consumer, when awoken, will process all the work available. 

I think exposing a channel as the the mechanism to read from a message queue is a better approach than what we have here.

PR #2038, opened 2022-02-21
[General] Root snapshot QC validation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6145

### Context
This PR implements validation of QCs that are part of the root snapshot.
A few important points that were implemented in different way comparing to what is described in ticket:
- Implemented validation for both root QC and cluster QCs
- Implemented the check as separate function, not as part of `IsValidRootSnapshot`. (`IsValidRootSnapshot` is used in ~130 tests which breaks all of them since we use fixtures for root snapshot). 
- Extra check were added in `FlowNodeBuilder` and `finalize` tool.


state/protocol/badger/validity.go
```diff
@@ -264,3 +270,96 @@ func IsValidRootSnapshot(snap protocol.Snapshot, verifyResultID bool) error {
...
+func validateClusterQC(cluster protocol.Cluster) error {
+	clusterBlock := cluster.RootBlock()
+	clusterRootBlock := &model.Block{
+		BlockID:     clusterBlock.ID(),
+		View:        clusterBlock.Header.View,
+		ProposerID:  clusterBlock.Header.ProposerID,
+		QC:          nil,
+		PayloadHash: clusterBlock.Header.PayloadHash,
+		Timestamp:   clusterBlock.Header.Timestamp,
+	}
```
zhangchiqing, line 351: could we [extract it into a function](https://github.com/onflow/flow-go/blob/master/cmd/bootstrap/run/cluster_qc.go#L23-L30), and reuse here?

cmd/bootstrap/cmd/finalize.go
```diff
@@ -212,6 +212,12 @@ func finalize(cmd *cobra.Command, args []string) {
 		log.Fatal().Err(err).Msg("the generated root snapshot is invalid")
 	}
 
+	// validate the generated root snapshot QCs
+	err = badger.IsValidRootSnapshotQCs(snapshot)
```
zhangchiqing, line 216: could we move `IsValidRootSnapshotQCs` into `IsValidRootSnapshot`?

zhangchiqing (approved):
Looks good. Some minor suggestions.

cmd/bootstrap/cmd/finalize.go
durkmurder: We cannot because `IsValidRootSnapshot` is used in ~130 tests, moving it will break every test or require us to generate a completely valid QC instead of fixtures.

state/protocol/badger/validity.go
```diff
@@ -264,3 +270,88 @@ func IsValidRootSnapshot(snap protocol.Snapshot, verifyResultID bool) error {
...
+
+// validateRootQC performs validation of root QC
+// Returns nil on success
+func validateRootQC(snap protocol.Snapshot) error {
+	rootBlock, err := snap.Head()
+	if err != nil {
+		return fmt.Errorf("could not get root block: %w", err)
+	}
+
+	identities, err := snap.Identities(filter.HasRole(flow.RoleConsensus))
```
jordanschalm, line 310:
```suggestion
	identities, err := snap.Identities(filter.IsVotingConsensusCommitteeMember)
```

jordanschalm (approved):
Just one comment, other LGTM. Thanks for this

PR #2040, opened 2022-02-22
[Consensus] `VoteAggregator` pruning

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6158

### Context

Currently `VoteAggregator` is not pruning stale collectors because it's not connected to finalization events. This behavior is patched on mainnet. This PR introduces a proper fix for this problem by reacting to finalization events in event-driven fashion. 

### What was done

- Implemented logic for handling finalization events in compliance engine
- Updated compliance engine pruning
- Connected compliance engine to finalization distributor(both collection & consensus)
- Added tests

NOTE: I haven't refactored compliance engine workers model since as I recall we will be refactoring or possibly rewriting compliance engine either way.


engine/consensus/compliance/engine.go
```diff
@@ -399,3 +402,28 @@ func (e *Engine) BroadcastProposalWithDelay(header *flow.Header, delay time.Dura
...
+			return
+		case <-finalizationNotifier:
+			finalized, err := e.state.Final().Head()
+			if err != nil {
+				e.log.Fatal().Err(err).Msg("could not retrieve last finalized block")
+			}
+			e.core.ProcessFinalizedBlock(finalized)
+		}
+	}
+}
```
zhangchiqing, line 429: This approach works. 

But I have another idea here. The current approach queries the finalized block each time, we could technically get rid of this query if we consume [Protocol Event](https://github.com/onflow/flow-go/blob/master/state/protocol/events.go#L27), instead of [HotStuff Event](https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/consumer.go#L31). The HotStuff Event only provides block view, but no block height, whereas the ProtocolEvent `OnBlockFinalized` provides the block header. 

But if we do so, we might have additional work to do: we don't want to block the consume call. 

Your current approach has done this by using the event notifier. Event Notifier will cache up to 1 unprocessed event, so that if it takes time to prune mempool/votecollectors, pushing a new event to the event notifier will not be blocked. 

If we want to get rid of the query by using the Protocol Event `OnBlockFinalized`, then we will need to implement a way that ensures:
1. the consumer should not block the protocol state `Mutate` call when processing `OnBlockFinalized` event.
2. If consumer is slow processing events, the event queue should keep the newest event which is the latest finalized block height, and drop the old event.

Thoughts?

durkmurder: With current approach we satisfy both 1) and 2), it's the same approach we use in sealing engine. I believe it's not worth to over-engineer it at this moment since we are planning to update compliance engine either way.

engine/collection/compliance/core.go
```diff
@@ -319,3 +319,15 @@ func (c *Core) prunePendingCache() {
...
+// ProcessFinalizedBlock performs pruning of stale data based on finalization event
+// removes pending blocks below the finalized height and prunes vote aggregator
+func (c *Core) ProcessFinalizedBlock(final *flow.Header) {
+	// remove all pending blocks at or below the finalized height
+	c.pending.PruneByHeight(final.Height)
+
+	// always record the metric
+	c.mempoolMetrics.MempoolEntries(metrics.ResourceClusterProposal, c.pending.Size())
+
+	c.voteAggregator.PruneUpToView(final.View)
```
zhangchiqing, line 332: Now we have a proper fix than the hotfix to the original issue. Do we also want to add some logs for debugging purpose. The pending mempool size is recorded with metrics, [maybe we could add a debug level log for vote aggregator vote collectors size](https://github.com/onflow/flow-go/pull/2016/files)?

zhangchiqing (approved):
Looks good

engine/collection/compliance/core.go
durkmurder: I can add a log, but we can also make a metric for vote aggregator size, maybe that will be more informative, most of the time it should be low, but if finalization gets delayed than we will see spikes.

zhangchiqing: yeah, both work for me.

engine/collection/compliance/engine_test.go
```diff
@@ -310,3 +311,23 @@ func (cs *ComplianceSuite) TestProcessUnsupportedMessageType() {
...
+	done := atomic.NewBool(false)
+	cs.voteAggregator.On("PruneUpToView", finalizedBlock.Header.View).Run(func(mock.Arguments) {
+		done.Toggle()
+	}).Return(nil).Once()
+	cs.engine.OnFinalizedBlock(finalizedBlock.ID())
+
+	// matching engine has at least 100ms ticks for processing events
+	time.Sleep(1 * time.Second)
+
+	require.Eventually(cs.T(), done.Load, time.Second, time.Millisecond*20)
```
jordanschalm, line 330: Any reason we need both a sleep and an `Eventually` here? `Eventually` on its own will wait until the condition holds.
```suggestion

	require.Eventually(cs.T(), done.Load, time.Second, time.Millisecond*20)
```

engine/consensus/compliance/engine_test.go
```diff
@@ -210,3 +211,23 @@ func (cs *ComplianceSuite) TestProcessUnsupportedMessageType() {
...
+	done := atomic.NewBool(false)
+	cs.voteAggregator.On("PruneUpToView", finalizedBlock.View).Run(func(mock.Arguments) {
+		done.Toggle()
+	}).Return(nil).Once()
+	cs.engine.OnFinalizedBlock(finalizedBlock.ID())
+
+	// matching engine has at least 100ms ticks for processing events
+	time.Sleep(1 * time.Second)
+
+	require.Eventually(cs.T(), done.Load, time.Second, time.Millisecond*20)
```
jordanschalm, line 230: Same comment here - I think we could remove the sleep. 

```suggestion

	require.Eventually(cs.T(), done.Load, time.Second, time.Millisecond*20)
```

jordanschalm (approved):

AlexHentschel (changes requested):
I have the feeling that this PR works around some legacy problems in the code. Instead, it would be great if we could update the old code, to work organically with the newer design patterns. 

Structurally, I think it would be good to change the following aspects, which I hope would make the code more modular, less interdependent and less brittle: 
1.  If a component, specifically VoteAggregator, needs some event from a different component, it should **directly subscribe to the event-producer.** I think it is not good style to route events through other components (like what is done here with the finalization event -- I only did that as a hot fix):
    * The critical property for a component is the ability to process inbound events in its own worker threads. Hence, VoteAggregator is its own stand-alone component, i.e. a vertey within the consensus node's internal data flow graph. 
    *  If we start routing events through other components, it will become very hard to track where components get events from. 
    * Furthermore, I am worried that the code will become more brittle, because if some bug in the Compliance engine prevents proper cleanup, the VoteAggregator will also be affected.
    
    Therefore, I would suggest that `VoteAggregator` subscribes to the HotStuff notification [`OnFinalizedBlock `](https://github.com/onflow/flow-go/blob/2d23c7cf7784499fff77968a2f7d075f139ad06e/consensus/hotstuff/consumer.go#L31) (not the similar [Protocol Event](https://github.com/onflow/flow-go/blob/master/state/protocol/events.go#L27) and neither the [Block `Identifier`](https://github.com/onflow/flow-go/blob/d12d1611f35c39eed70f7f8b883aaf49e83f47a4/consensus/hotstuff/notifications/pubsub/finalization_distributor.go#L11))
2. In my opinion, it would be best if the **Compliance engine would directly consume the HotStuff notification [`OnFinalizedBlock `](https://github.com/onflow/flow-go/blob/2d23c7cf7784499fff77968a2f7d075f139ad06e/consensus/hotstuff/consumer.go#L31)** (not the similar [Protocol Event](https://github.com/onflow/flow-go/blob/master/state/protocol/events.go#L27) and neither the [Block `Identifier`](https://github.com/onflow/flow-go/blob/d12d1611f35c39eed70f7f8b883aaf49e83f47a4/consensus/hotstuff/notifications/pubsub/finalization_distributor.go#L11)). Reasoning:
   * The HotStuff notification `OnFinalizedBlock ` already contains a pointer to HotStuff's `Block` representation. I think this is fully sufficient for pruning (details below). Thereby we avoid any additional lookup.
   * The reason we currently do the data base lookup https://github.com/onflow/flow-go/blob/d12d1611f35c39eed70f7f8b883aaf49e83f47a4/engine/consensus/compliance/engine.go#L422 is because `PendingBlockBuffer` and the `PendingClusterBlockBuffer` only prune by height. But this convention was introduced by an engineer who was not fully aware that for HotStuff the primary quantity is `View` (not height). 
    * I believe that changing `PendingBlockBuffer` and the `PendingClusterBlockBuffer` to prune by view would be more intuitive and not much work. 
    * This change would unlock other simplifications, 

> we are planning to update compliance engine either way

For this reason, I am happy if we keep the _internal implementation_ of the Compliance engine a bit more messy. But I think it is worth-while to still clean up the API of the Compliance engine and dependencies (`PendingBlockBuffer` and the `PendingClusterBlockBuffer`)

engine/collection/compliance/engine_test.go
durkmurder: I forgot to remove sleep :(

engine/collection/compliance/engine.go
```diff
@@ -416,3 +419,28 @@ func (e *Engine) BroadcastProposalWithDelay(header *flow.Header, delay time.Dura
...
 	return e.BroadcastProposalWithDelay(header, 0)
 }
+
+// OnFinalizedBlock implements the `OnFinalizedBlock` callback from the `hotstuff.FinalizationConsumer`
+//  (1) Informs sealing.Core about finalization of respective block.
+// CAUTION: the input to this callback is treated as trusted; precautions should be taken that messages
+// from external nodes cannot be considered as inputs to this function
+func (e *Engine) OnFinalizedBlock(flow.Identifier) {
+	e.finalizationEventsNotifier.Notify()
+}
```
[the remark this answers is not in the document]
durkmurder: It's the same event actually, we just deliver only ID.

engine/consensus/compliance/engine.go
```diff
@@ -399,3 +402,28 @@ func (e *Engine) BroadcastProposalWithDelay(header *flow.Header, delay time.Dura
...
 	return e.BroadcastProposalWithDelay(header, 0)
 }
+
+// OnFinalizedBlock implements the `OnFinalizedBlock` callback from the `hotstuff.FinalizationConsumer`
+//  (1) Informs sealing.Core about finalization of respective block.
+// CAUTION: the input to this callback is treated as trusted; precautions should be taken that messages
+// from external nodes cannot be considered as inputs to this function
+func (e *Engine) OnFinalizedBlock(flow.Identifier) {
+	e.finalizationEventsNotifier.Notify()
+}
```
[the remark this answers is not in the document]
durkmurder: It's the same event, we just need to change signature of pubsub.

engine/collection/compliance/core.go
durkmurder: @zhangchiqing https://github.com/onflow/flow-go/pull/2040/commits/e5b277f1b7f75a323465bd9d64942631e208a69c

consensus/hotstuff/voteaggregator/vote_collectors.go
```diff
@@ -105,15 +109,18 @@ func (v *VoteCollectors) getCollector(view uint64) (hotstuff.VoteCollector, bool
 // kept and the method call is a NoOp.
 func (v *VoteCollectors) PruneUpToView(lowestRetainedView uint64) {
 	v.lock.Lock()
-	defer v.lock.Unlock()
```
AlexHentschel, line 108: Was wondering what your motivation is to replace the `defer unlock` with explicit statements. It seems the code would be much easier to accidentally break with subsequent revisions.

consensus/hotstuff/voteaggregator/vote_collectors.go
```diff
@@ -129,5 +136,13 @@ func (v *VoteCollectors) PruneUpToView(lowestRetainedView uint64) {
...
 	}
+	from := v.lowestRetainedView
 	v.lowestRetainedView = lowestRetainedView
+	sizeAfter := len(v.collectors)
+	v.lock.Unlock()
+
+	v.log.Debug().
+		Uint64("from", from).
+		Uint64("to", lowestRetainedView).
+		Msgf("pruned vote collectors, before: %d, after: %d", sizeBefore, sizeAfter)
```
AlexHentschel, line 147: should we change the values for `sizeBefore` and `sizeAfter` to explicit key-value pairs, e.g. 

```suggestion
	v.log.Debug().
		Uint64("prior_lowest_retained_view", from).
		Uint64("lowest_retained_view", lowestRetainedView).
		Int("prior_size", sizeBefore).
		Int("size", sizeAfter).
		Msgf("pruned vote collectors")
```

AlexHentschel (approved):
Looks great. 

Brief request to also update the comment here (we still mention height, while it now should be view): https://github.com/onflow/flow-go/blob/e5b277f1b7f75a323465bd9d64942631e208a69c/engine/common/follower/engine.go#L405-L406

consensus/hotstuff/voteaggregator/vote_collectors.go
durkmurder: true, but I wanted to keep logging outside of critical section, maybe it's not worth it, but I always feel uncomfortable when locking more than needed.

AlexHentschel: No strong feeling about it either. I wouldn’t worry too much about a log message extra within the scope of the lock … its not that we are computing much for this lock message. Either way is fine.

durkmurder:
bors merge

PR #2041, opened 2022-02-22
Khalil/1662 epoch test suite updates

kc1116:
This PR adds some small updates to the epoch integration test suite:
- Make sure we don't pause the container we connect to with the flow client in the AN suite
- Increase the wait time for WaitForSealedView from 2 min -> 3min to decrease flakiness in SN suite

### DIFF
--- a/integration/tests/epochs/epoch.go
+++ b/integration/tests/epochs/epoch.go
@@ -206,9 +206,17 @@ func (s *Suite) runTestEpochJoinAndLeave(role flow.Role, checkNetworkHealth node
 
 	env := utils.LocalnetEnv()
 
-	// grab the first container of this node role type, this is the container we will replace
-	containerToReplace := s.getContainerToReplace(role)
-	require.NotNil(s.T(), containerToReplace)
+	var containerToReplace *testnet.Container
+
+	// replace access_2, avoid replacing access_1 the container used for client connections
+	if role == flow.RoleAccess {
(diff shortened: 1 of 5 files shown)


integration/tests/epochs/epoch.go
```diff
@@ -206,9 +206,17 @@ func (s *Suite) runTestEpochJoinAndLeave(role flow.Role, checkNetworkHealth node
...
+
+	// replace access_2, avoid replacing access_1 the container used for client connections
+	if role == flow.RoleAccess {
+		containerToReplace = s.getContainerToReplace(role, "access_2")
+		require.NotNil(s.T(), containerToReplace)
+	} else {
+		// grab the first container of this node role type, this is the container we will replace
+		containerToReplace = s.getContainerToReplace(role, "")
+		require.NotNil(s.T(), containerToReplace)
+	}
```
zhangchiqing, line 219: Would it be more clear to have two functions:
```suggestion
	if role == flow.RoleAccess {
		containerToReplace = s.getContainerToReplaceByName("access_2")
		require.NotNil(s.T(), containerToReplace)
	} else {
		// grab the first container of this node role type, this is the container we will replace
		containerToReplace = s.getContainerToReplaceByRole(role)
		require.NotNil(s.T(), containerToReplace)
	}
```

jordanschalm (approved):

PR #2042, opened 2022-02-22
Khalil/5893 secretsdb key soft enforcement

kc1116:
This PR adds soft enforcement for encrypted secrets DB . 

- While attempting to read the secrets db encryption key if it is missing the node will exit
- Encryption can be explicitly bypassed using **_--insecure-secrets-db_** (soft enforcement), this will be deprecated in the future when we make encryption a hard requirement

Issue: https://github.com/dapperlabs/flow-go/issues/5893


cmd/scaffold.go
```diff
@@ -146,6 +146,8 @@ func (fnb *FlowNodeBuilder) BaseFlags() {
 	fnb.flags.StringVar(&fnb.BaseConfig.DynamicStartupEpochPhase, "dynamic-startup-epoch-phase", "EpochPhaseSetup", "the target epoch phase for dynamic startup <EpochPhaseStaking|EpochPhaseSetup|EpochPhaseCommitted")
 	fnb.flags.StringVar(&fnb.BaseConfig.DynamicStartupEpoch, "dynamic-startup-epoch", "current", "the target epoch for dynamic-startup, use \"current\" to start node in the current epoch")
 	fnb.flags.DurationVar(&fnb.BaseConfig.DynamicStartupSleepInterval, "dynamic-startup-sleep-interval", time.Minute, "the interval in which the node will check if it can start")
+
+	fnb.flags.BoolVar(&fnb.BaseConfig.InsecureSecretsDB, "--insecure-secrets-db", false, "allow the node to start up without an secrets DB encryption key")
```
jordanschalm, line 150:
```suggestion
	fnb.flags.BoolVar(&fnb.BaseConfig.InsecureSecretsDB, "insecure-secrets-db", false, "allow the node to start up without an secrets DB encryption key")
```

Pretty sure the `--` is not expected when defining the flag name

jordanschalm (commented):
Looks good, just one minor formatting note on the flag. 

Before merging this to `master`, let's make sure to have a plan to send out a heads-up to operators in advance cc @vishalchangrani

cmd/scaffold.go
kc1116: Typo good catch

cmd/scaffold.go
```diff
@@ -561,6 +563,23 @@ func (fnb *FlowNodeBuilder) initSecretsDB() {
 	})
 }
 
+// loadSecretsEncryptionKey attempt to read an encryption key for the secrets DB from the canonical path
+func (fnb *FlowNodeBuilder) loadSecretsEncryptionKey() []byte {
+	encryptionKey, err := loadSecretsEncryptionKey(fnb.BootstrapDir, fnb.NodeID)
+	if errors.Is(err, os.ErrNotExist) {
+		if fnb.NodeRole == flow.RoleConsensus.String() {
```
jordanschalm, line 570: The logic of the additional constraint on consensus nodes is split between this function and `initSecretsDB` currently. To fully understand the logic for this we need to read both functions `loadSecretsEncryptionKey`.

We could make this function agnostic to the role-specific constraints (like the other functions which read bootstrap files, which just read and return the file contents) and consolidate the special logic for consensus nodes into `initSecretsDB`, to make it a bit easier to follow.

Eg.
```go
func loadSecretsEncryptionKey() ([]byte key, bool fileExists, err error) { ... }
```

cmd/scaffold.go
```diff
@@ -538,15 +540,20 @@ func (fnb *FlowNodeBuilder) initSecretsDB() {
...
-	if errors.Is(err, os.ErrNotExist) {
+
+	role, err := flow.ParseRole(fnb.NodeRole)
+	if err != nil {
+		fnb.Logger.Fatal().Str("role", fnb.NodeRole).Err(err).Msg("unable to parse node role")
+	}
+
+	// NOTE: SN nodes need to explicitly set --insecure-secrets-db to true in order to
+	// disable secrets database encryption
+	if role == flow.RoleConsensus && fnb.InsecureSecretsDB {
```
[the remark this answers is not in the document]
jordanschalm: outdated

cmd/scaffold.go
kc1116: Good catch, I left all the logic in initSecretsDB theres not much need to add separate func at the moment https://github.com/onflow/flow-go/pull/2042/commits/29136d315045e543a9be3631b7e51b774f8ff689

jordanschalm (approved):

kc1116:
bors merge

PR #2043, opened 2022-02-22
fix for missing cache

jwinkler2083233:
This place in root_block.go could be much faster, if the ID for an empty block is cached.

Here's what the CPU profile reading is, before the change:

On Access Node:
```
4.31%--github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```

On Collection Node:
```
0.66%--github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```

and on Consensus Node:
```
0.26%     0.00%  app      app                [.] github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```

Here is data from after the change:

On Collection Node:
```
0.55%--github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```

On Consensus Node:
```
0.21%     0.00%  app      app                [.] github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```

On Access Node (main benefit):
```
3.70%     0.01%  app      app                [.] github.com/onflow/flow-go/state/cluster.CanonicalRootBlock
```


state/cluster/root_block.go
```diff
@@ -13,17 +14,24 @@ func CanonicalClusterID(epoch uint64, participants flow.IdentityList) flow.Chain
...
 // CanonicalRootBlock returns the canonical root block for the given
 // cluster in the given epoch. It contains an empty collection referencing
 func CanonicalRootBlock(epoch uint64, participants flow.IdentityList) *cluster.Block {
 
 	chainID := CanonicalClusterID(epoch, participants)
 	payload := cluster.EmptyPayload(flow.ZeroID)
+
+	once.Do(func() {
+		cachedHash = payload.Hash()
+	})
```
zhangchiqing, line 29: I think we could even pre-calculate the payload hash for the root block:

```go
var rootBlockPayload = cluster.EmptyPayload(flow.ZeroID)
var rootBlockPayloadHash = rootBlockPayload.Hash()
```

state/cluster/root_block.go
```diff
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"sync"
```
zhangchiqing, line 5:
```suggestion
```

zhangchiqing (approved):
Looks good. Nice improvement

state/cluster/root_block.go
jwinkler2083233: I like it.  thx.

vishalchangrani (approved):

PR #2046, opened 2022-02-22
[Backport to v0.24]  Add message queue to transaction `ingest` engine

jordanschalm:
Backport https://github.com/onflow/flow-go/pull/2035


zhangchiqing (approved):

engine/collection/ingest/engine.go
```diff
@@ -65,13 +73,44 @@ func New(
...
+				return ok
+			},
+			Store: pendingTransactions,
+		},
+	)
+
 	e := &Engine{
-		unit:                 engine.NewUnit(),
+		lm:                   lifecycle.NewLifecycleManager(),
+		wg:                   new(sync.WaitGroup),
```
durkmurder, line 106: why do we need a wait group? I don't see where we wait

jordanschalm: It's for the processing loop; https://github.com/onflow/flow-go/pull/2046/files#diff-09bd1de20e1e535512432a5e994d212b95f89937bda8a4df0020e9875db5e3bfR132-R133

peterargue: rather than using a `LifecycleManager` and `WaitGroup`, maybe this could use a `ComponentManager`. Right now this mixes the interfaces a bit and it may get confusing.

e.g. after engine is instantiated, add something like
```
	builder := component.NewComponentManagerBuilder().
		AddWorker(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
			ready()
			e.loop(ctx)
		})

	e.cm = builder.Build()
	e.Component = e.cm
```

then you can ditch the explicit ready/done/start methods since the `Component` interface is provided by `cm`

engine/collection/ingest/engine.go
```diff
@@ -87,75 +126,122 @@ func New(
...
-// a blocking manner. It returns the potential processing error when done.
-func (e *Engine) Process(channel network.Channel, originID flow.Identifier, event interface{}) error {
-	return e.unit.Do(func() error {
-		return e.process(originID, event)
-	})
-}
+// processAvailableMessages is called when the message queue is non-empty. It
+// will process transactions while the queue is non-empty, then return.
+func (e *Engine) processAvailableMessages() error {
+	for {
```
peterargue, line 216: this should probably check if the context is cancelled.

smnzhu:
Added some comments to https://github.com/onflow/flow-go/pull/2035 which I hope we can address / discuss before merging to master, but since this is an urgent fix it looks good to be deployed first.

smnzhu (approved):

engine/collection/ingest/engine.go
smnzhu: @jordanschalm I think his point is that nowhere do we actually ever call `wg.Wait()`, so it's not actually be used.

jordanschalm:
Merging this as-is to address the immediate issues on Mainnet. Will address additional suggestions in https://github.com/onflow/flow-go/pull/2035

yhassanzadeh13 (approved):

PR #2047, opened 2022-02-23
[Consensus] SignerIndices Optimization Interface changes (1/4)

zhangchiqing:
## This PR represents the equivalent of a feature branch

For https://github.com/dapperlabs/flow-go/issues/6147

The base branch is the latest master.

This PR made changes to the interfaces in order to present as a proposal for the signer indices optimization.

The main change is to change QuorumCertificate's SignerID field into SignerIndices.

See details and reasonings in comments.

Note: The test cases are still broken, and linting errors are everywhere. They will be addressed once the interface changes are confirmed.


model/flow/quorum_certificate.go
```diff
@@ -14,7 +14,7 @@ type QuorumCertificate struct {
 	// In order to distinguish the signature types, the SigData has to be deserialized. Specifically,
 	// the field `SigData.SigType` (bit vector) indicates for each signer which sig type they provided.
 	// For collection cluster, the SignerIDs includes all the staking sig signers.
-	SignerIDs []Identifier
```
zhangchiqing, line 17: Q: Why changing the field here?

A: I was evaluating two options:
1. Keeping the QC's `SignerIDs` field unchanged, and decode the signer indices from the header and query the signer IDs when creating the QC.
2. Change the field here

Option 1 seems better since it can keep the QC's structure unchanged, however it makes a lot of things harder, for instance, the [`BlockFromFlow`](https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/model/block.go#L21) function is a pure function that can convert a block header into a hotstuff Block. Since header uses signerIndices but QC uses signer IDs, then we need to query signerID by signerIndices first. And that query is stateful, and would have to be mocked in tests, it would be more complex to do. 

Instead, I picked Option 2. Basically, we are delaying the query to convert signer indices into signer IDs until its needed by certain modules. And since some modules can work directly with signer indices, keeping it as is could skip some back-and-force conversions.

model/flow/quorum_certificate.go
```diff
@@ -14,7 +14,7 @@ type QuorumCertificate struct {
...
+	SignerIndices []byte
```
zhangchiqing, line 17: Q: why not combining `SignerIndices` and `SigData` since their type are both `[]byte`?

A: SigData contains signature, we don't want to validate signature before validating the signer IDs, because signature validation is expensive. 

Keeping them separated allows us to decode the signer indices alone, and check the total stakes represents the majority before decoding and validating SigData field.

jordanschalm (commented):
To me this problem is mainly one of storage/network optimization. From that perspective it makes sense to perform the conversion on the API boundaries of the network layer and the storage layer. Currently the proposal is applying essentially an encoding optimization to the model used by the application logic and requiring the application logic to manage conversion between the efficient encoding and the useful representation.

For simplicity I think we also want to consider the ordered list of consensus participants fixed over the duration of an epoch. If we have the consensus committee [A,B,C,D], and node C is slashed or ejected, D still has index 3. So we don't need to worry about changes on a block-by-block basis, just an epoch-by-epoch basis.

For applying the optimization for the network messages, what about something like:
```go
// sending messages out
func (e *Engine) BroadcastProposal(proposal *flow.Header) {
  optimizedHeader := NewOptimizedHeader(proposal, e.SignerIndicesForEpochProvider(proposal.View))
  // ...
}

// receiving inbound messages
func (e *Engine) Process(event interface{}) {
  optimizedHeader := event.(*flow.OptimizedHeader)
  header := FromOptimizedHeader(optimizedHeader, e.SignerIndicesForEpochProvider(optimizedHeader.View))
}
```

It seems like the surface area of the change required to implement this will be much smaller if we do the conversion right before sending the message over the wire.

consensus/hotstuff/validator/validator.go
```diff
@@ -46,13 +47,20 @@ func (v *Validator) ValidateQC(qc *flow.QuorumCertificate, block *model.Block) e
 
 	// Retrieve full Identities of all legitimate consensus participants and the Identities of the qc's signers
 	// IdentityList returned by hotstuff.Committee contains only legitimate consensus participants for the specified block (must have positive stake)
+	// TODO: filter.Any can be removed
 	allParticipants, err := v.committee.Identities(block.BlockID, filter.Any)
```
zhangchiqing, line 51: When validating QC, we need to:
1. decode the signer IDs
2. check if the QC represents enough stakes
3. check if the QC's signature is correct.

If step 1 is failed, then step 2, 3 can be skipped. 

There is some data needed for both step 2 and 3, which is the signer identities, but it's not needed by step 1. But why I'm querying all the participants here before doing step 1?

It's a trade off. 

We could technically query all participants' IDs by view here from the committee module, who could cache all signer IDs with canonical order by view for each epoch. If step 1 failed, then we could skip step 2 and 3. But in the happy path where the validation will pass, where it need to query identities anyway for step 2 and 3. 

So rather, we just query all the identities here, which will be used by step 1,2 and 3. We simplifies the committee API by wasting one query to the committee module in the unhappy case of step 1 fails.

AlexHentschel: 👍 very convincing reasoning

AlexHentschel:
Some thoughts on @jordanschalm [comment](https://github.com/onflow/flow-go/pull/2047#pullrequestreview-894351242)

# Aspect (1)
> For simplicity I think we also want to consider the ordered list of consensus participants fixed over the duration of an epoch. If we have the consensus committee [A,B,C,D], and node C is slashed or ejected, D still has index 3. So we don't need to worry about changes on a block-by-block basis, just an epoch-by-epoch basis.

Conceptually, I agree with this. However, I think there are some consistency considerations here:
* Ideally, [`Identities`](https://github.com/onflow/flow-go/blob/5a49e2e05e08d192aa15f9e5833c8f0833772195/consensus/hotstuff/committees/consensus_committee.go#L72)  would return the nodes in canonical order. 
* At the moment, [`Identities`](https://github.com/onflow/flow-go/blob/5a49e2e05e08d192aa15f9e5833c8f0833772195/consensus/hotstuff/committees/consensus_committee.go#L72) only returns nodes that are allowed to vote. 
* If we implemented your suggestion, there could be a discrepancy (depending on implementation details):
   1. Node C would still have an index, but it wouldn't be included in the return value of `Identities(..)`; meaning it is impossible for an application logic that only has access to `Committee` to convert indices to `IdentityList` and vice versa.
   2. Another option would be to also include ejected nodes in the output of `Identities(..)`. This has the disadvantage that at every point the application logic has to check that an entity that it is not ejected, which I consider error prone.  
   3. A third option would be to integrate the logic for converting between indices and `IdentityList` into the committee. But I think this also has its own risks: engineers might not be aware that they cannot do the conversion themselves and thereby create security vulnerabilities. 

Nevertheless, from my perspective, there is a _significant_ advantage of a 'per-epoch' signer vector representation:
* It allows a node that has fallen behind to do some sanity checks when it receives new blocks where it is still missing the connecting blocks: it can verify the signature on the QC, and thereby convince itself that this is a legitimate extension of the chain. In other words, it will be impossible for any byzantine node to make a node request blocks that don't exist.

AlexHentschel:
Regarding **Aspect (2)**:
> it makes sense to perform the conversion on the API boundaries of the network layer and the storage layer. 

I think there is an implicit assumption here, that is very important to be explicit about:
* In the mature system, we are going to have a few thousand nodes (1k Verifiers and probably equally many Access Nodes). In order to implement conversion of indices to Identifiers, we could:
  1. Always use the entire set of nodes as a base line. This is wasteful. Lets say we have 500 consensus nodes and 3000 nodes in total. Then, we have to represent each signer set by a 3000-bit vector. For consensus nodes, this still provides a significant space saving, compared to enumerating identities (about a factor of 30). Nevertheless, a 500-bit vector would be much nicer. 
  2. However, for result approver sets, we only assign about 80 - 100 verifiers per chunk. Using a 3000 bit vector is still more efficient than enumerating the approver IDs directly. Though, we are still wasting lot of space using a 3000-bit vector compared to just 100 bit, when taking the set of assigned verifiers for each chunk as a basis. Here, we would be wasting space by 30x for each chunk. And there are a lot of chunks (The architecture goal is 1000 chunks per block).
* Therefore, I would conclude that we should be working towards a context-aware representation of the signer set. A QC can only include signatures from consensus nodes. Hence, the signer vector only should have as many bits as we have consensus nodes. For a seal, the vector of the chunk approvers should only have the length equal to the assigned verifiers. 

The other detail that is important to consider is:
* The primary needs to sign the block. Authority to sign messages is (for security reasons) only given to very specific components in the code. So what does the application logic sign then? The representation where all the signer sets are Identifiers? That would mean that _any_ node has to re-compute the set of assigned verifiers for each chunk before they can assert the validity of the Block. Specifically for seals, it is non-trivial to track the assigned verifiers for each chunk. Alternatively, we'll could go back to always encoding the entire set of nodes ... which I don't want to because it is so wasteful. 
* Second, if we consider a context-aware decoding logic, I think there is little disadvantage to moving this decoding logic into the application layer (essentially Leo's proposal) 


Lastly:
* A lot of the crypto already works with indices. So when we do a conversion on the network layer and the storage layer boundaries, it means that we are converting Indices -> Identifiers -> Indices. Which seems unnecessarily complex.

AlexHentschel:
So in summary, my my feeling about using indices only as a transport-layer optimization is :
* considering the immediate next step and the current state of the code base, it is the least time-intensive approach for now
* But I think the methodology would not be very efficient to represent approver sets for seals, which are arguably the most space-critical, because we are aiming to have a lot of chunks per block.  

Hence, I favour Leo's approach of integrating the signer indices into the application logic. Instead of keeping it out of the core protocol logic, I think we could embrace the more compact representation of an identity. For example, instead of the block proposer including its full identity, it could include only its index w.r.t. the consensus committee. Essentially, the indices would become the canonical way to refer to nodes. 
This is going to be much more work now, but my feeling is this would set us on a preferable route for the future allowing us to easily adopt the optimization for other data structures in the future. 🤷 🔮

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -256,7 +256,7 @@ func (p *CombinedVoteProcessorV2) Process(vote *model.Vote) error {
```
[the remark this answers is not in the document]
AlexHentschel: We (Leo, Jordan, myself) discussed this today and Leo raised the following important point, which I had missed before:
* The Random Beacon committee is a subset of the consensus committee. Hence, the signer indices are different. So we anyway need to convert. Converting `Identifier` -> index might be less confusing than converting `'index` -> `index`.

AlexHentschel (approved):
👍  on the approach.

consensus/hotstuff/signature.go
```diff
@@ -124,7 +124,7 @@ type Packer interface {
 	// sig is the aggregated signature data.
 	// Expected error returns during normal operations:
 	//  * none; all errors are symptoms of inconsistent input data or corrupted internal state.
-	Pack(blockID flow.Identifier, sig *BlockSignatureData) ([]flow.Identifier, []byte, error)
+	Pack(blockID flow.Identifier, sig *BlockSignatureData) ([]byte, []byte, error)
```
jordanschalm, line 127:
```suggestion
	Pack(blockID flow.Identifier, sig *BlockSignatureData) (signerIndices []byte, signatures []byte, err error)
```
Could name the return values here for clarity - since they are the same type.

consensus/hotstuff/validator/validator.go
```diff
@@ -46,13 +47,20 @@ func (v *Validator) ValidateQC(qc *flow.QuorumCertificate, block *model.Block) e
...
-		return newInvalidBlockError(block, model.NewInvalidSignerErrorf("some qc signers are duplicated or invalid consensus participants at block %x", block.BlockID))
+
+	signerIndices, err := packer.DecodeSignerIndices(qc.SignerIndices)
+	if err != nil {
+		return newInvalidBlockError(block, fmt.Errorf("qc.SignerIndices is invalid: %w", err))
+	}
+
+	signers, err := filterByIndices(allParticipants, signerIndices)
+	if err != nil {
+		return newInvalidBlockError(block, fmt.Errorf("could not find signers by indices: %w", err))
```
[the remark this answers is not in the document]
jordanschalm: 👍  for integrating into `Committee`. I'm not sure about the protocol state though, based on what we discussed on Friday.

The `Committee` knows about the filtering which defines the ordered list we generate the indices from. But the protocol state doesn't. How do we tell the protocol state that this set of indices represents random beacon participants and that set of indices represents verifiers for block X?

jordanschalm (approved):

consensus/hotstuff/signature.go
tarakby: How about we add a new type for bit vectors (defined under some common module maybe under `flow-go/module`) to differentiate bit vectors from `[]byte` ? 
The common module would provide the conversions to/from `[]flow.Identifier` and `BitVector`.

model/flow/epoch.go
```diff
@@ -128,32 +128,22 @@ type EpochCommit struct {
 // gathered by the ClusterQC smart contract. It contains the aggregated
 // signature over the root block for the cluster as well as the set of voters.
 type ClusterQCVoteData struct {
-	SigData  crypto.Signature // the aggregated signature over all the votes
-	VoterIDs []Identifier     // the set of voters that contributed to the qc
+	SigData      crypto.Signature // the aggregated signature over all the votes
+	VoterIndices []byte           // the set of voters that contributed to the qc
```
tarakby, line 132: This would be a `BitVector` too

tarakby (approved):

consensus/hotstuff/signature.go
zhangchiqing: Yeah, we can do that, [I've issued up,](https://github.com/onflow/flow-go/pull/2047#discussion_r822216975) will do it in a separate PR, since there is another PR already depend on this.

PR #2055, opened 2022-02-23
Change execution receipt received to info log

smnzhu:
Change this to info log so that we can use it in metrics dashboard


zhangchiqing (approved):

engine/consensus/matching/core.go
```diff
@@ -141,6 +140,21 @@ func (c *Core) ProcessReceipt(receipt *flow.ExecutionReceipt) error {
...
+		Hex("execution_data_id", receipt.ExecutionResult.ExecutionDataID[:]).
+		Hex("previous_result", receipt.ExecutionResult.PreviousResultID[:]).
+		Hex("block_id", receipt.ExecutionResult.BlockID[:]).
+		Hex("executor_id", receipt.ExecutorID[:]).
+		Logger()
+
+	if c.receipts.HasReceipt(receipt) {
+		log.Debug().Msg("skipping processing of already known receipt")
+		return false, nil
+	}
```
smnzhu, line 156: Should I put this after the duration metric tracking right below this? Or is this fine?

zhangchiqing: It is fine to be here, I think for that metrics, we are more interested how long a receipt is actually being process.

module/mempool/consensus/execution_tree.go
```diff
@@ -84,6 +84,16 @@ func (et *ExecutionTree) getEquivalenceClass(result *flow.ExecutionResult, block
 	return vertex.(*ReceiptsOfSameResult), nil
 }
 
+func (et *ExecutionTree) HasReceipt(receipt *flow.ExecutionReceipt) bool {
+	et.RLock()
+	defer et.RUnlock()
+	vertex, found := et.forest.GetVertex((&receipt.ExecutionResult).ID())
```
zhangchiqing, line 90: can we compute the ID before the read lock?

AlexHentschel: +100

AlexHentschel (approved):
Looks good. 

Quick request to fix a typo in related code:  please remove repetition of 'the block' 👇 
https://github.com/onflow/flow-go/blob/29a5139fdc2afd89a804139c69c1314e1a16d62a/model/flow/execution_result.go#L51

smnzhu:
bors merge

smnzhu:
bors merge

PR #2057, opened 2022-02-23
Refactor epochs related contract deployments, use blueprints pattern

kc1116:
This PR updates the transactions related to epochs in the fvm bootstrap procedure, it moves these transactions to the blueprints package.


jordanschalm (commented):
Could we move over `registerNodeTransaction` as well? Otherwise looks great 🧹🧹

jordanschalm (approved):

fvm/blueprints/epochs.go
```diff
@@ -1,13 +1,275 @@
...
 	"github.com/onflow/cadence"
 	jsoncdc "github.com/onflow/cadence/encoding/json"
 	"github.com/onflow/flow-core-contracts/lib/go/templates"
 
+	"github.com/onflow/flow-go/model/bootstrap"
+
+	"github.com/onflow/flow-go/module/epochs"
+
 	"github.com/onflow/flow-go/model/flow"
 )
```
joshuahannan, line 17: Is is possible to use templates from the core contracts repo and SDK here instead of hard-coding the transaction text?

joshuahannan (approved):

kc1116:
bors merge

fvm/blueprints/epochs.go
kc1116: It looks like most of the hardcoded transactions either don't have a corresponding tx in **_templates_** or they are different slightly.

joshuahannan: that could mean it might be worth adding some of them, but I guess that make sense

kc1116:
bors merge

PR #2069, opened 2022-02-25
[Network] add metrics and log to track blocked queue workers

smnzhu:
Increase visibility into which topics (and hence engines) the queue workers are blocked on.


zhangchiqing (approved):
Looks good

jordanschalm (approved):
Looks great, thanks for this!

In order to deploy, we'll also need to port to `v0.24` and greenlist the new metrics ([example](https://github.com/dapperlabs/dapper-flow-hosting/pull/344/files)).

smnzhu:
bors merge

PR #2071, opened 2022-02-25
Enable metrics for unstaked access node

m4ksio:
Looks like we don't use flag to enable/disable metrics for AN, they are always on for staked AN, so I made it consistent.

### DIFF
--- a/cmd/access/node_builder/unstaked_access_node_builder.go
+++ b/cmd/access/node_builder/unstaked_access_node_builder.go
@@ -135,9 +135,11 @@ func (builder *UnstakedAccessNodeBuilder) Initialize() error {
 
 	builder.enqueueConnectWithStakedAN()
 
-	builder.EnqueueMetricsServerInit()
-	if err := builder.RegisterBadgerMetrics(); err != nil {
-		return err
+	if builder.BaseConfig.MetricsEnabled {
+		builder.EnqueueMetricsServerInit()
+		if err := builder.RegisterBadgerMetrics(); err != nil {
+			return err
+		}
 	}
 
 	builder.PreInit(builder.initUnstakedLocal())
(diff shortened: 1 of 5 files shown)


cmd/access/node_builder/unstaked_access_node_builder.go
```diff
@@ -135,6 +135,11 @@ func (builder *UnstakedAccessNodeBuilder) Initialize() error {
 
 	builder.enqueueConnectWithStakedAN()
 
+	builder.EnqueueMetricsServerInit()
```
peterargue, line 138: This will mean metrics are also enabled for DPS. maybe we can just export `metricsEnabled` from `cmd.node_builder` and check it here?

m4ksio: Good point. Hmm, ideally consensus follower should behave like an library not like a node.
So rather than enabling/disabling metrics we should expose them to be processed by calling host.
I just wonder if this PR is "good for now" ™️ and adding a ticket for further refactor?

peterargue (approved):

cmd/access/node_builder/unstaked_access_node_builder.go
peterargue: I'd say it's good for now if you don't think this change is going to cause problems with running DPS for mainnet16 (on v0.24)

smnzhu: So once the Access Node / Observer have been refactored and separated into their own folders, they would probably both call the base [`Initialize()`](https://github.com/onflow/flow-go/blob/c622f81817fb148fd96a7576db23cbcfbf5b809b/cmd/scaffold.go#L1020) anyways.

At that point, if we want to keep supporting consensus follower, we will probably want to add new code for it instead of reusing the Observer code, and just initialize the components that are actually needed.

smnzhu: https://github.com/onflow/flow-go/issues/2075

smnzhu (approved):

koko1123 (approved):

m4ksio:
Requested re-review as I changed approach a bit.
Now metrics can be switched on and off, and they are off by default for consensus follower but start in UAN.

peterargue (approved):

smnzhu (approved):

PR #2076, opened 2022-02-25
add admin-tool to enable/disable profiler

zhangchiqing:
This PR added a new admin command `set-profiler-enabled` and also added README for how to use the admin tool


module/metrics/server.go
```diff
@@ -20,14 +20,12 @@ type Server struct {
...
-func NewServer(log zerolog.Logger, port uint, enableProfilerEndpoint bool) *Server {
+func NewServer(log zerolog.Logger, port uint) *Server {
 	addr := ":" + strconv.Itoa(int(port))
 
 	mux := http.NewServeMux()
 	mux.Handle("/metrics", promhttp.Handler())
-	if enableProfilerEndpoint {
-		mux.Handle("/debug/pprof/", http.DefaultServeMux)
-	}
+	mux.Handle("/debug/pprof/", http.DefaultServeMux)
```
zhangchiqing, line 28: I don’t find a way to dynamically enable/disable this endpoint. So I ended up keep it enabled. Thoughts?

simonhf: Who has access to the port? Is it only internal? Otherwise, this could be a kind of attack vector?

integration/localnet/bootstrap.go
```diff
@@ -313,6 +315,7 @@ func prepareService(container testnet.ContainerConfig, i int) Service {
 			fmt.Sprintf("--tracer-enabled=%t", true),
 			"--profiler-dir=/profiler",
 			"--profiler-interval=2m",
+			fmt.Sprintf("--admin-addr=:%v", 9002+n),
```
simonhf, line 318: Is the `+n` so that the `admin-addr` is not unique and does not collide on e.g. localnet? If so, I noticed on localnet that the `--ingress-addr=collection_2:9000` option uses a fixed port but appears to use a custom network interface to differentiate. Would that be an alternative option to using `+n`? Pros? Cons?

cmd/scaffold.go
```diff
@@ -1050,6 +1048,8 @@ func (fnb *FlowNodeBuilder) Initialize() error {
 func (fnb *FlowNodeBuilder) RegisterDefaultAdminCommands() {
 	fnb.AdminCommand("set-log-level", func(config *NodeConfig) commands.AdminCommand {
 		return &common.SetLogLevelCommand{}
+	}).AdminCommand("set-profiler-enabled", func(config *NodeConfig) commands.AdminCommand {
+		return &common.SetProfilerEnabledCommand{}
```
simonhf, line 1052: What would be the example curl command to enable the profiler on a particular node?

Is it only possible to enable the profiler, but not disable it?

Presumably this is just a way to start the profiler later when the node is running? But the interval and duration cannot be set via admin?

module/metrics/server.go
zhangchiqing: Reverted

integration/localnet/bootstrap.go
zhangchiqing: Reverted, this will be a fixed internal port

cmd/scaffold.go
zhangchiqing: yes, changing the interval and duration dynamically is a bit harder, maybe we can do that later in a separate PR

admin/commands/common/set_profiler_enabled.go
```diff
@@ -0,0 +1,37 @@
...
+func (s *SetProfilerEnabledCommand) Handler(ctx context.Context, req *admin.CommandRequest) (interface{}, error) {
+	enabled := req.ValidatorData.(bool)
+	debug.SetProfilerEnabled(enabled)
+	return "ok", nil
+}
+
+func (s *SetProfilerEnabledCommand) Validator(req *admin.CommandRequest) error {
+	enabled, ok := req.Data.(string)
+	if !ok {
+		return errors.New("the input must be true or false")
```
Kay-Zee, line 26: should we mention we're also looking for a string?

admin/commands/common/set_profiler_enabled.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	if enabled == "true" {
```
Kay-Zee, line 28: i guess we don't think it's worth doing strings.EqualFold?

admin/commands/common/set_profiler_enabled.go
```diff
@@ -0,0 +1,37 @@
...
+		req.ValidatorData = true
+	} else if enabled == "false" {
+		req.ValidatorData = false
+	} else {
+		return fmt.Errorf("unknown value, the input must be true or false")
```
Kay-Zee, line 33: might as well put the input here for better error messaging, no?

cmd/scaffold.go
```diff
@@ -461,14 +461,17 @@ func (fnb *FlowNodeBuilder) initMetrics() {
 func (fnb *FlowNodeBuilder) initProfiler() {
 	// note: by default the Golang heap profiling rate is on and can be set even if the profiler is NOT enabled
 	runtime.MemProfileRate = fnb.BaseConfig.profilerMemProfileRate
+
 	if !fnb.BaseConfig.profilerEnabled {
```
Kay-Zee, line 465: Don't we have to remove this line now? Otherwise the auto profiler will never get created during initialization, if we set the profiler to be false during startup

admin/commands/common/set_profiler_enabled.go
zhangchiqing: I'd like to keep it simple and stupid.

admin/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+### To change log level
+```
+curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-log-level", "data": "debug"}'
+```
+
+### To turn on profiler
+```
+curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-log-level", "data": "debug"}'
```
simonhf, line 18: Copy and paste typo? This line looks identical to the `change log level` line, or?

integration/localnet/bootstrap.go
```diff
@@ -439,13 +460,14 @@ func prepareExecutionService(container testnet.ContainerConfig, i int) Service {
 	service.Ports = []string{
 		fmt.Sprintf("%d:%d", ExecutionAPIPort+2*i, RPCPort),
 		fmt.Sprintf("%d:%d", ExecutionAPIPort+(2*i+1), SecuredRPCPort),
+		fmt.Sprintf("%d:%d", AdminToolLocalPort+n, AdminToolPort),
```
simonhf, line 463: Instead of the `+n` mechanism, why not use the `container.ContainerName` mechanism like `--ingress-addr` above? This way there is no need for the `+n` and the name and port is more predictable regardless of localnet configuration? E.g. `fmt.Sprintf("--admin-addr=%s:%d", container.ContainerName, 9002),`? Pros? Cons?

zhangchiqing: `fmt.Sprintf("%d:%d", AdminToolLocalPort+n, AdminToolPort)` will add a port mapping as `3702:9002`. 9002 is the internal port of that container, and `3702` will map that port to your local port, so that we can talk to the admin tool port `9002` via the local port `3702`. Since every node is using `9002` as admin tool port, and we only has one `9002` port, so we will map each node's 9002 to a different port without overlapping. 

If we use `container name`, how it works to access a port via container name. we can't send request to `curl http://127.0.0.1:localnet_consensus_1_1`, right?

integration/localnet/README.md
```diff
@@ -264,3 +264,11 @@ flow transactions send transfer_tokens.cdc 9999.9 <ACCOUNT_ADDRESS> -n localnet
 > replace `<ACCOUNT_ADDRESS>` in the command above with an address that you created on your localnet.
 
 After the transaction is sealed, the account with `<ACCOUNT_ADDRESS>` should have the balance increased by 9999.9 tokens.
+
+# admin tool
+The admin tool is enabled by default for all node type except access node.
+
+For instance, in order to use admin tool to change log level, first find the local port that maps to `9002` which is the admin tool address, if the local port is `3702`, then run:
```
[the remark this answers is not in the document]
zhangchiqing: >What are the steps to find the "local port"?

`docker ps -a` will show the mapping. 

>Why not implement like the --ingress-addr=collection_2:9000

That is for `node-to-node` communication. But in our case, we need `host-to-node` communication, where a client outside of the node group is sending a request to the node.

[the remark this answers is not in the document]
smnzhu: > docker ps -a will show the mapping.

Can we briefly explain this in the documentation here?

integration/localnet/bootstrap.go
```diff
@@ -331,6 +335,11 @@ func prepareService(container testnet.ContainerConfig, i int) Service {
 		},
 	}
 
+	// TODO: having trouble enabling admin tool for access node. skip Access node for now.
```
smnzhu, line 338: Can you describe the error? Were there logs? It definitely works on access nodes (Access Nodes on the current mainnet has it enabled), so it's probably something that can be fixed. I might be able to help here.

smnzhu (commented):
I would suggest passing the profiler object into the admin command constructor, instead of using a function which changes a global variable as you have here.

This makes it trivial to also support modifying the interval and duration, as @simonhf suggested.

For example, see: https://github.com/onflow/flow-go/blob/3df577c3b0eb3fdbbfb064bd22e1dfaf36bbc911/admin/commands/state_synchronization/read_execution_data.go#L72-L76

zhangchiqing:
@smnzhu the challenge is not from how we pass arguments to the admin tool, but how the admin tool apply the change to interval and duration. 

First, the admin tool doesn't have access to the AutoProfiler instance which is owned by FlowNodeBuilder object. I used a global value for controlling enable/disable the profiler, but I can't do it for the interval and duration.

`Interval` is used by the `LaunchPeriodically` function to create an internal ticker, which is not exposed to be updated. Making it dynamic would require bigger changes. 

So I decided to just let admin tool control enable/disable. And revisit this later.

integration/localnet/bootstrap.go
zhangchiqing: Access nodes kept crashing on startup, due to port to be used. I'm not sure which module is also using `9002`, so I changed to `9005`, it worked, but crashed again for a different reason, saying grpc server is fatal to start.

admin/commands/common/set_profiler_enabled.go
smnzhu: We actually don't need to use a string at all, booleans are natively supported in JSON. So I'd suggest instead using a bool:

```
enabled, ok := req.Data.(bool)
```

smnzhu (approved):

vishalchangrani (approved):

zhangchiqing:
bors merge

integration/localnet/bootstrap.go
simonhf: > If we use container name, how it works to access a port via container name. we can't send request to curl `http://127.0.0.1:localnet_consensus_1_1`, right?

I assumed we would send a curl request to `http://consensus_1:9000` and that docker would handle it. But now I'm assuming -- because of your answer below -- that this kind of domain only works inside the node group, thus making it not possible to send such a curl request outside the node group.

And this [1] from 2016 also confirms that docker hosts cannot see the internal docker network :-(

Sorry about misunderstanding and thinking that docker was more capable...

[1] https://stackoverflow.com/questions/36151981/local-hostnames-for-docker-containers

PR #2079, opened 2022-02-26
[network] Add read and write deadlines for direct messages

smnzhu:
Currently, there are no read / write deadlines for the direct message streams we use. This means that a write to a stream or a read from one could block forever.

Additionally, we were missing code to reset the stream if there was some error encountered upon writing to it.


network/p2p/middleware.go
```diff
@@ -392,17 +406,31 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
 	}
 	_, isStaked := m.ov.Identities().ByNodeID(nodeID)
 
+	ctx, cancel := context.WithTimeout(m.ctx, LargeMsgUnicastTimeout)
+
+	deadline, _ := ctx.Deadline()
```
Kay-Zee, line 411: do we not check the bool here since we already know there will be a deadline for sure?

network/p2p/middleware.go
```diff
@@ -392,17 +406,31 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
...
+	err = s.SetReadDeadline(deadline)
+	if err != nil {
+		log.Err(err).Msg("failed to set read deadline for stream")
```
Kay-Zee, line 414: is logging here sufficient? not other action needed? I guess it would just default to the same behaviour as if we didn't try to set a deadline?

network/p2p/readConnection.go
```diff
@@ -44,7 +45,6 @@ func newReadConnection(ctx context.Context,
 	}
 
 	c := readConnection{
-		ctx:        ctx,
```
Kay-Zee, line 47: what was this for?

network/p2p/middleware.go
smnzhu: Yes, that's the reason.

network/p2p/readConnection.go
smnzhu: Sorry, I think I need to add this back

vishalchangrani: this was context that the middleware passed in so that when the middleware was stopping it could tell the readConnection to stop and exit as well.

vishalchangrani: line [72](https://github.com/onflow/flow-go/pull/2079/files#diff-f8c0a7795966e3b2bf497901c5128a0e462624582e4a6a40d133f7041300c148L72) has an `rc.ctx` while would need this to be set.

network/p2p/middleware.go
smnzhu: So, there are two options here.

We could either say, if we fail to set the deadline, then we abort and close the stream immediately.

Otherwise, we could just fallback to the default behavior, which means potentially that the read / write could block forever.

I'm happy to go with either approach, but in the end I think in practice we will never fail to set the deadline. None of the implementations ever returns an error: https://github.com/libp2p/go-yamux/blob/cd22a37b789cf6e1fe4f2c2cff6da5df7a49dfac/stream.go#L457-L486

network/p2p/readConnection.go
jordanschalm: I think removing this will cause a panic here:
https://github.com/onflow/flow-go/blob/b4f37e0c537af6c437147604ebabd43e6265f1c3/network/p2p/readConnection.go#L72

Maybe we could remove it from `readConnection` completely and pass the context into `receiveLoop`

smnzhu: @jordanschalm that's a much better idea

network/p2p/middleware.go
```diff
@@ -347,6 +351,32 @@ func (m *Middleware) SendDirect(msg *message.Message, targetID flow.Identifier)
...
+	defer func() {
+		if success {
+			// close the stream immediately
+			err = stream.Close()
+			if err != nil {
+				err = fmt.Errorf("failed to close the stream for %s: %w", targetID, err)
+			}
+
+			// OneToOne communication metrics are reported with topic OneToOne
+			m.metrics.NetworkMessageSent(msg.Size(), metrics.ChannelOneToOne, msg.Type)
```
zhangchiqing, line 365: Why do we remove this logic?
```
	if _, isStaked := m.ov.Identities().ByNodeID(targetID); !isStaked {
		channel = metrics.ChannelOneToOneUnstaked
	}
```

zhangchiqing (approved):
The fix looks good. 

If it did work, I would suggest to extract the stream handling (close and reset) into a function/module to separate from the middleware module.

network/p2p/middleware.go
smnzhu: This was from previously, when we were using same libp2p host on Access Node for both staked and public network.

Now we use two separate ones, so we don't need this anymore, because the metric name already tells us whether it is staked or public network

jordanschalm (approved):

smnzhu:
> The fix looks good.
> 
> If it did work, I would suggest to extract the stream handling (close and reset) into a function/module to separate from the middleware module.

@zhangchiqing Yeah, that sounds good.

I'll port the current version first, and then work on updates on this version which will go into master

smnzhu:
bors merge

PR #2087, opened 2022-02-28
[Observer] Add bootstrap command to generate observer networking key

peterargue:
This adds a new command to generate the networking key used by observers on the public libp2p network.

Note: this key uses a different format from the normal staked networking key.

### DIFF
--- a/cmd/bootstrap/README.md
+++ b/cmd/bootstrap/README.md
@@ -80,7 +80,7 @@ go run -tags relic ./cmd/bootstrap key --address "example.com:1234" --role "cons
      but is not required at node start
 
 
-### Phase 2: Generating final root information
+## Phase 2: Generating final root information
 
 This step will generate the entire root information for all nodes (incl. keys for all Dapper-controlled nodes).
 
@@ -147,3 +147,15 @@ go run -tags relic ./cmd/bootstrap finalize \
    - root Quorum Certificate for `ClusterBlockProposal` for collector cluster with ID `<ClusterID>`
    - REQUIRED at NODE START by all collectors of the respective cluster
    - file can be made accessible to all nodes at boot up (or recovery after crash)
+
+## Generating networking key for Observer
+
+This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for observers.
+
+```bash
+go run -tags relic ./cmd/bootstrap observer-network-key  -o ./path/
+```
+
+In this example, the following file will be generated: `./path/network-key`
+
+This key must be kept secret as it's used to encrypt and sign network requests sent by the observers.
--- a/cmd/bootstrap/README.md
+++ b/cmd/bootstrap/README.md
@@ -150,7 +150,7 @@ go run -tags relic ./cmd/bootstrap finalize \
 
 ## Generating networking key for Observer
 
-This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for observers.
+This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for Observers.
 
 ```bash
 go run -tags relic ./cmd/bootstrap observer-network-key  -o ./path/
--- a/cmd/bootstrap/README.md
+++ b/cmd/bootstrap/README.md
@@ -153,9 +153,7 @@ go run -tags relic ./cmd/bootstrap finalize \
 This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for Observers.
 
 ```bash
-go run -tags relic ./cmd/bootstrap observer-network-key  -o ./path/
+go run -tags relic ./cmd/bootstrap observer-network-key  -f ./path/network-key
 ```
-
-In this example, the following file will be generated: `./path/network-key`
 
 This key must be kept secret as it's used to encrypt and sign network requests sent by the observers.
--- a/cmd/bootstrap/cmd/observer_network_key.go
+++ b/cmd/bootstrap/cmd/observer_network_key.go
@@ -8,7 +8,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/onflow/flow-go/cmd/bootstrap/utils"
-	"github.com/onflow/flow-go/crypto"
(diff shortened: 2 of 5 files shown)


cmd/bootstrap/cmd/observer_network_key.go
```diff
@@ -0,0 +1,62 @@
...
+	seed := validateSeed(flagNetworkSeed)
+	networkKey, err := utils.GenerateUnstakedNetworkingKey(seed)
+	if err != nil {
+		log.Fatal().Err(err).Msg("could not generate network key")
+	}
+	log.Info().Msg("generated network key")
+
+	// hex encode and write to file
+	output := make([]byte, hex.EncodedLen(crypto.PrKeyLenECDSASecp256k1))
+	hex.Encode(output, networkKey.Encode())
```
m4ksio, line 59: Seems that internal implementation detail (`crypto.PrKeyLenECDSASecp256k1` ) is leaking here, we should just encode generated key without reallocating any lengths

cmd/bootstrap/README.md
```diff
@@ -147,3 +147,15 @@ go run -tags relic ./cmd/bootstrap finalize \
    - root Quorum Certificate for `ClusterBlockProposal` for collector cluster with ID `<ClusterID>`
    - REQUIRED at NODE START by all collectors of the respective cluster
    - file can be made accessible to all nodes at boot up (or recovery after crash)
+
+## Generating networking key for Observer
+
+This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for observers.
```
smnzhu, line 153:
```suggestion
This generates the networking key used by observers to connect to the public libp2p network. It is a different key format than staked nodes and should only be used for Observers.
```

cmd/bootstrap/cmd/observer_network_key.go
```diff
@@ -0,0 +1,61 @@
...
+	"path"
+
+	"github.com/spf13/cobra"
+
+	"github.com/onflow/flow-go/cmd/bootstrap/utils"
+	model "github.com/onflow/flow-go/model/bootstrap"
+)
+
+// observerNetworkKeyCmd represents the `observer-network-key` command which generates required network key
+// for an observer,  and writes it to the default path within the provided directory. Used by new observer
```
smnzhu, line 15:
```suggestion
// for an Observer,  and writes it to the default path within the provided directory. Used by new Observer
```

cmd/bootstrap/cmd/observer_network_key.go
```diff
@@ -0,0 +1,61 @@
...
+	// if the file already exists, exit
+	keyExists, err := pathExists(path.Join(flagOutdir, model.FilenameObserverNetworkKey))
+	if err != nil {
+		log.Fatal().Err(err).Msg("could not check if network-key exists")
+	}
+
+	if keyExists {
+		log.Warn().Msg("network key already exists, exiting...")
+		return
+	}
```
smnzhu, line 46: nit: I think it would be more clear if the messages here were something more along the lines of "could not check if <file_name> exists" or "<file_name> already exists, exiting...", replacing with whatever file name the user chose

cmd/bootstrap/README.md
```diff
@@ -147,3 +147,15 @@ go run -tags relic ./cmd/bootstrap finalize \
...
+
+```bash
+go run -tags relic ./cmd/bootstrap observer-network-key  -o ./path/
+```
+
+In this example, the following file will be generated: `./path/network-key`
```
smnzhu, line 159: Instead of hardcoding a file name, might be better to just let the user specify the entire file name instead of just the path.

smnzhu (approved):
Overall good, a few small comments

cmd/bootstrap/README.md
peterargue: yea. that's better. I thought the outputDir command was required, but looks like it's not. so I'll add a new flag for the file path

cmd/bootstrap/cmd/observer_network_key.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// generate unstaked networking private key
+	networkKey, err := utils.GenerateUnstakedNetworkingKey(flagNetworkSeed)
+	if err != nil {
+		log.Fatal().Err(err).Msg("could not generate network key")
+	}
+	log.Info().Msg("generated network key")
+
+	// hex encode and write to file
+	output := make([]byte, hex.EncodedLen(networkKey.Size()))
```
m4ksio, line 68: Is `.Size()` returning encoded length? 
I'm not sure if we need really to do this pre-allocation, can't we just encode whatever we get?
This might lead to some subtle bugs in the future

m4ksio (approved):

cmd/bootstrap/cmd/observer_network_key.go
peterargue: fair enough. I updated to use the actual slice len. the preallocation is needed. that's actually what the [hex library does](https://cs.opensource.google/go/go/+/master:src/encoding/hex/hex.go;l=106-109?q=hex.go&ss=go%2Fgo:src%2F) for `EncodeToString` except without the cast

cmd/bootstrap/cmd/observer_network_key.go
```diff
@@ -0,0 +1,79 @@
...
+		log.Fatal().Err(err).Msgf("could not check if %s exists", flagOutputFile)
+	}
+
+	if keyExists {
+		log.Warn().Msgf("%s already exists, exiting...", flagOutputFile)
+		return
+	}
+
+	// generate unstaked networking private key
+	networkKey, err := utils.GenerateUnstakedNetworkingKey(flagNetworkSeed)
```
tarakby, line 61: What are "observers" exactly? are these a sub-type of unstaked access nodes?
I'm asking to make sure `GenerateUnstakedNetworkingKey` is what we need and not `GenerateNetworkingKey`.

peterargue: Observer is the new name for unstaked access node.

tarakby: ah okay 😅  It makes sense to use that function then.

peterargue:
bors merge

Issue #2081, opened 2022-02-26
[Network] Enable per-topic message size and timeout limits


synzhu:
Context: https://github.com/onflow/flow-go/blob/f78981ee3d43aeeda13969c03f025c3a8d1db9aa/network/p2p/middleware.go#L424-L428 and https://github.com/onflow/flow-go/blob/f78981ee3d43aeeda13969c03f025c3a8d1db9aa/network/p2p/middleware.go#L462-L463

In addition to registering a different protocol / stream handler per topic, we also need to make sure to send to the correct protocol for outgoing messages.

PR #2091, opened 2022-03-01
[Network] Direct Message API

smnzhu:
TODO: 
- [x] Set custom direct message configs for large message channels (e.g chunk data pack)

closes https://github.com/onflow/flow-go/issues/2081


Kay-Zee:
Very outdated, either re-open or move to a separate PR if we still need the contents

PR #2093, opened 2022-03-01
[Observability] Adds metrics for HeroCache

yhassanzadeh13:
Implements metrics module for the `HeroCache`-based mempools.


module/mempool/herocache/dns_cache.go
```diff
@@ -16,12 +17,23 @@ type DNSCache struct {
...
+			stdmap.WithBackData(
+				herocache.NewCache(
+					sizeLimit,
+					herocache.DefaultOversizeFactor,
+					heropool.LRUEjection,
+					logger,
+					txtCollector))),
+		ipCache: stdmap.NewBackend(
+			stdmap.WithBackData(
+				herocache.NewCache(sizeLimit,
```
simonhf, line 32: Any reason why `sizeLimit` is not on its own line?

module/metrics.go
```diff
@@ -94,13 +94,13 @@ type CleanerMetrics interface {
...
+	// CacheEntries report the total number of cached items
 	CacheEntries(resource string, entries uint)
-	// report the number of times the queried item is found in the cache
+	// CacheHit report the number of times the queried item is found in the cache
 	CacheHit(resource string)
-	// report the number of items the queried item is not found in the cache, nor found in the database
+	// CacheNotFound report the number of items the queried item is not found in the cache, nor found in the database
 	CacheNotFound(resource string)
-	// report the number of items the queried item is not found in the cache, but found in the database
+	// CacheMiss report the number of items the queried item is not found in the cache, but found in the database
```
simonhf, line 103: s~(number of) items~$1 times~ ? :-)

module/metrics.go
```diff
@@ -94,13 +94,13 @@ type CleanerMetrics interface {
...
 
 type CacheMetrics interface {
-	// report the total number of cached items
+	// CacheEntries report the total number of cached items
 	CacheEntries(resource string, entries uint)
-	// report the number of times the queried item is found in the cache
+	// CacheHit report the number of times the queried item is found in the cache
 	CacheHit(resource string)
-	// report the number of items the queried item is not found in the cache, nor found in the database
+	// CacheNotFound report the number of items the queried item is not found in the cache, nor found in the database
```
simonhf, line 101: s~(number of) items~$1 times~ ? :-)

module/metrics.go
```diff
@@ -424,3 +424,29 @@ type PingMetrics interface {
...
+
+type HeroCacheMetrics interface {
+	// BucketAvailableSlots keeps track of number of available slots in buckets of cache.
+	BucketAvailableSlots(uint64, uint64)
+
+	// OnSuccessfulWrite is called whenever a new entity is successfully added to the cache.
+	OnSuccessfulWrite()
+
+	// OnEntityEjectedAtFullCapacity is called whenever adding a new entity to the cache results in ejection of another entity.
+	// This normally happens when the cache is full.
```
simonhf, line 436: s~(This normally happens) (when the cache is full)~$1 -- and is expected -- $2~ ? :-)

module/metrics.go
```diff
@@ -424,3 +424,29 @@ type PingMetrics interface {
...
+	OnEntityEjectedAtFullCapacity()
+
+	// OnEmergencyKeyEjection is called whenever a bucket is found full and all of its keys are valid.
+	// Hence, adding a new entity to that bucket will replace the oldest valid key inside that bucket.
```
simonhf, line 440: I've noticed a couple of comments where the first reference to the `key` is called `entity` and the second reference to the `key` is called `key`. I imagine this makes things more confusing for some readers? What about consistently referring to keys as entities or vice-versa? Or whenever talking about a key as an entity then always refer to it as e.g. `entity (i.e. key in heroCache terminology)`?

So this sentence might be the following:

`Hence, adding a new entity [1] to that bucket will replace the oldest valid entity [1] inside that bucket.`
`[1] note: entity is a key in heroCache terminology`

module/metrics.go
```diff
@@ -424,3 +424,29 @@ type PingMetrics interface {
...
+	OnEmergencyKeyEjection()
```
simonhf, line 441: This `OnEmergencyKeyEjection()` function refers to `Key`, but above the similar `OnEntityEjectedAtFullCapacity()` function refers to `Entity`. Better to have a consistent referring to the same thing?

module/metrics/herocache.go
```diff
@@ -0,0 +1,163 @@
...
+
+	successfulReadCount   prometheus.Counter
+	unsuccessfulReadCount prometheus.Counter
+
+	successfulWriteCount   prometheus.Counter
+	unsuccessfulWriteCount prometheus.Counter
+
+	fullCapacityEntityEjectionCount prometheus.Counter
+	emergencyKeyEjectionCount       prometheus.Counter
+}
```
simonhf, line 24: Similar to above, consider more structured names, e.g.:
```
	CountKeyGetSuccess prometheus.Counter
	CountKeyGetFailure prometheus.Counter

	CountKeyPutSuccess prometheus.Counter
	CountKeyPutFailure prometheus.Counter

	CountKeyEjectionAtFullCapacity prometheus.Counter
	CountKeyEjectionDueToEmergency prometheus.Counter
```

module/metrics/herocache.go
```diff
@@ -0,0 +1,163 @@
...
+func CollectionNodeTransactionsCacheMetrics(registrar prometheus.Registerer, epoch uint64) *HeroCacheCollector {
+	return NewHeroCacheCollector(namespaceCollection, fmt.Sprintf("%s_%d", ResourceTransaction, epoch), registrar)
+}
+
+func NewHeroCacheCollector(nameSpace string, cacheName string, registrar prometheus.Registerer) *HeroCacheCollector {
+
+	normalizedBucketSlotAvailableHistogram := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: nameSpace,
+		Subsystem: subsystemHeroCache,
+		Buckets:   []float64{0, 0.05, 0.1, 0.25, 0.5, 0.75, 1},
```
simonhf, line 49: Is it worth adding a comment about how these floating point numbers work?

Here's my assumption about how they work and how they would relate to heroCache:

heroCache generally uses 16 slots per bucket and therefore the available slots per bucket will be a number ranging from zero to 16, i.e. 17 different possible values. Whereas, the floats represent 7 possible values / prometheus 'buckets' not to be confused with heroCache 'buckets' :-) Have these 7 possible values been carefully chosen and if so, should this choosing process be commented on too?

So for the zero to 17 available slots:
* zero will map to 0,
* 1 will map to 0.625, <-- will this hit or miss the 0.05 bucket?
* 2 will map to 0.125, <-- will this hit or miss the 0.1 bucket?
* 3 will map to 0.1875,
* 4 will map to 0.25, <-- will this hit or miss the 0.25 bucket?
...
* 17 will map to 1.

Is it intended that the available slots 0, 1 , and 2 get mapped to prometheus buckets 0, 0.05, and 0.1?

Another question: Since the difference between 17 possible values and 7 buckets is only 10 then why not have 17 buckets and match the default number of heroCache slots?

module/mempool/herocache/dns_cache.go
yhassanzadeh13: fixed https://github.com/onflow/flow-go/pull/2093/commits/15f1b2ec9914a42214993228d49bdbc41c000cb1

module/metrics.go
yhassanzadeh13: This is not within my context, I just fixed a `godoc` to be complaint with Golang best practices, i.e., `godoc` must start with the name of its corresponding program entity. In this case, it must start with `CacheMiss`. That is the only change it is applying. Any other comment is better consulted with the original developer.

module/metrics.go
yhassanzadeh13: https://github.com/onflow/flow-go/commit/7e88c0596bcc0c137114a5a2d024915a5b481201

module/metrics/herocache.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2093/commits/ce5f7bf2d1adfddda0c749a6bde2aa922883a70d

module/metrics/herocache.go
yhassanzadeh13: Added documentation here: https://github.com/onflow/flow-go/pull/2093/commits/2af1b4de49cf63fdc968e8da2690ca8925d92205

> Another question: Since the difference between 17 possible values and 7 buckets is only 10 then why not have 17 buckets and match the default number of heroCache slots?

Each bucket adds a time series per collector. Currently, we have 6 buckets, and adding 11 more means 11 more time-series per mempool that utilizes HeroCache. Currently 4 mempools use `HeroCache` and hence it adds 44 extra timeseries. This can simply go beyond 100 more timeseries when we switch more mempools to `HeroCache`. Not a scalable solution.

module/metrics.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2093/commits/ee25f8d0ea77ce4c4a014170bfdb71209409646b

module/metrics.go
simonhf: FYI I'm only suggesting that there is a simple typo :-)
```
// CacheMiss report the number of items the queried item is not found in the cache, but found in the database
// CacheMiss report the number of times the queried item is not found in the cache, but found in the database
```

module/metrics.go
simonhf: In the other comment, I requested the original developer @zhangchiqing to have a quick look (@jordanschalm the other original developer is on vacation AFAIK).

module/metrics.go
zhangchiqing: yes, correct, should be `number of times`.

yhassanzadeh13: fixed in https://github.com/onflow/flow-go/pull/2093/commits/be72fd8857db97c13673034623438c30895ac6fd.

module/metrics.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2093/commits/2cc141ff142bd3b3550643a33829df81db8cac8e

simonhf (approved):
Thanks for the name refactoring :-) The specifically metrics related changes LGTM.

However, the non-metrics related changes are a bit out of my wheelhouse. Would be good if another two reviewers could review that who are more familiar with the code? :-)

cmd/access/node_builder/access_node_builder.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/onflow/flow-go/module/metrics"
```
zhangchiqing, line 39: Asking this question here, which has nothing to do this line of code. Just so, we can discuss in this thread.

My question is: What would be the performance implication for this change?

The reason I ask is that, a while ago we [discovered EN's memory was abused](https://github.com/dapperlabs/flow-go/issues/5952#issue-1031456155) from [the report ](https://www.notion.so/dapperlabs/Examining-Leo-s-multi-fork-CPU-spike-issue-f7cecbfb9c3c40409bc258a09f4aa13e#56c2880fd2d64332a198dce3fde5d4bc)that @simonhf made.

Both the top `alloc-space` and `alloc-objects` is `CacheHit`:
```
github.com/onflow/flow-go/module/metrics.(*CacheCollector).CacheHit
```

To fix that, [I disabled the metrics for cache](https://github.com/onflow/flow-go/pull/1518/files).

Now this PR is adding metrics for a different cache, which reminds me that issue. 

I think the reason it was an issue before is that whenever the cache API is visited, we will call into metrics API which somehow allocates objects on heap, as cache API is called frequently, it causes memory abuse.

To mitigate that, maybe we could record metrics less frequently? If we were to record `Size()` of a cache, instead of measuring once for every single call, we could record once for every 100 calls. However, we would have to keep track of the state of total called number, and ensure concurrent-safe, which would be a bit more complex.

Thoughts?

cc @Kay-Zee

zhangchiqing, line 39: Asking this question here, which has nothing to do this line of code. Just so, we can discuss in this thread.

My question is: What would be the performance implication for this change?

The reason I ask is that, a while ago we [discovered EN's memory was abused](https://github.com/dapperlabs/flow-go/issues/5952#issue-1031456155) from [the report ](https://www.notion.so/dapperlabs/Examining-Leo-s-multi-fork-CPU-spike-issue-f7cecbfb9c3c40409bc258a09f4aa13e#56c2880fd2d64332a198dce3fde5d4bc)that @simonhf made.

Both the top `alloc-space` and `alloc-objects` is `CacheHit`:
```
github.com/onflow/flow-go/module/metrics.(*CacheCollector).CacheHit
```

To fix that, [I disabled the metrics for cache](https://github.com/onflow/flow-go/pull/1518/files).

Now this PR is adding metrics for a different cache, which reminds me that issue. 

I think the reason it was an issue before is that whenever the cache API is visited, we will call into metrics API which somehow allocates objects on heap, as cache API is called frequently, it causes memory abuse.

To mitigate that, maybe we could record metrics less frequently? If we were to record `Size()` of a cache, instead of measuring once for every single call, we could record once for every 100 calls. However, we would have to keep track of the state of total called number, and ensure concurrent-safe, which would be a bit more complex.

Thoughts?

cc @Kay-Zee

yhassanzadeh13: @zhangchiqing per our conversation, I introduced a flag in https://github.com/onflow/flow-go/pull/2093/commits/1b389f31f211328c775674935999b982cfcc935f for enabling the `HeroCache` metrics collection. The flag is off by default, so there will be no heap overhead due to metrics collection, and we can turn it on for incident management and firefighting. Moreover I issued up [this ticket](https://app.zenhub.com/workspaces/flow-protocol-5a2ad022cee40e25b0885f82/issues/dapperlabs/flow-go/6206) to further benchmark and investigate the heap footprint of the `HeroCacheCollector`.

zhangchiqing (approved):

PR #2097, opened 2022-03-03
[FVM] Change fee deduction logic

janezpodhostnik:
# Fee deduction logic change

The [FLIP: Variable transaction fees](https://github.com/onflow/flow/pull/660) specifies that transaction fees will be calculated as `Fees = surgeFactor ( inclusionEffortCost * InclusionEffort + executionEffortCost * executionEffort)`. This calculation happens in the FlowFees core contract with the following [PR](https://github.com/onflow/flow-core-contracts/pull/279).

The FVM needs to provide the inclusionEffort and the executionEffort. 

This PR:
- imports and uses the changes to the the `FlowFees` core contract
- transaction fee deduction was changed to use `FlowFees` 
- sets the fee parameters so the fees are still a static 0.0001
  - sets the default (on bootstrap) `surgeFactor` to 1.0
  - sets the default (on bootstrap) `inclusionEffortCost` to 0.0001 and the static `inclusionEffort` to 1 0000 0000 (or 1.0)
  - sets the default `executionEffortCost` to 0.0 and the `executionEffort` to "computation" (which will be renamed to executionEffort)

The reason that the `executionEffortCost` is 0.0 and `executionEffort` is set to computation is that the computation usage will start appearing in the new `FlowFees.FeesDeducted` event emitted.

When switching over to dynamic execution fees the `inclusionEffortCost` and `executionEffortCost` will change from 0.0001, 0.0 to 0.00005, something non-zero.

## Breaking change ⚠️ 

`TransactionFeeDeductionFailedError` No longer contains the amount of fees that were supposed to be deducted. That number is calculated within the `FlowFees.deductTransactionFee` contract function and is not available to the FVM.


fvm/transactionInvoker.go
```diff
@@ -261,16 +263,12 @@ func (i *TransactionInvoker) deductTransactionFees(env *TransactionEnv, proc *Tr
 	}()
 
 	deductTxFees := DeductTransactionFeesInvocation(env, proc.TraceSpan)
-	_, err = deductTxFees(proc.Transaction.Payer)
+	// Hardcoded inclusion effort (of 1.0). Eventually this will be dynamic.
+	// Execution effort will be connected to computation used.
+	_, err = deductTxFees(proc.Transaction.Payer, 1_0000_0000, computationUsed)
```
ramtinms, line 268: 1_0000_0000 looks a bit odd to me, the use of 4 zeros instead of 3 zeros is not really human readable.

fvm/systemcontracts/system_contracts.go
```diff
@@ -153,6 +159,9 @@ var (
 	stakingContractAddressMainnet = flow.HexToAddress("8624b52f9ddcd04a")
 	// stakingContractAddressTestnet is the address of the FlowIDTableStaking contract on Testnet and Canary
 	stakingContractAddressTestnet = flow.HexToAddress("9eca2b38b18b5dfe")
+
+	flowFeesContractAddressMainnet = flow.HexToAddress("f919ee77447b7497")
+	flowFeesContractAddresTestnet  = flow.HexToAddress("912d5440f7e3769e")
 )
```
ramtinms, line 165: Why do we need to include these fee contracts into the system_contracts?

fvm/fvm_test.go
```diff
@@ -3085,7 +3085,7 @@ func TestTransactionFeeDeduction(t *testing.T) {
 		checkResult   func(t *testing.T, balanceBefore uint64, balanceAfter uint64, tx *fvm.TransactionProcedure)
 	}
 
-	txFees := fvm.DefaultTransactionFees.ToGoValue().(uint64)
+	txFees := uint64(1_0000)
```
ramtinms, line 3088: same here, 4 zeros is not human readable

fvm/blueprints/fees.go
```diff
@@ -46,18 +46,17 @@ func DeployStorageFeesContractTransaction(service flow.Address, contract []byte)
...
-transaction(transactionFee: UFix64, accountCreationFee: UFix64, minimumStorageReservation: UFix64, storageMegaBytesPerReservedFLOW: UFix64, restrictedAccountCreationEnabled: Bool) {
+transaction(accountCreationFee: UFix64, minimumStorageReservation: UFix64, storageMegaBytesPerReservedFLOW: UFix64, restrictedAccountCreationEnabled: Bool) {
     prepare(service: AuthAccount) {
         let serviceAdmin = service.borrow<&FlowServiceAccount.Administrator>(from: /storage/flowServiceAdmin)
             ?? panic("Could not borrow reference to the flow service admin!");
 
         let storageAdmin = service.borrow<&FlowStorageFees.Administrator>(from: /storage/storageFeesAdmin)
             ?? panic("Could not borrow reference to the flow storage fees admin!");
 
-        serviceAdmin.setTransactionFee(transactionFee)
```
ramtinms, line 60: I'm not sure if I understand why this has been added here, considering this is only activated when we create a new account?

fvm/systemcontracts/system_contracts.go
janezpodhostnik: In the [core contract changes](https://github.com/onflow/flow-core-contracts/pull/279) a new fee deduction function was added to the `FlowFees` contract (the old function was kept in place so that we can switch to the new calculation without braking the old one once the contract is updated). 

Since the fee deduction is now in FlowFees, we need the address of FlowFees to be known.

ramtinms: yeah but system contract is mostly used for epoch stuff, so they had no choice to hard code them here, I wonder if it should be computed on the fly or at least be in a different place

utils/unittest/execution_state.go
```diff
@@ -24,7 +24,7 @@ const ServiceAccountPrivateKeySignAlgo = crypto.ECDSAP256
 const ServiceAccountPrivateKeyHashAlgo = hash.SHA2_256
 
 // Pre-calculated state commitment with root account with the above private key
-const GenesisStateCommitmentHex = "d3518010ead6ce14c3994005fd8121691efece28232c8853ad60eaeccaed4fd2"
+const GenesisStateCommitmentHex = "4981dcc4690b2efe272ef9e5283311bf2e71472919c01fb71405fe86f991c2c2"
```
ramtinms, line 27: did we check what is changing genesis state commitment, is it the new content for fee contract?

fvm/systemcontracts/system_contracts.go
janezpodhostnik: I think it will always be the 4th address (i would have to verify). Would it be better to use that?

utils/unittest/execution_state.go
janezpodhostnik: Yes, this is because of the new contract. Not sure how I could verify that it is **only** because of the new contract.

SupunS (approved):
I mostly reviewed the general Go code. Looks fine to me.

ramtinms (approved):

janezpodhostnik:
bors merge

janezpodhostnik:
bors merge

PR #2101, opened 2022-03-03
[Consensus] SignerIndices Optimization (2/4)

zhangchiqing:
Build on top of https://github.com/onflow/flow-go/pull/2047

This PR implements the SignerIndices optimization.

Note:
1. The EpochCommit event from smart contract uses VoterID, therefore, I have to introduce a new data type for QC: `QuorumCertificateWithSignerIDs`. See reasoning in comment
2. The unit test cases are mostly fixed, but integration tests are not, because the [CollectionGuarantee also contains SignerIDs](https://github.com/onflow/flow-go/pull/2101/files#diff-16362fa5185fd4c29577be51a8b7b91eeb9178c8fc43ff4380d444cf28d294bbR14).  [See discussion here.](https://github.com/dapperlabs/flow-go/issues/6181). Will be implemented in separate PR.
3. `make lint` passes


module/finalizer/collection/finalizer.go
```diff
@@ -150,8 +150,10 @@ func (f *Finalizer) MakeFinal(blockID flow.Identifier) error {
 				Guarantee: flow.CollectionGuarantee{
 					CollectionID:     payload.Collection.ID(),
 					ReferenceBlockID: payload.ReferenceBlockID,
-					SignerIDs:        step.ParentVoterIDs,
-					Signature:        step.ParentVoterSigData,
+					// TODO use signer indices for collection guarantee signer IDs
```
zhangchiqing, line 153: https://github.com/dapperlabs/flow-go/issues/6181

consensus/hotstuff/signature/packer.go
```diff
@@ -30,50 +30,24 @@ func NewConsensusSigDataPacker(committees hotstuff.Committee) *ConsensusSigDataP
...
-		if ok {
-			signerIDs = append(signerIDs, beaconSigner)
-			sigTypes = append(sigTypes, hotstuff.SigTypeRandomBeacon)
-		} else {
-			return nil, nil, fmt.Errorf("random beacon signer %v not found in the committee at block: %v", beaconSigner, blockID)
-		}
-	}
-
-	// serialize the sig type for compaction
-	serialized, err := serializeToBitVector(sigTypes)
```
zhangchiqing, line 70: moved to `encodeSignerIndicesAndSigType`

consensus/hotstuff/signature/packer.go
```diff
@@ -85,59 +59,24 @@ func (p *ConsensusSigDataPacker) Pack(blockID flow.Identifier, sig *hotstuff.Blo
...
-				signerID, blockID, model.ErrInvalidFormat)
-		}
-
-		if sigType == hotstuff.SigTypeStaking {
-			stakingSigners = append(stakingSigners, signerID)
-		} else if sigType == hotstuff.SigTypeRandomBeacon {
-			randomBeaconSigners = append(randomBeaconSigners, signerID)
-		} else {
-			return nil, fmt.Errorf("unknown sigType %v, %w", sigType, model.ErrInvalidFormat)
-		}
```
zhangchiqing, line 140: all moved to `decodeSignerIndicesAndSigType`

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,88 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+	return signerIndices, serialized, nil
+}
+
+// decodeSignerIndicesAndSigType decodes sigType and use it to split the given signerIDs into two groups: staking sigers and random beacon signers.
+// it returns model.ErrInvalidFormat if decode failed or the decoded data doesn't match with the given signer IDs.
+func decodeSignerIndicesAndSigType(signerIDs []flow.Identifier, sigType []byte) ([]flow.Identifier, []flow.Identifier, error) {
+	// deserialize the compact sig types
+	sigTypes, err := deserializeFromBitVector(sigType, len(signerIDs))
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to deserialize sig types from bytes: %w", err)
```
zhangchiqing, line 229: `deserializeFromBitVector` returns `model.ErrInvalidFormat`, so no need to wrap

cmd/bootstrap/cmd/seal.go
```diff
@@ -43,9 +44,24 @@ func constructRootResultAndSeal(
 		RandomSource:       getRandomSource(flagBootstrapRandomSeed),
 	}
 
+	qcsWithSignerIDs := make([]*flow.QuorumCertificateWithSignerIDs, 0, len(clusterQCs))
```
zhangchiqing, line 47: ## Why adding `QuorumCertificateWithSignerIDs` type?

The smart contract emits a EpochCommit Event, in which the [ClusterQC contains `VoterIDs`](https://github.com/onflow/flow-go/blob/master/model/convert/service_event.go#L358-L363), and we need to convert the cadence value into golang struct to be used in protocol. It's better to keep the data type of `ClusterQC` similar to the same struct in smart contract, therefore,[ I had to revert the change to `ClusterQC` in order to keep using `VoterIDs` field,](https://github.com/onflow/flow-go/pull/2101/files#diff-1efaf8fd528651c4c22e9586d64243dae87ddb4efb83e5c175f621153f18852aR132), which makes the conversion side-effect free.

However the QC now uses SignerIndices, which means `flow.ClusterQCVoteDatasFromQCs` can't convert directly QC into ClusterQC. 

I could change `flow.ClusterQCVoteDatasFromQCs` to let it take `(qc, signerIDs)` , but it looks weird that the two inputs seems have no relation, but in fact the signerIDs must be converted from qc.SignerIndices.

In order to make it a bit type-safe, I decided to define the `QuorumCertificateWithSignerIDs`. Now the caller of `ClusterQCVoteDatasFromQCs` is responsible to convert QC into `QuorumCertificateWithSignerIDs`, such that clusterQC can be converted from it. 

The drawback of this approach is that now we have two QC definition `QuorumCertificate` and `QuorumCertificateWithSignerIDs`. Also weird. 

Thoughts?

model/flow/collectionGuarantee.go
```diff
@@ -9,10 +9,11 @@ import (
 // CollectionGuarantee is a signed hash for a collection, which is used
 // to announce collections to consensus nodes.
 type CollectionGuarantee struct {
-	CollectionID     Identifier       // ID of the collection being guaranteed
-	ReferenceBlockID Identifier       // defines expiry of the collection
-	SignerIDs        []Identifier     // list of guarantors
-	Signature        crypto.Signature // guarantor signatures
+	CollectionID     Identifier // ID of the collection being guaranteed
+	ReferenceBlockID Identifier // defines expiry of the collection
+	// TODO: to be replaced with signer indices, and add cluster ID
```
zhangchiqing, line 14: See https://github.com/dapperlabs/flow-go/issues/6181

state/protocol/convert.go
```diff
@@ -66,14 +67,26 @@ func ToEpochCommit(epoch Epoch) (*flow.EpochCommit, error) {
...
+	qcs := make([]*flow.QuorumCertificateWithSignerIDs, 0, len(clustering))
 	for i := range clustering {
 		cluster, err := epoch.Cluster(uint(i))
 		if err != nil {
 			return nil, fmt.Errorf("could not get epoch cluster (index=%d): %w", i, err)
 		}
-		qcs = append(qcs, cluster.RootQC())
+		qc := cluster.RootQC()
+		// TODO: double check cluster.Members returns canonical order
+		signerIDs, err := packer.DecodeSignerIdentifiersFromIndices(cluster.Members().NodeIDs(), qc.SignerIndices)
```
jordanschalm, line 78: Good catch. There is some inconsistency here we may want to resolve:

* `Cluster.Members` returns identities in the order [specified by the cluster assignment list](https://github.com/onflow/flow-go/blob/c59b5e4de1b59766d352209dbc856068c03f9225/model/flow/cluster.go#L59-L64) (respects order of `EpochSetup` event).
* `Epoch.InitialIdentities` returns identities in the [canonical order](https://github.com/onflow/flow-go/blob/bd910749460f48d77823af422bf8baeb798b2b13/model/convert/service_event.go#L317)
* `Snapshot.Identities` returns identities in the [canonical order](https://github.com/onflow/flow-go/blob/bd910749460f48d77823af422bf8baeb798b2b13/state/protocol/badger/snapshot.go#L227)

state/protocol/inmem/epoch.go
```diff
@@ -155,6 +156,7 @@ func (es *committedEpoch) Cluster(index uint) (protocol.Cluster, error) {
 		return nil, fmt.Errorf("failed to generate clustering: %w", err)
 	}
 
+	// TODO: double check ByIndex returns canonical order
 	members, ok := clustering.ByIndex(index)
```
jordanschalm, line 160: ⚠️ Currently it respects the assignment order in the `EpochSetup` event, which is not the same as the canonical ordering used by eg. `Snapshot.Identities`. (see previous comment)

zhangchiqing: Yes, will be addressed in separate PR

module/packer/signer_indices.go
```diff
@@ -0,0 +1,65 @@
...
+		if byt&mask > 0 {
+			signerIndices = append(signerIndices, index)
+		}
+	}
+
+	// remaining bits (if any), they must be all `0`s
+	remainings := byt << (8 - offset)
+	if remainings != byte(0) {
+		return nil, fmt.Errorf("the remaining bits are expected to be all 0s, but are %v", remainings)
+	}
```
zhangchiqing, line 58: referred to this function
https://github.com/onflow/flow-go/blob/master/consensus/hotstuff/signature/packer.go#L211-L227

cmd/bootstrap/cmd/seal.go
AlexHentschel: 👍 
I like this option, because we anyway package the bootstrapping information on its own. By allowing the bootstrapping information to be in its own custom format, we gain flexibility.

model/flow/quorum_certificate.go
```diff
@@ -25,3 +25,13 @@ type QuorumCertificate struct {
 	// from all signers.
 	SigData []byte
 }
+
+// QuorumCertificateWithSignerIDs is a QuorumCertificate model with signer IDs instead of indices.
+// it is needed for backward-compatibility with the FlowEpoch smart contract which generates QCs
+// using signer IDs.
+type QuorumCertificateWithSignerIDs struct {
```
AlexHentschel, line 32: To me, it would make sense to move this struct definition somewhere to the bootstrapping folder. This would highlight that this data structure is not used to for building blocks and needs explicit post-processing (as much of the bootstrapping data)

model/flow/quorum_certificate.go
```diff
@@ -25,3 +25,13 @@ type QuorumCertificate struct {
 	// from all signers.
 	SigData []byte
 }
+
+// QuorumCertificateWithSignerIDs is a QuorumCertificate model with signer IDs instead of indices.
+// it is needed for backward-compatibility with the FlowEpoch smart contract which generates QCs
+// using signer IDs.
```
AlexHentschel, line 31:
```suggestion
// QuorumCertificateWithSignerIDs is a QuorumCertificat, where the signing nodes are identified via their `flow.Identifier`s instead of indices. Working with IDs as opposed to 
// indices is less efficient, but simpler, because we don't require a canonical node order. 
// It is used for bootstrapping new Epochs, because the FlowEpoch smart contract has no
// notion of node ordering. 
```

module/packer/signer_ids.go
```diff
@@ -0,0 +1,49 @@
+package packer
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// EncodeSignerIdentifiersToIndices encodes the given signerIDs into compacted bit vector
```
AlexHentschel, line 9:
```suggestion
// EncodeSignerIdentifiersToIndices encodes the given signerIDs into compacted bit vector.
```

module/packer/signer_ids.go
```diff
@@ -0,0 +1,49 @@
...
+// fullIdentities represents all identities who are eligible to sign the given resource. It excludes
+// identities who are ineligible to sign the given resource. For example, fullIdentities in the context
+// of a cluster consensus quorum certificate would include authorized members of the cluster and
+// exclude ejected members of the cluster, or unejected collection nodes from a different cluster.
```
AlexHentschel, line 13:
```suggestion
// exclude ejected members of the cluster, or any collection nodes from a different cluster.
```

consensus/hotstuff/committees/cluster_committee.go
```diff
@@ -61,8 +61,12 @@ func NewClusterCommittee(
 	return com, nil
 }
 
-func (c *Cluster) Identities(blockID flow.Identifier, selector flow.IdentityFilter) (flow.IdentityList, error) {
-
+// Identities returns the identities of all the collection nodes members within the same cluster at the given block.
+// The order of the identities is the canonical order
```
AlexHentschel, line 65:
```suggestion
// Identities returns the identities of all cluster members that are authorized to
// participate at the given block. The order of the identities is the canonical order.
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -68,6 +68,8 @@ func NewConsensusCommittee(state protocol.State, me flow.Identifier) (*Consensus
 	return com, nil
 }
 
+// Identities returns the identities of all the consensus participants at the given block.
+// The order of the identities is the canonical order
```
AlexHentschel, line 72:
```suggestion
// Identities returns the identities of all authorized consensus participants at the given block.
// The order of the identities is the canonical order.
```

consensus/hotstuff/integration/instance_test.go
```diff
@@ -381,6 +391,10 @@ func NewInstance(t require.TestingT, options ...Option) *Instance {
 	return &in
 }
 
+func toIndices(all []flow.Identifier, signers []flow.Identifier) ([]byte, error) {
```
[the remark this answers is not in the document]
AlexHentschel: agree. Personally prefer less wrapping.

consensus/hotstuff/signature.go
```diff
@@ -127,10 +127,9 @@ type Packer interface {
 	Pack(blockID flow.Identifier, sig *BlockSignatureData) (signerIndices []byte, sigData []byte, err error)
 
 	// Unpack de-serializes the provided signature data.
-	// blockID is the block that the aggregated sig is signed for
 	// sig is the aggregated signature data
 	// It returns:
 	//  - (sigData, nil) if successfully unpacked the signature data
 	//  - (nil, model.ErrInvalidFormat) if failed to unpack the signature data
-	Unpack(blockID flow.Identifier, signerIDs []flow.Identifier, sigData []byte) (*BlockSignatureData, error)
+	Unpack(signerIDs []flow.Identifier, sigData []byte) (*BlockSignatureData, error)
```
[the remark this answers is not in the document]
AlexHentschel: I think the method `Unpack` could work with full identities. (Technically also [`Pack`](https://github.com/onflow/flow-go/blob/3ab8b5916e6fcd66ae276e7c927ab9baae894a6a/consensus/hotstuff/signature.go#L134))

I think it anyway has access to the `Committee`  and retrieves the full identities. I think we needlessly convert forth and back between IDs and Identities multiple times. Maybe that is an optimization for a later PR.

AlexHentschel (commented):
first set of comments. Only got 1/3 through the PR.

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,91 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
 }
+
+// encodeSignerIndicesAndSigType encodes the given stakingSigners and beaconSigners into signer indices and sig types.
+// The given fullMembers provides the canonical order of signer index for each signer.
+// For instance, assuming fullMembers are [A,B,C,D,E,F], stakingSigners are [D,E], beaconSigners are [A,B], then
+// the signerIndices will be [byte([1,1,0,1,1,0,0,0])],
+// 			bit 1 indicates the node at that index signed, bit 0 indicates the node at that index didn't sign
+// the sigType will be [byte([1,1,0,0,0,0,0,0])],
+// 			bit 1 indicates the signer at the same index in signerIndices signed random beacon sig
+// 			bit 0 indicates the signer at the same index in signerIndices signed staking sig
```
AlexHentschel, line 179: I like the example. I think a little extension of the doc would be helpful here. Also I would suggest to choose a different example, which more directly illustrates a subtle but important difference:

```suggestion
// As an example consider the case where we have a committee C of 10 nodes in canonical oder
//            C = [A,B,C,D,E,F,G,H,I,J]
// where nodes [B,F] are stakingSigners and beaconSigners are [C,E,G,I,J].
//  * First return parameter: QC.signerIndices
//    - We start with a bit vector v that has |C| number of bits
//    - If a node contributed either as staking signer or beacon signer,
//      we set the respective bit to 1:
//          [A,B,C,D,E,F,G,H,I,J]
//             ↓     ↓ ↓     ↓ ↓
//           0,1,1,0,1,1,1,0,1,1
//    - Lastly, right-pad the resulting bit vector with 0 to full bytes. We have 10 committee members,
//      so we pad to 2 bytes:
//           01101110 11000000
//  * second return parameter: signature types
//    - Here, we restrict our focus on the signers, which we encoded in the previous step.
//      In out example, nodes [B,C,E,F,G,I,J] signed in canonical order. This is exactly the same order,
//      as we have represented the signer in the last step.
//    - For these 5 nodes in their canonical order, we encode each node's signature type as
// 			bit-value 1: node was in beaconSigners
// 			bit-value 0: node was in stakingSigners
//      This results in the bit vector
//            [B,C,E,F,G,I,J]
//             ↓ ↓ ↓ ↓ ↓ ↓ ↓
//             0,1,0,1,1,1,1
//    - Again, we right-pad with zeros to full bytes, As we only had 7 signers, the sigType slice is 1byte long
//             01011110
// Prerequisites:
//  * The inputs `stakingSigners` and `beaconSigners` are treated as sets, i.e. they
//    should not contain any duplicates.
//  * A node can be listed in either `stakingSigners` or `beaconSigners`. A node appearing in both lists
//    constitutes an illegal input.
//  * `stakingSigners` must be a subset of `fullMembers`
//  * `beaconSigners` must be a subset of `fullMembers`
// Violation of any of these prerequisites results in an error return.
```

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,91 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+func encodeSignerIndicesAndSigType(fullMembers []flow.Identifier, stakingSigners flow.IdentifierList, beaconSigners flow.IdentifierList) ([]byte, []byte, error) {
+	stakingSignersLookup := stakingSigners.Lookup()
+	beaconSignersLookup := beaconSigners.Lookup()
+
+	indices := make([]int, 0, len(fullMembers))
+	sigType := make([]hotstuff.SigType, 0, len(fullMembers))
```
AlexHentschel, line 185:
```suggestion
	sigType := make([]hotstuff.SigType, 0, len(stakingSigners)+len(beaconSigners))
```

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,91 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+		}
+	}
+
+	if len(stakingSignersLookup) > 0 {
+		return nil, nil, fmt.Errorf("unknown staking signers %v", stakingSignersLookup)
+	}
+
+	if len(beaconSignersLookup) > 0 {
+		return nil, nil, fmt.Errorf("unknown beacon signers %v", beaconSignersLookup)
+	}
```
AlexHentschel, line 211: There are other illegal inputs that could trigger these sanity checks. For example, a node being listed in both `stakingSigners` _and_ `beaconSigners` constitutes an illegal input, which would trigger the latter check. I think we could consolidate both checks and write:

```suggestion
	if len(stakingSignersLookup) > 0 {
		return nil, nil, fmt.Errorf("unknown staking signers %v", stakingSignersLookup)
	}
	if len(beaconSignersLookup) > 0 {
		return nil, nil, fmt.Errorf("unknown or duplicated beacon signers %v", beaconSignersLookup)
	}
```

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,91 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+
+// encodeSignerIndicesAndSigType encodes the given stakingSigners and beaconSigners into signer indices and sig types.
+// The given fullMembers provides the canonical order of signer index for each signer.
+// For instance, assuming fullMembers are [A,B,C,D,E,F], stakingSigners are [D,E], beaconSigners are [A,B], then
+// the signerIndices will be [byte([1,1,0,1,1,0,0,0])],
+// 			bit 1 indicates the node at that index signed, bit 0 indicates the node at that index didn't sign
+// the sigType will be [byte([1,1,0,0,0,0,0,0])],
+// 			bit 1 indicates the signer at the same index in signerIndices signed random beacon sig
+// 			bit 0 indicates the signer at the same index in signerIndices signed staking sig
+func encodeSignerIndicesAndSigType(fullMembers []flow.Identifier, stakingSigners flow.IdentifierList, beaconSigners flow.IdentifierList) ([]byte, []byte, error) {
```
AlexHentschel, line 180: For me, it would be more intuitive to rename  `fullMembers` to `orderedCommittee`

consensus/hotstuff/integration/instance_test.go
```diff
@@ -216,15 +217,21 @@ func NewInstance(t require.TestingT, options ...Option) *Instance {
...
 		func(votes []*model.Vote) *flow.QuorumCertificate {
-			voterIDs := make([]flow.Identifier, 0, len(votes))
+			voterIDs := make(flow.IdentifierList, 0, len(votes))
 			for _, vote := range votes {
 				voterIDs = append(voterIDs, vote.SignerID)
 			}
+
+			signerIndices, err := packer.EncodeSignerIdentifiersToIndices(in.participants.NodeIDs(), voterIDs)
+			if err != nil {
+				panic(fmt.Errorf("could not encode signer indices: %w", err))
```
durkmurder, line 227: I think this panic just makes things more complicated, can we stick to require.NoError?

consensus/hotstuff/signature/packer.go
zhangchiqing: Very nice and detailed explanation. Thanks 👍

consensus/hotstuff/signature.go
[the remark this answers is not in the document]
zhangchiqing: OK, I will explore this in later PR, with test cases checking for the changes.

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -112,6 +112,7 @@ func (c *CombinedVerifier) VerifyVote(signer *flow.Identity, sigData []byte, blo
 //  - model.ErrInvalidSignature if a signature is invalid
 //  - error if running into any unexpected exception (i.e. fatal error)
 func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
+	// TODO: remove this check, because we've checked the total stake have passed threshold
```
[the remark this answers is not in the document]
zhangchiqing: Even if we remove this check, the `Unpack` will fail anyway if `signers` is empty.

Also we are not supposed to return `model.ErrInvalidFormat` if sanity check fails, which is what confused me. When I saw this, I thought there was unhappy case from a user error, but actually not. 

Since the later operation will fail even without this check, I would prefer to remove this check here to avoid confusion.

consensus/integration/signer_test.go
```diff
@@ -26,15 +26,12 @@ func (s *Signer) CreateVote(block *model.Block) (*model.Vote, error) {
...
-	}
 	qc := &flow.QuorumCertificate{
-		View:      votes[0].View,
-		BlockID:   votes[0].BlockID,
-		SignerIDs: voterIDs,
-		SigData:   nil,
+		View:    votes[0].View,
+		BlockID: votes[0].BlockID,
+		// TODO: fix
+		SignerIndices: nil,
```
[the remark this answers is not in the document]
zhangchiqing: https://github.com/onflow/flow-go/pull/2140/files#diff-0651c91205a7f0e131c435303abb380ea71331d7149406658db82852a8235174
This doesn't need to be fixed actually, because the entire signature validation is mocked away

engine/collection/compliance/core.go
```diff
@@ -83,7 +83,7 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Hex("payload_hash", header.PayloadHash[:]).
 		Time("timestamp", header.Timestamp).
 		Hex("proposer", header.ProposerID[:]).
-		Int("num_signers", len(header.ParentVoterIDs)).
+		// Int("num_signers", len(header.ParentVoterIDs)).
```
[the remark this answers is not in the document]
zhangchiqing: I commented this out because I haven't thought about what to replace with, but also don't want to forget about it.

Now the header only has ParentVoterIndices, I don't want to take extra query to parse the actual signers. Maybe we could simply log the entire `ParentVoterIndices` field?

module/finalizer/collection/finalizer_test.go
```diff
@@ -342,7 +342,7 @@ func TestFinalizer(t *testing.T) {
 			prov.AssertCalled(t, "SubmitLocal", &messages.SubmitCollectionGuarantee{
 				Guarantee: flow.CollectionGuarantee{
 					CollectionID: block1.Payload.Collection.ID(),
-					SignerIDs:    block1.Header.ParentVoterIDs,
+					SignerIDs:    []flow.Identifier{}, // block1.Header.ParentVoterIndices,
```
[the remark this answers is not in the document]
zhangchiqing: [Fixed here](https://github.com/onflow/flow-go/pull/2140/files#diff-d54be921364d5207f28a38390df5a4315f47566cf8fda39a663044bf1aaaf2ee)

consensus/hotstuff/integration/instance_test.go
zhangchiqing: [Fixed here](https://github.com/onflow/flow-go/pull/2140/files#diff-3aaac0dad8c18f64b4dea3aa6b2ea3dc9330028ca6cb5d4c56887b25491165d8)

model/flow/quorum_certificate.go
zhangchiqing: Yeah, makes sense, will address in later PR

module/finalizer/collection/finalizer.go
```diff
@@ -150,8 +150,10 @@ func (f *Finalizer) MakeFinal(blockID flow.Identifier) error {
...
+					// add cluster ID/chain ID
+					SignerIDs: nil, // step.ParentVoterIndices,
+					Signature: step.ParentVoterSigData,
```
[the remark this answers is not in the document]
zhangchiqing: [Added TODO here](https://github.com/onflow/flow-go/pull/2140/files#diff-0a46088b36f83a26444b535d8dc03540cefb433cd71118ec732a126d235c7dc2R155)

state/protocol/convert.go
zhangchiqing: The Cluster.Members is created from the AssignmentList. So [I fixed from the source](https://github.com/onflow/flow-go/pull/2140/files#r828751067) by sorting the assignment list so that when cluster list is built off the assignment, it's already sorted.

consensus/hotstuff/verification/combined_verifier_v2.go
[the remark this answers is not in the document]
durkmurder: Agree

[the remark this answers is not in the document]
AlexHentschel: #  ⚠️ 

I think there are two aspects to [this discussion](https://github.com/onflow/flow-go/pull/2101/files#r829188365):
1. Should we have a check here? 
2. If the check fails, what error type should we return. 

## My thoughts regarding 1:
* This is an implementation of the [`Verifier` interface](https://github.com/onflow/flow-go/blob/619a82e30c2c478b21b957d063db06a8436ff638/consensus/hotstuff/verifier.go). 
* The place where we check that  the `total stake have passed threshold` is in the `Validator`: https://github.com/onflow/flow-go/blob/619a82e30c2c478b21b957d063db06a8436ff638/consensus/hotstuff/validator/validator.go#L70-L71 So this call from the `Validator` into the `Verifier.VerifyQC` happens through an API boundary
    ```
        Validator [higher level code] --- API boundary ---> Verifier [lower level code]
    ```

* Now there are some important considerations:
   1.  The API specification [`Verifier.VerifyQC`](https://github.com/onflow/flow-go/blob/619a82e30c2c478b21b957d063db06a8436ff638/consensus/hotstuff/verifier.go#L38-L51) does _not_ mention the prerequisite that that higher level code must check _first_ that the total stake has passed the threshold. 
   2. The assumption that "total stake have passed threshold" is made by the `Verifier`. In other words, we are making **assumptions in the lower-level code about ordering of operations in the higher level code.** I like to avoid any such assumptions if possible with reasonable effort.
   3. Lastly, I am not convinced about this statement  
        > the Unpack will fail anyway if signers is empty.

        If both `signers` and `sigData` are empty or nil, are we sure that `Unpack` will fail? I think it might actually succeed. But the crypto operations (specifically `AggregateBLSPublicKeys`) will return an unexpected error, thereby crashing the receiving node.  

    In which order the higher level code executes its operations should ideally be up the higher level code and the lower level code should be agnostic to it. Otherwise, we make the code brittle; an engineer might re-order operations in the higher level code without being aware about implicit assumptions in the lower level code. After such a re-ordering of logic, the code might have uncovered edge cases, which could be exploited by malicious replicas. 

### Conclusion for 1: 
Ideally, the code in `CombinedVerifier.VerifyQC` should handle any input without making assumptions about higher level code. This also included the case of empty inputs like `signers` and `sigData` as the ultimately were generated by an untrusted external actor. I think this is relatively simple to implement (see below). 


## My thoughts regarding 2:
* In general, an empty `signers` slice indicates an invalid QC. 
* Nevertheless, @zhangchiqing  has a good point:
   * _If_ the higher-level logic already checked that the "total stake have passed threshold", `signers` being empty is an internal bug. However, there is no strong reason that the `Validator` must perform in the checks in this order. 
   * _If_ the `Validator` checked the signatures first (calling [`Verifier.VerifyQC`](https://github.com/onflow/flow-go/blob/619a82e30c2c478b21b957d063db06a8436ff638/consensus/hotstuff/verifier.go#L38-L51) _first_), `len(signers) == 0` would indicate a byzantine input. 
  
### Conclusion for 2:  
I think that the semantics of this failure case depends on the order of checks within `Validator`. Therefore, we should allow `Validator` to distinguish the failure case of `len(signers) == 0` from other failure cases. Thereby it can make its own decision whether this error is expected (caused by a byzantine input) or a symptom of an internal bug).

## Argument also applies to `CombinedVerifierV3`

In `CombinedVerifierV3`, we do _not_ check whether `signers` is empty. In a world where we _first_ check the signatures, this code would actually be vulnerable to a crash attack. This is because `AggregateBLSPublicKeys` fails with an internal error if there are no signatures to aggregate: https://github.com/onflow/flow-go/blob/619a82e30c2c478b21b957d063db06a8436ff638/consensus/hotstuff/verification/combined_verifier_v3.go#L145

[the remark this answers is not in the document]
AlexHentschel: P.S. will be addressing this in my PR with suggested revisions.

module/packer/signer_indices.go
```diff
@@ -0,0 +1,65 @@
+package packer
```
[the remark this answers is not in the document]
tarakby: This file could at least re-use the tools from `ledger/utils` like `MakeBitVector` and `SetBit`.

module/packer/signer_indices.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+)
+
+// EncodeSignerIndices encodes indices into compacted bit vector. Each bit represents whether the identity at
+// that index is a signer.
+// Note, the indices in the first argument must be in the strict increasing order,
```
tarakby, line 9: if we re-use the tools from `ledger/utils`, I think this condition won't be required anymore.

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,118 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+//  * First return parameter: QC.signerIndices
+//    - We start with a bit vector v that has |C| number of bits
+//    - If a node contributed either as staking signer or beacon signer,
+//      we set the respective bit to 1:
+//          [A,B,C,D,E,F,G,H,I,J]
+//             ↓     ↓ ↓     ↓ ↓
+//           0,1,1,0,1,1,1,0,1,1
+//    - Lastly, right-pad the resulting bit vector with 0 to full bytes. We have 10 committee members,
+//      so we pad to 2 bytes:
+//           01101110 11000000
```
tarakby, line 186: Fyi, this is how bytes look like only for big endian cpus. 
```suggestion
//           01101110 11000000   (if big endianness)
```

consensus/hotstuff/signature/packer.go
```diff
@@ -229,3 +168,118 @@ func deserializeFromBitVector(serialized []byte, count int) ([]hotstuff.SigType,
...
+//      as we have represented the signer in the last step.
+//    - For these 5 nodes in their canonical order, we encode each node's signature type as
+// 			bit-value 1: node was in beaconSigners
+// 			bit-value 0: node was in stakingSigners
+//      This results in the bit vector
+//            [B,C,E,F,G,I,J]
+//             ↓ ↓ ↓ ↓ ↓ ↓ ↓
+//             0,1,0,1,1,1,1
+//    - Again, we right-pad with zeros to full bytes, As we only had 7 signers, the sigType slice is 1byte long
+//             01011110
```
tarakby, line 199:
```suggestion
//             01011110   (if big endianness)
```

consensus/hotstuff/committees/cluster_committee.go
```diff
@@ -74,14 +78,12 @@ func (c *Cluster) Identities(blockID flow.Identifier, selector flow.IdentityFilt
 
 	// use the initial cluster members for root block
 	if isRootBlock {
-		return c.initialClusterMembers.Filter(selector), nil
+		return c.initialClusterMembers, nil
```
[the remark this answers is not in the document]
jordanschalm: Queries for the cluster's root block return the initial cluster members, without excluding any nodes ejected in the future. The reason is that the root block is defined at the same time as the initial cluster members are defined -- any ejections after that point shouldn't apply. Any ejections before that point should be reflected in the initial cluster members.

consensus/hotstuff/integration/instance_test.go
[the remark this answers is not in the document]
zhangchiqing: https://github.com/onflow/flow-go/pull/2140/commits/6bba92656391d2840baec9323edde561412e9c2d

module/packer/signer_ids.go
```diff
@@ -0,0 +1,49 @@
...
+func EncodeSignerIdentifiersToIndices(fullIdentities []flow.Identifier, signerIDs flow.IdentifierList) ([]byte, error) {
```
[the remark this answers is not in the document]
zhangchiqing: Addressed https://github.com/onflow/flow-go/pull/2204/files#diff-a1983c6bd364d16ae09eb3dcc8abd9970e72cc66df5e8901d1434f188567d5b3R65

consensus/hotstuff/verification/combined_verifier_v2.go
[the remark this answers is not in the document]
zhangchiqing: Addressed https://github.com/onflow/flow-go/pull/2204/files#diff-0cd49adfc8c2df8b741ff1252f2dc39210c56f5f9a506d21154e259b01b66cdeR123

engine/collection/compliance/core.go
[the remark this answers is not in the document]
zhangchiqing: Addressed [here](https://github.com/onflow/flow-go/pull/2140/files#diff-6491debb977ace8c99ddb1edbd37fca0420cc659243a52ebb03505ba7efe78fdR86)

module/packer/signer_indices.go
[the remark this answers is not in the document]
zhangchiqing: Addressed https://github.com/onflow/flow-go/pull/2204/files#diff-a1983c6bd364d16ae09eb3dcc8abd9970e72cc66df5e8901d1434f188567d5b3R189

consensus/hotstuff/signature/packer.go
zhangchiqing: https://github.com/onflow/flow-go/pull/2204/files#diff-a1983c6bd364d16ae09eb3dcc8abd9970e72cc66df5e8901d1434f188567d5b3R233

PR #2102, opened 2022-03-03
Flaky Test Monitor - BigQuery update and full refactoring of unit tests

gomisha:
This PR updates flaky test monitor in 2 main ways:

1) Level 1 JSON output conforms to BigQuery JSON schema requirements so that level 1 summary data can be uploaded to BigQuery via REST

2) Level 1 and 2 unit test major refactoring so that the majority of tests now use structs to represent Level 1 Summary instead of JSON files. This was done to lower the cost to maintain many JSON files as the schema will inevitably evolve. There is now a single level 1 test and a single level 2 test that uses real level 1 JSON files. All the remaining level 1 and level 2 tests use structs instead to represent level 1 data.


tools/flaky_test_monitor/level1/process_summary1_results_test.go
```diff
@@ -101,58 +154,24 @@ func runProcessSummary1TestRun(t *testing.T, jsonExpectedActualFile string) {
...
+	require.Equal(t, len(expectedLevel1Summary.Rows), len(actualLevel1Summary.Rows))
 
-		// check all outputs of each test result
-		require.Equal(t, len(expectedTestResult.Output), len(actualTestResult.Output), fmt.Sprintf("TestResult[%d].Test: %s", testResultIndex, actualTestResult.Test))
-		for testResultOutputIndex := range expectedTestResult.Output {
-			require.Equal(t, expectedTestResult.Output[testResultOutputIndex], actualTestResult.Output[testResultOutputIndex], fmt.Sprintf("TestResult[%d] Output[%d]", testResultIndex, testResultOutputIndex))
-		}
+	// check that all expected results are in actual results
+	for actualRowIndex := range actualLevel1Summary.Rows {
+		require.Contains(t, actualLevel1Summary.Rows, expectedLevel1Summary.Rows[actualRowIndex], printLevel1TestResult(expectedLevel1Summary.Rows[actualRowIndex].TestResult, "expected not in actual"))
```
smnzhu, line 169: This checks that the actual result contains the expected as a substring, but do we actually want to check that they are exactly equal?

tools/flaky_test_monitor/level1/process_summary1_results_test.go
```diff
@@ -101,58 +154,24 @@ func runProcessSummary1TestRun(t *testing.T, jsonExpectedActualFile string) {
...
+	}
 
-		require.Equal(t, expectedTestResult.Package, actualTestResult.Package)
-		require.Equal(t, expectedTestResult.Test, actualTestResult.Test)
-		require.Equal(t, expectedTestResult.Elapsed, actualTestResult.Elapsed, fmt.Sprintf("TestResult[%d].Test: %s", testResultIndex, actualTestResult.Test))
-		require.Equal(t, expectedTestResult.Result, actualTestResult.Result)
+	// check that all actual results are in expected results
+	for exptedRowIndex := range expectedLevel1Summary.Rows {
```
smnzhu, line 173: similar comment here

tools/flaky_test_monitor/level2/process_summary2_results.go
```diff
@@ -13,96 +13,107 @@ import (
...
+				saveFailureMessage(level1TestResultRow.TestResult)
+			case "skip":
+				level2TestResult.Skipped++
+
+				// don't count skip as a run
+				level2TestResult.Runs--
+
+				// truncate last duration - don't count durations of skips
+				level2TestResult.Durations = level2TestResult.Durations[:len(level2TestResult.Durations)-1]
+			case "-100":
```
smnzhu, line 62: is -100 just a random magic value we chose to represent no result?

tools/flaky_test_monitor/level1/process_summary1_results_test.go
```diff
@@ -6,89 +6,142 @@ import (
...
-		"3 count nil test with normal tests": "test-result-nil-test-others-normal-3-count-pass.json",
+		"3 count nil test with normal tests": {
+			ExpectedLevel1Summary: testdata.GetTestData_Leve1_3CountNilWithNormalTests(),
+			RawJSONTestRunFile:    "test-result-nil-test-others-normal-3-count-pass.json",
+		},
+	}
+
+	require.NoError(t, os.Setenv("COMMIT_DATE", "2021-09-21T18:06:25-07:00"))
+	require.NoError(t, os.Setenv("COMMIT_SHA", "46baf6c6be29af9c040bc14195e195848598bbae"))
+	require.NoError(t, os.Setenv("JOB_STARTED", "2021-09-21T21:06:25-07:00"))
```
smnzhu, line 90: nit: define these values as variables?

tools/flaky_test_monitor/common/testdata/test_data.go
```diff
@@ -0,0 +1,888 @@
...
+	level1Summaries = append(level1Summaries, level1Summary3)
+	level1Summaries = append(level1Summaries, level1Summary4)
+	level1Summaries = append(level1Summaries, level1Summary5)
+	level1Summaries = append(level1Summaries, level1Summary6)
+	level1Summaries = append(level1Summaries, level1Summary7)
+	level1Summaries = append(level1Summaries, level1Summary8)
+	level1Summaries = append(level1Summaries, level1Summary9)
+	level1Summaries = append(level1Summaries, level1Summary10)
+	return level1Summaries
+}
```
smnzhu, line 888: Could you add some comments here? It's not super clear what this function is doing. In particular, what is the meaning of the loop iteration counts?

tools/flaky_test_monitor/level1/process_summary1_results_test.go
gomisha: It checks by the entire element, not by substring. It checks that every expected level 1 result row row element (not a substring) is in the actual result list (and vice versa). This makes it more adaptable to expected or actual results being in a different order and still passing.

[`require.contains()`](https://pkg.go.dev/github.com/stretchr/testify/require#Contains) uses "asserts that the specified string, list(array, slice...) or map contains the specified substring or element"

tools/flaky_test_monitor/level1/process_summary1_results_test.go
gomisha: Yes, we want to test that all actual level 1 result rows are in the actual level 1 result rows and vice versa.

tools/flaky_test_monitor/level2/process_summary2_results.go
gomisha: Yes, it's explained in the Level 1 parser in more detail so I copied the comments from there to here so it's more clear:

https://github.com/onflow/flow-go/pull/2102/commits/424731838609dbd97fea06074b1afb62c1edfad8

```
// for tests that don't have a result generated (e.g. using fmt.Printf() with no newline in a test)
// we want to highlight these tests so they show up at the top in Granfa
// we do this by simulating a really low fail rate so that the average success rate
// will be very low compared to other tests so it will show up on the "flakiest test" panel
```

tools/flaky_test_monitor/level1/process_summary1_results_test.go
gomisha: extracted to `testdata` package

https://github.com/onflow/flow-go/pull/2102/commits/e0672bd5a3e9c83c1446787cfefcbd583630e55b

tools/flaky_test_monitor/common/testdata/test_data.go
gomisha: I added comments and more explanation for all the test data functions so it's more clear how they're used by the level 1 and 2 unit tests. This function has to create a collection of level 1 summaries that are used as inputs to the level 2 parser. The loop iterations model how many times `go test` was run (e.g. `go test -count=10`) to generate level 1 summaries. We then take a collection of all the level 1 summaries to generate a level 2 summary.

https://github.com/onflow/flow-go/pull/2102/commits/b8fcd03c948869f46a4a3566d1b3f6ee22388fba

tools/flaky_test_monitor/level1/process_summary1_results_test.go
smnzhu: So, what you actually want to do here is check that `actualLevel1Summary.Rows` and `expectedLevel1Summary.Rows` contain the exact same elements right? Should we use something like `assert.EqualValues` here?

gomisha: I was hoping this would work, but it doesn't. I tried this:

`require.EqualValues(t, expectedLevel1Summary.Rows, actualLevel1Summary.Rows)`

but that causes many assertion failures.

> So, what you actually want to do here is check that actualLevel1Summary.Rows and expectedLevel1Summary.Rows contain the exact same elements right?

Yes, that's right. I used to compare each element one at a time between expected and actual lists but that was brittle because the order mattered. I found using `require.contains()` is a good way to do assertions without requiring the order is the same.

smnzhu: https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

gomisha: That works - thanks! Updated to use `require.ElementsMatch()`

https://github.com/onflow/flow-go/pull/2102/commits/4275b7c8535e8b65e1ee92d636efe2b93e320b4e

tools/flaky_test_monitor/common/testdata/test_data.go
gomisha: I added more comments on the different loop counters so it's more clear on why we use different loop counters to construct level 1 expected data.

https://github.com/onflow/flow-go/pull/2102/commits/0958e2012babadc413e4227a551d9d7b8ebbbbb5

tools/flaky_test_monitor/level2/process_summary2_results.go
```diff
@@ -13,96 +13,107 @@ import (
...
+				level2TestResult.NoResult++
+				// don't count no result as a run
+				level2TestResult.Runs--
+
+				// truncate last duration - don't count durations of no-results
+				level2TestResult.Durations = level2TestResult.Durations[:len(level2TestResult.Durations)-1]
```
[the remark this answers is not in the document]
gomisha: Updated to count no-results as failures and added a new field to count no-results so we can view all the tests with no-results in Grafana

https://github.com/onflow/flow-go/pull/2102/commits/efcb8f09b5efd63a963b4eb949ac8c793af1eec2

tools/flaky_test_monitor/level2/process_summary2_results.go
gomisha: removed using magic value and treating no-results as a failure but with additional bool field set to true

https://github.com/onflow/flow-go/pull/2102/commits/efcb8f09b5efd63a963b4eb949ac8c793af1eec2

smnzhu (approved):

Kay-Zee (approved):

Issue #2048, opened 2022-02-23
HTTP API: Explicitly define transaction success


sideninja:
## Problem Definition
Deciding a “success” (transaction did what it was intended to do) of a transaction using AN APIs (from the developers point of view) is not straightforward and might be prone to mistakes

When a transaction is sealed its success is determined by the presence of an error message which is not great and prone to mistakes. If a transaction was successful it should be more explicit in the response with an additional field.

The error message is not structured data but just a string `"error_message": "[Error Code: 1006] invalid proposal key: public key 0 on account df9c30eb2252f1fa does not have a valid signature: [Error Code: 1009] invalid envelope key: public key 0 on account df9c30eb2252f1fa does not have a valid signature: signature is not valid"`

There are cases where the error message is not present but the transaction wasn’t “successful” - expired (which might be hard to realize during testing). More documentation on this is needed either way but can we build a better response for that

## Proposed Solution

To solve above mentioned problems I propose additional field on the response `result` indicating the failure or success of a transaction and `errors` containing structured error responses possible to be understood by client software:

```
{
    "result": "success|failure|pending",
    "status": "unknown|pending|finalized|executed|sealed|expired",
    "errors": [{
        "code": int,
        "message": "..."
    }, ...]
    ... rest of the response stays the same ... 
}  
```

Error example would be:
```
{
    "result": "failure",
    "status": "sealed",
    "errors": [{
        "code": 1006,
        "message": "invalid proposal key: public key 0 on account df9c30eb2252f1fa does not have a valid signature"
    }, {
        "code": 1009,
        "message": "invalid envelope key: public key 0 on account df9c30eb2252f1fa does not have a valid signature: signature is not valid"    
    }]
    ... rest of the response stays the same ... 
}  
```

or success example (note we omit error altogether):
```
{
    "result": "success",
    "status": "sealed",
    ... rest of the response stays the same ... 
}  
```
I would even suggest adding an optional developer error message that could provide more guidance on error resolution, example would be (but I’m not strongly opinionated about it since we need to provide value with this message not just rephrase the error so although I feel it’s great to have that it really depends if we can make it so that it provides more value (which is arguably not true for my example, but is just an example)):
```
{
    "result": "failure",
    "status": "sealed",
    "errors": [{
        "code": 1006,
        "message": "invalid proposal key: public key 0 on account df9c30eb2252f1fa does not have a valid signature"
    }, {
        "code": 1009,
[...]

PR #2106, opened 2022-03-04
[HTTP API] Transaction success status

sideninja:
Closes: #2048 

Deciding a “success” (transaction did what it was intended to do) of a transaction using AN APIs (from the developers point of view) is not straightforward and might be prone to mistakes

When a transaction is sealed its success is determined by the presence of an error message which is not great and prone to mistakes. If a transaction was successful it should be more explicit in the response with an additional field.

This PR implements transaction execution status field.

P.S. there is a lot of file changes that just come from including linting in the model generation, form now on model generation will be part of make action.


vishalchangrani:
Adding a few more reviewers since we need two approvals to merge to master.

vishalchangrani (approved):
lgtm..there are lint errors though.

koko1123 (approved):
looks good to me! thanks for thrwoing in the test and `make generate-openapi`, it was definitely a bother for me to figure out when I first tried doing this!

peterargue (approved):

PR #2108, opened 2022-03-04
[Execution] enable cadence runtime tracing given node level flag

ramtinms:
This PR sets a tracing flag on cadence runtime when tracing is enabled on the node level by an extra flag.


janezpodhostnik (approved):

m4ksio:
I get the idea, but maybe it should be separate flag? 
We need an ability to set up different logging levels for different components, but extra flag should do before we get there

ramtinms:
@m4ksio thats a good suggestion, made it a separate flag for cadence runtime tracing.

m4ksio (approved):

SupunS (approved):

PR #2114, opened 2022-03-07
Define ToggleUploaderCommand

smnzhu:
add command to toggle uploaders on / off


engine/execution/computation/manager.go
```diff
@@ -28,6 +29,14 @@ import (
 	"github.com/onflow/flow-go/utils/logging"
 )
 
+var uploadEnabled bool
+
+func SetUploaderEnabled(enabled bool) {
```
peterargue, line 34: I think we'll need to also call this function from the component startup:
https://github.com/onflow/flow-go/blob/master/cmd/execution/main.go#L217
https://github.com/onflow/flow-go/blob/master/cmd/execution/main.go#L247

vishalchangrani (approved):

engine/execution/computation/manager.go
vishalchangrani: set default to true on line 32 now

peterargue (approved):

PR #2115, opened 2022-03-07
explicitly setting the uploader to enabled/disabled based on the enab…

vishalchangrani:
…le-blockdata-upload flag passed at startup

### DIFF
--- a/cmd/execution/main.go
+++ b/cmd/execution/main.go
@@ -218,6 +218,7 @@ func main() {
 			return nil
 		}).
 		Component("GCP block data uploader", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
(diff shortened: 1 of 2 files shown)


cmd/execution/main.go
```diff
@@ -218,6 +218,7 @@ func main() {
 			return nil
 		}).
 		Component("GCP block data uploader", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+			computation.SetUploaderEnabled(enableBlockDataUpload)
```
peterargue, line 221: might want to do that inside the conditional incase the bucket name is empty. It should also get set to true in the s3 case (line 253)

peterargue (approved):

smnzhu (commented):
What about if we just set the default value of the variable to be true?

PR #2117, opened 2022-03-07
Log computation metering

janezpodhostnik:
A minimal version of https://github.com/onflow/flow-go/pull/1631
This only adds one log to each transaction.

This is needed to get mode data for variable execution fees

### DIFF
--- a/fvm/handler/computation.go
+++ b/fvm/handler/computation.go
@@ -7,9 +7,11 @@ type ComputationMeter interface {
 	// Limit gets computation limit
 	Limit() uint64
 	// AddUsed adds more computation used to the current computation used
-	AddUsed(used uint64) error
+	AddUsed(intensity uint64, feature string)
 	// Used gets the current computation used
 	Used() uint64
+	// Intensities gets the intensities of features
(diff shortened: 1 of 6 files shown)


fvm/handler/computation.go
```diff
@@ -27,36 +29,38 @@ type SubComputationMeter interface {
...
 func (c *computationMeter) Limit() uint64 {
 	return c.limit
 }
 
-func (c *computationMeter) AddUsed(used uint64) error {
-	c.used += used
-	return nil
+func (c *computationMeter) AddUsed(intensity uint64, feature string) {
+	c.intensities[feature] += intensity
+	if feature != "function_or_loop_call" {
```
ramtinms, line 45: what about statements?

janezpodhostnik: This is the label used for computation coming from cadence:

```
func (e *ScriptEnv) SetComputationUsed(used uint64) error {
	return e.computationHandler.AddUsed(used)
	e.computationHandler.AddUsed(used, "function_or_loop_call")
	return nil
}
```

Which includes everything we measured before (including statements).

PR #2119, opened 2022-03-08
[Execution State] Add logging for CPU, RAM, and checkpoint

### LINKED ISSUE #2074: [Execution State] Add extra logging in execution node
## Problem

Some error(s), warnings, and other useful information are not currently logged.

Examples:
- It can be useful to log CPU info (model, family, logical cores, etc.) for various reasons.  If EN specs fluctuate between Broadwell and Cascade Lake with substantial GHz difference, then having that info can be useful when deploying optimizations and making comparisons to past performance.
- It can be useful to log a warning if some activity based on dynamic or stored data is going to use a suspiciously large percent of total installed RAM.  If we ever run out of RAM, then such a log can help us save time investigating crash loops, etc.
- There's at least one error not logged which was noticed during PR #[1944](https://github.com/onflow/flow-go/pull/1944) but it's out of scope for that PR and located outside checkpoint code. 

## The Proposed Solution

- Log some useful info (e.g. cpu info), warnings (see above), and errors.
- Avoid spamming the log with noise.

fxamacker:
### Changes

Add logs with CPU and RAM info to provide context when we use logs on EN to compare performance.

- CPU: ModelName, MHz, Family, Model, Stepping, PhysicalCores, LogicalCores
- RAM: Total, Free

While at it, also add extra logs for:
- Checkpoint: creation, loading, and replaying

Closes #2074
Updates #1884 https://github.com/dapperlabs/flow-go/issues/6114

### Caveats

This PR uses a popular 3rd-party Go library `shirou/gopsutil`.  It is cross-platform and is used by well-known projects such as `go-ethereum`.

For EN, if we only need to support linux_amd64 then we can avoid using 3rd-party library and just parse `/proc/cpuinfo` and `/proc/meminfo`.  I don't think we need to reinvent the wheel but I'm open to replacing the 3rd-party package if requested.


cmd/execution/main.go
```diff
@@ -330,6 +334,12 @@ func main() {
 				}
 			}
 
+			hwLogger := node.Logger.With().Str("system", "hardware").Logger()
+			err = logHardware(hwLogger)
```
zhangchiqing, line 338: Wonder why do we log it inside of "execution state ledger"?

If it's logging static hardware spec, we could move it to its own module:

```
		Module("log hardware spec", func(node *cmd.NodeConfig) error {
			hwLogger := node.Logger.With().Str("system", "hardware").Logger()
			err = logHardware(hwLogger)
			if err != nil {
				hwLogger.Error().Err(err)
			}
			return nil
		}).
```

zhangchiqing (approved):

cmd/execution/main.go
fxamacker: Great idea!  Done in 22c46fcb90cc15bf508d23c58e9bc307db240de1.

ramtinms (approved):

PR #2121, opened 2022-03-08
[BFT Testing] Adds attacker implementation

yhassanzadeh13:
This PR implements attacker logic that is wrapper around the plug-an-play logic of attack orchestrator(s). We envision having a spectrum of attack orchestrators (e.g., double voting orchestrators). On the contrary, the attacker implementation is the same and provides a lifecycle management component that is capable of accommodating an orchestrator, and running it, while abstracting away implementation details of the BFT testing framework.


insecure/adversary/attacker_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+				// received event by orchestrator must be an expected one.
+				require.Contains(t, events, e)
+				seen[e] = struct{}{}
+				orchestratorWG.Done()
+
+			}).Return(nil)
+
+			// sends all messages concurrently to the attacker (imitating corruptible conduits sending
+			// messages conccurently to attacker).
```
gomisha, line 68: nit: spelling: conccurently

insecure/adversary/attacker_test.go
```diff
@@ -0,0 +1,195 @@
...
+		msgs[i] = m
+		events[i] = e
+	}
+
+	return msgs, events
+}
+
+// withAttacker is a test helper that creates and starts an attacker server, prepares a set of message fixtures and runs the given scenario
+// against the attacker and generated messages.
+func withAttacker(
```
gomisha, line 150: can you add more high level explanation of how exactly `withAttacker()` works? I find it confusing to follow the logic of this function to understand what exactly it's doing

insecure/adversary/attacker_test.go
```diff
@@ -0,0 +1,195 @@
...
+	go func() {
+		select {
+		case err := <-errChan:
+			t.Error("attacker startup encountered fatal error", err)
+		case <-ctx.Done():
+			return
+		}
+	}()
+
+	// starts attacker
```
gomisha, line 176: can you add comments here to better explain what exactly is happening?

insecure/adversary/attacker_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2121/commits/7766095508a35da50eb9f382dd8b1cd3f99e4ef6

insecure/adversary/attacker_test.go
yhassanzadeh13: Tried decomposing the test logic into smaller components in https://github.com/onflow/flow-go/pull/2121/commits/7766095508a35da50eb9f382dd8b1cd3f99e4ef6 and added more comments.

insecure/adversary/attacker_test.go
yhassanzadeh13: The logic is quite self-explanatory and common across our codebase. So, let's discuss it in more context here instead of adding redundant comments to the code that may stale over the time:

Starting the attacker starts its sub-components. In the Flow codebase, we define the notion of [`Component`](https://github.com/onflow/flow-go/blob/57f89d4e96259f08fe84163c91ecd32484401b45/module/component/component.go#L202) as an abstraction of a _control_ logic. At the very broad level, the protocol logic is decomposed into self-contained `Component`s each with a certain responsibility. Smaller `Component`s are composed together and shaped into bigger ones with more general responsibilities. Examples of such in our BFT testing implementation are `Attacker` and `Orchestrator`. The `Orchestrator` implements a certain type of attack (e.g., the wintermute), and is _composed_ into the bigger `Attacker` component that is responsible for managing the orchestrator together with a bunch of corrupted conduit factories. 

In such a design, we cannot let components commence their work at the initialization time. The reason is some components depend on others, and some are composed into bigger ones. There is no guarantee provided by Golang that all components are initialized together even if we attempt to run them concurrently. This provides motivations to decouple the components _initialization_ from _startup_. 

One may see the _startup_ as triggering a tree of events. The biggest component (i.e., the node) starts first. As part of its startup, it starts its constituent components, which in turn each component starts its subcomponents, etc. 

But how do we make sure that a component is started successfully? by looking at its `Ready` method. The `Ready` returns a channel that gets closed when the component is (and hence all its sub-component are) ready. That is why we do `attacker.Start(...)` and then wait for `attacker.Ready`'s returned channel to get closed. 

But how we can make sure that a fault in the attacker's startup does not block the entire test? the answer is the `RequireClosesBefore` helper function of our unit-testing. It fails the entire test if the given channel is not closed prior to a deadline. So the following means start the attacker and if it fails to get started completely within 1 second, fail the entire test. 

```
	attacker.Start(attackCtx)
	unittest.RequireCloseBefore(t, attacker.Ready(), 1*time.Second, "could not start attacker on time")
```

insecure/adversary/attacker_test.go
```diff
@@ -0,0 +1,214 @@
...
+	mockinsecure "github.com/onflow/flow-go/insecure/mock"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/model/libp2p/message"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/network"
+	"github.com/onflow/flow-go/network/codec/cbor"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+const attackerAddress = "localhost:8000"
```
vishalchangrani, line 27: Here is the example of the Access node test which allows dynamic ports instead of a fixed on - 
1. Test sets the port to 0 https://github.com/onflow/flow-go/blob/ceaa10f04aa3a05687dbc8197937c2f0a88f64d8/engine/access/rate_limit_test.go#L93-#L96
2. The engine code exposes a function retrieve the port https://github.com/onflow/flow-go/blob/3efe3a9375de4c61ff7568e3047ba640da835c3d/engine/access/rpc/engine.go#L244
3. The test waits for the port to be opened and then reads it - https://github.com/onflow/flow-go/blob/ceaa10f04aa3a05687dbc8197937c2f0a88f64d8/engine/access/rate_limit_test.go#L117-#L124

insecure/adversary/attacker.go
```diff
@@ -0,0 +1,148 @@
...
+			ln, err := net.Listen(networkingProtocolTCP, attacker.address)
+			if err != nil {
+				ctx.Throw(fmt.Errorf("could not listen on specified address: %w", err))
+			}
+			attacker.server = s
+
+			wg := sync.WaitGroup{}
+			wg.Add(1)
+			go func() {
+				wg.Done()
```
vishalchangrani, line 56: the `wg.Done()` wouldn't guarantee that the gRPC server starts severing. I had a similar issue and resorted to checking a non-nil port was returned.

insecure/event.go
```diff
@@ -0,0 +1,19 @@
...
+// An event is the protocol-level representation of an outgoing message of a corruptible conduit.
+// The corruptible conduit relays the message to the attacker instead of dispatching it through the Flow network.
+// The attacker decodes the message into an event and relays it to the orchestrator.
+type Event struct {
+	CorruptedId flow.Identifier     // identifier of corrupted conduit
+	Channel     network.Channel     // channel of the event on the corrupted conduit
+	Content     interface{}         // the protocol-level event itself.
+	Protocol    Protocol            // networking-layer protocol that this event was meant to send on.
+	TargetNum   uint32              // number of randomly chosen targets (used in multicast protocol).
+	TargetIds   flow.IdentifierList // set of target identifiers
```
gomisha, line 18: `TargetIds` is the list of all nodes a message is meant for? Is there a distinction between honest / dishonest nodes in the list? I feel like this should be made more clear.

insecure/event.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Event represents the data model that is exchanged between the attacker and the attack orchestrator.
+// An event is the protocol-level representation of an outgoing message of a corruptible conduit.
+// The corruptible conduit relays the message to the attacker instead of dispatching it through the Flow network.
+// The attacker decodes the message into an event and relays it to the orchestrator.
+type Event struct {
+	CorruptedId flow.Identifier     // identifier of corrupted conduit
+	Channel     network.Channel     // channel of the event on the corrupted conduit
+	Content     interface{}         // the protocol-level event itself.
```
gomisha, line 15: isn't this the message? It might be clearer to re-name this field to `Message` and change the comment to use `message` instead of `event`

insecure/adversary/attacker_test.go
yhassanzadeh13: Thanks, applied here: https://github.com/onflow/flow-go/pull/2121/commits/aed11d2e0c8cee21f550bd9cbe723229992e524a

insecure/adversary/attacker.go
vishalchangrani: my bad - now seeing that the listener (and hence the port) is already open by this time.

vishalchangrani (approved):
lgtm

insecure/event.go
yhassanzadeh13: Provided more details in https://github.com/onflow/flow-go/pull/2121/commits/139adaf0eaf7cf27f1ccb54783f4645c7550a96d

insecure/event.go
yhassanzadeh13: Renamed and clarified in https://github.com/onflow/flow-go/pull/2121/commits/139adaf0eaf7cf27f1ccb54783f4645c7550a96d

gomisha (approved):

PR #2129, opened 2022-03-10
[FVM] refactor computation and memory metering

ramtinms:
This pr
- bumps cadence version to support computation metering interface changes 
- moves computation metering logic into the state to enable recording metrics when programs are getting cached and replying when returned from cache (prevents execution forks)
- adds basic metering that behaves similar to the cadence runtime limits before moving responsibility to FVM (+tests)
- enables building more complex computation and memory metering logic that depends on the execution state (e.g. reading values from service account)


fvm/transactionInvoker.go
```diff
@@ -243,23 +244,6 @@ func (i *TransactionInvoker) deductTransactionFees(env *TransactionEnv, proc *Tr
...
-		if err != nil {
-			// The error merr (from discarding the subMeter) will be hidden by err (transaction fee deduction error)
-			// as it has priority. So log merr.
-			i.logger.Error().Err(merr).
-				Msg("error discarding computation meter in deductTransactionFees (while also handling a deductTransactionFees error)")
-			return
-		}
-		err = merr
-	}()
-
```
ramtinms, line 262: I believe we don't need this anymore since we disable the metering

fvm/meter/basic/meter.go
```diff
@@ -0,0 +1,88 @@
+package basic
+
+import (
+	"github.com/onflow/flow-go/fvm/errors"
+	interfaceMeter "github.com/onflow/flow-go/fvm/meter"
+)
+
+var _ interfaceMeter.Meter = &Meter{}
+
+// Meter is a basic meter, which collects memory and computation usage
```
ramtinms, line 10: note to self: update/improve docs here.

janezpodhostnik (approved):
Looks good.

How come the base for this branch is `cadence-issue-1127`?

fvm/transactionEnv.go
```diff
@@ -545,15 +530,19 @@ func (e *TransactionEnv) GenerateUUID() (uint64, error) {
...
 
 func (e *TransactionEnv) GetComputationLimit() uint64 {
-	return e.computationHandler.Limit()
+	return uint64(e.sth.State().TotalComputationLimit())
 }
 
 func (e *TransactionEnv) SetComputationUsed(used uint64) error {
-	return e.computationHandler.AddUsed(used)
+	// TODO update this interface with the proper cadence ones
+	if e.sth.EnforceComputationLimits() {
```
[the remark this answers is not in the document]
ramtinms: I was thinking to change the method to match the cadence one and call this method everywhere else in the fvm code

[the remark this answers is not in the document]
janezpodhostnik: another solution could be adding something like this to the env:
```
func (e *TransactionEnv) meterComputation(kind, intensity uint) error {
	if e.sth.EnforceComputationLimits() {
		return e.sth.State().MeterComputation(kind, intensity)
	}
	return nil
}
```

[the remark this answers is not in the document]
ramtinms: yeah thats what I meant, but I kept that change for after the cadence interface changes are merged, that method deosn't have to be private and I should be public so cadence call also use it

fvm/context.go
```diff
@@ -59,9 +60,18 @@ func newContext(ctx Context, opts ...Option) Context {
 const AccountKeyWeightThreshold = 1000
 
 const (
-	DefaultGasLimit                     = 100_000 // 100K
-	DefaultEventCollectionByteSizeLimit = 256_000 // 256KB
+	DefaultComputationLimit             = 100_000     // 100K
+	DefaultMemoryLimit                  = 100_000_000 // 100G
```
turbolent, line 64: Is this per transaction? This seems very high, no? Or is it intentionally high until we find "average" / "normal" uses?

fvm/context.go
```diff
@@ -59,9 +60,18 @@ func newContext(ctx Context, opts ...Option) Context {
...
+	DefaultEventCollectionByteSizeLimit = 256_000     // 256KB
 	DefaultMaxNumOfTxRetries            = 3
+
+	// service transaction is a transaction have service account as payer
+	// and signed by service account, higher limits are allowed
+	// for service transactions to consider flexibility for network services
+	// such as governance transactions, etc.
+	ServiceTransactionComputationLimit             = 100_000_000_000 // 100B
```
turbolent, line 72: Did you mean 100T? Are we sure about that? 100G already seems very high
```suggestion
	ServiceTransactionComputationLimit             = 100_000_000_000 // 100T
```

fvm/context.go
```diff
@@ -110,10 +121,18 @@ func WithChain(chain flow.Chain) Option {
 	}
 }
 
-// WithGasLimit sets the gas limit for a virtual machine context.
-func WithGasLimit(limit uint64) Option {
```
turbolent, line 114: Thank you for making the naming consistent 👍

fvm/fvm_test.go
```diff
@@ -3492,3 +3492,121 @@ func TestStorageUsed(t *testing.T) {
 
 	assert.Equal(t, cadence.NewUInt64(5), script.Value)
 }
+
+func TestEnforcingComputationLimit(t *testing.T) {
```
turbolent, line 3496: Good idea to have the same tests here 👍

turbolent (commented):
Thank you for also already starting on integration / preparing for the memory metering 👍

fvm/context.go
ramtinms: yeah, I set it to a large number until we realize what should be actual values, also don't think its connected so, meterMemory calls are not received yet.

fvm/context.go
ramtinms: so we didn't have any limit before, this is only the case that if service account accidentally forgets to set a limit, and submit an always for loop, I was not sure what we should set for them, and didn't want to cause any disturbance to the reward payout and etc, @Kay-Zee any idea what we could do as a upperbound for service transactions?

fvm/context.go
ramtinms: I was worried to replace them with math.MaxUint64

fvm/context.go
ramtinms: I just noticed dkg test package is dependent on the emulator and the emulator is dependent on this method, added this method as well as a deprecated method, until we update the emulator and other places, we could remove it in the follow up PRs.

janezpodhostnik (approved):

fvm/meter/basic/meter.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return nil
+}
+
+// TotalComputationUsed returns the total computation used
+func (m *Meter) TotalComputationUsed() uint {
+	return m.computationUsed
+}
+
+// TotalComputationUsed returns the total computation limit
```
SupunS, line 75:
```suggestion
// TotalComputationLimit returns the total computation limit
```

fvm/state/state.go
```diff
@@ -165,18 +175,50 @@ func (s *State) Set(owner, controller, key string, value flow.RegisterValue, enf
...
+func (s *State) MeterComputation(kind, intensity uint) error {
+	return s.meter.MeterComputation(kind, intensity)
+}
+
+// TotalComputationUsed returns total computation used
+func (s *State) TotalComputationUsed() uint {
+	return s.meter.TotalComputationUsed()
+}
+
+// TotalComputationUsed returns total computation limit
```
SupunS, line 198:
```suggestion
// TotalComputationLimit returns total computation limit
```

SupunS (approved):

fvm/state/state.go
ramtinms: updated at 2d0182d

fvm/meter/basic/meter.go
```diff
@@ -0,0 +1,97 @@
...
+	computationLimit uint
+	memoryUsed       uint
+	memoryLimit      uint
+}
+
+// NewMeter constructs a new Meter
+func NewMeter(computationLimit, memoryLimit uint) *Meter {
+	return &Meter{
+		computationLimit: uint(computationLimit),
+		memoryLimit:      uint(memoryLimit),
```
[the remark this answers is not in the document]
ramtinms: updated at 2d0182d

fvm/context.go
ramtinms: this seems to be a separate problem to be solved, move the logic outside of this PR.

fvm/context.go
ramtinms: set it to 2GB for now, but it's not enforced since there is no call to meter memory yet. would adjust in that PR

PR #2135, opened 2022-03-10
5437 - remove unnecessary MakeID calls

jwinkler2083233:
This will fix a few unnecessary calls to MakeID()

Here's an example profile with MakeID(), taken recently:

```
 event 'cpu-clock', 4000 Hz, Event count (approx.): 46514024409                                                                   
  Children      Self  Shared Object           Symbol                                                                                              
+   33.27%     0.00%  app                     [.] runtime.goexit.abi0
+   12.55%     0.20%  [kernel]                [k] do_syscall_64
+   11.22%    11.06%  app                     [.] github.com/onflow/flow-go/crypto/hash.keccakF1600.abi0
+    8.11%     3.90%  app                     [.] runtime.mallocgc
+    6.31%     0.03%  app                     [.] github.com/onflow/flow-go/model/flow.MakeID
+    6.28%     0.00%  app                     [.] github.com/dgraph-io/badger/v2.(*DB).View

```


pattyshack:
most of the extra calls are in the error path.  don't think this helps much

PR #2136, opened 2022-03-10
Jwinkler2083233/6185 reduce mem keypart

jwinkler2083233:


jwinkler2083233:
Looks like this was also done in a similar way by Faye.

PR #2141, opened 2022-03-11
Remove double instantiation of admin server for access node

smnzhu:
Admin server is already initialized in scaffold here: https://github.com/onflow/flow-go/blob/c8a7461a5befd1458022396ef3e2cb26c9d3f231/cmd/scaffold.go#L1137

So we should not enqueue it again in access node. If we do, we cause it to be instantiated twice, which leads to "bind already in use" error.


cmd/access/node_builder/staked_access_node_builder.go
```diff
@@ -97,8 +97,6 @@ func (builder *StakedAccessNodeBuilder) Initialize() error {
 		return err
 	}
 
-	builder.EnqueueAdminServerInit()
```
zhangchiqing, line 100: I see. Is this function still being used? If not, we could just remove

zhangchiqing (approved):
Nice, thanks for the fix

cmd/access/node_builder/staked_access_node_builder.go
smnzhu: The same exact function is being called in scaffold.go.

The problem was that if we call it again in the access node builder, it will be called twice.

peterargue (approved):

PR #2144, opened 2022-03-11
[BFT Testing] Implements attack network

yhassanzadeh13:
This PR implements the attack network that is acting as the place holder to mount an attack orchestrator (e.g., the Wintermute orchestrator) on it. The attack network sits between the mounted orchestrator and the corrupted nodes and relays their exchanged messages in both directions.


insecure/attacknetwork/attackNetwork.go
```diff
@@ -0,0 +1,244 @@
...
+				a.logger.Info().Msg("attack network closed processing stream")
+				return stream.SendAndClose(&empty.Empty{})
+			}
+			if err != nil {
+				return fmt.Errorf("could not read corrupted node's stream: %w", err)
+			}
+
+			if err = a.processMessageFromCorruptedNode(msg); err != nil {
+				a.logger.Fatal().Err(err).Msg("could not process message of corrupted node")
+				return stream.SendAndClose(&empty.Empty{})
```
peterargue, line 163: this will never be called since `Fatal()` will call `os.Exit(1)`. maybe instead of a fatal error, return a custom error type that the caller can use to decide what to do. alternatively, you could accept an `irrecoverable.SignalerContext` and throw the error.

insecure/attacknetwork/attackNetwork.go
```diff
@@ -0,0 +1,244 @@
...
+
+	payload, err := a.codec.Encode(event)
+	if err != nil {
+		return nil, fmt.Errorf("could not encode event: %w", err)
+	}
+
+	return &insecure.Message{
+		ChannelID: channel.String(),
+		OriginID:  corruptedId[:],
+		Targets:   num,
```
peterargue, line 239: nit: should this be `TargetCount`? `Targets` seems to imply it's a list of targets

insecure/conduitMaster.go
```diff
@@ -5,6 +5,9 @@ import (
 	"github.com/onflow/flow-go/network"
 )
 
+// CorruptedFactoryPort defines the port on which attacker dials a corruptible conduit factory.
+const CorruptedFactoryPort = 5000
+
```
peterargue, line 10: not specific to this line, but a general comment. Let's use a more inclusive word Instead of `master`. Maybe `Coordinator` or `Controller`?

peterargue (commented):
looks great. added a few comments

insecure/conduitMaster.go
yhassanzadeh13: applied in https://github.com/onflow/flow-go/pull/2144/commits/7cca5cab58bd76956d2ca057beedc6eaa032fbf9

insecure/attacknetwork/attackNetwork.go
yhassanzadeh13: fixed in https://github.com/onflow/flow-go/pull/2144/commits/9049d4ffec6c0e6a9f033366587586d029fec535

insecure/attacknetwork/attackNetwork.go
yhassanzadeh13: Thanks for pointing it out. 

This code is part of the `Observe` method, which implements a client streaming gRPC interface. The interface is generated by compiling a `.proto` file, and cannot be refactored to accept an `irrecoverable.SignalerContext`. Main reason is this `Observe` method is supposed to get called by the gRPC stub solely. 

> this will never be called since `Fatal()` will call `os.Exit(1)`.  

Correct, so the purpose of the code is to safely crash the attacker process in the event that messages arriving from a (remote) corrupted node is not successfully getting through the process path. On a happy path, the attacker must be able to process every incoming message of remote corrupted nodes. If the happy path of processing an incoming message is broken, the testing framework should not continue, and we do it by crashing the attacker. Since attacker is running on the main process that runs the test, crashing it through `os.Exit(1)` will crash the test with the fatal level log. 

The subsequent return was _initially_ there to let the code compile (i.e., all paths of the method ends with a return). However, realized that we no longer need it with current shape of the code, so removed it in https://github.com/onflow/flow-go/pull/2144/commits/3a05692dba6ab4bfe20278bf029773774804efa5

peterargue (approved):
looks great

gomisha (approved):

PR #2147, opened 2022-03-14
[FVM, EN] pass request context and add timeout for script execution

ramtinms:
This PR: 
- updates FVM to accept a context.Context when constructing a script procedure and check for context cancellation or timeout on every meter compution call during the execution of a script. (this is only limited to script procedures since transactions have to have a deterministic execution and can not support this) 
- updates execution computation manager to append a timeout limit to the request context. This way, script execution can be terminated when we have reached the timeout or it was cancelled by upper layers (e.g. grpc connection was dropped or client timed out) 
- sets the default timeout to 10 seconds but adjustable by a flag on EN nodes.


fvm/script.go
```diff
@@ -14,35 +15,57 @@ import (
...
+	return &ScriptProcedure{
+		ID:             proc.ID,
+		Script:         proc.Script,
+		RequestContext: reqContext,
+		Arguments:      proc.Arguments,
+	}
 }
 
-func (proc *ScriptProcedure) WithArguments(args ...[]byte) *ScriptProcedure {
+func NewScriptWithContextAndArgs(code []byte, reqContext context.Context, args ...[]byte) *ScriptProcedure {
```
ramtinms, line 62: did add this extra method to avoid copying on every script

cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -151,7 +152,7 @@ func NewBalanceReporter(chain flow.Chain, view state.View) *balanceProcessor {
 		accounts: accounts,
 		st:       st,
 		prog:     prog,
-		intf:     fvm.NewScriptEnvironment(ctx, vm, sth, prog),
+		intf:     fvm.NewScriptEnvironment(context.Background(), ctx, vm, sth, prog),
```
ramtinms, line 155: its fine to use background context here, since the scope of collecting account balance is anyway limited.

fvm/scriptEnv.go
```diff
@@ -439,7 +443,16 @@ func (e *ScriptEnv) meterComputation(kind, intensity uint) error {
 }
 
 func (e *ScriptEnv) MeterComputation(kind common.ComputationKind, intensity uint) error {
-	return e.meterComputation(uint(kind), intensity)
+	select {
```
janezpodhostnik, line 446: Should this change be in the private `meterComputation` receiver instead? Otherwise, calling `meterComputation` will go around this logic. 

Another question is if `e.sth.EnforceComputationLimits` is disabled, should we check if the context is done. Usually `e.sth.EnforceComputationLimits` is disabled for parts that should not be interrupted, though that is not critical in scripts.

janezpodhostnik (approved):

turbolent (approved):
Nice! 

Next step is probably passing it to Cadence?

engine/execution/computation/manager_test.go
```diff
@@ -292,18 +288,15 @@ func TestExecuteScript_ShortScriptsAreNotLogged(t *testing.T) {
...
-	view := delta.NewView(func(_, _, _ string) (flow.RegisterValue, error) {
-		return nil, nil
-	})
 	header := unittest.BlockHeaderFixture()
 
 	eds := new(state_synchronization.ExecutionDataService)
 	edCache := new(state_synchronization.ExecutionDataCIDCache)
 
-	manager, err := New(log, metrics.NewNoopCollector(), nil, nil, nil, vm, ctx, DefaultProgramsCacheSize, committer.NewNoopViewCommitter(), 1*time.Second, nil, eds, edCache)
+	manager, err := New(log, metrics.NewNoopCollector(), nil, nil, nil, vm, ctx, DefaultProgramsCacheSize, committer.NewNoopViewCommitter(), 1*time.Second, DefaultScriptExecutionTimeLimit, nil, eds, edCache)
```
turbolent, line 296: Maybe chop down the argument list here, or even consider defining a struct which bundles all the parameters of the function

ramtinms: did chop down, we are going to revisit all these tests soon, so we could consider using a param object as we refactor.

PR #2150, opened 2022-03-15
Upgrade to Secure Cadence

janezpodhostnik:
Upgrade Cadence to Secure Cadence [secure-cadence-m1](github.com/onflow/cadence@secure-cadence-m1).

Uses  core contracts from [lib/go/contracts/secure-cadence-m2](https://github.com/onflow/flow-core-contracts/releases/tag/lib%2Fgo%2Fcontracts%2Fsecure-cadence-m2)

Used in the emulator in PR [Upgrade emulator to use Secure Cadence](https://github.com/onflow/flow-emulator/pull/127)

⚠️ DO NOT MERGE


cmd/util/cmd/epochs/cmd/templates.go
```diff
@@ -50,7 +50,7 @@ transaction(name: String,
 
             // the below arguments are unused and are safe to be left empty
             collectorClusters: [],
-            clusterQCs: [],
-            dkgPubKeys: [])
+            clusterQCs: [] as [FlowClusterQC.ClusterQC],
+            dkgPubKeys: [] as [String])
```
SupunS, line 54: isn't this needed for the preceding `collectorClusters: [],` argument also

SupunS: I'm actually more curious about how this worked 🤔  This change (type inferring changes - which I assume the cause of this) was done many moons ago

janezpodhostnik: Checking again, it seems that the change here does not affect the results of any tests. I do not know if the reason is because this code is not used in a test, or is there some other reason.

I made these changes together with making the changes in `fvm/blueprints/epochs.go`. Those definitely affect tests. Without those changes the following tests fail:
```
--- FAIL: TestFungibleTokenTracker (0.36s)
panic: failed to deploy Epoch contract: [Error Code: 1101] cadence runtime error Execution failed:
error: cannot infer type from array literal:  requires an explicit type annotation
  --> 0d705cc7942cbee81c7bb25b82977f305f6ec04f28e14bfd5924b4b33536d5a4:27:14
   |
27 | 		clusterQCs: [],
   |               ^

error: cannot infer type from array literal:  requires an explicit type annotation
  --> 0d705cc7942cbee81c7bb25b82977f305f6ec04f28e14bfd5924b4b33536d5a4:28:14
   |
28 | 		dkgPubKeys: [],
   |               ^
```

turbolent: I was wondering about these changes before, as I expected the type assertions not to be necessary.
Shouldn't type inference propagate the parameter type to the argument here?

SupunS: Thanks, @janezpodhostnik for double-checking.

@turbolent This particular case (extra arguments) is not supported yet (along with generics) https://github.com/onflow/cadence/blob/481cb23a7d6bac00083cf5127a86a17b900c99aa/runtime/sema/check_invocation_expression.go#L450-L451. This one should be easy to add, but generic is the complicated one. This was documented in the original issue, but can't seem to fund it now. I'll create an issue anyway for these particular ones.

janezpodhostnik: What is the way forward here? Do I check all cadence scripts in the `flow-go` and `flow-core-contracts` for empty arrays, dictionaries in arguments and update them?

SupunS: This is actually strange. I checked this scenario and it works in Cadence `v0.23.2` (IDK how/why it worked in the first place). But doesn't work in the master. So some recent must have triggered this. I'll have a look at this today.

SupunS: Ok, so I figured it. It was this change: https://github.com/onflow/cadence/pull/1027/files#diff-6cc33849623674ce4f600cda1e45c49e822a86b35be9bc61f4ed3ca83d8f0671R79-R92. Earlier, we used to assume empty dictionaries without context-type is to be Never type. But now we are more strict and expect the user to add an explicit annotation. I can special case the empty array/dictionary case to have the old behaviour, but I think the current behaviour is correct.

Assuming the current behaviour is correct: this should only affect empty arrays/dictionaries passed into a function with generics, or that use extra arguments (e.g: signer.contracts.add() where contract inits' args are passed, apart from the add() methods arguments). Maybe checking usages of signer.contracts.add() would be a good idea and might be sufficient.

Normal function calls should be working as expected.

janezpodhostnik: I checked all usages of `.contracts.add(` and only found this reference and the one in `fvm/blueprints/epochs.go` that use empty arrays or empty maps. I updated both usages.

cmd/util/ledger/reporters/fungible_token_tracker.go
```diff
@@ -151,7 +152,7 @@ func (r *FungibleTokenTracker) worker(
 
 		for _, domain := range domains {
 			storageMap := storage.GetStorageMap(owner, domain)
-			itr := storageMap.Iterator()
+			itr := storageMap.Iterator(&interpreter.Interpreter{})
```
janezpodhostnik, line 155: @SupunS TestOrderedMapMigration is now failing (after upgrading to the latest commit of the memory-metering branch) and I think it relates to this change, but I am not sure how to fix it.

Any Ideas?

```
--- FAIL: TestOrderedMapMigration (0.00s)
    --- FAIL: TestOrderedMapMigration/sort_values (0.00s)
panic: cannot convert stored value: migrations.RawStorable [recovered]
	panic: cannot convert stored value: migrations.RawStorable

goroutine 10 [running]:
testing.tRunner.func1.2(0x4add640, 0xc0002c2740)
	/usr/local/go/src/testing/testing.go:1143 +0x332
testing.tRunner.func1(0xc000102d80)
	/usr/local/go/src/testing/testing.go:1146 +0x4b6
panic(0x4add640, 0xc0002c2740)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/onflow/cadence/runtime/interpreter.MustConvertStoredValue(...)
	/Users/janez/go/pkg/mod/github.com/onflow/cadence@v0.21.3-0.20220316204141-2e995e2195c8/runtime/interpreter/storage.go:48
github.com/onflow/cadence/runtime/interpreter.StoredValue(0x4e55340, 0xc000102c00, 0x4e69740, 0xc000556150, 0x4e6d030, 0xc000560e10, 0xc000556168, 0x4e69740)
	/Users/janez/go/pkg/mod/github.com/onflow/cadence@v0.21.3-0.20220316204141-2e995e2195c8/runtime/interpreter/storage.go:42 +0xdf
github.com/onflow/cadence/runtime/interpreter.StorageMap.SetValue(0xc000561380, 0xc000102c00, 0xc000482af7, 0x3, 0x4e58360, 0xc000556168)
	/Users/janez/go/pkg/mod/github.com/onflow/cadence@v0.21.3-0.20220316204141-2e995e2195c8/runtime/interpreter/storagemap.go:127 +0x194
github.com/onflow/flow-go/cmd/util/ledger/migrations.(*OrderedMapMigration).migrate(0xc000506090, 0xc000427e80, 0x5, 0x5, 0x40, 0x10, 0xc000427dd0, 0x2, 0x2)
	/Users/janez/Programing/dl/flow-go/cmd/util/ledger/migrations/ordered_map_migration.go:237 +0x41c
github.com/onflow/flow-go/cmd/util/ledger/migrations.TestOrderedMapMigration.func2(0xc000102d80)
	/Users/janez/Programing/dl/flow-go/cmd/util/ledger/migrations/ordered_map_migration_test.go:88 +0xb4c
testing.tRunner(0xc000102d80, 0xc000561050)
	/usr/local/go/src/testing/testing.go:1193 +0xe5
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1238 +0x2b3
```

SupunS: Just saw this. Having a look now 👀

SupunS: It was due to an API change in `interpreter.Value` interface. Pushed a fix: 8c597980e5956cd90c5b7aa90b2d94e1ef9ce505

turbolent, line 155: Maybe move the interpreter instantiation outside the loop. Same below

access/validator.go
```diff
@@ -227,9 +228,16 @@ func (v *TransactionValidator) checkExpiry(tx *flow.TransactionBody) error {
...
+}
+
+func (noopMemoryGauge) MeterMemory(_ common.MemoryUsage) error {
+	return nil
+}
+
 func (v *TransactionValidator) checkCanBeParsed(tx *flow.TransactionBody) error {
 	if v.options.CheckScriptsParse {
-		_, err := parser2.ParseProgram(string(tx.Script))
+		_, err := parser2.ParseProgram(string(tx.Script), noopMemoryGauge{})
```
turbolent, line 240: It should be possible to pass nil here. If not, we should fix this on the Cadence side.

If it is needed at the moment, maybe declare a global for the noop gauge and re-use it across calls.

turbolent (approved):

janezpodhostnik:
Merged v0.25 into this branch. Please recheck if everything is ok.

turbolent (approved):

SupunS (approved):

janezpodhostnik:
replaced with https://github.com/onflow/flow-go/pull/2337

PR #2156, opened 2022-03-16
[FVM] support cadence memory metering

ramtinms:


fvm/transactionInvoker.go
```diff
@@ -289,7 +289,7 @@ func valueDeclarations(ctx *Context, env Environment) []runtime.ValueDeclaration
 			Kind:           common.DeclarationKindFunction,
 			IsConstant:     true,
 			ArgumentLabels: nil,
-			Value: interpreter.NewHostFunctionValue(
+			Value: interpreter.NewUnmeteredHostFunctionValue(
```
ramtinms, line 292: I'm not sure about this @janezpodhostnik

janezpodhostnik: IMO this is fine. Its only one un-metered function, and we know what this function is supposed to do.

engine/execution/computation/computer/computer_test.go
```diff
@@ -502,6 +502,10 @@ func (e *testRuntime) SetResourceOwnerChangeHandlerEnabled(_ bool) {
 	panic("SetResourceOwnerChangeHandlerEnabled not expected")
 }
 
+func (e *testRuntime) SetInvalidatedResourceValidationEnabled(_ bool) {
```
janezpodhostnik, line 505: This should go away if you pull from [janez/secure-cadence](https://github.com/onflow/flow-go/tree/janez/secure-cadence) again.

PR #2161, opened 2022-03-17
Update Cadence version to secure-cadence-m1

SupunS:
Uses the https://github.com/onflow/cadence/tree/secure-cadence-m1 tag


janezpodhostnik (approved):
LGTM

Issue #2100, opened 2022-03-03
[Access API] Add block_id and transaction_index to GetTransactionResult parameters to avoid duplicate transaction hashes


koko1123:
## Problem Definition

In engine/access/rpc/backend/backend_transactions.go, when looking up a transaction result for a particular transaction hash, the Access Node does:
`block, err := b.lookupBlock(txID)`
This block then gets used to query the Execution Node for a particular transaction result within a specific block.
Since users can seemingly resubmit the same transaction (with the same transaction hash) multiple times in different blocks, the Access Node finds a previous block with the same transaction hash when it does the lookupBlock(txID), and thus the wrong transaction result gets returned

## Proposed Solution

Define a new endpoint `GetTransactionResultByIndex` which takes in block_id and transaction_index parameters, where transaction_index represents the index within the block (not within a collection). Both of these are required to guarantee unambiguity. We will then also need to add the `GetTransactionResultByIndex` API to the ExecutionAPI, and the Access node handler should forward the request to the execution node.

smnzhu:
@koko1123 Updated the Proposed Solution, please review it again.

Also, we probably want to create a separate issue for the `flow-go-sdk` to support this new endpoint.

koko1123:
As we have implemented the EN side of these changes we've encountered a constraint:

1. Transactions are currently stored and indexed in ANs by `TransactionId`, not by `blockId` and `index`
2. As the status of the transaction in same `transactionId` at the inter/intra block level, the indexed status stored will reflect the *most recent* status, instead of the status at the exact block/index
3. This means that we do not have accurate data for previous sporks at the AN level
4. We are implementing the indexing scheme in 1., that will track this for future sporks
5. for mainnet16 and 15, we'll have to backfill transaction data for this new store, which we can attend to later.

m4ksio:
> As we have implemented the EN side of these changes we've encountered a constraint:
> 
> 1. Transactions are currently stored and indexed in ANs by `TransactionId`, not by `blockId` and `index`

That is not correct, I might have mislead you earlier. We key both transactions results by blockID and txID. So this cover duplicated txs in different blocks, only failing when we have two txs in same block.
For events we are in a better position, as we key events by blockID, txID txIndex (and event index) - which means we can query events without a miss.

I think we might be able to regenerate the index for tx results unless we have two txs in one block, but that should be easy to check

bluesign:
[...]

PR #2166, opened 2022-03-18
Store and retrieve transaction results by EN RPC

m4ksio:
EN part of #2100

As this is intended to deploy mid-spork I added a new index (block_id, tx_index) rather than modify existing one. 
For a new spork we should probably combine indexes and use badger traversal to retrieve data needed and avoid duplicating tx results (although they are rather small.

Notice, this will only allow of retrieval by index for new data, which has been executed after this change is live. Previously executed data won't be available but rebuilding index for past data is a separate issue.


go.mod
```diff
@@ -106,3 +106,5 @@ require (
 )
 
 replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
+
+//replace github.com/onflow/flow => ../flow
```
koko1123, line 110: why aren't we using the local version?

storage/badger/transaction_results.go
```diff
@@ -13,14 +15,21 @@ import (
...
+	db         *badger.DB
+	cache      *Cache
+	indexCache *Cache
 }
 
 func KeyFromBlockIDTransactionID(blockID flow.Identifier, txID flow.Identifier) string {
 	return fmt.Sprintf("%x%x", blockID, txID)
 }
 
+func KeyFromBlockIDIndex(blockID flow.Identifier, txIndex uint32) string {
```
koko1123, line 27: smart!

koko1123 (approved):
Great work on such short notice!

storage/mocks/storage.go
```diff
@@ -600,6 +600,21 @@ func (mr *MockEventsMockRecorder) ByBlockIDTransactionID(arg0, arg1 interface{})
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ByBlockIDTransactionID", reflect.TypeOf((*MockEvents)(nil).ByBlockIDTransactionID), arg0, arg1)
 }
 
+// ByBlockIDTransactionIndex mocks base method
```
ramtinms, line 603: any reason for having to mocks files on under mocks and one under mock ?

go.mod
```diff
@@ -59,13 +59,13 @@ require (
...
 	github.com/onflow/cadence v0.23.2
-	github.com/onflow/flow v0.2.3-0.20220131193101-d4e2ca43a621
+	github.com/onflow/flow v0.2.3-0.20220315185722-b1c4393db996
 	github.com/onflow/flow-core-contracts/lib/go/contracts v0.10.2-0.20220303222312-6b9de4f3fea5
 	github.com/onflow/flow-core-contracts/lib/go/templates v0.10.1
 	github.com/onflow/flow-emulator v0.20.3
 	github.com/onflow/flow-go-sdk v0.24.0
 	github.com/onflow/flow-go/crypto v0.24.3
-	github.com/onflow/flow/protobuf/go/flow v0.2.4-0.20220304041411-6d91cd04a33a
+	github.com/onflow/flow/protobuf/go/flow v0.2.4-0.20220315185722-b1c4393db996
```
ramtinms, line 68: do we need to release a new version there ?

ramtinms (approved):
Looks good to me, just put two minor comments

go.mod
m4ksio: Yes, once we merge the changes from other repo

storage/mocks/storage.go
m4ksio: Two different mocking libraries maybe?

PR #2171, opened 2022-03-21
[FVM] signature verification refactoring

tarakby:
- remove the user-domain tag completely from the repo, as this is not a protocol constant anymore. 
- remove not needed `SignatureVerifier` interface to verify signatures. 
- separate the verification functions of runtime and transaction signatures that support different algorithms for a better clarity and to avoid future errors.
- add extra checks for transaction verification inputs (although they are supposed to be valid as they are read from the state).
- make the public function `SignWithTag` unable to sign other that the transaction tag for safety. 
- adjust existing tests, add new tests
- avoid using `fmt.Sprintf` for key deduplication (performance)


fvm/crypto/crypto.go
```diff
@@ -160,71 +164,83 @@ func VerifySignatureFromRuntime(
...
+	valid, err := pk.Verify(signature, message, hasher)
 	if err != nil {
 		// All inputs are guaranteed to be valid at this stage.
 		// The check for crypto.InvalidInputs is only a sanity check
 		if crypto.IsInvalidInputsError(err) {
 			return false, err
 		}
-		panic(fmt.Errorf("verify signature failed with unexpected error %w", err))
+		// TODO: check if we should panic or bubble the error up
+		panic(fmt.Errorf("verify transaction signature failed with unexpected error %w", err))
```
tarakby, line 243: @ramtinms, wondering if we should panic or return an error here. This is a case of unexpected error (error in allocating memory for instance..)

fvm/crypto/crypto.go
```diff
@@ -160,71 +164,83 @@ func VerifySignatureFromRuntime(
...
+	if pk.Algorithm() != crypto.ECDSAP256 && pk.Algorithm() != crypto.ECDSASecp256k1 {
+		// TODO: check if we should panic
+		// This case only happens in production if there is a bug
+		return false, errors.NewValueErrorf(pk.Algorithm().String(), "is not supported in transactions")
+	}
+	// hashing compatibility
+	if hashAlgo != hash.SHA2_256 && hashAlgo != hash.SHA3_256 {
+		// TODO: check if we should panic
+		// This case only happens in production if there is a bug
+		return false, errors.NewValueErrorf(hashAlgo.String(), "is not supported in transactions")
```
tarakby, line 227: @ramtinms : If this error happens then there is a bug. Should we panic?

janezpodhostnik: I think a FVM `errors.Failure` should be returned instead.

fvm/crypto/hash.go
```diff
@@ -17,38 +20,36 @@ import (
...
-	case hash.SHA2_384:
-		hasher = hash.NewSHA2_384()
-	case hash.SHA3_384:
-		hasher = hash.NewSHA3_384()
-	case hash.Keccak_256:
-		hasher = hash.NewKeccak_256()
-	default:
-		return nil, errors.New("hashing algorithm is not a supported for prefixed algorithm")
+	hasherCreator := hasherCreators[algo]
+	if hasherCreator == nil {
```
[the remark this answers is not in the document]
tarakby: Yep, good suggestion 👌🏼

access/validator.go
```diff
@@ -267,13 +267,16 @@ func (v *TransactionValidator) checkAddresses(tx *flow.TransactionBody) error {
...
+	}
+	observedSigs := make(map[uniqueKey]bool)
 	for _, sig := range append(tx.PayloadSignatures, tx.EnvelopeSignatures...) {
-		keyStr := sig.UniqueKeyString()
-		if observedSigs[keyStr] {
+		if observedSigs[uniqueKey{sig.Address, sig.KeyIndex}] {
 			return DuplicatedSignatureError{Address: sig.Address, KeyIndex: sig.KeyIndex}
 		}
-		observedSigs[keyStr] = true
+		observedSigs[uniqueKey{sig.Address, sig.KeyIndex}] = true
```
ramtinms, line 279: I assume this change is mostly for performance reasons, could you please check we have tests for this?

fvm/context.go
```diff
@@ -32,7 +31,6 @@ type Context struct {
 	ServiceEventCollectionEnabled bool
 	AccountFreezeAvailable        bool
 	ExtensiveTracing              bool
-	SignatureVerifier             crypto.SignatureVerifier
```
ramtinms, line 35: this change might need some changes on the emulator and other places that has their own context and environments.

model/flow/constants.go
```diff
@@ -42,22 +42,25 @@ const DefaultMaxAddressIndex = 20_000_000
...
+//
+// Signatures on Flow that needs to be scoped to a certain domain need to
+// have the same length in order to avoid tag collision issues, when prefixing the
+// message to sign.
 const DomainTagLength = 32
 
 const TransactionTagString = "FLOW-V0.0-transaction"
 
-const UserTagString = "FLOW-V0.0-user"
-
```
ramtinms, line 50: this might be in use by go-sdk or others ?

fvm/crypto/crypto.go
ramtinms: we usually panic if there is a small chance of non-deterministic reasons that caused this error.

fvm/crypto/crypto.go
```diff
@@ -160,71 +164,85 @@ func VerifySignatureFromRuntime(
...
+		return false, errors.NewUnknownFailure(fmt.Errorf(
+			pk.Algorithm().String(), "is not supported in transactions"))
+	}
+	// hashing compatibility
+	if hashAlgo != hash.SHA2_256 && hashAlgo != hash.SHA3_256 {
+		// TODO: check if we should panic
+		// This case only happens in production if there is a bug
+		return false, errors.NewUnknownFailure(fmt.Errorf(
+			hashAlgo.String(), "is not supported in transactions"))
+	}
```
ramtinms, line 230: not sure about this, you mean if there is some problematic old keys we would stop the network? I think we should just error and add a watch on the error type.

fvm/crypto/crypto.go
```diff
@@ -146,11 +154,7 @@ func VerifySignatureFromRuntime(
 			return false, errors.NewValueErrorf(sigAlgo.String(), "cannot use hashing algorithm type %s with signature signature algorithm type %s",
 				hashAlgo, sigAlgo)
 		}
-
-		// tag length compatibility
-		if len(tag) > flow.DomainTagLength {
-			return false, errors.NewValueErrorf(tag, "tag length (%d) is larger than max length allowed (%d bytes).", len(tag), flow.DomainTagLength)
-		}
+		// tag constraints are checked when initializing a prefix-hasher
```
ramtinms, line 157: you mean we check the constraints inside NewPrefixedHashing ?

fvm/crypto/crypto.go
```diff
@@ -160,71 +164,85 @@ func VerifySignatureFromRuntime(
...
-	publicKey crypto.PublicKey,
+	pk crypto.PublicKey,
 	hashAlgo hash.HashingAlgorithm,
 ) (bool, error) {
 
-	var hasher hash.Hasher
+	// check ECDSA compatibilites
+	if pk.Algorithm() != crypto.ECDSAP256 && pk.Algorithm() != crypto.ECDSASecp256k1 {
+		// TODO: check if we should panic
+		// This case only happens in production if there is a bug
```
ramtinms, line 220: we should probably run some analysis on the mainnet data to verify all data are okey

ramtinms (approved):
Looks good to me. thanks for clearing things up.

model/flow/transaction.go
```diff
@@ -273,16 +273,18 @@ func (tb *TransactionBody) SignEnvelope(
...
 
-// SignMessageWithTag signs the message (tag + message) with the specified account key.
+// SignMessageWithTag signs the data (transaction_tag + message) with the specified private key
+// and hasher.
 //
-// This function returns an error if the signature cannot be generated.
+// This function returns an error if:
+//  - crypto.InvalidInputsError if the private key cannot sign with the given hasher
+//  - other error if an unexpected error occurs
 func (tb *TransactionBody) SignMessageWithTag(
```
janezpodhostnik, line 282: Does it make sense to rename this to `SignTransactionMessage`?

model/flow/constants.go
janezpodhostnik: It is in use, this is a breaking change for the `flow-go-sdk/examples`. We should keep that in mind.

fvm/transactionVerifier.go
```diff
@@ -42,6 +33,14 @@ func (v *TransactionSignatureVerifier) Process(
 	return v.verifyTransactionSignatures(proc, *ctx, sth)
 }
 
+func newInvalidEnvelopeSignatureError(txSig flow.TransactionSignature, err error) error {
```
janezpodhostnik, line 36: Nice, I like this approach!

fvm/crypto/hash.go
```diff
@@ -17,38 +20,36 @@ import (
...
 
 // paddedDomainTag converts a string into a padded byte array.
-func paddedDomainTag(s string) ([flow.DomainTagLength]byte, error) {
-	var tag [flow.DomainTagLength]byte
-	if len(s) > flow.DomainTagLength {
-		return tag, fmt.Errorf("domain tag %s cannot be longer than %d characters", s, flow.DomainTagLength)
+func paddedDomainTag(s string) ([tagLength]byte, error) {
+	var tag [tagLength]byte
+	if len(s) > tagLength {
+		return tag, fmt.Errorf("domain tag cannot be longer than %d characters, got %d", tagLength, len(s))
```
janezpodhostnik, line 30: The error no longer includes the tag text. Is this ok?

janezpodhostnik (approved):
Some minor comments.

access/validator.go
tarakby: Yes, it is for performance. Here are the tests https://github.com/onflow/flow-go/blob/master/fvm/transactionVerifier_test.go#L43-L68

fvm/context.go
tarakby: I already checked the emulator doesn't use the interface. If it does, I'll also update the emulator. 
Is there another tool that you think uses its own context?

model/flow/constants.go
tarakby: I'll update the sdk too after merging this PR: https://github.com/onflow/flow-go-sdk/issues/273

fvm/crypto/crypto.go
tarakby: I tagged you about this [here](https://github.com/onflow/flow-go/pull/2171#discussion_r832795320). I suggested to panic, Janez suggested the Failure error. As mentioned in the comment, that case can only happen if there is a bug in the code. These keys are all account keys, they are supposed to be checked and added previously, and they are not supposed to be problematic. 

What do you suggest to use here?

fvm/crypto/crypto.go
tarakby: Correct

fvm/crypto/crypto.go
tarakby: These are all keys already checked and added using `AddAccountKey`, so this check is only a sanity check, and if it fails then we have am internal bug somewhere.

This is the same as above. I suggested to panic and Janez suggest `Failure` error. What do you suggest?

model/flow/transaction.go
tarakby: How about just `Sign`, so it would be a method `TransactionBody.Sign` ?

fvm/crypto/hash.go
tarakby: I will add it back, that's better.

janezpodhostnik (approved):

model/flow/transaction.go
tarakby: https://github.com/onflow/flow-go/pull/2171/commits/dcd8a01ce5f9031960837913cf9fccb0a4e0f859

ramtinms (approved):
Looks good to me

tarakby:
bors merge

PR #2193, opened 2022-03-23
Test storage capacity update on FLOW transfer

robert-e-davidson3:
Test for verifying the fix for this bug: https://github.com/onflow/cadence/issues/1367


janezpodhostnik (approved):
Looks good. 

Isn't there a similar test somewhere already. This looks familiar. Maybe it was only in cadence...?

robert-e-davidson3:
> Looks good.
> 
> Isn't there a similar test somewhere already. This looks familiar. Maybe it was only in cadence...?

You may be remembering this test. I wrote it to verify the original change but it depended on cadence changes that weren't in yet.

SaveTheRbtz (approved):

PR #2195, opened 2022-03-24
Remove migrations already applied to canary for v0.25

janezpodhostnik:


turbolent:
@janezpodhostnik The next spork won't have the Secure Cadence yet, so the contracts should only be updated for the Cadence release to be deployed, i.e. v0.23

janezpodhostnik:
> @janezpodhostnik The next spork won't have the Secure Cadence yet, so the contracts should only be updated for the Cadence release to be deployed, i.e. v0.23

I was honestly confused what you were saying util I noticed that snuk in. Sorry, that was unintentional, I will remove it.

turbolent (approved):

PR #2199, opened 2022-03-24
Add ID check tests for AN response structures

koko1123:
Adding in more tests for the AN response structures with `ID()` interfaces and payload hash, to ensure sub-structures are converted correctly


engine/common/rpc/convert/convert_test.go
```diff
@@ -30,12 +30,114 @@ func TestConvertAccountKey(t *testing.T) {
...
+
+	msg := convert.EventToMessage(setupEvent)
+	converted := convert.MessageToEvent(msg)
+	assert.Equal(t, setupEvent.ID(), converted.ID())
+
+	commitEvent := unittest.EventFixture(flow.ServiceEventCommit, uint32(1), uint32(1), txId, 0)
+
+	msg = convert.EventToMessage(commitEvent)
+	converted = convert.MessageToEvent(msg)
+	assert.Equal(t, setupEvent.ID(), converted.ID())
```
smnzhu, line 98: nit: is there any way to systematically iterate through all Event types so that when new ones are added they are automatically covered by this test? For example, maybe define a new function `ServiceEventTypes()` in `service_event.go` right underneath the service event definitions which returns a list of all existing service events, with a comment saying that it must be updated every time a new event is added?

smnzhu (approved):

engine/common/rpc/convert/convert_test.go
koko1123: worth figuring out let's see if there's a golang way to iterate through it

peterargue (approved):

PR #2201, opened 2022-03-24
Remove comparison of key in MTrie payload.Equals() and add API comment about payload key

### LINKED ISSUE #2198: [Execution State] In MTrie, payload.Equals() shouldn't compare payload's key
## Problem

In MTrie, payload key shouldn't be included in payload equality test.

Also, the expectation that MTrie expects payload's key to never change doesn't appear to be documented.

Thanks @AlexHentschel for pointing out payload key is auxiliary data and for suggesting we add a comments.  👍 

## Solution

- Modify `payload.Equals()` so it doesn't compare payload's key.
- Add a comments to API to indicate MTrie expects payload's key to never change.

Although payload key was compared in `payload.Equals()` for about a year, this is an edge case that isn't triggered by our current use of MTrie.  So this change isn't expected to cause compatibility issues.

fxamacker:
Changes:

- Modify `payload.Equals()` so it doesn't compare payload's key.

- Add a comments to API to indicate MTrie expects payload's key to never change.

Although payload key was compared in payload.Equals() for about a year, this is an edge case that isn't triggered by our current use of MTrie. So this change isn't expected to cause compatibility issues.

Closes #2198


tarakby:
Thanks @fxamacker for working on the improvement and apologies for my late comment on the other PR. 

1. I'm not sure what was the outcome of the other PR discussion. @AlexHentschel mentioned that he prefers keeping the mtrie code to work for all use cases (whether paths are derived from keys or not) but I might have misunderstood the answer. I might have also missed an off-line follow up on this topic. 

2. Now, let's assume we are fine with making the mtrie implementation work only for paths that are derived from keys. I was thinking the `Payload.Equals` function could still compare both keys and values, since they are the fields of a `Payload`. My suggestion was about the [specific case](https://github.com/onflow/flow-go/blob/master/ledger/complete/mtrie/trie/trie.go#L365) in the mtrie update algorithm, where we are in a sub-case where paths are [guaranteed](https://github.com/onflow/flow-go/blob/master/ledger/complete/mtrie/trie/trie.go#L362) to be equal (which means keys should be equal too with a high probability). In that case, we could use a little optimization to compare only values of the payload using `func (p *Value) Equals(other *Value) bool`.  I've checked that the other current usage of `Payload.Equals` is in `TripProof.Equals` and it works well with only comparing values. However, a future usage of `Payload.Equals` may assume it includes keys comparison. What do you think?

tarakby (approved):
Thank you for the details and adjusting the tests 👌🏼

PR #2203, opened 2022-03-24
[BFT Testing] Implements integration tests for corruptible conduit factory framework

yhassanzadeh13:
This PR implements several integration tests composing components of the corruptible conduit framework, to evaluate the correctness of their composability and integrity. It also adds end-to-end integration testing for the framework.


insecure/attacknetwork/attackNetwork.go
```diff
@@ -54,31 +54,7 @@ func NewAttackNetwork(
 	// setting lifecycle management module.
 	cm := component.NewComponentManagerBuilder().
 		AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
-			// starts up gRPC server of attack network at given address.
```
yhassanzadeh13, line 57: Just moving some codes from the constructor of `AttackNetwork` to its `start` method in order to satisfy some causality between the component's startup.

insecure/attacknetwork/testHelpers.go
```diff
@@ -0,0 +1,90 @@
+package attacknetwork
```
yhassanzadeh13, line 1: This file does not contain any new changes, just some shared test fixtures moved to this file for better organization.

insecure/testHelpers.go
```diff
@@ -0,0 +1,89 @@
+package insecure
```
yhassanzadeh13, line 1: This file does not contain any new changes, just encapsulating some of our test fixtures into a separate file.

insecure/fixtures.go
```diff
@@ -0,0 +1,89 @@
+package insecure
```
yhassanzadeh13, line 1: This file does not contain any new changes, just moving some of our fixtures to here.

gomisha (approved):
all questions answered during PR review meeting

insecure/corruptible/factory.go
```diff
@@ -2,44 +2,111 @@ package corruptible
...
+	factory := &ConduitFactory{
+		myId:   myId,
+		codec:  codec,
+		logger: logger.With().Str("module", "corruptible-conduit-factory").Logger(),
+	}
+
+	cm := component.NewComponentManagerBuilder().
+		AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+			factory.start(ctx, address)
+			factory.ctx = ctx
```
peterargue, line 63: why store the context on the factory instead of using the one passed into start?

insecure/corruptible/factory.go
```diff
@@ -142,13 +209,14 @@ func (c *ConduitFactory) registerAttacker(ctx context.Context, address string) e
...
-	clientConn, err := grpc.Dial(address)
+	clientConn, err := grpc.Dial(address,
+		grpc.WithTransportCredentials(grpcinsecure.NewCredentials()))
 	if err != nil {
 		return fmt.Errorf("could not establish a client connection to attacker: %w", err)
 	}
 
 	attackerClient := insecure.NewAttackerClient(clientConn)
-	c.attackerObserveClient, err = attackerClient.Observe(ctx)
+	c.attackerObserveClient, err = attackerClient.Observe(c.ctx)
```
peterargue, line 219: why use the factory's context instead of the one passed in? generally, it's best to [avoid storing contexts](https://go.dev/blog/context-and-structs)

peterargue (commented):
looks good. one nit and a comments about context passing

insecure/attacknetwork/corruptedConnector.go
```diff
@@ -52,6 +60,10 @@ func (c *CorruptedConnector) Connect(ctx context.Context, targetId flow.Identifi
 	return &CorruptedNodeConnection{stream: stream}, nil
 }
 
+func (c *CorruptedConnector) WithAttackerAddress(address string) {
```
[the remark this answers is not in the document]
yhassanzadeh13: The short answer is: that the attacker address _may be_ determined post-startup rather than the initialization time. 

The long answer is: for testing, we use `localhost:0` as the attacker address so that the actual attacker port number will be picked by the kernel at the startup of the attacker. Hence, in testing, we may not have the correct port number of the attacker till it successfully starts up. This is to avoid port collision and flakey behavior during the test execution. Moreover, the `AttackNetwork` is mostly meant to be utilized as an integration test tool, hence a dynamic runtime port selection benefits the health of testing as well. Accordingly, this configuration is set to be independent of the constructor of the `CoruptedConnector`, so that the port number of the attacker is set when the attacker is set post-startup.

insecure/corruptible/factory.go
yhassanzadeh13: The short answer is: to guarantee a graceful shutdown and avoid message loss. 


The long answer is: this is a `gRPC` method invoked by a remote attacker. As the result of invocation, it tries bouncing back a `gRPC` connection to the attacker (i.e., `attackerClient`). The input context is coming from the _remote_ attacker's side and is influenced by the attacker's lifecycle, for the rest of this comment, let's call it **the attacker's context**. 

Now if we use the incoming attacker's context to create a `gRPC` connection back to the attacker, we are making the lifetime of this connection totally coupled with the attacker's life cycle. This poses an issue: when the attacker shutdown _starts_ by its component manager, the contexts generated at the attacker's side are getting canceled right away. The cancellation of context should gracefully stop the attack network by first letting messages en route on connections to be delivered, and then tearing down the connections. 

However, when we use attacker's context to create a connection back to it, the resulted connection will be canceled the moment that the parent context at the attacker side is canceled and _before_ the `gRPC` server at the attacker's side shuts down gracefully. So, there is a chance of message loss. 

This message loss is verifiable by switching the code above to use the passed `ctx` and then the test gets flakey since the moment attacker starts shutting down, all connections are dropped immediately and messages en route will be dropped. 

Accordingly, the context to create connections back to attacker should be local to the `factory` and managed by it. The simplest solution is to create a `context.Background()` every time  `factory` is creating a connection to the attacker, i.e., 

```
c.attackerObserveClient, err = attackerClient.Observe(context.Background())
```
However, the problem is those connections may outlive the `factory` itself, since they may reside even when the factory itself is shut down. That is the reason behind having a local context as a member field to the `factory` and using it to create connections back to the attacker, hence making sure that: 
1. Those connections are not controlled by the attacker's lifecycle. 
2. Those connections are torn down as the procedure of `factory`'s shutdown. 

P.S: when the attacker is shut down, the `gRPC` connection is closed at the attacker side, and when all messages are delivered, an `EOF` is delivered on the connection telling the `factory` that the connection is over.

insecure/corruptible/factory.go
yhassanzadeh13: please refer to my answer to the next comment.

peterargue (approved):

PR #2206, opened 2022-03-25
Refactor root checkpoint extraction program

### LINKED ISSUE #2205: [Execution State] Refactor root checkpoint extraction program
## Problem

Root checkpoint extraction program can be improved.  It's not checking for the max supported checkpoint file format (it assumes forward compatibility) and it is not closing the checkpoint file.

Thanks @m4ksio for suggesting upper version boundry to make the program a bit more future-proof!  👍 

## The Proposed Solution

- Add `MaxVersion` const as v5 and use it to check max supported checkpoint file format
- Extract the code that reads the checkpoint file into checkpointer.go
- Close the checkpoint file

fxamacker:
Changes include:

- Add MaxVersion to reject unsupported version early
- Extract code reading the checkpoint file from CLI program into checkpointer.go
- Close checkpoint file

Closes #2205


ramtinms (approved):

ledger/complete/wal/checkpointer.go
```diff
@@ -776,6 +782,48 @@ func readCheckpointV5(f *os.File) ([]*trie.MTrie, error) {
 	return tries, nil
 }
 
+// ReadLastTrieRootHashFromCheckpoint returns last trie's root hash from checkpoint file f.
+func ReadLastTrieRootHashFromCheckpoint(f *os.File) (hash.Hash, error) {
```
AlexHentschel, line 786: could we please include a short documentation of the returned error. Something like the following would already be sufficient:

```suggestion

// All returned errors indicate that the given checkpoint file is either corrupted or incompatible. As the function is side-effect free, all failures are simply a no-op.
func ReadLastTrieRootHashFromCheckpoint(f *os.File) (hash.Hash, error) {
```

cmd/util/cmd/execution-state-extract/cmd.go
```diff
@@ -119,45 +117,16 @@ func run(*cobra.Command, []string) {
...
-			}
+		rootHash, err := wal.ReadLastTrieRootHashFromCheckpoint(f)
+		if err != nil {
+			log.Fatal().Err(err).Msgf("failed to read last root hash in root checkpoint: %s", err.Error())
 		}
 
-		n, err = f.Read(stateCommitment[:])
-		if err != nil || n != hash.HashLen {
-			log.Fatal().Err(err).Msg("failed to read state commitment from root checkpoint")
+		stateCommitment, err = flow.ToStateCommitment(rootHash[:])
```
AlexHentschel, line 127: may I request a brief update / extension of `ToStateCommitment`'s goDoc: 
https://github.com/onflow/flow-go/blob/8a2063475ac150fecd883950dbda81153357a7ff/model/flow/ledger.go#L111-L113
* Please correct the typo `StateComitment` -> `StateCommitment` 
* Please include a brief description of the expected error returns. Something like the following would already suffice:
    ```golang
    // All returned errors indicate that the given byte slice is not a valid root hash of an execution state.
    // As the function is side-effect free, all failures are simply a no-op.
    ```

AlexHentschel (approved):
Looks great. Thank you. Sorry for the delayed review. 
My only request is to document error returns in accordance with our [Guidelines](https://github.com/onflow/flow-go/blob/a9748ffe23d830fed2ea61912078337e56fd5a46/CodingConventions.md#error-handling)

cmd/util/cmd/execution-state-extract/cmd.go
fxamacker: Thanks! 👍   Done in commit 7d8cba1.

PR #2211, opened 2022-03-26
Docker labels

m4ksio:
Adds docker labels with git pointers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,6 +209,7 @@ docker-ci-integration:
 .PHONY: docker-build-collection
 docker-build-collection:
 	docker build -f cmd/Dockerfile  --build-arg TARGET=collection --build-arg COMMIT=$(COMMIT)  --build-arg VERSION=$(IMAGE_TAG) --build-arg GOARCH=$(GOARCH) --target production \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -209,6 +209,7 @@ docker-ci-integration:
 .PHONY: docker-build-collection
 docker-build-collection:
 	docker build -f cmd/Dockerfile  --build-arg TARGET=collection --build-arg COMMIT=$(COMMIT)  --build-arg VERSION=$(IMAGE_TAG) --build-arg GOARCH=$(GOARCH) --target production \
+		--label "git_commit=${COMMIT}" --label "git_tag=${IMAGE_TAG}" \
```
zhangchiqing, line 212: What is label for? The `--build-arg` already contains COMMIT and IMAGE_TAG?

m4ksio: `--build-arg` passes the value to a `Dockerfile` for building. `--label` adds a metadata to an image which can be later seen by inspecting the image itself

zhangchiqing (approved):

Makefile
zhangchiqing: I see. Good point. Thanks

PR #2214, opened 2022-03-28
[FVM] Fix computation/memory meter error text - v0.25

janezpodhostnik:
related: [[FVM] Fix computation/memory meter error text](https://github.com/onflow/flow-go/pull/2215)

During testing with the emulator, I discovered that the metering "limit reached" error text did not include the correct limit.

The error used the limit at the meters internal precision, instead of the actual limit.

I also changed the tests to tests for this.


janezpodhostnik:
included with https://github.com/onflow/flow-go/pull/2216

PR #2216, opened 2022-03-28
[FVM] Fix metering limits - v0.25

janezpodhostnik:
[master](https://github.com/onflow/flow-go/pull/2217) version

This fixes three things:

1. During testing with the emulator, I discovered that the metering "limit reached" error text did not include the correct limit. The error used the limit at the meters internal precision, instead of the actual limit. (now covered in the weighted meter tests)

2. When testing I noticed that setting a gas limit of 999 I was only able to get a transaction to 941 execution effort before it failed. This was due to the limit enforcement happening after the fee deduction which added computation and potentially failed the transaction.

3. When the transaction failed, the fee deduction code also failed due to `meter.MergeMeter` still emitting an error even though limit enforcement was off.

The new `fvm/fvm_test.go` test covers 2. and 3. It also makes sure that if the limit is reached (or overshot) the user is only charged up to their gas limit.

### DIFF
--- a/fvm/transactionInvoker.go
+++ b/fvm/transactionInvoker.go
@@ -188,7 +188,12 @@ func (i *TransactionInvoker) Process(
 
 	// if there is still no error check if all account storage limits are ok
 	if txError == nil {
+		// disable the computation/memory limit checks on storage checks,
+		// so we don't error from computation/memory limits on this part.
+		// We cannot charge the user for this part, since fee deduction already happened.
+		sth.DisableAllLimitEnforcements()
 		txError = NewTransactionStorageLimiter().CheckLimits(env, sth.State().UpdatedAddresses())
+		sth.EnableAllLimitEnforcements()
 	}
 
 	// it there was any transaction error clear changes and try to deduct fees again
--- a/fvm/fvm_test.go
+++ b/fvm/fvm_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strconv"
+	"strings"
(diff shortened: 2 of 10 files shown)


fvm/transactionInvoker.go
```diff
@@ -188,7 +188,12 @@ func (i *TransactionInvoker) Process(
 
 	// if there is still no error check if all account storage limits are ok
 	if txError == nil {
+		// disable the computation/memory limit checks on storage checks,
+		// so we don't error from computation/memory limits on this part.
+		// We cannot charge the user for this part, since fee deduction already happened.
+		sth.DisableAllLimitEnforcements()
 		txError = NewTransactionStorageLimiter().CheckLimits(env, sth.State().UpdatedAddresses())
+		sth.EnableAllLimitEnforcements()
```
turbolent, line 196: What exactly does `CheckLimits` do? If code / data executed here is influenced by the user (for example, we always run a loop, but it iterates over user-provided data), then it could be potentially exploited

janezpodhostnik: No, Its not user provided data.

`CheckLimits` iterates trough all accounts where the storage used changed, and checks if they are within the limit still.

It is a predetermined action.

The only influence over that action the transaction has is how many different accounts it touches. If it would touch a lot of different accounts, this would take longer, and potentially throw off metering. However it would not circumvent metering, as it would still cost to "touch" an account in the first place.

turbolent: I see, that makes sense, thank you for the explanation!

PR #2217, opened 2022-03-28
[FVM] Fix metering limits

janezpodhostnik:
[v0.25](https://github.com/onflow/flow-go/pull/2216) version

This fixes three things:

1. During testing with the emulator, I discovered that the metering "limit reached" error text did not include the correct limit. The error used the limit at the meters internal precision, instead of the actual limit. (now covered in the weighted meter tests)

2. When testing I noticed that setting a gas limit of 999 I was only able to get a transaction to 941 execution effort before it failed. This was due to the limit enforcement happening after the fee deduction which added computation and potentially failed the transaction.

3. When the transaction failed, the fee deduction code also failed due to `meter.MergeMeter` still emitting an error even though limit enforcement was off.

The new `fvm/fvm_test.go` test covers 2. and 3. It also makes sure that if the limit is reached (or overshot) the user is only charged up to their gas limit.

### DIFF
--- a/fvm/fvm_test.go
+++ b/fvm/fvm_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/onflow/cadence"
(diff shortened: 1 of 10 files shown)


ramtinms (approved):

janezpodhostnik:
bors merge

PR #2223, opened 2022-03-29
Reduce default checkpoint distance flag from 40 to 20 to speedup EN startup

### LINKED ISSUE #2207: [Execution Node] EN startup time can be further reduced by reducing number of WAL segments
## Problem

EN startup time can be improved.

Currently, checkpoint creation is triggered when more than 40 WAL files accumulate.  Replaying each WAL file can take very roughly about 10-20 seconds each (depending on system specs, other activity, etc.)

Prior to PR #1944, it wasn't feasible to create checkpoints more frequently because checkpoint creation took 12-15+ hours on mainnet.  After PR #1944, checkpoint creation should take very roughly around 15 minutes while consuming much less RAM.

EN startup time can be faster by having fewer WALs to replay during EN startup.

## The Proposed Solution

- Compare the benchmarks (checkpoint file loading + replaying WALS) using 40 WALs vs 20 WALs.
- If needed, configure EN to use a different number of WALs to trigger checkpoint creation.

## Preliminary Benchmarks

On benchnet, using loading checkpoint and replaying WALs takes about:

- 10 mins 10 secs (40 WALs, current configuration)
- 6 mins 40 secs (20 WALs, possible configuration)

Updates https://github.com/dapperlabs/flow-go/issues/6114

fxamacker:
Reduce checkpoint distance flag to trigger creation of checkpoints more frequently.

Prior to #1944 checkpoint creation took 11-15+ hours so it wasn't feasible to run checkpoints more frequently.

This change will cause fewer WAL segments to trigger checkpoint creation, which will speedup EN startup by having fewer WALs to replay during startup.

On benchnet, replaying 20 WALs is 3.5 minutes faster than replaying 40 WALs.

This change will also speedup checkpoint file creation by the same duration because that also requires replaying WALs.

closes #2207 
updates https://github.com/dapperlabs/flow-go/issues/6114

### DIFF
--- a/cmd/execution/main.go
+++ b/cmd/execution/main.go
@@ -145,7 +145,7 @@ func main() {
 			flags.StringVar(&triedir, "triedir", datadir, "directory to store the execution State")
 			flags.StringVar(&executionDataDir, "execution-data-dir", filepath.Join(homedir, ".flow", "execution_data_blobstore"), "directory to use for Execution Data blobstore")
 			flags.Uint32Var(&mTrieCacheSize, "mtrie-cache-size", 500, "cache size for MTrie")
-			flags.UintVar(&checkpointDistance, "checkpoint-distance", 40, "number of WAL segments between checkpoints")
+			flags.UintVar(&checkpointDistance, "checkpoint-distance", 20, "number of WAL segments between checkpoints")
 			flags.UintVar(&checkpointsToKeep, "checkpoints-to-keep", 5, "number of recent checkpoints to keep (0 to keep all)")
 			flags.UintVar(&stateDeltasLimit, "state-deltas-limit", 100, "maximum number of state deltas in the memory pool")
 			flags.UintVar(&cadenceExecutionCache, "cadence-execution-cache", computation.DefaultProgramsCacheSize, "cache size for Cadence execution")


zhangchiqing (approved):

cmd/execution/main.go
```diff
@@ -145,7 +145,7 @@ func main() {
 			flags.StringVar(&triedir, "triedir", datadir, "directory to store the execution State")
 			flags.StringVar(&executionDataDir, "execution-data-dir", filepath.Join(homedir, ".flow", "execution_data_blobstore"), "directory to use for Execution Data blobstore")
 			flags.Uint32Var(&mTrieCacheSize, "mtrie-cache-size", 500, "cache size for MTrie")
-			flags.UintVar(&checkpointDistance, "checkpoint-distance", 40, "number of WAL segments between checkpoints")
+			flags.UintVar(&checkpointDistance, "checkpoint-distance", 20, "number of WAL segments between checkpoints")
```
m4ksio, line 148: just maybe we should use some const rather than magic number

m4ksio (approved):

cmd/execution/main.go
fxamacker: Great point!  I'll open an issue/PR to replace magic numbers in the flags. E.g.

image

PR #2225, opened 2022-03-29
Resolve cycle dependency from NewClusterList

zhangchiqing:


model/flow/cluster.go
```diff
@@ -45,61 +42,6 @@ func (clusters ClusterList) Assignments() AssignmentList {
 	return assignments
 }
 
-// NewClusterList creates a new cluster list based on the given cluster assignment
```
zhangchiqing, line 48: moved to model/flow/factory/cluster_list.go in order to resolve cycle dependency, since NewClusterList depends on order.Canonical, which depends on flow.Identifier

jordanschalm (approved):

PR #2227, opened 2022-03-29
Fail state extract early

smnzhu:
if the root checkpoint file already exists, we should fail early instead of waiting until the very end after all the work has already been done

### DIFF
--- a/module/state_synchronization/storage.go
+++ b/module/state_synchronization/storage.go
@@ -1,0 +1,35 @@
+package state_synchronization
+
+import "github.com/onflow/flow-go/model/flow"
+
+type Storage struct {
+}
+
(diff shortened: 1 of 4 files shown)


module/state_synchronization/storage.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (s *Storage) ReceivedRoot(rootID flow.Identifier, chunkExecutionDataIDs []flow.Identifier) {
+
+}
+
+func (s *Storage) ReceivedLevel(parentLevel, level []cid.Cid){
+
+}
+
+func (s *Storage)
```
janezpodhostnik, line 35: This file does not seem relevant to the change, and it stops abruptly.

PR #2229, opened 2022-03-30
[FVM] Move execution intensities debug logging

janezpodhostnik:
This PR moves the execution intensities debug logging to before the fee deduction and storage limit checks.

The reasoning behind this is that:
- payer is only limited on the execution that happens before this point
- payer (or rather the transaction) cannot affect the fee deduction or storage limit checks
- payer is only charged for the execution that happens before this point

This only changes logging so it is safe to port to v0.25, which is something I would like to do, after this is merged to master.


ramtinms (approved):
Looks good to me

SupunS (approved):

PR #2233, opened 2022-03-30
Fix bootstrap constraint check

zhangchiqing:
See comments.


cmd/bootstrap/cmd/constraints.go
```diff
@@ -50,12 +50,12 @@ func checkConstraints(partnerNodes, internalNodes []model.NodeInfo) {
 			if _, exists := internals.ByNodeID(node.NodeID); exists {
 				clusterInternalCount++
 			}
-			if clusterInternalCount <= clusterPartnerCount*2 {
```
zhangchiqing, line 53: This check should be done once we have the accurate number, rather than in the middle of calculate the count.

We never ran into a problem until it's captured by the changes in the test cases from the signer indices optimization.

jordanschalm (approved):
Good catch - thank you

durkmurder (approved):

PR #2238, opened 2022-03-31
Update the private message struct: remove Orig from DKGMessage

kc1116:
This PR makes some updates to the message structs in the DKG package;

- Remove the Orig field from DKGMessage
- Add Orig field to BroadcastDKGMessage wrapper
- Add new wrapper PrivateDKGMessage with Orig field

These changes increase the resiliency to byzantine message attacks in the DKG messaging engine by setting the Orig field in the DKG broker when private incoming messages are received as opposed to the sender setting the Orig field when they send a private message. Using the OriginID we get from the network layer we look up the committee member index of the sender of the private message. It is preferable to include the least possible data in the message to avoid verification on the receiver side. BroadCastDKGMessage' are a different case where these messages are communicated using the DKG smart contract as the intermediary. This requires us to explicitly set the Orig field for broadcast messages and check this field when reading broadcast messages.

[Resilience against Impersonation and Masquerade Attacks (docs) ](https://www.notion.so/dapperlabs/Resilience-against-Impersonation-and-Masquerade-Attacks-61189d46e5374c2b9af36edacf71435f)


model/messages/dkg.go
```diff
@@ -37,8 +35,22 @@ type PrivDKGMessageOut struct {
 
 // BroadcastDKGMessage is a wrapper around a DKGMessage intended for broadcasting.
 // It contains a signature of the DKGMessage signed with the staking key of the
-// sender.
+// sender. DTTF that BroadcastDKGMessage are communicated via the DKG smart contract
```
jordanschalm, line 38: What does DTTF mean?

module/dkg/broker.go
```diff
@@ -415,31 +415,36 @@ func (b *Broker) listen() {
...
-	err := b.checkMessageInstanceAndOrigin(msg)
-	if err != nil {
-		b.log.Err(err).Msg("bad message")
-		return
-	}
-	// check that the message's origin matches the sender's flow identifier
-	nodeID := b.committee[msg.Orig].NodeID
-	if !bytes.Equal(nodeID[:], originID[:]) {
-		b.log.Error().Msgf("bad message: OriginID (%v) does not match committee member %d (%v)", originID, msg.Orig, nodeID)
+	if memberIndex, ok := b.committee.GetIndex(originID); ok {
```
jordanschalm, line 418: We can invert this conditional to keep the main logic flow at the top level of indentation.

```
memberIndex, exists := b.committee.GetIndex(originID)
if !exists {
    // log error
    return
}

// happy-path logic
// ... 
```

model/messages/dkg.go
```diff
@@ -37,8 +35,22 @@ type PrivDKGMessageOut struct {
...
+// we must also include the Orig or DKG committee index of the sender so that when these
+// messages are consumed we can verify if BroadcastDKGMessage was sent by a DKG committee member.
```
jordanschalm, line 40: We can allow the whiteboard to tell us the sender of a broadcast message, in the same way the network layer does for private message, which may simplify things here. 

The first field of the value we read from the smart contract https://github.com/onflow/flow-go/blob/d75191e2b0077c47fe5c46678baeca1df5cd0bb9/module/dkg/client.go#L141-L145 is the node ID of the message sender. (eg. `nodeID := val.(cadence.Struct).Fields[0]`

module/dkg/broker.go
```diff
@@ -415,31 +415,36 @@ func (b *Broker) listen() {
...
+		if err != nil {
+			b.log.Err(err).Msg("bad message")
+			return
+		}
+		// check that the message's origin matches the sender's flow identifier
+		nodeID := b.committee[memberIndex].NodeID
+		if !bytes.Equal(nodeID[:], originID[:]) {
+			b.log.Error().Msgf("bad message: OriginID (%v) does not match committee member %d (%v)", originID, memberIndex, nodeID)
+			return
+		}
```
jordanschalm, line 429: Not new code here, but I don't think this check is necessary besides as a sanity check against internal bugs. Since we are reading `memberIndex` from `committee`, this can only happen if `GetIndex` has a bug. I'm inclined to remove the check, but let me know what you think.

module/dkg/broker_test.go
```diff
@@ -177,80 +176,14 @@ func TestReceivePrivateMessage_Valid(t *testing.T) {
...
 	)
 
 	unittest.RequireCloseBefore(t, doneCh, 50*time.Millisecond, "message not received")
 }
 
-// TestProcessPrivateMessage_InvalidOrigin checks that incoming DKG messages are
-// discarded if their origin is invalid, or if there is a discrepancy between
-// the origin defined in the message, and the network identifier of the origin
-// (as provided by the network utilities).
-func TestProcessPrivateMessage_InvalidOrigin(t *testing.T) {
```
jordanschalm, line 192: 👍 for removing the case testing origin ID / origin index mismatch. But I think we should still have a test case for messages from a non-committee member.

model/messages/dkg.go
kc1116: "due to the fact", i'll update

module/dkg/broker.go
kc1116: I agree I wanted to remove it but left it just in case in what was there for some reason I could not see.

kc1116: https://github.com/onflow/flow-go/pull/2238/commits/7a287cfa9c3d71c3b4f593d01f47ba48e182dd29

module/dkg/broker.go
kc1116: https://github.com/onflow/flow-go/pull/2238/commits/7a287cfa9c3d71c3b4f593d01f47ba48e182dd29

model/messages/dkg.go
kc1116: https://github.com/onflow/flow-go/pull/2238/commits/578700dc80f04d5ffc1c75c4975b31e2f159f81e

I added json ignore tags so that Orig and NodeID aren't sent when broadcasting messages, instead I am using the nodeID returned from the contract to later set the Orig field when processing the broadcast messages during reading

module/dkg/broker_test.go
kc1116: https://github.com/onflow/flow-go/pull/2238/commits/4ecc851cf54089b33217d7479e3bdaeeaab358cc

model/messages/dkg.go
tarakby: I agree we do not need to add `nodeID` or `orig` as fields. However, how can we read the whiteboard data when printed in a file?
Debugging is still important at this stage. If I could keep one of the 2 fields, I would prefer to keep `orig`. It's only a few characters more per message.

tarakby: I agree we do not need to add `nodeID` or `orig` as fields. However, how can we read the whiteboard data when printed in a file?
Debugging is still important at this stage. If I could keep one of the 2 fields, I would prefer to keep `orig`. It's only a few characters more per message.

model/messages/dkg.go
```diff
@@ -37,8 +35,22 @@ type PrivDKGMessageOut struct {
...
 	Signature crypto.Signature
 }
+
+// PrivateDKGMessage is a wrapper around DKGMessage intended for use when communicating
+// incoming private DKG messages from the messaging engine. This wrapper adds Orig or DKG committee index of the sender
+// which is needed when processing the DKG message in the crypto library.
+// PrivDKGMessageIn component flow:		message_engine -> broker -> controller
+// At the point where the private message reaches the broker, the broker will get the DKG committee index of the sender
+// validate and attach it to an instance of PrivateDKGMessage that will then be forwarded to the controller.
+type PrivateDKGMessage struct {
```
tarakby, line 53: - This structure seems to be used only when receiving a private message (not when sending). It is mainly another version of `PrivDKGMessageIn` that substitute the ID with an index. It would be good to have it called something to reflect the "In" aspect, maybe one should be declared in the broker, and the other in the controller, instead of both in model/messages ?

- A different design choice is made for broadcast messages  while it is the same case as Private messages. The struct `BroadcastDKGMessage` is used by both the controller and broker, by having both the ID or index fields. I don't have a preference for either design choice, but it would be good if we keep the same consistent choice for both broadcast and private messages.

module/dkg/broker.go
```diff
@@ -473,7 +480,7 @@ func (b *Broker) prepareBroadcastMessage(data []byte) (messages.BroadcastDKGMess
 //   or the signature could not be checked
 // TODO differentiate errors
 func (b *Broker) verifyBroadcastMessage(bcastMsg messages.BroadcastDKGMessage) (bool, error) {
-	err := b.checkMessageInstanceAndOrigin(bcastMsg.DKGMessage)
+	err := b.checkMessageInstanceAndOrigin(bcastMsg.Orig, bcastMsg.DKGMessage)
```
tarakby, line 483: - this function should not check the origin anymore, because we look up the origin from the local ID->index map. Since the local map is trusted by construction, the origin is also trusted. I suggest to rename it to `checkMessageInstance`. It is also called for incoming private messages, but the origin there shouldn't be checked either for the same reason.
- Let's update the doc too on line 474 and remove the origin mention.

kc1116: https://github.com/onflow/flow-go/pull/2238/commits/df9129e27e41825e65077a23fa471f3f37b239b2
https://github.com/onflow/flow-go/pull/2238/commits/152aca29e7879589906fabd7e14d52ea1d929cbb

model/messages/dkg.go
kc1116: I decided to add CommitteeMemberIndex to PrivDKGMessageIn and remove the wrapper
https://github.com/onflow/flow-go/pull/2238/commits/df9129e27e41825e65077a23fa471f3f37b239b2

tarakby (approved):
Thanks for the updates 👌🏼

kc1116:
bors merge

PR #2239, opened 2022-03-31
Add extraction report for usage within external tooling

sjonpaulbrown:
## Description
This PR adds an extraction report that is the first report that is executed after the root.checkpoint is generated. The intent behind the generation of this report is to output information that is necessary for external processes and tooling. This report can be extended as needed as additional information is required. An example of the report is below:

```
{
  "CurrentStateCommitment": "dccf36df9aabcb11a5bd5a97bbc446eb384b37d01859b3e1520bcfa446906a2b",
  "EpochCounter": 511,
  "PreviousStateCommitment": "dccf36df9aabcb11a5bd5a97bbc446eb384b37d01859b3e1520bcfa446906a2b"
}
```

This report has been tested against canary state to ensure that it can properly generate the report along with ensuring that the EpochCounter extraction changes do not break the other reporters.

Along with this report, changes have been made to support the output of a status file **after** the root checkpoint is written to disk. This is to be used with wrapper automation to safely determine when the root checkpoint can be uploaded. Also, changes have been made to support the execution of this single extraction report **before** we write the root checkpoint to disk. This is intended to optimize the spork process by outputting this information as soon as possible.


cmd/util/ledger/reporters/export_reporter.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (e *ExportReporter) Name() string {
+	return "ExportReporter"
+}
+
+func (e *ExportReporter) Report(payload []ledger.Payload, o ledger.ExportOutputs) error {
+	script, _, err := ExecuteCurrentEpochScript(e.Chain, payload)
+
+	if err == nil && script.Err == nil && script.Value != nil {
```
janezpodhostnik, line 33: I recommend inverting this if statement and exiting early in case we failed to get the epoch counter. 

That way the else statement isn't needed. So in short:

```
if not_good {
// log and return
}
// report the good stuff
```

cmd/util/ledger/reporters/epoch_counter_reporter.go
```diff
@@ -17,30 +17,17 @@ import (
 
 // EpochCounterReporter reports the current epoch counter from the FlowEpoch smart contract.
 type EpochCounterReporter struct {
-	Chain flow.Chain
-	Log   zerolog.Logger
+	Chain                   flow.Chain
+	Log                     zerolog.Logger
+	PreviousStateCommitment flow.StateCommitment
```
janezpodhostnik, line 22: This does not look like its used anywhere on the `EpochCounterReporter`.

janezpodhostnik (commented):
Just some minor things. Looks good otherwise!

cmd/util/ledger/reporters/export_reporter.go
sjonpaulbrown: I agree. This is poorly structured. I will update this to remove the need for the else clause.

janezpodhostnik (approved):
LGTM

ledger/ledger.go
```diff
@@ -321,10 +321,14 @@ func (v Value) MarshalJSON() ([]byte, error) {
 // Migration defines how to convert the given slice of input payloads into an slice of output payloads
 type Migration func(payloads []Payload) ([]Payload, error)
 
+type ExportOutputs struct {
```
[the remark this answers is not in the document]
ramtinms: I'm not sure if I follow, from the separation of concern perspective, Reporter doesn't know anything about an export or a migration, so I would suggest we keep this as part of the specific reporter that needs to export this data instead of requiring it as input for all the reporters. This could be injected as a function to return the previousStateCommitment to the reporter and could be updated with when migration is happening. 

Another alternative if we would like cross reporter and migration data passing, is to update reporter and migration to accept a context.Context object and collect the data as needed.

vishalchangrani (approved):

sjonpaulbrown:
@ramtinms I have refactored the method in which I am passing the state commitment into that report. This separates the concerns and removes the need to inject outputs into the Report method. As you have time, if you could review and provide feedback, I can make changes as needed.

ramtinms (approved):
Thanks for doing it

sjonpaulbrown:
@ramtinms I have updated this PR and would appreciate another review as I have added some additional things. One of the changes addresses a bug that I noticed as the ledger state was not being properly injected, and instead of getting the state commitment, we were getting the default hash in the export report. After ensuring that the `newState` var was passed in, everything worked as expected.

Also, after receiving feedback during the last spork, I added two additional features here. I have updated the export to support the execution of the export report **before** writing the checkpoint to disk. This is an optimization that would allow us to finalize the root block during a spork without waiting for the root checkpoint to be written to disk. This is done by updating the reports to be a map that can be used run that single report.

Along with that, I have added support for writing an additional file called `checkpoint_status.json` that will be used as an indicator for the completion of the root checkpoint. As the root checkpoint file can be "available" **before** it has been fully written, I wanted to create a separate file that can be reliably used to indicate that the checkpoint is ready and is fully copied/written to disk.

These optimizations will help streamline the spork automation used during state extraction, and they have all been tested to ensure that they work as expected within the context of this automation.

cc: @zhangchiqing - During the last spork, you mentioned updating this tool to run the epoch report **before** writing the root checkpoint. This will add that support. If you would also like to review or have any additional thoughts, feel free to let me know, and I can make changes accordingly.

ledger/complete/ledger.go
```diff
@@ -433,3 +441,30 @@ func (l *Ledger) keepOnlyOneTrie(state ledger.State) error {
...
+
+	l.Info().
+		Str("timeTaken", elapsed.String()).
+		Str("name", r.Name()).
+		Msg("reporter done")
+	if err != nil {
+		return ledger.State(hash.DummyHash),
+			fmt.Errorf("error running reporter (%s): %w", r.Name(), err)
+	}
+	return ledger.State(hash.DummyHash), nil
```
ramtinms, line 462: this seems to always return dummyHash?

sjonpaulbrown: I will remove this so that it only returns an error or nil from the function. There is no reason that a hash needs to be returned.

ledger/complete/ledger.go
```diff
@@ -265,7 +267,8 @@ func (l *Ledger) Checkpointer() (*wal.Checkpointer, error) {
 func (l *Ledger) ExportCheckpointAt(
 	state ledger.State,
 	migrations []ledger.Migration,
-	reporters []ledger.Reporter,
+	reporters map[string]ledger.Reporter,
```
ramtinms, line 270: I'm not sure if I follow the change here, if we are interested in access to name of reporter we could add Name as a method to reporter interface

ledger/complete/ledger.go
```diff
@@ -265,7 +267,8 @@ func (l *Ledger) Checkpointer() (*wal.Checkpointer, error) {
...
+	extractionReportName string,
```
ramtinms, line 271: same here, it seems this to be related to a single reporter and not something ledger package should be aware of

sjonpaulbrown:
The decision has been made to merge this and refactor the [ExportCheckpointAt method](https://github.com/onflow/flow-go/blob/97b9e28a6e73b64389485ee63136dfa05d389d94/ledger/complete/ledger.go#L267). We would like to refactor this to more cleanly execute reports at different stages or execute the checkpoint writing in parallel with the report execution.

PR #2240, opened 2022-03-31
Rebuild tx index index

m4ksio:
A small utility which populated missing entries in (block_id, tx_index) which was added mid-spork for mainnet16/testnet33.
I have run it on snapshot of data for mainnet16 and it worked fine.

New version started writing new index with deployment so going forward the data is there, and it will be there in next sporks.
We only miss entries from beginning of a spork up until deployment, so it's a one off operation.

While reviews are required, not sure if its worth merging to master after we run it current networks.


cmd/util/cmd/read-badger/cmd/transactions.go
```diff
@@ -39,3 +53,162 @@ var transactionsCmd = &cobra.Command{
...
+
+				totalTxs++
+				txIndex++
+			}
+
+			blockNo++
+
+			if blockNo%10000 == 0 {
+				log.Info().Int("blocks", blockNo).Int("total_tx", totalTxs).Msg("processing progress")
+			}
```
koko1123, line 189: What dies this check do? it is just logging for every 10k blocks?

koko1123 (approved):
looks fine to me, I'm assuming you'll get the lint issues :)

cmd/util/cmd/read-badger/cmd/transactions.go
m4ksio: Yes, it takes quite long to process (20-30mins)

cmd/util/cmd/read-badger/cmd/transactions.go
```diff
@@ -39,3 +54,162 @@ var transactionsCmd = &cobra.Command{
...
+				if _, has := megaMap[txID][blockID]; has {
+					panic(fmt.Sprintf("duplicated tx %s in block %s", txID.String(), blockID.String()))
+				}
+
+				megaMap[txID][blockID] = struct{}{}
+
+				_, err := storages.TransactionResults.ByBlockIDTransactionIndex(blockID, txIndex)
+				if err != nil {
+					if errors.Is(err, storage.ErrNotFound) {
+						transactionResult, err := storages.TransactionResults.ByBlockIDTransactionID(blockID, txID)
```
peterargue, line 160: should this report the error?

m4ksio: No. It was previously, but turns out not every block received gets executed, so there might not be a tx result for given block and that's OK.
That's why I added a check for heights - there should be at least one tx for every height.

peterargue (approved):
Looks good. Added one comment about a missing error check. generally could use some more comments to make it easier to understand the flow, but it's probably only going to be used once

cmd/util/cmd/read-badger/cmd/transactions.go
m4ksio: Yep, record not found we can skip, but some other problems should bubble up, good catch

PR #2242, opened 2022-04-01
Split CI integration job into categories

smnzhu:
Splits CI integration job into multiple categories.

Currently with how the job is setup, the first category which fails will exit the job, skipping the remaining categories. This makes determining the full set of failed tests a tedious process. Furthermore, even when tests succeed, occasionally (rarely) they leave behind orphaned containers, which can mess with the results of the subsequent tests.

Separating them provides better separation, and also better parallelism so that the CI jobs complete more quickly.


.github/workflows/flaky-test-monitor.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       COMMIT_SHA: ${{ github.sha }} 
       RUN_ID: ${{ github.run_id }}
       RESULTS_FILE: test-results
-      TEST_OUTPUT_FILE: test-output
```
Kay-Zee, line 38: any reason we removed this?

bors.toml
```diff
@@ -1,9 +1,52 @@
...
+  "Integration Tests (1.16, integration-verification)",
+  "Integration Tests (1.17, integration-mvp)",
+  "Integration Tests (1.17, integration-ghost)",
+  "Integration Tests (1.17, integration-network)",
+  "Integration Tests (1.17, integration-epochs)",
+  "Integration Tests (1.17, integration-access)",
+  "Integration Tests (1.17, integration-collection)",
+  "Integration Tests (1.17, integration-consensus)",
+  "Integration Tests (1.17, integration-execution)",
+  "Integration Tests (1.17, integration-verification)",
```
Kay-Zee, line 25: this is pretty noisy, it makes me want to move to just a single go version. Thoughts?

Kay-Zee (approved):

.github/workflows/flaky-test-monitor.yml
smnzhu: The `run-tests.sh` script no longer writes to a file, it just prints to stdout. This way it can be used for both CI and flaky test monitor. For flaky test monitor, we just need to then pipe the output to a file.

Kay-Zee (approved):
LGTM

tarakby (approved):

PR #2243, opened 2022-04-01
[Consensus] Active Pacemaker multi message BLS aggregation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6214

### Context

This PR implements multi message BLS aggregation which is required for building timeout certificate. 
Design is very similar to `WeightedSignatureAggregator`, the only difference is that we don't have a lower layer like `SignatureSameMessageAggregator` because it doesn't introduce any real value(remember, we aggregate signature over many distinct messages).


consensus/hotstuff/timeoutcollector/aggregation_test.go
```diff
@@ -0,0 +1,253 @@
+package timeoutcollector
+
+import (
+	"math/rand"
```
tarakby, line 4: It would be better to test with different data at each iteration: we could use `math/rand` and call `Seed(time.Now().UnixNano())` OR just use `crypto/rand` instead.

tarakby (commented):
Clean code 👌🏼 
I left a few comments in the code, and another high-level comment about the design. 

Since the new aggregator structure does not fix the messages upfront, the design of the pacemaker assumes the verifier would receive the messages to verify against at the same time as receiving the aggregated signature. 
In this case, the pacemaker business logic will need to enforce some checking on the messages beyond verifying the cryptographic signature (I believe this is already the case). In particular, the business logic should ensure past messages cannot be re-used in newer rounds (to avoid replay attacks).

consensus/hotstuff/signature.go
```diff
@@ -107,6 +107,43 @@ type WeightedSignatureAggregator interface {
...
+	TotalWeight() uint64
+
+	// Aggregate aggregates the signatures and returns the aggregated signature.
+	// The function performs a final verification and errors if the aggregated
+	// signature is not valid. This is required for the function safety since
+	// `TrustedAdd` allows adding invalid signatures.
+	// Expected errors during normal operations:
+	//  - model.InsufficientSignaturesError if no signatures have been added yet
+	//  - model.InvalidSignatureIncludedError if some signature(s), included via TrustedAdd, are invalid
+	Aggregate() ([]flow.Identifier, []byte, error)
```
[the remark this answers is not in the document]
durkmurder: I didn't include messages, since other component will be building data for reconstructing messages. In other words I don't really need it. If you think we need to add it for general usage we can add it.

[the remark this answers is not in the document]
tarakby: We could:
 - either return `signers map[flow.Identifier][]byte , sig Signature, error`  for generic use cases. 
 - OR only return `sig Signature, error` because the list of signers in an array doesn't add any information about the aggregated signature.
 
 > since other component will be building data for reconstructing messages.
 So when the aggregator node broadcasts/sends this aggregated signature to other nodes, it doesn't need to also mention the list of signers and their messages? 
(FYI, I know nothing about the pace maker design)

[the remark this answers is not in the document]
durkmurder: We have a mixed case, providing some context, we have a timeout certificate, which is similar to quorum certificate but has multiple messages. 

```golang
type TimeoutCertificate struct {
	View      uint64
	SignerIDs []Identifier
	TOHighQCViews []uint64 // highest QC view
	TOHighestQC *flow.QuorumCertificate // highest QC built based on TOHighQCViews
	SigData   []byte // Aggregated signature over multiple messages, 
									 // one message consist of (View, HighestQCView)
}
```

This structure will be filled after aggregation process. 
`SigData` - is the aggregated signature itself
`SignerIDs` - is the list of signers contributed to creating TC
`TOHighQCViews` - is the array of views which is used to create the messages for sig verification(also used for other validity checks)

So when receiving TC to verify `SigData` we will reconstruct messages using `View` and `TOHighQCViews`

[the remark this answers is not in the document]
tarakby: Thanks for the context, I understand better now. 

The module's API itself looks fine, but if we look at it as part of the wider design, there might be a conceptual safety issue. This module collects signatures and messages via `TrustedAdd`, I assume each signature has been already verified using `Verify` (like in the QC case). In order to guarantee that the aggregated signature at the end is safe, we execute an internal extra verification of the aggregated signature. That verification is there as a safety measure to make sure calls to `TrustedAdd` and `Verify` are consistent. The last verification guarantees that the aggregated signature is valid against the **set of messages collected internally** (and not another set of messages). However, the design seems to be collection the messages in another module (the one collecting  `View` and `TOHighQCViews`). 
The extra verification inside `SignatureAggregator` doesn't add much to the validity of the aggregated signature against `TOHighQCViews`.

Still thinking about the best way to have a safe design 🤔

jordanschalm (commented):
Nice work. 

One thing we should discuss is how we want to handle the process of adding a new TO. We might ingest some TO `A` from node `a`, then ingest some other TO `B`, where `B.HighQC.View>A.HighQC.View` and `B.HighQC` resolves to a committee where `a` is ejected. If we optimistically added `A.SigData` to the signature aggregator, then we would need to start over the aggregation.

On the other hand, if we wait to add `A` until we have until TOs, then we're doing a lot of the work of this component in the layer above.

My gut feeling is that optimistically adding TOs, then having a recovery path where we filter previously cached invalid TOs and restart aggregation would be OK, since committee change should be pretty infrequent. But if we don't go down that path it would impact the design of this component.

consensus/hotstuff/signature.go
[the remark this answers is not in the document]
tarakby: Note that the the last sanity verification (different messages) isn't cheap like the one in QC. Some options to address the issue above:

1. run another safety verification of the aggregated signature against a message derived from `View` and `TOHighQCViews`, when creating `TimeoutCertificate`.  [expensive option because the safety verification is run twice]
2. rename `Aggregate` to `UnsafeAggregate` and remove the first safety verification from the module. Keep the second verification when creating `TimeoutCertificate`. 
3. Update the the aggregator to accept `TOHighQCView uint64` instead of `message []byte` in `Verify` and `TrustedAdd`. `Aggregate` would return `signature crypto.Signature, SignerIDs []Identifier, TOHighQCViews []uint64` that are all safe to go into `TimeoutCertificate`.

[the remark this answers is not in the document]
durkmurder: What if we don't cache messages and pass them in `Aggregate` as parameter?
On other side I am tempted to do `UnsafeAggregate` and then validate resulting TC(not only signature but whole TC).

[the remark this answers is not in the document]
tarakby: > What if we don't cache messages and pass them in Aggregate as parameter

that's also possible

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,187 @@
...
+// The constructor errors if:
+// - the list of identities is empty
+// - if the length of keys does not match the length of identities
+// - if one of the keys is not a valid public key.
+//
+// A multi message sig aggregator is used for one aggregation only. A new instance should be used for each
+// signature aggregation task in the protocol.
+func NewWeightedMultiMessageSigAggregator(ids flow.IdentityList, // list of all authorized signers
+	pks []crypto.PublicKey, // list of corresponding public keys used for signature verifications
+	dsTag string, // domain separation tag used by the signature) *WeightedMultiMessageSignatureAggregator
```
jordanschalm, line 55:
```suggestion
	dsTag string, // domain separation tag used by the signature
```

consensus/hotstuff/signature.go
```diff
@@ -107,6 +107,42 @@ type WeightedSignatureAggregator interface {
 	Aggregate() ([]flow.Identifier, []byte, error)
 }
 
+// WeightedMultiMessageSignatureAggregator aggregates signatures of the same signature scheme but
+// possibly distinct message from different signers. Only public keys are agreed upon upfront.
+// It's similar to WeightedSignatureAggregator but supports distinct messages and  messages are not known upfront.
+// It is also recommended to only aggregate signatures generated with keys representing
+// equivalent security-bit level.
```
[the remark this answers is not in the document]
durkmurder: It means that keys generated by different methods shouldn't be mixed together. For instance staking and beacon keys.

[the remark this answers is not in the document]
durkmurder: It might be not relevant here

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,187 @@
...
+// represent the list of all possible signers.
+// The constructor errors if:
+// - the list of identities is empty
+// - if the length of keys does not match the length of identities
+// - if one of the keys is not a valid public key.
+//
+// A multi message sig aggregator is used for one aggregation only. A new instance should be used for each
+// signature aggregation task in the protocol.
+func NewWeightedMultiMessageSigAggregator(ids flow.IdentityList, // list of all authorized signers
+	pks []crypto.PublicKey, // list of corresponding public keys used for signature verifications
```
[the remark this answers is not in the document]
durkmurder: makes sense, agree

consensus/hotstuff/signature.go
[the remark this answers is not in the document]
AlexHentschel: I think it might be helpful here to consider the reasons why we have the `TrustedAdd` in the `WeightedSignatureAggregator` (same message):
* **Vote Aggregation**: same message, but potentially two different signatures from the same voter:
   
   This is the scenario  for which we designed `WeightedSignatureAggregator`. Votes (depending on the voting scheme) could contain a staking _and_ a beacon signatures, which must either _both_ be added or not added. Therefore, we have _two_ aggregators, whose internal state needs to be evolved in synchrony, ideally without too much locking.  
   * Separating `Verify` and `TrustedAdd` allowed us to _safely_ address this case in a performant manner, because we can verify both signatures upfront. In other words, we check that it is safe to add the votes upfront, before we commence any state-changing operations. 
   * As @tarakby explained, the `WeightedSignatureAggregator` does a posterior validation of the signature aggregate _before_ it returns the result. This protects from accidental misuses, where a faulty signature is added via `TrustedAdd`. So `WeightedSignatureAggregator` guarantees correctness for any aggregate it returns (a very important property!). 
     When aggregating signatures over the _same_ message, the posterior validation adds only marginal overhead (essentially 1 extra validation on top of the 90+ validations we anyway do using `Verify`). 
* **Timeout Aggregation**: each signer potentially signs a different message, but each signer only provides a _single_ signature:

  This is the scenario  for which we designing `WeightedMultiMessageSignatureAggregator`. While it would be nice to have consistent API, a safety guarantee is much more important.  

  * Note that in this case, there is only a single struct that carries the relevant state: a single instance of `WeightedMultiMessageSignatureAggregator`. Therefore, we could work with the method
    ```golang
    VerifyAndAdd(signerID flow.Identifier, sig crypto.Signature, msg []byte)
    ```
  * We can design the code such that congestion is not a problem, if we check signature validity internally _before_ we lock the internal state. 
  * With `VerifyAndAdd`, the aggregator can itself assert that all input signatures are valid. Therefore, the aggregator  can guarantee correctness for any aggregate it returns without posterior validation of the result. 
    For this scenario, posterior validation is much more expensive because different messages are signed. 

In summary, I would suggest to favour safety over API consistency

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// WeightedMultiMessageSignatureAggregator implements consensus/hotstuff.WeightedMultiMessageSignatureAggregator.
+// It aggregates BLS signatures for many messages from different signers.
+// Only public keys needs to be agreed upon upfront.
+// Each signer is allowed to sign at most once.
+// Aggregation uses BLS scheme. Mitigation against rogue attacks is done using Proof Of Possession (PoP)
+// This module does not verify PoPs of input public keys, it assumes verification was done outside this module.
+// Implementation is thread-safe.
+type WeightedMultiMessageSignatureAggregator struct {
```
AlexHentschel, line 27: I really commend your effort to implement a reusable component. But as [Tarak has pointed out](https://github.com/onflow/flow-go/pull/2243#discussion_r843390102), this is quite non-trivial in general. For such a component to be universally useful _and safe_, it would need to
* > output should have the information that can be shared with another party and allow that party to verify the aggregated signature.
* It should guarantee that any aggregate it returns is correct (defensive programming). 

If achieving this is possible with reasonable overhead, I am happy. Though, I feel that we have designed a component with a general API, which is easily to misuse, requires explicit _extra_ actions to guarantee result correctness (check of the aggregate must be explicitly added by the engineer), and does not retain the necessary information to validate its result. 

### Suggestion
In addition, note that timeout aggregation is actually a _mixture_, where _all_ replicas sign the _same_ current view but different `QC.Views`. This is important, because we explicitly utilize the knowledge that the current view is the same for all signs for space efficiency. 
In other words, this signature scheme is super specific for timeout aggregation. Therefore, I think it warrants its own special-purpose aggregator. 

I feel we should rename this struct to `TimeoutSignatureAggregator` and include all the timeout-specific details:
* it knows the view
* When we add a timeout, we feed in: `SignerID`, `SigData`, and `HighestQC.View`.
* We add timeouts via `VerifyAndAdd`

The `TimeoutSignatureAggregator` would then:
* guarantee that it always returns a valid aggregate
* along with the aggregate sig return the `TOHighQCViews` slice (which we can directly use in the TC)
* own all the information that is required to verify the produced signature

consensus/hotstuff/signature.go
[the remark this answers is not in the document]
tarakby: Thanks @AlexHentschel for the context.

It looks like the use cases of `WeightedMultiMessageSignatureAggregator` and `WeightedSignatureAggregator` are different. After the context you added, I agree there is no need for `WeightedMultiMessageSignatureAggregator` to separate `Add` and `Verify` and we could have these in a single step `VerifyAndAdd`.  In that case `Aggregate` is a safe function without the need to add a final verification. 

The wider safety issue still remains though, and there is final step of verification needed: the signature returned by `WeightedMultiMessageSignatureAggregator` is safe against the messages collected internally by the module, but is not safe against the `TOHighQCViews`, which are collected by a separate module. 

Some options to fix this are:

1. run another safety verification of the aggregated signature against a message derived from View and `TOHighQCViews`, when creating `TimeoutCertificate`. [run an expensive safety verification].
2. Update the the aggregator to accept `TOHighQCView uint64` instead of `message []byte` in `VerifyAndAdd`. `Aggregate` would return `signature crypto.Signature, SignerIDs []Identifier, TOHighQCViews []uint64` that are all safe to go into TimeoutCertificate.

consensus/hotstuff/timeoutcollector/aggregation.go
tarakby: I think the suggestion here corresponds to the option 2 in my suggestion [here](https://github.com/onflow/flow-go/pull/2243#discussion_r860016309).

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -69,43 +79,33 @@ func NewWeightedMultiMessageSigAggregator(ids flow.IdentityList, // list of all
...
 		idToInfo:      idToInfo,
-		idToSignature: make(map[flow.Identifier]crypto.Signature),
+		idToSignature: make(map[flow.Identifier]sigInfo),
+		view:          view,
 	}, nil
 }
 
-// Verify verifies the signature under the stored public keys.
+// VerifyAndAdd verifies the signature under the stored public keys and adds signature with corresponding
+// highest QC view to the internal set. Internal set and collected weight is modified iff signature _is_ valid.
```
tarakby, line 91: I think the signature being valid isn't enough to update the internal set.
```suggestion
// highest QC view to the internal set. Internal set and collected weight is modified iff the signer ID is not a duplicate and signature _is_ valid.
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -69,43 +79,33 @@ func NewWeightedMultiMessageSigAggregator(ids flow.IdentityList, // list of all
...
+// The total weight of all collected signatures (excluding duplicates) is returned regardless
+// of any returned error.
 // Expected errors during normal operations:
 //  - model.InvalidSignerError if signerID is invalid (not a consensus participant)
+//  - model.DuplicatedSignerError if the signer has been already added
 //  - model.ErrInvalidSignature if signerID is valid but signature is cryptographically invalid
 // The function is thread-safe.
-func (a *TimeoutSignatureAggregator) Verify(signerID flow.Identifier, sig crypto.Signature, msg []byte) error {
+func (a *TimeoutSignatureAggregator) VerifyAndAdd(signerID flow.Identifier, sig crypto.Signature, highestQCView uint64) (totalWeight uint64, exception error) {
```
tarakby, line 99: It is better to verify the signature cryptographically only when the signer ID is not duplicate, because the cryptographic verification is more expensive.

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -158,10 +169,19 @@ func (a *TimeoutSignatureAggregator) UnsafeAggregate() ([]flow.Identifier, []byt
...
 		//  * if some signature(s), included via TrustedAdd, could not be decoded
 		if crypto.IsInvalidInputsError(err) {
-			return nil, nil, model.NewInvalidSignatureIncludedErrorf("signatures with invalid structure were included via TrustedAdd: %w", err)
+			return nil, nil, nil, model.NewInvalidSignatureIncludedErrorf("signatures with invalid structure were included via TrustedAdd: %w", err)
 		}
-		return nil, nil, fmt.Errorf("BLS signature aggregation failed: %w", err)
+		return nil, nil, nil, fmt.Errorf("BLS signature aggregation failed: %w", err)
+	}
+
+	valid, err := crypto.VerifyBLSSignatureManyMessages(pks, aggSignature, messages, hashers)
```
tarakby, line 180: As @AlexHentschel explained here, if the module does not support `TrustedAdd` anymore, the final verification isn't needed. The aggregated signature is guaranteed to be valid because adding and verifying are now always coupled. The purpose of `VerifyAndAdd` is to remove this final expensive check.

durkmurder: I would have the extra verification to be 100% sure that we are dealing with correct sig, maybe we could remove that later ? This gives strong guarantee that produced signature is always valid and node won't be slashed. So API safety as priority

consensus/hotstuff/timeoutcollector/aggregation.go
durkmurder: Makes sense, will change this

durkmurder:
@AlexHentschel @tarakby @jordanschalm I've made an important change to increase data locality. Strictly speaking aggregator has all needed data to build TC except highest QC, originally I was planning to handle highest QC and high QC views in separate "builder" component but since high QC views are moved inside of aggregator it seems to me it's also feasible to move knowledge of highest QC as well. This allows us to have one lock instead of two. As result we have all needed data to build TC as result of `Aggregate`.

consensus/hotstuff/timeoutcollector/aggregation.go
tarakby: In the current implementation (with only `VerifyAndAdd` updating the internal set), there is no risk of the aggregated signature being invalid. API safety checks should only be added if there is a risk the API caller can misuse the available functions. This is not the case here because the aggregated signature is valid by construction. A vague analogy is to write a code that computes `sum(a,b) = a+b` but then add a check `sum(a,b) - a == b` for safety. The check isn't necessary and correctness of the implementation should be maintained by unit tests. 

However, I agree with you that there is nothing in our code that guarantees no one will add a `TrustedAdd` function in the future 😅  I don't think Go can enforce only `VerifyAndAdd` can touch the internal set. If that happens then the extra check becomes required.

We usually don't add checks of code correctness within the implementation itself (that's what tests rather do), however I let you and @AlexHentschel decide if you want to keep the extra check. From a cryptographic point of view, that check is not needed here.

consensus/hotstuff/signature.go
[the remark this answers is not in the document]
AlexHentschel: thanks Tarak for the insights. 

> 2. Update the the aggregator to accept TOHighQCView uint64 instead of message []byte in VerifyAndAdd. Aggregate would return signature crypto.Signature, SignerIDs []Identifier, TOHighQCViews []uint64 that are all safe to go into TimeoutCertificate.

this would be my preference

consensus/hotstuff/timeoutcollector/aggregation.go
AlexHentschel: The final verification of the aggregate is particularly expensive in this case, as the signatures are over different messages. Therefore, my preference would be to skip the posterior check, as it impacts the hot-path in consensus.

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+	sig        crypto.Signature
+	highQCView uint64
+}
+
+// TimeoutSignatureAggregator implements consensus/hotstuff.TimeoutSignatureAggregator.
+// It performs timeout specific BLS aggregation over multiple distinct messages.
+// We perform timeout signature aggregation for some concrete view, because of that our messages
+// are build in next way: hash(view, highestQC.View), where `highestQC` is a value contributed
+// by different replicas.
+// View and public keys needs to be agreed upon upfront.
```
AlexHentschel, line 32:
```suggestion
// View and the identities of all authorized replicas are 
// specified when the TimeoutSignatureAggregator is instantiated.
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+// Each signer is allowed to sign at most once.
+// Aggregation uses BLS scheme. Mitigation against rogue attacks is done using Proof Of Possession (PoP)
+// This module does not verify PoPs of input public keys, it assumes verification was done outside this module.
```
AlexHentschel, line 35: nice. Thanks for explicitly pointing this out in the documentation!

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+	view          uint64                         // view for which we are aggregating signatures
+	highestQC     *flow.QuorumCertificate        // highest QC submitted via VerifyAndAdd
+}
+
+var _ hotstuff.TimeoutSignatureAggregator = (*TimeoutSignatureAggregator)(nil)
+
+// NewTimeoutSignatureAggregator returns a multi message signature aggregator initialized with a predefined view
+// for which we aggregate signatures, list of flow identities,
+// their respective public keys and a domain separation tag. The identities
+// represent the list of all possible signers.
```
AlexHentschel, line 52:
```suggestion
// represent the list of all authorized signers.
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+// The constructor errors if:
+// - the list of identities is empty
+// - if one of the keys is not a valid public key.
+//
+// A multi message sig aggregator is used for one aggregation only. A new instance should be used for each
```
AlexHentschel, line 57:
```suggestion
// A multi message sig aggregator is used for aggregating timeouts for a single view only. A new instance should be used for each
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+	idToInfo := make(map[flow.Identifier]signerInfo)
+	for _, id := range ids {
+		idToInfo[id.NodeID] = signerInfo{
+			pk:     id.StakingPubKey,
+			weight: id.Weight,
+		}
+	}
+
+	return &TimeoutSignatureAggregator{
+		hasher:        crypto.NewBLSKMAC(dsTag),
```
AlexHentschel, line 84:
```suggestion
		hasher:        crypto.NewBLSKMAC(dsTag), // concurrency safe
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+// This module does not verify PoPs of input public keys, it assumes verification was done outside this module.
+// Implementation is thread-safe.
+type TimeoutSignatureAggregator struct {
+	lock          sync.RWMutex
+	hasher        hash.Hasher
+	idToInfo      map[flow.Identifier]signerInfo // auxiliary map to lookup signer weight and public key (only gets updated by constructor)
+	idToSignature map[flow.Identifier]sigInfo    // signatures indexed by the signer ID
+	totalWeight   uint64                         // total accumulated weight
+	view          uint64                         // view for which we are aggregating signatures
+	highestQC     *flow.QuorumCertificate        // highest QC submitted via VerifyAndAdd
```
AlexHentschel, line 44: suggesting to track the newest QC on a higher code level. Technically, we could even include a QC that is newer than `TOHighQCViews`

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+
+	aggSignature, err := crypto.AggregateBLSSignatures(signatures)
+	if err != nil {
+		// invalidInputsError for:
+		//  * empty `signatures` slice, i.e. sharesNum == 0, which we exclude by earlier check
+		//  * if some signature(s), included via TrustedAdd, could not be decoded
+		if crypto.IsInvalidInputsError(err) {
+			return nil, model.NewInvalidSignatureIncludedErrorf("signatures with invalid structure were included via TrustedAdd: %w", err)
+		}
+		return nil, fmt.Errorf("BLS signature aggregation failed: %w", err)
```
AlexHentschel, line 194: agree with [Tarak's comment](https://github.com/onflow/flow-go/pull/2243#discussion_r861228522): 
* As we don't have the `TrustedAdd` method, this is _not_ an expected failure path anymore, but a symptom of an internal bug. Hence, no need to provide a sentinel to the higher logic. 

```suggestion
		// invalidInputsError for:
		//  * empty `signatures` slice, i.e. sharesNum == 0, which we exclude by earlier check
		//  * if some signature(s) could not be decoded, which should be impossible since we check all signatures before adding them
		// Hence, any error here is a symptom of an internal bug
		return nil, fmt.Errorf("unexpected internal error during BLS signature aggregation: %w", err)
```

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	valid, err := crypto.VerifyBLSSignatureManyMessages(pks, aggSignature, messages, hashers)
+	if err != nil {
+		return nil, fmt.Errorf("signature verification failed: %w", err)
+	}
+	// since we add only verified signatures, then we MUST have a valid signature at this point
+	if !valid {
+		return nil, fmt.Errorf("constructed invalid BLS signature")
+	}
```
AlexHentschel, line 204:
```suggestion
```

AlexHentschel (approved):
very nice. Mostly stylistic suggestions.
Great documentation 🏅 👏

consensus/hotstuff/verification/common.go
```diff
@@ -19,3 +19,17 @@ func MakeVoteMessage(view uint64, blockID flow.Identifier) []byte {
...
+// to contribute to Active Pacemaker protocol. Each replica signs with the highest QC view
+// known to that replica.
+func MakeTimeoutMessage(view uint64, highQCView uint64) []byte {
+	msg := flow.MakeID(struct {
+		View       uint64
+		HighQCView uint64
+	}{
+		View:       view,
+		HighQCView: highQCView,
+	})
```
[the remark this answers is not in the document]
durkmurder: @tarakby I have made this optimization, let me know if you see any problems.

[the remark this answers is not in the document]
tarakby: I confirm the new encoding is injective so the optimization is valid.

consensus/hotstuff/timeoutcollector/aggregation.go
```diff
@@ -0,0 +1,185 @@
...
+	highQCViews := make([]uint64, 0, sharesNum)
+	for id, info := range a.idToSignature {
+		signatures = append(signatures, info.sig)
+		signers = append(signers, id)
+		highQCViews = append(highQCViews, info.highQCView)
+	}
+
+	aggSignature, err := crypto.AggregateBLSSignatures(signatures)
+	if err != nil {
+		// invalidInputsError for:
```
tarakby, line 177:
```suggestion
		// unexpected error for:
```

consensus/hotstuff/timeoutcollector/aggregation_test.go
```diff
@@ -0,0 +1,250 @@
...
+	subSet := signersNum / 2
+	expectedWeight := uint64(0)
+	var wg sync.WaitGroup
+	for i, sig := range sigs[subSet:] {
+		wg.Add(1)
+		// test thread safety
+		go func(i int, sig crypto.Signature) {
+			defer wg.Done()
+			index := i + subSet
+			// test TrustedAdd
```
tarakby, line 94:
```suggestion
			// test VerifyAndAdd
```

tarakby (approved):
Clean code and tests! Thanks for your patience with this PR :)

durkmurder:
bors merge

PR #2254, opened 2022-04-05
Memory metering

dsainati1:
Work on https://github.com/dapperlabs/cadence-private-issues/issues/36

For the upcoming Cadence changes (not yet released) to add memory metering, we need the FVM to actually do something with the recorded memory usage. In this PR, the FVM tracks the total used memory so far by multiplying the amount of abstract memory consumed by a weight (not yet computed) to produce a total, and then throws an error if that total exceeds the maximum allowed usage.


dsainati1:
Closing this in favor of a better alternative as discussed with @janezpodhostnik

PR #2256, opened 2022-04-05
Add test for referencing non-existent address

janezpodhostnik:
Below is the call stack from the test:

image

This happens when borrowing the flowTokenReceiver from an account that does not exist.

I don't think the step marked in red in the screenshot should happen.

Fixed by https://github.com/onflow/cadence/pull/1557, but we should still add the new FVM test


turbolent:
Great investigation, I fixed this on the Cadence side in onflow/cadence#1557

turbolent (approved):
Thank you for adding a test for this 👍

ramtinms (approved):

janezpodhostnik:
bors merge

PR #2258, opened 2022-04-05
Upgrade `go-retry` `v0.1.0->v0.2.3`

jordanschalm:
`go-retry` had a bug where backoffs failed to handle integer overflow, resulting in negative time values which undercut the duration cap. This could cause a perpetual tight loop in components using `retry` backoffs, resulting in significant CPU and memory usage and impact on network performance. See [here](https://www.notion.so/dapperlabs/Apr-4-Canary-Consensus-Slowdown-Following-EECC-e4136747a6934f5e906c95bdafe4b318) for more details.

This PR upgrades `go-retry` from `v0.1.0->v0.2.3`. It also adds a test case validating that retries behave as expected, in the component where this problem manifested.


module/epochs/machine_account_test.go
```diff
@@ -178,3 +182,33 @@ func TestMachineAccountChecking(t *testing.T) {
...
+	calls := 0
+	last := time.Now()
+	retry.Do(context.Background(), backoff, func(ctx context.Context) error {
+		calls++
+		fmt.Println(calls, time.Since(last))
+
+		// assert the time between calls is greater than the cap
+		// an overflow would
+		if calls > 1 {
+			assert.Greater(t, time.Since(last), time.Microsecond*9)
```
kc1116, line 204: Should we be checking if time.Since(last) is equal to 10 micro seconds after the first backoff ?

kc1116 (commented):
Nice job and good write up. I have one question regarding the test.

module/epochs/machine_account_test.go
jordanschalm: Since there is jitter in the backoff after the duration cap, it won't be exactly 10µs

kc1116: Gotcha, could we check if time.Since(last) <= capped duration + max possible jitter ?

jordanschalm: Checking against an upper bound could cause a flakey test, since the test will sleep for at least `backoff`, but possibly more. I changed the test to check `backoff.Next` directly so we don't need to worry about inconsistencies of measuring wallclock time in the test.

kc1116: Looks good

zhangchiqing (approved):

jordanschalm:
bors merge

jordanschalm:
bors merge

Issue #1744, opened 2021-12-09
[Execution State] Memory usage and performance optimisation


ramtinms:
This Epic captures some of the improvements to the execution state code (also known as Ledger) in terms of reducing memory and improving the speed of operations and proof verifications.

fxamacker:
Testnet memory reduction results after Atree Inlining & Deduplication:

![image](https://github.com/user-attachments/assets/99222f5a-fec0-45f3-bf38-9ea4c47c98dd)

NOTE:  testnet and mainnet results can differ, depending on usage, data, etc.

See atree epic 292 for more details & charts:
- https://github.com/onflow/atree/issues/292
- https://github.com/onflow/atree/releases/tag/v0.8.0-rc.5

PR #2265, opened 2022-04-06
(WIP) [EN Performance] Split Node into LeafNode and InterimNode

fxamacker:
### Changes

Split Node into LeafNode and InterimNode and remove unnecessary fields to reduce memory:
- 16 bytes for each leaf node
- 24 bytes for each interim node

Create Node interface with read-only functions for LeafNode and InterimNode.

Updates Epic #1744 

### Caveats

- 🚧 WIP - Looking into why benchnet test shows only 4.4GB savings (expected more).
- 🐢 Calling interface methods which cannot be inlined will have a performance cost.
- ⚠️ Need to decide if the memory savings doesn't outweigh the performance cost.


Kay-Zee:
@fxamacker @ramtinms @m4ksio What's the status on this issue? Getting quite stale.

fxamacker:
@Kay-Zee This was paused because it sacrifices EN performance to gain memory (for both reads and updates).  In August Spork, checkpoint v5 memory reductions (especially after PR 3050 got deployed to mainnet before October Spork) made it less important for EN to reduce memory by sacrificing performance.

I can close this for now and it can be reopened if needed.

Thoughts?

cc: @ramtinms

fxamacker:
Confirmed with @ramtinms that this can be closed for now.

PR #2276, opened 2022-04-07
Fix result indexing by block

koko1123:
Adds simple seal looping logic on block finalizing handler
modifies unit tests for the same


m4ksio (approved):
Would be great to add a test showing actual change - ie, how only sealed results are indexed.

smnzhu (approved):

PR #2278, opened 2022-04-07
Memory weight assignment

dsainati1:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/31

This depends on https://github.com/onflow/flow-go-sdk/pull/265 being merged into `flow-go-sdk`, and on updating `flow-go` itself to use the newest version of that and of secure Cadence.

This hardcodes weights into the FVM for memory metering. Ideally these would be defined in a contract, but in the interest of speed we can hard code them here for now. 

This also updates some other parts of the code to match changes to the Cadence API.


fvm/meter/weighted/meter.go
```diff
@@ -187,10 +306,10 @@ func (m *Meter) MemoryIntensities() interfaceMeter.MeteredMemoryIntensities {
 
 // TotalMemoryUsed returns the total memory used
 func (m *Meter) TotalMemoryUsed() uint {
-	return uint(m.memoryUsed >> MeterInternalPrecisionBytes)
```
janezpodhostnik, line 190: I think there is one more of these in the merge function. Also maybe we can rename the `MeterInternalPrecisionBytes` to `MeterExecutionInternalPrecisionBytes` or similar.

fvm/meter/weighted/meter.go
```diff
@@ -23,10 +23,129 @@ var (
...
+		common.MemoryKindNominalType:       0,
+		common.MemoryKindOptionalType:      0,
+		common.MemoryKindReferenceType:     0,
+		common.MemoryKindRestrictedType:    0,
+		common.MemoryKindVariableSizedType: 0,
+	}
 )
 
+func _() {
+	// A compiler error signifies that we have not accounted for all memory kinds
```
janezpodhostnik, line 144: If we are worried about all the memory kinds being present in the map, maybe the `SetMemoryWeights` and the `WithMemoryWeights ` functions should also check for this.

janezpodhostnik (commented):
Left some minor comments, but looks good to me.

robert-e-davidson3 (approved):

fvm/meter/weighted/meter.go
```diff
@@ -114,7 +241,7 @@ func (m *Meter) MergeMeter(child interfaceMeter.Meter, enforceLimits bool) error
 	if basic, ok := child.(*Meter); ok {
 		childMemoryUsed = basic.memoryUsed
 	} else {
-		childMemoryUsed = uint64(child.TotalMemoryUsed()) << MeterInternalPrecisionBytes
+		childMemoryUsed = uint64(child.TotalMemoryUsed()) << MeterExecutionInternalPrecisionBytes
```
janezpodhostnik, line 244: No shifting here, because this is about memory, not computation.

go.mod
```diff
@@ -105,3 +105,7 @@ require (
 )
 
 replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
+
+replace github.com/onflow/cadence => ../cadence
```
janezpodhostnik, line 109: This should be removed before merging.

janezpodhostnik (approved):
Two more minor things. Looks good otherwise.

Would it make sense to add a fvm test that sends a transfer transaction and just checks that the memory intensities are non-0?

fvm/meter/weighted/meter.go
```diff
@@ -75,6 +197,14 @@ func WithComputationWeights(weights ExecutionEffortWeights) WeightedMeterOptions
 
 // WithMemoryWeights sets the weights for the memory intensities
 func WithMemoryWeights(weights ExecutionMemoryWeights) WeightedMeterOptions {
+	if int(common.MemoryKindLast)-len(weights) != 1 {
+		return func(m *Meter) {
+			for kind, weight := range weights {
+				m.memoryWeights[kind] = weight
```
robert-e-davidson3, line 203: I think this will overwrite `DefaultMemoryWeights`, making this unexpectedly destructive. I don't think that will ever matter in production but might mess up tests.

fvm/meter/weighted/meter.go
```diff
@@ -75,6 +197,14 @@ func WithComputationWeights(weights ExecutionEffortWeights) WeightedMeterOptions
...
+			}
+		}
+	}
+
 	return func(m *Meter) {
 		m.memoryWeights = weights
 	}
```
robert-e-davidson3, line 210: I found this confusing despite knowing what was going on. Might just be be but I think this is clearer:

```suggestion
	// Replace all weights
	if int(common.MemoryKindLast)-len(weights) == 1 {
		return func(m *Meter) {
			m.memoryWeights = weights
		}
	}

	// Replace just some weights
	return func(m *Meter) {
		for kind, weight := range weights {
			m.memoryWeights[kind] = weight
		}
	}	
```

fvm/meter/weighted/meter.go
```diff
@@ -153,16 +283,21 @@ func (m *Meter) ComputationIntensities() interfaceMeter.MeteredComputationIntens
...
 }
 
 // SetMemoryWeights sets the memory weights
 func (m *Meter) SetMemoryWeights(weights ExecutionMemoryWeights) {
+	if int(common.MemoryKindLast)-len(weights) != 1 {
+		for kind, weight := range weights {
+			m.memoryWeights[kind] = weight
+		}
+	}
 	m.memoryWeights = weights
```
robert-e-davidson3, line 301: This is doing a bit of work then throwing it out with `m.memoryWeights = weights`. It also might be overwriting `DefaultMemoryWeights`.

dsainati1: Right yeah, this is meant to be an else

fvm/meter/weighted/meter.go
```diff
@@ -75,6 +196,16 @@ func WithComputationWeights(weights ExecutionEffortWeights) WeightedMeterOptions
 
 // WithMemoryWeights sets the weights for the memory intensities
 func WithMemoryWeights(weights ExecutionMemoryWeights) WeightedMeterOptions {
+	// Replace some weights
+	if int(common.MemoryKindLast)-len(weights) != 1 {
```
SupunS, line 200: Might be simpler to compare with `DefaultMemoryWeights`?
i.e:
```suggestion
	if len(weights) != len(DefaultMemoryWeights) {
```

fvm/meter/weighted/meter.go
```diff
@@ -153,17 +284,23 @@ func (m *Meter) ComputationIntensities() interfaceMeter.MeteredComputationIntens
...
 // TotalComputationLimit returns the total computation limit
 func (m *Meter) TotalComputationLimit() uint {
-	return uint(m.computationLimit >> MeterInternalPrecisionBytes)
+	return uint(m.computationLimit >> MeterExecutionInternalPrecisionBytes)
 }
 
 // SetMemoryWeights sets the memory weights
 func (m *Meter) SetMemoryWeights(weights ExecutionMemoryWeights) {
-	m.memoryWeights = weights
+	if int(common.MemoryKindLast)-len(weights) != 1 {
```
SupunS, line 297: same here, can compare with `DefaultMemoryWeights`

SupunS (approved):
Looks fine to me. Just a one suggestion/question

PR #2280, opened 2022-04-08
Drop checkpoint files from OS file cache to prevent file cache growing to hundreds of GB

fxamacker:
### Changes

After reading or creating a checkpoint file, drop it from the OS file cache.

Without this change, checkpointing will add over 100GB to OS file cache each time because each checkpoint file is over 50GB.

This change helps prevent Grafana EN Memory Usage chart from showing hundreds of GB as used because it doesn't exclude the OS file cache.

Huge thanks to @zhangchiqing 👍  for trying the `/bin/dd` idea on testnet to confirm this approach works and that the drop in file cache appears on Grafana EN Memory Usage chart.

### Test Results

On benchnet using Debian 10, creating a checkpoint file showed:
- first few minutes: OS file cache climbs to old checkpoint file size and immediately drops to under 2 GB.
- next few minutes: OS file cache remained under 2 GB while replaying WAL segments.
- last few minutes: OS file cache climbs to newly created checkpoint file size and immediately drops to under 2 GB.

The WAL segments are included in the 2 GB OS file cache because the tests were repeated.

### Other Approaches

Given the simplicity of this change, I decided not to add layer of indirection (run python script instead of /bin/dd directly) or cron job or systemd service to handle this and etc.

### Caveats

This solution uses `/bin/dd` or `/usr/bin/dd` if GOOS is linux.  On other OS and if `dd` isn't found in /bin or /usr/bin, it does nothing.

For now, we don't bother dropping WAL files from OS file cache to keep this change very simple.


ramtinms (approved):

ledger/complete/wal/syncrename.go
```diff
@@ -60,6 +60,11 @@ func (s *SyncOnCloseRenameFile) Close() error {
 		return fmt.Errorf("error while renaming from %s to %s: %w", s.file.Name(), s.targetName, err)
 	}
 
+	err = requestDropFromOSFileCache(s.targetName)
```
zhangchiqing, line 63: Why doing it here? Is it when the checkpoint is generated? In order to drop the checkpoint from memory right after the checkpoint is generated?

ledger/complete/wal/checkpointer.go
```diff
@@ -769,3 +773,43 @@ func readCheckpointV5(f *os.File) ([]*trie.MTrie, error) {
...
+	s = strings.ReplaceAll(s, ">", "")
+	s = strings.ReplaceAll(s, "<", "")
+	s = strings.ReplaceAll(s, "*", "")
+
+	_, err = os.Stat(s)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("sanitized filename %s does not exist", s)
+	}
+
+	cmd := exec.Command(cmdFileName, "if="+s, "iflag=nocache", "count=0")
```
zhangchiqing, line 813: Better to INFO log this operation

ledger/complete/wal/checkpointer.go
```diff
@@ -769,3 +773,43 @@ func readCheckpointV5(f *os.File) ([]*trie.MTrie, error) {
 
 	return tries, nil
 }
+
+// requestDropFromOSFileCache requests the specified file
+// be dropped from OS file cache.
+// CAUTION: Returns nil without doing anything if GOOS != linux.
```
zhangchiqing, line 779:
```suggestion
// The use case is that when a new checkpoint is generated, OS file cache will still cache the entire checkpoint file in memory, causing the memory usage to stay high. Dropping it could free up the memory.
// CAUTION: Returns nil without doing anything if GOOS != linux.
```

zhangchiqing (approved):
Nice work 💯

ledger/complete/wal/syncrename.go
fxamacker: > Is it when the checkpoint is generated? In order to drop the checkpoint from memory right after the checkpoint is generated?

Yes.  This is the `Close()` of the custom checkpoint file writer.  The caller doesn't have access to checkpoint filename, so this is a convenient and safe spot.

ledger/complete/wal/checkpointer.go
fxamacker: Great suggestion! 👍   Added logging in 3bf1320f82fe6abaff0aeac682c213970e1fdc7b.

Unfortunately, adding it required more changes than anticipated because the caller didn't have access to a logger.

zhangchiqing: I see. Thanks for adding it. Actually it's not required to pass the logger all the way down. To me, it's also OK to simply [use global logger like this](https://github.com/onflow/flow-go/blob/petera/execution-data-requester/cmd/bootstrap/cmd/check_machine_account.go#L91)

fxamacker:
I updated Other Approaches section to note possibility of dropping segments of file and also `O_DIRECT`.

If needed, we can revisit finer-grained solution (dropping segments of file periodically instead of waiting until the end to drop all) after observing this simpler/safer PR on mainnet.

PR #2287, opened 2022-04-11
Use block ID from seal

Kay-Zee:
index against the block id in the seal, rather than the current block ID


smnzhu (approved):

vishalchangrani:
bors seem to have incorrectly closed the PR, re-opening it.

PR #2291, opened 2022-04-11
[BFT Testing] implements wintermute attack orchestrator

yhassanzadeh13:
Implements Wintermute attack orchestrator. The orchestrator takes control of a subset of corrupted execution and verification nodes, and performs the following attack logic:
- Orchestrator corrupts the result of the first incoming execution receipt from any of the corrupted execution node.
- The orchestrator sends the corrupted execution result to all corrupted execution nodes.
- If Orchestrator receives any chunk data pack request for a corrupted chunk from a corrupted verification node, it replies it with attestation for that chunk.
- If Orchestrator receives any chunk data pack response from a corrupted execution node to an honest verification node, it drops the response if it is for one of the corrupted chunks.
- Any other incoming messages to the orchestrator are passed through, i.e., are sent as they are in the original Flow network without any tampering.


insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,244 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
 // In Corruptible Conduit Framework for BFT testing, corrupted nodes relay their outgoing events to
 // the attacker instead of dispatching them to the network.
 func (o *Orchestrator) HandleEventFromCorruptedNode(event *insecure.Event) error {
+	o.Lock()
+	defer o.Unlock()
+
+	switch protocolEvent := event.FlowProtocolEvent.(type) {
+
+	case *flow.ExecutionReceipt:
```
gomisha, line 64: can we add this comment to elaborate this case: 
`// orchestrator received execution receipt from EN after EN executed a block`

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,244 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
...
+		if err := o.handleExecutionReceiptEvent(event); err != nil {
+			return fmt.Errorf("could not handle execution receipt event: %w", err)
+		}
+	case *messages.ChunkDataRequest:
```
gomisha, line 68: can we add this comment to elaborate this case:
`// orchestrator received chunk data request from VN after VN assigned a chunk, and VN now requesting that chunk to verify it`

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,244 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
...
+		if err := o.handleChunkDataPackRequestEvent(event); err != nil {
+			return fmt.Errorf("could not handle chunk data pack request event: %w", err)
+		}
+	case *messages.ChunkDataResponse:
```
gomisha, line 72: can we add this comment to elaborate this case:
`// orchestrator received chunk data response from EN when EN wants to respond to a chunk data request from VN`

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,244 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
...
+		Str("channel", string(receiptEvent.Channel)).
+		Uint32("targets_num", receiptEvent.TargetNum).
+		Str("target_ids", fmt.Sprintf("%v", receiptEvent.TargetIds)).Logger()
+
+	if o.state != nil {
+		// non-nil state means an execution result has already been corrupted.
+		if receipt.ExecutionResult.ID() == o.state.originalResult.ID() {
+			// receipt contains the original result that has been corrupted.
+			// corrupted result must have already been sent to this node, so
+			// just discards it.
```
gomisha, line 133: nit grammar: `just discard it.`

yhassanzadeh13: https://github.com/onflow/flow-go/pull/2291/commits/f0555a9a48ce8f31c919b1e4ce76cea49a48f013

insecure/wintermute/attackOrchestrator.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2291/commits/e752e4fd7d551717520bec62d1d135e957815633

insecure/wintermute/attackOrchestrator.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2291/commits/303245f3b9ac6f6e72e75f8b3ee7f6fa6e0cbdb7

AlexHentschel:
I am not sure, though I feel there might be a bit of an imprecision in the terminology here:
* Here, you define that `CorruptedId` is the _Identifier of a conduit_:
   https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/event.go#L20 
   * Presumably, a single node generally has multiple conduits (one conduit per channel?). So for a single node ID, there should be multiple `CorruptedId`
* However, I feel there is a conflicting definition in other places in the code:
   https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L32
   Here, we specifically have the Identifiers of the corrupted nodes, and then ask, whether the ID of a particular conduit is part of this list:
   https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L103

In short, I feel the code (or documentation?) is conflating conduits and nodes Identifiers.

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,259 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
...
+		sent, err := o.replyWithAttestation(chunkDataPackRequestEvent)
+		if err != nil {
+			return fmt.Errorf("could not reply with attestation: %w", err)
+		}
+		if sent {
+			return nil
+		}
+	}
+
+	// no result corruption yet conducted, hence bouncing back the chunk data request.
```
AlexHentschel, line 216: I am not sure that this statement is accurate. I think we can also reach this code _after_ some results already have been corrupted:
* the corrupted execution nodes will only corrupt a single result and then move back to honest behaviour 
* So lets look at the honest results that they emit after they corrupted their result:
  *  for the subsequent honest results, this check will be `false`: https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L275-L276 Hence, we will _not_ return here: https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L211-L213
  *  instead, we continue to https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L216-L217 despite some result already being corrupted.

In my opinion, the implementation does what we desire. Just that the documentation needs to to be revised:
```suggestion
	// chunk data request is for an honest result, hence the corrupted verifier can follow the protocol and send its honest chunk data request 
```

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,259 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
...
+				// Illegal state! chunk data pack response for a CORRUPTED chunk to a CORRUPTED verification node.
+				// Request must have never been reached to corrupted execution node, and must have been replaced with
+				// an attestation.
+				lg.Fatal().
+					Msg("orchestrator received a chunk data response for corrupted chunk to a corrupted verification node")
+			}
+		}
+	}
+
+	// no result corruption yet conducted, hence bouncing back the chunk data request.
```
AlexHentschel, line 262: same here: I think we can reach this code even after results were corrupted:


In my opinion, the implementation does what we desire. Just that the documentation needs to to be revised:
```suggestion
	// chunk data response is for an honest result, hence the corrupted execution node can follow the protocol and send its honest response 
```

AlexHentschel (approved):
Looks good. I like the logic. Its very cool to see that we can encode the entire attack logic for Wintermute in the relatively compact [`wintermute/attackOrchestrator.go`](https://github.com/onflow/flow-go/blob/8c57979c888feef7e0beda782419b3723f74ee36/insecure/wintermute/attackOrchestrator.go#L88) 👏 

Most of my comments are w.r.t. documentation. While the flow is coarsely described, I feel we gloss over some important details thereby introducing imprecision and ambiguity that might mislead engineers trying to understand the code. I tried to give detailed explanations in my comments where I think the lack of precision might be misleading. 
Overall, very nice work. Thanks.

insecure/wintermute/attackOrchestrator.go
```diff
@@ -1,30 +1,52 @@
...
 	"github.com/onflow/flow-go/insecure"
 	"github.com/onflow/flow-go/model/flow"
 )
 
 // Orchestrator encapsulates a stateful implementation of wintermute attack orchestrator logic.
+// The attack logic works as follows:
+// 1. Orchestrator corrupts the result of the first incoming execution receipt from any of the corrupted execution node.
+// 2. Orchestrator sends the corrupted execution result to all corrupted execution nodes.
+// 3. If Orchestrator receives any chunk data pack request for a corrupted chunk from a corrupted verification node, it replies it with an attestation
+//    for that chunk.
```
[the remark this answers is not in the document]
gomisha: Updated to be more clear that the orchestrator prompts corrupted verifier to approve corrupted chunk.
https://github.com/onflow/flow-go/pull/2291/commits/f63f2a7e03ae5f884683c09c6a2dfe0fef00bce4

insecure/wintermute/attackOrchestrator.go
```diff
@@ -1,30 +1,52 @@
...
+// 4. If Orchestrator receives any chunk data pack response from a corrupted execution node to an honest verification node, it drops the response
+//    if it is for one of the corrupted chunks.
```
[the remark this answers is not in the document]
gomisha: The orchestrator doesn't instruct the corrupted execution node to drop the response. Instead, the orchestrator doesn't send anything back to the execution node to send to the VN, thereby indirectly causing the execution node to wintermute the honest verification node.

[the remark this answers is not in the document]
gomisha: Updated comments to make this more clear
https://github.com/onflow/flow-go/pull/2291/commits/4691894e610fb97f0236eae048ce14c9ae4290a7

insecure/wintermute/attackOrchestrator.go
gomisha: updated 
https://github.com/onflow/flow-go/pull/2291/commits/7b8117d079e05b99b72e0ca4fe0dfe2e0b53c43c

insecure/wintermute/attackOrchestrator.go
gomisha: Nice catch! Updated docs to reflect this clarification (couldn't commit suggestion to outdated code)

https://github.com/onflow/flow-go/pull/2291/commits/40594f7566d507a2b082b79e93de638c9fae9b7b

insecure/wintermute/attackOrchestrator.go
```diff
@@ -34,7 +56,259 @@ func (o *Orchestrator) WithAttackNetwork(network insecure.AttackNetwork) {
 // In Corruptible Conduit Framework for BFT testing, corrupted nodes relay their outgoing events to
 // the attacker instead of dispatching them to the network.
```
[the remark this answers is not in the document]
gomisha: updated with suggestion

gomisha (approved):
Addressed all PR review comments

PR #2292, opened 2022-04-12
Execution data storage updates

smnzhu:
~~Split execution data service into Adder and Getter, and implement Storage to track the status of Execution Data requests and do bookkeeping of Execution Data blobs.~~

This version of the requester supports processing both Execution Receipts as well as sealed Results simultaneously, thus allowing us to optimistically download Execution Data before it is sealed.

The requester consists of four main components:
* Dispatcher: This processes Receipts received from the ingestion engine and sealed Execution Results delivered by Finalization Distributor, and dispatches jobs to download the Execution Data associated with them. 
* Handler: This handles downloading the Execution Data from the network for the jobs submitted by the Dispatcher. It is equipped with appropriate retry logic depending on the type of error encountered while executing a job.
* Fulfiller: This processes sealed heights received from the Dispatcher and job results received from the Handler, and updates the persisted state / fires notifications for each newly fulfilled height. 
* Notifier: This receives notifications submitted by the Fulfiller, and sends them to all notification subscribers.

The storage module tracks all necessary metadata regarding the progress of the execution data requester / provider. In particular, it tracks the latest fulfilled height, the set of blob Cids per height, and latest height containing each blob Cid. It provides `Prune` and `Update` methods to ensure that pruning and updating never happens concurrently.

TODO:
- [ ] Add admin command to update pruning config
- [ ] Design doc
- [ ] Integration tests
- [x] Write a basic test to check that Prune actually deletes everything we expect
- [ ] Test handler edge cases (duplicate CID's in the same level, cid not found in the blobservice, etc)


module/state_synchronization/storage.go
```diff
@@ -0,0 +1,957 @@
...
+	} else if (errors.Is(err1, badger.ErrKeyNotFound) && err2 == nil) || (err1 == nil && errors.Is(err2, badger.ErrKeyNotFound)) {
+		return fmt.Errorf("db is in inconsistent state")
+	} else {
+		return fmt.Errorf("failed to initialize storage: %w", err1)
+	}
+
+	return nil
+}
+
+func (s *Storage) replay(storedDataLowerBound uint64, latestIncorporatedHeight uint64) error {
```
zhangchiqing, line 586: Is there any assumption that storedDataLowerBound must be lower than latestIncorporatedHeight? Do we need any sanity check here?

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,957 @@
...
+
+// Prune removes all data from storage corresponding to block heights lower than or equal to the given height.
+// This includes all blob tree records, blob records, and blob data.
+// This should only be called with height <= latest incorporated height.
+func (s *Storage) Prune(height uint64) ([]cid.Cid, error) {
+	storedDataLowerBoundKey := makeGlobalStateKey(globalStateStoredDataLowerBound)
+	storedDataLowerBoundValue := make([]byte, 8)
+	binary.LittleEndian.PutUint64(storedDataLowerBoundValue, height)
+
+	if err := s.db.Update(func(txn *badger.Txn) error {
```
zhangchiqing, line 680: We are updating the database before actually deleting pruning (deleting Cids). If the node crashes right after updating the database, how can we recover from the inconsistency between database and data store?

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,957 @@
...
+		return 0, err
+	}
+
+	return storedDataLowerBound, nil
+}
+
+// Prune removes all data from storage corresponding to block heights lower than or equal to the given height.
+// This includes all blob tree records, blob records, and blob data.
+// This should only be called with height <= latest incorporated height.
+func (s *Storage) Prune(height uint64) ([]cid.Cid, error) {
```
zhangchiqing, line 675: ### Abstraction
I think this storage module lacks of levels of abstraction. The exposed implementation details makes the main logic not that easy to understand.

We need provide abstraction for database operation, so that the logic here could work with higher level abstraction. 

As an example. 

I think the logic could be more clear with higher level abstractions: (pseudo code)
```
func (s) PruneUpToHeight(height) ([]cid, error) {
  pruned := s.executionDataSyncTracker.GetPrunedHeight(s.db)
  
  if height <= pruned {
    // preventing the pruned height go backwards
    return nil
  }
  
  for h := pruned + 1; h <= height; h++ {
    err := s.pruneByHeight(height)
  }

  // only update after the operation is done, in case it crashed before updating db,
  // we will prune again after restart, and the pruning operation is a no-op if has done before.
  s.executionDataSyncTracker.SetPrunedHeight(s.db, height)
  return nil
}
```

The encoding from uint64 to bytes should only stay in the low level DB operation implementation.

The low level database operation should provide abstraction for getting and setting key value pairs with the most convenient data type for the upper level application, where as the low level deals with encoding.

### Data Type
We are storing multiple simple data type under different keys with one transaction. I wonder if makes sense to construct a structured object and [encoding it with msgpack](https://github.com/onflow/flow-go/blob/master/storage/badger/operation/common.go#L73) and store it under a single key.

zhangchiqing (commented):
Haven't finish reviewing yet. But I'd like to provide some early feedbacks, and hear your thoughts on them. Especially about the storage abstraction.

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,957 @@
...
+
+		return nil
+	}); err != nil {
+		return nil, nil, err
+	}
+
+	return cidsToDelete, blockHeights, nil
+}
+
+func (s *Storage) init(startHeight uint64) error {
```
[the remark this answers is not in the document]
smnzhu: Sure, I'll add some logging inside here

module/state_synchronization/cid_comparator.go
```diff
@@ -0,0 +1,50 @@
...
+	cid         cid.Cid
+	isRoot      bool
+	chunkIndex  uint
+	height      uint
+	blockID     flow.Identifier
+	blockHeight uint64
+}
+
+// compareTo returns a negative integer, zero, or a positive integer as
+// a has higher, equal, or lower priority than b
```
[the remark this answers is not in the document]
smnzhu: right, I think that's right

module/state_synchronization/cid_comparator.go
```diff
@@ -0,0 +1,50 @@
...
+			}
+
+			return int(b.height - a.height)
+		}
+
+		// nodes in different blob trees at the same block height have equal priority
+		return 0
+	} else {
+		return -1
+	}
```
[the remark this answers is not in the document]
smnzhu: I've actually removed this for now, because I'm considering just removing task prioritization entirely and adding it back later if we actually see the need for it.

module/state_synchronization/cid_comparator.go
```diff
@@ -0,0 +1,50 @@
...
+				if b.isRoot {
+					return 0
+				} else {
+					return -1
+				}
+			} else if b.isRoot {
+				return 1
+			}
+
+			return int(b.height - a.height)
```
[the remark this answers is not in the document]
smnzhu: height is the height of the blob in the blob tree, not the block height. Deeper nodes have higher priority, which means conversely that higher blobs have lower priority.

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,957 @@
...
+	} else {
+		return int(maxItemCountByWriteSize)
+	}
+}
+
+type ExecutionDataRecord struct {
+	BlockID     flow.Identifier
+	BlockHeight uint64
+	RootID      flow.Identifier
+	Completed   bool
```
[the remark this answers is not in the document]
smnzhu: [3b8f81a](https://github.com/onflow/flow-go/pull/2292/commits/3b8f81a26e5e6d8c48b7aff16d5bd5c089b0442b)

module/state_synchronization/storage.go
smnzhu: It definitely should be less than or equal to the latestIncorporatedHeight. This is checked by the `Check` function. But I think you're right that we should check this here as well, since this is called during the actual creation of the storage.

module/state_synchronization/storage.go
smnzhu: Yes, this is supposed to be handled by `replay`. Can you check that this is handled properly?

module/state_synchronization/storage.go
smnzhu: Thanks, I agree with this, but the storage module is actually intended to be a low level API. It is only used by the Adder / Getter, and all other code should only use the Adder and Getter. Does this make sense?

module/state_synchronization/getter.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if edRoot.BlockID != blockID {
+		return nil, &MismatchedBlockIDError{blockID, edRoot.BlockID}
+	}
+
+	return edRoot, nil
+}
+
+func (s *executionDataGetterImpl) GetChunkExecutionDatas(
```
zhangchiqing, line 227: Would this method being called multiple times?

Is this method a stateless method in a sense if called the second time, it won't "remember" the result it returned from the first time?

module/state_synchronization/getter.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	g, gCtx := errgroup.WithContext(ctx)
+
+	chunkExecutionDatas := make([]*ChunkExecutionData, len(edRoot.ChunkExecutionDataIDs))
+	for i, chunkID := range edRoot.ChunkExecutionDataIDs {
+		i := i
+		chunkID := chunkID
+
+		g.Go(func() error {
```
zhangchiqing, line 258: Can we explain the download strategy here?

From what I've understood. If a result has multiple chunks, the CIDs between chunks are concurrently downloaded. 

And for a single CID from a chunk, it will download level by level, until reaching `maxBlobTreeDepth`. For each level, the CIDs within the same level are downloaded _sequentially_. And we won't download CIDs from the next level, until CIDs from the current level are all downloaded. 

In other words, we only concurrently download CIDs for chunks, but within a single chunk, CIDs are downloaded sequentially. Do we consider downloading them concurrently as well?

smnzhu: The CIDs within the same level are downloaded in parallel. This happens under the hood by Bitswap. It is not possible do download different levels in parallel, because the date from the previous level tells us what we need to download in the next level.

module/state_synchronization/getter.go
smnzhu: It is not supposed to be called multiple times, but if it does get called again, it will load the data it already has on disk.

The way `BlobService` works is that it wraps the logic of checking the local blobstore before making a bitswap request into an interface, which you can call to transparently fetch needed blobs. So under the hood, when you call `BlobService.GetBlobs`, what actually happens is it first checks the blobstore if it already has any of those blobs, and if so, it sends those back immediately. It then sends bitswap requests for any ones it still needs.

So in the case `GetChunkExecutionDatas` gets called twice, the second time all of the data will be loaded from the local blobstore.

module/state_synchronization/storage.go
smnzhu: > We are storing multiple simple data type under different keys with one transaction. I wonder if makes sense to construct a structured object and [encoding it with msgpack](https://github.com/onflow/flow-go/blob/master/storage/badger/operation/common.go#L73) and store it under a single key.

I would have liked to do this for blob records and blob data, but the blob data is currently being populated by the blockservice third party library, and it would be too much work to mess with the implementation of that.

module/state_synchronization/getter.go
zhangchiqing: Let's say the rootCID maps to 3 CIDs, so the first level has 3 CID: C1, C2, C3.

C1 has 2 CIDs: CA and CB.
C2 has 2 CIDs: CM and CN.
C3 has 2 CIDs: CX and CY.

so level 2 has 6 CIDs: [CA, CB, CM, CN, CX, CY].

If the getter has downloaded for C1, and learned C1 has two children: CA and CB, can we parallelize downloading for CA and CB while downloading C2 and C3? Why downloading CA from level 2 is blocked by downloading C2 in level 1?

smnzhu: Ah, that's not quite how it works. What we do is we concatenate all the binary data we get from C1, C2, and C3, and then deserialize the entire piece of data into either another list of CID's, or an Execution Data.

So, to say that "C1 has 2 CIDs: CA and CB" is not accurate, because each CID on its own doesn't represent a self-contained piece of data. You need to concatenate all the data from an entire level in order to deserialize something that makes sense.

smnzhu: I think what you have described might be conceptually cleaner, but it is somewhat more complex to implement, because now we need to handle cases such as, what to do if the subtree rooted at C1 has a different depth than the one rooted at C2? We should not allow this to happen, because otherwise it allows the structure of an Execution Data tree to be non-deterministic.

smnzhu: In practice, the tree for a single ChunkExecutionData will be no more than 2 levels deep, because this already allows for ~32GB of data in total

module/state_synchronization/getter.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to deserialize execution data root blob: %w", err)
+	}
+
+	edRoot, ok := v.(*ExecutionDataRoot)
+	if !ok {
+		return nil, fmt.Errorf("execution data root blob is not an ExecutionDataRoot")
+	}
+
+	if edRoot.BlockID != blockID {
```
zhangchiqing, line 220: I'm not sure how useful is this validation, which was the main purpose of storing it along with the chunk CIDs into the `ExecutionDataRoot`.

Because, a malicious node could also create a `ExecutionDataRoot` with good blockID and bad CIDs. For example,

An honest node uses adder to add `(blockID1, [C1, C2, C3])`, which returns `rootCIDGood`
A malicious node uses adder to add `(blockID1, [C1, C3, C4])`, which returns `rootCIDBad`. 
Now if a follower node uses getter to get `(blockID1, rootCIDBad)`, it will receive `(blockID1, [C1, C3, C4])`. It will not consider it as invalid, beacuse it is able to get chunks from the bad root CID, and the blockID also matches. 

You might say the follower node will get the blockID and rootCID from the sealed result, so it should have good (blockID, rootCID) pair to begin with, but in this case, a good rootCID will always return `(goodBlockID, [C1, C2, C3])`, so it's impossible for blockID to be different. 

Therefore, I think it's not even necessary to put BlockID into `ExecutionDataRoot`. `ExecutionDataRoot` should just be `[]CID`. It's equally secure as containing BlockID

smnzhu: In theory that should be correct, but I think the reason I added the BlockID is just to make sure that every Execution Data Root ID is definitely unique. To be fair, I can't think of any specific scenario that would cause them to not already be unique, even without it.

Another reason is that, somewhere down the line, we may decide to optimistically download Execution Data for blocks before they are even sealed. Then, the Block ID in the Execution Data just serves as a sanity check.

smnzhu: It's probably not strictly necessary, but I don't see a super compelling reason not to keep it, and it might make certain things simpler. For example, we can use it to easily know what block an Execution Data corresponds to.

zhangchiqing: OK, in that case, this `MismatchedBlockIDError` should be a fatal error probably a software bug. 

The current error type sounds like as if it's originated from a malicious node, which lead me thought through the above reasoning. 

I'm OK to keep this sentinel error, but better to explain in comments that the caller should ensure the blockID and rootCID are valid ones from sealed results. And `MismatchedBlockIDError` is a fatal error.

module/state_synchronization/getter.go
zhangchiqing: Now a different question:

I had the impression that `Prune` is only for Getter, because BlobService would store too many data. But since BlobService is used by both Adder and Getter to add and get data by CID, the Adder would also have the same problem as Getter for storing too many data. So Adder will also need Prune as well, right?

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,996 @@
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return errors, nil
+}
+
+// Check checks the integrity of the storage and returns any inconsistencies that were found.
+func (s *Storage) Check(options CheckOptions) error {
```
zhangchiqing, line 438: Can you explain why we need to check the storage? 

I understand it's for integrity. But I wonder what could cause the inconsistency. 

The database (the storage layer) is supposed to guarantee data integrity as long as its public method is used to store data. 

Can you provide an example that could possibly cause inconsistency?

We could simply a lot here if the storage could provide consistency guarantee.

module/state_synchronization/execution_data_service.go
```diff
@@ -2,430 +2,44 @@ package state_synchronization
...
+// ExecutionDataAdder handles adding execution data to a blobservice
+type ExecutionDataAdder interface {
+	// AddChunkExecutionData constructs a blob tree for the given ChunkExecutionData
+	// and adds it to the blobservice, and then returns the root CID.
+	AddChunkExecutionData(
+		ctx context.Context,
+		blockID flow.Identifier,
+		chunkIndex int,
+		ced *ChunkExecutionData,
+	) (cid.Cid, error)
```
zhangchiqing, line 21: I'm not sure whether this method is necessary to be public. This might cause mistake from the caller who added 1 chunk, but didn't add the rest chunks. 

Instead we should just provide one method to add all chunks. So that there are only two results after the call, either nothing was added, or all chunks have been added.

module/state_synchronization/adder.go
```diff
@@ -0,0 +1,285 @@
...
+type executionDataAdderImpl struct {
+	logger        zerolog.Logger
+	metrics       module.ExecutionDataAdderMetrics
+	maxBlobSize   int
+	blobBatchSize int
+	serializer    *serializer
+	blobService   network.BlobService
+}
+
+func (s *executionDataAdderImpl) AddExecutionDataRoot(
```
zhangchiqing, line 71: I'm proposing a refactor of the `Getter` and `Adder` implementation. This is the pseudo code.

The main benefit of the refactor is to provide high conhesion abstractions, so that the function body of `AddExecutionDataRoot` and `GetChunkExecutionDatas` is simplified and self-explains the main logic. 
```golang
func Chunkify(ComputationResult) []*ChunkExectuionData

type ChunkifiedExecutionData struct {
  BlockID flow.Identifier
  Chunks []*ChunkExecutionData
}

type ExecutionDataRoot struct {
  BlockID flow.Identifier
  ChunkCIDs []cid.CID
}

type ChunkExecutionData struct { 	Collection *flow.Collection
	Events     flow.EventsList
	TrieUpdate *ledger.TrieUpdate
}

// AddExecutionData adds the given list of chunk execution data to blob service, and return the root CID of 
// chunksExecutionData is generated from execution.ChunkifyExecutionData(computationResult), which should be a stateless computation
//  from some other package. 
func (adder *ExecutionDataAdder) AddChunkifiedExecutionData(ctx, executionData *ChunkifiedExecutionData) (flow.Identifier, error) {
  // adding ChunkExectuionData could be possibly run concurrently, splitting the concurrency control and 
  // a single job operation into, and using by composing them. 
  chunkCIDs := runConcurrently(executionData.Chunks, func(chunkExecutionData) (cid.CID, error) {
    return adder.addChunkExecutionDataToBlobService(ctx, chunkExecutionData)
  })
  
  // the reason we encode the block ID and chunk CIDs together is so that the blockID could be used
  // by the Getter to verify the rootCID.   rootBlob := adder.codec.Encode(&ExecutionDataRoot{     BlockID: executionData.Chunks,
    ChunkCIDs: chunkCIDs,
  })
  // compute the root CID of all the execution data  chunks
  rootCIDIdentifier := flow.CIDToID(rootBlob.CID())   return rootCIDIdentifier, nil } 
// addChunkExecutionData adds a single chunkExectuionData to the blobService, and return the Cid 
// for the given chunkExecutionData func (*ExecutionDataAdder) addChunkExecutionDataToBlobService(ctx, chunkExecutionData) cid.CID, error

// symmetric Encode and Decode methods to convert between Blob and ExecutionDataRoot
type ExecutionDataRootCodec interface {
	Encode(*ExecutionDataRootCodec) blobs.Blob
       Decode(blob *blobs.Blob) *ExecutionDataRootCodec
}

// GetChunkExecutionDatas gets all the chunk execution data for the given rootCID. The caller must ensure
// the validity of blockID and rootCIDIdentifier, which is supposed to be taken from sealed result to guarantee
// validity. 
// The blockID is taken for sanity check. If `MismatchedBlockIDError` error is returned, the caller should treat it
// as a fatal error. 
func (*Getter) GetChunkExecutionDatas(ctx, blockID, rootCIDIdentifier) ([]*ChunkExecutionData, error) {
  // rootID is a bit confusing as if it was the root hash of the trie.
  // rootCIDIdentifier although is a bit longer but more clear about what it actually is.
  rootCID := flow.IDtoCID(rootCIDIdentifier)

  // an invalid rootCID would fail to retrieve its blob
  blob, err := getter.blobGetter.GetBlob(ctx, rootCID)
  executionDataRoot := getter.codec.Decode(blob)

  // is this check not strictly necessary, but as a sanity check, because the caller is supposed to call with a good (blockID, rootCID) 
  // from a sealed result. Fail this check should be a fatal error, and the caller is responsible to handle this sentinel error properly.
  if blockID != executionDataRoot.BlockID {     return nil, MismatchedBlockIDError
  }
  
  chunks, err := runConcurrently(executionDataRoot.ChunkCIDs, func(cid) (*ChunkExecutionData, error) {
    return getter.getChunkExecutionDataByCID(ctx, cid)
  })

  return chunks, err
}

func (*Getter) getChunkExecutionDataByCID(ctx, cid) (*ChunkExecutionData, error)
```

module/state_synchronization/adder.go
```diff
@@ -0,0 +1,285 @@
...
+	ctx context.Context,
+	blockID flow.Identifier,
+	chunkExecutionDataCIDs []cid.Cid,
```
zhangchiqing, line 74: Taking `chunkExecutionDataCIDs` would assume the caller has called the BlobService to break the chunks execution data into CIDs, but I think this should be part of the Adder's responsibility in order for Adder to guarantee data integrity. Otherwise, the caller might run into mistake that called with wrong Cid slice, which would be hard to debug. 

Refer to my suggested API design in the above comment, where I suggest the Adder should only provide a `AddChunkifiedExecutionData` which takes the ChunkExecutionData, and convert it into slice of CIDs internally.

smnzhu: Originally I was gonna do this, but I think I actually want to incorporate the adding of the individual Chunk Execution Data as part of the block execution pipeline, here: https://github.com/onflow/flow-go/blob/7234bce41a91a71fad6244bcd3d70f5491cc905b/engine/execution/computation/computer/computer.go#L149-L185

What do you think about this?

module/state_synchronization/adder.go
```diff
@@ -0,0 +1,285 @@
...
+				return nil, ctx.Err()
+			}
+
+			// the blob channel is closed, signal to the caller that no more
+			// blobs can be received
+			if i == 0 {
+				return nil, err
+			}
+
+			break
```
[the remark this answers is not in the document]
smnzhu: the only error that we actually expect to see is the one telling us that the blob channel is closed. If this happens in the middle of a batch, then we want to return the batch

module/state_synchronization/getter.go
smnzhu: Potentially, it could use Prune, yes. But for the Adder (execution nodes), it can do something simpler which is just configure the underlying badger db with TTL expiry of all keys.

Pruning is probably the better solution, but using TTL is easier because we can avoid dealing with all of that stuff

module/state_synchronization/adder.go
zhangchiqing: I see. I just have one concerns for that:
Adding the execution data is lower priority compare to finish executing the block. If we integrate it into the block execution pipeline, it might slow down the generation of the execution receipt, which will slow down time to seal. Better let EN focus on block execution, and running on a different thread for adding execution data. But there is a challenge here, the progress of adding execution data might be slower than block execution. If we've executed block 1000, but only added block 200, and then crashed/restarted, how to resume adding the executed blocks? how to ensure that eventually all executed block will be added to state sync service?

module/state_synchronization/storage.go
smnzhu: So in theory, if I've implemented things correctly, there should be no inconsistency :)

But I don't think it's a smart idea to assert this just by analyzing the code, and without actually doing any tests.

This function serves to actually validate that the consistency is maintained.

module/state_synchronization/getter.go
zhangchiqing: Using TTL might get the situation out of our control, especially in a case of running into a cadence bug. By the time, we fixed the bug, and past data might have all been pruned before being added?

TTL is usually used for cache. But our scenario here is not a cache. I think the fine-grained prune logic is necessary to guarantee the network has enough seeder for each block's execution data, right?

module/state_synchronization/storage.go
smnzhu: It is used in all the tests, including the ones which I haven't pushed yet. I think there is no harm making it part of the 
API as well. Nodes can run this check once upon each restart, just as a sanity check.

module/state_synchronization/adder.go
smnzhu: But the point is that the Execution Data ID is part of the Execution Result, so it cannot generate the receipt until it generates this ID anyways

module/state_synchronization/getter.go
smnzhu: Yeah I have same opinion as you theoretically. I just thought we could save some work and do this in a later PR. But now that I think about it more closely, I don't think we really save much work anyways.

To incorporate Pruning into the execution node, we need to prevent any new updates to the db while pruning happens, which means during the pruning it will not be able to store new Execution Data. It will be a bit of work to build in the synchronization mechanisms into the block executor logic to enable this.

module/state_synchronization/storage.go
zhangchiqing: It's not clear to me what needs to be stored in storage. The ExecutionData is stored in BitSwap. To me, we only need to store the height for the highest incorporated execution data, which is just a height. 

So can you explain what needs to be stored in storage, this helps me better understand why they would get inconsistent and why we need to check.

module/state_synchronization/adder.go
zhangchiqing: I see. But I think only the generation of CIDs is needed for generation the ExecutionDataID, the upload of the execution data should not block the block execution right?

smnzhu: Perhaps one way to solve the problem is to redesign the Adder implementation, so that the actual storing of data to the BlobService is done by separate worker routines? So we could just generate and return the CIDs immediately, while the actual data will be stored asynchronously.

module/state_synchronization/execution_data_service.go
```diff
@@ -2,430 +2,44 @@ package state_synchronization
...
-// BlobNotFoundError is returned when the blobservice failed to find a blob.
-type BlobNotFoundError struct {
-	cid cid.Cid
-}
-
-func (e *BlobNotFoundError) Error() string {
-	return fmt.Sprintf("blob %v not found", e.cid.String())
+const DefaultMaxBlobSize = 1 << 20 // 1MiB
+
+// ExecutionDataAdder handles adding execution data to a blobservice
```
jordanschalm, line 12: I think some extended documentation here would help with understanding the code, eg:
* Adder corresponds to data producer/provider (execution node)
* Adding data to the blob service stores it on the node's disk for future requests, and makes it available to other nodes via bitswap

Likewise for the Getter interface

module/state_synchronization/getter.go
```diff
@@ -0,0 +1,313 @@
...
+			if blob, err = s.findBlob(ctx, blobChan, c, cachedBlobs, logger); err != nil {
+				return err
+			}
+		}
+
+		cidCounts[c] -= 1
+
+		if cidCounts[c] == 0 {
+			delete(cachedBlobs, c)
+		}
```
jordanschalm, line 106: Is this to be able to free memory taken by a cached blob earlier in the for loop?

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,995 @@
...
+
+var _ StatusTrackerFactory = (*Storage)(nil)
+
+// badger key prefixes
+const (
+	prefixGlobalState         byte = iota + 1 // global state variables
+	prefixExecutionDataStatus                 // tracks the status of each execution data
+	prefixBlobRecord                          // tracks the latest height among all blob trees containing a blob
+	prefixBlobTreeRecord                      // tracks the set of blobs for each blob tree
+)
```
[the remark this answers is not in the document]
smnzhu: This is a separate db instance, yes. I can mention that in the comments

module/state_synchronization/adder.go
```diff
@@ -0,0 +1,285 @@
...
+		s.metrics.ChunkExecutionDataAddFinished(time.Since(start), success, blobTreeSize)
+	}()
+
+	cids, totalBytes, err := s.addBlobs(ctx, ced, logger)
+	if err != nil {
+		return cid.Undef, fmt.Errorf("failed to add chunk execution data blobs: %w", err)
+	}
+
+	for {
+		if logger.GetLevel() >= zerolog.DebugLevel {
```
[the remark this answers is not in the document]
smnzhu: Sure, I can do that

module/state_synchronization/adder.go
```diff
@@ -0,0 +1,285 @@
...
+		return flow.ZeroID, fmt.Errorf("failed to add execution data root to blobservice: %w", err)
+	}
+
+	success = true
+	size = buf.Len()
+
+	return flow.CidToId(rootBlob.Cid())
+}
+
+func (s *executionDataAdderImpl) AddChunkExecutionData(
```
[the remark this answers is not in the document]
smnzhu: The documentation is on the interface, in execution_data_service.go. Can you check if those documentation look good?

module/state_synchronization/getter.go
smnzhu: I don't quite remember tbh, it looks like that might have been the reason.

module/state_synchronization/getter.go
smnzhu: This also means that every once in a while, when pruning is happening, Execution Nodes will experience a pause in the block execution pipeline because Execution Data cannot be stored until the pruning is done. This would be infrequent (we can configure the pruning to only happen once per day or something, for example), but still, I wonder if this would be another reason to separate CID generation from the actual submission of the data blobs to badger / Bitswap as suggested [here](https://github.com/onflow/flow-go/pull/2292/files/7234bce41a91a71fad6244bcd3d70f5491cc905b#r849989908)?

module/state_synchronization/adder.go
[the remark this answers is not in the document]
smnzhu: The documentation for this is on the interface, in execution_data_service.go. Please comment there if you think anything needs to be changed / added

module/state_synchronization/storage.go
smnzhu: Bitswap is not a data store, it is just a data transfer protocol. The actual data is stored in a Blockstore, which in our case is backed by badger. Bitswap requires you to pass in a [Blockstore](https://pkg.go.dev/github.com/ipfs/go-ipfs-blockstore) implementation which it uses to check if we have data requested by peers, and to store data that we've requested once we receive it. The implementation we use is https://pkg.go.dev/github.com/ipfs/go-ds-badger.

The key thing to understand here is that the underlying badger db used by the Blockstore implementation that we pass to Bitswap, is the same db that we use in `Storage`. This is so that we can prune (remove) data blobs while simultaneously updating our tracking metadata (blob records, blob tree records, etc) in a transactionally consistent manner. 

For what it's worth, I am very open to suggestions on how we can decouple the two DB's while still guaranteeing consistency. So far I haven't been able to think of anything.

module/state_synchronization/adder.go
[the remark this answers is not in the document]
smnzhu: [6384547](https://github.com/onflow/flow-go/pull/2292/commits/6384547cd9cc23e8177a9ba7669ca8f1926d1c20)

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,996 @@
...
+	}); err != nil {
+		return nil, err
+	}
+
+	latestIncorporatedHeight, err := s.GetLatestIncorporatedHeight()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest incorporated height: %w", err)
+	}
+
+	if numExecutionDataStatuses > 0 {
```
jordanschalm, line 370:
```suggestion

	// second, ensure execution data statuses are consistent with the blob tree index
	if numExecutionDataStatuses > 0 {
```

jordanschalm (commented):
This review focussed on `storage.go`.

module/state_synchronization/getter.go
```diff
@@ -0,0 +1,313 @@
...
+	targetLogger.Debug().Msg("blob not found")
+
+	return nil, &BlobNotFoundError{target}
+}
+
+// getBlobs gets the given CIDs from the blobservice, reassembles the blobs, and deserializes the reassembled data into an object.
+//
+// blobs are fetched from the blobservice and sent over a blob channel as they arrive, to a separate goroutine, which performs the
+// deserialization in a streaming fashion.
+func (s *executionDataGetterImpl) getBlobs(ctx context.Context, cids []cid.Cid, logger zerolog.Logger) (interface{}, uint64, error) {
```
zhangchiqing, line 161: Since the result is either a slice of CIDs or a ChunkExecutionData, better to make it clear here. if `isMoreCIDs` is true, read moreCIDs, if false, read `chunkExecutionData`.
```suggestion
func (s *executionDataGetterImpl) getBlobs(ctx context.Context, cids []cid.Cid, logger zerolog.Logger) (moreCIDs []cid.CID , chunkExecutionData *ChunkExecutionData, isChunkExecutionData bool, depth uint64, err error) {
```

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,996 @@
...
+					"inconsistency detected: blob record height for Cid %s is %d, but blob tree record exists at height %d",
+					c.String(), latestHeightContainingBlob, blockHeight,
+				)
+			} else if latestHeightContainingBlob == blockHeight {
+				if err := txn.Delete(blobRecordKey); err != nil {
+					return fmt.Errorf("failed to delete blob record for Cid %s: %w", c.String(), err)
+				}
+
+				blobKey := network.BlobServiceDatastoreKeyFromCid(c).Bytes()
+				if err := txn.Delete(blobKey); err != nil {
```
zhangchiqing, line 578:
```suggestion
				// BitSwap store all data blobs using BlobStore, which is a key-value store. 
				// data blobs are stored under the key that is determined by CID of the blob. 
				// in order to delete the Blob for a given CID, we first find the blob key, and then simply
				// remove value by the key.
				// we first find the blobKey 
				blobKey := network.BlobServiceDatastoreKeyFromCid(c).Bytes()
				if err := txn.Delete(blobKey); err != nil {
```

module/state_synchronization/storage.go
```diff
@@ -0,0 +1,996 @@
...
+	return executionDataStatusKey
+}
+
+func parseExecutionDataStatusKey(key []byte) uint64 {
+	return binary.LittleEndian.Uint64(key[1:])
+}
+
+const blobTreeRecordKeyLength = 1 + 8 + blobs.CidLength
+
+func makeBlobTreeRecordKey(blockHeight uint64, c cid.Cid) []byte {
```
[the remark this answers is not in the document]
zhangchiqing: Yeah, the storage is a key-value store, we need to make it clear what key-value pairs are we storing in the storage.

My understanding is currently have:
```
last_incorporated_height: <height> // to find the highest height that has incorporated 
lowest_pruned_height: <height> // to find the next unpruned height (lowest_pruned_height + 1) to prune 
executionDataStatus_<height> : inprogress | complete // to track whether the execution data for the height has been fully downloaded
blobRecord_<cid>: blob // to find the blob by cid
blobTreeRecord_<height>_<cid>: nil // to find cid by height, useful to know what cids to prune for a certain height
```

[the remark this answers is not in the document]
zhangchiqing: I don't quite understand what is `blobTreeRecord_<height>_<cid>` for. Also why do we need to track `latestHeightSeen` and `earliestBlobTreeRecordHeight`, `highestBlobTreeRecordHeight` during `checkBlobTreeRecords`.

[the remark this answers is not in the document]
zhangchiqing: ⚠️ Note execution node executes incorporated blocks, meaning they have not finalized yet. So it's possible there are multiple blocks at the same height got executed, in which case the `executionDataStatus_<height>` can't track for a single block, also `blobTreeRecord_<height>_<cid>` would index cids for all blocks at the same height.

module/state_synchronization/adder.go
zhangchiqing: >but I think I actually want to incorporate the adding of the individual Chunk Execution Data as part of the block execution pipeline, here:

The execution node executes incorporated blocks (not finalized yet). But access nodes usually only care about the execution state for the sealed blocks. When answering user's state query, they are ran against the state from sealed blocks. So it's not necessary to store or sync execution data for blocks that haven't sealed yet. If we integrate adding individual chunk execution data to the block execution pipeline, then we are possibly storing for blocks that eventually been abandoned as they are conflicting with finalized forks. 

So I think we don't have to integrate it into the block execution pipeline. Instead, having another thread going through each finalized block, which eventually will be sealed, and store execution data for them. Thoughts?

zhangchiqing: Another thought: If we are not integrating them, then we probably don't need  `AddChunkExecutionData` to add for each individual chunk, but just adding for all chunks.

module/state_synchronization/adder.go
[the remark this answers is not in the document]
jordanschalm: I think we could still add the 2nd/3rd suggestions above to the ` execution_data_service.go` interface documentation. They aren't included there, and to me it wasn't obvious at first.

zhangchiqing (commented):
Haven't finished reviewing, but spotted some misunderstanding. Just to point out that first.

jordanschalm (commented):
First round of comments.

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+	maxItemCountByWriteSize := db.MaxBatchSize() / totalSizePerItem
+
+	if maxItemCountByWriteCount < maxItemCountByWriteSize {
+		return int(maxItemCountByWriteCount)
+	} else {
+		return int(maxItemCountByWriteSize)
+	}
+}
+
+type TrackBlobsFn func(uint64, ...cid.Cid) error
```
jordanschalm, line 87: Please add documentation for the exported types and functions. Some suggestions:
* is the height value inclusive or exclusive
* are there any expected errors?

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+func (s *storage) batchDeleteItemLimit() int {
+	itemsPerBatch := 256
+	maxItemsPerBatch := getBatchItemCountLimit(s.db, 2, blobRecordKeyLength+latestHeightKeyLength)
+	if maxItemsPerBatch < itemsPerBatch {
+		itemsPerBatch = maxItemsPerBatch
+	}
+	return itemsPerBatch
+}
+
+func (s *storage) Prune(height uint64) error {
```
jordanschalm, line 322: Please document this function. For example:
* what data is pruned?
* pruning locks the tracker (other writes are blocked during the pruning)
* prunes up to and including `height`
* what happens if we prune a height that isn't fulfilled?

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+
+		if err := txn.Set(prunedHeightKey, prunedHeightValue); err != nil {
+			return fmt.Errorf("failed to set pruned height value: %w", err)
+		}
+
+		return nil
+	})
+}
+
+func (s *storage) Update(f UpdateFn) error {
```
jordanschalm, line 184: Please add documentation for this function. For example:
* used to add new blob CIDs to the tracker
* blob CIDs may be for either sealed or unsealed heights
* the same CID can be added multiple times, for different heights
* can the same CID be added multiple times, for the same height?

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/module/blobs"
+)
+
+// badger key prefixes
+const (
+	prefixGlobalState  byte = iota + 1 // global state variables
+	prefixLatestHeight                 // tracks the latest height containing each blob
```
jordanschalm, line 20: Can we explicitly document which height we're referring to here and elsewhere in this file where we refer to height? Given an execution receipt like:
```
A<-B(Receipt_A)
```
are we referring to `A.height` - the height of the executed block - or `B.height` - the height of the container block.

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+
+func (s *storage) Prune(height uint64) error {
+	blobRecordPrefix := []byte{prefixBlobRecord}
+	itemsPerBatch := s.batchDeleteItemLimit()
+	var batch []*deleteInfo
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.setPrunedHeight(height)
```
jordanschalm, line 330: ⚠️ unhandled error

jordanschalm (commented):
Adding comments for `tracker/storage.go`

jordanschalm (commented):
Most of the new components added don't have associated tests. Is there a follow-up issue for adding tests in a separate PR?

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get pruned height: %w", err)
+	}
+
+	fulfilledHeight, err := storage.GetFulfilledHeight()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get fulfilled height: %w", err)
+	}
+
+	fulfilledHeightsIn, fulfilledHeightsOut := util.UnboundedChannel()
```
jordanschalm, line 65: I am wary of using unbounded channels like this. In comparison to unbuffered channels or message stores with observable and capped sizes, it is more difficult to detect bottlenecks or reason about the memory usage of messages awaiting processing. 

The usage here is one where the scope is limited enough to be pretty confident it will work as expected, but I think it's a pattern we should avoid in general on Flow, so I'm inclined to suggest we consider removing it even for this usage. 

Some suggestions for alternatives:
* caller updates mutex-protected values (fulfilled height, threshold, height range) directly, uses `Notifier` to notify pruning thread for every change
* replace `UnboundedChannel` with something similar that buffers only the maximum input in the channel, drop other inputs

jordanschalm (commented):
Pruner comments

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/onflow/flow-go/module/executiondatasync/execution_data"
+	"github.com/onflow/flow-go/module/executiondatasync/tracker"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+)
+
+type jobResult struct {
+	executionData *execution_data.BlockExecutionData
+	resultID      flow.Identifier
+	blockHeight   uint64
```
jordanschalm, line 20:
```suggestion
	blockHeight   uint64 // height of the executed block
```

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+	err           error
+}
+
+type fulfiller struct {
```
jordanschalm, line 24: Please add documentation for this component. For example:
* Accepts sealed results from Dispatcher and completed execution data download jobs from Handler
* Defn: a block is "fulfilled" when it has been sealed and we have downloaded the execution data for that block
* Detects when a block is fulfilled, and:
  * notifies subscribers via Notifier
  * updates the fulfilled height in execution state download tracker

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+func (f *fulfiller) fulfill() error {
+loop:
+	if j, ok := f.fulfilled[f.fulfilledHeight+1]; ok {
+		f.fulfilledHeight++
+		delete(f.fulfilled, f.fulfilledHeight)
+		delete(f.sealedResults, f.fulfilledHeight)
+		f.notifier.notify(j.blockHeight, j.executionData)
+
+		goto loop
+	}
```
jordanschalm, line 98:
```suggestion
	for {
		job, ok := f.fulfilled[f.fulfilledHeight+1]
		if !ok {
			break
		}
		
		f.fulfilledHeight++
		delete(f.fulfilled, f.fulfilledHeight)
		delete(f.sealedResults, f.fulfilledHeight)
		f.notifier.notify(job.blockHeight, job.executionData)
	}
```
Better to use the more common loop style.

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+			jmap = make(map[flow.Identifier]*jobResult)
+			f.jobResults[j.blockHeight] = jmap
+		}
+
+		jmap[j.resultID] = j
+	}
+}
+
+func (f *fulfiller) handleSealedResult(ctx irrecoverable.SignalerContext, sealedResultID flow.Identifier) {
+	f.sealedHeight++
```
jordanschalm, line 149: ⚠️  There is a critical undocumented assumption here. For `fulfiller`'s `sealedHeight` state to be correct, sealed results must be delivered exactly once and in the correct order. 

It would be better for this function to accept the block height corresponding to `sealedResultID`, then we can set `f.sealedHeight = max(f.sealedHeight, sealedResultHeight)` and have less rigid constraints on inputs. If we need to keep the assumption about inputs, please note it in the function documentation.

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+	executionData *execution_data.BlockExecutionData
+	resultID      flow.Identifier
+	blockHeight   uint64
+	err           error
+}
+
+type fulfiller struct {
+	jobResults    map[uint64]map[flow.Identifier]*jobResult
+	sealedResults map[uint64]flow.Identifier
+	fulfilled     map[uint64]*jobResult
```
jordanschalm, line 27:
```suggestion
	// jobResults tracks all completed jobs for unsealed heights, including potentially for
	// invalid results which will never be sealed
	// height -> resultID -> jobResult
	jobResults    map[uint64]map[flow.Identifier]*jobResult 
	// sealedResults tracks the result ID which is sealed at each height
	sealedResults map[uint64]flow.Identifier
	// fulfilled tracks the completed, successful job for the most recent sealed heights
	// used to trigger notifications
	fulfilled     map[uint64]*jobResult 
```

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (f *fulfiller) handleSealedResult(ctx irrecoverable.SignalerContext, sealedResultID flow.Identifier) {
+	f.sealedHeight++
+	f.sealedResults[f.sealedHeight] = sealedResultID
+
+	if completedJobs, ok := f.jobResults[f.sealedHeight]; ok {
+		// once we know the sealed result ID, we can discard all unneeded job results
+		delete(f.jobResults, f.sealedHeight)
+
+		for resultID, j := range completedJobs {
```
jordanschalm, line 156:
```suggestion
		for resultID, job := range completedJobs {
```
Better to avoid too many one-character variables.

module/executiondatasync/requester/fulfiller.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/onflow/flow-go/module/executiondatasync/tracker"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+)
+
+type jobResult struct {
+	executionData *execution_data.BlockExecutionData
+	resultID      flow.Identifier
+	blockHeight   uint64
+	err           error
```
jordanschalm, line 21:
```suggestion
	// err is set when a critical, non-retryable error was encountered while downloading execution data for this result
	err           error
```

jordanschalm (commented):
Comments for Dispatcher, Fulfiller

jordanschalm (commented):
Completed reviewing all files. Nice work on the separation of components in the Requester, and the detailed metrics.

Some general comments: 
* I have some concerns about the use of unbounded channels in several of the components. I think we should put size limits on these channels, add metrics for visibility into their size, or both.
* Many of the new components do not have test coverage. Given the TODO in the PR description, my understanding is that tests will be added a follow-up PR. I suggest that we merge this PR to a feature branch (not master) until we add test coverage for the new components. 
* I think we can improve the documentation for the components added in this PR. I left comments where I thought we should add or extend documentation, but will summarize here in case I missed anything:
  * For each new component, documentation explaining what the component does and how it relates to the other components.
  * Godoc for most exported types, unless their purpose is obvious or trivial
  * Documentation for private methods which are more complex or rely on assumptions about their inputs
  * For all public methods, documentation of expected errors

engine/execution/messages.go
```diff
@@ -17,18 +18,18 @@ type ComputationOrder struct {
...
+	StateCommitments        []flow.StateCommitment
+	Proofs                  [][]byte
+	Events                  []flow.EventsList
+	EventsHashes            []flow.Identifier
+	ServiceEvents           flow.EventsList
+	TransactionResults      []flow.TransactionResult
+	ComputationUsed         uint64
+	StateReads              uint64
+	TrieUpdates             []*ledger.TrieUpdate
+	ExecutionDataProvideJob *provider.ProvideJob
```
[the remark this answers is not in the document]
smnzhu: It's because the storing of the blobs happens asynchronously. I can split the done channel out separately, but at some point somewhere we will still need to check the error and log it.

engine/execution/ingestion/engine.go
```diff
@@ -638,6 +642,16 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
 	if err != nil {
 		e.log.Err(err).Msg("failed in process block's children")
 	}
+
+	err, ok := <-computationResult.ExecutionDataProvideJob.Done
```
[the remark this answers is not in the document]
smnzhu: 1. I don't think it actually causes block execution to halt, because block execution happens in a separate gorouting for each block
2. We can just make the storing of the execution happen synchronously to simplify things.

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+	})
+}
+
+func (s *storage) Update(f UpdateFn) error {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return f(s.trackBlobs)
+}
+
+func (s *storage) SetFulfilledHeight(height uint64) error {
```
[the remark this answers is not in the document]
smnzhu: It's up to the caller to ensure this.

state/protocol/badger/mutator.go
```diff
@@ -633,6 +633,21 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+			err := operation.IndexExecutionResult(seal.BlockID, seal.ResultID)(tx)
+			if err == nil {
+				continue
+			}
+
+			if !errors.Is(err, storage.ErrAlreadyExists) {
+				return fmt.Errorf("could not index sealed execution result: %w", err)
+			}
+
+			if err := operation.ReindexExecutionResult(seal.BlockID, seal.ResultID)(tx); err != nil {
```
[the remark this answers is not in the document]
smnzhu: Can you rewrite this comment? There are a lot of typos and I have a hard time understanding what you're saying here.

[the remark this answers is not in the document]
smnzhu: When it was already previously indexed, before it was sealed

cmd/access/node_builder/access_node_builder.go
```diff
@@ -375,6 +387,97 @@ func (builder *FlowAccessNodeBuilder) BuildConsensusFollower() AccessNodeBuilder
...
+				return nil, fmt.Errorf("cannot get the sealed block: %w", err)
+			}
+
+			trackerDir := filepath.Join(builder.executionDataDir, "tracker")
+			trackerStorage, err = tracker.OpenStorage(
+				trackerDir,
+				sealed.Height,
+				node.Logger,
+				tracker.WithPruneCallback(func(c cid.Cid) error {
+					// TODO: use a proper context here
```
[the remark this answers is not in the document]
smnzhu: atm, not sure where the context would come from...

engine/execution/ingestion/engine.go
```diff
@@ -638,6 +642,16 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
...
+	if ok {
+		e.log.Err(err).
+			Hex("block_id", logging.Entity(executableBlock)).
+			Str("execution_data_id", receipt.ExecutionResult.ExecutionDataID.String()).
+			Msg("failed to provide execution data blobs")
+	}
+
+	e.executionDataPruner.NotifyFulfilledHeight(executableBlock.Height())
```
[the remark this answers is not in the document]
smnzhu: i think we'll just make the Provider synchronous as discussed

module/executiondatasync/tracker/storage.go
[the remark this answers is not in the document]
smnzhu: What if we instead add more comments to the function declaration stating that it should never be called for a height above the fulfilled height?

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,454 @@
...
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/module/blobs"
+)
+
+// badger key prefixes
+const (
+	prefixGlobalState  byte = iota + 1 // global state variables
+	prefixLatestHeight                 // tracks the latest height containing each blob
+	prefixBlobRecord                   // tracks the set of blobs at each height
```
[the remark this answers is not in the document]
smnzhu: The `PruneCallback` should be responsible for handling this. In the current code, what would happen is that the callback will try to delete the blob from the blobstore, which will be a noop if the blob doesn't exist.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, opt := range opts {
+		opt(p)
+	}
+
+	return p, nil
+}
+
+func (p *Pruner) NotifyFulfilledHeight(height uint64) {
+	if util.CheckClosed(p.cm.ShutdownSignal()) {
```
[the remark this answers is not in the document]
smnzhu: Because it's an unbounded channel, the in channel will always accept new items. So we should check the shutdownsignal first before sending to the channel, otherwise the send may succeed even if the shutdownsignal is closed

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (p *Pruner) loop(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+	ready()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case h := <-p.fulfilledHeightsOut:
+			p.lastFulfilledHeight = h.(uint64)
```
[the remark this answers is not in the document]
smnzhu: They are guaranteed to be strictly increasing, but we also can take the max just to be safe.

module/executiondatasync/pruner/pruner.go
smnzhu: It's probably not necessary here, since we only really care about processing the most recent item, but I introduced it mainly for other usecases in the requester where we need to process every single item, yet we don't want the producers to be blocked waiting for the consumers to finish processing.

For this specific case, I'm fine replacing it with one of your suggestions.

module/executiondatasync/requester/dispatcher.go
```diff
@@ -0,0 +1,264 @@
...
+	blocks storage.Blocks,
+	results storage.ExecutionResults,
+	handler *handler,
+	fulfiller *fulfiller,
+	logger zerolog.Logger,
+	metrics module.ExecutionDataRequesterMetrics,
+) *dispatcher {
+	receiptsIn, receiptsOut := util.UnboundedChannel()
+	finalizedBlocksIn, finalizedBlocksOut := util.UnboundedChannel()
+	resultInfosIn, resultInfosOut := util.UnboundedChannel()
```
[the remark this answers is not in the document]
smnzhu: as I mentioned above, the reason I initially introduced them is because they allow the rate of production and consumption to be independent of each other. This is better in situations where overall the consumer is able to keep up with the rate of consumption, but there may be moments during which the consumer is processing items and producers would otherwise block during those moments if we didn't use an unbuffered channel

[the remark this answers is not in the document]
smnzhu: Let me know what you think about this, I'm not sure which approach is better in terms of tradeoffs.

module/executiondatasync/requester/dispatcher.go
```diff
@@ -0,0 +1,264 @@
...
+					Msg("could not retrieve corresponding block for execution receipt, skipping")
+				d.metrics.ReceiptSkipped()
+				continue
+			}
+
+			d.resultInfosIn <- &resultInfo{
+				resultID:        receipt.ExecutionResult.ID(),
+				blockID:         receipt.ExecutionResult.BlockID,
+				blockHeight:     block.Header.Height,
+				executionDataID: receipt.ExecutionResult.ExecutionDataID,
```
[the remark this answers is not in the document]
smnzhu: the default would already be false. Are you just suggesting this because it's more clear?

module/executiondatasync/requester/notifier.go
```diff
@@ -0,0 +1,119 @@
...
+
+	subs   chan *subRequest
+	unsubs chan *subRequest
+
+	subscriptions map[*notificationSub]struct{}
+
+	logger zerolog.Logger
+
+	cm *component.ComponentManager
+	component.Component
```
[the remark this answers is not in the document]
smnzhu: just because I didn't want the `ShutdownSignal` to be publicly accessible method

module/blobs/blob_store.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+)
+
+type Blobstore interface {
```
SaveTheRbtz, line 12: just curious:  Why do we copy [Blockstore](https://pkg.go.dev/github.com/ipfs/go-ipfs-blockstore#Blockstore) interface instead of using it directly?

module/blobs/blob_store.go
```diff
@@ -0,0 +1,82 @@
...
+	// the CIDs in the Blobstore can be read. It should respect
+	// the given context, closing the channel if it becomes Done.
+	AllKeysChan(ctx context.Context) (<-chan cid.Cid, error)
+
+	// HashOnRead specifies if every read blob should be
+	// rehashed to make sure it matches its CID.
+	HashOnRead(enabled bool)
+}
+
+var ErrNotFound = errors.New("blobstore: blob not found")
```
SaveTheRbtz, line 37: nit: CID would be useful here

module/blobs/blob_store.go
smnzhu: Block is confusing terminology in a blockchain codebase

engine/execution/computation/computer/computer.go
```diff
@@ -198,13 +209,20 @@ func (e *blockComputer) executeBlock(
 		if err != nil {
 			return nil, fmt.Errorf("cannot merge view: %w", err)
 		}
+
+		col := collection.Collection()
+		executionData.ChunkExecutionDatas = append(executionData.ChunkExecutionDatas, &execution_data.ChunkExecutionData{
+			Collection: &col,
+			Events:     res.Events[i],
+		})
```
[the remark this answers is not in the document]
smnzhu: [048b47d](https://github.com/onflow/flow-go/pull/2292/commits/048b47d36ec6ca3bc63d7513027f190786e84ac7)

cmd/execution/main.go
```diff
@@ -371,79 +388,61 @@ func main() {
...
+			var prunerMetrics module.ExecutionDataPrunerMetrics = metrics.NewNoopCollector()
+			if node.MetricsEnabled {
+				prunerMetrics = metrics.NewExecutionDataPrunerCollector()
+			}
 
+			executionDataPruner, err = pruner.NewPruner(
+				node.Logger,
+				prunerMetrics,
+				executionDataTracker,
+				pruner.WithHeightRangeTarget(65000),
```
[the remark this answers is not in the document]
smnzhu: [6c9e6b9](https://github.com/onflow/flow-go/pull/2292/commits/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7)

cmd/access/node_builder/access_node_builder.go
```diff
@@ -375,6 +387,97 @@ func (builder *FlowAccessNodeBuilder) BuildConsensusFollower() AccessNodeBuilder
...
+			if node.MetricsEnabled {
+				prunerMetrics = metrics.NewExecutionDataPrunerCollector()
+			}
+
+			executionDataPruner, err := pruner.NewPruner(
+				node.Logger,
+				prunerMetrics,
+				trackerStorage,
+				pruner.WithHeightRangeTarget(65000),
+				pruner.WithThreshold(65000),
```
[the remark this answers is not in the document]
smnzhu: [6c9e6b9](https://github.com/onflow/flow-go/pull/2292/commits/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7)

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/onflow/flow-go/module/component"
+	"github.com/onflow/flow-go/module/executiondatasync/tracker"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+	"github.com/rs/zerolog"
+)
+
+const (
+	defaultHeightRangeTarget = uint64(400000)
+	defaultThreshold         = uint64(100000)
```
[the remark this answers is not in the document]
smnzhu: [75cef3d](https://github.com/onflow/flow-go/pull/2292/commits/75cef3d460d44245cc7475f175d4e8df7959a399)

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+type Pruner struct {
```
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+
+	fulfilledHeightsIn    chan<- interface{}
+	fulfilledHeightsOut   <-chan interface{}
+	thresholdChan         chan uint64
+	heightRangeTargetChan chan uint64
+
+	lastFulfilledHeight uint64
+	lastPrunedHeight    uint64
+	heightRangeTarget   uint64
+	threshold           uint64
```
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/pruner/pruner_test.go
```diff
@@ -0,0 +1,172 @@
...
+	pruner.Start(signalerCtx)
+
+	pruned := make(chan struct{})
+	trackerStorage.On("Prune", uint64(6)).Return(func(height uint64) error {
+		close(pruned)
+		return nil
+	}).Once()
+
+	pruner.NotifyFulfilledHeight(16)
+	<-pruned
```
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/pruner/pruner.go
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (p *Pruner) NotifyFulfilledHeight(height uint64) {
+	if util.CheckClosed(p.cm.ShutdownSignal()) {
+		return
+	}
+
+	p.fulfilledHeightsIn <- height
+}
+
+func (p *Pruner) SetHeightRangeTarget(heightRangeTarget uint64) error {
```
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,157 @@
...
+	defaultThreshold         = uint64(100000)
+)
+
+type Pruner struct {
+	storage tracker.Storage
+
+	fulfilledHeightsIn    chan<- interface{}
+	fulfilledHeightsOut   <-chan interface{}
+	thresholdChan         chan uint64
+	heightRangeTargetChan chan uint64
```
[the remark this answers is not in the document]
smnzhu: [61680e0](https://github.com/onflow/flow-go/pull/2292/commits/61680e0cdb7d599afffcad7e5ef5e2f15c868de5)

module/executiondatasync/requester/fulfiller.go
smnzhu: job is the name of the struct, so I use j to avoid confusion

module/executiondatasync/requester/handler.go
```diff
@@ -0,0 +1,434 @@
...
+	return v, result.totalSize, nil
+}
+
+type retrieveBlobsResult struct {
+	err       error
+	totalSize uint64
+}
+
+// retrieveBlobs retrieves the blobs for the given CIDs from the blobservice.
+func (h *handler) retrieveBlobs(parent context.Context, cids []cid.Cid) (<-chan blobs.Blob, <-chan *retrieveBlobsResult) {
```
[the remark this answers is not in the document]
smnzhu: Gonna leave it on handler for now, because `findBlob` uses `h.maxBlobSize`

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+
+func (s *storage) Prune(height uint64) error {
+	blobRecordPrefix := []byte{prefixBlobRecord}
+	itemsPerBatch := s.batchDeleteItemLimit()
+	var batch []*deleteInfo
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if err := s.setPrunedHeight(height); err != nil {
```
SaveTheRbtz, line 384: do we want to set pruned height even if prunning fails?

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+		return err
+	}
+
+	if err := s.db.View(func(txn *badger.Txn) error {
+		it := txn.NewIterator(badger.IteratorOptions{
+			PrefetchValues: false,
```
SaveTheRbtz, line 390: nit: any reason to not prefetch?

module/executiondatasync/provider/provider.go
```diff
@@ -0,0 +1,234 @@
...
+
+	errCh := p.storeBlobs(ctx, blockHeight, blobCh)
+	g, gCtx := errgroup.WithContext(ctx)
+
+	chunkDataIDs := make([]cid.Cid, len(executionData.ChunkExecutionDatas))
+	for i, chunkExecutionData := range executionData.ChunkExecutionDatas {
+		i := i
+		chunkExecutionData := chunkExecutionData
+
+		g.Go(func() error {
```
SaveTheRbtz, line 145: nit: depending on the possible number of chunks in the executionData you may consider upperbounding concurrency here.  (Or ideally, creating a worker pool for the `Provider` during `New`)

SaveTheRbtz (commented):
Feel free to ignore nits or followup on them in later diffs.

PR #2293, opened 2022-04-12
Fix race detector issues.

SaveTheRbtz:
While exploring code base I've stumbled upon a couple of race-detector problems. Most of them are harmless issues that only affect test flakiness.  Nevertheless, here is a sample of interesting ones:

* Unprotected `OnSucceed` callback access in badger batches: https://github.com/onflow/flow-go/commit/35babdbc398dbff812d7abf69054e17534623126
* Potential error clobbering in ledger migration's "storage-used" calculation: https://github.com/onflow/flow-go/commit/c42159fc719ed97414922515d362492e1b8b956a
* EpochMgr `epochs` access without lock on `Done`: https://github.com/onflow/flow-go/commit/a0c6cee65888f427a3482e81670cbb88fcce7f9b
* `BLS12_381` verification accesses uninitialized memory: https://github.com/onflow/flow-go/pull/2293#discussion_r850080079
* Unprotected global access in tests https://github.com/onflow/flow-go/commit/d26ca03a58b1eff8295b267de6a0fce845587044
* Lock being copied by value by the mock framework: https://github.com/onflow/flow-go/commit/a89101eb0ffb5aff16deea5e1a12926a44262a93
* `time.Sleep` references in tests: https://github.com/onflow/flow-go/commit/05a87c703eafc1f158851cb011d8e2da72e149ce
* Ignore `runtime.noescape` in `checkptr`: https://github.com/onflow/flow-go/pull/2293/commits/1288c1dbcfa19c4249cdc0ae88078e7f4112b4d5

I needed to disable the network topology test in `-race` since it was triggering the goroutine limit of the race detector
https://github.com/onflow/flow-go/commit/bb24d8f519cffe37f900aeb7f945e1db5f9be2c0.  For this one we'll need to wait until the support of infinite number of goroutines is released https://github.com/golang/go/commit/d6a1ffd624bd0d6dbf3a15070e378749612b35c9 is available (go likely 1.19).

Also there is a race condition in herocache that can be solved by either:
* Making Counter and Timestamp atomic: https://github.com/SaveTheRbtz/flow-go/commit/d1cc40686c5e51a4728f07305bcfea0fb2a43936
* Removing `logTelemetry` methods from the "read-only" methods protected only by `RLock`: https://github.com/SaveTheRbtz/flow-go/commit/6b0dcbdbefaaa1cb2d652b1bdbd9c5505ca298e9
What is the better approach here? (cc: @yhassanzadeh13)

Testify also has couple of nasty races in their `Suite` implementation if coupled with spawning gorouptines: https://github.com/stretchr/testify/issues/1168#issuecomment-1072082613: https://github.com/onflow/flow-go/commit/485161434ad83641a2d8b1f86f0dc74d72a30fa2 https://github.com/onflow/flow-go/commit/71b3fdc2d8d6c228042118bcdf9bdcc9960a1e88

Looking at [race detector](https://go.dev/doc/articles/race_detector)'s ability to find timing/concurrency related problems in both tests and code it looks like it should likely be a first-class citizen in the Flow's CI/CD pipeline. WDYT about its ROI?
Seems like @huitseeker did look at this problem some time ago but since then race detector work was mostly abandoned.  Maybe it is possible to make it an optional part of the CI/CD pipeline, like fuzzing and benchmarking?

### DIFF
--- a/cmd/util/ledger/migrations/storage_used_update_migration.go
+++ b/cmd/util/ledger/migrations/storage_used_update_migration.go
@@ -2,12 +2,15 @@ package migrations
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"os"
 	"path"
 	"runtime"
 	"sync"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 
 	fvm "github.com/onflow/flow-go/fvm/state"
 	"github.com/onflow/flow-go/ledger"
@@ -72,26 +75,27 @@ func (m *StorageUsedUpdateMigration) Migrate(payload []ledger.Payload) ([]ledger
 
 	storageUsed := make(map[string]uint64)
 	storageUsedChan := make(chan accountPayloadSize, workerCount)
-	payloadChan := make(chan indexedPayload)
+	payloadChan := make(chan indexedPayload, workerCount)
 	storageUsedPayloadChan := make(chan accountStorageUsedPayload, workerCount)
 	storageUsedPayload := make(map[string]int)
 
-	inputWG := &sync.WaitGroup{}
+	inputEG, ctx := errgroup.WithContext(context.Background())
 	outputWG := &sync.WaitGroup{}
 
 	outputWG.Add(1)
 	go func() {
+		defer outputWG.Done()
 		for payloadSize := range storageUsedChan {
 			if _, ok := storageUsed[payloadSize.Address]; !ok {
 				storageUsed[payloadSize.Address] = 0
 			}
 			storageUsed[payloadSize.Address] = storageUsed[payloadSize.Address] + payloadSize.StorageUsed
 		}
-		outputWG.Done()
 	}()
 
 	outputWG.Add(1)
 	go func() {
+		defer outputWG.Done()
 		for su := range storageUsedPayloadChan {
 			if _, ok := storageUsedPayload[su.Address]; ok {
 				m.Log.Error().
@@ -100,17 +104,15 @@ func (m *StorageUsedUpdateMigration) Migrate(payload []ledger.Payload) ([]ledger
 			}
 			storageUsedPayload[su.Address] = su.Index
 		}
-		outputWG.Done()
 	}()
 
 	for i := 0; i < workerCount; i++ {
-		inputWG.Add(1)
-		go func() {
+		inputEG.Go(func() error {
 			for p := range payloadChan {
-				var id flow.RegisterID
-				id, err = KeyToRegisterID(p.Payload.Key)
+				id, err := KeyToRegisterID(p.Payload.Key)
 				if err != nil {
 					log.Error().Err(err).Msg("error converting key to register ID")
+					return err
 				}
 				if len([]byte(id.Owner)) != flow.AddressLength {
 					// not an address
@@ -127,19 +129,21 @@ func (m *StorageUsedUpdateMigration) Migrate(payload []ledger.Payload) ([]ledger
 					StorageUsed: uint64(registerSize(id, p.Payload)),
 				}
 			}
-			inputWG.Done()
-		}()
+			return nil
+		})
 	}
 
+Loop:
 	for i, p := range payload {
-		payloadChan <- indexedPayload{
-			Index:   i,
-			Payload: p,
+		select {
+		case <-ctx.Done():
+			break Loop
+		case payloadChan <- indexedPayload{Index: i, Payload: p}:
 		}
 	}
 
 	close(payloadChan)
-	inputWG.Wait()
+	err = inputEG.Wait()
 	close(storageUsedChan)
 	close(storageUsedPayloadChan)
 	outputWG.Wait()
--- a/consensus/hotstuff/votecollector/combined_vote_processor_v2_test.go
+++ b/consensus/hotstuff/votecollector/combined_vote_processor_v2_test.go
@@ -451,7 +451,7 @@ func TestCombinedVoteProcessorV2_PropertyCreatingQCCorrectness(testifyT *testing
 			"staking signers: %v, beacon signers: %v\n\t"+
 			"required weight: %v", stakingSignersCount, beaconSignersCount, minRequiredWeight)
 
-		stakingTotalWeight, collectedShares := uint64(0), atomic.NewUint64(0)
+		stakingTotalWeight, collectedShares := *atomic.NewUint64(0), *atomic.NewUint64(0)
 
 		// setup aggregators and reconstructor
 		stakingAggregator := &mockhotstuff.WeightedSignatureAggregator{}
@@ -469,7 +469,7 @@ func TestCombinedVoteProcessorV2_PropertyCreatingQCCorrectness(testifyT *testing
 		stakingAggregatorLock := &sync.Mutex{}
 
 		stakingAggregator.On("TotalWeight").Return(func() uint64 {
-			return stakingTotalWeight
+			return stakingTotalWeight.Load()
 		})
 		reconstructor.On("EnoughShares").Return(func() bool {
 			return collectedShares.Load() >= beaconSignersCount
@@ -497,7 +497,9 @@ func TestCombinedVoteProcessorV2_PropertyCreatingQCCorrectness(testifyT *testing
 			// check that aggregated signers are part of all votes signers
 			// due to concurrent processing it is possible that Aggregate will return less that we have actually aggregated
 			// but still enough to construct the QC
+			stakingAggregatorLock.Lock()
 			require.Subset(t, aggregatedStakingSigners, blockSigData.StakingSigners)
+			stakingAggregatorLock.Unlock()
 			require.Nil(t, blockSigData.RandomBeaconSigners)
 			require.Nil(t, blockSigData.AggregatedRandomBeaconSig)
 			require.GreaterOrEqual(t, uint64(len(blockSigData.StakingSigners)),
@@ -560,7 +562,7 @@ func TestCombinedVoteProcessorV2_PropertyCreatingQCCorrectness(testifyT *testing
 				signerID := args.Get(0).(flow.Identifier)
 				stakingAggregatorLock.Lock()
 				defer stakingAggregatorLock.Unlock()
-				stakingTotalWeight += sigWeight
+				stakingTotalWeight.Add(sigWeight)
 				aggregatedStakingSigners = append(aggregatedStakingSigners, signerID)
 			}).Return(uint64(0), nil).Maybe()
 		}
--- a/crypto/bls12381_utils.c
+++ b/crypto/bls12381_utils.c
@@ -99,7 +99,7 @@ ctx_t* relic_init_BLS12_381() {
     if (ALLOC != AUTO) return NULL;
 
     // initialize relic core with a new context
-    ctx_t* bls_ctx = (ctx_t*) malloc(sizeof(ctx_t));
+    ctx_t* bls_ctx = (ctx_t*) calloc(1, sizeof(ctx_t));
     core_set(bls_ctx);
     if (core_init() != RLC_OK) return NULL;
 
--- a/engine/execution/ingestion/engine_test.go
+++ b/engine/execution/ingestion/engine_test.go
@@ -66,6 +66,8 @@ type testingContext struct {
 	identity            *flow.Identity
 	broadcastedReceipts map[flow.Identifier]*flow.ExecutionReceipt
 	collectionRequester *module.MockRequester
+
+	mu *sync.Mutex
 }
 
 func runWithEngine(t *testing.T, f func(testingContext)) {
@@ -189,6 +191,8 @@ func runWithEngine(t *testing.T, f func(testingContext)) {
 		snapshot:            snapshot,
 		identity:            myIdentity,
 		broadcastedReceipts: make(map[flow.Identifier]*flow.ExecutionReceipt),
+
+		mu: &sync.Mutex{},
 	})
 
 	<-engine.Done()
@@ -202,7 +206,6 @@ func (ctx *testingContext) assertSuccessfulBlockComputation(
 	expectBroadcast bool,
 	newStateCommitment flow.StateCommitment,
 	computationResult *execution.ComputationResult) *protocol.Snapshot {
-
 	if computationResult == nil {
 		computationResult = executionUnittest.ComputationResultForBlockFixture(executableBlock)
 	}
@@ -253,12 +256,12 @@ func (ctx *testingContext) assertSuccessfulBlockComputation(
 
 	mocked.RunFn =
 		func(args mock.Arguments) {
-			//lock.Lock()
-			//defer lock.Unlock()
-
 			blockID := args[1].(*flow.Header).ID()
 			commit := args[2].(flow.StateCommitment)
+
+			ctx.mu.Lock()
 			commits[blockID] = commit
+			ctx.mu.Unlock()
 			onPersisted(blockID, commit)
 		}
 
@@ -303,7 +306,9 @@ func (ctx *testingContext) assertSuccessfulBlockComputation(
 				assert.True(ctx.t, valid)
 			}
 
+			ctx.mu.Lock()
 			ctx.broadcastedReceipts[receipt.ExecutionResult.BlockID] = receipt
+			ctx.mu.Unlock()
 		}).
 		Return(nil)
 
@@ -334,25 +339,20 @@ func (ctx *testingContext) stateCommitmentExist(blockID flow.Identifier, commit 
 }
 
 func (ctx *testingContext) mockStateCommitsWithMap(commits map[flow.Identifier]flow.StateCommitment) {
-	lock := sync.Mutex{}
+	mocked := ctx.executionState.On("StateCommitmentByBlockID", mock.Anything, mock.Anything)
+	// https://github.com/stretchr/testify/issues/350#issuecomment-570478958
+	mocked.RunFn = func(args mock.Arguments) {
 
-	{
-		mocked := ctx.executionState.On("StateCommitmentByBlockID", mock.Anything, mock.Anything)
-		// https://github.com/stretchr/testify/issues/350#issuecomment-570478958
-		mocked.RunFn = func(args mock.Arguments) {
-			// prevent concurrency issue
-			lock.Lock()
-			defer lock.Unlock()
-
-			blockID := args[1].(flow.Identifier)
-			commit, ok := commits[blockID]
-			if ok {
-				mocked.ReturnArguments = mock.Arguments{commit, nil}
-				return
-			}
-
-			mocked.ReturnArguments = mock.Arguments{flow.StateCommitment{}, storageerr.ErrNotFound}
+		blockID := args[1].(flow.Identifier)
+		ctx.mu.Lock()
+		commit, ok := commits[blockID]
+		ctx.mu.Unlock()
+		if ok {
+			mocked.ReturnArguments = mock.Arguments{commit, nil}
+			return
 		}
+
+		mocked.ReturnArguments = mock.Arguments{flow.StateCommitment{}, storageerr.ErrNotFound}
 	}
 }
 
--- a/engine/execution/state/delta/view.go
+++ b/engine/execution/state/delta/view.go
@@ -24,7 +24,7 @@ type View struct {
 	// TODO we can add a flag to disable capturing spockSecret
 	// for views other than collection views to improve performance
 	spockSecret       []byte
-	spockSecretLock   sync.Mutex
+	spockSecretLock   *sync.Mutex // using pointer instead, because using value would cause mock.Called to trigger race detector
 	spockSecretHasher hash.Hasher
 	readFunc          GetRegisterFunc
 }
@@ -48,6 +48,7 @@ func AlwaysEmptyGetRegisterFunc(owner, controller, key string) (flow.RegisterVal
 func NewView(readFunc GetRegisterFunc) *View {
 	return &View{
 		delta:             NewDelta(),
+		spockSecretLock:   &sync.Mutex{},
 		regTouchSet:       make(map[string]flow.RegisterID),
 		readFunc:          readFunc,
 		spockSecretHasher: hash.NewSHA3_256(),
--- a/engine/unit_test.go
+++ b/engine/unit_test.go
@@ -1,6 +1,7 @@
 package engine_test
 
 import (
+	"sync"
 	"testing"
 	"time"
 
@@ -19,13 +20,21 @@ func TestReadyDone(t *testing.T) {
 // Test that if a function is run by LaunchPeriodically and
 // takes longer than the interval, the next call will be blocked
 func TestLaunchPeriod(t *testing.T) {
+	lock := sync.Mutex{}
+
 	u := engine.NewUnit()
 	unittest.RequireCloseBefore(t, u.Ready(), time.Second, "ready did not close")
 	logs := make([]string, 0)
 	u.LaunchPeriodically(func() {
+		lock.Lock()
 		logs = append(logs, "running")
+		lock.Unlock()
+
 		time.Sleep(100 * time.Millisecond)
+
+		lock.Lock()
 		logs = append(logs, "finish")
+		lock.Unlock()
 	}, 50*time.Millisecond, 0)
 
 	// 100 * 3 is to ensure enough time for 3 periodic function to finish
@@ -34,11 +43,13 @@ func TestLaunchPeriod(t *testing.T) {
 	<-time.After((100*3 + 30) * time.Millisecond)
 
 	// This can pass
+	lock.Lock()
 	require.Equal(t, []string{
 		"running", "finish",
 		"running", "finish",
 		"running",
 	}, logs)
+	lock.Unlock()
 	unittest.RequireCloseBefore(t, u.Done(), time.Second, "done did not close")
 
 	require.Equal(t, []string{
--- a/module/lifecycle/lifecycle.go
+++ b/module/lifecycle/lifecycle.go
@@ -55,11 +55,12 @@ func (lm *LifecycleManager) OnStop(shutdownFns ...func()) {
 		return
 	}
 	lm.shutdownCommenced = true
+	waitForStartup := lm.startupCommenced
 	lm.stateTransition.Unlock()
 
 	close(lm.shutdownSignal)
 	go func() {
-		if lm.startupCommenced {
+		if waitForStartup {
 			<-lm.started
 			for _, fn := range shutdownFns {
 				fn()
--- a/module/lifecycle/lifecycle_test.go
+++ b/module/lifecycle/lifecycle_test.go
@@ -25,9 +25,10 @@ func (suite *LifecycleManagerSuite) SetupTest() {
 func (suite *LifecycleManagerSuite) TestConcurrentStart() {
 	var numStarts uint32
 
+	lm := suite.lm
 	for i := 0; i < 10; i++ {
 		go func() {
-			suite.lm.OnStart(func() {
+			lm.OnStart(func() {
 				atomic.AddUint32(&numStarts, 1)
 			})
 		}()
@@ -46,9 +47,10 @@ func (suite *LifecycleManagerSuite) TestConcurrentStop() {
 
 	var numStops uint32
 
+	lm := suite.lm
 	for i := 0; i < 10; i++ {
 		go func() {
-			suite.lm.OnStop(func() {
+			lm.OnStop(func() {
 				atomic.AddUint32(&numStops, 1)
 			})
 		}()
--- a/network/topology/topology_test.go
+++ b/network/topology/topology_test.go
@@ -1,3 +1,6 @@
+//go:build !race
+// +build !race
+
 package topology_test
 
 import (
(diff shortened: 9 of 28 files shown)


network/topology/topology_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build !race
+// +build !race
```
tarakby, line 2: It's not possible to run this test when `race` is enabled?

crypto/hash/xor_unaligned_asbytes_norace.go
```diff
@@ -0,0 +1,47 @@
...
+package hash
+
+import "unsafe"
+
+func (b *storageBuf) asBytes() *[maxRate]byte {
+	// re-using a trick from https://github.com/golang/go/blob/master/src/runtime/stubs.go#L178:
+	// to hide the input pointer from escape analysis and avoid
+	// an escape to to the heap. The 0 xor tricks the escape analysis tool
+	// to think "ptr" and "b" are not related.
+	ptr := uintptr(unsafe.Pointer(b)) ^ 0 // nolint:staticcheck
```
tarakby, line 45: I'm curious what kind of race conditions this line has caused, and if there are ways to avoid them other than removing the line?

tarakby (commented):
Great first PR 👌🏼  Thanks for contributing!

I focused on the changes under `crypto/` and I let the other reviewers look at the other parts.

crypto/hash/xor_unaligned_asbytes_norace.go
SaveTheRbtz: [race detector implicitly](https://go-review.googlesource.com/c/go/+/201783/) enables `checkptr` so [it fails with the following](https://go.dev/play/p/vc0QHhoOZrY?v=gotip):
```
$ go run -race main.go
fatal error: checkptr: pointer arithmetic result points to invalid allocation

goroutine 1 [running]:
runtime.throw({0x4edf95?, 0x4906d9?})
	/usr/lib/go/src/runtime/panic.go:992 +0x71 fp=0xc000119ec8 sp=0xc000119e98 pc=0x462951
runtime.checkptrArithmetic(0x0?, {0x0, 0x0, 0xc000119f70?})
	/usr/lib/go/src/runtime/checkptr.go:69 +0xaa fp=0xc000119ef8 sp=0xc000119ec8 pc=0x437d2a
main.(*storageBuf).asBytes(...)
```

It is possible to disable it with `-d=checkptr=0` but given that race detector enabled build is already quite slow, I considered that it makes sense to just disable this optimization there.

SaveTheRbtz: (Alternatively, i can add `//go:nocheckptr` to the function, which may be more obvious)

crypto/bls12381_utils.c
```diff
@@ -99,7 +99,7 @@ ctx_t* relic_init_BLS12_381() {
     if (ALLOC != AUTO) return NULL;
 
     // initialize relic core with a new context
-    ctx_t* bls_ctx = (ctx_t*) malloc(sizeof(ctx_t));
+    ctx_t* bls_ctx = (ctx_t*) calloc(1, sizeof(ctx_t));
```
SaveTheRbtz, line 102: @tarakby while here, also changed `malloc` to `calloc` since the latter zero-initializes the context.  ([given that default context](https://github.com/relic-toolkit/relic/blob/main/src/relic_core.c#L99) is `static` I would assume it expects that)  This fixes `msan` complaint:
```
==2776351==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x14a7aef in bls_verify_ep /home/rbtz/tmp/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls_core.c:145:9
    #1 0x14a7898 in bls_verify /home/rbtz/tmp/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls_core.c:328:12
    #2 0x14a1184 in _cgo_01218c0baead_Cfunc_bls_verify /tmp/go-build/cgo-gcc-prolog:75:11
    #3 0x527bef in runtime.asmcgocall.abi0 /home/rbtz/tmp/go/src/runtime/asm_amd64.s:765

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/rbtz/tmp/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls_core.c:145:9 in bls_verify_ep
Exiting
```

crypto/hash/xor_unaligned_asbytes_norace.go
SaveTheRbtz: Yeah, you are totally right: adding an annotation is way more straightforward:  https://github.com/onflow/flow-go/pull/2293/commits/1288c1dbcfa19c4249cdc0ae88078e7f4112b4d5

tarakby: I like the `//go:nocheckptr` idea. I tested and it works well 👌🏼. Thank you for the details!

crypto/bls12381_utils.c
tarakby: Interesting catch!

May I know how you  run the sanitizer?

From a quick look at the code, I think the warning is a sign of another racing issue in the code. I suspect it is not caught because the race Go tooling doesn't analyze things beyond Cgo.  The context  is a shared and not protected struct, initialized by the `init()` of the Go package and used by the function in logs `bls_verify_ep`. Is it possible that a thread is accessing the [freshly created context](https://github.com/SaveTheRbtz/flow-go/blob/combined/crypto/bls12381_utils.c#L103) before it gets [initialized](https://github.com/SaveTheRbtz/flow-go/blob/combined/crypto/bls12381_utils.c#L104) ?

I can dig further into that, and look generally at racing issues in the crypto tests, though I'll create another issue for that work.

SaveTheRbtz: I was using the following test:
```
$ CC=/usr/bin/clang CXX=/usr/bin/clang++ go test -c ./engine/execution/ingestion -tags=relic -msan
$ ./ingestion.test
==2776351==WARNING: MemorySanitizer: use-of-uninitialized-value
...
```
From what I can understand, it is complaining about: ```core_get()->code == RLC_OK```.  That said, I can't find the path where `code` can be uninitialized on `bls_verify`, unless it is called between these two lines:
```
    core_set(bls_ctx);
    if (core_init() != RLC_OK) return NULL;
```

SaveTheRbtz: The problem is that IIUC `lsInstance.init()` is called in the `init` function of the module, so it should not be able to race.... unless, maybe, there are some background threads created by the relic itself?  Or maybe one of flow's `init` functions launches a goroutine?  Though in these cases I would assume that bug should not be reliably triggered...

tarakby: > unless it is called between these two lines

Yes exactly, that's what I think. The fix you added (setting to zero) makes the second line check valid, which I think is a happy coincidence, but the racing issue might still be there.   

>is called in the init function of the module, so it should not be able to race

That was my understanding too, but not after your test :) (I do not think Relic is running anything in the backgroud)

I cannot run `-msan` because it's only supported on `linux`, so we can keep `calloc` for now, and I'll find a way to test later. 
Thank you again for catching this 🙏🏼

SaveTheRbtz: Seems like there is also a potential race condition there: 
```
$ CC=/usr/bin/clang CXX=/usr/bin/clang++ CGO_CFLAGS="-fsanitize=thread" CGO_LDFLAGS="-fsanitize=thread" ~/go/bin/dlv test --build-flags="-tags=relic" ./engine/execution/ingestion
...
WARNING: ThreadSanitizer: data race (pid=3757976)
  Read of size 8 at 0x000005738ae0 by thread T9:
    #0 fp_get_sign /home/rbtz/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls12381_utils.c:239:26 (debug.test+0x15ac0a3)
    #1 ep_write_bin_compact /home/rbtz/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls12381_utils.c:272:24 (debug.test+0x15abf9c)
    #2 bls_sign_ep /home/rbtz/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls_core.c:82:5 (debug.test+0x15adda8)
    #3 bls_sign /home/rbtz/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls_core.c:93:5 (debug.test+0x15adcdb)
    #4 _cgo_01218c0baead_Cfunc_bls_sign /tmp/go-build/cgo-tsan-prolog:33:2 (debug.test+0x15a9aec)
    #5 runtime.asmcgocall.abi0 /usr/lib/go/src/runtime/asm_amd64.s:821 (debug.test+0x5676a3)

  Previous write of size 8 at 0x000005738ae0 by thread T8:
    #0 precomputed_data_set /home/rbtz/go/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/bls12381_utils.c:54:14 (debug.test+0x15ab550)
    #1 _cgo_01218c0baead_Cfunc_precomputed_data_set /tmp/go-build/cgo-tsan-prolog:336:2 (debug.test+0x15aa855)
    #2 runtime.asmcgocall.abi0 /usr/lib/go/src/runtime/asm_amd64.s:821 (debug.test+0x5676a3)

  Location is global 'bls_prec' of size 8 at 0x000005738ae0 (debug.test+0x000005738ae0)```

Likely because `blsInstance.reInit()` is called unconditionally on each sign/verify.  Theoretically it should be safe, since this data should static after `init()` but if it is -- is there a reason to call it there?  If it is not static, then it can potentially race between different signers/verifiers.

File: engine/execution/state/delta/view.go [hunk 4]
```diff
@@ -48,6 +48,7 @@ func AlwaysEmptyGetRegisterFunc(owner, controller, key string) (flow.RegisterVal
 func NewView(readFunc GetRegisterFunc) *View {
 	return &View{
 		delta:             NewDelta(),
+		spockSecretLock:   &sync.Mutex{},
```
jordanschalm, line 51: What is it about this usage which makes the existing `sync.Mutex` field problematic? 

There are several other similar patterns in the code base, for example:
https://github.com/onflow/flow-go/blob/051bb593a2dc79e686b62a86dbf310d418d96375/engine/common/synchronization/request_heap.go#L14-L25

https://github.com/onflow/flow-go/blob/63d0eac7c47c7385adb008dde95c0639facea488/engine/consensus/approvals/chunk_collector.go#L11-L25

SaveTheRbtz: This is only a problem if you copy value of the structure (or read it in a different goroutine after the write).  In the view case it was read by the `assert.Equal`.  Generally, having non-pointer `Mutex` slightly is faster (no dereference, hence no potential for a TLB walk) but there are cases where it triggers race detector.

If you want I can move the cases you've mentioned to use the pointer to be more error-proof.

zhangchiqing: Can you provide the race detector command and the test case that triggers the race without this change, and won't trigger with this change?

engine/unit_test.go
```diff
@@ -19,13 +20,21 @@ func TestReadyDone(t *testing.T) {
...
 func TestLaunchPeriod(t *testing.T) {
+	lock := sync.Mutex{}
+
 	u := engine.NewUnit()
 	unittest.RequireCloseBefore(t, u.Ready(), time.Second, "ready did not close")
 	logs := make([]string, 0)
 	u.LaunchPeriodically(func() {
+		lock.Lock()
 		logs = append(logs, "running")
+		lock.Unlock()
```
zhangchiqing, line 31: I guess @SaveTheRbtz must have found this issue with the race detector, right?

However, when I ran the test with `-race` I ran into the following error. (Test passes when ran it without `-race`).

It's the same for both master and your branch. I wonder if you ran into the same issue?

```
▸ go test -race . -v -run=TestLaunchPeriod  -tags=relic -count=1
fatal error: checkptr: pointer arithmetic result points to invalid allocation

goroutine 1 [running, locked to thread]:
runtime.throw({0x4f5a7b4?, 0x406e879?})
        /usr/local/opt/go/libexec/src/runtime/panic.go:992 +0x71 fp=0xc00029f668 sp=0xc00029f638 pc=0x4037d71
runtime.checkptrArithmetic(0x25?, {0x0, 0x0, 0x57cf620?})
        /usr/local/opt/go/libexec/src/runtime/checkptr.go:69 +0xaa fp=0xc00029f698 sp=0xc00029f668 pc=0x4008e8a
github.com/onflow/flow-go/crypto/hash.(*storageBuf).asBytes(...)
        /Users/leo/gopath/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/hash/xor_unaligned.go:48
github.com/onflow/flow-go/crypto/hash.(*spongeState).appendBuf(...)
        /Users/leo/gopath/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/hash/keccak.go:122
github.com/onflow/flow-go/crypto/hash.(*spongeState).write(0xc00029f740, {0xc00003d2c0, 0x25, 0x25})
        /Users/leo/gopath/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/hash/keccak.go:160 +0x2a5 fp=0xc00029f718 sp=0xc00029f698 pc=0x43248e5
github.com/onflow/flow-go/crypto/hash.ComputeSHA3_256(0xc00029f8f0, {0xc00003d2c0, 0x25, 0x25})
        /Users/leo/gopath/pkg/mod/github.com/onflow/flow-go/crypto@v0.24.3/hash/sha3.go:48 +0xcd fp=0xc00029f8d0 sp=0xc00029f718 pc=0x432646d
github.com/onflow/flow-go/model/flow.MakeID({0x4e93320, 0xc000310f80})
        /Users/leo/flow-go/model/flow/identifier.go:119 +0x66 fp=0xc00029f920 sp=0xc00029f8d0 pc=0x45c3766
github.com/onflow/flow-go/model/cluster.Payload.Hash(...)
        /Users/leo/flow-go/model/cluster/payload.go:53
github.com/onflow/flow-go/state/cluster.init()
        /Users/leo/flow-go/state/cluster/root_block.go:18 +0x22d fp=0xc00029fac0 sp=0xc00029f920 pc=0x4bf6f0d
runtime.doInit(0x57752c0)
        /usr/local/opt/go/libexec/src/runtime/proc.go:6222 +0x126 fp=0xc00029fbf0 sp=0xc00029fac0 pc=0x4047786
runtime.doInit(0x5783740)
        /usr/local/opt/go/libexec/src/runtime/proc.go:6199 +0x71 fp=0xc00029fd20 sp=0xc00029fbf0 pc=0x40476d1
runtime.doInit(0x577ae80)
        /usr/local/opt/go/libexec/src/runtime/proc.go:6199 +0x71 fp=0xc00029fe50 sp=0xc00029fd20 pc=0x40476d1
runtime.doInit(0x5777800)
        /usr/local/opt/go/libexec/src/runtime/proc.go:6199 +0x71 fp=0xc00029ff80 sp=0xc00029fe50 pc=0x40476d1
runtime.main()
        /usr/local/opt/go/libexec/src/runtime/proc.go:233 +0x1d3 fp=0xc00029ffe0 sp=0xc00029ff80 pc=0x403a413
runtime.goexit()
        /usr/local/opt/go/libexec/src/runtime/asm_amd64.s:1571 +0x1 fp=0xc00029ffe8 sp=0xc00029ffe0 pc=0x406d141
FAIL    github.com/onflow/flow-go/engine        0.389s
FAIL

▸ go version
go version go1.18 darwin/amd64
```

SaveTheRbtz: You can try applying this diff https://github.com/onflow/flow-go/pull/2293/commits/1288c1dbcfa19c4249cdc0ae88078e7f4112b4d5 to the hasher inside `$GOPATH/pkg/mod/github.com/onflow/flow-go` (IIUC `crypto` package is installed separately)

engine/execution/state/delta/view.go
SaveTheRbtz: It fails down the line in the computation tests:
```CC=/usr/bin/clang CXX=/usr/bin/clang++ ~/tmp/go/bin/go test -tags=relic -race ./engine/execution/computation/computer```

with:
```
==================
WARNING: DATA RACE
Read at 0x00c000012800 by goroutine 26:
  reflect.typedmemmove()
      /home/rbtz/tmp/go/src/runtime/mbarrier.go:178 +0x0
  reflect.packEface()
      /home/rbtz/tmp/go/src/reflect/value.go:123 +0xae
  reflect.valueInterface()
      /home/rbtz/tmp/go/src/reflect/value.go:1394 +0x18a
  reflect.Value.Interface()
      /home/rbtz/tmp/go/src/reflect/value.go:1364 +0xc9
  fmt.(*pp).printValue()
      /home/rbtz/tmp/go/src/fmt/print.go:722 +0xca
  fmt.(*pp).printValue()
      /home/rbtz/tmp/go/src/fmt/print.go:876 +0x12be
  fmt.(*pp).printArg()
      /home/rbtz/tmp/go/src/fmt/print.go:712 +0xdf4
  fmt.(*pp).doPrintf()
      /home/rbtz/tmp/go/src/fmt/print.go:1122 +0xe38
  fmt.Sprintf()
      /home/rbtz/tmp/go/src/fmt/print.go:219 +0x67
  github.com/stretchr/testify/mock.Arguments.Diff()
      /home/rbtz/go/pkg/mod/github.com/stretchr/testify@v1.7.1-0.20210824115523-ab6dc3262822/mock/mock.go:803 +0x175
  github.com/stretchr/testify/mock.(*Mock).findExpectedCall()
      /home/rbtz/go/pkg/mod/github.com/stretchr/testify@v1.7.1-0.20210824115523-ab6dc3262822/mock/mock.go:295 +0x144
  github.com/stretchr/testify/mock.(*Mock).MethodCalled()
      /home/rbtz/go/pkg/mod/github.com/stretchr/testify@v1.7.1-0.20210824115523-ab6dc3262822/mock/mock.go:401 +0xb9
  github.com/stretchr/testify/mock.(*Mock).Called()
      /home/rbtz/go/pkg/mod/github.com/stretchr/testify@v1.7.1-0.20210824115523-ab6dc3262822/mock/mock.go:391 +0x1a4
  github.com/onflow/flow-go/engine/execution/computation/computer/mock.(*ViewCommitter).CommitView()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/mock/view_committer.go:21 +0x1a4
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockCommitter).Run()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:434 +0x195
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockComputer).executeBlock.func3()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:178 +0x39

Previous write at 0x00c000012800 by goroutine 24:
  sync/atomic.AddInt32()
      /home/rbtz/tmp/go/src/runtime/race_amd64.s:305 +0xb
  sync/atomic.AddInt32()
      <autogenerated>:1 +0x1a
  github.com/onflow/flow-go/engine/execution/state/delta.(*View).SpockSecret()
      /home/rbtz/porn/flow-go/engine/execution/state/delta/view.go:255 +0xda
  github.com/onflow/flow-go/engine/execution/state/delta.(*View).MergeView()
      /home/rbtz/porn/flow-go/engine/execution/state/delta/view.go:218 +0x2bb
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockComputer).executeBlock()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:197 +0x11d8
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockComputer).ExecuteBlock()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:103 +0x48f
  github.com/onflow/flow-go/engine/execution/computation/computer_test.TestBlockExecutor_ExecuteBlock.func1()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer_test.go:84 +0x11ea
  testing.tRunner()
      /home/rbtz/tmp/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /home/rbtz/tmp/go/src/testing/testing.go:1306 +0x47

Goroutine 26 (running) created at:
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockComputer).executeBlock()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:177 +0xe64
  github.com/onflow/flow-go/engine/execution/computation/computer.(*blockComputer).ExecuteBlock()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer.go:103 +0x48f
  github.com/onflow/flow-go/engine/execution/computation/computer_test.TestBlockExecutor_ExecuteBlock.func1()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer_test.go:84 +0x11ea
  testing.tRunner()
      /home/rbtz/tmp/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /home/rbtz/tmp/go/src/testing/testing.go:1306 +0x47

Goroutine 24 (running) created at:
  testing.(*T).Run()
      /home/rbtz/tmp/go/src/testing/testing.go:1306 +0x726
  github.com/onflow/flow-go/engine/execution/computation/computer_test.TestBlockExecutor_ExecuteBlock()
      /home/rbtz/porn/flow-go/engine/execution/computation/computer/computer_test.go:45 +0xcb
  testing.tRunner()
      /home/rbtz/tmp/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /home/rbtz/tmp/go/src/testing/testing.go:1306 +0x47
```

File: engine/unit_test.go (hunk 5 above)
Comment (zhangchiqing) [reply engine/unit_test.go:31]: I'm still running into this error even with [the diff](https://github.com/onflow/flow-go/commit/1288c1dbcfa19c4249cdc0ae88078e7f4112b4d5) on top of latest master or your branch.

Tried on MacOS with golang v1.17 and 1.18.

Comment (SaveTheRbtz) [reply engine/unit_test.go:31]: Interesting, do you see it also applied to files in `$GOPATH/pkg/mod/github.com/onflow/flow-go`?

**Review by zhangchiqing — APPROVED** [review]
Great work @SaveTheRbtz !

I was able to reproduce the race conditions and confirm they are fixed on your branch. Thanks for fixing so many races issue 👍

File: module/lifecycle/lifecycle.go [hunk 6]
```diff
@@ -49,17 +49,22 @@ func (lm *LifecycleManager) OnStart(startupFns ...func()) {
...
+
 	lm.stateTransition.Lock()
 	if lm.shutdownCommenced {
 		lm.stateTransition.Unlock()
 		return
 	}
 	lm.shutdownCommenced = true
+	if lm.startupCommenced {
+		waitForStartup = true
+	}
```
AlexHentschel, line 62: Good catch!

wouldn't it be more compact to write:
```suggestion
	waitForStartup := lm.startupCommenced
```
?

AlexHentschel (approved):
Thanks for the many concurrency improvements.

module/lifecycle/lifecycle_test.go
```diff
@@ -25,9 +25,10 @@ func (suite *LifecycleManagerSuite) SetupTest() {
 func (suite *LifecycleManagerSuite) TestConcurrentStart() {
 	var numStarts uint32
 
+	lm := suite.lm
 	for i := 0; i < 10; i++ {
 		go func() {
-			suite.lm.OnStart(func() {
+			lm.OnStart(func() {
```
AlexHentschel, line 31: Just wondering, is this an algorithmic change?

zhangchiqing: I think this happens because `suite.lm` is shared between multiple test cases. 
The next test case will run the same`SetupTest` method, which will update `suite.lm`. However, that's a concurrent write, which is a race to a concurrent read here from `suite.lm.OnStart`.

The fix was to create a new value `lm`. So that the concurrent write from other test cases to `suite.lm` will not be a race to the read to `lm`.

In other words, it's purely a test case bug rather than an algorithmic change.

yhassanzadeh13:
> @yhassanzadeh13 A friendly ping for possible solutions of the HeroCache races:
> 
> * Making Counter and Timestamp atomic: [SaveTheRbtz@d1cc406](https://github.com/SaveTheRbtz/flow-go/commit/d1cc40686c5e51a4728f07305bcfea0fb2a43936)
> * Removing `logTelemetry` methods from the "read-only" methods protected only by `RLock`: [SaveTheRbtz@6b0dcbd](https://github.com/SaveTheRbtz/flow-go/commit/6b0dcbdbefaaa1cb2d652b1bdbd9c5505ca298e9)
> 
> Or maybe there is a better solution there?

Thanks @SaveTheRbtz for the suggestions. I would recommend going with making `interactionCounter` and `lastTelemetryDump` atomic as you have already done in https://github.com/SaveTheRbtz/flow-go/commit/d1cc40686c5e51a4728f07305bcfea0fb2a43936. HeroCache is a young project and we still need insights into its read and write paths. Hence, removing `logTelemetry` from read-only methods may not be a viable solution.

consensus/hotstuff/votecollector/combined_vote_processor_v2_test.go
```diff
@@ -451,7 +451,7 @@ func TestCombinedVoteProcessorV2_PropertyCreatingQCCorrectness(testifyT *testing
 			"staking signers: %v, beacon signers: %v\n\t"+
 			"required weight: %v", stakingSignersCount, beaconSignersCount, minRequiredWeight)
 
-		stakingTotalWeight, collectedShares := uint64(0), atomic.NewUint64(0)
+		stakingTotalWeight, collectedShares := atomic.Uint64{}, atomic.Uint64{}
```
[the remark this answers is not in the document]
SaveTheRbtz: yeah, I've tried to not introduce any additional pointer de-references.. I can do something like
```
stakingTotalWeight, collectedShares := *atomic.NewUint64(0), *atomic.NewUint64(0)
```
This should have a good balance between expressiveness and runtime perf (not that it matters too much in a test though.)

crypto/bls12381_utils.c
tarakby: Agree, that is another race condition that I'm aware of for some time : there is a single static relic context, initialized with a BLS-friendly context during `init()`. The crypto library today uses only one setup (context) but could be evolved to use more contexts in the future. Whenever a BLS related function is called, no assumptions about the Relic context value are made, so `reinit()` is called to set up the BLS context each time. This is obviously not thread safe, but since the lib for now is based only on one context, the same context value is being assigned over and over, and the race condition doesn't cause errors. 

A temporary solution would be to not call `reinit()` at all because the call is obsolete (I can do this in another PR). If more contexts are supported in the future, then a thread safe solution needs to be implemented.

tarakby: I created this [issue](https://github.com/dapperlabs/flow-internal/issues/1699)   to address these two race conditions. You'll get access to the repo soon :) Thanks for reporting these points.

tarakby (approved):

SaveTheRbtz:
rebased over the latest changes.  What are the next steps here?

tarakby:
@SaveTheRbtz feel free to merge the PR once Github allows you to do so (required tests are passing + branch is up to date with master). FYI, you can use the merging bot by commenting `bors merge` on the PR, so that it is merged automatically once tests have passed and you don't have to wait for them.

engine/execution/ingestion/engine_test.go
```diff
@@ -66,6 +66,8 @@ type testingContext struct {
 	identity            *flow.Identity
 	broadcastedReceipts map[flow.Identifier]*flow.ExecutionReceipt
 	collectionRequester *module.MockRequester
+
+	mu sync.Mutex
```
Kay-Zee, line 70: Go Vet is mad because adding this as a value rather than a pointer means when it gets passed around the lock gets copied

I believe what you're intending to do would still work with this as a pointer, we'd just have to make sure to initialize the lock

cmd/util/ledger/migrations/storage_used_update_migration.go
```diff
@@ -13,6 +14,7 @@ import (
 	"github.com/onflow/flow-go/ledger"
 	"github.com/onflow/flow-go/ledger/common/utils"
 	"github.com/onflow/flow-go/model/flow"
+	"golang.org/x/sync/errgroup"
```
Kay-Zee, line 17: I think our lint expects a specific order for the imports, something like:

```
import (
    "context" // standard lib

    "golang.org/x/sync/errgroup" // external deps

    "github.com/onflow/flow-go/model/flow" // internal deps
)
```

I think this file may have been incorrect previously, but touching it caused the lint to flag it, so you'll have to re-arrange to get it to pass the linter.

Kay-Zee (commented):
@SaveTheRbtz The linter isn't happy right now, put in a few things I noticed that may help address the linter's complaints

SaveTheRbtz:
@Kay-Zee thanks. fixed linter warnings and did the final:
```
git rebase --autosquash -i master
```

after tests pass, let me try this bors thingy

SaveTheRbtz:
> lock Permission denied
> 
> Existing reviewers: [click here to make SaveTheRbtz a reviewer](https://app.bors.tech/repositories/35927/add-reviewer/SaveTheRbtz)

@tarakby seems like I may need your help there since my first day is technically next week, hence no perms =)

zhangchiqing:
bors merge

PR #2294, opened 2022-04-12
Add config to limit caching of new proposals while syncing

jordanschalm:
Adds a new config which will cause consensus followers and participants to drop new proposals if the proposal is sufficiently ahead of the node's locally finalized state. This is a simple way to prevent memory exhaustion when a node is synchronizing from a very old state.

### DIFF
--- a/cmd/access/node_builder/access_node_builder.go
+++ b/cmd/access/node_builder/access_node_builder.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/pflag"
 
 	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/module/compliance"
 
 	"github.com/onflow/flow-go/cmd"
 	"github.com/onflow/flow-go/consensus"
@@ -90,6 +91,7 @@ type AccessNodeConfig struct {
 	apiRatelimits                map[string]int
 	apiBurstlimits               map[string]int
 	rpcConf                      rpc.Config
+	followerConfig               compliance.Config
 	ExecutionNodeAddress         string // deprecated
 	HistoricalAccessRPCs         []access.AccessAPIClient
 	logTxTimeToFinalized         bool
@@ -127,6 +129,7 @@ func DefaultAccessNodeConfig() *AccessNodeConfig {
 			PreferredExecutionNodeIDs: nil,
 			FixedExecutionNodeIDs:     nil,
 		},
+		followerConfig:               compliance.DefaultConfig(),
 		ExecutionNodeAddress:         "localhost:9000",
 		logTxTimeToFinalized:         false,
 		logTxTimeToExecuted:          false,
@@ -309,6 +312,7 @@ func (builder *FlowAccessNodeBuilder) buildFollowerEngine() *FlowAccessNodeBuild
 			builder.FollowerCore,
 			builder.SyncCore,
 			node.Tracer,
+			compliance.WithSkipNewProposalsThreshold(builder.followerConfig.SkipNewProposalsThreshold),
 		)
 		if err != nil {
 			return nil, fmt.Errorf("could not create follower engine: %w", err)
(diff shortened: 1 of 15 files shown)


engine/common/follower/engine.go
```diff
@@ -215,6 +224,18 @@ func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.Bl
 		return fmt.Errorf("could not check proposal: %w", err)
 	}
 
+	// ignore proposals which are too far ahead of our local finalized state
```
zhangchiqing, line 227:
```suggestion
	// ignore proposals which are too far ahead of our local finalized state
	// instead, rely on sync engine to catch up finalization more effectively, and avoid
	//  large subtree of blocks to be cached. 
```

zhangchiqing (approved):

engine/common/follower/engine.go
jordanschalm: 👍 updated https://github.com/onflow/flow-go/pull/2294/commits/5d1c56881355211d1502c12de749fb832ae99ac8

cmd/access/node_builder/access_node_builder.go
```diff
@@ -442,6 +446,7 @@ func (builder *FlowAccessNodeBuilder) extraFlags() {
 		flags.UintVar(&builder.rpcConf.MaxHeightRange, "rpc-max-height-range", defaultConfig.rpcConf.MaxHeightRange, "maximum size for height range requests")
 		flags.StringSliceVar(&builder.rpcConf.PreferredExecutionNodeIDs, "preferred-execution-node-ids", defaultConfig.rpcConf.PreferredExecutionNodeIDs, "comma separated list of execution nodes ids to choose from when making an upstream call e.g. b4a4dbdcd443d...,fb386a6a... etc.")
 		flags.StringSliceVar(&builder.rpcConf.FixedExecutionNodeIDs, "fixed-execution-node-ids", defaultConfig.rpcConf.FixedExecutionNodeIDs, "comma separated list of execution nodes ids to choose from when making an upstream call if no matching preferred execution id is found e.g. b4a4dbdcd443d...,fb386a6a... etc.")
+		flags.Uint64Var(&builder.followerConfig.SkipNewProposalsThreshold, "follower-skip-proposals-threshold", defaultConfig.followerConfig.SkipNewProposalsThreshold, "threshold at which new proposals are discarded rather than cached, if their height is this much above local finalized height")
```
smnzhu, line 449: Instead of adding this new flag for all node types, could we just add it as a shared flag in scaffold.go?

engine/collection/compliance/core.go
```diff
@@ -108,6 +117,20 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		return fmt.Errorf("could not check proposal: %w", err)
 	}
 
+	// ignore proposals which are too far ahead of our local finalized state
+	// instead, rely on sync engine to catch up finalization more effectively, and avoid
+	// large subtree of blocks to be cached.
+	final, err := c.state.Final().Head()
+	if err != nil {
+		return fmt.Errorf("could not get latest finalized header: %w", err)
+	}
```
smnzhu, line 126: I would suggest using [FinalizedHeaderCache](https://github.com/onflow/flow-go/blob/a2c545d78087b26f97ccea09022f61c3b124d5bb/engine/common/synchronization/finalized_snapshot.go#L32) instead, so we can avoid accessing the state every time.

engine/common/follower/engine.go
```diff
@@ -215,6 +224,20 @@ func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.Bl
...
+	// instead, rely on sync engine to catch up finalization more effectively, and avoid
+	// large subtree of blocks to be cached.
+	final, err := e.state.Final().Head()
+	if err != nil {
+		return fmt.Errorf("could not get latest finalized header: %w", err)
+	}
```
smnzhu, line 233: I would suggest using [FinalizedHeaderCache](https://github.com/onflow/flow-go/blob/a2c545d78087b26f97ccea09022f61c3b124d5bb/engine/common/synchronization/finalized_snapshot.go#L32) instead, so we can avoid accessing the state every time.

engine/consensus/compliance/core.go
```diff
@@ -137,6 +146,20 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
 		return fmt.Errorf("could not check proposal: %w", err)
 	}
 
+	// ignore proposals which are too far ahead of our local finalized state
+	// instead, rely on sync engine to catch up finalization more effectively, and avoid
+	// large subtree of blocks to be cached.
+	final, err := c.state.Final().Head()
+	if err != nil {
+		return fmt.Errorf("could not get latest finalized header: %w", err)
+	}
```
smnzhu, line 155: I would suggest using [FinalizedHeaderCache](https://github.com/onflow/flow-go/blob/a2c545d78087b26f97ccea09022f61c3b124d5bb/engine/common/synchronization/finalized_snapshot.go#L32) instead, so we can avoid accessing the state every time.

cmd/access/node_builder/access_node_builder.go
jordanschalm: I didn't do this because we're dealing with 3 different engines across the different node roles, but we could use the same flag name for `compliance` on `consensus` nodes and `follower` on the others.

engine/collection/compliance/core.go
jordanschalm: I don't think the `FinalizedHeaderCache` is a great way to cache this data, which is why I avoided using it here. Some reasoning below:

We already have resource-specific caches, including for [headers](https://github.com/onflow/flow-go/blob/2366de82cbc5e6c68eadbadbe246784c16fc46aa/storage/badger/headers.go#L22). We are all but guaranteed to always find the latest finalized header in this cache without touching the database.

In addition to retrieving the header, `state.Final` will query for the `finalized_height` pointer and the corresponding `height->block_id` lookup from the database. These lookups would benefit from something like `FinalizedHeaderCache`. However, if we are going to cache these values, why not cache them in the protocol state itself? Then all protocol state users get the performance improvement, without needing to incorporate another component.

smnzhu: Ah I haven't looked into what has been implemented for caching in the protocol state itself. It probably makes more sense to do that.

smnzhu (approved):

engine/collection/compliance/core.go
jordanschalm: Added an issue for it: https://github.com/dapperlabs/flow-go/issues/6224

jordanschalm:
bors merge

PR #2300, opened 2022-04-13
Feature/observer poc

szegedim:
This pull request is a Proof of Concept of an observer service/node implementation.
1. It refactors code moving unstaked parts into the follower library and it creates an observer image.
2. Observers start up with an access node bootstrapped that gets certain messages over 
3. The terms access node, observer service, follower library, and Flow API Service are introduced.
4. Limitation and a future work is to detach the observer from the private network replaced by a follower on a separate public bound address.

Please review.


szegedim:
This pull request is superseded by https://github.com/onflow/flow-go/pull/2309

PR #2305, opened 2022-04-13
Memory weight assignment from contract

janezpodhostnik:
Based on core contracts https://github.com/onflow/flow-core-contracts/pull/285


dsainati1 (approved):
thanks for doing this!

fvm/meter/weighted/meter.go
```diff
@@ -199,9 +199,7 @@ func WithComputationWeights(weights ExecutionEffortWeights) WeightedMeterOptions
 func WithMemoryWeights(weights ExecutionMemoryWeights) WeightedMeterOptions {
 	if int(common.MemoryKindLast)-len(weights) != 1 {
 		return func(m *Meter) {
-			for kind, weight := range weights {
```
janezpodhostnik, line 202: @dsainati1 This was causing the tests to fail, as the the memory weights are assigned like this:
```
		memoryWeights:          DefaultMemoryWeights,
```

so this code caused changes to the `DefaultMemoryWeights`

engine/execution/computation/execution_verification_test.go
```diff
@@ -270,8 +270,8 @@ func TestTransactionFeeDeduction(t *testing.T) {
 		checkResult   func(t *testing.T, cr *execution.ComputationResult)
 	}
 
-	txFees := uint64(1_0000)
-	fundingAmount := uint64(1_0000_0000)
+	txFees := uint64(1_000)
```
janezpodhostnik, line 273: This was actually set to the wrong number.

fvm/context.go
```diff
@@ -60,9 +60,9 @@ func newContext(ctx Context, opts ...Option) Context {
 const AccountKeyWeightThreshold = 1000
 
 const (
-	DefaultComputationLimit             = 100_000    // 100K
-	DefaultMemoryLimit                  = 10_000_000 // 10G
-	DefaultEventCollectionByteSizeLimit = 256_000    // 256KB
+	DefaultComputationLimit             = 100_000        // 100K
+	DefaultMemoryLimit                  = 10_000_000_000 // 10G
```
janezpodhostnik, line 64: comment was 10G which fits better in this case.

dsainati1 (approved):

PR #2308, opened 2022-04-13
[BFT Testing] Integrating  BFT testing framework with testnet

yhassanzadeh13:
Enables overriding enqueued components in the scaffold so that we can override default conduit factory with a corrupted one and build corrupted images at compile time.

It also creates corrupted `cmd`s for ENs and VNs.

This PR lays the groundwork for running BFT integration tests by updating the way in which the testnet network is brought up to allow for Corrupted Conduit Framework to be used for integration tests

Another PR will deal with the actual running of the integration tests.


network/p2p/conduit/conduit.go
```diff
@@ -5,18 +5,33 @@ import (
...
+	"github.com/onflow/flow-go/module/irrecoverable"
 	"github.com/onflow/flow-go/network"
 )
 
 // DefaultConduitFactory is a wrapper around the network Adapter.
 // It directly passes the incoming messages to the corresponding methods of the
 // network Adapter.
 type DefaultConduitFactory struct {
+	component.Component
+	cm      *component.ComponentManager
```
jordanschalm, line 18:
```suggestion
	*component.ComponentManager
```
Directly embedding `ComponentManager` will make `DefaultConduitFactory` implement `component.Component`  without having two fields for the same object -- is that why you have `component.Component` embedded currently?

network/p2p/conduit/conduit.go
```diff
@@ -5,18 +5,33 @@ import (
...
 	adapter network.Adapter
 }
 
 func NewDefaultConduitFactory() *DefaultConduitFactory {
-	return &DefaultConduitFactory{}
+	d := &DefaultConduitFactory{}
+	cm := component.NewComponentManagerBuilder().
+		AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
```
jordanschalm, line 25: Why do we need this worker?

cmd/verification_build.go
```diff
@@ -0,0 +1,396 @@
+package cmd
```
jordanschalm, line 1: Should this file be called `verification_builder.go`?

jordanschalm (approved):
Nice work 🎸 

A few minor comments, but the approach generally looks good to me.

cmd/verification_build.go
gomisha: renamed https://github.com/onflow/flow-go/pull/2308/commits/5f6b80addf2b674dbc0f6c6e9a41dc9a277fac26

insecure/cmd/corrupted_builder.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"net"
+
+	"github.com/onflow/flow-go/cmd"
+	"github.com/onflow/flow-go/insecure/corruptible"
+	"github.com/onflow/flow-go/module"
+)
+
+// CorruptibleConduitFactoryPort is the port number that gRPC server of the conduit factory of corrupted nodes is listening on.
+const CorruptibleConduitFactoryPort = "4300"
```
smnzhu, line 13: What is this different port used for? Can you maybe explain at a high level how the corruptible conduit system works?

I was not a reviewer on the previous PR's, but based on what I'm seeing in this PR, here is my approximate understanding:
Basically, we have a gRPC server running on node A that allows us to manually send messages to it, which then get delivered to the node via the corrupted conduit as if it had actually arrived from another node B on the network?
Are these "corrupted" messages interleaved with the "real" messages received from node B? Is there a mechanism that enables us to drop certain "real" messages sent by node B?

Btw, not sure if you have seen this, but libp2p has developed [TestGround](https://github.com/testground/testground) which they used to do testing for the libp2p implementation. It may be useful to look into this for inspiration.

smnzhu (commented):
There are lot of refactoring changes here so I haven't gone line by line into the details, but I've just raised a few questions / concerns on things that caught my attention. I don't see any glaring problems at the moment.

network/p2p/conduit/conduit.go
gomisha: added clarification https://github.com/onflow/flow-go/pull/2308/commits/a42ae4221c601f2e6f44880991b5679a20788a5b

insecure/cmd/corrupted_builder.go
gomisha: The port is used for communicating between the "orchestrator" and each corrupted node, on their own private network. To find a detailed explanation of the design search for "Corruptible Conduits Framework" in Nortion.

Corrupted nodes communicate on their own "attack network" that's outside of the regular protocol libp2p network.

Corrupted messages are not interleaved real messages. Corrupted messages are delivered to the orchestrator which then decides what to do with them - drop them, alter them, or pass them through unaltered. The orchestrator sends the messages it receives back to the conduit which then sends them on the network. Each corrupted conduit is "dumb" and only sends its messages to the orchestrator and then sends whatever the orchestrator sends it back. If the orchestrator wants to drop a message, it doesn't send anything back to the corrupted conduit and the message never gets sent.

cmd/scaffold.go
```diff
@@ -142,8 +147,8 @@ func (fnb *FlowNodeBuilder) BaseFlags() {
 		"incoming message cache size at networking layer")
 	fnb.flags.UintVar(&fnb.BaseConfig.guaranteesCacheSize, "guarantees-cache-size", bstorage.DefaultCacheSize, "collection guarantees cache size")
 	fnb.flags.UintVar(&fnb.BaseConfig.receiptsCacheSize, "receipts-cache-size", bstorage.DefaultCacheSize, "receipts cache size")
-	fnb.flags.StringVar(&fnb.BaseConfig.topologyProtocolName, "topology", defaultConfig.topologyProtocolName, "networking overlay topology")
-	fnb.flags.Float64Var(&fnb.BaseConfig.topologyEdgeProbability, "topology-edge-probability", defaultConfig.topologyEdgeProbability,
+	fnb.flags.StringVar(&fnb.BaseConfig.TopologyProtocolName, "topology", defaultConfig.TopologyProtocolName, "networking overlay topology")
+	fnb.flags.Float64Var(&fnb.BaseConfig.TopologyEdgeProbability, "topology-edge-probability", defaultConfig.TopologyEdgeProbability,
```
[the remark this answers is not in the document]
yhassanzadeh13: yes

network/p2p/conduit/conduit.go
yhassanzadeh13: Fixed in https://github.com/onflow/flow-go/pull/2308/commits/43fb4c956c72a35ef26aeaf9ce3801bfa439b29b

smnzhu (approved):

PR #2309, opened 2022-04-14
Feature/observer poc 0414

szegedim:
This pull request is a Proof of Concept of an observer service/node implementation.

1. It extends https://github.com/onflow/flow-go/pull/2300 with Github specific changelist comments
2. It refactors code moving unstaked parts into the follower library and it creates an observer image.
3. Observers start up with an access node bootstrapped that sends certain messages over to the access node.
4. The terms access node, observer service, follower library, and Flow API Service are introduced.
5. Limitation and a future work is to detach the observer from the private network replaced by a follower on a separate public bound address.
Please review.

PR #2310, opened 2022-04-14
Feature/observer poc 0414a

szegedim:
This pull request is a Proof of Concept of an observer service implementation.

. It extends https://github.com/onflow/flow-go/pull/2309 with Github specific changelist comments.
. It refactors code moving unstaked access node parts into the follower library, and it creates code for an observer service image.
. Observers start up with an access node bootstrapped that sends certain messages over to the access node.
. The terms access node, observer service, follower library, and Flow API Service are introduced.
. Limitations: We need to detach the observer from the private network replaced by a follower on a separate public bound address.
Please review.


szegedim:
Please refer to https://github.com/onflow/flow-go/pull/2320
We will embrace smaller feature complete pull requests instead of big ones.

PR #2317, opened 2022-04-14
Khalil/1689 Add authorized sender pubsub topic validator

kc1116:
In order to increase resiliency to byzantine messages  this PR adds a new topic validator to the default pubsub topic validators for non-public channels. Leveraging a static mapping of message_type => list of authorized roles and the fact that the first byte of each message payload contains the message type the authorized sender validator will reject messages that; 

1.  have an unknown message type
2. role of sender is not in the list of authorized roles for the message type
3. sender has a weight <= 0
4. sender is an ejected node

[Notion page with related docs](https://www.notion.so/dapperlabs/Resilience-against-Impersonation-and-Masquerade-Attacks-61189d46e5374c2b9af36edacf71435f)


network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("sender is not authorized to send this message type")
+	}
+
+	return nil
+}
+
+// isActiveNode checks that the node has a weight > 0 and is not ejected
+func isActiveNode(nodeID flow.Identifier, weight uint64, ejected bool) error {
+	if weight <= 0 {
+		return fmt.Errorf("node %s has an invalid weight of %d is not an active node", nodeID, weight)
```
jordanschalm, line 143: ⚠️  Nodes with zero weight need to be able to participate in network communications, for a limited number of messages. We should only discard messages at the network layer on the sole basis of a node's identity, if a node is ejected. In particular, this is needed for the DKG and for block synchronization for new nodes. More details [here](https://www.notion.so/dapperlabs/Epoch-Networking-0c7b7a6ecd904a9fb8c023069488fe51).

In the interest of keeping with the approach here, my suggestion would be to implement these restrictions alongside the role-based restrictions added in this PR. For each message code, in addition to an allow-list of sender roles, we can include a tag for whether the message can be sent be a zero-weight node.

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,151 @@
...
+		Logger()
+
+	return func(ctx context.Context, from peer.ID, msg *message.Message) pubsub.ValidationResult {
+		identity, ok := getIdentity(from)
+		if !ok {
+			log.Warn().Str("peer_id", from.String()).Msg("could not get identity of sender")
+			return pubsub.ValidationReject
+		}
+
+		msgType := msg.Payload[0]
```
jordanschalm, line 95: We could include a comment here as well noting the dependency on the encoding being used.

cmd/scaffold.go
```diff
@@ -237,6 +237,8 @@ func (fnb *FlowNodeBuilder) EnqueueResolver() {
 
 func (fnb *FlowNodeBuilder) EnqueueNetworkInit() {
 	fnb.Component("network", func(node *NodeConfig) (module.ReadyDoneAware, error) {
+		//NOTE: if the codec used in the network component is ever changed any code relying on
+		// the message format specific to the codec must be updated. i.e: the AuthorizedSenderValidator.
```
jordanschalm, line 241: 👍

network/codec/cbor/codec.go
```diff
@@ -36,7 +36,7 @@ func (c *Codec) NewDecoder(r io.Reader) network.Decoder {
 	return &Decoder{dec: dec}
 }
 
-// Given a Golang interface 'v', return a []byte 'envelope'.
+// Encode Given a Golang interface 'v', return a []byte 'envelope'.
```
jordanschalm, line 39:
```suggestion
// Encode will, given a Golang interface 'v', return a []byte 'envelope'.
```

network/codec/cbor/codec.go
```diff
@@ -74,9 +74,9 @@ func (c *Codec) Encode(v interface{}) ([]byte, error) {
 	return dataBytes, nil
 }
 
-// Given a []byte 'envelope', eturn a Golang interface 'v'.
+// Decode Given a []byte 'envelope', return a Golang interface 'v'.
```
jordanschalm, line 77:
```suggestion
// Decode will, given a []byte 'envelope', return a Golang interface 'v'.
```

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	if shouldReceive {
+		// assert we can receive the next message
+		msg, err := sub.Next(ctx)
+		require.NoError(t, err)
+		require.Equal(t, expectedData, msg.Data)
+	} else {
+		wg.Add(1)
+		go func() {
+			_, err := sub.Next(ctx)
+			require.Error(t, err)
```
jordanschalm, line 438: Is the assumption that this is getting an error saying "there are no further messages"? If possible, we could check the error type / sentinel to confirm.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// assert that the nodes are connected as expected
+	require.Eventually(t, func() bool {
+		return len(sn1.pubSub.ListPeers(topic.String())) > 0 &&
+			len(sn2.pubSub.ListPeers(topic.String())) > 0 &&
+			len(an1.pubSub.ListPeers(topic.String())) > 0
+	}, 3*time.Second, 100*time.Millisecond)
+
+	timedCtx, cancel5s := context.WithTimeout(context.Background(), 1000*time.Second)
```
jordanschalm, line 184: If `cancel5s` means "5 seconds", it's 1000 seconds currently.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	translator, err := NewFixedTableIdentityTranslator(ids)
+	require.NoError(t, err)
+
+	// setup hooked logger
+	hookCalls := 0
+	hook := zerolog.HookFunc(func(e *zerolog.Event, level zerolog.Level, message string) {
+		if level == zerolog.WarnLevel {
+			hookCalls++
+		}
+	})
```
jordanschalm, line 152:
```suggestion
	hookCalls := atomic.NewUint32(0)
	hook := zerolog.HookFunc(func(e *zerolog.Event, level zerolog.Level, message string) {
		if level == zerolog.WarnLevel {
			hookCalls.Inc()
		}
	})
```

We had [a PR](https://github.com/onflow/flow-go/pull/2293) recently that pointed out a few cases where our tests had races. I think we would be vulnerable to a race here as well, because `authorizedSenderValidator` will be called by 3 nodes concurrently. This suggestion is taken from one of the fixes they contributed.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+
+	// sn2 also does not receive the message via gossip from the sn1 (event after the 1 second hearbeat)
+	timedCtx, cancel = context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	checkReceive(timedCtx, t, nil, sub2, &wg, false)
+
+	unittest.RequireReturnsBefore(t, wg.Wait, 5*time.Second, "could not receive message on time")
+
+	// expecting 1 warn calls for each rejected message from unauthorized node
+	require.Equalf(t, 1, hookCalls, "expected 1 warning to be logged")
```
jordanschalm, line 232: Shouldn't we get at least 2 warnings? (One from each consensus node which rejected the block proposal from AN)

Or does SN2 never receive the proposal message at all? (it is discarded before being gossiped by SN1)

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	an1 := createNode(t, identity3.NodeID, privateKey3, sporkId)
+
+	topic := engine.TopicFromChannel(engine.SyncCommittee, sporkId)
+
+	ids := flow.IdentityList{identity1, identity2, identity3}
+	translator, err := NewFixedTableIdentityTranslator(ids)
+	require.NoError(t, err)
+
+	// setup hooked logger
+	hookCalls := 0
```
jordanschalm, line 254: I think we should use a synchronized variable here as well (same suggestion as previous test)

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+		}
+		return ids.ByNodeID(fid)
+	})
+
+	// node1 is connected to node2, and the an1 is connected to node1
+	// an1 <-> sn1 <-> sn2
+	require.NoError(t, sn1.AddPeer(context.TODO(), *host.InfoFromHost(sn2.host)))
+	require.NoError(t, an1.AddPeer(context.TODO(), *host.InfoFromHost(sn1.host)))
+
+	// sn1 and sn2 subscribe to the topic with the topic validator
```
jordanschalm, line 275: It looks like only SN1 subscribed with the `authorizedSenderValidator`.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	an1 := createNode(t, identity3.NodeID, privateKey3, sporkId)
+
+	topic := engine.TopicFromChannel(engine.SyncCommittee, sporkId)
+
+	ids := flow.IdentityList{identity1, identity2, identity3}
+	translator, err := NewFixedTableIdentityTranslator(ids)
+	require.NoError(t, err)
+
+	// setup hooked logger
+	hookCalls := 0
```
jordanschalm, line 350: synchronized variable here as well

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+
+	// sn2 also gets the message (as part of the libp2p loopback of published topic messages)
+	checkReceive(timedCtx, t, data1, sub2, nil, true)
+
+	// an1 also gets the message
+	checkReceive(timedCtx, t, data1, sub3, nil, true)
+
+	var wg sync.WaitGroup
+	// "eject" sn2 to ensure messages published by ejected nodes get rejected
+	identity2.Weight = 0
```
jordanschalm, line 408: ⚠️ Setting weight to 0 is not the same as ejection.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +107,349 @@ func TestTopicValidator(t *testing.T) {
...
+	checkReceive(timedCtx, t, nil, sub1, &wg, false)
+
+	unittest.RequireReturnsBefore(t, wg.Wait, 5*time.Second, "could not receive message on time")
+
+	// expecting 1 warn calls for each rejected message from ejected node
+	require.Equalf(t, 1, hookCalls, "expected 1 warning to be logged")
+}
+
+// TestAuthorizedSenderValidator_InvalidWeight tests that the authorized sender validator rejects messages from nodes with invalid weight
+func TestAuthorizedSenderValidator_InvalidWeight(t *testing.T) {
```
jordanschalm, line 332: Mentioned previously, but nodes with 0 weight must be able to participate in limited network communication. We'll need to update this test case.

network/validator/pubsub/authorized_sender_validator.go
kc1116: Got it, so just to confirm: 
- It is safe to discard messages from ejected nodes ([via this link](https://www.notion.so/dapperlabs/Epoch-Networking-0c7b7a6ecd904a9fb8c023069488fe51)
- Certain messages must be allowed for nodes with 0 weight ?

Do you know where I can find this list of limited messages wrt the second point ?

jordanschalm: > Do you know where I can find this list of limited messages wrt the second point ?

I don't think we have a list anywhere, but it is:
* everything needed to synchronize chain state (sync requests/responses, block responses, pushed blocks)
* DKG messages

In terms of channels:
* `consensus-committee`
* `sync-committee`
* `dkg-committee`
* `push-blocks`

kc1116:
```
	// consensus
	authorizedRolesMap[cborcodec.CodeBlockProposal] = flow.RoleList{flow.RoleConsensus}
	authorizedRolesMap[cborcodec.CodeBlockVote] = flow.RoleList{flow.RoleConsensus}

	// protocol state sync
	authorizedRolesMap[cborcodec.CodeSyncRequest] = flow.Roles()
	authorizedRolesMap[cborcodec.CodeSyncResponse] = flow.Roles()
	authorizedRolesMap[cborcodec.CodeRangeRequest] = flow.Roles()
	authorizedRolesMap[cborcodec.CodeBatchRequest] = flow.Roles()
	authorizedRolesMap[cborcodec.CodeBlockResponse] = flow.Roles()
        authorizedRolesMap[cborcodec.CodeDKGMessage] = flow.RoleList{flow.RoleConsensus}
```

This should be the full list, can you confirm ?

jordanschalm: That looks correct. Only change is, we can remove `CodeBlockVote` (zero-weight nodes aren't voting).

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,151 @@
...
+	// dkg
+	authorizedRolesMap[cborcodec.CodeDKGMessage] = flow.RoleList{flow.RoleConsensus} // sn nodes for next epoch
+}
+
+// AuthorizedSenderValidator using the getIdentity func will check if the role of the sender
+// is part of the authorized roles list for the type of message being sent. A node is considered
+// to be authorized to send a message if all of the following are true.
+// 1. The message type is a known message type (initialized in the authorizedRolesMap).
+// 2. The authorized roles list for the message type contains the senders role.
+// 3. The node has a weight > 0 and is not ejected
```
peterargue, line 82: after discussing with Alex, we can skip the weight/ejected checks here since it would be hard to match the message with a specific view/block outside of the application layer. The weight check will need to go into the application layer (or at least outside of the topic validator)

network/validator/pubsub/authorized_sender_validator.go
peterargue: Maybe we could add a method to the cbor codec module that returns this value to make the dependency explicit

network/validator/pubsub/authorized_sender_validator.go
peterargue: I noted in a separate comment, but I discussed with with Alex, and we should skip the weight checks in the topic validator since it's not practical to know the specific view/block a message relates to outside of the application layer. We can still do the ejected checks since those take effect immediately.

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,151 @@
...
+func init() {
+	// initialize the authorized roles map the first time this package is imported.
+	initializeAuthorizedRolesMap()
+}
+
+// authorizedRolesMap is a mapping of message type to a list of roles authorized to send them.
+var authorizedRolesMap map[uint8]flow.RoleList
+
+// initializeAuthorizedRolesMap initializes authorizedRolesMap.
+func initializeAuthorizedRolesMap() {
```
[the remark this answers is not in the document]
kc1116: I like this idea

network/p2p/topic_validator_test.go
kc1116: typo from copying test

network/validator/pubsub/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/topic_validator_test.go
kc1116: I've updated the comment with the reason why sn2 does not subscribe with the topic validator https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/validator/pubsub/authorized_sender_validator.go
kc1116: removed 0 weight checks and test so we don't need this list anymore https://github.com/onflow/flow-go/pull/2317/commits/079042c5ef705456cae43752bd375f29afdec63c

network/validator/pubsub/authorized_sender_validator.go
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/8041cc2e0d443b18962bc42df45da023424810e5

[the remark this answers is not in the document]
kc1116: @jordanschalm can you review changes in this commit ^

network/p2p/topic_validator_test.go
kc1116: This is just waiting for ctx.Done to be called under the hood. After which a context deadline error is returned. What this is asserting in the test is that we wait some certain amount of time without getting a message before we conclude that the message was rejected.

network/p2p/topic_validator_test.go
kc1116: Good catch, this was leftover from initial debugging

kc1116: https://github.com/onflow/flow-go/pull/2317/commits/e00f7c7b3648c6ddc855a9e759131932bc7e7158

network/p2p/topic_validator_test.go
kc1116: Yes the second point is true that SN2 does not receive the message via gossip, this is noted in the comments of the test as well

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/commit/079042c5ef705456cae43752bd375f29afdec63c

network/p2p/middleware.go
```diff
@@ -499,7 +499,10 @@ func (m *Middleware) Subscribe(channel network.Channel) error {
 	var validators []psValidator.MessageValidator
 	if !engine.PublicChannels().Contains(channel) {
 		// for channels used by the staked nodes, add the topic validator to filter out messages from non-staked nodes
-		validators = append(validators, psValidator.StakedValidator(m.ov.Identity))
+		validators = append(validators,
+			psValidator.StakedValidator(m.ov.Identity),
```
jordanschalm, line 503: I think we can remove this validator. All it does is check that the identity exists, which `AuthorizedSenderValidator` will also do.

kc1116: I like the idea of keeping it for the sake of readability. I can look at this middleware.go file and automatically deduce that being a staked node is a requirement for non-public channels. What do you think ?

jordanschalm: For readability, I'm more inclined to include that in the documentation for `AuthorizedSenderValidator`, or use a comment here. Using both `AuthorizedSenderValidator` and `StakedValidator` separately could also lead the reader to incorrectly infer that an "authorized sender" may _not_ be staked.

kc1116: Okay that makes sense i'll add a comment

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/onflow/flow-go/network/message"
+)
+
+func init() {
+	// initialize the authorized roles map the first time this package is imported.
+	initializeChannelToMsgCodesMap()
+}
+
+// channelToMsgCodes is a mapping of network channels to codes of the messages communicated on them. This will be used to check the list of authorized roles associated with the channel
+var channelToMsgCodes map[network.Channel][]uint8
```
peterargue, line 25: A couple of comments on this data structure:

1. Given the coupling of these codec types with the authorizer, how about moving the consts to a higher level namespace that's used by both the `AuthorizedSenderValidator` and codec. The implementation probably doesn't need to change, but at least then it's more clear that they are used for more than the encoding. That way the codec can change the implementation and maintain compatibility with the validator by providing a transparent mapping to the official const.

2. If the codec consts/functions are moved, this mapping should probably live along side that code.

network/p2p/topic_validator_test.go
```diff
@@ -109,31 +109,328 @@ func TestTopicValidator(t *testing.T) {
...
-		require.Error(t, err)
-		wg.Done()
-	}()
+	checkReceive(timedCtx, t, nil, sub2, &wg, false)
+
+	unittest.RequireReturnsBefore(t, wg.Wait, 5*time.Second, "could not receive message on time")
+}
+
+// TestAuthorizedSenderValidator_Authorized tests that the authorized sender validator rejects messages from nodes that are not authorized to send the message
+func TestAuthorizedSenderValidator_UnAuthorized(t *testing.T) {
```
peterargue, line 131: This tests that a node is not authorized because they have the wrong role. Can you add a test for the case where they are not staked.

Can you also add a test for the case where the message type was sent in the wrong channel

peterargue (commented):
Added a few minor comments. Overall this is looking really good

network/p2p/topic_validator_test.go
kc1116: Already done [here](https://github.com/onflow/flow-go/pull/2317/files#diff-f818a0f175186a5b2b5e837819b1f1fe7c16d9fe5f24c989100e4eb7e07ab7c3R239)

network/p2p/middleware.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/9920d68c30fdb7d362ee86723c1ffc4babeb62e0

network/validator/pubsub/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/21addd94142ee7c23ad7689d3eeefca741c6ff03

peterargue: It's good to keep them all together. I'm suggesting we move the message codes out of the codec package to reduce the coupling between the specific cbor encoding currently used, and the authorizer. Right now we need a bunch of warnings to make sure if the codec changes the authorizer is updated. This creates a higher likelihood someone accidentally introduces a bug. If we decouple the two packages, that reduces the likelihood of that type of bug, and adds flexibility to make changes within the codec/use a different codec without having to modify the authorizer.

for example, we could define a new type `type MessageType uint8` that has methods for existing functionality (e.g. `String()`, `ToInterface()` and put it in a common namespace like `network/messages`. Then codec can use the new type/methods, and provide a method that returns the `MessageType` or a payload.

It's fine if you'd rather that go in a different PR. We could add it later

network/p2p/topic_validator_test.go
peterargue: Does that condition get exercised in the test. I thought that test included all of the identities in the identity list, so none of the message would come from an unknown/unstaked node.

network/validator/pubsub/authorized_sender_validator.go
kc1116: There is a direct coupling at the moment from codec -> codes . The JSON codec and cbor encode the message type differen't ways. The also don't start at the same iota see [cbor](https://github.com/onflow/flow-go/blob/master/network/codec/cbor/envelope.go#L6) and [json](https://github.com/onflow/flow-go/blob/master/network/codec/json/envelope.go#L12) . This is why we need the warnings because we actually currently can not decouple the codes from the codec.

network/p2p/topic_validator_test.go
kc1116: Yes see [here](https://github.com/onflow/flow-go/blob/f3ce2727c2a0423da7bfaefd4137e30c48f22fbe/network/p2p/topic_validator_test.go#L146) we are trying to broadcast a block proposal message on the sync committee channel. This could be improved by adding a regex on the log statement instead of relying on the log counter only.

network/validator/pubsub/authorized_sender_validator.go
kc1116: It would be nice to create a separate issue to update the json codec to match the cbor codec and then decouple the codes. I can create and add it to the BFT messages epic.

network/p2p/topic_validator_test.go
peterargue: sorry, I missed that. Should have read the comments. 

My last comment was about the not staked test. Your link pointed to line 239, but it looks like that conditional wouldn't match in the test. maybe I'm being dense

kc1116: Sorry I totally missed that in your previous message i'll add that test case (unstaked node cannot send)

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,129 @@
...
+	roles, err := getRoles(channel, code)
+	if err != nil {
+		return err
+	}
+
+	if !roles.Contains(identity.Role) {
+		return fmt.Errorf("sender is not authorized to send this message type")
+	}
+
+	if identity.Ejected {
```
peterargue, line 82: we should check this immediately after resolving the identity to minimize resource use

peterargue (approved):
Looks great! Thank @kc1116 

Added a few small comments. I can incorporate them in my next PR as well.

network/validator/pubsub/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/33b0bd14a2d1b9438fe9a8a173a5b57cb967bad0

network/p2p/topic_validator_test.go
smnzhu: Maybe we should check that the error is actually what we expect (e.g the context error).

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+// 4. The node is not ejected
+func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getIdentity func(peer.ID) (*flow.Identity, bool)) MessageValidator {
+	log = log.With().
+		Str("component", "authorized_sender_validator").
+		Str("network_channel", channel.String()).
+		Logger()
+
+	// use cbor codec to add explicit dependency on cbor encoded messages adding the message type
+	// to the first byte of the message payload, this adds safety against changing codec without updating this validator
+	codec := cborcodec.NewCodec()
```
smnzhu, line 34: could we have the codec be a parameter?

smnzhu: We could add the message type as the first byte regardless of what the codec is

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+func getRoles(channel network.Channel, msgTypeCode uint8) (flow.RoleList, error) {
+	// echo messages can be sent by anyone
+	if msgTypeCode == cborcodec.CodeEcho {
+		return flow.Roles(), nil
+	}
+
+	// cluster channels have a dynamic channel name
+	if msgTypeCode == cborcodec.CodeClusterBlockProposal || msgTypeCode == cborcodec.CodeClusterBlockVote || msgTypeCode == cborcodec.CodeClusterBlockResponse {
+		return channels.ClusterChannelRoles(channel), nil
+	}
```
smnzhu, line 104: do we eventually plan to check for messages sent on the wrong cluster channel? e.g collection node sending messages on a different cluster channel than the cluster it belongs to.

network/validator/pubsub/authorized_sender_validator.go
smnzhu: My understanding is that currently, the logic of prepending a single-byte "code" to an encoded message has been put directly into the cbor codec implementation: https://github.com/onflow/flow-go/blob/ce0c5aa063f8cbcfb0de058a932dfcc93d0bb5a3/network/codec/cbor/codec.go#L49

However, there's no reason this code-prepending should be specific to Cbor. Instead, we should ideally refactor the code so that we have a "meta-codec" which can take in any arbitrary codec, and encodes messages by prepending a single-byte code to the encoding provided by the underlying codec.

network/validator/pubsub/authorized_sender_validator.go
peterargue: That should be checked, but it's out of scope for the masquerade protection. It's something that belongs in the engine since the application layer should have the context needed to determine the cluster membership for the view

network/validator/pubsub/authorized_sender_validator.go
peterargue: I'm planning to refactor the codec parts in a future PR for the unicast case. Essentially, I want to break the message codes out to a separate `MessageType` module, and add a method to the codec interface that returns the `MessageType` for a given message. That should abstract away the details about the specific encoding and how that's represented within the message.

kc1116: @peterargue is there a ticket for the codec refactoring ?

cmd/node_builder.go
```diff
@@ -210,6 +210,10 @@ func DefaultBaseConfig() *BaseConfig {
 	homedir, _ := os.UserHomeDir()
 	datadir := filepath.Join(homedir, ".flow", "database")
 
+	// NOTE: if the codec used in the network component is ever changed any code relying on
+	// the message format specific to the codec must be updated. i.e: the AuthorizedSenderValidator.
+	codecFactory := func() network.Codec { return cbor.NewCodec() }
```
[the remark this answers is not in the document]
kc1116: `codecFactory := cbor.NewCodec` returns a pointer

network/validator/pubsub/authorized_sender_validator.go
kc1116: We can rearrange the plumbing to make codec a parameter, this will make this PR too big. Best to do all the codec work at once.

kc1116:
@peterargue @smnzhu unless you guys have any other comments I will merge this today.

network/p2p/topic_validator_test.go
kc1116: @smnzhu https://github.com/onflow/flow-go/pull/2317/commits/cb4db5bdc3159bca985e9adbcb832f5964e2186e

peterargue:
looks good to me

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+		Logger()
+
+	// use cbor codec to add explicit dependency on cbor encoded messages adding the message type
+	// to the first byte of the message payload, this adds safety against changing codec without updating this validator
+	codec := cborcodec.NewCodec()
+
+	return func(ctx context.Context, from peer.ID, msg *message.Message) pubsub.ValidationResult {
+		identity, ok := getIdentity(from)
+		if !ok {
+			log.Warn().Str("peer_id", from.String()).Msg("could not get identity of sender")
```
yhassanzadeh13, line 39:
```suggestion
			log.Warn().Str("peer_id", from.String()).Msg("could not verify identity of sender")
```

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+			return pubsub.ValidationReject
+		}
+
+		if identity.Ejected {
+			log.Warn().
+				Err(fmt.Errorf("node %s is an ejected node", identity.NodeID)).
```
yhassanzadeh13, line 45:
```suggestion
				Err(fmt.Errorf("sender %s is an ejected node", identity.NodeID)).
```

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+				Str("role", identity.Role.String()).
+				Msg("rejecting message")
+			return pubsub.ValidationReject
+		}
+
+		// attempt to decode the flow message type from encoded payload
+		code, what, err := codec.DecodeMsgType(msg.Payload)
+		if err != nil {
+			log.Warn().
+				Err(err).
```
yhassanzadeh13, line 56: Please also append the flow identifier of the sender.

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		if err := isAuthorizedSender(identity, channel, code); err != nil {
+			log.Warn().
+				Err(err).
+				Str("peer_id", from.String()).
+				Str("role", identity.Role.String()).
+				Str("message_type", what).
+				Str("network_channel", channel.String()).
+				Msg("rejecting message")
```
yhassanzadeh13, line 70:
```suggestion
				Msg("sender is not authorized, rejecting message")
```

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+
+			return pubsub.ValidationReject
+		}
+
+		return pubsub.ValidationAccept
+	}
+}
+
+// isAuthorizedSender checks if node is an authorized role and is not ejected
```
yhassanzadeh13, line 79:
```suggestion
// isAuthorizedSender checks if node is an authorized role and is not ejected.
```
`godoc`s must end with a period.

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// cluster channels have a dynamic channel name
+	if msgTypeCode == cborcodec.CodeClusterBlockProposal || msgTypeCode == cborcodec.CodeClusterBlockVote || msgTypeCode == cborcodec.CodeClusterBlockResponse {
+		return channels.ClusterChannelRoles(channel), nil
+	}
+
+	// get message type codes for all messages communicated on the channel
+	codes, ok := cborcodec.ChannelToMsgCodes[channel]
+	if !ok {
+		return nil, fmt.Errorf("could not get message codes for unknown channel")
```
yhassanzadeh13, line 109:
```suggestion
		return nil, fmt.Errorf("could not get message codes for unknown channel: %s", channel)
```

yhassanzadeh13 (approved):

network/validator/pubsub/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/2317/commits/45c69ab5ae35a670a6634a30a2ad9c0884aed2ac

PR #2321, opened 2022-04-18
[Access node] Add random en lookup on receipt lookup failure

koko1123:
Add logic to lookup random execution nodes on insufficient receipts instead of a blank list/custom error

### DIFF
--- a/engine/access/rpc/backend/backend.go
+++ b/engine/access/rpc/backend/backend.go
@@ -396,11 +396,11 @@ func executionNodesForBlockID(
 		receiptCnt := len(executorIDs)
 		// if less than minExecutionNodesCnt execution receipts have been received so far, then return random ENs
 		if receiptCnt < minExecutionNodesCnt {
-			executorIdentities, err := state.Final().Identities(filter.HasRole(flow.RoleExecution))
+			executionNodes, err := state.AtBlockID(blockID).Identities(filter.HasRole(flow.RoleExecution))
 			if err != nil {
 				return nil, fmt.Errorf("failed to retreive execution IDs for block ID %v: %w", blockID, err)
 			}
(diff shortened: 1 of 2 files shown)


engine/access/rpc/backend/backend.go
```diff
@@ -394,9 +394,13 @@ func executionNodesForBlockID(
...
 		receiptCnt := len(executorIDs)
-		// if less than minExecutionNodesCnt execution receipts have been received so far, then throw an error
+		// if less than minExecutionNodesCnt execution receipts have been received so far, then return random ENs
 		if receiptCnt < minExecutionNodesCnt {
-			return flow.IdentityList{}, InsufficientExecutionReceipts{blockID: blockID, receiptCount: receiptCnt}
+			executionNodes, err := state.AtBlockID(blockID).Identities(filter.HasRole(flow.RoleExecution))
+			if err != nil {
+				return nil, fmt.Errorf("failed to retreive execution IDs for block ID %v: %w", blockID, err)
+			}
+			return executionNodes.Sample(maxExecutionNodesCnt), nil
```
smnzhu, line 403: Instead of returning right away, should we still continue to the block below (`chooseExecutionNodes`, etc)? I think that function already does most of the same as what you're doing here.

koko1123: so just remove this `if`?

smnzhu: No, I mean inside this `if`, just set `executorIDs` to `state.AtBlockID(blockID).Identities(filter.HasRole(flow.RoleExecution).NodeIDs()`

smnzhu (approved):

peterargue (approved):

smnzhu:
bors merge

PR #2327, opened 2022-04-19
Fixed some tests for AddPublicKey and RemoveePublicKey

robert-e-davidson3:


admin/tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
janezpodhostnik, line 1: Are these needed?

robert-e-davidson3: No. I ran `go fmt ./...` by habit and it did that. I can undo those changes if they're a problem.

janezpodhostnik:
Merging this into [janez/memory-weight-assignment-from-contract](https://github.com/onflow/flow-go/tree/janez/memory-weight-assignment-from-contract) and adding more test fixes there.

PR #2328, opened 2022-04-19
Support block-scoped and epoch-scoped identity queries in HotStuff committee

jordanschalm:
This PR modifies `hotstuff.Committee` to support epoch-scoped identity queries. 
* The interface is split into:
  * `Replicas` and `*ByEpoch` methods are used for validating votes (committee is static over epoch)
  * `DynamicCommittee` and `*ByBlock` methods are used for validating whether a replica may propose a block 
* The new consensus committee implementation relies on a new assumption for EECC:
  * https://github.com/dapperlabs/flow-go/issues/6227 will be implemented. This will mean that the `Committee` implementation can rely on a fallback epoch being injected into the protocol state which can be queried, rather than using the absence of a committed next epoch as evidence to trigger epoch fallback behaviour.

#### Critical Changes
  * [Interface change](https://github.com/onflow/flow-go/pull/2328/files#diff-7a8c7272cf23ec746eba199e12c834b4cdadcc85e57191c90dfd8cbe5f8303a7)
  * [Security-sensitive API usage](https://github.com/onflow/flow-go/pull/2328/files#diff-8e933e078b2e6031be4617d76cfe25046e38ac012720b757b382869e6a509aea)
  * [New Committee implementation](https://github.com/onflow/flow-go/pull/2328/files#diff-7236d580d88b376c2040caf16dec5857f2c23366052f90e6f161d6d4646ca5b3)

#### Outstanding Todos

- [x] Handing new sentinel `ErrViewForUnknownEpoch` - TODOs added in this PR, will be handled when front-loading verification
- [x] There are still many tests which haven't been updated yet - they are marked with:
```
unittest.SkipUnless(t, unittest.TEST_TODO, "COMMITTEE_BY_VIEW - updating in next pr")
```
- [x] Return "identity skeleton" model for epoch-scoped queries - to implement in subsequent PR (https://github.com/dapperlabs/flow-go/issues/6232)
- [ ] Add f+1 threshold to Replicas in addition to 2f+1 threshold, for creating timeout.


consensus/hotstuff/committee.go
```diff
@@ -5,40 +5,36 @@ import (
 	"github.com/onflow/flow-go/state/protocol"
 )
```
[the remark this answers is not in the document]
jordanschalm: Thanks for the suggestions. I agree creating two models for "initial identity" and "dynamic identity" makes sense, though I think fully implementing this should be in the scope of a separate PR. I'll add TODOs for now where that model type should be used, and implement the new model in a separate PR: https://github.com/dapperlabs/flow-go/issues/6232.

AlexHentschel:
I think also these methods in the Epoch interface should not return full identities: https://github.com/onflow/flow-go/blob/d6666fc16c39b6790001cfa6df1f13ac318d3065/state/protocol/epoch.go#L69-L74
but rather Identity Skeletons

consensus/hotstuff/committee.go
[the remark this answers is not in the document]
jordanschalm: Renamed the interfaces in https://github.com/onflow/flow-go/pull/2328/commits/57357180aa72cecffb60f7b247b7a17056cbee44

consensus/hotstuff/committees/cluster_committee.go
```diff
@@ -119,6 +120,22 @@ func (c *Cluster) Identity(blockID flow.Identifier, nodeID flow.Identifier) (*fl
 	return identity, nil
 }
 
+// IdentitiesByEpoch returns the initial cluster members for this epoch. Since clusters
+// only exist for one epoch, we don't need to check the view.
+func (c *Cluster) IdentitiesByEpoch(_ uint64, selector flow.IdentityFilter) (flow.IdentityList, error) {
```
AlexHentschel, line 125: I think it might increase clarity if we explicitly point out that the view here is the view within the local collector cluster. Just to avoid confusion. 

same here: 
* https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/committees/cluster_committee.go#L131
* https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/committees/cluster_committee.go#L139

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,83 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/model/flow/filter"
 	"github.com/onflow/flow-go/state/protocol"
-	"github.com/onflow/flow-go/state/protocol/seed"
 )
 
-var errSelectionNotComputed = fmt.Errorf("leader selection for epoch not yet computed")
+// ErrViewForUnknownEpoch is returned when a by-view query is made with a view
+// outside all cached epochs. This can happen when a query is made for a view in the
```
AlexHentschel, line 17: wording suggestion that is easier to understand for my non-native-english-speaking brain 🤷  

```suggestion
// ErrViewForUnknownEpoch is returned when Epoch information is queried for a view that is 
// outside of all cached epochs. This can happen when a query is made for a view in the
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,83 @@ import (
...
+	}
+	finalView, err := epoch.FinalView()
+	if err != nil {
+		return nil, fmt.Errorf("could not get final view: %w", err)
+	}
+	leaders, err := leader.SelectionForConsensus(epoch)
+	if err != nil {
+		return nil, fmt.Errorf("could not get leader selection: %w", err)
+	}
+	initialidentities, err := epoch.InitialIdentities()
```
AlexHentschel, line 46:
```suggestion
	initialIdentities, err := epoch.InitialIdentities()
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,83 @@ import (
...
+		finalView:        finalView,
+		leaders:          leaders,
+		initialCommittee: initialCommittee,
+		dkg:              dkg,
+	}
+	return epochInfo, nil
+}
 
 // Consensus represents the main committee for consensus nodes. The consensus
 // committee persists across epochs.
```
AlexHentschel, line 67:
```suggestion
// committee might be active for multiple successive epochs.
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,83 @@ import (
...
 
-var errSelectionNotComputed = fmt.Errorf("leader selection for epoch not yet computed")
+// ErrViewForUnknownEpoch is returned when a by-view query is made with a view
+// outside all cached epochs. This can happen when a query is made for a view in the
+// next epoch, if that epoch is not committed yet. This can also happen when an
+// old epoch is queried (>3 in the past), even if that epoch does exist in storage.
+var ErrViewForUnknownEpoch = fmt.Errorf("by-view query for unknown epoch")
+
+// staticEpochInfo contains leader selection and the initial committee for one epoch.
+type staticEpochInfo struct {
```
AlexHentschel, line 23: Is the `staticEpochInfo` supposed to be an immutable data structure? If so, it would be great to explicitly point this out in the goDoc.

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,83 @@ import (
...
-		leaders: make(map[uint64]*leader.LeaderSelection),
+		state:  state,
+		me:     me,
+		epochs: make(map[uint64]*staticEpochInfo),
 	}
 
 	final := state.Final()
 
 	// pre-compute leader selection for current epoch
 	current := final.Epochs().Current()
```
AlexHentschel, line 89: As best practise, I prefer to to only query final once, because the return value can theoretically change between different queries. We query twice in this method: [here](https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/committees/consensus_committee.go#L89) and [here](https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/committees/consensus_committee.go#L101)

I don't think there is an issue with the way it is currently implemented, because finality can probably not progress within the constructor. Nevertheless, I think it is no extra hassle to stick to our best practise and write
```golang
	eq := final.Epochs()
	current := eq.Current()
        ...
	previous := eq.Previous()
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -90,202 +138,202 @@ func (c *Consensus) Identity(blockID flow.Identifier, nodeID flow.Identifier) (*
...
+//
+// TODO: trigger this via an event from protocol state instead
+//   * previously the asynchrony of protocol events was not acceptable, because any
+//     unknown view errors for LeaderForView were fatal
+//   * now, we handle these errors, since we may validate messages from future views -
+//     so long as we eventually cache a newly committed epoch, liveness is not compromised
+func (c *Consensus) tryPrepareNextEpoch() (*staticEpochInfo, bool, error) {
+	next := c.state.Final().Epochs().Next()
+	committed, err := protocol.IsEpochCommitted(next)
+	fmt.Println("committed? ", committed, err)
```
AlexHentschel, line 273: print should be _eventually_ removed

state/protocol/util.go
```diff
@@ -69,3 +70,17 @@ func IsSporkRootSnapshot(snapshot Snapshot) (bool, error) {
 	}
 	return true, nil
 }
+
+// IsEpochCommitted returns whether the epoch is committed.
```
AlexHentschel, line 74: Lets be explicit with documentation:
```suggestion
// IsEpochCommitted returns whether the epoch is committed.
// No errors are expected during normal operations. 
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -90,202 +138,202 @@ func (c *Consensus) Identity(blockID flow.Identifier, nodeID flow.Identifier) (*
...
-		c.mu.Unlock()
-		return selection.LeaderForView(view)
-	}
+// IdentityByEpoch returns the identity for the given node ID, in the epoch which
+// contains the given view.
+//
+// Error returns:
+//   * ErrViewForUnknownEpoch if no committed epoch containing the given view is known.
+//     This is an expected error and must be handled.
+//   * unspecific error in case of unexpected problems and bugs
```
AlexHentschel, line 163: `InvalidSignerError` should be added as an expected error:
```suggestion
// During normal operations, the following errors are expected:
//   * model.ErrViewForUnknownEpoch if no committed epoch containing the given view is known.
//     This is an expected error and must be handled.
//   * model.InvalidSignerError if nodeID was not listed by the Epoch Setup event as an authorized consensus participants
```

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -90,202 +138,202 @@ func (c *Consensus) Identity(blockID flow.Identifier, nodeID flow.Identifier) (*
...
-	if err == nil {
-		return id, nil
-	}
-	if !errors.Is(err, errSelectionNotComputed) {
-		return flow.ZeroID, err
+// IdentitiesByEpoch returns the committee identities in the epoch which contains
+// the given view.
+//
+// Error returns:
+//   * ErrViewForUnknownEpoch if no committed epoch containing the given view is known.
```
AlexHentschel, line 145:
```suggestion
//   * model.ErrViewForUnknownEpoch if no committed epoch containing the given view is known.
```

AlexHentschel (commented):
next set of comments:

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 func TestProduceVote(t *testing.T) {
+	unittest.SkipUnless(t, unittest.TEST_TODO, "COMMITTEE_BY_VIEW - updating in next pr")
```
durkmurder, line 16: No need to fix this test, I will update it either way.

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -13,16 +13,16 @@ type SafetyRules struct {
 	signer        hotstuff.Signer
 	forks         hotstuff.ForksReader
 	persist       hotstuff.Persister
-	committee     hotstuff.Committee // only produce votes when we are valid committee members
-	lastVotedView uint64             // need to keep track of the last view we voted for so we don't double vote accidentally
+	committee     hotstuff.Replicas // only produce votes when we are valid committee members
```
durkmurder, line 16: @AlexHentschel I think we need to use dynamic committee here, replica always votes for block and we don't want ejected nodes to vote on block. No need to fix this, I just want to double check this.

AlexHentschel: Thanks for noticing this @durkmurder. I would agree that we need to use the dynamic committee here. 
Though, I have a slightly different view on the reasoning for why it should be the dynamic committee:
* For an ejected node, it is of little concern whether or not it votes. Especially if the node is byzantine, why wouldn't it vote? Of course we desire honest nodes not to vote once we enter the epoch where they are unstaking. But I think this would also be achieved by the static committee.  
* I think we need the dynamic committee on the recipients' end:
  * An honest replica should not vote for a block from an ejected node. Hence, the dynamic committee. 
  * Ideally, an honest replica should also exclude votes from ejected replicas.

AlexHentschel:
We could move the `WeightThreshold` into the `hotstuff.Replicas` API: 
https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/validator/validator.go#L58-L59

This would allow us to pre-compute the value once per Epoch and store it in the `staticEpochInfo`

consensus/hotstuff/votecollector/combined_vote_processor_v3.go
```diff
@@ -31,7 +31,7 @@ import (
 // the proposer's vote (decorator pattern).
 // nolint:unused
 type combinedVoteProcessorFactoryBaseV3 struct {
-	committee   hotstuff.Committee
+	committee   hotstuff.Replicas
```
AlexHentschel, line 34: same argument as for [combined_vote_processor_v2.go](https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/votecollector/combined_vote_processor_v2.go), I think we should use the `DynamicCommittee` here to reject votes from ejected replicas. 

## ⚠️ 
if you follow my suggestion, the following line needs to be updated:
https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/votecollector/combined_vote_processor_v3.go#L89

consensus/hotstuff/votecollector/factory.go
```diff
@@ -63,7 +63,7 @@ func (f *VoteProcessorFactory) Create(log zerolog.Logger, proposal *model.Propos
 // NewStakingVoteProcessorFactory implements hotstuff.VoteProcessorFactory for
 // members of a collector cluster. For their cluster-local hotstuff, collectors
 // only sign with their staking key.
-func NewStakingVoteProcessorFactory(committee hotstuff.Committee, onQCCreated hotstuff.OnQCCreated) *VoteProcessorFactory {
+func NewStakingVoteProcessorFactory(committee hotstuff.Replicas, onQCCreated hotstuff.OnQCCreated) *VoteProcessorFactory {
```
AlexHentschel, line 66: same argument as for [combined_vote_processor_v2.go](https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/votecollector/combined_vote_processor_v2.go), I think we should use the `DynamicCommittee` for all factories here to reject votes from ejected replicas.

consensus/hotstuff/votecollector/staking_vote_processor.go
```diff
@@ -29,14 +29,14 @@ import (
 // by `votecollector.VoteProcessorFactory` which adds the logic to verify
 // the proposer's vote (decorator pattern).
 type stakingVoteProcessorFactoryBase struct {
-	committee   hotstuff.Committee
+	committee   hotstuff.Replicas
```
AlexHentschel, line 32: same argument as for [combined_vote_processor_v2.go](https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/votecollector/combined_vote_processor_v2.go), I think we should use the `DynamicCommittee` here to reject votes from ejected replicas. 

⚠️ not to forget updating https://github.com/onflow/flow-go/blob/4de4231e43a65fe672a54aab4f84971ecac03db6/consensus/hotstuff/votecollector/staking_vote_processor.go#L58

engine/testutil/nodes.go
```diff
@@ -696,11 +696,23 @@ type RoundRobinLeaderSelection struct {
 	me         flow.Identifier
 }
```
AlexHentschel, line 698:
```suggestion
var _ hotstuff.Replicas = (*Consensus)(nil)
var _ hotstuff.DynamicCommittee = (*Consensus)(nil)
```

engine/testutil/nodes.go
```diff
@@ -696,11 +696,23 @@ type RoundRobinLeaderSelection struct {
...
-func (s *RoundRobinLeaderSelection) Identities(blockID flow.Identifier, selector flow.IdentityFilter) (flow.IdentityList, error) {
+func (s *RoundRobinLeaderSelection) IdentitiesByBlock(_ flow.Identifier, selector flow.IdentityFilter) (flow.IdentityList, error) {
 	return s.identities.Filter(selector), nil
 }
 
-func (s *RoundRobinLeaderSelection) Identity(blockID flow.Identifier, participantID flow.Identifier) (*flow.Identity, error) {
+func (s *RoundRobinLeaderSelection) IdentityByBlock(_ flow.Identifier, participantID flow.Identifier) (*flow.Identity, error) {
+	id, found := s.identities.ByNodeID(participantID)
+	if !found {
+		return nil, fmt.Errorf("not found")
```
AlexHentschel, line 706:
```suggestion
		return nil, model.NewInvalidSignerErrorf("not found")
```

engine/testutil/nodes.go
```diff
@@ -696,11 +696,23 @@ type RoundRobinLeaderSelection struct {
...
+}
+
+func (s *RoundRobinLeaderSelection) IdentitiesByEpoch(_ uint64, selector flow.IdentityFilter) (flow.IdentityList, error) {
+	return s.identities.Filter(selector), nil
+}
+
+func (s *RoundRobinLeaderSelection) IdentityByEpoch(_ uint64, participantID flow.Identifier) (*flow.Identity, error) {
 	id, found := s.identities.ByNodeID(participantID)
 	if !found {
 		return nil, fmt.Errorf("not found")
```
AlexHentschel, line 718:
```suggestion
		return nil, model.NewInvalidSignerErrorf("not found")
```

AlexHentschel (approved):
Looks good. Thanks for adding much detailed goDoc 👏 and the nice error handing!

Many of my comments are suggestions where to use the `hotstuff.Replicas` vs the `hotstuff.DynamicCommittee`. I am trying to already anticipate the design where we front-load the QC checks.

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -11,35 +11,82 @@ import (
...
+// next epoch, if that epoch is not committed yet. This can also happen when an
+// old epoch is queried (>3 in the past), even if that epoch does exist in storage.
+var ErrViewForUnknownEpoch = fmt.Errorf("by-view query for unknown epoch")
+
+// staticEpochInfo contains leader selection and the initial committee for one epoch.
+type staticEpochInfo struct {
+	firstView        uint64
+	finalView        uint64
+	leaders          *leader.LeaderSelection
+	initialCommittee flow.IdentityList
```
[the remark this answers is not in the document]
jordanschalm: Adding TODO

consensus/hotstuff/committees/consensus_committee.go
jordanschalm: Added to godoc

consensus/hotstuff/committees/consensus_committee.go
jordanschalm: We query a particular snapshot only once though, on line 86 (`final := state.Final()`). Once we capture the snapshot once, querying it multiple times afterward is safe.

consensus/hotstuff/committees/consensus_committee.go
jordanschalm: removed

consensus/hotstuff/committees/consensus_committee.go
jordanschalm: added

consensus/hotstuff/committees/consensus_committee.go
```diff
@@ -90,202 +138,202 @@ func (c *Consensus) Identity(blockID flow.Identifier, nodeID flow.Identifier) (*
...
+// given epoch, including leader selection. Calling prepareEpoch multiple times
+// for the same epoch returns cached static epoch information.
 //
-// Returns the leader selection for the given epoch.
-func (c *Consensus) prepareLeaderSelection(epoch protocol.Epoch) (*leader.LeaderSelection, error) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// Input must be a committed epoch.
+//
+func (c *Consensus) prepareEpoch(epoch protocol.Epoch) (*staticEpochInfo, error) {
```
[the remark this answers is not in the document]
jordanschalm: added sanity check

consensus/hotstuff/validator/validator.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // Validator is responsible for validating QC, Block and Vote
 type Validator struct {
-	committee hotstuff.Committee
+	committee hotstuff.DynamicCommittee
```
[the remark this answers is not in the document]
jordanschalm: Agreed, added a note

consensus/hotstuff/validator/validator.go
```diff
@@ -45,8 +45,8 @@ func (v *Validator) ValidateQC(qc *flow.QuorumCertificate, block *model.Block) e
 	}
 
 	// Retrieve full Identities of all legitimate consensus participants and the Identities of the qc's signers
-	// IdentityList returned by hotstuff.Committee contains only legitimate consensus participants for the specified block (must have positive weight)
-	allParticipants, err := v.committee.Identities(block.BlockID, filter.Any)
+	// IdentityList returned by hotstuff.DynamicCommittee contains only legitimate consensus participants for the specified block (must have positive weight)
```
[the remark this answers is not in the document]
jordanschalm: adjusted the comment

consensus/hotstuff/validator/validator.go
```diff
@@ -107,6 +107,15 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
+	// check the proposer is a valid committee member at the most recent state on this fork
+	_, err = v.committee.IdentityByBlock(block.BlockID, block.ProposerID)
+	if model.IsInvalidSignerError(err) {
+		return newInvalidBlockError(block, fmt.Errorf("proposer %x is leader but is not a valid committee member", block.ProposerID))
+	}
+	if err != nil {
+		return fmt.Errorf("unexpected error checking proposer standing for block %x: %w", block.BlockID, err)
+	}
+
 	// check that we have the parent for the proposal
```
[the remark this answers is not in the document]
jordanschalm: added a TODO

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -30,15 +30,15 @@ import (
 // by `votecollector.VoteProcessorFactory` which adds the logic to verify
 // the proposer's vote (decorator pattern).
 type combinedVoteProcessorFactoryBaseV2 struct {
-	committee   hotstuff.Committee
+	committee   hotstuff.Replicas
```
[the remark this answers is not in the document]
jordanschalm: Just wondering, why do we only want to accept votes from un-ejected replicas? We will accept QCs with votes from ejected replicas, so these QCs are considered valid. 

I understand that we _can_ reject votes from ejected replicas here, so I'm guessing the argument is to err on the side of caution because it's just as easy to use the dynamic committee in this case. But is there any particular attack we're avoiding by doing this?

consensus/hotstuff/validator/validator.go
```diff
@@ -107,6 +107,15 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
 		return newInvalidBlockError(block, fmt.Errorf("proposer %s is not leader (%s) for view %d", block.ProposerID, leader, block.View))
 	}
 
+	// check the proposer is a valid committee member at the most recent state on this fork
```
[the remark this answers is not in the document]
jordanschalm: added TODO

consensus/hotstuff/safetyrules/safety_rules.go
jordanschalm: Copying this question here: https://github.com/onflow/flow-go/pull/2328#discussion_r867209653 regarding which of the these statements we care about more 😅 

> For an ejected node, it is of little concern whether or not it votes

> an honest replica should also exclude votes from ejected replicas.

jordanschalm:
TODO: Add `f+1` threshold to `Replicas` in addition to `2f+1` threshold, for creating timeout.

durkmurder (approved):
Checked last changes related to `DynamicCommittee`, looks good to me.

PR #2334, opened 2022-04-20
[Tool] add rollback executed height

zhangchiqing:
This PR adds the a rollback-executed-height command to the util cmd. 

This command allows execution node operator to roll back the executed block height, in order to re-execute blocks. 

It will remove results, state commitment, and transaction results to every block above the given height.

### DIFF
--- a/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
+++ b/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
@@ -97,6 +97,8 @@ func run(*cobra.Command, []string) {
 
 }
 
+// use badger instances directly instead of stroage interfaces so that the interface don't
+// need to include the Remove methods
 func removeExecutionResultsFromHeight(
 	protoState protocol.State,
 	transactionResults *badger.TransactionResults,
--- a/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
+++ b/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
@@ -228,13 +228,13 @@ func removeForBlockID(
 	}
 
 	// remove own execution results index
-	err = myReceipts.RemoveByBlockID(blockID)
+	err = myReceipts.RemoveIndexByBlockID(blockID)
 	if err != nil {
 		if !errors.Is(err, storage.ErrNotFound) {
-			return fmt.Errorf("could not remove own result by BlockID %v: %w", blockID, err)
+			return fmt.Errorf("could not remove own receipt by BlockID %v: %w", blockID, err)
 		}
 
-		log.Warn().Msgf("own result not found for block %v", blockID)
+		log.Warn().Msgf("own receipt not found for block %v", blockID)
 	}
 
 	// remove events
@@ -247,6 +247,16 @@ func removeForBlockID(
 	err = serviceEvents.RemoveByBlockID(blockID)
 	if err != nil {
 		return fmt.Errorf("could not remove service events by blockID %v: %w", blockID, err)
+	}
+
+	// remove execution result index
+	err = results.RemoveIndexByBlockID(blockID)
+	if err != nil {
+		if !errors.Is(err, storage.ErrNotFound) {
+			return fmt.Errorf("could not remove result by BlockID %v: %w", blockID, err)
+		}
+
+		log.Warn().Msgf("result not found for block %v", blockID)
 	}
 
 	return nil
--- a/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
+++ b/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
@@ -11,6 +11,7 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/state/protocol"
 	"github.com/onflow/flow-go/storage"
+	"github.com/onflow/flow-go/storage/badger"
 )
 
 var (
@@ -54,7 +55,13 @@ func run(*cobra.Command, []string) {
 	state, err := common.InitProtocolState(db, storages)
 
 	if err != nil {
-		log.Fatal().Msg("could not init states")
+		log.Fatal().Err(err).Msg("could not init states")
+	}
+
+	// cast into storage module
+	headers, ok := storages.Headers.(*badger.Headers)
+	if !ok {
+		log.Fatal().Msg("could not use badger.Headers as storage.Headers")
 	}
 
 	err = removeExecutionResultsFromHeight(
@@ -74,7 +81,7 @@ func run(*cobra.Command, []string) {
 		log.Fatal().Err(err).Msgf("could not get block header at height %v", flagHeight)
 	}
 
-	err = storages.Headers.RollbackExecutedBlock(header)
+	err = headers.RollbackExecutedBlock(header)
 	if err != nil {
 		log.Fatal().Err(err).Msgf("could not roll back executed block at height %v", flagHeight)
 	}
--- a/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
+++ b/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
@@ -170,7 +170,8 @@ func removeForBlockID(
 	transactionResults storage.TransactionResults,
 	results storage.ExecutionResults,
 	chunks storage.ChunkDataPacks,
-	blockID flow.Identifier) error {
+	blockID flow.Identifier,
+) error {
 	result, err := results.ByBlockID(blockID)
 	if errors.Is(err, storage.ErrNotFound) {
 		log.Info().Msgf("result not found for block %v", blockID)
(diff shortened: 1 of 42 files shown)


cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
```diff
@@ -0,0 +1,206 @@
...
+	Short: "Rollback the executed height",
+	Run:   run,
+}
+
+func init() {
+
+	// execution results from height + 1 will be removed
+	Cmd.Flags().Uint64Var(&flagHeight, "height", 0,
+		"the height of the block to update the highest executed height")
+	_ = Cmd.MarkFlagRequired("from-height")
```
jordanschalm, line 32: These are using different flag names

cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if err != nil {
+		return fmt.Errorf("could not find result for block %v: %w", blockID, err)
+	}
+
+	// remove chunks
+	for _, chunk := range result.Chunks {
+		chunkID := chunk.ID()
+		err := chunks.Remove(chunkID)
+		if errors.Is(err, storage.ErrNotFound) {
```
jordanschalm, line 180: Would be good to log here - I don't think there is an expected path where we have an ER but not the chunks (or only some of the chunks).

cmd/util/cmd/rollback-executed-height/cmd/root.go
```diff
@@ -0,0 +1,38 @@
...
+
+var (
+	flagDatadir string
+)
+
+// run with `./remove-execution-fork --datadir /var/flow/data/protocol`
+var rootCmd = &cobra.Command{
+	Use:   "remove-execution-result",
+	Short: "remove execution result",
+	Run:   run,
```
jordanschalm, line 19: We should change the command name or the directory name to be consistent (either `remove-execution-result` or `rollback-execution-result`)

jordanschalm (approved):

AlexHentschel:
Quick question: 
* `execution.state.State` has a reference to `storage.ExecutionReceipts`, where the consensus follower store receipts that were _observed in blocks_: https://github.com/onflow/flow-go/blob/dc2e3c2f6dd977524d2d2329022adbdc84a3b078/engine/execution/state/state.go#L87
I don't think this is used. Could we remove it (as part of this PR)?

engine/execution/state/state.go
```diff
@@ -336,6 +336,22 @@ func (s *state) GetExecutionResultID(ctx context.Context, blockID flow.Identifie
 func (s *state) SaveExecutionResults(ctx context.Context, header *flow.Header, endState flow.StateCommitment,
 	chunkDataPacks []*flow.ChunkDataPack, executionReceipt *flow.ExecutionReceipt, events []flow.EventsList, serviceEvents flow.EventsList,
 	results []flow.TransactionResult) error {
+	err := s.saveExecutionResults(ctx, header, endState, chunkDataPacks, executionReceipt, events, serviceEvents, results, false)
+	if err == nil {
+		return nil
+	}
+
+	// if result already exists, we try again by force re-indexing the result by the block ID,
+	if storage.IsResultAlreadyExistsErr(err) {
```
AlexHentschel, line 345: I find this error handling too coarse. I feel there are a lot of reasons why `ResultAlreadyExistsErr` could be returned, but I am not sure that forcing an overwrite in all cases is the appropriate response. Essentially, if you are responding to all cases with a forced override, you have to argue that this is the correct action in _all possible cases_ and that this behaviour is safe in a byzantine environment and won't get the EN accidentally slashed. 
To me the current approach feels like a "continue on a best effort basis", which is a pattern we are aiming to avoid. 

My assumption why you added this code is: 
* The EN maintains an index  BlockID -> resultID for its _own_ results.
* You don't clean this up in [`removeExecutionResultsFromHeight`](https://github.com/onflow/flow-go/blob/7b041a45902c4cce7fee7c293bab3711e6273fbb/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go#L93), so when the EN later executes blocks again, it potentially finds that the block it just executed already has a resultID associated with it (specifically the result the EN had _before_ you removed the results). 




### Suggestion
* I would find it much cleaner if you also cleaned up this secondary index from BlockID -> resultID in [`removeExecutionResultsFromHeight`](https://github.com/onflow/flow-go/blob/7b041a45902c4cce7fee7c293bab3711e6273fbb/cmd/util/cmd/rollback-executed-height/cmd/rollback-executed-height.go#L93)
* thereby, this work-around due to an the incomplete cleanup would not be necessary.

storage/badger/operation/transaction_results.go
```diff
@@ -69,3 +71,30 @@ func LookupTransactionResultsByBlockIDUsingIndex(blockID flow.Identifier, txResu
 
 	return traverse(makePrefix(codeTransactionResultIndex, blockID), txErrIterFunc)
 }
+
+// RemoveTransactionResultsByBlockID removes the transaction results for the given blockID
+func RemoveTransactionResultsByBlockID(blockID flow.Identifier) func(*badger.Txn) error {
+	return func(txn *badger.Txn) error {
+		var txResults []flow.TransactionResult
+
+		err := SkipNonExist(LookupTransactionResultsByBlockID(blockID, &txResults))(txn)
```
AlexHentschel, line 80: Not sure how performance relevant this code is. But I assume in a common case, we are removing thousands maybe even hundred out thousand entries. So it might take a while.

There are some comments for performance [here](https://github.com/dgraph-io/badger/issues/598) (specifically doing key-only iteration as opposed to de-serializing each individual transaction result, just to delete the key)

storage/results.go
```diff
@@ -28,7 +28,7 @@ type ExecutionResults interface {
 	ForceIndex(blockID flow.Identifier, resultID flow.Identifier) error
 
 	// BatchIndex indexes an execution result by block ID in a given batch
```
AlexHentschel, line 30: A short documentation what the behaviour of the `ForceIndex` flag is would be great. Thanks.

engine/execution/state/state.go
zhangchiqing: We need this check here, not because we didn't remove the index, it's because we can't remove the result due to the fact a result might be included in a block's payload already. Removing the result might cause EN unable to load this block for re-executing it. So the result can't be removed during rolling back.

The benefit of keeping the index is that, we can tell if we are executing a block for the first time or the second time by checking if this index exists. 

If we are executing for the first time, then we should not run into ResultAlreadyExist error. 
If we are executing for the second time, then there are still two cases: 
1. the produced result is the same result as the first time
2. the produced result is a different one

For the second case, we should not run into ResultAlreadyExist error, but for the first case, we will. 

Can we distinguish the two cases and only handle this error if it's the first case? 

Yes, but we have to query the result by ID before inserting the result. Unfortunately, we can't do it because saving the execution results is done with a batch operation. Badger doesn't support any DB read within a batch update( the [WriteBatch struct](https://github.com/dgraph-io/badger/blob/master/batch.go#L170) only has `Set` method, no `Get` method). 

One workaround is to query the index and result before batch inserting result related data. Given the safe assumption that index and result will not be removed during the node is running, the query result should still hold true when later batch operation is run in a separate DB transaction.

The only drawback is that the two additional query slows down the saving of the result, given this is very rare to happen. 

But even we add the query, I think it's essentially the same as with this check. My point is that adding the query could avoid this check, but they are handling (distinguishing) the same cases. So I'm not sure we would gain much. 

Thoughts?

storage/badger/operation/common.go
```diff
@@ -84,6 +84,29 @@ func insert(key []byte, entity interface{}) func(*badger.Txn) error {
 	}
 }
 
+// batchUpdate will encode the given entity with JSON and update the binary data
+// under the given key in the badger DB. It will error if the key does not exist
+// yet.
+// batchUpdate assume it's the item of the key exists already
```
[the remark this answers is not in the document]
zhangchiqing: Nice suggestion. 👍

engine/execution/state/state.go
zhangchiqing: I think what you says makes sense @AlexHentschel 

We should remove all the results related data including events, service events, all kinds of index, as if the block was never executed. Except the result and the receipt itself, because it's possible they have been included in a block's payload.

I'm able to remove this check, since we actually won't hit this error, because the batch write would overwrite the result if already exists.

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	log.Info().Msgf("removed height from %v. removed for %v finalized blocks, and %v pending blocks",
+		fromHeight, finalRemoved, pendingRemoved)
+
+	return nil
+}
+
+// removeForBlockID remove block execution related data for a given block.
+func removeForBlockID(
```
zhangchiqing, line 182: I have addressed all the review comments. 

The `removeForBlockID` will remove all the results related data including events, serviceEvents, etc. 

And I've verified with a test case that save the results data, and remove it, and then save again. Also tested unhappy path that save again with a different result to simulate the case where there is an execution fork.

Tests are passing. 

Also tested with localnet, rollback to a height, and after starting backup, the node is able to re-execute blocks, and continue executing new blocks. 

cc @AlexHentschel

AlexHentschel (approved):
Nice work! Very clean. Great Tests 👏

PR #2336, opened 2022-04-21
Advise Linux to drop files from page cache (it was holding 425 GB of which 198+ GB were 3 checkpoint files)

fxamacker:
This PR advises Linux to evict entire files from its page cache without requiring `/bin/dd` or cgo.  

Changes include:

- At startup, advise Linux to evict all checkpoint files (up to 5 x 66GB) from Linux page cache.
- Use sys/unix package to advise Linux to evict from cache:
  - existing checkpoint file after it is loaded
  - new checkpoint file after it is created

This prevents 132+ GB from accumulating in Linux page cache after each checkpointing.

### Next Steps

A separate PR will advise Linux to evict segments of checkpoint files during use, which will provide more fine-grained control   as described in Other Approaches section of PR #2280.


ledger/complete/wal/fadvise.go
```diff
@@ -0,0 +1,44 @@
...
+// fadviseSegmentNoLinuxPageCache advises Linux to evict the
+// file segment from Linux page cache.  This calls unix.Fadvise
+// which in turn calls posix_fadvise with POSIX_FADV_DONTNEED.
+// If GOOS != "linux" then this function does nothing.
+// CAUTION: If fsync=true, this will call unix.Fsync which
+// can cause performance hit especially when used inside a loop.
+func fadviseSegmentNoLinuxPageCache(fd uintptr, off, len int64, fsync bool) error {
+	if runtime.GOOS != "linux" {
+		return nil
+	}
```
turbolent, line 28: Maybe just use a build tag instead to move this check to compile-time?

fxamacker: Great suggestion! 👍 Will do.

ledger/complete/wal/syncrename.go
```diff
@@ -53,6 +53,13 @@ func (s *SyncOnCloseRenameFile) Close() error {
 		return fmt.Errorf("cannot sync file %s: %w", s.file.Name(), err)
 	}
 
+	// s.file.Sync() was already called, so we pass fsync=false
+	err = evictFileFromLinuxPageCache(s.file, false, s.logger)
+	if err != nil && s.logger != nil {
```
zhangchiqing, line 58:
```suggestion
	if err != nil  {
```

I don't think we need to handle the `logger == nil` case. The logger should always be available, if not, it's a bug. 

Same in other places.

ledger/complete/wal/checkpointer.go
```diff
@@ -773,50 +776,75 @@ func readCheckpointV5(f *os.File) ([]*trie.MTrie, error) {
...
 
-	cmd := exec.Command(cmdFileName, "if="+s, "iflag=nocache", "count=0")
+	return evictFileFromLinuxPageCache(f, fsync, logger)
+}
 
-	if logger != nil {
-		logger.Info().Msgf("run %q to drop file from OS file cache", cmd.String())
+// evictFileFromLinuxPageCache advises Linux to evict a file from Linux page cache.
+// A use case is when a new checkpoint is loaded or created, Linux may cache big
+// checkpoint files in memory until evictFileFromLinuxPageCache causes them to be
```
zhangchiqing, line 829:
```suggestion
// checkpoint files in memory until evictFileFromLinuxPageCache is called to force them to be
```

zhangchiqing (approved):

ledger/complete/wal/fadvise.go
fxamacker: Done in commit 15cdac5fe04f3997141e1ddc8a8e02f86987d0f0.

ledger/complete/wal/checkpointer.go
fxamacker: `evictFileFromLinuxPageCache` calls `posix_fadvise` under the hood.  And `posix_fadvise` docs say that it's not binding advice and merely constitutes an expectation on behalf of the application.

In benchnet tests, the file is always removed from Linux page cache after calling `evictFileFromLinuxPageCache`.

ledger/complete/wal/syncrename.go
fxamacker: Thanks for the suggestion! :+1:   Done in commit dd7d9551f630bda3c693aa0e9f931b64d44dae6a.

ramtinms (approved):

PR #2346, opened 2022-04-25
Permissionless contract updates

turbolent:
We are planning to enable permissionless deployment (disable `restrictedDeploymentEnabled`) after a test period after the Secure Cadence spork.

However, Secure Cadence will also make breaking changes. To allow developers with existing contracts to update their contracts we will allow *updates* to always be permissionless, and have `restrictedDeploymentEnabled` only contract *new* deployments.

### DIFF
--- a/fvm/fvm_test.go
+++ b/fvm/fvm_test.go
@@ -549,7 +549,7 @@ func TestBlockContext_DeployContract(t *testing.T) {
 
 		assert.Error(t, tx.Err)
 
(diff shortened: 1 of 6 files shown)


janezpodhostnik (approved):
Nice change!

One thing I was thinking about during this review is that if both contract updates and contract  deployments were controllable by some on chain thing, this update could be rolled out without a spork. 

But I guess this is not relevant as we want to tie this together with secure cadence deploy right?

ramtinms (approved):
Looks good to me

PR #2349, opened 2022-04-26
[Component Manager] Fix race condition

smnzhu:
closes https://github.com/dapperlabs/flow-go/issues/6220

Fixes race condition introduced in https://github.com/onflow/flow-go/pull/1543


peterargue (commented):
can you add some notes about what the race condition is as comments (or in the PR description).

module/component/component.go
```diff
@@ -218,14 +218,14 @@ func (c *ComponentManager) Start(parent irrecoverable.SignalerContext) {
 		// goroutine and the parent's are scheduled. If the parent is scheduled first, any errors
 		// thrown within workers would not have propagated, and it would only receive the done signal
 		defer func() {
+			cancel() // shutdown all workers
+			<-c.shutdownSignal
```
jordanschalm, line 222:
```suggestion
			cancel() // shutdown all workers
			// wait for shutdown signal before signalling the component is done
			// this guarantees that ShutdownSignal is closed before Done
			<-c.shutdownSignal
```

module/component/component_manager_test.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/onflow/flow-go/utils/unittest"
 )
 
-const CHANNEL_CLOSE_LATENCY_ALLOWANCE = 20 * time.Millisecond
+const CHANNEL_CLOSE_LATENCY_ALLOWANCE = 25 * time.Millisecond
```
jordanschalm, line 20: Should we add the test reproducing the race? It's `TestComponentManagerShutdown` in https://github.com/onflow/flow-go/commit/803b173a753dffe658c9c0695f59cff5a1f2d893

jordanschalm (commented):
Thanks for the fix Simon, this looks good to me.

I tested with [the case](https://github.com/onflow/flow-go/commit/803b173a753dffe658c9c0695f59cff5a1f2d893) I was using to reproduce earlier, and it passes now. I think we should add that test case in this PR as well.

smnzhu:
> can you add some notes about what the race condition is as comments (or in the PR description).

@peterargue See the issue referenced in the description: https://github.com/dapperlabs/flow-go/issues/6220

peterargue (approved):
Looks good. Probably worth including Jordan's script.

jordanschalm (approved):

smnzhu:
bors merge

PR #2355, opened 2022-04-28
Don't push :latest tag by default

m4ksio:
Don't push the latest docker tag for by default.


peterargue (approved):
looks good. couple of thoughts:
* do we use the `latest` tagged images from the registry for anything?
* if we don't you could probably just get rid of the `*-latest` targets. if someone really wants to tag latest, they could `make docker-push-* IMAGE_TAG=latest`

m4ksio:
> looks good. couple of thoughts:
> 
> * do we use the `latest` tagged images from the registry for anything?

I'm not sure really. I would assume some local operations might, but this change shouldn't affect local images

> * if we don't you could probably just get rid of the `*-latest` targets. if someone really wants to tag latest, they could `make docker-push-* IMAGE_TAG=latest`

Hmm, yes, but I think its easier to have a new target, plus wont this approach push docker image with latest twice instead of proper image version?

sjonpaulbrown (approved):

PR #2357, opened 2022-04-29
Add weights for atree values

dsainati1:
This removes weights for array and dictionary length, since this was the old method of metering these values. Instead, we add weights for the underlying atree representation of these values.


fvm/meter/weighted/meter.go
```diff
@@ -56,6 +54,11 @@ var (
 		common.MemoryKindBoundFunction:           25,
 		common.MemoryKindBigInt:                  50,
 
+		common.MemoryKindAtreeArrayDataSlab:     72,
+		common.MemoryKindAtreeArrayMetaDataSlab: 1024,
+		common.MemoryKindAtreeMapDataSlab:       72,
+		common.MemoryKindAtreeMapMetaDataSlab:   1024,
```
dsainati1, line 60: @fxamacker I added these weights based on our slack discussion, but let me know if these values seem correct to you.

For context for others, we previously metered length (slabs) as if they consumed 1kb of space. In actuality this value is how much data they can hold, not how much memory require in and of themselves. Instead we should just meter the overhead of the data slabs themselves.

fxamacker: @dsainati1 I think the initial estimates for atree slabs are:

```
- atree array DataSlab overhead: 72 bytes
- atree array MetaDataSlab: 1 KiB
- atree map DataSlab overhead: 140 bytes
- atree map element (key+value) overhead: 64 bytes
- atree array MetaDataSlab: 1 KiB 
```

But based on @SupunS findings on Friday, I am going to look into the 720-byte discrepancy for empty struct.  So we probably will tweak the overhead listed above.

dsainati1: There is a separate overhead for each element of the map?

fxamacker: > There is a separate overhead for each element of the map?

Yes, because we have to store map key's hash digest (and related overhead) for retrieval.

dsainati1: Hmm, sounds like we would need another memory kind for this, unless you think we can just fold that overhead into the cost of MapDataSlab?

fxamacker: Yeah, I think it might be easier to have another memory kind, especially because we are still tweaking weight for each type.

turbolent (approved):

janezpodhostnik (approved):

fvm/meter/weighted/meter.go
```diff
@@ -36,12 +36,10 @@ var (
...
-		common.MemoryKindArrayBase:               57,
-		common.MemoryKindArrayLength:             1024,
-		common.MemoryKindDictionaryBase:          33,
-		common.MemoryKindDictionarySize:          1024,
-		common.MemoryKindCompositeBase:           233,
-		common.MemoryKindCompositeSize:           1024,
+		common.MemoryKindArrayBase:               57 + 48,
+		common.MemoryKindDictionaryBase:          33 + 96,
+		common.MemoryKindCompositeBase:           233 + 96,
+		common.MemoryKindSimpleComposite:         73 + 96,
```
fxamacker, line 42: Not sure, but I was under the impression we are not going to use atree map for `SimpleComposite` type.

fxamacker (commented):
Nice work!  Just left a comment about weight of `SimpleComposite`.

fvm/meter/weighted/meter.go
dsainati1: oh yea good point, that's my mistake.

fvm/meter/weighted/meter.go
```diff
@@ -36,12 +36,10 @@ var (
...
 		common.MemoryKindNumber:                  8,
-		common.MemoryKindArrayBase:               57,
-		common.MemoryKindArrayLength:             1024,
-		common.MemoryKindDictionaryBase:          33,
-		common.MemoryKindDictionarySize:          1024,
-		common.MemoryKindCompositeBase:           233,
-		common.MemoryKindCompositeSize:           1024,
+		common.MemoryKindArrayBase:               57 + 48,
+		common.MemoryKindDictionaryBase:          33 + 96,
+		common.MemoryKindCompositeBase:           233 + 96,
```
turbolent, line 41: It seems like the amounts are "split" into two parts for some reason – if possible, add a comment explaining why

turbolent (approved):
Looks good

go.mod
```diff
@@ -95,3 +95,4 @@ require (
 )
 
 replace mellium.im/sasl => github.com/mellium/sasl v0.2.1
+replace github.com/onflow/cadence => ../cadence
```
turbolent, line 98: This should not be committed

```suggestion
```

turbolent (approved):
Looks good, feel free to merge once the temporary replace statement is removed

fvm/meter/weighted/meter.go
turbolent: Thank you for adding the comment 👌

SupunS (approved):

PR #2362, opened 2022-05-02
Add checksum to signer indices

zhangchiqing:
Close https://github.com/dapperlabs/flow-go/issues/6179

This PR adds checksum to signer indices. 

### Why adding checksum?
After using signer indices to replace signer IDs, we lose certain level of debuggability. If we run into invalid QC issue, we can't exclude the case where the Committee.Identities had a bug that not always return the same list for the same block.
The checksum could provide the same debuggability as when signerIDs was used. Since for the above case, it's easy to confirm if the same list of signerIDs were used before matching sig type and verifying the signatures.

### How checksum is added
Checksum is first computed from the canonical list of identities that are used for generating signer indices.
The computed checksum will be prefixed to the generated signer indices, so that the decoder can use it to compare with it's own canonical list to check if both the encoder and decoder see the same canonical identity list

### DIFF
--- a/engine/execution/execution_test.go
+++ b/engine/execution/execution_test.go
@@ -128,7 +128,7 @@ func TestExecutionFlow(t *testing.T) {
 	// we need 1 byte
 	child.Header.ParentVoterIndices = voterIndices
 
-	log.Info().Msgf("child block ID: %v, indices: %v", child.Header.ID(), child.Header.ParentVoterIndices)
+	log.Info().Msgf("child block ID: %v, indices: %x", child.Header.ID(), child.Header.ParentVoterIndices)
 
 	collectionNode := testutil.GenericNodeFromParticipants(t, hub, colID, identities, chainID)
 	defer collectionNode.Done()
@@ -282,9 +282,6 @@ func makePanicBlock(t *testing.T, conID *flow.Identity, colID *flow.Identity, ch
 	// make collection
 	col := &flow.Collection{Transactions: []*flow.TransactionBody{tx}}
 
-	signerIndices, err := signature.EncodeSignersToIndices(
-		[]flow.Identifier{conID.NodeID}, []flow.Identifier{conID.NodeID})
-	require.NoError(t, err)
 	clusterChainID := cluster.CanonicalClusterID(1, flow.IdentityList{colID})
 	// make block
 	block := unittest.BlockWithParentAndProposerFixture(parent, conID.NodeID, 1)
@@ -292,6 +289,11 @@ func makePanicBlock(t *testing.T, conID *flow.Identity, colID *flow.Identity, ch
 		[]flow.Identifier{conID.NodeID}, []flow.Identifier{conID.NodeID})
 	require.NoError(t, err)
 	block.Header.ParentVoterIndices = voterIndices
+
+	signerIndices, err := signature.EncodeSignersToIndices(
+		[]flow.Identifier{colID.NodeID}, []flow.Identifier{colID.NodeID})
+	require.NoError(t, err)
+
 	block.SetPayload(flow.Payload{
 		Guarantees: []*flow.CollectionGuarantee{
 			{CollectionID: col.ID(), SignerIndices: signerIndices, ChainID: clusterChainID, ReferenceBlockID: ref.ID()},
@@ -310,7 +312,7 @@ func makeSuccessBlock(t *testing.T, conID *flow.Identity, colID *flow.Identity, 
 	require.NoError(t, err)
 
 	signerIndices, err := signature.EncodeSignersToIndices(
-		[]flow.Identifier{conID.NodeID}, []flow.Identifier{conID.NodeID})
+		[]flow.Identifier{colID.NodeID}, []flow.Identifier{colID.NodeID})
 	require.NoError(t, err)
 	clusterChainID := cluster.CanonicalClusterID(1, flow.IdentityList{colID})
 
@@ -331,21 +333,9 @@ func makeSuccessBlock(t *testing.T, conID *flow.Identity, colID *flow.Identity, 
 	return tx, col, block, proposal, seq + 1
 }
 
-// Test the following behaviors:
-// (1) ENs sync statecommitment with each other
-// (2) a failed transaction will not change statecommitment
-//
-// We prepare 3 transactions in 3 blocks:
-// tx1 will deploy a contract
-// tx2 will always panic
-// tx3 will be succeed and change statecommitment
-// and then create 2 EN nodes, both have tx1 executed. To test the synchronization,
-// we send tx2 and tx3 in 2 blocks to only EN1, and check that tx2 will not change statecommitment for
-// verifying behavior (1);
-// and check EN2 should have the same statecommitment as EN1 since they sync
-// with each other for verifying behavior (2).
-// TODO: state sync is disabled, we are only verifying 2) for now.
-func TestExecutionStateSyncMultipleExecutionNodes(t *testing.T) {
+// Test a successful tx should change the statecommitment,
+// but a failed Tx should not change the statecommitment.
+func TestFailedTxWillNotChangeStateCommitment(t *testing.T) {
 	hub := stub.NewNetworkHub()
 
 	chainID := flow.Emulator
--- a/engine/execution/execution_test.go
+++ b/engine/execution/execution_test.go
@@ -2,6 +2,7 @@ package execution_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
@@ -16,6 +17,7 @@ import (
 	"github.com/onflow/flow-go/engine/testutil"
 	testmock "github.com/onflow/flow-go/engine/testutil/mock"
 	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/model/flow/order"
 	"github.com/onflow/flow-go/model/messages"
 	"github.com/onflow/flow-go/module/signature"
 	"github.com/onflow/flow-go/network/mocknetwork"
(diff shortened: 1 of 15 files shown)


module/signature/checksum.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// CheckSumLen is fixed to be 16 bytes, which is also md5.Size
+const CheckSumLen = 32
+
+// CheckSumFromIdentities returns checksum for the given identities
+func CheckSumFromIdentities(identities []flow.Identifier) [CheckSumLen]byte {
+	return flow.MakeID(EncodeIdentities(identities))
```
zhangchiqing, line 15: I tried to use `crypto/md5.Sum` method, but the linter complains it's too weak, so I switched to sha3.

zhangchiqing: I'm still working on the test cases. But I would like to get some opinions on the choice of
1. using `flow.MakeID` to compute the check sum
2. encodes identities by concatenating the bytes.

@AlexHentschel @tarakby

tarakby: To confirm the use-case of the checksum, is this only a measure to find honest bugs? Or are we considering malicious behaviours when a malicious QC would target a collision (find 2 list of identities that have the same checksum, so that the receiver doesn't catch the inconsistency) ?

tarakby: I'll continue the review depending on the answer to the question :)

zhangchiqing: The checksum is for debugging. 

Even if such collision (2 list of identifiers having the same checksum) exists, the receiver will still reject the malicious QC because the signature will not match with the signer. And the block will be rejected. This is not supposed to happen in a non-BFT environment, except for software bug, in which case, the checksum will be used to debug. 

That being said, this checksum is only useful in non-BFT environment for debugging purpose. In a BFT environment, the checksum is not very useful.

module/signature/checksum.go
```diff
@@ -0,0 +1,74 @@
...
```
tarakby, line 15: I suggest to use a CRC of 32 bits as a checksum since the purpose is only to detect honest mistakes (bugs).  You can use this function to return the checksum : https://pkg.go.dev/hash/crc32@go1.18.1#ChecksumIEEE.
It returns a `uint32` but you can convert it into `[]byte` if needed.

module/signature/checksum.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// CheckSumLen is fixed to be 16 bytes, which is also md5.Size
+const CheckSumLen = 32
```
tarakby, line 11: If we use CRC32, the length would be 4. 

```suggestion
// CheckSumLen is fixed to be 4 bytes
const CheckSumLen = 4
```

module/signature/checksum.go
tarakby: The encoding into bytes (point 2.) is fine 👌🏼

module/signature/checksum_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+
+	t.Run("different identities, different checksum", func(t *testing.T) {
+		ids := unittest.IdentifierListFixture(4)
+		require.NotEqual(t, signature.CheckSumFromIdentities(ids), signature.CheckSumFromIdentities(ids[1:]))     // subset
+		require.NotEqual(t, signature.CheckSumFromIdentities(ids[1:]), signature.CheckSumFromIdentities(ids[:2])) // overlap
+		require.NotEqual(t, signature.CheckSumFromIdentities(ids[:2]), signature.CheckSumFromIdentities(ids[2:])) // no overlap
+	})
+
+	t.Run("checksum length always 16", func(t *testing.T) {
```
tarakby, line 35:
```suggestion
	t.Run("checksum length always constant", func(t *testing.T) {
```

engine/execution/execution_test.go
```diff
@@ -57,7 +59,7 @@ func TestExecutionFlow(t *testing.T) {
 		unittest.WithKeys,
 	)
 
-	identities := unittest.CompleteIdentitySet(colID, conID, exeID, verID)
+	identities := unittest.CompleteIdentitySet(colID, conID, exeID, verID).Sort(order.Canonical)
```
zhangchiqing, line 62: Tests were flakey because the identities were not sorted. 

It's now fixed by sorting it first

engine/execution/execution_test.go
```diff
@@ -95,8 +97,14 @@ func TestExecutionFlow(t *testing.T) {
...
-	block.Header.ParentVoterIndices = unittest.SignerIndicesByIndices(1, []int{0})
-	signerIndices := unittest.SignerIndicesByIndices(1, []int{0})
+	voterIndices, err := signature.EncodeSignersToIndices(
+		[]flow.Identifier{conID.NodeID}, []flow.Identifier{conID.NodeID})
+	require.NoError(t, err)
+	block.Header.ParentVoterIndices = voterIndices
+	signerIndices, err := signature.EncodeSignersToIndices(
+		[]flow.Identifier{colID.NodeID}, []flow.Identifier{colID.NodeID})
+	require.NoError(t, err)
+	fmt.Println("===========", fmt.Sprintf("%x", voterIndices))
```
durkmurder, line 107: it is a leftover?

engine/execution/ingestion/engine.go
```diff
@@ -455,7 +455,7 @@ func (e *Engine) handleBlock(ctx context.Context, block *flow.Block) error {
 	})
 
 	if err != nil {
-		return fmt.Errorf("could not enqueue block: %w", err)
+		return fmt.Errorf("could not enqueue block %v: %w", blockID, err)
```
durkmurder, line 458:
```suggestion
		return fmt.Errorf("could not enqueue block %x: %w", blockID, err)
```

integration/testnet/network.go
```diff
@@ -1198,6 +1199,11 @@ func BootstrapNetwork(networkConf NetworkConfig, bootstrapDir string) (*Bootstra
 		return nil, fmt.Errorf("could not create bootstrap state snapshot: %w", err)
 	}
 
+	err = badger.IsValidRootSnapshotQCs(snapshot)
```
durkmurder, line 1202: :+1:

integration/testnet/network.go
```diff
@@ -1323,19 +1330,22 @@ func setupClusterGenesisBlockQCs(nClusters uint, epochCounter uint64, confs []Co
...
 			}
 		}
-		if len(cluster) != len(participants) { // sanity check
+		if len(cluster) != len(clusterNodeInfos) { // sanity check
 			return nil, nil, nil, fmt.Errorf("requiring a node info for each cluster participant")
 		}
 
 		// generate qc for root cluster block
-		qc, err := run.GenerateClusterRootQC(participants, bootstrap.ToIdentityList(participants), block)
+		fmt.Println("clusterNodeInfos", clusterNodeInfos)
```
durkmurder, line 1345: do we need this print?

module/signature/checksum.go
```diff
@@ -0,0 +1,100 @@
...
+	signerIndices := checkSumPrefixedSignerIndices[CheckSumLen:]
+
+	return sum, signerIndices, nil
+}
+
+// CompareAndExtract reads the checksum from the given checkSumPrefixedSIgnerIndices
+// bytes, and compare with the checksum of the given identifier list.
+// it returns the signer indices if the checksum matches.
+// it returns error if splitting the checksum fails or
+// 										 the splitted checksum doesn't match
```
durkmurder, line 74:
```suggestion
// the splitted checksum doesn't match
```

module/signature/checksum.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	var sum [CheckSumLen]byte
+	copy(sum[:], checkSumPrefixedSignerIndices[:CheckSumLen])
+	signerIndices := checkSumPrefixedSignerIndices[CheckSumLen:]
+
+	return sum, signerIndices, nil
+}
+
+// CompareAndExtract reads the checksum from the given checkSumPrefixedSIgnerIndices
```
durkmurder, line 70:
```suggestion
// CompareAndExtract reads the checksum from the given checkSumPrefixedSignerIndices
```

module/signature/checksum_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/stretchr/testify/require"
+	"pgregory.net/rapid"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/signature"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestCheckSum(t *testing.T) {
```
durkmurder, line 14: can we add minimal godoc for these tests?

tarakby (approved):

module/signature/checksum.go
```diff
@@ -0,0 +1,99 @@
...
+
+// CheckSumLen is fixed to be 4 bytes
+const CheckSumLen = 4
+
+func checksum(data []byte) [CheckSumLen]byte {
+	// since the checksum is only for detecting honest mistake,
+	// crc32 is enough
+	sum := crc32.ChecksumIEEE(data)
+	// converting the uint32 checksum value into [4]byte
+	return [CheckSumLen]byte{
```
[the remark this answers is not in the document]
durkmurder: I would vote for Tarak's suggestion, more obvious so to say

durkmurder (approved):
Very solid. Looks good to me.

zhangchiqing:
bors merge

PR #2363, opened 2022-05-03
Use go 1.18 to build and run binary

Kay-Zee:
Recently, had a discussion about moving up to 1.18, as there are some optimizations, plus enables the ability to explore generics for future features.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
       - 'v[0-9]+.[0-9]+'
 
 env:
-  GO_VERSION: 1.17
+  GO_VERSION: 1.18
 
 jobs:
   golangci:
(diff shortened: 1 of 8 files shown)


jordanschalm (commented):
Nice, I guess the new mockery library fixes the [mock generation issues we had earlier](https://axiomzen.slack.com/archives/C015G65HR2P/p1650489831418979)?

AlexHentschel (approved):
Thanks for the upgrade.
Looks good from my end. Its probably best to make sure @yhassanzadeh13 looks over the networking changes.

peterargue (approved):
Looks good. Thanks for figuring this all out @Kay-Zee!

Not specific to this PR, but it looks like we've stopped testing against multiple versions of go. I think we should at least run unit and integration tests against all versions we official support in `go.mod`

zhangchiqing (approved):

yhassanzadeh13 (approved):
Checked the networking-related changes and they look good, thanks for refactoring.

PR #2365, opened 2022-05-03
add runtime memory metering to test on mainnet

j1010001:
DO NOT MERGE


Kay-Zee:
Tested and deployed, closing

PR #2367, opened 2022-05-03
[Consensus] Block header change

durkmurder:
### Context

As short-term measure to provide an easy and reliable way for syncing blocks for Active Pacemaker without much effort we have decided to include `LastViewTC` in `flow.Header` which will allow us to use current sync engine to sync block proposals. 

For long-term solution we will implement a request-reply mechanism for syncing last proposals, this will allow us to remove `LastViewTC` from header.

As for know including `LastViewTC` into header is the easiest solution to achieve block <-> proposal conversion without writing new modules.


model/flow/header.go
```diff
@@ -43,6 +44,9 @@ type Header struct {
 
 	ProposerSigData []byte // signature of the proposer over the new block. Not a single cryptographic
 	// signature since the data represents cryptographic signatures serialized in some way (concatenation or other)
+
+	LastViewTC *TimeoutCertificate // Timeout certificate for previous view, it can be nil
+	// it has to be present if previous round ended with timeout.
```
jordanschalm, line 49:
```suggestion
	// LastViewTC is the timeout certificate for previous view. It can be nil and
	// has to be present if previous round ended with timeout.
	LastViewTC *TimeoutCertificate 
```

Not a problem with this PR, just noticed it here: we aren't formatting these comments properly as GoDoc. Comments for a field shouldn't be below the field, otherwise they aren't associated with that field and don't show in documentation tools ([godoc site example](https://pkg.go.dev/github.com/onflow/flow-go/model/flow#Header)). 

For such an important model, I think it's worthwhile to fix all the fields here.

jordanschalm (approved):

AlexHentschel (approved):
🎉

PR #2368, opened 2022-05-03
Update cd.yml to ignore daily tags created through automation

sjonpaulbrown:
## Description
Exclude daily tags that will be created on a schedule. We do not want these to kickoff new container builds using this workflow.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -4,7 +4,7 @@ on:
   push:
     tags: 
       - '*'
-      - "daily-*"
+      - "!daily-*"
 
 jobs:
   docker-push:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     tags: 
       - '*'
+      - "daily-*"
```
Kay-Zee, line 7: this looks like it's just adding `daily-*`, aren't suppose to have something indicating that it should be excluded, rather than included?

sjonpaulbrown: I missed this when I pushed an update. This should be `!daily-*`

Kay-Zee (approved):

PR #2371, opened 2022-05-03
[DO NOT MERGE] avoid race condition in validating blocks

zhangchiqing:


engine/common/synchronization/engine.go
```diff
@@ -310,6 +310,9 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockRe
 			OriginID: originID,
 			Block:    block,
 		}
+		// tempfix: to help nodes falling far behind to catch up.
+		// it avoids the race condition in compliance engine and hotstuff to validate blocks
+		time.Sleep(150 * time.Millisecond)
```
jordanschalm, line 315: Should this go in `compliance/core.go` instead? (eg. [immediately after passing the block to HotStuff](https://github.com/onflow/flow-go/blob/5d1c56881355211d1502c12de749fb832ae99ac8/engine/consensus/compliance/core.go#L346)). This is introducing the delay while adding a block to the queue, but if there are any blocks in the queue already, we aren't guaranteed to see this delay between the compliance layer validation and HotStuff validation.

AlexHentschel, line 315: Looks good. 
Hopefully this will give us a good test signal. For production, a lower value 50ms or 100ms might yield better tradeoff.

AlexHentschel (approved):

jordanschalm (commented):
Just a comment on ensuring we're always seeing the delay where we expect. For a hotfix to validate the problem, I think the current approach is OK as well.

engine/common/synchronization/engine.go
AlexHentschel: I am worried about putting it in the hot path. Every block needs to pass through this and I am worried that it might impact nodes on the happy path. In contrast, Leo has put the logic in the sync engine. The live blocks should already have a latency of more than 150ms and we only throttle the sync engine as the known source of the problem.

PR #2372, opened 2022-05-04
Outdated blocks should not be logged as errors

AlexHentschel:
Conceptually, this is expected during normal operations that some _other_ node has fallen behind and sending outdated messages, such as block proposals.  Despite this being an expected scenario, outdated blocks as logged as errors ([code](https://github.com/onflow/flow-go/blob/f93a31395b2b780f2cbabba1c5303e2fe43d745b/engine/common/follower/engine.go#L349-L357): errors are bubbled up the call stack, where they arrive at the networking layer, which just logs them.) We have seen that such harmless conditions can be distracting during incidents [slack](https://axiomzen.slack.com/archives/CEEGK3HGC/p1651601982120259?thread_ts=1651599613.010539&cid=CEEGK3HGC) potentially sending engineers down the wrong rabbit hole. 

This PR attempts a preliminary fix and only handle this particular case of outdated blocks. Generally, the Flower engine is older code which lacks adequate error handling, including the protocol. So there is much more work beyond this PR.

### DIFF
--- a/engine/common/follower/engine.go
+++ b/engine/common/follower/engine.go
@@ -313,6 +313,9 @@ func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.Bl
 // the finalized state; if a parent of children is validly processed, it means
 // the children are also still on a valid chain and all missing links are there;
 // no need to do all the processing again.
+// Expected errors during normal operations:
+// * engine.OutdatedInputError if the block proposal is orphaned
+// * engine.InvalidInputError if the block proposal is invalid
 func (e *Engine) processBlockProposal(ctx context.Context, proposal *messages.BlockProposal) error {
 
 	span, ctx := e.tracer.StartSpanFromContext(ctx, trace.FollowerProcessBlockProposal)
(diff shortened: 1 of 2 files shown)


engine/common/follower/engine.go
```diff
@@ -344,19 +347,17 @@ func (e *Engine) processBlockProposal(ctx context.Context, proposal *messages.Bl
...
-	}
-
-	// if the error is a known outdated extension of the protocol state, then
-	// the input is outdated
-	if state.IsOutdatedExtensionError(err) {
-		return engine.NewOutdatedInputErrorf("outdated extension of protocol state: %w", err)
-	}
-
 	if err != nil {
+		// The proposal must represent a valid extension of the protocol state, otherwise it is invalid.
```
jordanschalm, line 351: It might be clearer to use the same log levels and messages here as the other engines where this case is handled:
https://github.com/onflow/flow-go/blob/5d1c56881355211d1502c12de749fb832ae99ac8/engine/consensus/compliance/core.go#L253-L262

AlexHentschel: 👍 implemented in https://github.com/onflow/flow-go/pull/2372/commits/412fd2a54fa706a5773c9e4bf3ea58d646ae546f

Only comment: it is not _exactly_ the same log message, as the same symptom has different explanations from the view point of the node dealing with the orphaned block

durkmurder (approved):

PR #2373, opened 2022-05-04
add jobqueue README

zhangchiqing:
Adding a README for jobqueue to explain its design and patterns.


module/jobqueue/README.md
```diff
@@ -0,0 +1,63 @@
...
+When Job consumer finds new job, it uses the Worker interface to process each job. Each job is also an interface, so it's user's responsiblity to handle the conversion between job and the actual job data type. So that the modules.Job can return a job to converted back to the actual job data for worker to work on.
+
+In the scenario of processing finalized blocks, implementing symmetric functions like BlockToJob and JobToBlock are recommended for this conversion.
+
+In order to report job completion, the worker needs to call job consumer's NotifyJobIsDone method.
+
+## Pipeline Pattern
+Multiple jobqueues can be combined to form a pipeline. This is useful in the scenario that the first job queue will process each finalized block and create jobs to process data depending on the block, and having the second job queue to process each job created by the worker of the first job queue.
+
+For instance, verification node uses 2-jobqueue pipeline to find chunks from each block and create jobs if the block has chunks that it needs to verify, and the second job queue will allow verification node to verify each chunk with a max number of workers.
```
peterargue, line 63:
```suggestion
For instance, verification node uses 2-jobqueue pipeline to find chunks from each block and create jobs if the block has chunks that it needs to verify, and the second job queue will allow verification node to verify each chunk with a max number of workers.

## Considerations

### Push vs Pull
The jobqueue architecture is optimized for "pull" style processes, where the consumer pulls jobs from a source when workers are available. All current implementations are using this style since it lends well to asynchronously processing jobs based on block heights.

Some use cases require "push" style jobs where there is a job producer that creates jobs, and a consumer that processes work from the producer. This is possible with the jobqueue, but requires the producer persist the jobs to a database, then implement the `Head` and `AtIndex` methods that allow accessing jobs by sequential `uint64` indexes.

In existing pull use cases, the sync engine is effectively the job producer.
```

peterargue (approved):
looks great. added a few suggestions

module/jobqueue/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+The most common use case is to work on each finalized block async.
+
+For instance, the verification node verify each finalized block. The block processing needs to be async because if not, the verification node might get overwhelmed by the large amount of blocks to be processed when they got finalized quickly during the period that the node went back online and catching up blocks from other peers.
+
+So the goal for the jobqueue system are:
+1. guarantee each finalized block will be processed eventually
+2. in the event of a crash failure, the worker can be scheduled so that no finalized block will be skipped.
+3. allow concurrent processing of multiple jobs (i.e, finalized blocks)
+4. the number of concurrent workers are configable so that the node won't get overwhelmed when too many jobs are created (i.e too many blocks are finalized in a short period of time)
```
smnzhu, line 13:
```suggestion
4. the number of concurrent workers are configable so that the node won't get overwhelmed by too many jobs being created (i.e. too many blocks being finalized in a short period of time)
```

module/jobqueue/README.md
```diff
@@ -0,0 +1,63 @@
+## JobQueue Design Goal
+
+The jobqueue package implemented a reusable job queue system for async message processing.
+
+The most common use case is to work on each finalized block async.
+
+For instance, the verification node verify each finalized block. The block processing needs to be async because if not, the verification node might get overwhelmed by the large amount of blocks to be processed when they got finalized quickly during the period that the node went back online and catching up blocks from other peers.
```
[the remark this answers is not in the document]
zhangchiqing: >So is the job queue only creating a job for each new finalized block? or can it be used for anything?

The finalized block is the first scenario for the job queue. Since each finalized block can be indexed by height, we use a value uint64 value as job index to find each job. 

So it can be used in other things if jobs can be indexed by uint64 and are consecutive, which means it cannot be used for each incorporated block, because incorporated blocks cannot be indexed by height.

module/jobqueue/README.md
```diff
@@ -0,0 +1,70 @@
...
+3. max search ahead is reached (will explain later)
+
+`Check` method is concurrent safe, meaning even if job consumer is notified concurrently about new jobs available, job consumer will check at most once to find new jobs.
+
+Whenever a worker finishes a job, job consumer will also call `Check` internally.
+
+### Storing job consuming progress in storage.ConsumerProgress
+Job consumer stores the last processed job index in `storage.ConsumerProgress`, so that on startup, the job consumer can read the last processed job index from storage and compare with the last available job index from `module.Jobs`'s `Head` method to resume job processing.
+
+This ensures each job will be processed at least once. Note: given the at least once execution, the `Worker` should gracefully handle duplicate runs of the same job.
```
SaveTheRbtz, line 50: You may want to mention that on job execution failure process will die with `Fatal`

module/jobqueue/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Job consumer relies on the modules.Jobs to find jobs. However, modules.Jobs doesn't provide a way to notify as soon as a new job is available. So it's consumer's job to keep track of the values returned by module.Jobs's `Head` method and find jobs that are new.
+
+### Using Check method to notify job consumer for checking new jobs
+Job consumer provides the `Check` method for users to notify new jobs available.
+
+Once called, job consumer will iterate through each height with the `AtIndex` method. It stops when one of the following condition is true:
+1. no job was found at a index
+2. no more workers to work on them
+3. max search ahead is reached (will explain later)
```
SaveTheRbtz, line 41: Not sure what does max search ahead means here.  Looking at the code `maxProcessing` is the only limit there (which is the same as max number of workers).

SaveTheRbtz (approved):

module/jobqueue/README.md
zhangchiqing: Good catch.

The max search ahead [is implemented on a different branch,](https://github.com/onflow/flow-go/pull/1895/files#diff-2731207c7c63d225021f36118b9379bcb3e4394d9ef95e314020a55ff0715500R33) but not merged to master yet.

I added it to the TODOs, which are the features extended by @peterargue ['s PR. ](https://github.com/onflow/flow-go/pull/1895/files#diff-8eb19c06c1184bbe002ad8b1ccaf04394e602c64077c96943b2947757fb92c63)

zhangchiqing:
bors merge

PR #2374, opened 2022-05-04
[FVM] Merge default execution/memory weights and state weights

janezpodhostnik:
This concerns both Memory and Execution weights.

The default weights dictionary is hardcoded in the FVM in `fvm/meter/weighted/meter.go`.
The state weights dictionary is in the service account's storage: https://github.com/onflow/flow-core-contracts/blob/master/transactions/FlowServiceAccount/set_execution_effort_weights.cdc

Current logic is:
- In each transaction
- Check if the weights dictionary in the state exists
- If it does take the dictionary from state
- If not take the default dictionary

The change in this PR is:
- In each transaction
- Check if the weights dictionary in the state exists
- If it does, take the default weights dictionary and update it with all the entries from the state dictionary (merging them)
- If not take the default dictionary

The reasoning behind this is a very special scenario that might occur:
If someone sends a lot of transaction exploiting something that has a 0 weight (or no weight at all) and this causes a crash loop on the execution nodes (at the same height), one way out of this is to update the default weights and redeploy the nodes.

### DIFF
--- a/fvm/fvm_test.go
+++ b/fvm/fvm_test.go
@@ -3652,7 +3652,9 @@ func TestSettingExecutionWeights(t *testing.T) {
 		fvm.WithTransactionFee(fvm.DefaultTransactionFees),
 		fvm.WithExecutionEffortWeights(
 			weightedMeter.ExecutionEffortWeights{
-				common.ComputationKindStatement: 1 << weightedMeter.MeterExecutionInternalPrecisionBytes,
+				common.ComputationKindStatement:          1 << weightedMeter.MeterExecutionInternalPrecisionBytes,
+				common.ComputationKindLoop:               0,
+				common.ComputationKindFunctionInvocation: 0,
 			},
 		),
 	).withContextOptions(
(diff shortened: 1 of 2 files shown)


turbolent (approved):
Nice!

fvm/fvm_test.go
```diff
@@ -3652,7 +3652,9 @@ func TestSettingExecutionWeights(t *testing.T) {
 		fvm.WithTransactionFee(fvm.DefaultTransactionFees),
 		fvm.WithExecutionEffortWeights(
 			weightedMeter.ExecutionEffortWeights{
-				common.ComputationKindStatement: 1 << weightedMeter.MeterExecutionInternalPrecisionBytes,
+				common.ComputationKindStatement:          1 << weightedMeter.MeterExecutionInternalPrecisionBytes,
+				common.ComputationKindLoop:               0,
+				common.ComputationKindFunctionInvocation: 0,
```
ramtinms, line 3657: not sure why need these zero, wasn't the case these were counted as 1 before?

janezpodhostnik: The test is intended as if these were 0. Which means that after this change I need to explicitly set them to 0 in the state, to override the FWM hardcoded defaults (which are 1)

ramtinms (approved):

PR #2376, opened 2022-05-05
Allow disabling/enabling permissionless deployment with a transaction

janezpodhostnik:
The logic is:
 If there is no boolean at the expected location in the state, take the value from the context (`ctx.RestrictedDeploymentEnabled`). If there is a boolean at the expected location take that value instead.

This allows for a gradual transition from having this defined in the context to having it defined in the state.

The expected location in the state is on the service account  @`/storage/isContractDeploymentRestricted`


fvm/bootstrap.go
```diff
@@ -36,6 +36,9 @@ type BootstrapProcedure struct {
 	storagePerFlow                   cadence.UFix64
 	restrictedAccountCreationEnabled cadence.Bool
 
+	// TODO: restrictedContractDeployment should be a bool after RestrictedDeploymentEnabled is removed from the context
+	restrictedContractDeployment *bool
```
turbolent, line 40: Why is this a pointer to a bool, and not just a bool? Maybe add a comment explaining the three different states of this field

janezpodhostnik: added in [5221e00](https://github.com/onflow/flow-go/pull/2376/commits/5221e00a23f4b4cf5be62059433e7f38246e8546)

ramtinms (approved):
looks good to me,

turbolent (approved):

janezpodhostnik:
bors merge

j1010001 (approved):

PR #2379, opened 2022-05-06
changing the block height range for GetEventsForBlockHeightRange REST…

vishalchangrani:
… API call to match with the gRPC range limit

It was [reported](https://github.com/onflow/fcl-js/issues/1135#) that the GetEventForBlockHeight REST API call had a different limit than the one for the equivalent gRPC call.

I [found](https://github.com/onflow/fcl-js/issues/1135#issuecomment-1119849049) that the block height range limit for the REST API was 50 while the one for the gRPC API was 250.

This change does the following:
1. Separate out the range limits for the GetEventsForBlockHeight and GetBlocks REST API calls.
2. Define the limit for the `GetEventsForBlockHeight` as 250 (independently of the gRPC API constant incase it is needed to make the height ranges for REST and gRPC different in the future)

### DIFF
--- a/engine/access/rest/request/get_block.go
+++ b/engine/access/rest/request/get_block.go
@@ -9,7 +9,8 @@ import (
 const heightQuery = "height"
 const startHeightQuery = "start_height"
 const endHeightQuery = "end_height"
-const MaxAllowedHeights = 50
+const MaxAllowedBlockRequestHeightRange = 50
+const MaxAllowedEventRequestHeightRange = 250
 const idParam = "id"
 
 type GetBlock struct {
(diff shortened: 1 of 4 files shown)


peterargue (approved):
Thanks for the fix!

engine/access/rest/request/get_block.go
```diff
@@ -9,7 +9,7 @@ import (
 const heightQuery = "height"
 const startHeightQuery = "start_height"
 const endHeightQuery = "end_height"
-const MaxAllowedHeights = 50
+const MaxAllowedBlockRequestHeightRange = 50
```
sideninja, line 12: would the name `MaxRequestHeightRange` make sense (it's a bit more concise)?

vishalchangrani: make sense - I changed the names to `MaxBlockRequestHeightRange` and `MaxEventRequestHeightRange` to distinguish between the two types of request.

sideninja (approved):

vishalchangrani:
bors merge

PR #2381, opened 2022-05-06
[Collection] Log collection proposer

zhangchiqing:
This PR helps to answer the question: who created the cluster block that later turns into a collection guarantee.

To answer this, we need a log that ties the cluster block ID and the collection guarantee ID, and this PR adds it.

### DIFF
--- a/engine/collection/compliance/core.go
+++ b/engine/collection/compliance/core.go
@@ -89,7 +89,9 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Uint64("block_view", header.View).
 		Hex("block_id", logging.Entity(header)).
 		Hex("parent_id", header.ParentID[:]).
-		Hex("payload_hash", header.PayloadHash[:]).
+		Hex("ref_block_id", proposal.Payload.ReferenceBlockID[:]).
+		Hex("collection_id", logging.Entity(proposal.Payload.Collection)).
+		Int("tx_count", proposal.Payload.Collection.Len()).
 		Time("timestamp", header.Timestamp).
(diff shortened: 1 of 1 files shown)


engine/collection/compliance/core.go
```diff
@@ -89,7 +89,9 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Uint64("block_view", header.View).
 		Hex("block_id", logging.Entity(header)).
 		Hex("parent_id", header.ParentID[:]).
-		Hex("payload_hash", header.PayloadHash[:]).
+		Hex("ref_block_id", proposal.Payload.ReferenceBlockID[:]).
+		Hex("collection_id", logging.Entity(proposal.Payload.Collection)).
+		Int("tx_count", proposal.Payload.Collection.Len()).
```
zhangchiqing, line 94: we don't submit collection guarantee if the finalized cluster block has no tx. So it's useful to also log the tx count for debugging purpose.

engine/collection/compliance/core.go
```diff
@@ -89,7 +89,9 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Uint64("block_view", header.View).
 		Hex("block_id", logging.Entity(header)).
```
zhangchiqing, line 90: Since we are logging the `collection_id` and `block_id`, we can find what block is turned into a guarantee by following the `OnFinalizedBlock` log where it logs the finalized block id. Any collection_id associated with the finalized block id will be turned into a guarantee. Useful for debugging

jordanschalm (approved):

tarakby (approved):

PR #2385, opened 2022-05-09
Robust Transaction Deduplication

jordanschalm:
This PR improves transaction de-duplication in collection nodes. It replaces a heuristic which de-duplicated based on the contents of a fixed number of ancestor collections, with an reference height index which enables de-duplicating transactions correctly in all cases including when the collection production rate significantly outpaces the consensus block rate. 

See also https://github.com/onflow/flow-go/pull/2386 targeting the `v0.25` branch.

## Changes
* Add an index mapping reference block height to collection ID, for non-empty finalized collections
* Modify transaction inclusion validation when building and validating collections to take into account potentially conflicting collections at any point in the cluster state ancestry

### DIFF
--- a/module/builder/collection/builder.go
+++ b/module/builder/collection/builder.go
@@ -72,7 +72,7 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 
 	// first we construct a proposal in-memory, ensuring it is a valid extension
 	// of chain state -- this can be done in a read-only transaction
-	err := b.db.View(func(tx *badger.Txn) error {
+	err := b.db.View(func(btx *badger.Txn) error {
 
 		// STEP ONE: Load some things we need to do our work.
 		// TODO (ramtin): enable this again
@@ -80,7 +80,7 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 		// defer b.tracer.FinishSpan(parentID, trace.COLBuildOnSetup)
 
 		var parent flow.Header
-		err := operation.RetrieveHeader(parentID, &parent)(tx)
+		err := operation.RetrieveHeader(parentID, &parent)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve parent: %w", err)
 		}
@@ -88,19 +88,25 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 		// retrieve the height and ID of the latest finalized block ON THE MAIN CHAIN
 		// this is used as the reference point for transaction expiry
 		var refChainFinalizedHeight uint64
-		err = operation.RetrieveFinalizedHeight(&refChainFinalizedHeight)(tx)
+		err = operation.RetrieveFinalizedHeight(&refChainFinalizedHeight)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve main finalized height: %w", err)
 		}
 		var refChainFinalizedID flow.Identifier
-		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(tx)
+		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve main finalized ID: %w", err)
+		}
+		// pre-compute the minimum possible reference block height for transactions
+		// included in this collection (actual reference height may be greater)
+		minPossibleRefHeight := refChainFinalizedHeight - uint64(flow.DefaultTransactionExpiry-b.config.ExpiryBuffer)
+		if minPossibleRefHeight > refChainFinalizedHeight {
+			minPossibleRefHeight = 0 // overflow check
 		}
 
 		// retrieve the finalized boundary ON THE CLUSTER CHAIN
 		var clusterFinal flow.Header
-		err = procedure.RetrieveLatestFinalizedClusterHeader(parent.ChainID, &clusterFinal)(tx)
+		err = procedure.RetrieveLatestFinalizedClusterHeader(parent.ChainID, &clusterFinal)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve cluster final: %w", err)
 		}
@@ -126,30 +132,10 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 		// keep track of transactions to enforce rate limiting
 		limiter := newRateLimiter(b.config, parent.Height+1)
 
-		// look up previously included transactions in UN-FINALIZED ancestors
-		ancestorID := parentID
-		clusterFinalID := clusterFinal.ID()
-		for ancestorID != clusterFinalID {
-			ancestor, err := b.clusterHeaders.ByBlockID(ancestorID)
-			if err != nil {
-				return fmt.Errorf("could not get ancestor header (%x): %w", ancestorID, err)
-			}
-
-			if ancestor.Height <= clusterFinal.Height {
-				return fmt.Errorf("should always build on last finalized block")
-			}
-
-			payload, err := b.payloads.ByBlockID(ancestorID)
-			if err != nil {
-				return fmt.Errorf("could not get ancestor payload (%x): %w", ancestorID, err)
-			}
-
-			collection := payload.Collection
-			for _, tx := range collection.Transactions {
-				lookup.addUnfinalizedAncestor(tx.ID())
-				limiter.addAncestor(ancestor.Height, tx)
-			}
-			ancestorID = ancestor.ParentID
+		// first, look up previously included transactions in UN-FINALIZED ancestors
+		err = b.populateUnfinalizedAncestryLookup(parentID, clusterFinal.Height, lookup, limiter)
+		if err != nil {
+			return fmt.Errorf("could not populate un-finalized ancestry lookout (parent_id=%x): %w", parentID, err)
 		}
 
 		// TODO (ramtin): enable this again
@@ -157,35 +143,10 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 		// b.tracer.StartSpan(parentID, trace.COLBuildOnFinalizedLookup)
 		// defer b.tracer.FinishSpan(parentID, trace.COLBuildOnFinalizedLookup)
 
-		//TODO for now we check a fixed # of finalized ancestors - we should
-		// instead look back based on reference block ID and expiry
-		// ref: https://github.com/dapperlabs/flow-go/issues/3556
-		limit := clusterFinal.Height - flow.DefaultTransactionExpiry
-		if limit > clusterFinal.Height { // overflow check
-			limit = 0
-		}
-
-		// look up previously included transactions in FINALIZED ancestors
-		ancestorID = clusterFinal.ID()
-		ancestorHeight := clusterFinal.Height
-		for ancestorHeight > limit {
-			ancestor, err := b.clusterHeaders.ByBlockID(ancestorID)
-			if err != nil {
-				return fmt.Errorf("could not get ancestor header (%x): %w", ancestorID, err)
-			}
-			payload, err := b.payloads.ByBlockID(ancestorID)
-			if err != nil {
-				return fmt.Errorf("could not get ancestor payload (%x): %w", ancestorID, err)
-			}
-
-			collection := payload.Collection
-			for _, tx := range collection.Transactions {
-				lookup.addFinalizedAncestor(tx.ID())
-				limiter.addAncestor(ancestor.Height, tx)
-			}
-
-			ancestorID = ancestor.ParentID
-			ancestorHeight = ancestor.Height
+		// second, look up previously included transactions in FINALIZED ancestors
+		err = b.populateFinalizedAncestryLookup(minPossibleRefHeight, refChainFinalizedHeight, lookup, limiter)
+		if err != nil {
+			return fmt.Errorf("could not populate finalized ancestry lookup: %w", err)
 		}
 
 		// STEP THREE: build a payload of valid transactions, while at the same
@@ -196,14 +157,15 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 		// b.tracer.StartSpan(parentID, trace.COLBuildOnCreatePayload)
 		// defer b.tracer.FinishSpan(parentID, trace.COLBuildOnCreatePayload)
 
+		// keep track of the actual smallest reference height of all included transactions
 		minRefHeight := uint64(math.MaxUint64)
-		// start with the finalized reference ID (longest expiry time)
 		minRefID := refChainFinalizedID
 
 		var transactions []*flow.TransactionBody
 		var totalByteSize uint64
 		var totalGas uint64
 		for _, tx := range b.transactions.All() {
+			fmt.Println("considering transaction for inclusion: ", tx.ID())
 
 			// if we have reached maximum number of transactions, stop
 			if uint(len(transactions)) >= b.config.MaxCollectionSize {
@@ -252,7 +214,8 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 
 			// ensure the reference block is not too old
 			txID := tx.ID()
-			if refChainFinalizedHeight-refHeader.Height > uint64(flow.DefaultTransactionExpiry-b.config.ExpiryBuffer) {
+			if refHeader.Height < minPossibleRefHeight {
+				fmt.Println("expired tx", txID, refHeader.Height, minPossibleRefHeight)
 				// the transaction is expired, it will never be valid
 				b.transactions.Rem(txID)
 				continue
@@ -260,11 +223,13 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 
 			// check that the transaction was not already used in un-finalized history
 			if lookup.isUnfinalizedAncestor(txID) {
+				fmt.Println("confict with unfinalized ancestor: ", txID)
 				continue
 			}
 
 			// check that the transaction was not already included in finalized history.
 			if lookup.isFinalizedAncestor(txID) {
+				fmt.Println("confict with finalized ancestor: ", txID)
 				// remove from mempool, conflicts with finalized block will never be valid
 				b.transactions.Rem(txID)
 				continue
@@ -283,6 +248,8 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 
 			// update per-payer transaction count
 			limiter.transactionIncluded(tx)
+
+			fmt.Println("adding transaction: ", txID)
 
 			transactions = append(transactions, tx)
 			totalByteSize += txByteSize
@@ -344,4 +311,86 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 	}
 
 	return proposal.Header, err
+}
+
+// populateUnfinalizedAncestryLookup traverses the unfinalized ancestry backward
+// to populate the transaction lookup (used for deduplication) and the rate limiter
+// (used to limit transaction submission by payer).
+//
+// The traversal begins with the block specified by parentID (the block we are
+// building on top of) and ends with the oldest unfinalized block in the ancestry.
+func (b *Builder) populateUnfinalizedAncestryLookup(parentID flow.Identifier, finalHeight uint64, lookup *transactionLookup, limiter *rateLimiter) error {
+
+	ancestorID := parentID
+	for {
+		ancestor, err := b.clusterHeaders.ByBlockID(ancestorID)
+		if err != nil {
+			return fmt.Errorf("could not retrieve ancestor header: %w", err)
+		}
+
+		if ancestor.Height <= finalHeight {
+			break
+		}
+
+		payload, err := b.payloads.ByBlockID(ancestorID)
+		if err != nil {
+			return fmt.Errorf("could not retrieve ancestor payload: %w", err)
+		}
+
+		for _, tx := range payload.Collection.Transactions {
+			lookup.addUnfinalizedAncestor(tx.ID())
+			limiter.addAncestor(ancestor.Height, tx)
+		}
+		ancestorID = ancestor.ParentID
+	}
+
+	return nil
+}
+
+// populateFinalizedAncestryLookup traverses the reference block height index to
+// populate the transaction lookup (used for deduplication) and the rate limiter
+// (used to limit transaction submission by payer).
+//
+// The traversal is structured so that we check every collection whose reference
+// block height translates to a possible constituent transaction which could also
+// appear in the collection we are building.
+func (b *Builder) populateFinalizedAncestryLookup(minRefHeight, maxRefHeight uint64, lookup *transactionLookup, limiter *rateLimiter) error {
+
+	// Let E be the global transaction expiry constant, measured in blocks
+	//
+	// 1. a cluster block's reference block height is equal to the lowest reference
+	//    block height of all its constituent transactions
+	// 2. a transaction with reference block height T is eligible for inclusion in
+	//    any collection with reference block height C where C<=T<C+E
+	//
+	// Therefore, to guarantee that we find all possible duplicates, we need to check
+	// all collections with reference block height in the range (C-E,C+E). Since we
+	// know the actual maximum reference height included in this collection, we can
+	// further limit the search range to (C-E,maxRefHeight]
+
+	// the finalized cluster blocks which could possibly contain any conflicting transactions
+	var clusterBlockIDs []flow.Identifier
+	start := minRefHeight - flow.DefaultMaxCollectionByteSize + 1
+	end := maxRefHeight
+	err := b.db.View(operation.LookupClusterBlocksByReferenceHeightRange(start, end, &clusterBlockIDs))
+	if err != nil {
+		return fmt.Errorf("could not lookup finalized cluster blocks by reference height range [%d,%d]: %w", start, end, err)
+	}
+
+	for _, blockID := range clusterBlockIDs {
+		header, err := b.clusterHeaders.ByBlockID(blockID)
+		if err != nil {
+			return fmt.Errorf("could not retrieve cluster header (id=%x): %w", blockID, err)
+		}
+		payload, err := b.payloads.ByBlockID(blockID)
+		if err != nil {
+			return fmt.Errorf("could not retrieve cluster payload (block_id=%x): %w", blockID, err)
+		}
+		for _, tx := range payload.Collection.Transactions {
+			lookup.addFinalizedAncestor(tx.ID())
+			limiter.addAncestor(header.Height, tx)
+		}
+	}
+
+	return nil
 }
--- a/module/builder/collection/builder.go
+++ b/module/builder/collection/builder.go
@@ -90,8 +90,7 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 	//
 	// A collection with overlapping expiry window can be finalized or un-finalized.
 	// * to find all non-expired and finalized collections, we make use of an index
-	//   (main_chain_finalized_height -> cluster_block_id), to search for a range of main chain heights
-	//   which could be only referenced by collections with overlapping expiry windows.
+	//   (main_chain_finalized_height -> cluster_block_ids with respective reference height), to search for a range of main chain heights	//   which could be only referenced by collections with overlapping expiry windows.
 	// * to find all overlapping and un-finalized collections, we can't use the above index, because it's
 	//   only for finalized collections. Instead, we simply traverse along the chain up to the last
 	//   finalized block. This could possibly include some collections with expiry windows that DON'T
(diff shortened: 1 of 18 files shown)


module/builder/collection/builder.go
```diff
@@ -345,3 +306,88 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
...
+}
+
+// populateFinalizedAncestryLookup traverses the reference block height index to
+// populate the transaction lookup (used for deduplication) and the rate limiter
+// (used to limit transaction submission by payer).
+//
+// The traversal is structured so that we check every collection whose reference
+// block height translates to a possible constituent transaction which could also
+// appear in the collection we are building.
+func (b *Builder) populateFinalizedAncestryLookup(minRefHeight, maxRefHeight uint64, lookup *transactionLookup, limiter *rateLimiter) error {
```
AlexHentschel, line 351: We have unified functions for fork traversal here:
* https://github.com/onflow/flow-go/blob/6aba4c0f95d4d7e2a71d8c34ff1ec2829f285aec/state/fork/traversal.go#L16
* https://github.com/onflow/flow-go/blob/6aba4c0f95d4d7e2a71d8c34ff1ec2829f285aec/state/fork/traversal.go#L45

They are designed to be highly customizable. Usage examples:
* https://github.com/onflow/flow-go/blob/6aba4c0f95d4d7e2a71d8c34ff1ec2829f285aec/module/builder/consensus/builder.go#L305
* https://github.com/onflow/flow-go/blob/6aba4c0f95d4d7e2a71d8c34ff1ec2829f285aec/module/validation/seal_validator.go#L191 

Not sure, but hopefully we are able to use this here and avoid replicating this logic in many different places over and over.

jordanschalm: 👍  Good point - I refactored the un-finalized ancestry lookup to use the fork traversal functions. I don't think it's applicable to finalized ancestry though, since we're traversing using the reference height index instead of by block parent.

module/builder/collection/builder.go
```diff
@@ -72,35 +73,41 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
...
 		}
 		var refChainFinalizedID flow.Identifier
-		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(tx)
+		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve main finalized ID: %w", err)
 		}
+		// pre-compute the minimum possible reference block height for transactions
+		// included in this collection (actual reference height may be greater)
+		minPossibleRefHeight := refChainFinalizedHeight - uint64(flow.DefaultTransactionExpiry-b.config.ExpiryBuffer)
```
ramtinms, line 103: is it possible that we end up in a situation where ExpiryBuffer is not large enough? like some sort of collection processing built up on consensus nodes, or consensus nodes not finalizing?

ramtinms (approved):
Looks good to me,

module/builder/collection/builder.go
jordanschalm: If we just traverse the fork for finalized blocks, we don't know when to stop the traversal. It's also possible we might unnecessarily check collections that can't possibly contain a duplicate transaction. That's the reason for adding the index in this PR. The index tells us, given a collection's reference block height, exactly which other collections can possibly contain a conflicting transaction. 

We need to check for duplicates in any collection with an expiry range that overlaps with the candidate collection's expiry range. Since expiry is defined by heights on the main chain, and we're traversing the cluster chain, we can't just traverse backward a fixed number of collections. We also can't assume that, if we observe a collection with a reference height outside the range overlapping with the candidate collection, it means that no other older ancestor will have a reference height that does overlap with the candidate collection. Consecutive collections may not have strictly increasing reference block heights.

durkmurder (approved):
Looks good, very clean

module/builder/collection/builder.go
durkmurder: Thank you for explanation, yeah, that makes sense

module/builder/collection/builder.go
```diff
@@ -72,35 +73,41 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
 
 	// first we construct a proposal in-memory, ensuring it is a valid extension
 	// of chain state -- this can be done in a read-only transaction
-	err := b.db.View(func(tx *badger.Txn) error {
+	err := b.db.View(func(btx *badger.Txn) error {
```
zhangchiqing, line 76: This `View` transaction now contains lots of db queries and other operation. It will increase the chance of race condition and failure. 

I would suggest to break this view transaction into multiple ones. And move some operations out of this view transaction, such as removing txs from transactions mempool

module/builder/collection/builder.go
```diff
@@ -72,35 +73,41 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
...
 		var refChainFinalizedID flow.Identifier
-		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(tx)
+		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve main finalized ID: %w", err)
 		}
+		// pre-compute the minimum possible reference block height for transactions
+		// included in this collection (actual reference height may be greater)
+		minPossibleRefHeight := refChainFinalizedHeight - uint64(flow.DefaultTransactionExpiry-b.config.ExpiryBuffer)
+		if minPossibleRefHeight > refChainFinalizedHeight {
```
zhangchiqing, line 104: Do we need to check if this minPossibleRefHeight is higher than the root height?

module/builder/collection/builder.go
```diff
@@ -126,66 +133,21 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
...
-			}
-
-			collection := payload.Collection
-			for _, tx := range collection.Transactions {
-				lookup.addUnfinalizedAncestor(tx.ID())
-				limiter.addAncestor(ancestor.Height, tx)
-			}
-			ancestorID = ancestor.ParentID
+		// first, look up previously included transactions in UN-FINALIZED ancestors
+		err = b.populateUnfinalizedAncestryLookup(parentID, clusterFinal.Height, lookup, limiter)
```
zhangchiqing, line 137: Do we have a cap on the number of finalized cluster blocks to traverse back regardless finalized or unfinalized?

In the event of finalization slow or finalization halt, and cluster chain grows at the normal speed, then it will make the range of finalized cluster blocks very big. In this case, we might look up too many cluster blocks for checking dups. In this worst case, are checking so many cluster blocks indeed necessary to prevent dups?

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+	}
+	end := maxRefHeight
+	err := m.db.View(operation.LookupClusterBlocksByReferenceHeightRange(start, end, &clusterBlockIDs))
+	if err != nil {
+		return nil, fmt.Errorf("could not lookup finalized cluster blocks by reference height range [%d,%d]: %w", start, end, err)
+	}
+
+	for _, blockID := range clusterBlockIDs {
+		// TODO: could add LightByBlockID and retrieve only tx IDs
+		payload, err := m.payloads.ByBlockID(blockID)
```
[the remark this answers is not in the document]
jordanschalm: There is no guarantee about the rate at which collections are produced compared to the rate at which blocks are produced, so there isn't a bound on the worst case in terms of number of cluster blocks to check. The payload does have a maximum size though (100 transactions by default. But if consensus halts and we keep building collections forever, then the number of cluster blocks we need to query will keep growing. 

Practically:
* when healthy, both blocks and collections are continually built, so the average size of the query is about the same as it is currently (slightly larger because we build collections faster than blocks)
* if consensus halts for long enough to become a problem here, we would turn off transaction ingestion until the problem is resolved, so that newly created empty collections wouldn't get included in these queries.

module/builder/collection/builder.go
```diff
@@ -72,35 +73,41 @@ func (b *Builder) BuildOn(parentID flow.Identifier, setter func(*flow.Header) er
...
+		err = operation.LookupBlockHeight(refChainFinalizedHeight, &refChainFinalizedID)(btx)
 		if err != nil {
 			return fmt.Errorf("could not retrieve main finalized ID: %w", err)
 		}
+		// pre-compute the minimum possible reference block height for transactions
+		// included in this collection (actual reference height may be greater)
+		minPossibleRefHeight := refChainFinalizedHeight - uint64(flow.DefaultTransactionExpiry-b.config.ExpiryBuffer)
+		if minPossibleRefHeight > refChainFinalizedHeight {
+			minPossibleRefHeight = 0 // overflow check
+		}
```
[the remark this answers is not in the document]
jordanschalm: Added check that the config is valid in the constructor: https://github.com/onflow/flow-go/pull/2385/commits/c15e68b2ff74b323d21c8836ea1043d15b9603f4

module/builder/collection/builder.go
jordanschalm: 👍  Good point, there is a lot here that doesn't need to be in the transaction: updated in https://github.com/onflow/flow-go/pull/2385/commits/c15e68b2ff74b323d21c8836ea1043d15b9603f4

module/builder/collection/builder.go
```diff
@@ -60,277 +53,244 @@ func NewBuilder(
 		apply(&b.config)
 	}
 
-	return &b
+	// sanity check config
+	if b.config.ExpiryBuffer >= flow.DefaultTransactionExpiry {
```
zhangchiqing, line 57:
```suggestion
	if b.config.ExpiryBuffer > flow.DefaultTransactionExpiry {
```

I think it's OK to be equal, which is also consistent with the error message

module/builder/collection/builder.go
```diff
@@ -60,277 +53,244 @@ func NewBuilder(
...
+	limiter := newRateLimiter(b.config, parent.Height+1)
+
+	// RATE LIMITING: the builder module can be configured to limit the
+	// rate at which transactions with a common payer are included in
+	// blocks. Depending on the configured limit, we either allow 1
+	// transaction every N sequential collections, or we allow K transactions
+	// per collection.
+
+	// first, look up previously included transactions in UN-FINALIZED ancestors
+	fmt.Println("cluster chain final height: ", clusterChainFinalizedBlock.Height, clusterChainFinalizedBlock.ID())
```
zhangchiqing, line 137: do we need this log?

module/builder/collection/builder.go
```diff
@@ -60,277 +53,244 @@ func NewBuilder(
...
+	err = b.populateUnfinalizedAncestryLookup(parentID, clusterChainFinalizedBlock.Height, lookup, limiter)
```
zhangchiqing, line 138: why do we need to distinguish `populateUnfinalizedAncestryLookup` and `populateFinalizedAncestryLookup `?

Are they guaranteed to be connected? More importantly, would there be a gap between the last block of `populateFinalizedAncestryLookup` and the start block of `populateUnfinalizedAncestryLookup`? In this case, some transations will not be missed for dedup. 

I wonder why not computing the the `minRefFinalizedHeight` based on `minPossibleRefHeight` and the new index. So that we can just iterate from parentID all the way to the `minRefFinalizedHeight`, which is guaranteed to cover the range of blocks without any gap.

module/builder/collection/builder.go
jordanschalm: If they're equal, we would only allow transactions which have exactly our latest finalized block as the reference block. Practically it would mean very close to 0 transactions would be included. Maybe it's technically valid, but I think using `>=` is better in case there is a typo like `ExpiryBuffer=flow.DefaultTransactionExpiry`

module/builder/collection/builder.go
```diff
@@ -60,277 +53,244 @@ func NewBuilder(
...
-	var proposal cluster.Block
+	var proposal cluster.Block                 // proposal we are building
+	var parent flow.Header                     // parent of the proposal we are building
+	var clusterChainFinalizedBlock flow.Header // finalized block on the cluster chain
+	var refChainFinalizedHeight uint64         // finalized height on reference chain
+	var refChainFinalizedID flow.Identifier    // finalized block ID on reference chain
 
 	startTime := time.Now()
 
+	// STEP ONE: Load some things we need to do our work.
```
zhangchiqing, line 75:
```suggestion
	// STEP ONE: build a lookup for excluding duplicated transactions. This is briefly how it works:
	// expired transaction won't be included in new collection block, so it's not possible for expired tx
	// to be duplicated.
	// therefore, we only need a lookup for all non-expired transactions.
	// non-expired transactions only exists in non-expired collections, because a valid collection
	// must only include non-expired transactions.
	// in other words, we can build the transaction lookup by finding all the non-expired collections.
	// Note, in order to make the search algorithm simpler, finding all non-expired collections might 
	// end up including some expired collections. But that's OK, because expired transactions are
	// excluded in new collection block anyway.
	// a non-expired collection can be finalized or un-finalized. 
	// - to find all non-expired and finalized collections, we make use of an index 
	//    (main_chain_finalized_height -> cluster_block_id), to search for a range of main chain heights 
	//     which could be only referenced by non-expired collections. 
	// - to find all non-expired and un-finalized collections, we can't use the above index, because it's
	//    only for finalized collections. Instead, we simply traverse along the chain up to the last 
	//    finalized block. This could possibly include some expired collections, but it's OK as explained
	//    above.
	// after combining both the finalized and un-finalized collection blocks that are not expired, 
	// we can iterate through their transactions, and build a lookup for excluding duplicated transactions.
```

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+			txID := tx.ID()
+			_, duplicated := includedTransactions[txID]
+			if duplicated {
+				duplicateTxIDs = append(duplicateTxIDs, txID)
+			}
+		}
+		return nil
+	}, fork.ExcludingHeight(finalHeight))
+
+	return duplicateTxIDs, err
```
zhangchiqing, line 230: `checkDupeTransactionsInUnfinalizedAncestry` has duplicated code with `populateUnfinalizedAncestryLookup`. 
This is how you can extract the common part:

```
func iterateThroughTransactionsInUnfinalizedAcenstery(clusterHeaders, payloads, parentID, finalHeight, reducer func(blockHeight, txID)) error {
  return fork.TraverseBackward(clusterHeaders, parentID, func(ancestor *flow.Header) error {
		payload, err :=payloads.ByBlockID(ancestor.ID())
		if err != nil {
			return fmt.Errorf("could not retrieve ancestor payload: %w", err)
		}

		for _, tx := range payload.Collection.Transactions {
			reducer(ancester.Height, tx.ID())
		}
		return nil
	}, fork.ExcludingHeight(finalHeight))
}

// for builder
iterateThroughTransactionsInUnfinalizedAcenstery(clusterHeaders, payloads, finalHeight, func(blockHeight, txID) {
  lookup.addUnfinalizedAncestor(txID)
  limiter.addAncestor(blockHeight)
})

// for mutator
iterateThroughTransactionsInUnfinalizedAcenstery(clusterHeaders, payloads, finalHeight, func(blockHeight, txID) {
  _, duplicated := includedTransaction[txID]
  if duplicated {
    duplicatedTx = append(duplicatedTx, txID)
  }
})
```

same for `checkDupeTransactionsInFinalizedAncestry`

storage/badger/operation/common.go
```diff
@@ -226,6 +226,12 @@ func lookup(entityIDs *[]flow.Identifier) func() (checkFunc, createFunc, handleF
 	}
 }
 
+func withPrefetchValues(prefetch bool) func(*badger.IteratorOptions) {
+	return func(options *badger.IteratorOptions) {
+		options.PrefetchValues = false
```
zhangchiqing, line 231:
```suggestion
		options.PrefetchValues = prefetch
```

zhangchiqing (approved):
Looks good. just had some minor comments

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+		start = 0 // overflow check
+	}
+	end := maxRefHeight
+	err := m.db.View(operation.LookupClusterBlocksByReferenceHeightRange(start, end, &clusterBlockIDs))
+	if err != nil {
+		return nil, fmt.Errorf("could not lookup finalized cluster blocks by reference height range [%d,%d]: %w", start, end, err)
+	}
+
+	for _, blockID := range clusterBlockIDs {
+		// TODO: could add LightByBlockID and retrieve only tx IDs
```
zhangchiqing, line 263: This sounds like optional, but would be very nice to have, because it avoids computing hashes on the transactions.

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+	// further limit the search range to (C-E,maxRefHeight]
+
+	// the finalized cluster blocks which could possibly contain any conflicting transactions
+	var clusterBlockIDs []flow.Identifier
+	start := minRefHeight - flow.DefaultTransactionExpiry + 1
+	if start > minRefHeight {
+		start = 0 // overflow check
+	}
+	end := maxRefHeight
+	err := m.db.View(operation.LookupClusterBlocksByReferenceHeightRange(start, end, &clusterBlockIDs))
```
zhangchiqing, line 257: this will be on a hot path, when every extending a new cluster block, we will look back X finalized cluster blocks, which means a certain finalized blocks might get queried over and over for many times.

Maybe we could optimize by maintaining some cache here.

module/builder/collection/builder.go
zhangchiqing: If we do want to log, it's better to log with zerolog. 

It actually makes sense to me that the builder to have a logger module, and be able to log things.

The builder right now does a lot of work and takes a lot of time to build a block. 

It would be good to log some useful metrics to understand which part is the bottleneck. 

For instance, I would be curious to know who many cluster blocks in total did builder actually iterate through to build the lookup. And see if that makes a big impact to block building time.

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+
+// checkDupeTransactionsInFinalizedAncestry checks for duplicate transactions in the finalized
+// ancestry, and returns a list of all duplicates if there are any.
+func (m *MutableState) checkDupeTransactionsInFinalizedAncestry(includedTransactions map[flow.Identifier]struct{}, minRefHeight, maxRefHeight uint64) ([]flow.Identifier, error) {
+	var duplicatedTxIDs []flow.Identifier
+
+	// Let E be the global transaction expiry constant, measured in blocks
+	//
+	// 1. a cluster block's reference block height is equal to the lowest reference
+	//    block height of all its constituent transactions
```
AlexHentschel, line 241: trying to increase mathematical rigour here a bit:
```suggestion
	// Let E be the global transaction expiry constant, measured in blocks. For each
	// T ∈ `includedTransactions`, we have to decide whether the transaction
	// already appeared in _any_ finalized cluster block.
	// Notation:
	//   - consider a valid cluster block C and let c be its reference block height
	//   - consider a transaction T ∈ `includedTransactions` and let t denote its
	//     reference block height
	//
	// Boundary conditions:
	// 1. C's reference block height is equal to the lowest reference block height of
	//    all its constituent transactions. Hence, C can contain T if and only if c <= t.
```

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+	// 2. a transaction with reference block height T is eligible for inclusion in
+	//    any collection with reference block height C where C<=T<C+E
```
AlexHentschel, line 243: ## ⚠️ 

Here you are assuming collection `C` can only contain `T` if T's reference block height is not exceeding `C+E`. This is an _upper bound_. But as far as I can see:
* we don't enforce the _upper bound_ in the mutator 
* I also couldn't convince myself that the builder adheres to this convention. There might be many unfinalized blocks from main consensus, which TXs could use as their reference blocks.

AlexHentschel (commented):
Thanks. Great documentation, and tests. Very clean code 👏 
However, I didn't manage to convince myself that the dedup algorithm is actually edge-case free. I feel we don't enforce the necessary assumptions in the builder or the mutator (?) See [this](https://github.com/onflow/flow-go/pull/2385#discussion_r872006862) comment

state/cluster/badger/mutator.go
AlexHentschel, line 243: ⚠️ 
I think the conditions in the code might be potentially inconsistent (?)
* Elsewhere in the code, we include the boundary value https://github.com/onflow/flow-go/blob/c15e68b2ff74b323d21c8836ea1043d15b9603f4/model/flow/constants.go#L21
* while in here, we _exclude_ the boundary value

Is that consistent?

module/builder/collection/builder.go
jordanschalm: Thanks for this - I adjusted the wording a bit to be more specific: we are searching for collections which have an overlapping expiry window with our collection under construction. Not necessarily collections which are expired vs not expired.

state/cluster/badger/mutator.go
jordanschalm: I benchmarked the lookup operation with 6,000 cluster blocks (about 5x what we would expect given the block/collection production rates on Mainnet) and it takes about 1ms. The data is stored in the key, which is generally stored in memory already by Badger. I don't think a separate cache will help much here.

state/cluster/badger/mutator.go
```diff
@@ -218,3 +205,74 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+		if err != nil {
+			return fmt.Errorf("could not retrieve ancestor payload: %w", err)
+		}
+
+		for _, tx := range payload.Collection.Transactions {
+			txID := tx.ID()
+			_, duplicated := includedTransactions[txID]
+			if duplicated {
+				duplicateTxIDs = append(duplicateTxIDs, txID)
+			}
```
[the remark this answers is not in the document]
jordanschalm: My thinking was to help with debugging why the duplicates were included.

state/cluster/badger/mutator.go
jordanschalm: The different ranges are talking about inclusion in collections and blocks.

For collections, we set the collection reference height to min(tx.ReferenceBlock.Height) | tx ∈ collection. (The reference heights of transactions in the collection is inclusive at the lower bound.)

For blocks, it's impossible for a block to contain a transaction which references the block it is within, hence why the reference heights of transactions in blocks is exclusive at the lower bound. The number of reference heights which are valid for inclusion in the block is equal to the expiry window (600).

I think it is more that the documentation is confusing than that the code is inconsistent. The "reference height inclusion window" for blocks and collection don't need to be exactly the same (and practically aren't, because of the expiry buffer for collections).

state/cluster/badger/mutator.go
```diff
@@ -117,6 +118,7 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		// check that all transactions within the collection are valid
 		minRefID := flow.ZeroID
 		minRefHeight := uint64(math.MaxUint64)
+		maxRefHeight := uint64(0)
 		for _, flowTx := range payload.Collection.Transactions {
 			refBlock, err := m.headers.ByBlockID(flowTx.ReferenceBlockID)
 			if errors.Is(err, storage.ErrNotFound) {
```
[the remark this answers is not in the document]
jordanschalm: 👍 Good point - we should have an `UnverifiableExtension` sentinel like we have for messages.

state/cluster/badger/mutator.go
```diff
@@ -117,6 +118,7 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		// check that all transactions within the collection are valid
 		minRefID := flow.ZeroID
 		minRefHeight := uint64(math.MaxUint64)
+		maxRefHeight := uint64(0)
 		for _, flowTx := range payload.Collection.Transactions {
 			refBlock, err := m.headers.ByBlockID(flowTx.ReferenceBlockID)
```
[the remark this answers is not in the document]
jordanschalm: > I feel a collector should only accept a collection where it knows that all the reference blocks are finalized.

👍 Agreed. This makes reasoning about the de-duplication correctness much easier as well. 

> Don't we need to check the same for each transaction's reference block?

We definitely do.

state/cluster/badger/mutator.go
jordanschalm: Agreed, I think enforcing that all reference blocks are finalized addresses this.

state/cluster/badger/mutator.go
jordanschalm: I didn't think it was worth doing this here, because:
* the common logic is only a few lines of code, and the only two usages are in different packages
* if we implement the [ID-only retrieval lookup for the builder](https://github.com/onflow/flow-go/pull/2385#discussion_r871884647), which I think we should, there is little re-factorable common logic left

storage/badger/operation/common.go
jordanschalm: Good catch, thanks 🙇

jordanschalm:
Hey @AlexHentschel & @zhangchiqing,

Thank you for the feedback. I addressed the documentation-related notes, simple fixes, and issues Alex pointed out about edge cases in the Builder (not in the `Mutator`). My thinking is we can address only the building logic here and address validation in a separate BFT-scoped issue. 

I've added 2 new issues capturing the feedback that wasn't actioned here:
* https://github.com/dapperlabs/flow-go/issues/6244 - performance/observability
* https://github.com/dapperlabs/flow-go/issues/6245 - BFT requirements

module/builder/collection/builder.go
```diff
@@ -60,277 +53,277 @@ func NewBuilder(
...
-			Height:      parent.Height + 1,
-			PayloadHash: payload.Hash(),
-			Timestamp:   time.Now().UTC(),
-
-			// NOTE: we rely on the HotStuff-provided setter to set the other
-			// fields, which are related to signatures and HotStuff internals
+		// retrieve the main chain header that was used as reference
+		refHeader, err := b.mainHeaders.ByBlockID(tx.ReferenceBlockID)
+		if errors.Is(err, storage.ErrNotFound) {
+			continue // in case we are configured with liberal transaction ingest rules
```
zhangchiqing, line 231: Question here:

Once the builder has built a lookup, it will start iterate through `transaction.All()` in the mempool, but is it a problem if there are too many tx to iterate through? the tx mempool has a size cap, so the worst case is the tx mempool being full. 

The reason I think of this is that when a collection node is falling behind mainchain, it will receive lots of txs with unknown `tx.ReferenceBlockID`, and it seems we still keep it in the mempool, this would cause the falling behind collection node to be slow during catch up, when it's entering a view where it needs to propose a block. 

I wonder if make sense to drop a tx if reference block ID is unknown. Do we already check reference block ID before adding it to the mempool, if so, this ErrNotFound case should not happen, otherwise it's a bug.

AlexHentschel: We could two a 2-step approach:
* ideally, all transactions in the mempool should have a _finalized_ block as their reference block (see [here](https://github.com/onflow/flow-go/pull/2385#discussion_r871985113) for reasoning)
* we could cache all TXs whose reference block is _not_ finalized
* when the collector locally confirms that another block on the main chain has been finalized (`OnBlockFinalized` event), it can release the corresponding TXs into the mempool. 
* pruning heurisitic for the cache:

state/cluster/badger/mutator.go
```diff
@@ -218,3 +216,81 @@ func (m *MutableState) Extend(block *cluster.Block) error {
...
+	//     reference block height
+	//
+	// Boundary conditions:
+	// 1. C's reference block height is equal to the lowest reference block height of
+	//    all its constituent transactions. Hence, C can contain T if and only if c <= t.
+	// 2. a transaction with reference block height T is eligible for inclusion in
+	//    any collection with reference block height C where C<=T<C+E
+	//
+	// Therefore, to guarantee that we find all possible duplicates, we need to check
+	// all collections with reference block height in the range (C-E,C+E). Since we
```
AlexHentschel, line 264: * I would suggest to update point 2 to conform to the notation.
* The argument for the overall range is slightly confusing to me.

```suggestion
	// 2. For T to be eligible for inclusion in collection C, _none_ of the transactions within C are allowed 
	// to be expired w.r.t. C's reference block. Hence, for collection C to potentially contain T, it must satisfy t < c + E.  
	//
	// Therefore, for collection C to potentially contain transaction T, it must satisfy t - E < c <= t. In other words, 
	// we only need to inspect collections with reference block height c ∈ (t-E, t]. Consequently, for a set of transactions, 
	// with `minRefHeight` (`maxRefHeight`) being the smallest (largest) reference block height, we only need to inspect collections with c ∈ (minRefHeight-E, maxRefHeight].
```

module/builder/collection/builder.go
```diff
@@ -60,277 +53,277 @@ func NewBuilder(
...
+	//   * therefore, C will contain only transactions with reference heights in [R,R+E)
+	// * if we are validating C:
+	//   * honest validators only consider C valid if all its transactions have reference heights in [R,R+E)
+	//
+	// Therefore, to check for duplicates, we only need a lookup for transactions in collection
+	// with expiry windows that overlap with our collection under construction.
+	//
+	// A collection with overlapping expiry window can be finalized or un-finalized.
+	// * to find all non-expired and finalized collections, we make use of an index
+	//   (main_chain_finalized_height -> cluster_block_id), to search for a range of main chain heights
```
AlexHentschel, line 93: the index returns a _set_ of blocks for a given `main_chain_finalized_height`

```suggestion
	//   (main_chain_finalized_height -> cluster_block_ids with respective reference height), to search for a range of main chain heights
```

AlexHentschel (approved):
Very nice. Largely suggestions for the goDoc

module/builder/collection/builder.go
jordanschalm: The configuration for whether to ingest transactions with unknown reference blocks is in the [`TransactionValidator`](https://github.com/onflow/flow-go/blob/2df4ced13a14d61a6b784490af9b3cf02cc96d07/access/validator.go#L199), instantiated in the [`ingest engine`](https://github.com/onflow/flow-go/blob/57f89d4e96259f08fe84163c91ecd32484401b45/engine/collection/ingest/engine.go#L61). We do not accept transactions with unknown reference blocks in the collection node (`AllowUnknownReferenceBlockID=false`). The builder is tolerant of unknown reference blocks so that it respects the transaction ingest rules set in the `ingest` engine, regardless of what they are.

zhangchiqing: Looks like even though we have this option `AllowUnknownReferenceBlockID`, [the collection node is always using](https://github.com/onflow/flow-go/blob/master/engine/collection/ingest/engine.go#L64) `AllUnknownReferenceBlockID=false`, so it's not an issue.

storage/badger/headers.go
```diff
@@ -138,6 +138,17 @@ func (h *Headers) ByHeight(height uint64) (*flow.Header, error) {
...
+func (h *Headers) BlockIDByHeight(height uint64) (flow.Identifier, error) {
+	tx := h.db.NewTransaction(false)
+	defer tx.Discard()
+
+	blockID, err := h.retrieveIdByHeightTx(height)(tx)
+	if err != nil {
+		return flow.ZeroID, fmt.Errorf("could not lookup block id by height %d: %w", height, err)
+	}
+	return blockID, nil
+}
```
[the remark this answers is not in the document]
jordanschalm: I added this as a more efficient way to check whether a reference block is finalized, but modifying the `Headers` interface broke `flow-emulator`. I included making use of this method as an item in the follow-up issue: https://github.com/dapperlabs/flow-go/issues/6244

jordanschalm:
bors merge

PR #2386, opened 2022-05-09
[v0.25] Robust Transaction Deduplication

jordanschalm:
This is the same as https://github.com/onflow/flow-go/pull/2385, with an added feature needed only for mid-spork deployment on `testnet34` and `mainnet17`:
* on startup, build the reference block height index if none already exists

⚠️  **For reviewers, the bulk of the difference in this PR compared to #2385 is in [this file](https://github.com/onflow/flow-go/pull/2386/files#diff-4892eee52c975e9f3919a5d884908bd090aa4683210b469d64b9b417e2d2d725) (and the corresponding test).**

### DIFF
--- a/cmd/collection/main.go
+++ b/cmd/collection/main.go
@@ -6,16 +6,13 @@ import (
 
 	"github.com/spf13/pflag"
 
-	"github.com/onflow/flow-go/cmd/util/cmd/common"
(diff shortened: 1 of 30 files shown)


zhangchiqing (approved):

ramtinms (approved):
looks good to me

PR #2390, opened 2022-05-09
Separate permission setting of contract deploy/update and removal

robert-e-davidson3:
Closes https://github.com/dapperlabs/cadence-private-issues/issues/40

Contracts can be deployed, updated, or removed by the service account and accounts it authorizes to do so. That authorization applies to all three operations but we want to permit more fine-grained control: deploy/update and removal should be two separate permissions.

This is somewhat related to https://github.com/onflow/flow-go/pull/2346.

### DIFF
--- a/fvm/blueprints/contracts.go
+++ b/fvm/blueprints/contracts.go
@@ -20,7 +20,7 @@ const ContractRemovalAuthorizedAddressesPathIdentifier = "authorizedAddressesToR
 const IsContractDeploymentRestrictedPathDomain = "storage"
 const IsContractDeploymentRestrictedPathIdentifier = "isContractDeploymentRestricted"
 
-const setContractDeploymentAuthorizersTransactionTemplate = `
+const setContractOperationAuthorizersTransactionTemplate = `
 transaction(addresses: [Address], path: StoragePath) {
 	prepare(signer: AuthAccount) {
 		signer.load<[Address]>(from: path)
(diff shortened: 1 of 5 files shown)


fvm/blueprints/contracts.go
```diff
@@ -48,6 +51,28 @@ func SetContractDeploymentAuthorizersTransaction(serviceAccount flow.Address, au
...
+	arg2, err := jsoncdc.Encode(cadence.Path{
+		Domain:     ContractRemovalAuthorizedAddressesPathDomain,
+		Identifier: ContractRemovalAuthorizedAddressesPathIdentifier,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return flow.NewTransactionBody().
+		SetScript([]byte(setContractDeploymentAuthorizersTransactionTemplate)).
```
turbolent, line 70: Maybe rename `setContractDeploymentAuthorizersTransactionTemplate` to something that indicates it is more generally used, e.g. also for setting the removal authorizers here. Maybe `setContractOperationAuthorizersTransactionTemplate`?

fvm/blueprints/contracts.go
```diff
@@ -48,6 +51,28 @@ func SetContractDeploymentAuthorizersTransaction(serviceAccount flow.Address, au
 		AddArgument(arg2), nil
 }
 
+// SetContractRemovalAuthorizersTransaction returns a transaction for updating list of authorized accounts allowed to remove contracts
+func SetContractRemovalAuthorizersTransaction(serviceAccount flow.Address, authorized []flow.Address) (*flow.TransactionBody, error) {
+	arg1, err := jsoncdc.Encode(utils.AddressSliceToCadenceValue(utils.FlowAddressSliceToCadenceAddressSlice(authorized)))
```
turbolent, line 56: Maybe improve naming of `arg1` and `arg2` here, and also in `SetContractDeploymentAuthorizersTransaction`

turbolent (commented):
Nice work!

fvm/blueprints/contracts.go
robert-e-davidson3: Good point.

Looking at this, I realize I should factor out logic and just pass in the right path. I'll push that shortly.

turbolent (approved):
Nice!

m4ksio (approved):

PR #2391, opened 2022-05-09
[Crypto] fix race conditions in DKG tests

tarakby:
Addresses https://github.com/dapperlabs/flow-internal/issues/1700
It fixes race conditions in DKG tests. All tests using `--race` flag are now passing.

This is a follow up on the work that started by https://github.com/onflow/flow-go/pull/2293


tarakby:
> Were these tests reported as flaky by the Flaky Test Monitor?

No, they were not flakey. But running them with the race detector (`--race`) was showing race conditions. I guess the Flakey Test Monitor doesn't look specifically at race conditions for now right?

crypto/dkg_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build relic
 // +build relic
```
jordanschalm, line 2: Do we need both of these?

jordanschalm (approved):

kc1116 (approved):

crypto/dkg_test.go
tarakby: IDE has been adding them automatically lately, though the `// +` seems enough.

tarakby:
bors merge

PR #2394, opened 2022-05-10
Dan/1536 add cadence execution time metrics

danielholmes839:
WIP


module/metrics/transaction.go
```diff
@@ -81,11 +83,28 @@ func NewTransactionCollector(transactionTimings mempool.TransactionTimings, log
 			Subsystem: subsystemTransactionSubmission,
 			Help:      "counter for the success/failure of transaction submissions",
 		}, []string{"result"}),
+		executeScriptRTT: promauto.NewCounter(prometheus.CounterOpts{
+			Name:      "execute_script_rtt",
+			Namespace: namespaceAccess,
+			Subsystem: "TEMP",
```
danielholmes839, line 89: I'm not sure if I should create a dedicated Subsystem for these metrics.

kc1116: You should be able to use the existing transaction timing subsystem, I don't think these metrics warrant a new subsystem.

engine/access/rpc/backend/backend_scripts.go
```diff
@@ -99,15 +100,20 @@ func (b *backendScripts) executeScriptOnExecutionNode(
 	// encode to MD5 as low compute/memory lookup key
 	encodedScript := md5.Sum(script) //nolint:gosec
 
+	// record start time
+	execStartTime := time.Now()
```
peterargue, line 104: I think the start time should be taken within the loop so it only includes the time spent executing on the EN that returned a successful result.

danielholmes839: Fixed

kc1116: @peterargue Do we not want to take into account the time it takes us to find a successful execution node ?

module/metrics/transaction.go
```diff
@@ -81,11 +84,65 @@ func NewTransactionCollector(transactionTimings mempool.TransactionTimings, log
 			Subsystem: subsystemTransactionSubmission,
 			Help:      "counter for the success/failure of transaction submissions",
 		}, []string{"result"}),
+		executeScriptDuration: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name:      "execute_script_rtt_duration",
+			Namespace: namespaceAccess,
+			Subsystem: subsystemTransactionSubmission,
+			Help:      "counter for the duration in ms of the round trip time for executing a script",
+		}, []string{"script_size"}),
```
kc1116, line 92: Should this label be different ? Why are we using script_size for rrt_duration

module/metrics/transaction.go
```diff
@@ -81,11 +84,65 @@ func NewTransactionCollector(transactionTimings mempool.TransactionTimings, log
...
+}
+
+func (tc *TransactionCollector) GetTransactionResultRTT(dur time.Duration, size int) {
+	tc.scriptSizeHist(size, "GetTransactionResult")
+	tc.transactionResultDuration.With(prometheus.Labels{
+		"script_size": tc.scriptSizeLabel(size),
+	}).Add(float64(dur.Milliseconds()))
+}
+
+func (tc *TransactionCollector) scriptSizeLabel(size int) string {
```
kc1116, line 124: Is there a way to get more accurate size labels ? Maybe only use the kb conversion for very large scripts.

danielholmes839: Currently using Vishal's recommendation https://github.com/dapperlabs/flow-internal/issues/1536#issuecomment-1122955228

engine/access/rpc/backend/backend_scripts.go
peterargue: no, this metric is looking to measure how long it takes to execute the script from the ANs perspective. We have other metrics that capture processing time of the gRPC endpoint that we could use for total time.

module/metrics/transaction.go
```diff
@@ -21,6 +21,9 @@ type TransactionCollector struct {
 	timeToExecuted             prometheus.Summary
 	timeToFinalizedExecuted    prometheus.Summary
 	transactionSubmission      *prometheus.CounterVec
+	executeScriptDuration      *prometheus.CounterVec
+	transactionResultDuration  *prometheus.CounterVec
```
peterargue, line 25: These should probably be `prometheus.Histogram` to track the value over time. Count is usually used for values where we want to track how often/how many of something occur.

e.g. https://github.com/onflow/flow-go/blob/master/module/metrics/execution.go#L302-L308

module/metrics/transaction.go
```diff
@@ -21,6 +21,9 @@ type TransactionCollector struct {
...
+	scriptSize                 *prometheus.HistogramVec
```
peterargue, line 26: This should also be a `prometheus.Histogram`

e.g. https://github.com/onflow/flow-go/blob/master/module/metrics/state_synchronization.go#L70-L76

module/metrics.go
```diff
@@ -400,7 +400,17 @@ type ExecutionMetrics interface {
 	ExecutionBlockDataUploadFinished(dur time.Duration)
 }
 
+type BackendScriptsMetrics interface {
+	// Record the round trip time while executing a script
+	ExecuteScriptRTT(dur time.Duration, size int)
```
peterargue, line 405: Maybe use `ScriptExecuted` to align with the naming convention used by other metrics

module/metrics.go
```diff
@@ -400,7 +400,17 @@ type ExecutionMetrics interface {
...
+}
+
 type TransactionMetrics interface {
+	BackendScriptsMetrics
+
+	// Record the round trip time while getting a transaction result
+	GetTransactionResultRTT(dur time.Duration, size int)
```
peterargue, line 412: maybe `TransactionResultFetched`

peterargue (commented):
great work. This is coming together.

module/metrics/transaction.go
```diff
@@ -81,11 +84,67 @@ func NewTransactionCollector(transactionTimings mempool.TransactionTimings, log
...
+		}, []string{"script_size"}),
+		transactionResultDuration: promauto.NewHistogramVec(prometheus.HistogramOpts{
+			Name:      "transaction_result_fetched_duration",
+			Namespace: namespaceAccess,
+			Subsystem: subsystemTransactionSubmission,
+			Help:      "histogram for the duration in ms of the round trip time for getting a transaction result",
+			Buckets:   []float64{1, 100, 500, 1000, 2000, 5000},
+		}, []string{"script_size"}),
+		scriptSize: promauto.NewHistogramVec(prometheus.HistogramOpts{
+			Name:      "script_size",
```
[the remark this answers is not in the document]
danielholmes839: I'm not sure. I watched one video that recommended using labels for this sort of thing. I think they said something like otherwise you end up with too many metrics. You could end up with 10 metrics instead of 1 metric with 10 labels. We would only have two right now though.

With multiple metrics do you know if it's easy to combine them into one histogram? I feel like being able to see the distribution of script size across all methods at the same time would be useful.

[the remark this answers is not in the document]
vishalchangrani: I agree with what you said but had the following concerns:
1. A transaction is not really a script. So may be we can the label "payload_size" ?
2. Also, if reducing the number of metrics is better then can't the first two metrics - transaction result duration and script execution duration be combine? (I don't mind it either ways).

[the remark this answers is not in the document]
danielholmes839: Ok I think using two metrics makes sense. I've pushed those changes.

vishalchangrani (approved):
lgtm

koko1123 (approved):
Looks good! Thanks for persisting through it!

PR #2396, opened 2022-05-10
Log MemoryUsed by transactions. 

dsainati1:
Add `MemoryUsed` to the transaction process, and then log it afterwards so we can pick it up in Grafana.


fvm/transaction.go
```diff
@@ -77,6 +78,8 @@ func (proc *TransactionProcedure) Run(vm *VirtualMachine, ctx Context, st *state
 		}
 	}
 
+	proc.MemoryUsed = uint64(st.State().TotalMemoryUsed())
```
janezpodhostnik, line 81: can you move this to `transactionInvoker` and duplicate the logic for computation used:
- https://github.com/onflow/flow-go/blob/8f7942581efe7d86f7f7121a7cb6c1daf9fd3499/fvm/transactionInvoker.go#L248
- https://github.com/onflow/flow-go/blob/8f7942581efe7d86f7f7121a7cb6c1daf9fd3499/fvm/transactionInvoker.go#L167

That way we have that logic together.

janezpodhostnik (changes requested):

janezpodhostnik (approved):
LGTM

engine/execution/computation/computer/computer.go
```diff
@@ -407,19 +405,18 @@ func (e *blockComputer) executeTransaction(
...
+	runtime.ReadMemStats(&m)
+	memAllocAfter := m.TotalAlloc
+
 	evt := e.log.With().
 		Hex("tx_id", txResult.TransactionID[:]).
 		Str("block_id", res.ExecutableBlock.ID().String()).
 		Str("traceID", traceID).
 		Uint64("computation_used", txResult.ComputationUsed).
+		Uint64("memory_used", tx.MemoryUsed).
+		Uint64("memAlloc", memAllocAfter-memAllocBefore).
```
SupunS, line 417: not something introduced in this PR, and probably not a big deal: wouldn't it be nice to use snake-case for all names for consistency? e.g: `memory_alloc`

SupunS (approved):
Is it possible to add a test-case?

fvm/fvm_test.go
```diff
@@ -3539,6 +3539,7 @@ func TestSettingExecutionWeights(t *testing.T) {
 			tx := fvm.Transaction(txBody, 0)
 			err = vm.Run(ctx, tx, view, programs)
 			require.NoError(t, err)
+			require.Greater(t, tx.MemoryUsed, uint64(20_000_000_000))
```
dsainati1, line 3542: @SupunS should check that the `MemoryUsed` field is properly set/updated.

fvm/transactionInvoker.go
```diff
@@ -246,6 +247,7 @@ func (i *TransactionInvoker) Process(
 	// if tx failed this will only contain fee deduction logs
 	proc.Logs = append(proc.Logs, env.Logs()...)
 	proc.ComputationUsed = proc.ComputationUsed + computationUsed
+	proc.MemoryUsed = proc.MemoryUsed + memoryUsed
```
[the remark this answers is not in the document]
dsainati1: If our memory limit is lower than the max value for an integer, we should never have to worry about this overflowing right? The transaction should abort long before we near the point of overflow.

SaveTheRbtz:
Please note that `runtime.ReadMemStats` is a very expensive function to call since it calls `stopTheWorld` and waits for all goroutines to park:
```
// stopTheWorld stops all P's from executing goroutines, interrupting
// all goroutines at GC safe points and records reason as the reason
// for the stop. On return, only the current goroutine's P is running.
...
```

PR #2402, opened 2022-05-11
Override faulty event payload

m4ksio:


janezpodhostnik (approved):
LGTM!

I double checked:
- tx id
- event type id
- event values

ramtinms (approved):

PR #2403, opened 2022-05-11
Chunk reloading blocks

m4ksio:


engine/execution/ingestion/engine.go
```diff
@@ -346,30 +352,39 @@ func (e *Engine) reloadUnexecutedBlocks() error {
 			}
 		}
 
-		finalized, pending, err := e.unexecutedBlocks()
-		if err != nil {
-			return fmt.Errorf("could not reload unexecuted blocks: %w", err)
-		}
+		for {
```
zhangchiqing, line 355: what is this for loop for?

m4ksio: To reload all the blocks - we terminate when there are no unexecuted blocks left

engine/execution/ingestion/engine.go
```diff
@@ -243,12 +243,15 @@ func (e *Engine) finalizedUnexecutedBlocks(finalized protocol.Snapshot) ([]flow.
...
+			return unexecuted, nil
+		}
 	}
 
 	return unexecuted, nil
 }
 
-func (e *Engine) pendingUnexecutedBlocks(finalized protocol.Snapshot) ([]flow.Identifier, error) {
+func (e *Engine) pendingUnexecutedBlocks(finalized protocol.Snapshot, maxBlocks int) ([]flow.Identifier, error) {
 	pendings, err := finalized.ValidDescendants()
```
zhangchiqing, line 255: The most common case is there are too many unexecuted and finalized blocks, so maxBlocks here would be `0`. And we could `ValidDescendants`. 
```suggestion
	if maxBlocks == 0 {
		return []flow.Identifier{}, nil  
	}
	pendings, err := finalized.ValidDescendants()
```

engine/execution/ingestion/engine.go
zhangchiqing: But reloading block doesn't execute them, if we reload them in batches, it still won't execute any block if there are too many batches.

I think we should just load up to `maxBlocks` number of blocks, and then start executing.

m4ksio: it does. Execution starts because they are put in the queue https://github.com/onflow/flow-go/pull/2403/files#diff-e5c0dea70cf688717060471583da13cc5269a6f661e955613ed7ef698354cc1aR378

engine/execution/ingestion/engine.go
```diff
@@ -346,30 +355,39 @@ func (e *Engine) reloadUnexecutedBlocks() error {
...
+					return fmt.Errorf("could not reload block: %v, %w", blockID, err)
+				}
+
+				e.log.Debug().Hex("block_id", blockID[:]).Msg("reloaded block")
+			}
+
+			if len(finalized) == 0 && len(pending) == 0 {
+				break
 			}
+			log.Info().Msgf("reloaded unexecuted blocks chunk of %d", len(finalized)+len(pending))
```
zhangchiqing, line 389:
```suggestion
			log.Info().Msgf("reloaded unexecuted blocks chunk of %d finalized, %d pending", len(finalized), len(pending))
```

zhangchiqing (approved):

ramtinms:
@m4ksio do we still need this considering the base is v0.25 ?

m4ksio:
No longer needed

PR #2404, opened 2022-05-11
Add weights to newly introduced memory kinds

SupunS:
Add weights to newly introduced memory kinds from https://github.com/onflow/cadence/pull/1631 and https://github.com/onflow/cadence/pull/1647.

Also update the cadence version to the HEAD of [feature/memory-metering](https://github.com/onflow/cadence/commits/feature/memory-metering)


turbolent (approved):
👍

PR #2407, opened 2022-05-12
Add another test for FVM exceeding memory limits

dsainati1:


fvm/fvm_test.go
```diff
@@ -3504,6 +3504,43 @@ func TestSettingExecutionWeights(t *testing.T) {
...
 		},
 	))
+
+	t.Run("transaction allocating high memory should fail", newVMTest().withBootstrapProcedureOptions(
+		fvm.WithMinimumStorageReservation(fvm.DefaultMinimumStorageReservation),
+		fvm.WithAccountCreationFee(fvm.DefaultAccountCreationFee),
+		fvm.WithStorageMBPerFLOW(fvm.DefaultStorageMBPerFLOW),
+		fvm.WithExecutionMemoryWeights(weightedMeter.DefaultMemoryWeights),
+	).withContextOptions(
+		fvm.WithMemoryLimit(10_000_000_000),
```
SupunS, line 3514: Would it make sense to have a much lower value as the memory limit, to avoid this test case from consuming lot of memory?

robert-e-davidson3 (approved):

ramtinms:
@dsainati1 do we need to make a PR to the master considering secure-cadence has been merged into master ?

dsainati1:
Huh to be honest I forgot about this PR; I'm not sure it's necessary because the new test isn't adding any coverage really, I think we can just close this.

PR #2410, opened 2022-05-12
add flaky skip reason so tests can run in cron job

kc1116:
This PR replaces the TEST_RESOURCE_INTENSIVE skip reason with the FLAKY skip reason so that the tests are not skipped entirely and can run in the cron job. Based on the metrics of other epoch join and leave tests for [LN](https://dapperlabs.grafana.net/d/toErpbynz/single-test-metrics?orgId=1&from=now-30d&to=now&var-package=github.com%2Fonflow%2Fflow-go%2Fintegration%2Ftests%2Fepochs&var-dataset_name=dapperlabs-data.production_src_flow_test_metrics&var-skipped_tests_table=skipped_tests&var-test_results_table=test_results&var-query_limit=100000&var-test=TestEpochJoinAndLeaveLN) and [SN](https://dapperlabs.grafana.net/d/toErpbynz/single-test-metrics?orgId=1&from=now-30d&to=now&var-package=github.com%2Fonflow%2Fflow-go%2Fintegration%2Ftests%2Fepochs&var-dataset_name=dapperlabs-data.production_src_flow_test_metrics&var-skipped_tests_table=skipped_tests&var-test_results_table=test_results&var-query_limit=100000&var-test=TestEpochJoinAndLeaveSN) nodes the pass rate is very high, so these tests should have a similar pass rate.

### DIFF
--- a/integration/tests/epochs/epoch_join_and_leave_an_test.go
+++ b/integration/tests/epochs/epoch_join_and_leave_an_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestEpochJoinAndLeaveAN(t *testing.T) {
-	unittest.SkipUnless(t, unittest.TEST_RESOURCE_INTENSIVE, "epochs join/leave tests should be run on an machine with adequate resources")
+	unittest.SkipUnless(t, unittest.TEST_FLAKY, "epochs join/leave tests should be run on an machine with adequate resources")
 	suite.Run(t, new(EpochJoinAndLeaveANSuite))
 }
 
--- a/integration/tests/epochs/epoch_join_and_leave_vn_test.go
+++ b/integration/tests/epochs/epoch_join_and_leave_vn_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 func TestEpochJoinAndLeaveVN(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jordanschalm (approved):

integration/tests/epochs/epoch_join_and_leave_an_test.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func TestEpochJoinAndLeaveAN(t *testing.T) {
-	unittest.SkipUnless(t, unittest.TEST_RESOURCE_INTENSIVE, "epochs join/leave tests should be run on an machine with adequate resources")
+	unittest.SkipUnless(t, unittest.TEST_FLAKY, "epochs join/leave tests should be run on an machine with adequate resources")
```
gomisha, line 12: should the reason be updated? Is this to test if running on cron job will be ok for these tests and they might not require bigger hardware?

kc1116: Yes so I would like to keep the message the same because it's the reason why the test are labeled as flaky.

kc1116:
bors merge

PR #2412, opened 2022-05-12
Implement Connection Pooling for RPC Connections in AN

lolpuddle:
This change allows AN to pool connections to LN and EN so that the connections may be re-used, and do not need to be re-initialized. Closes [3351](https://github.com/dapperlabs/flow-go/issues/3351)


engine/access/rpc/engine.go
```diff
@@ -128,11 +130,20 @@ func New(log zerolog.Logger,
 	// wrap the unsecured server with an HTTP proxy server to serve HTTP clients
 	httpServer := NewHTTPServer(unsecureGrpcServer, config.HTTPListenAddr)
 
+	cacheSize := config.ConnectionPoolSize
+	if cacheSize == 0 {
+		cacheSize = 100
+	}
+	cache, err := lru.New(cacheSize)
+	if err != nil {
+		// fmt.Errorf("cannot create conenction pool cache %w", err)
```
lolpuddle, line 139: Uncertain about what the best way of handling failure to init the cache would be, also uncertain about a good default size to set the cache to (above)

koko1123: if thew Cache fails to initialize, so should the server, right? not being able to connect to EN/CN in it's current state should be fatal

koko1123: also let's start with a size like `10` and go from there

engine/access/rpc/backend/connection_factory.go
```diff
@@ -61,16 +68,33 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
 	return conn, nil
 }
 
+func (cf *ConnectionFactoryImpl) retrieveConnection(address string) (*grpc.ClientConn, error) {
```
peterargue, line 71: this could accept a `time.Duration` and pass that into `createConnection`, that way we can maintain the existing timeouts

engine/access/rpc/engine.go
```diff
@@ -128,11 +130,22 @@ func New(log zerolog.Logger,
...
 	httpServer := NewHTTPServer(unsecureGrpcServer, config.HTTPListenAddr)
 
+	cacheSize := config.ConnectionPoolSize
+	if cacheSize == 0 {
+		cacheSize = backend.DefaultConnectionPoolSize
+	}
+	cache, err := lru.New(cacheSize)
+	if err != nil {
+		log.Err(err).Msg("could not initialize connection pool cache")
+		return nil, err
```
peterargue, line 140: You can just wrap the err with context and return it, then let the caller log it.

if you unravel the calls, what will happen is the RPC engine returns the error to the component factory method, which then returns it to the builder, which finally throws a fatal error. So in the end, you'll get the error log message and fatal error for free

engine/access/rpc/backend/connection_factory.go
```diff
@@ -61,16 +68,33 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
...
+	if res, ok := cf.ConnectionsCache.Get(address); ok {
+		conn = res.(*grpc.ClientConn)
+	}
+	if conn == nil || conn.GetState() != connectivity.Ready {
+		conn, err := cf.createConnection(address, cf.ExecutionNodeGRPCTimeout)
+		if err != nil {
+			return nil, err
+		}
+		cf.ConnectionsCache.Add(address, conn)
+	}
```
peterargue, line 82: I think it would be helpful to add metrics that track connection reuse. Maybe metrics for total connections and connections reused.

engine/access/rpc/backend/connection_factory.go
```diff
@@ -61,16 +68,33 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
 	return conn, nil
 }
 
+func (cf *ConnectionFactoryImpl) retrieveConnection(address string) (*grpc.ClientConn, error) {
+	var conn *grpc.ClientConn
+	if res, ok := cf.ConnectionsCache.Get(address); ok {
+		conn = res.(*grpc.ClientConn)
+	}
+	if conn == nil || conn.GetState() != connectivity.Ready {
+		conn, err := cf.createConnection(address, cf.ExecutionNodeGRPCTimeout)
```
peterargue, line 77: otherwise, `conn` here is a different variable than the one returned on line 83
```suggestion
		var err error
		conn, err = cf.createConnection(address, cf.ExecutionNodeGRPCTimeout)
```

engine/access/rpc/backend/connection_factory_test.go
```diff
@@ -33,6 +36,14 @@ func TestProxyAccessAPI(t *testing.T) {
 	connectionFactory := new(ConnectionFactoryImpl)
 	// set the collection grpc port
 	connectionFactory.CollectionGRPCPort = cn.port
+	// set the connection pool cache size
+	cache, _ := lru.New(5)
+	connectionFactory.ConnectionsCache = cache
+	// set metrics reporting
+	metrics := new(modulemock.TransactionMetrics)
+	metrics.On("ConnectionFromPoolRetrieved").Return(nil)
+	metrics.On("TotalConnectionsInPool", testifymock.Anything, testifymock.Anything).Return(nil)
```
peterargue, line 45: Rather than mocking the metrics, you can just use the `metrics.NewNoopCollector()`

peterargue (approved):
looks good. great work @lolpuddle!

koko1123 (approved):
nice work!

PR #2424, opened 2022-05-13
[FVM] fix a bug with invalid signature error type

tarakby:
A bug in a recent PR swaps the error sentinels returned by `TransactionSignatureVerifier` for invalid envelope and payload signatures. 
This PR fixes the bug and adds missing tests for error types:
 - `InvalidEnvelopeSignatureError` should be returned when a transaction has an invalid envelope signature.
 - `InvalidPayloadSignatureError` should be returned when a transaction has an invalid payload signature.
 - `InvalidEnvelopeSignatureError` should be returned when a transaction has invalid envelope and payload signatures. (this test is skipped for now since the `TransactionSignatureVerifier` doesn't implement these checks in the right order, though this doesn't have any security impact)


janezpodhostnik (approved):
looks good to me.

Just a note: this might affect some tests in flow-cli or flow-emulator or flow-go-sdk.

ramtinms (approved):

tarakby:
bors merge

PR #2429, opened 2022-05-16
Add `util` command to retrieve snapshot at any height

jordanschalm:
This is useful for recovery in certain cases. We can retrieve a snapshot from a suitable point in time, then start up the node using the snapshot, rather than requiring the node to sync all history.


jordanschalm:
Add comment to PR - pick block right before current epoch

zhangchiqing (approved):

PR #2430, opened 2022-05-16
Add more transactions to the simulator

SupunS:
Adding some more operations to the transaction generator, to help narrow down/analyze estimations of more granular operations.

Depends on: https://github.com/onflow/flow-go/pull/2351


fvm/parameter_calibration/TransactionGenerator.go
```diff
@@ -265,6 +265,14 @@ var Pool = TransactionTypePool{
 			body:     `let vaultRef = signer.borrow<&FlowToken.Vault>(from: /storage/flowTokenVault)!`,
 			name:     "get signer vault",
 		},
+		&SimpleTxType{
+			paramMax: 4394,
```
SupunS, line 269: @janezpodhostnik I'm not sure what value should go here and the rest of the newly added ones. This is a random value (copied from another tx). Need to update it with a proper value.

janezpodhostnik: The generator tries to generate transactions that are between 500 ms and 0 ms by varying the random loop length upper bound. To do this it needs to know approximately how many iterations of the loop are needed to get to 500 ms `N`. As the generator runs `N` is updated and gets progressively better. The paramMax is the initial guess for N. Is it is to low the first few transactions might be really short, but if it's to high they might be absurdly long.

As the generator runs it periodically outputs the "slope" ( how long it takes to run one iteration), which can be used for the initial guess for future runs. `paramMax = 500/slope`

PR #2436, opened 2022-05-17
PR Migrate ingestion engine to new Component interface

danielholmes839:
PR to migrate the access node ingestion engine to follow to the `Component` interface

[Reference migration](https://github.com/onflow/flow-go/pull/2035/files#diff-09bd1de20e1e535512432a5e994d212b95f89937bda8a4df0020e9875db5e3bf) of the collection node ingestion engine

Questions / Feedback needed
- I did not include metrics for the queues or configurable max queue lengths. See [this](https://github.com/onflow/flow-go/pull/2035/files#diff-09bd1de20e1e535512432a5e994d212b95f89937bda8a4df0020e9875db5e3bfR76).  Should these features be added?
-  I added a [`DisableProcessBackground` method  ](https://github.com/onflow/flow-go/pull/2436/files#diff-d322c324f42da4ad3934bfa1be878f6a9e25df0ed579872c72dc26fed0df9d93R188)to avoid calling mocks [here](https://github.com/onflow/flow-go/pull/2436/files#diff-d322c324f42da4ad3934bfa1be878f6a9e25df0ed579872c72dc26fed0df9d93R200) during certain tests. The code that's avoided was not executed during the tests before the changes. Is there a preferred solution? I tried fixing the mocks but that wasn't going well.
- Multiple small changes were made to the unit tests


engine/access/ingestion/engine.go
```diff
@@ -92,10 +102,45 @@ func New(
 	blocksToMarkExecuted *stdmap.Times,
 	rpcEngine *rpc.Engine,
 ) (*Engine, error) {
+	executionReceiptsRawQueue, err := fifoqueue.NewFifoQueue(
+		fifoqueue.WithCapacity(10_000),
```
peterargue, line 106: can you make `10_000` a constant?

engine/access/ingestion/engine_test.go
```diff
@@ -134,28 +143,26 @@ func (suite *Suite) TestOnFinalizedBlock() {
...
 			delete(needed, collID)
+			wg.Done()
 		},
 	)
 
 	// process the block through the finalized callback
 	suite.eng.OnFinalizedBlock(&hotstuffBlock)
-
-	// wait for engine shutdown
-	done := suite.eng.unit.Done()
```
peterargue, line 150: I think you'll still need to shutdown the engine at the end of the test, otherwise it will leak go routines. Similar to what you did in `TestComponentShutdown`

peterargue (commented):
great job! added some comments

engine/access/ingestion/engine.go
```diff
@@ -111,100 +159,204 @@ func New(
...
-		err := e.requestMissingCollections(ctx)
-		if err != nil {
-			e.log.Error().Err(err).Msg("requesting missing collections failed")
-		}
-	})
-	e.unit.LaunchPeriodically(e.updateLastFullBlockReceivedIndex, defaultFullBlockUpdateInterval, time.Duration(0))
-	return readyChan
+func (e *Engine) DisableProcessBackground() {
+	// disable the processBackground worker from starting during testing to avoid calling mocks
+	e.processBackgroundDisabled = true
```
smnzhu, line 193: It is better to avoid this, this isn't a good practice. The proper way to address this would be to setup the mocks correctly.

smnzhu (changes requested):

peterargue (approved):
nice work!

PR #2437, opened 2022-05-17
Memory limit exception for service account

turbolent:
Closes dapperlabs/cadence-private-issues#50

Just like we already have an exception for the interaction limit for the service account, also add an exception for the memory limit.

### DIFF
--- a/fvm/fvm_test.go
+++ b/fvm/fvm_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/onflow/flow-go/crypto"
 	"github.com/onflow/flow-go/crypto/hash"
+
 	"github.com/onflow/flow-go/engine/execution/testutil"
 	exeUtils "github.com/onflow/flow-go/engine/execution/utils"
 	"github.com/onflow/flow-go/fvm"
@@ -3570,6 +3571,7 @@ func TestTransactionFeeDeduction(t *testing.T) {
 }
 
 func TestSettingExecutionWeights(t *testing.T) {
+
 	t.Run("transaction should fail with high weights", newVMTest().withBootstrapProcedureOptions(
 		fvm.WithMinimumStorageReservation(fvm.DefaultMinimumStorageReservation),
 		fvm.WithAccountCreationFee(fvm.DefaultAccountCreationFee),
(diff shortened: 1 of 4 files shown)


fvm/fvm_test.go
```diff
@@ -3607,12 +3609,58 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+				AddAuthorizer(accounts[0]).
+				SetPayer(accounts[0])
+
+			err = testutil.SignTransaction(txBody, accounts[0], privateKeys[0], 0)
+			require.NoError(t, err)
+
+			tx := fvm.Transaction(txBody, 0)
+			err = vm.Run(ctx, tx, view, programs)
+			require.NoError(t, err)
+			require.Greater(t, tx.MemoryUsed, uint64(20_000_000_000))
```
Kay-Zee, line 3657: is this greater than or equal to? or is there something about how this is setup that it's never equal to the limit we input? like it always adds a bool size to our input number or something?

Also, since we use this to set + check, we might as well pull it into a var, maybe just before setting it to `memoryWeights[common.MemoryKindBool]`

fvm/fvm_test.go
```diff
@@ -3642,11 +3690,12 @@ func TestSettingExecutionWeights(t *testing.T) {
 			tx := fvm.Transaction(txBody, 0)
 			err = vm.Run(ctx, tx, view, programs)
 			require.NoError(t, err)
-			require.Greater(t, tx.MemoryUsed, uint64(20_000_000_000))
+			require.Equal(t, uint64(0), tx.MemoryUsed)
```
Kay-Zee, line 3693: oh, the return is zero if there is no limit set? interesting.

fvm/state/state_holder.go
```diff
@@ -35,7 +35,7 @@ func (s *StateHolder) SetActiveState(st *State) {
 	s.activeState = st
 }
 
-// SetActiveState sets active state
+// SetPayerIsServiceAccount sets if the payer is the service account
```
Kay-Zee, line 38: Thanks for the fix here

SupunS (approved):

Kay-Zee (commented):
I noticed another var, `EnforceComputationLimits`, will that need this treatment as well?

ramtinms (approved):

turbolent:
@Kay-Zee As discussed, this PR is regarding memory metering, so computation limit is out-of-scope for this PR. You might be right, but I know too little about FVM

fvm/fvm_test.go
turbolent: Total memory usage depends on many weights which results in a very particular concrete value here. 
I kept the assertion as it was before (the diff looks like added code, but it is actually existing code): We assert that at least the very high weight for booleans was correctly included in the total memory usage.

PR #2438, opened 2022-05-17
Add method for timeout weight threshold to `Committee`

jordanschalm:


consensus/hotstuff/committee.go
```diff
@@ -38,12 +38,20 @@ type Replicas interface {
...
-	WeightThresholdForView(view uint64) (uint64, error)
+	QuorumThresholdForView(view uint64) (uint64, error)
+
+	// TimeoutThresholdForView returns the minimum total weight of observed timeout objects
+	// required to safely timeout for the given view. This weight threshold is computed
+	// using the total weight of the initial committee and is static over the course of
+	// an epoch.
+	// Returns the following expected errors for invalid inputs:
+	//   * model.ErrViewForUnknownEpoch if no epoch containing the given view is known
+	TimeoutThresholdForView(view uint64) (uint64, error)
```
durkmurder, line 54: :warning: To create TC we need the same threshold as for QC(2f+1). We need f+1 threshold for a specific liveness optimization called Bracha timeout mechanism. In our architecture proposal, Algorithm 2(Pacemaker) emits local timeout whenever it observes f+1 timeouts, this is still not enough to construct TC(2f+1) but based on quorum intersection lemma, already enough to understand that quorum eventually will be reached. Lemma 8 uses this property. 

Hopefully this context is useful. We need to change naming and comment.

durkmurder (approved):
Looks good, thank you

consensus/hotstuff/committee.go
```diff
@@ -38,12 +38,20 @@ type Replicas interface {
 	//   * model.ErrViewForUnknownEpoch if no epoch containing the given view is known
 	LeaderForView(view uint64) (flow.Identifier, error)
 
-	// WeightThresholdForView returns the minimum weight required to form a QC
+	// QuorumThresholdForView returns the minimum total weight of votes required to form a QC
 	// for the given view. This weight threshold is computed using the total weight
```
AlexHentschel, line 42: I feel the goDoc is a bit over specific. In addition to a QC, we also use this threshold for a Timeout Certificate. 

```suggestion
	// QuorumThresholdForView returns the minimum total weight for a supermajority 
	// at the given view. This weight threshold is computed using the total weight
```

consensus/hotstuff/committees/threshold.go
```diff
@@ -1,6 +1,7 @@
 package committees
 
-// WeightThresholdToBuildQC returns the weight that is minimally required for building a QC.
+// WeightThresholdToBuildQC returns the weight (sum of unique, valid votes for this view)
+// that is minimally required for building a QC.
```
AlexHentschel, line 4:
```suggestion
// that is minimally required for a supermajority.
```

consensus/hotstuff/committees/threshold_test.go
```diff
@@ -17,3 +18,16 @@ func TestComputeWeightThreshold(t *testing.T) {
 		assert.False(t, boundaryValue < float64(threshold-1))
 	}
 }
+
+// TestComputeTOWeightThreshold tests computing the HotStuff safety threshold
+// for producing a TO.
+func TestComputeTOWeightThreshold(t *testing.T) {
```
AlexHentschel, line 24: * Could we rename `i` to `totalWeight`? 
* I would suggest to change the test slightly. There are two equations that must be satisfied for `threshold` to be correct
   * `threshold * 3 > totalWeight`
   * `(threshold - 1) * 3  <= totalWeight`

   The reason why I prefer these checks is that we don't imitate the computation of the threshold.

AlexHentschel (approved):
nice. my suggestions are only for aesthetics

jordanschalm:
bors merge

jordanschalm:
bors retry

PR #2441, opened 2022-05-17
[BFT Testing] Corruptible Conduit Factory authenticating dictated messages. 

yhassanzadeh13:
This PR enables Corruptible Conduit Factory to fill in the execution receipt and result approvals it receives from the attacker by adding authentication metadata, e.g., signature, identifier, etc. The execution receipt and result approvals are wintermute-related contexts.


insecure/corruptible/factory.go
```diff
@@ -297,11 +328,21 @@ func (c *ConduitFactory) sendOnNetwork(event interface{},
...
 
 	case insecure.Protocol_MULTICAST:
 		return c.adapter.MulticastOnChannel(channel, event, num, targetIds...)
 	default:
 		return fmt.Errorf("unknown protocol for sending on network: %d", protocol)
 	}
 }
+
+func (c *ConduitFactory) generateExecutionReceipt(result *flow.ExecutionResult) (*flow.ExecutionReceipt, error) {
+	// TODO: fill spock secret with dictated spock data from attacker.
```
yhassanzadeh13, line 341: Note that current state of Flow Protocol does not check SPOCK proofs generated by ENs and VNs, hence, we leave the SPOCK secret empty for now, as enabling CCF to generate SPOCK proof needs extra efforts which is out of the scope of this PR. The wintermute attack should work with empty SPOCK proof.

insecure/corruptible/factory.go
```diff
@@ -32,18 +39,21 @@ type ConduitFactory struct {
 	cm                    *component.ComponentManager
 	logger                zerolog.Logger
 	codec                 network.Codec
-	myId                  flow.Identifier
+	me                    module.Local
```
gomisha, line 42: can this be renamed to something more clear? `mod` or `module`?

insecure/corruptible/factory.go
```diff
@@ -32,18 +39,21 @@ type ConduitFactory struct {
...
+	receiptHasher         hash.Hasher
+	spockHasher           hash.Hasher
+	approvalHasher        hash.Hasher
 }
 
 func NewCorruptibleConduitFactory(
 	logger zerolog.Logger,
 	chainId flow.ChainID,
-	myId flow.Identifier,
+	me module.Local,
```
gomisha, line 56: same as before - is there a reason it's name `me`? Can it be renamed to be more clear?

insecure/wintermute/helpers.go
```diff
@@ -201,24 +201,29 @@ func orchestratorOutputSanityCheck(
...
 
 	for _, outputEvent := range outputEvents {
 		switch event := outputEvent.FlowProtocolEvent.(type) {
 		case *flow.ExecutionReceipt:
-			// makes sure sender is a corrupted execution node.
-			ok := corrEnIds.Contains(outputEvent.CorruptedNodeId)
-			require.True(t, ok)
-			// uses union to avoid adding duplicate.
-			bouncedReceipts = bouncedReceipts.Union(flow.IdentifierList{event.ID()})
-		case *flow.ExecutionResult:
```
gomisha, line 215: why are we not checking for execution results anymore?

insecure/wintermute/helpers.go
```diff
@@ -201,24 +201,29 @@ func orchestratorOutputSanityCheck(
...
-			resultId := event.ID()
-			if dictatedResults[resultId] == nil {
-				dictatedResults[resultId] = flow.IdentifierList{}
+			if len(event.ExecutorSignature.Bytes()) != 0 {
+				// a receipt with a non-empty signature is a pass-through receipt.
```
gomisha, line 211: can we clarify this more? Does a receipt with a non-empty signature ALWAYS mean it's from a non-corrupted node and that's why it's passed through?

insecure/wintermute/helpers.go
```diff
@@ -201,24 +201,29 @@ func orchestratorOutputSanityCheck(
 	// keeps a map of (corrupted results ids -> execution node ids)
 	dictatedResults := make(map[flow.Identifier]flow.IdentifierList)
 
-	// keeps a list of all bounced back events.
-	bouncedReceipts := flow.IdentifierList{}
+	// keeps a list of all pass through receipts.
```
gomisha, line 204: can we clarify this more - WHY are keeping a list of passed through receipts?

insecure/wintermute/helpers.go
```diff
@@ -232,7 +237,7 @@ func orchestratorOutputSanityCheck(
 	// number of bounced receipts should match the expected value.
```
gomisha, line 237:
```suggestion
	// number of passed through receipts should match the expected value.
```

insecure/corruptible/factory.go
```diff
@@ -194,16 +223,16 @@ func (c *ConduitFactory) processAttackerMessage(msg *insecure.Message) error {
 // RegisterAttacker is a gRPC end-point for this conduit factory that lets an attacker register itself to it, so that the attacker can
 // control it.
 // Registering an attacker on a conduit is an exactly-once immutable operation, any second attempt after a successful registration returns an error.
-func (c *ConduitFactory) RegisterAttacker(ctx context.Context, in *insecure.AttackerRegisterMessage) (*empty.Empty, error) {
+func (c *ConduitFactory) RegisterAttacker(_ context.Context, in *insecure.AttackerRegisterMessage) (*empty.Empty, error) {
```
gomisha, line 226: do we still need to pass the context even though it's not used anymore by `RegisterAttacker()`?

insecure/corruptible/factory.go
peterargue: `me` is the conventional name we use for the local node info

insecure/corruptible/factory.go
```diff
@@ -178,6 +191,30 @@ func (c *ConduitFactory) processAttackerMessage(msg *insecure.Message) error {
 		return fmt.Errorf("could not decode message: %w", err)
 	}
 
+	switch e := event.(type) {
+	case *flow.ExecutionReceipt:
+		if len(e.ExecutorSignature) == 0 {
```
peterargue, line 196: is this something that will always be true, or could it be specific to the attack orchestrator? Wondering if this check belongs in the `generateExecutionReceipt` implementation

peterargue (approved):
looks good. I had one question

insecure/corruptible/factory.go
yhassanzadeh13: It is specific to the attack orchestrator basically deciding whether to generate an execution receipt and override the current event with it, or just to pass the current event as it is without any modification. Hence, the logic is better to be separated from the creational pattern of `generateExecutionReceipt` that is responsible to generate an execution receipt out of the given result. 

It also applies a better level of information hiding and encapsulation, i.e., `generateExecutionReceipt` only receives the result part of the receipt, and does not need to be concern about other parts of that, while the inspection of the receipt is happening at the orchestrator level.

PR #2442, opened 2022-05-17
[Tool] Export transactions for height range

zhangchiqing:
Close https://github.com/dapperlabs/flow-go/issues/6236

This PR adds a util commandline and admin tool to get transactions by height range to run on execution node or access node.

## How to test
1. start the network, and start the loader to create some txs
```
cd integration/localnet
make init
make start
make load
```
2. run admin tool against execution node 
```
curl localhost:3701/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-transactions", "data": { "start-height": 340, "end-height": 343 }}'
```
This will return all transactions for blocks between height 340 to 343.

3. stop the network, and run util
```
make stop
cd ../../
go run --tags=relic ./cmd/util/main.go export-json-transactions --output-dir ./ --datadir ./integration/localnet/data/access/adbeb15af18a5e4a84d9201cee5205f112d4eda982d5540efba6106b8b02e3d3/protocol --start-height 340 --end-height 343
```
open `transactions.json` to see all transactions grouped by block and collection.


admin/commands/storage/read_transactions.go
```diff
@@ -0,0 +1,98 @@
...
+		return err
+	}
+
+	endHeight, err := findUint64(input, "end-height")
+	if err != nil {
+		return err
+	}
+
+	if endHeight-startHeight > 99 {
+		return fmt.Errorf("getting transactions for more than 100 blocks at a time might have an impact to node's performance")
```
zhangchiqing, line 89: This is to protect from accidentially querying too many blocks

jordanschalm (approved):

cmd/util/cmd/export-json-transactions/cmd.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// create JSON file writer
+	outputFile := filepath.Join(outputDir, "transactions.json")
+	fi, err := os.Create(outputFile)
+	if err != nil {
+		return fmt.Errorf("could not create block output file %v, %w", outputFile, err)
+	}
+	defer fi.Close()
+
+	blockWriter := bufio.NewWriter(fi)
```
SaveTheRbtz, line 91: not sure whether bufio is useful here, given that data is stored in an array of bytes

cmd/util/cmd/export-json-transactions/cmd.go
```diff
@@ -0,0 +1,122 @@
...
+func run(*cobra.Command, []string) {
+	log.Info().Msg("start exporting transactions")
+	err := ExportTransactions(flagDatadir, flagOutputDir, flagStartHeight, flagEndHeight)
+	if err != nil {
+		log.Fatal().Err(err).Msg("cannot get export transactions")
+	}
+	log.Info().Msg("transactions exported")
+}
+
+func writeJSONTo(writer io.Writer, jsonData []byte) error {
```
SaveTheRbtz, line 57: This looks like an re-implementation of the [os.WriteFile](https://pkg.go.dev/os#WriteFile)

cmd/util/cmd/export-json-transactions/cmd.go
```diff
@@ -0,0 +1,122 @@
...
+	blockWriter := bufio.NewWriter(fi)
+
+	// build all blocks first before writing to disk
+	// TODO: if the height range is too high, consider streaming json writing for each block
+	blocks, err := finder.GetByHeightRange(startHeight, endHeight)
+	if err != nil {
+		return err
+	}
+
+	log.Info().Msgf("exporting transactions for %v blocks", endHeight-startHeight+1)
```
SaveTheRbtz, line 100: nit: generally instead of %v you want to use .Int*() methods

cmd/util/cmd/export-json-transactions/transactions/range_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		require.Len(t, fetched[0].Collections[0].Transactions, 1)
+		require.Len(t, fetched[0].Collections[1].Transactions, 2)
+		require.Len(t, fetched[1].Collections[0].Transactions, 3)
+
+		require.Equal(t,
+			fetched[0].Collections[0].Transactions[0].TxID,
+			col1.Collection.Transactions[0].ID().String(),
+		)
+
```
SaveTheRbtz, line 91: you may want to try a couple of error cases like 
```
f.GetByHeightRange(5, 4)
f.GetByHeightRange(99998, 99999)
f.GetByHeightRange(0, 999999999)
```

cmd/util/cmd/export-json-transactions/cmd.go
zhangchiqing: This is to leave room for future improvement if needed, which could minimize the memory usage. 

If someone query a large range or a block that contains a lot of blocks, which will take lots of memory space to hold these blocks, before saving them to disk as JSON. 

A better way is to pipeline it, query one block at a time, and then write it to the json file, and release the data for the block, and work on the next block. This way we only hold memory for up to a block's data rather than the entire range of blocks data. 

So I made a `RunByHeightRange` function which takes a `run func(*BlockData) error` for the bufio to write data. 

I didn't have time to implement the optimal solution yet. But still kept the bufio here in case needed.

cmd/util/cmd/export-json-transactions/cmd.go
zhangchiqing: Yeah, I was thinking to keep the option of using Writer for pipelining the block processing

cmd/util/cmd/export-json-transactions/cmd.go
zhangchiqing: Sorry, not quite sure what's your suggested change, do you mean use `%d` instead?

cmd/util/cmd/export-json-transactions/transactions/range_test.go
zhangchiqing: tests are added

cmd/util/cmd/export-json-transactions/cmd.go
SaveTheRbtz: I meant the static-typed methods of the logger:
```
log.Info().Uint64("numBlocks", endHeight-startHeight+1).Msg("exporting transactions")
```

This allows for:
* no memory allocations
* no `interface{}` formatting overhead (actually, no overhead at all since it is now `Msg` instead of `Msgf`).
* ease of the followup "grepping" via `| json | numBlocks > 100` in [LogQL](https://grafana.com/docs/loki/latest/logql/).

SaveTheRbtz (approved):

zhangchiqing:
bors merge

zhangchiqing:
bors retry

PR #2444, opened 2022-05-18
[Consensus] Active Pacemaker `SafetyRules` implementation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6193

### Context

This PR implements internal logic for `SafetyRules` interface. `SafetyRules` contains logic for persisting protocol safety. It's responsible for:
- creating signed votes
- creating signed timeouts
- voting/timeout creation rules
- persisting of local state

This PR has major rework of `Persister` in this PR we need only the `SafetyData` part but I've implemented also the `LivenessData` part(needed for `PaceMaker` PR)


consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -8,30 +8,31 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 )
 
-// SafetyRules produces votes for the given block
+// SafetyRules is a dedicated module that is responsible for persisting chain safety by producing
+// votes and timeouts. It follows voting and timeout rules for creating votes and timeouts respectively.
+// Caller can be sure that created vote or timeout doesn't break safety and can be used in consensus process.
+// SafetyRules relies on hotstuff.Persister to store latest state of hotstuff.SafetyData.
 type SafetyRules struct {
```
jordanschalm, line 15:
```suggestion
// SafetyRules relies on hotstuff.Persister to store latest state of hotstuff.SafetyData.
// 
// The voting rules implemented by SafetyRules are:
// 1. Replicas vote strictly in increasing rounds
// 2. Each block has to include a TC or a QC from the previous round. 
//   a. [Happy path] If the previous round resulted in a QC then new QC should extend it.
//   b. [Recovery path] If the previous round did *not* result in a QC, the leader of the 
//   subsequent round *must* include a valid TC for the previous round in its block.
//
// NOT safe for concurrent use. 
type SafetyRules struct {
```

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +105,76 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
+	}
+
+	// This check satisfies voting rule 1 and 2a.
+	if blockView == qcView+1 {
+		return nil
+	}
+
+	// This check satisfies voting rule 1 and 2b.
+	lastViewTC := proposal.LastViewTC
+	if lastViewTC != nil {
```
jordanschalm, line 156: I think it is worth factoring this into a `safeToExtend` function, like in the formal algorithm.
```suggestion
	return isSafeToExtend(blockView, qcView, lastViewTC)
```

What do you think about handling the failure cases first to reduce nesting and move the errors closer to the conditions that cause them? eg:
```
if blockView != lastViewTC.View+1 {
  // error
}

if qcView < lastViewTC.TOHighestQC.View {
  // error
}

// We satisfy voting rule 1 and 2b ...
return nil

File: consensus/hotstuff/safetyrules/safety_rules.go [hunk 3]
```diff
@@ -93,16 +105,76 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
+			}
+		} else {
+			return fmt.Errorf("last view TC %d is not sequential for block %d", lastViewTC.View, blockView)
+		}
+	}
+
+	return fmt.Errorf("block's view %d is not sequential %d, last view TC not included", blockView, qcView)
 }
 
 // IsSafeToTimeout checks if it's safe to timeout with proposed data, if timing out won't break safety rules.
```
jordanschalm, line 171:
```suggestion
// IsSafeToTimeout checks if it's safe to timeout with proposed data, if timing out won't break safety rules.
// highestQC is the valid QC with the greatest view that we have observed.
// lastViewTC is the TC for the previous view (may be nil)
```

I think the TC argument here is the TC for the last view, not the highest TC.

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +105,76 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
-func (v *SafetyRules) IsSafeToTimeout(curView uint64, highestQC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) bool {
-	panic("to be implemented")
+func (r *SafetyRules) IsSafeToTimeout(curView uint64, highestQC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) bool {
+	if highestQC.View < r.safetyData.LockedOneChainView ||
+		curView+1 <= r.safetyData.HighestAcknowledgedView ||
+		curView <= highestQC.View {
+		return false
+	}
+
+	return (curView == highestQC.View+1) || (curView == highestTC.View+1)
```
jordanschalm, line 179:
```suggestion
	if curView == highestQC.View+1 {
		return true
	}
	
	// sanity check: lastViewTC cannot be nil if last view did not result in QC
	if lastViewTC == nil { 
		return false
	}
	if curView == lastViewTC.View+1 {
		return true
	}
	return false
```

Suggestion to handle `lastViewTC` being unexpectedly nil.

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +105,76 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
 //  * (nil, model.NoTimeoutError): If the safety module decides that it is not safe to timeout under current conditions.
 //    This is a sentinel error and _expected_ during normal operation.
 // All other errors are unexpected and potential symptoms of uncovered edge cases or corrupted internal state (fatal).
-func (v *SafetyRules) ProduceTimeout(curView uint64, highestQC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
-	panic("to be implemented")
+func (r *SafetyRules) ProduceTimeout(curView uint64, highestQC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
```
jordanschalm, line 108:
```suggestion
func (r *SafetyRules) ProduceTimeout(curView uint64, lastViewTC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
```

(Also would update the godoc, similarly to `IsSafeToTimeout`)

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +105,76 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
+		return lastTimeout, nil
+	}
+
+	if !r.IsSafeToTimeout(curView, highestQC, highestTC) {
+		return nil, model.NoTimeoutError{Msg: "not safe to time out under current conditions"}
+	}
+
+	timeout, err := r.signer.CreateTimeout(curView, highestQC, highestTC)
+	if err != nil {
+		return nil, fmt.Errorf("could not timeout at view %d: %w", curView, err)
```
jordanschalm, line 120:
```suggestion
		return nil, fmt.Errorf("could not create timeout at view %d: %w", curView, err)
```

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -45,42 +46,53 @@ func New(
 //  * (nil, model.NoVoteError): If the voter decides that it does not want to vote for the given block.
 //    This is a sentinel error and _expected_ during normal operation.
 // All other errors are unexpected and potential symptoms of uncovered edge cases or corrupted internal state (fatal).
```
jordanschalm, line 48: We should take a pass over the godoc here, eg `SafetyRules will ask Forks whether a vote is a safe node or not` is outdated.

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -45,42 +46,53 @@ func New(
...
 	}
-	if curView <= v.lastVotedView {
-		return nil, fmt.Errorf("current view (%d) must be larger than the last voted view (%d)", curView, v.lastVotedView)
+
+	err := r.IsSafeToVote(proposal)
+	if err != nil {
+		return nil, fmt.Errorf("not safe to vote for proposal %x: %w", proposal.Block.BlockID, err)
+	}
+
+	// we expect that only valid proposals are submitted for voting
```
jordanschalm, line 61: Could we add this assumption to the method godoc as well?

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -45,42 +46,53 @@ func New(
...
+	// we need to make sure that proposer is not ejected to decide to vote or not
+	_, err = r.committee.IdentityByBlock(block.BlockID, block.ProposerID)
+	if model.IsInvalidSignerError(err) {
+		return nil, model.NoVoteError{Msg: "not voting - proposer ejected"}
```
jordanschalm, line 65: Technically from just this check we can't assume the proposer is ejected. It could also be either non-existent or have zero weight.
```suggestion
		// the proposer must be ejected since the proposal has already been validated,
		// which ensures that the proposer was a valid committee member at the start of the epoch
		return nil, model.NoVoteError{Msg: "not voting - proposer ejected"}
```

jordanschalm (commented):
First round of comments

jordanschalm (commented):
Still working on reviewing the tests - second round of comments

state/protocol/badger/state.go
```diff
@@ -236,14 +238,33 @@ func (state *State) bootstrapStatePointers(root protocol.Snapshot) func(*badger.
...
+			// in this case we are dealing with non-root QC
+			parentView = segment.Blocks[len(segment.Blocks)-2].Header.View
+		}
+
+		livenessData := &hotstuff.LivenessData{
+			CurrentView: highest.Header.View + 1,
+			// TODO(active-pacemaker): update last view TC if available
+			//LastViewTC:  highest.Header.LastViewTC,
+			HighestQC: model.BlockFromFlow(highest.Header, parentView).QC,
+		}
```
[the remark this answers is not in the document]
durkmurder: :1st_place_medal:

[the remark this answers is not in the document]
durkmurder: but does it work for bootstrapping from sealing segment?

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,423 @@
...
+	exception := errors.New("invalid-signer-identity")
+	s.committee.On("IdentityByBlock", s.proposal.Block.BlockID, s.proposal.Block.ProposerID).Return(nil, exception).Once()
 
-	_, err := voter.ProduceVote(block, curView)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "expecting block for current view")
-	require.False(t, model.IsNoVoteError(err))
+	vote, err := s.safety.ProduceVote(s.proposal, s.proposal.Block.View)
+	require.Nil(s.T(), vote)
+	require.ErrorAs(s.T(), err, &exception)
```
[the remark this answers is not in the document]
durkmurder: Well, both `As` and `Is` are checking if this error is in the chain I believe, don't they?

state/protocol/badger/state.go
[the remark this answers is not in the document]
jordanschalm: It should work for all cases - it is what we use to bootstrap HotStuff (eg. [here](https://github.com/onflow/flow-go/blob/e175c1e049241dc80f871f554944ba2d6336d989/cmd/scaffold.go#L782))

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
jordanschalm: They both check the chain of errors, but `As` is checking type equality (any error from `errors` package would match, including some error instance other than `exception`) whereas `Is` is checking instance equality (only the specific `exception` error created in the test case would match). Here's an example: https://go.dev/play/p/Malz3OGjaoo

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,423 @@
...
+					helper.WithTCView(s.bootstrapBlock.View+1),
+					helper.WithTCHighestQC(TOHighestQC))))
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
+		require.Nil(s.T(), vote)
+	})
+
+	s.signer.AssertNotCalled(s.T(), "CreateVote")
+	s.signer.AssertNotCalled(s.T(), "PutSafetyData")
 }
```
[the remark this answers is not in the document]
durkmurder: I've added extra test case for 1. Regarding 2., from safety perspective if we see qc.View = proposal.View - 1, then we vote for this block, we don't care what QC or TC is included. From validation side we have to check if indeed included QC is higher or equal to TC's highest QC.

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
durkmurder: Thanks for example but as I understand it's still good in our case since we use `errors` package only for exceptions, and point of those test is to find out if the exception or sentinel was returned. `errors` can't be cast to our custom sentinels(they don't implement `As`). Isn't that right? 
I've changed everything to `ErrorIs`

consensus/hotstuff/verification/combined_signer_v2.go
```diff
@@ -90,6 +94,25 @@ func (c *CombinedSigner) CreateVote(block *model.Block) (*model.Vote, error) {
 	return vote, nil
 }
 
+// CreateTimeout will create a signed timeout object for the given view.
```
jordanschalm, line 97:
```suggestion
// CreateTimeout will create a signed timeout object for the given view.
// Timeout objects are only signed with the staking key (not beacon key).
```

consensus/hotstuff/verification/combined_signer_v3.go
```diff
@@ -91,6 +94,25 @@ func (c *CombinedSignerV3) CreateVote(block *model.Block) (*model.Vote, error) {
 	return vote, nil
 }
 
+// CreateTimeout will create a signed timeout object for the given view.
```
jordanschalm, line 97:
```suggestion
// CreateTimeout will create a signed timeout object for the given view.
// Timeout objects are only signed with the staking key (not beacon key).
```

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
jordanschalm: Thanks for updating, I'll try to better explain the reasoning 👇 

> point of those test is to find out if the exception or sentinel was returned

We wouldn't accidentally confuse a sentinel for an exception, this is more the case I'm thinking of:

```go
func ProduceVote() {
  // a component we do not mock failing can cause an exception to be returned
  // we don't expect this error as part of this hypothetical test case
  ok := component1.CheckSomething()
  if !ok {
    return errors.New("some exception")
  }

  // we mock component2 to return an exception that we expect to be returned
  err := component2.DoSomething()
  if err != nil {
    return err
  }
  // ...
}
```
If we use `As`, we're not sure whether the error returned is from `component2` (expected) or `component1` (unexpected). If we use `Is`, we can be sure exactly the error we expected (from `component2`) was returned. 

It's more a general good practice than a concrete problem with the test as it exists currently.

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,463 @@
...
+
+	s.signer.AssertNotCalled(s.T(), "CreateVote")
+	s.persister.AssertNotCalled(s.T(), "PutSafetyData")
+}
 
-	// create voter
-	block, _, voter := createVoter(blockView, lastVotedView, isBlockSafe, isCommitteeMember)
+// TestProduceTimeout_ShouldTimeout tests that we can produce timeout in cases where
+// last view was successful or not. Also tests last timeout caching.
+func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
```
jordanschalm, line 347: Suggestion for additional test cases:
* we should be able to produce a vote for a view, then timeout for the same view
* we should be unable to timeout for a view, then vote for the same view

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
jordanschalm: > Regarding 2., from safety perspective if we see qc.View = proposal.View - 1, then we vote for this block, we don't care what QC or TC is included

Agreed. This shouldn't make a difference in the Safety Module. I'll add some comments to validation section of Implementation Proposal.

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
durkmurder: Thanks, that makes sense!

AlexHentschel (commented):
Looks great, but didn't quite get through the PR entirely. Mostly stylistic comments. 

My only critical comments are regarding the  `SafetyRules`:
* I feel we have a few checks there, which we treat as expected errors, while they are in fact symptoms of internal bugs.

consensus/hotstuff/persister.go
```diff
@@ -3,16 +3,15 @@ package hotstuff
 // Persister is responsible for persisting state we need to bootstrap after a
 // restart or crash.
 type Persister interface {
```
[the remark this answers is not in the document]
durkmurder: Yes, that is right, no sentinel errors

consensus/hotstuff/pacemaker.go
```diff
@@ -7,6 +7,17 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 )
 
+type LivenessData struct {
+	// CurrentView is the current active view tracked by PaceMaker. It's updated whenever
+	// PaceMaker sees evidence(QC or TC) for advancing to next view.
+	CurrentView uint64
+	// HighestQC is the highest QC(by view) observed by PaceMaker. QC can be observed on its own or as a part of TC.
+	HighestQC *flow.QuorumCertificate
```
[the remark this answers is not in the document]
durkmurder: Updated all HighestQC -> NewestQC

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +118,86 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
-func (v *SafetyRules) ProduceTimeout(curView uint64, highestQC *flow.QuorumCertificate, highestTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
-	panic("to be implemented")
+func (r *SafetyRules) ProduceTimeout(curView uint64, highestQC *flow.QuorumCertificate, lastViewTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
+	lastTimeout := r.safetyData.LastTimeout
+	if lastTimeout != nil && lastTimeout.View == curView {
+		return lastTimeout, nil
+	}
+
+	if !r.IsSafeToTimeout(curView, highestQC, lastViewTC) {
+		return nil, model.NewNoTimeoutErrorf("not safe to time out under current conditions")
```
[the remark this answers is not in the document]
durkmurder: Same reasoning as above, I believe that's the main validation logic and we won't have "internal" logic to rely on in this case. It's not a sanity check.

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -93,16 +118,86 @@ func (v *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
...
+// highestQC is the valid QC with the greatest view that we have observed.
+// lastViewTC is the TC for the previous view (might be nil)
+func (r *SafetyRules) IsSafeToTimeout(curView uint64, highestQC *flow.QuorumCertificate, lastViewTC *flow.TimeoutCertificate) bool {
+	if highestQC.View < r.safetyData.LockedOneChainView ||
+		curView+1 <= r.safetyData.HighestAcknowledgedView ||
+		curView <= highestQC.View {
+		return false
+	}
+
+	return (curView == highestQC.View+1) || (curView == lastViewTC.View+1)
```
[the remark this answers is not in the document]
durkmurder: It would be a bug if we rely on very specific implementation of modules which provide `newestQC`, `curView` and `lastViewTC`. Don't have strong feelings here.

AlexHentschel:
Could you quickly rename `highQCView` to `newestQCView` here: https://github.com/onflow/flow-go/blob/af9fc5c3cdca409ac82457bf4237f250b46f96d6/consensus/hotstuff/verification/common.go#L24
for consistency of terminology. Thanks

state/protocol/badger/state.go
```diff
@@ -236,14 +237,30 @@ func (state *State) bootstrapStatePointers(root protocol.Snapshot) func(*badger.
...
 
+		safetyData := &hotstuff.SafetyData{
+			LockedOneChainView:      highest.Header.View,
+			HighestAcknowledgedView: highest.Header.View,
+		}
+
+		rootQC, err := root.QuorumCertificate()
+		if err != nil {
+			return fmt.Errorf("could not get root QC: %w", err)
+		}
```
AlexHentschel, line 248: Per convention, all blocks in the sealing segment must be finalized; see here for reference: https://github.com/onflow/flow-go/blob/af9fc5c3cdca409ac82457bf4237f250b46f96d6/state/protocol/badger/state.go#L271

Therefore, the `highest` block in the sealing segment must have a QC. Per convention, the QC for `highest` should be contained in the `root` Snapshot (otherwise, the Snapshot is incomplete. I think we should add sanity checks here to enforce this convention

```suggestion
		// Per convention, all blocks in the sealing segment must be finalized. Therefore, a QC must
		// exist for the `highest` block in the sealing segment. The QC for `highest` should be
		// contained in the `root` Snapshot and returned by `root.QuorumCertificate()`. Otherwise,
		// the Snapshot is incomplete, because consensus nodes require this QC. To reduce the chance of
		// accidental misconfiguration undermining consensus liveness, we do the following sanity checks:
		//  * `rootQC` should not be nil
		//  * `rootQC` should be for `highest` block, i.e. its view and
		rootQC, err := root.QuorumCertificate()
		if err != nil {
			return fmt.Errorf("could not get root QC: %w", err)
		}
		if rootQC == nil {
			return fmt.Errorf("QC for highest (finalized) block in sealing segment cannot be nil")
		}
		if rootQC.View != highest.Header.View {
			return fmt.Errorf("root QC's view %d does not match the highest block in sealing segment (view %d)", rootQC.View, highest.Header.View)
		}
		if rootQC.BlockID != highest.Header.ID() {
			return fmt.Errorf("root QC is for block %v, which does not match the highest block %v in sealing segment", rootQC.BlockID, highest.Header.ID())
		}	
```

state/protocol/badger/state.go
```diff
@@ -236,14 +237,30 @@ func (state *State) bootstrapStatePointers(root protocol.Snapshot) func(*badger.
...
+
+		livenessData := &hotstuff.LivenessData{
+			CurrentView: highest.Header.View + 1,
+			LastViewTC:  highest.Header.LastViewTC,
```
AlexHentschel, line 252: I don't understand why this would be a correct initialization. Reasoning:
* Let `V` denote the view of the `highest` block in the sealing segment (i.e. `V := highest.Header.View`)
* If `highest.Header.LastViewTC` is set, it represents the TC for the view `V-1` 
* We specify that a consensus node should start at round `V+1`: https://github.com/onflow/flow-go/blob/af9fc5c3cdca409ac82457bf4237f250b46f96d6/state/protocol/badger/state.go#L251 So a valid TC for the "prior round" should have view `V`

But from my perspective, we can simply set `LastViewTC` to nil. Based on my previous comment, `rootQC`  should be for `highest` block. As the replica knows a QC for view `V`, it can enter view `V+1` according to consensus rules. No TC is required. 
```suggestion
```

consensus/participant.go
```diff
@@ -54,16 +54,15 @@ func NewParticipant(
 		option(&cfg)
 	}
 
-	// get the last view we started
-	started, err := modules.Persist.GetStarted()
+	// get the last stored safety data
+	safetyData, err := modules.Persist.GetSafetyData()
```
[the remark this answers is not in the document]
durkmurder: I guess so, I have implemented it this way to reflect our previous design, that's exactly the same way we have approached it. I will change.

state/protocol/badger/state.go
durkmurder: agree, good reasoning.

AlexHentschel:
Put up PR with some suggestions around `SafetyRules`: https://github.com/onflow/flow-go/pull/2489 (targeting your branch). Consolidated all the checks round QC and TC into a single function `validateEvidenceForEnteringView` . The remaining checks are mostly comparison with  `safetyData`

consensus/hotstuff/model/timeout.go
```diff
@@ -36,15 +36,15 @@ func (m TimeoutMode) String() string {
 type TimeoutObject struct {
 	// View is the view number which is replica is timing out
 	View uint64
-	// HighestQC is the highest QC known to the creator of this TimeoutObject
-	HighestQC *flow.QuorumCertificate
-	// LastViewTC is the timeout certificate for previous view if HighestQC.View != View - 1, else nil
+	// NewestQC is the highest QC known to the creator of this TimeoutObject
```
jordanschalm, line 39:
```suggestion
	// NewestQC is the newest QC (by view) known to the creator of this TimeoutObject
```

jordanschalm (commented):
Took another pass over the changed files - just a couple outstanding comments from my last review:
* [additional test cases](https://github.com/onflow/flow-go/pull/2444/files#r878282492)
* documenting dependency between Persister and protocol state bootstrapping

consensus/hotstuff/persister/persister.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 // Persister can persist relevant information for hotstuff.
+// Persister depends on protocol.State root snapshot bootstrapping to retrieve chainID.
```
jordanschalm, line 12:
```suggestion
// Persister depends on protocol.State root snapshot bootstrapping to set initial values for
// SafetyData and LivenessData. These values must be initialized before first use of Persister.
```

The initializing of the values was more what I meant.

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
-	require.False(t, model.IsNoVoteError(err))
+	s.signer.On("CreateTimeout", view, newestQC, (*flow.TimeoutCertificate)(nil)).Return(expectedTimeout, nil).Once()
+	timeout, err := s.safety.ProduceTimeout(view, newestQC, nil)
+	require.Nil(s.T(), timeout)
+	require.ErrorIs(s.T(), err, exception)
 }
 
-func testVotingWhileNonCommitteeMember(t *testing.T) {
-	blockView, curView, lastVotedView, isBlockSafe, isCommitteeMember := uint64(3), uint64(3), uint64(2), true, false
+// TestProduceTimeout_AfterVote tests a case where we first produce a timeout and then try to vote
```
jordanschalm, line 493: This doc seems to be the opposite of the implementation (maybe got mixed with the doc for the other suggested test case)

durkmurder: Thanks, I've messed up when was reordering and writing godoc.

jordanschalm (approved):

consensus/hotstuff/safetyrules/safety_rules_test.go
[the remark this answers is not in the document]
zhangchiqing: The problem with `errors.Is` is that once an error is unwrap the error, but `errors.As` is able to.  See example
https://go.dev/play/p/95CRSMECAep

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
-	blockView, curView, lastVotedView, isBlockSafe, isCommitteeMember := uint64(3), uint64(4), uint64(2), true, true
+// TestProduceVote_InvalidProposerIdentity tests that no vote is created if there was an exception retrieving proposer identity
+func (s *SafetyRulesTestSuite) TestProduceVote_InvalidProposerIdentity() {
+	*s.committee = mocks.DynamicCommittee{}
+	exception := errors.New("invalid-signer-identity")
+	s.committee.On("IdentityByBlock", s.proposal.Block.BlockID, s.proposal.Block.ProposerID).Return(nil, exception).Once()
+
+	vote, err := s.safety.ProduceVote(s.proposal, s.proposal.Block.View)
+	require.Nil(s.T(), vote)
+	require.ErrorIs(s.T(), err, exception)
```
AlexHentschel, line 219: If `SafetyRules` had a bug and erroneously wrapped the exception into a `NotVoteError` this test would still pass. I think it is important to test this here:
```suggestion
	require.ErrorIs(s.T(), err, exception)
	require.False(s.T(), model.IsNoVoteError(err))
```

Would suggest to include a clarifying comment in the test's goDoc:
> We are specifically testing that unexpected errors are handled correctly, i.e. 
> that SafetyRules does _not_ erroneously wrap unexpected exceptions into the expected NoVoteError.

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+	s.persister.AssertNotCalled(s.T(), "PutSafetyData")
+}
 
-	// create voter
-	block, _, voter := createVoter(blockView, lastVotedView, isBlockSafe, isCommitteeMember)
+// TestProduceVote_NodeEjected tests that no vote is created if voter is ejected
```
AlexHentschel, line 223: I think the more prominent case here is that the voter has zero weight. This is different than being ejected. An ejected node cannot participate at all anymore and is (eventually) cut out from all communication. In comparison, a consensus node with zero weight can still participate in the communication. It just can't vote anymore. Nodes have zero weight in the grace periods around the epochs where they are authorized to participate. 

```suggestion
// TestProduceVote_NodeNotAuthorizedToVote tests that no vote is created if the voter is not authorized to vote. 
// Nodes have zero weight in the grace periods around the epochs where they are authorized to participate. 
// We don't want zero-weight nodes to vote in the first place, to avoid unnecessary traffic. 
// Note: this also covers ejected nodes. In both cases, the committee will return an `InvalidSignerError`.
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+func (s *SafetyRulesTestSuite) TestProduceVote_InvalidVoterIdentity() {
+	*s.committee = mocks.DynamicCommittee{}
+	s.committee.On("Self").Return(s.ourIdentity.NodeID)
+	exception := errors.New("invalid-signer-identity")
+	s.committee.On("IdentityByBlock", s.proposal.Block.BlockID, s.proposal.Block.ProposerID).Return(s.proposerIdentity, nil).Maybe()
+	s.committee.On("IdentityByBlock", s.proposal.Block.BlockID, s.ourIdentity.NodeID).Return(nil, exception).Once()
+
+	vote, err := s.safety.ProduceVote(s.proposal, s.proposal.Block.View)
+	require.Nil(s.T(), vote)
+	require.ErrorIs(s.T(), err, exception)
```
AlexHentschel, line 246: * please test that the exception is not erroneously wrapped into a `NotVoteError` 
```suggestion
	require.ErrorIs(s.T(), err, exception)
	require.False(s.T(), model.IsNoVoteError(err))
```

Would suggest to include a clarifying comment in the test's goDoc:
> We are specifically testing that unexpected errors are handled correctly, i.e. 
> that SafetyRules does _not_ erroneously wrap unexpected exceptions into the expected NoVoteError.

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
-func testEqualLastVotedView(t *testing.T) {
-	// curView == lastVotedView
-	blockView, curView, lastVotedView, isBlockSafe, isCommitteeMember := uint64(3), uint64(3), uint64(3), true, true
+// TestProduceVote_CreateVoteException tests that no vote is created if vote creation raised an exception
+func (s *SafetyRulesTestSuite) TestProduceVote_CreateVoteException() {
+	exception := errors.New("create-vote-exception")
+	s.signer.On("CreateVote", s.proposal.Block).Return(nil, exception).Once()
+	vote, err := s.safety.ProduceVote(s.proposal, s.proposal.Block.View)
+	require.Nil(s.T(), vote)
+	require.ErrorIs(s.T(), err, exception)
```
AlexHentschel, line 256: same:
```suggestion
	require.ErrorIs(s.T(), err, exception)
	require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnUnsafeProposal() {
+	s.Run("invalid-block-view", func() {
+		// create block with block.View == block.QC.View
+		proposal := helper.MakeProposal(
+			helper.WithBlock(
+				helper.MakeBlock(
+					helper.WithParentBlock(s.bootstrapBlock),
+					helper.WithBlockView(s.bootstrapBlock.View))))
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
```
AlexHentschel, line 282: I think we are expecting an exception here, and not a `NotVoteError` (?)
```suggestion
		require.Error(s.T(), err)
		require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+	vote, err := s.safety.ProduceVote(s.proposal, s.proposal.Block.View)
+	require.Nil(s.T(), vote)
+	require.ErrorIs(s.T(), err, exception)
 }
 
-func testBelowLastVotedView(t *testing.T) {
-	// curView < lastVotedView
-	blockView, curView, lastVotedView, isBlockSafe, isCommitteeMember := uint64(3), uint64(3), uint64(4), true, true
+// TestProduceVote_VotingOnUnsafeProposal tests different scenarios where we try to vote on unsafe blocks
+func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnUnsafeProposal() {
```
AlexHentschel, line 273: ## ❓ (potentially missing test?)

Do we have a test that verifies the following rule:
* If a QC for the previous view is included in a block, the TC _must_ be nil? Otherwise, the block violates protocol rules.
* violating this rule should lead to an exception, because we expect the blocks to be already verified. 

Essentially rule 2 and 3 [here](https://github.com/onflow/flow-go/blob/3f1674597fdb15a2822f42db3e0101e2e8e136b4/consensus/hotstuff/safetyrules/safety_rules.go#L205-L216): 
> Happy Path [...] to prevent resource-exhaustion attacks, we require that no TC is included as part of the proof. 
> Following the Happy Path has priority over following the Recovery Path

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+	})
+	s.Run("no-last-view-tc", func() {
+		// create block where Block.View != Block.QC.View+1 and LastViewTC = nil
+		proposal := helper.MakeProposal(
+			helper.WithBlock(
+				helper.MakeBlock(
+					helper.WithParentBlock(s.bootstrapBlock),
+					helper.WithBlockView(s.bootstrapBlock.View+2))))
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
```
AlexHentschel, line 306: Exceptions should not be a `NoVoteError`:
```suggestion
		require.Error(s.T(), err)
		require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+		proposal := helper.MakeProposal(
+			helper.WithBlock(
+				helper.MakeBlock(
+					helper.WithParentBlock(s.bootstrapBlock),
+					helper.WithBlockView(s.bootstrapBlock.View+2))),
+			helper.WithLastViewTC(
+				helper.MakeTC(
+					helper.WithTCView(s.bootstrapBlock.View))))
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
```
AlexHentschel, line 321: Exceptions should not be a `NoVoteError`:
```suggestion
		require.Error(s.T(), err)
		require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+			helper.WithBlock(
+				helper.MakeBlock(
+					helper.WithParentBlock(s.bootstrapBlock),
+					helper.WithBlockView(s.bootstrapBlock.View+2))),
+			helper.WithLastViewTC(
+				helper.MakeTC(
+					helper.WithTCView(s.bootstrapBlock.View+1),
+					helper.WithTCHighestQC(TONewestQC))))
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
```
AlexHentschel, line 340: Exceptions should not be a `NoVoteError`:
```suggestion
		require.Error(s.T(), err)
		require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+	s.persister.On("PutSafetyData", expectedSafetyData).Return(nil).Once()
+
+	// creating new timeout should invalidate cache
+	otherTimeout, err = s.safety.ProduceTimeout(view+1, newestQC, lastViewTC)
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), otherTimeout)
+
+	// creating timeout for previous view(that was already cached) should result in error
+	timeout, err = s.safety.ProduceTimeout(view, newestQC, nil)
+	require.Error(s.T(), err)
```
AlexHentschel, line 427: Exceptions should not be a `NoVoteError`:
```suggestion
	require.Error(s.T(), err)
	require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+	require.Nil(s.T(), timeout)
```
AlexHentschel, line 428: To me, it seems this test would better fit into the [`TestProduceTimeout_NotSafeToTimeout`](https://github.com/onflow/flow-go/blob/3f1674597fdb15a2822f42db3e0101e2e8e136b4/consensus/hotstuff/safetyrules/safety_rules_test.go#L432) below. 🤷

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
 }
 
-func testVotingAgain(t *testing.T) {
-	blockView, curView, lastVotedView, isBlockSafe, isCommitteeMember := uint64(3), uint64(3), uint64(2), true, true
+// TestProduceTimeout_NotSafeToTimeout tests that we don't produce a timeout when it's not safe
+func (s *SafetyRulesTestSuite) TestProduceTimeout_NotSafeToTimeout() {
```
AlexHentschel, line 432: I think we expect all these cases to lead to exceptions (?)

```suggestion
// We expect that the EventHandler to feed only request timeouts for the current view, providing valid set of inputs.
// Hence, the cases tested here would be symptoms of an internal bugs, and therefore should not result in an NoVoteError.
func (s *SafetyRulesTestSuite) TestProduceTimeout_NotSafeToTimeout() {
```
Furthermore, I think we should add the assertion everywhere where we have `require.Error(s.T(), err)` in this test:
```golang
require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -1,165 +1,526 @@
...
+// TestProduceTimeout_CreateTimeoutException tests that no timeout is created if timeout creation raised an exception
+func (s *SafetyRulesTestSuite) TestProduceTimeout_CreateTimeoutException() {
+	view := s.proposal.Block.View
+	newestQC := helper.MakeQC(helper.WithQCView(view - 1))
+
+	exception := errors.New("create-vote-exception")
+	s.signer.On("CreateTimeout", view, newestQC, (*flow.TimeoutCertificate)(nil)).Return(nil, exception).Once()
+	vote, err := s.safety.ProduceTimeout(view, newestQC, nil)
+	require.Nil(s.T(), vote)
+	require.ErrorIs(s.T(), err, exception)
```
AlexHentschel, line 472:
```suggestion
	require.ErrorIs(s.T(), err, exception)
	require.False(s.T(), model.IsNoVoteError(err))
```

consensus/hotstuff/timeoutcollector/aggregation_test.go
```diff
@@ -171,15 +171,15 @@ func TestTimeoutSignatureAggregator_VerifyAndAdd(t *testing.T) {
...
-				weight, err = aggregator.VerifyAndAdd(ids[i].NodeID, sigs[(i+1)%signersNum], highQCViews[(i+1)%signersNum]) // different signature for same index
+				weight, err = aggregator.VerifyAndAdd(ids[i].NodeID, sigs[(i+1)%signersNum], newestQCViews[(i+1)%signersNum]) // different signature for same index
 				// weight should not change
 				require.Equal(t, expectedWeight, weight)
 				require.True(t, model.IsDuplicatedSignerError(err))
-				weight, err = aggregator.VerifyAndAdd(ids[(i+1)%signersNum].NodeID, sigs[(i+1)%signersNum], highQCViews[(i+1)%signersNum]) // different signature for same index
+				weight, err = aggregator.VerifyAndAdd(ids[(i+1)%signersNum].NodeID, sigs[(i+1)%signersNum], newestQCViews[(i+1)%signersNum]) // different signature for same index
 				// weight should not change
 				require.Equal(t, expectedWeight, weight)
 				require.ErrorAs(t, err, &model.ErrInvalidSignature)
```
AlexHentschel, line 185: ## ⚠️ 

based on Jordan's earlier comment, I think it would be better to write here
```suggestion
				require.ErrorIs(t, err, model.ErrInvalidSignature)
```

AlexHentschel (approved):
Thanks for the detailed tests. This is great. Only a few minor comments. 
Feel free to merge.

consensus/hotstuff/safetyrules/safety_rules_test.go
durkmurder: Actually we already have it in [TestProduceTimeout_NotSafeToTimeout](https://github.com/onflow/flow-go/blob/3f1674597fdb15a2822f42db3e0101e2e8e136b4/consensus/hotstuff/safetyrules/safety_rules_test.go#L432) :facepalm:

consensus/hotstuff/safetyrules/safety_rules_test.go
durkmurder: This test case is included in proposal [validation](https://github.com/onflow/flow-go/pull/2456/files#diff-c85759ef2971fa1acf44671dfb9dbf30e7f3485d3920d39686c2fdff08527518R403) only. I will add same test case here.

durkmurder:
bors merge

consensus/hotstuff/safetyrules/safety_rules_test.go
AlexHentschel: We are discussing the convention:
> If a QC for the previous view is included in a block, the TC must be nil

From my perspective, there are two angles to this discussion:
1. Will the proofs still hold, if `SafetyRules` does not enforce this convention?
2. Is our implementation of `SafetyRules` enforcing this convention. 

How I understand your argument, you are focusing on 1. Here, the answer is "yes, SafetyRules does not need to enforce this convention".  Hence, one could conclude 

In comparison, I am more focusing on 2. I think our tests for safety rules should be tailored towards our specific implementation of `SafetyRules`. 
* At the moment, our `SafetyRules` documentation states that it enforces this convention: https://github.com/onflow/flow-go/blob/b6f2aa302e3709d902fa1b61df18f250c6b3b218/consensus/hotstuff/safetyrules/safety_rules.go#L206-L209 Rules 1. and 2. directly imply that a block that violates our convention should not pass `validateEvidenceForEnteringView`
* Also the implementation has the specific code to reject blocks that violate the condition: https://github.com/onflow/flow-go/blob/b6f2aa302e3709d902fa1b61df18f250c6b3b218/consensus/hotstuff/safetyrules/safety_rules.go#L237-L239

In summary, I conclude that `SafetyRules` aims to enforce the convention.  Hence, I think the tests should verify the expected behaviour.

AlexHentschel: Btw, the convention is not only w.r.t. to blocks but also applies to timeouts. I would prefer to have tests, where we verify that violating this convention results in an error (not only for blocks but also for timeouts). 
Having said this, this is a soft preference. I am also ok with your position :-)

PR #2451, opened 2022-05-19
Add missing weights for cadence memory kinds

SupunS:
The main purpose of this PR was to add memory weight for the encoded-slab memory kind added in: https://github.com/onflow/cadence/pull/1662

In the process, I also found that there are some missing/zero weights in the weighted meter. So updated them with proper values, and added a test to ensure all weights are properly defined.

Also, sync with the changes added in https://github.com/onflow/cadence/pull/1667


fvm/meter/weighted/meter_test.go
```diff
@@ -344,3 +346,27 @@ func TestWeightedComputationMetering(t *testing.T) {
...
+func TestMemoryWeights(t *testing.T) {
+	for kind := common.MemoryKindUnknown + 1; kind < common.MemoryKindLast; kind++ {
+		weight, ok := weighted.DefaultMemoryWeights[kind]
+		assert.True(t, ok, fmt.Sprintf("missing weight for memory kind '%s'", kind.String()))
+
+		// CadenceDictionarySize memory kind currently have '0' weight.
+		// TODO: Add a proper weight
+		if kind == common.MemoryKindCadenceDictionarySize {
+			continue
+		}
```
SupunS, line 359: @robert-e-davidson3 The `MemoryKindCadenceDictionarySize` have a zero weight. Do you know whether it was intentional? Or need to add a proper weight?

robert-e-davidson3: @SupunS That's intentional. The CadenceDictionary is a Go array and each of its elements is already metered as `MemoryKindCadenceKeyValuePair`.

SupunS: Maybe we can remove this memory kind then?

robert-e-davidson3: We totally can. I included it because I was worried someone would wonder why it was missing.

I included a comment where I set the weight but looking back it wasn't super clear: https://github.com/onflow/flow-go/pull/2451/files#diff-915a90215e5e6d9bc62a61a96bdc993b9149273234e6665f4396622499e85823R97

fvm/meter/weighted/meter_test.go
```diff
@@ -344,3 +346,20 @@ func TestWeightedComputationMetering(t *testing.T) {
 		require.True(t, errors.IsMemoryLimitExceededError(err))
 	})
 }
+
+func TestMemoryWeights(t *testing.T) {
```
turbolent, line 350: Great idea to add a sanity check 👌

turbolent (approved):

robert-e-davidson3 (approved):

PR #2453, opened 2022-05-19
flaky tests fix test async uploader - Test_AsyncUploader/stopping_component_stops_retrying

gomisha:
This PR fixes the flaky behavior of `Test_AsyncUploader/stopping_component_stops_retrying`

[Grafana dashboard](https://dapperlabs.grafana.net/d/toErpbynz/single-test-metrics?orgId=1&var-package=github.com%2Fonflow%2Fflow-go%2Fengine%2Fexecution%2Fcomputation%2Fcomputer%2Fuploader&var-dataset_name=dapperlabs-data.production_src_flow_test_metrics&var-skipped_tests_table=skipped_tests&var-test_results_table=test_results&var-query_limit=100000&var-test=Test_AsyncUploader%2Fstopping_component_stops_retrying) (Flaky Test Monitor) showing that it's been failing consistently.

- there are many logging statements added - this is for debugging the sequence of events between goroutines and I chose to leave them in for future debugging
- left the conditional skip in place because I want it to run with the flaky test monitor for a while to make sure it's passing over a long period of time before removing it from quarantine

Tested and passing with:
`go test -count=10`
`go test -count=100`

image


engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -95,35 +99,75 @@ func Test_AsyncUploader(t *testing.T) {
...
+				wgUploadStarted.Done()
+
+				t.Log("DummyUpload func() waiting for component shutdown to start grID:", string(bytes.Fields(debug.Stack())[1]))
+				wgShutdownStarted.Wait()
+				t.Log("DummyUploader func() component shutdown started, about to return error grID:", string(bytes.Fields(debug.Stack())[1]))
+
+				// force an upload error to test that upload is never retried (because component is shut down)
+				// normally, we would see retry mechanism kick in and the callCount would be > 1
+				// but since component has started shutting down, we expect callCount to be 1
+				if callCount < 5 {
```
m4ksio, line 141: Why is this less than 5? Shouldn't this be called only once really?

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -95,35 +99,75 @@ func Test_AsyncUploader(t *testing.T) {
...
+		// this wait group ensures that async uploader won't send an error (to test if retry will kick in) until
+		// the component has initiated shutting down (which should stop retry from working)
+		wgShutdownStarted := sync.WaitGroup{}
+		wgShutdownStarted.Add(1)
+		t.Log("added 1 to wait group grID:", string(bytes.Fields(debug.Stack())[1]))
 
 		uploader := &DummyUploader{
 			f: func() error {
-				defer func() {
+				t.Log("DummyUploader func() - about to call wgUploadStarted.Done() grID:", string(bytes.Fields(debug.Stack())[1]))
```
m4ksio, line 130: I'm not sure we should use `debug.Stack()` on normal operation. I get it helps, but seems like potential pitfall.

engine/execution/computation/computer/uploader/uploader_test.go
```diff
@@ -95,35 +99,75 @@ func Test_AsyncUploader(t *testing.T) {
...
+		t.Log("about to close async uploader grID:", string(bytes.Fields(debug.Stack())[1]))
+
+		// wait until upload has started before shutting down the component
+		wgUploadStarted.Wait()
+
+		// stop component and check that it's fully stopped
+		t.Log("about to initiate shutdown grID: ", string(bytes.Fields(debug.Stack())[1]))
 		c := async.Done()
-		wg.Done()
-		<-c
```
m4ksio, line 125: Don't we need to read from a channel anymore?

m4ksio (approved):

engine/execution/computation/computer/uploader/uploader_test.go
Kay-Zee: Yeah, the comment also doesn't explain why it's 5, from reading the comment, i would assume it'd be `callCount <= 1`

Kay-Zee (approved):

engine/execution/computation/computer/uploader/uploader_test.go
gomisha: It's strictly for debugging purposes to log the goroutine id - it was very helpful in debugging this test and I think it's something we should use for tests with multiple goroutines. What's the potential pitfall?

engine/execution/computation/computer/uploader/uploader_test.go
gomisha: yes, this SHOULD be called only once - but we want the test to TRY and call it more than once to prove that it was only called once. If we changed it to `callCount < 1` that wouldn't prove that the test tried to call it more than once and wouldn't prove that stopping the component stopped the retry mechanism.

engine/execution/computation/computer/uploader/uploader_test.go
gomisha: this is done with the helper method on 168: `unittest2.RequireCloseBefore(t, c, 1*time.Second, "async uploader not closed in time")` which has a timeout in case reading form the channel results in a deadlock

engine/execution/computation/computer/uploader/uploader_test.go
gomisha: If there's a better way to log the goroutine ID, let's use that, but this is the most convenient way I found.

engine/execution/computation/computer/uploader/uploader_test.go
gomisha: added clarification comments to explain this
https://github.com/onflow/flow-go/pull/2453/commits/a6271a474bd1293fe7417f32b40ad2377de12b36

PR #2456, opened 2022-05-20
[Consensus] Active Pacemaker: entities validation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6192

### Context

This PR implements new validation rules for QC, proposal and TC.
Key changes:
- `hotstuff.Validator` has new API, now it performs validation of entities without requiring extra data(parent block)
- `hotstuff.Validator` performs validation based on epoch-wide information
- Added TC validation logic
- Handled TODOs required for front-loading of verification 

The most important change is reducing dependencies for validation, after this PR we will be able to validate entities as soon as we receive it from network, all we need to know is referred epoch.


consensus/hotstuff/validator/validator.go
```diff
@@ -118,63 +179,70 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
-	if vote.View != block.View {
-		return nil, newInvalidVoteError(vote, fmt.Errorf("vote's view %d is inconsistent with referenced block (view %d)", vote.View, block.View))
-	}
-
-	// TODO: handle model.ErrViewForUnknownEpoch when front-loading verification
-	voter, err := v.committee.IdentityByEpoch(block.View, vote.SignerID)
+// During normal operations, the following error returns are expected:
+//  * model.InvalidVoteError for invalid votes
+//  * model.ErrViewForUnknownEpoch if the vote refers unknown epoch
+// Any other error should be threaded as exception
```
jordanschalm, line 234:
```suggestion
// Any other error should be treated as exception
```

consensus/hotstuff/validator/validator.go
```diff
@@ -118,63 +179,70 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
+func (v *Validator) ValidateVote(vote *model.Vote) (*flow.Identity, error) {
+	voter, err := v.committee.IdentityByEpoch(vote.View, vote.SignerID)
 	if model.IsInvalidSignerError(err) {
 		return nil, newInvalidVoteError(vote, err)
 	}
 	if err != nil {
-		return nil, fmt.Errorf("error retrieving voter Identity at block %x: %w", block.BlockID, err)
+		return nil, fmt.Errorf("error retrieving voter Identity at view %x: %w", vote.View, err)
```
jordanschalm, line 241:
```suggestion
		return nil, fmt.Errorf("error retrieving voter Identity at view %d: %w", vote.View, err)
```

jordanschalm (commented):
First round of comments - haven't completed review yet.

consensus/hotstuff/validator/validator.go
```diff
@@ -118,63 +179,70 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
 		return newInvalidBlockError(block, fmt.Errorf("proposer %s is not leader (%s) for view %d", block.ProposerID, leader, block.View))
 	}
 
-	// check the proposer is a valid committee member at the most recent state on this fork
-	// TODO: move this check into SafetyRules when front-loading the verification
-	_, err = v.committee.IdentityByBlock(block.BlockID, block.ProposerID)
-	if model.IsInvalidSignerError(err) {
-		return newInvalidBlockError(block, fmt.Errorf("proposer %x is leader but is not a valid committee member", block.ProposerID))
```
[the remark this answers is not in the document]
durkmurder: https://github.com/onflow/flow-go/pull/2444/files#diff-ad0bb7646fd0dfa4dd9a4bc2e8d2ee50bdc755f9811449b612d3e875c855af86R79

consensus/hotstuff/validator/validator_test.go
```diff
@@ -522,12 +627,11 @@ func (qs *QCSuite) TestQCSignatureInvalid() {
...
 	// change the verifier to fail the QC signature
 	*qs.verifier = mocks.Verifier{}
-	qs.verifier.On("VerifyQC", qs.signers, qs.qc.SigData, qs.block).Return(
-		fmt.Errorf("%w", model.NewInvalidFormatErrorf("invalid sigType")))
+	qs.verifier.On("VerifyQC", qs.signers, qs.qc.SigData, qs.qc.View, qs.qc.BlockID).Return(model.NewInvalidFormatErrorf("invalid sigType"))
 
-	// the QC be considered as invalid
-	err := qs.validator.ValidateQC(qs.qc, qs.block)
-	assert.True(qs.T(), model.IsInvalidBlockError(err), "if the signature has an invalid format, an ErrorInvalidBlock error should be raised")
+	// the QC should no longer be validation
```
jordanschalm, line 632:
```suggestion
	// the QC should no longer pass validation
```

consensus/hotstuff/validator/validator_test.go
```diff
@@ -537,11 +641,173 @@ func (qs *QCSuite) TestQCSignatureInvalidFormat() {
...
+
+	// signers are a qualified majority at 7
+	s.signers = s.participants[:7]
+
+	var err error
+	s.indices, err = signature.EncodeSignersToIndices(s.participants.NodeIDs(), s.signers.NodeIDs())
+	require.NoError(s.T(), err)
+
+	rand.Seed(time.Now().UnixNano())
+	view := uint64(int(rand.Uint32()) + len(s.participants))
```
jordanschalm, line 685:
```suggestion
	view := uint64(int(rand.Uint32()))
```
Any reason to add `len(participants)` to view? It's not a problem really, just a bit confusing.

consensus/hotstuff/validator/validator_test.go
```diff
@@ -537,11 +641,173 @@ func (qs *QCSuite) TestQCSignatureInvalidFormat() {
...
+
+	// check the default happy case passes
+	err := s.validator.ValidateTC(s.tc)
+	assert.NoError(s.T(), err, "a valid TC should be accepted")
+}
+
+// TestTCEmptySigners tests if correct error is returned when signers are empty
+func (s *TCSuite) TestTCEmptySigners() {
+	s.verifier.On("VerifyTC", mock.Anything, []byte(s.tc.SigData), s.tc.View, s.tc.TOHighQCViews).Return(
+		fmt.Errorf("%w", model.NewInsufficientSignaturesErrorf("")))
```
jordanschalm, line 735: ⚠️  `VerifyTC` [does not return `InsufficientSignaturesError`](https://github.com/onflow/flow-go/blob/32eec3b47a466dc14712fa49c19f0ab83966f529/consensus/hotstuff/verification/staking_verifier.go#L120-L145) from what I can see. 

In the case of empty signers, [we will return `InvalidFormatError`](https://github.com/onflow/flow-go/blob/32eec3b47a466dc14712fa49c19f0ab83966f529/consensus/hotstuff/verification/staking_verifier.go#L122), which is considered invalid TC rather than internal error (opposite of what this test states). 

I suggest we modify `ValidateTC` to return consistent errors with `ValidateQC` for this error (eg. both should return `InsufficientSignaturesError` under the same circumstances). We could consider moving this test case to `TestTCInvalidSignature` where we test the other cases of `Verifier` returning various errors. 

Also noticed: 
* For `VerifyQC`, [`CombinedVerifier`](https://github.com/onflow/flow-go/blob/32eec3b47a466dc14712fa49c19f0ab83966f529/consensus/hotstuff/verification/combined_verifier_v2.go#L123) returns `InsufficientSignaturesError` but [`StakingVerifier`](https://github.com/onflow/flow-go/blob/32eec3b47a466dc14712fa49c19f0ab83966f529/consensus/hotstuff/verification/staking_verifier.go#L68) returns `InvalidFormatError`

durkmurder: fixed

consensus/hotstuff/verification/staking_verifier.go
```diff
@@ -92,7 +92,54 @@ func (v *StakingVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, bl
...
+// It is the responsibility of the calling code to ensure
+// that all `voters` are authorized, without duplicates. Return values:
+//  - nil if `sigData` is cryptographically valid
+//  - model.ErrInvalidFormat if `sigData` has an incompatible format
+//  - model.ErrInvalidSignature if a signature is invalid
+//  - unexpected errors should be treated as symptoms of bugs or uncovered
+//	  edge cases in the logic (i.e. as fatal)
+func verifyTC(voters flow.IdentityList, sigData []byte, view uint64, highQCViews []uint64, hasher hash.Hasher) error {
+	if len(voters) == 0 {
+		return model.NewInvalidFormatErrorf("empty list of signers")
```
jordanschalm:
```suggestion
		return model.NewInsufficientSignaturesErrorf("empty list of signers")
```
For consistency with `VerifyQC`. Though alternatively, maybe there is no real reason to differentiate between "invalid format" and "insufficient signers", and we could consolidate on using only `InvalidFormatError` for this case across both TC/QCs.

consensus/hotstuff/verification/staking_verifier.go
```diff
@@ -92,7 +92,54 @@ func (v *StakingVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, bl
...
+//	  edge cases in the logic (i.e. as fatal)
+func (v *StakingVerifier) VerifyTC(voters flow.IdentityList, sigData []byte, view uint64, highQCViews []uint64) error {
+	return verifyTC(voters, sigData, view, highQCViews, v.stakingHasher)
+}
+
+// verifyTC checks cryptographic validity of the TC's `sigData` for the given view.
+// It is the responsibility of the calling code to ensure
+// that all `voters` are authorized, without duplicates. Return values:
+//  - nil if `sigData` is cryptographically valid
+//  - model.ErrInvalidFormat if `sigData` has an incompatible format
```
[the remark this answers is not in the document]
jordanschalm: Agree about `IsInvalidInputsError`, was thinking the crypto method would return that for invalid `sigData` but seems not to be the case.

I think the specific code change suggestion still applies though. In particular, we do not return `ErrInvalidFormat` if `sigData` has an incompatible format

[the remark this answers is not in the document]
durkmurder: what do you mean by incompatible format? Incompatible format can happen only in case of `CombinedVerifier` where we have two sigs instead of one.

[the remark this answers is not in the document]
jordanschalm: What I mean is the documentation and implementation don't match. 

The function documentation says:
> Return values:
> - model.ErrInvalidFormat if `sigData` has an incompatible format

But in the implementation, we do:
```go
if len(voters) == 0 {
	return model.NewInvalidFormatErrorf("empty list of signers")
}
if len(voters) != len(highQCViews) {
	return model.NewInvalidFormatErrorf("signers and highQCViews mismatch")
}
```

My suggestion was to change the documentation to:
> - model.ErrInvalidFormat if `voters`/`highQCViews` have differing lengths, or are empty

because that is what we do in the implementation.

consensus/hotstuff/validator/validator.go
```diff
@@ -12,89 +12,143 @@ import (
...
+
+	// 1. Check if there is super-majority of votes
+	allParticipants, err := v.committee.IdentitiesByEpoch(tc.View)
+	if err != nil {
+		return fmt.Errorf("could not get consensus participants at view %d: %w", tc.View, err)
+	}
+
+	signers, err := signature.DecodeSignerIndicesToIdentities(allParticipants, tc.SignerIndices)
+	if err != nil {
+		if signature.IsDecodeSignerIndicesError(err) {
```
AlexHentschel, line 67: I specifically asked for this method to be removed, because I consider it an intrinsically unsafe pattern (see reasoning in description of [this PR](https://github.com/onflow/flow-go/pull/2517)). 
In master, `IsDecodeSignerIndicesError` had been replaced by `InvalidSignerIndicesError`.

consensus/hotstuff/validator/validator.go
```diff
@@ -12,89 +12,143 @@ import (
...
+			return newInvalidTCError(tc, fmt.Errorf("TC contains invalid signature(s): %w", err))
+		default:
+			return fmt.Errorf("cannot verify tc's aggregated signature (tc.View: %d): %w", tc.View, err)
+		}
 	}
+	return nil
+}
 
+// ValidateQC validates the QC
+// qc - the qc to be validated
```
AlexHentschel, line 105:
```suggestion
// ValidateQC validates the Quorum Certificate `qc`.
```

consensus/hotstuff/validator/validator.go
```diff
@@ -12,89 +12,143 @@ import (
...
 	if err != nil {
-		// Considerations about other errors that `VerifyQC` could return:
-		//  * model.InvalidSignerError: for the time being, we assume that _every_ HotStuff participant
-		//    is also a member of the random beacon committee. Consequently, `InvalidSignerError` should
-		//    not occur atm.
-		//    TODO: if the random beacon committee is a strict subset of the HotStuff committee,
-		//          we expect `model.InvalidSignerError` here during normal operations.
-		// * model.InsufficientSignaturesError: we previously checked the total weight of all signers
-		//   meets the supermajority threshold, which is a _positive_ number. Hence, there must be at
-		//   least one signer. Hence, receiving this error would be a symptom of a fatal internal bug.
```
AlexHentschel, line 92: I think it might be useful to keep this comment.

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +186,72 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
-		return newInvalidBlockError(block, fmt.Errorf("proposer %x is leader but is not a valid committee member", block.ProposerID))
+	// The Block must contain a proof that the primary legitimately entered the respective view.
+	// Transitioning to proposal.Block.View is possible either by observing a QC or a TC for the
+	// previous round. If and only if the QC is _not_ for the previous round we require a TC for
+	// the previous view to be present.
+	lastViewSuccessful := proposal.Block.View == proposal.Block.QC.View+1
+	if !lastViewSuccessful {
+		// check if proposal is correctly structured
+		if proposal.LastViewTC == nil {
+			return newInvalidBlockError(block, fmt.Errorf("last view has ended with timeout but proposal doesn't include LastViewTC"))
```
AlexHentschel, line 197: These checks happen at the compliance engine, which has no knowledge of what the current view is:

```suggestion
			return newInvalidBlockError(block, fmt.Errorf("QC in block is not for previous view, so expecting a TC but none is included in block"))
```

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +186,72 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
+		// check if included TC is for previous view
+		if proposal.Block.View != proposal.LastViewTC.View+1 {
+			return newInvalidBlockError(block, fmt.Errorf("expected TC for view %d got %d", proposal.Block.View-1, proposal.LastViewTC.View))
+		}
+
+		// Check if proposal extends either the newest QC specified in the TC, or a newer QC
+		// in edge cases a leader may construct a TC and QC concurrently such that TC contains
+		// an older QC - in these case we still want to build on the newest QC, so this case is allowed.
+		if proposal.Block.QC.View < proposal.LastViewTC.TOHighestQC.View {
+			return newInvalidBlockError(block, fmt.Errorf("proposal's QC is lower than locked QC"))
```
AlexHentschel, line 209:
```suggestion
			return newInvalidBlockError(block, fmt.Errorf("TC in block contains a newer QC than the block itself, which is a protocol violation"))
```

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +186,72 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
+	lastViewSuccessful := proposal.Block.View == proposal.Block.QC.View+1
+	if !lastViewSuccessful {
+		// check if proposal is correctly structured
+		if proposal.LastViewTC == nil {
+			return newInvalidBlockError(block, fmt.Errorf("last view has ended with timeout but proposal doesn't include LastViewTC"))
+		}
+
+		// check if included TC is for previous view
+		if proposal.Block.View != proposal.LastViewTC.View+1 {
+			return newInvalidBlockError(block, fmt.Errorf("expected TC for view %d got %d", proposal.Block.View-1, proposal.LastViewTC.View))
```
AlexHentschel, line 202:
```suggestion
			return newInvalidBlockError(block, fmt.Errorf("QC in block is not for previous view, so expecting a TC for view %d but got TC for view %d", proposal.Block.View-1, proposal.LastViewTC.View))
```

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +186,72 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
-	if !found {
-		// Forks is _allowed_ to (but obliged to) prune blocks whose view is below the newest finalized block.
-		if qc.View >= v.forks.FinalizedView() {
-			// If the parent block is equal or above the finalized view, then Forks should have it. Otherwise, we are missing a block!
-			return model.MissingBlockError{View: qc.View, BlockID: qc.BlockID}
+	if !lastViewSuccessful {
+		// check if included TC is valid
+		err = v.ValidateTC(proposal.LastViewTC)
+		if err != nil {
+			return newInvalidBlockError(block, fmt.Errorf("proposals TC's is not valid: %w", err))
```
AlexHentschel, line 231: ### ⚠️ 

I think not every error means that the block is invalid. 
```suggestion
		if model.IsInvalidTCError(err) {
			return newInvalidBlockError(block, fmt.Errorf("proposals TC's is not valid: %w", err))
		}
		fmt.Errorf("unexpected internal error while verifying the TC included in block: %w", err)
```

AlexHentschel (commented):
first batch of comments.

consensus/hotstuff/validator/validator.go
AlexHentschel: we can clean this up in a later PR. Just wanted to highlight the change.

AlexHentschel:
I think this method: https://github.com/onflow/flow-go/blob/b6f2aa302e3709d902fa1b61df18f250c6b3b218/consensus/hotstuff/committee.go#L54-L55
could also return a `model.ErrViewForUnknownEpoch` error. I think we should add this as documentation to the API

consensus/hotstuff/validator/validator.go
```diff
@@ -12,74 +12,140 @@ import (
...
 	}
-	if qc.View != block.View { // check view
-		return newInvalidBlockError(block, fmt.Errorf("qc's View %d doesn't match referenced block's View %d", qc.View, block.View))
+
+	// Validate QC
+	err = v.ValidateQC(highestQC)
+	if err != nil {
+		if model.IsInvalidQCError(err) {
+			return newInvalidTCError(tc, fmt.Errorf("invalid QC included in TC: %w", err))
+		}
```
AlexHentschel, line 87: I think we also require special handling of `ErrViewForUnknownEpoch` here:

```suggestion
		}
				if errors.Is(err, model.ErrViewForUnknownEpoch) {
			// We require each replica to be bootstrapped with a QC pointing to a finalized block. Consensus safety rules guarantee that
			// a QC at least as new as the root QC must be contained in any TC. This is because the TC must include signatures from a
			// supermajority of replicas, including at least one honest replica, which attest to their locally highest known QC. Hence, 
			// any QC included in a TC must be the root QC or newer. Therefore, we should know the Epoch for any QC we encounter.
			// receiving a `model.ErrViewForUnknownEpoch` is conceptually impossible, i.e. a symptom of an internal bug or invalid
			// bootstrapping information. 
			return fmt.Errorf("no Epoch information availalbe for QC that was included in TC; symptom of internal bug or invalid bootstrapping information: %s", err.Error())
		}
```

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +198,75 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
+
+	// Check signatures, keep the most expensive the last to check
+
+	// check if included QC is valid
+	err = v.ValidateQC(qc)
 	if err != nil {
-		return fmt.Errorf("unexpected error checking proposer standing for block %x: %w", block.BlockID, err)
+		if model.IsInvalidQCError(err) {
+			return newInvalidBlockError(block, fmt.Errorf("invalid qc included: %w", err))
+		}
```
AlexHentschel, line 235: same argument here:
```suggestion
		}
				if errors.Is(err, model.ErrViewForUnknownEpoch) {
			// We require each replica to be bootstrapped with a QC pointing to a finalized block. Consensus safety rules guarantee that
			// a QC at least as new as the root QC must be contained in any TC. This is because the TC must include signatures from a
			// supermajority of replicas, including at least one honest replica, which attest to their locally highest known QC. Hence, 
			// any QC included in a TC must be the root QC or newer. Therefore, we should know the Epoch for any QC we encounter.
			// receiving a `model.ErrViewForUnknownEpoch` is conceptually impossible, i.e. a symptom of an internal bug or invalid
			// bootstrapping information. 
			return fmt.Errorf("no Epoch information availalbe for QC that was included in TC; symptom of internal bug or invalid bootstrapping information: %s", err.Error())
		}
```

consensus/hotstuff/validator/validator.go
```diff
@@ -129,63 +198,75 @@ func (v *Validator) ValidateProposal(proposal *model.Proposal) error {
...
-		if qc.View >= v.forks.FinalizedView() {
-			// If the parent block is equal or above the finalized view, then Forks should have it. Otherwise, we are missing a block!
-			return model.MissingBlockError{View: qc.View, BlockID: qc.BlockID}
+	if !lastViewSuccessful {
+		// check if included TC is valid
+		err = v.ValidateTC(proposal.LastViewTC)
+		if err != nil {
+			if model.IsInvalidTCError(err) {
+				return newInvalidBlockError(block, fmt.Errorf("proposals TC's is not valid: %w", err))
+			}
```
AlexHentschel, line 245:
```suggestion
			}
						if errors.Is(err, model.ErrViewForUnknownEpoch) {
				// We require each replica to be bootstrapped with a QC pointing to a finalized block. Therefore, we should know the
				// Epoch for any QC.View and TC.View we encounter. Receiving a `model.ErrViewForUnknownEpoch` is conceptually impossible,
				// i.e. a symptom of an internal bug or invalid bootstrapping information.
				return fmt.Errorf("no Epoch information availalbe for QC that was included in TC; symptom of internal bug or invalid bootstrapping information: %s", err.Error())
			}
```

AlexHentschel (commented):
comments and suggestions regarding the error handling in [`Validator`](https://github.com/onflow/flow-go/blob/f03421d102a6f6934e0eb6491c85c7621b40fb29/consensus/hotstuff/validator/validator.go)

consensus/hotstuff/verifier.go
```diff
@@ -21,7 +20,7 @@ import (
 type Verifier interface {
```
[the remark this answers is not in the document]
durkmurder: Yeap, `VerifyVote` and `VerifyQC` can fail with it when querying DKG.

consensus/hotstuff/validator/validator.go
```diff
@@ -12,74 +12,140 @@ import (
...
+		// Considerations about other errors that `VerifyTC` could return:
+		// * model.InsufficientSignaturesError: we previously checked the total weight of all signers
+		//   meets the supermajority threshold, which is a _positive_ number. Hence, there must be at
+		//   least one signer. Hence, receiving this error would be a symptom of a fatal internal bug.
+		switch {
+		case model.IsInvalidFormatError(err):
+			return newInvalidTCError(tc, fmt.Errorf("TC's signature data has an invalid structure: %w", err))
+		case errors.Is(err, model.ErrInvalidSignature):
+			return newInvalidTCError(tc, fmt.Errorf("TC contains invalid signature(s): %w", err))
+		default:
```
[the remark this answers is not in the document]
durkmurder: I don't think `VerifyTC` can fail with this error

consensus/hotstuff/validator/validator.go
```diff
@@ -12,74 +12,140 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("could not get weight threshold for view %d: %w", tc.View, err)
+	}
+	if signers.TotalWeight() < threshold {
+		return newInvalidTCError(tc, fmt.Errorf("tc signers have insufficient weight of %d (required=%d)", signers.TotalWeight(), threshold))
 	}
-	if qc.View != block.View { // check view
-		return newInvalidBlockError(block, fmt.Errorf("qc's View %d doesn't match referenced block's View %d", qc.View, block.View))
+
+	// Validate QC
```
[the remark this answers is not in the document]
durkmurder: Oh, I definitely agree on this, I believe in most if the cases(if not all) we should check signature immediately and we have all means for that atm.

consensus/hotstuff/validator/validator.go
```diff
@@ -12,74 +12,140 @@ import (
...
-//  * model.InvalidBlockError if the QC is invalid
-func (v *Validator) ValidateQC(qc *flow.QuorumCertificate, block *model.Block) error {
-	if qc.BlockID != block.BlockID {
-		// Sanity check! Failing indicates a bug in the higher-level logic
-		return fmt.Errorf("qc.BlockID %s doesn't match block's ID %s", qc.BlockID, block.BlockID)
+//  * model.InvalidTCError if the TC is invalid
+//  * model.ErrViewForUnknownEpoch if the TC refers unknown epoch
+// Any other error should be treated as exception
+func (v *Validator) ValidateTC(tc *flow.TimeoutCertificate) error {
+	highestQC := tc.NewestQC
```
jordanschalm, line 38:
```suggestion
	newestQC := tc.NewestQC
```
suggestion to keep consistent terminology

consensus/hotstuff/validator/validator.go
```diff
@@ -12,74 +12,147 @@ import (
...
+//  * model.InvalidTCError if the TC is invalid
+//  * model.ErrViewForUnknownEpoch if the TC refers unknown epoch
+// Any other error should be treated as exception
+func (v *Validator) ValidateTC(tc *flow.TimeoutCertificate) error {
+	newestQC := tc.NewestQC
+
+	// The TC's view cannot be smaller than the view of the QC it contains.
+	// Note: we specifically allow for the TC to have the same view as the highest QC.
+	// This is useful as a fallback, because it allows replicas other than the designated
+	// leader to also collect votes and generate a QC.
```
jordanschalm, line 43: This explanation doesn't make sense to me.

We're saying that view `n` timed out, but the leader `L` for view `n+1` was able to collect votes for view `n`, so that `L` has:
* `QC_n` with view `n`
* `TC_n` with view `n`

If this happens, `L` may embed the `QC_n` in `TC_n`, in order to make progress using `QC_n` more quickly. But, why would `L` do this, when `L` could instead use `QC_n` directly and omit the TC altogether?

jordanschalm: Additionally, looking at [`ValidateProposal`](https://github.com/onflow/flow-go/pull/2456/files#diff-8e933e078b2e6031be4617d76cfe25046e38ac012720b757b382869e6a509aeaR218-R239), the case described here would be invalid and result in an error.

Suppose we have a `TC_n` containing a highest QC `QC_n`:

1. If we embed `QC_n` in the proposal, then `lastViewSuccessful` is true on line 218 and we return an error on line 238 because an unnecessary TC is included (check on line 236).

2. If we don't embed `QC_n` in the proposal, we must include a QC with view lower than `n` and we return an error line 234 (check on line 233).

consensus/hotstuff/validator/validator_test.go
```diff
@@ -221,6 +221,18 @@ func (ps *ProposalSuite) TestProposalQCInvalid() {
 		assert.ErrorIs(ps.T(), err, exception)
 		assert.False(ps.T(), model.IsInvalidBlockError(err))
 	})
+
+	ps.Run("verify-qc-err-view-for-unknown-epoch", func() {
+		*ps.verifier = mocks.Verifier{}
+		ps.verifier.On("VerifyQC", ps.voters, ps.block.QC.SigData, ps.parent.View, ps.parent.BlockID).Return(model.ErrViewForUnknownEpoch)
+		ps.verifier.On("VerifyVote", ps.voter, ps.vote.SigData, ps.block.View, ps.block.BlockID).Return(nil)
+
+		// check that validation fails and the failure case is recognized as an invalid block
```
jordanschalm, line 230:
```suggestion
		// check that validation fails and the failure is considered internal exception and NOT an InvalidBlock error
```

consensus/hotstuff/validator/validator_test.go
```diff
@@ -384,6 +396,35 @@ func (ps *ProposalSuite) TestProposalWithLastViewTC() {
 		err := ps.validator.ValidateProposal(proposal)
 		require.True(ps.T(), model.IsInvalidBlockError(err) && model.IsInvalidTCError(err))
 	})
+	ps.Run("verify-qc-err-view-for-unknown-epoch", func() {
+		// QC included in TC has view below QC included in proposal
```
jordanschalm, line 400:
```suggestion
```
I don't think this is relevant for this test case

jordanschalm (commented):
Took another pass. I have one question about the allowed case where a TC contains a QC for the same view. Everything else looks good to me - once we're aligned on the reasoning for this case, this is ready to merge from my side.

consensus/hotstuff/validator/validator.go
durkmurder: Hm, I am not 100% sure here. This comment was suggested by Alex and I think it's relevant for future extensions of the protocol where we might allow fallback leaders. 

@AlexHentschel

AlexHentschel: Yes, this ultimately for a performance optimization. While I don't necessarily want to implement the optimization right away, it would be nice to already implement our code such that it is compatible with this optimization. Up to now, my proposal wasn't written up, so I understand that there is a lot of confusion around it. Spent some time today for properly documenting it: https://www.notion.so/dapperlabs/Optimization-Reducing-impact-of-offline-primaries-ffa994aafcdd43ad99652e9c3b47a951

Regarding your questions @jordanschalm:
* > why would L do this, when L could instead use QC_n directly and omit the TC altogether

   I hope the notion doc clarifies this. 
* > case described here would be invalid and result in an error

  I think it is important to differentiate between the QC included in a block and the QC included in a TC: Multiple Vote Aggregators Optimization 3
  * We have a requirement that if `QC_1` is for the prior view, there should not be including a TC.  
  * Lets look at the situation you are describing: 
     * `TC_n` containing a highest `QC_n`. I don't think we have specified anywhere that the proposer for view `N+1` must use the TC. If it has a QC for the last round (here `QC_n`) it can just use that and omit the TC. This is actually preferable anyway, because a TC is probably more than twice as expensive to verify than a QC. 
     * The primary needs some logic to prevent it from accidentally including the TC as well as the QC both for the previous view. Though, I think this logic is very simple (only a few lines of code). Furthermore, I would almost guess that we would add this logic anyway even without the optimization to improve modularity of the code.

AlexHentschel (approved):
🎉 Perfect. 

I feel that spending some time on this PR really payed off. The code is super clean; well documented and exceptionally well tested. Great work.

consensus/hotstuff/validator/validator.go
jordanschalm: Thank you for the detailed explanation - I understand now. The piece I was missing was that the next proposal is two views later, and the diagrams in the notion doc make the flow very clear 👌

jordanschalm (approved):

PR #2457, opened 2022-05-20
Upgrade linter to 1.46.2

zhangchiqing:
## Context

The old version is able to detect type error, but fail to report the error accurately. Instead, it produces a lot of non-error as error, which is very confusing. 

Upgrading the linter to [the latest version](https://github.com/golangci/golangci-lint/tags), resolved the problem. 

I've tested. After I upgraded, I was able to find a type error on my branch which the tool was not able to report accurately before the upgrade.

There is a new warning in `make lint` after the upgrade, seems only matter if `generic` is used:
```
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
```

## Actions after the upgrade
You will need to run `make install-tools` to re-install the latest `golangci-lint`.


turbolent (approved):

kc1116 (approved):

zhangchiqing:
bors merge

PR #2461, opened 2022-05-20
Add Atree reporter to execution state extraction

### LINKED ISSUE #2460: [Execution] Add Atree report to execution state extraction
## Problem

Execution state extraction doesn't report any Atree-related information.

Specifically, it would be useful to know the number of payloads, number of first level hash collisions, etc. in the root checkpoint.

## The Proposed Solution

Report Atree-related info:

- number of FVM/storage/slab payloads
- number of slab array meta/data payloads
- number of slab map meta/data payloads
- number of first level hash collision groups in slab maps
- number of first level hash collisions in each collision group
- etc.

fxamacker:
### Changes

Report:
- number of FVM/storage/slab payloads
- number of slab array meta/data payloads
- number of slab map meta/data payloads
- number of first level hash collision groups in slab maps
- number of first level hash collisions in each collision group
- etc.

Closes #2460


turbolent (commented):
Great work! Thank you for adding this reporter

Is it necessary that we re-implement much of atree here? Could we expose information, types and functions from atree and use them here instead?

fxamacker:
@turbolent

> Is it necessary that we re-implement much of atree here? Could we expose information, types and functions from atree and use them here instead?

Great question!   I initially tried to avoid this approach.  Unfortunately, this was the main tradeoff because I didn't want to expose Atree implementation details (slab encoding) and have to support that in external projects that may rely on it.

We control atree and flow-go, plus this is a mostly self-contained report (which might be modified to report more details in the future).  So it was the lesser of two evils in this case (IMHO) to just copy some code I wrote from Atree into this repo.

I realize this is more code than maybe what Rob Pike had in mind in his [Go Proverbs](https://go-proverbs.github.io/) about copying code vs dependency.  😊

turbolent:
@fxamacker Haha, sounds good! Thank you for writing down your reasoning, that makes sense 👍

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -100,6 +100,10 @@ func extractExecutionState(
 				RWF:   reportFileWriterFactory,
 			},
 			"newFungibleTokenTracker": reporters.NewFungibleTokenTracker(log, reportFileWriterFactory, chain, []string{reporters.FlowTokenTypeID(chain)}),
+			"atree": &reporters.AtreeReporter{
+				Log: log,
+				RWF: reportFileWriterFactory,
+			},
```
ramtinms, line 106: How time-consuming is this reporter?  have we test it on mainnet data yet?

ramtinms (approved):
This would be very useful, only concern is if we need to run the reporters on every spork or not?

cmd/util/cmd/execution-state-extract/execution_state_extract.go
fxamacker: @ramtinms I just confirmed it takes about 4.5 mins on the benchnet using May 13 mainnet checkpoint file.  This PR uses concurrency because it used to take a lot longer.

If this isn't fast enough, I'm wondering if we can run this report concurrently while other reports are running.

zhangchiqing: Is this report done before the state extraction? 

I wonder if possible we can move this report post state extraction.

fxamacker: > Is this report done before the state extraction?

@zhangchiqing I think this report is done [after the state extraction](https://github.com/onflow/flow-go/blob/master/ledger/complete/ledger.go#L415-L420) (after new root checkpoint file is created).

turbolent:
@fxamacker is this already in 0.26?

fxamacker:
@turbolent Yes, I think so (I see the changes in v0.26 branch).

PR #2462, opened 2022-05-20
[Signer Indices] Validate guarantors

zhangchiqing:
This PR let consensus nodes check that guarantors are valid, which are needed for the signer indices optimization.


state/protocol/badger/mutator.go
```diff
@@ -322,6 +322,12 @@ func (m *MutableState) guaranteeExtend(ctx context.Context, candidate *flow.Bloc
 			return state.NewInvalidExtensionErrorf("payload includes expired guarantee (height: %d, limit: %d)",
 				ref.Height, limit)
 		}
+
+		// check the guarantors are correct
+		_, err = protocol.FindGuarantors(m, guarantee)
+		if err != nil {
```
AlexHentschel, line 328: ## ⚠️ 

It would be very important to me to consequently avoid this pattern of "seeing an error and attributing it without further consideration to one particular root cause". 

Reasoning: 
* Consider the case of a corrupted data base, returning some error on read. Or consider a bug in some of the code, where `FindGuarantors` returns an unexpected error. Your code would interpret any error as a symptom of the block being invalid. 
    * At some point in the not-too-distant future, this will result in a slashing challenge against the block proposer. If this accusation is false, the node raising an invalid challenge will be slashed (causing hundreds of thousands of dollars in damage)! 
    * Lets anticipate the point in time when we want to start slashing challenges. Do we want to go over every line of code and look for cases like this? I think that would be prohibitively expensive. 
* In addition, I think this type of undifferentiated error attribution also makes our code very brittle. Maybe at the moment, any error might be a symptom of a broken input. But some engineer might extend the logic of `FindGuarantors` in the future (maybe adding an additional sanity check). 
   * I think it would not be a reasonable ask that an engineer has to go through all usages of the function they extended with additional sanity checks and see whether they broke some implicit assumptions that go beyond the documented API. 
   * The API of `FindGuarantors` is: https://github.com/onflow/flow-go/blob/8c5e6ab0d29f46b4ea954fb2d7e2a40d74947998/state/protocol/util.go#L75-L79 This means that the sentinel errors (and only these!) indicate potentially byzantine inputs  `ErrIncompatibleBitVectorLength`, `ErrIllegallyPaddedBitVector`, `ErrInvalidChecksum`. Other errors are _not_ to be interpreted as symptom of a byzantine input. 


The unfortunate reality is that the protocol state does not properly differentiate between byzantine inputs and unexpected errors. **My ask to you**:
* For the code you add (or the parts of the code you touch), please add byzantine-compatible error handling. Please don't exacerbate the problem by taking the absence of byzantine-compatible error handling in existing code as a reason to ignore it in new code. 

### Problem:
There is a circular assumption in the code regarding detection of byzantine cases, which is not satisfied. Assume a byzantine input, where the `guarantee.ChainID` has some garbage value. 
* `FindGuarantors` delegates the responsibility of detecting this case to the protocol state: https://github.com/onflow/flow-go/blob/8c5e6ab0d29f46b4ea954fb2d7e2a40d74947998/state/protocol/util.go#L86-L87 
* And the protocol state assumes that `FindGuarantors` will spot this problem. 

I am hoping that by advocating for detailed considerations of byzantine inputs, we would increase our chances of finding such problems, se we create less technical debt for later. 

_Suggestion_
1. review the implementations of `Epoch`'s `ClusterByChainID` method and figure out what sentinel errors are expected during normal operations 
2. Document the expected sentinel errors in the [`Epoch` interface](https://github.com/onflow/flow-go/blob/8c5e6ab0d29f46b4ea954fb2d7e2a40d74947998/state/protocol/epoch.go#L80-L82)
3. `FindGuarantors` should handle the expected errors [here](https://github.com/onflow/flow-go/blob/8c5e6ab0d29f46b4ea954fb2d7e2a40d74947998/state/protocol/util.go#L85) accordingly. The goDoc should clearly specify what sentinel is returned in case the cluster is unknown.
4. Differentiate between expected errors and unexpected exceptions in here in line 328 of the `MutableState`.  Something like the following would be super great:
  ```golang
  if err != nil {
     if errors.Is(err, signature.ErrIllegallyPaddedBitVector) || errors.Is(err, signature.ErrIncompatibleBitVectorLength) || errors.Is(err, signature.ErrInvalidChecksum) ||  errors.Is(err, protocol.ErrClusterNotFound) {
         return state.NewInvalidExtensionErrorf("invalid collection guarantee %v: %w", guarantee.ID(), err)
     }
     return fmt.Errorf("unexpected internal error validating guarantors of collection %v: %w", guarantee.ID(), err)
  }
  ```

As part of your PR I would suggest to not touch anything at a higher level, i.e. do not touch the error handling in `MutableState` beyond extending line 328. But at least with this bottom-up approach, we are incrementally adding proper error piece by piece.

state/protocol/badger/mutator_test.go
```diff
@@ -1653,6 +1655,60 @@ func TestHeaderExtendHighestSeal(t *testing.T) {
 	})
 }
 
+// TestExtendInvalidGuarantee checks if Extend method will reject invalid blocks that contain
+// guarantees with invalid guarantors
+func TestExtendInvalidGuarantee(t *testing.T) {
```
AlexHentschel, line 1660: ## 🔍 (requesting to test an additional scenarios)

very nice test. Great documentation. Would be great to explicitly test the following scenarios: 
* `guarantee.ChainID` (identifying the collector cluster) has an unknown value? 
* illegal patting 
* invalid checksum

I think a [similar comment](https://github.com/onflow/flow-go/pull/2414#discussion_r876732747) here has lead to some confusion: 
* the purpose of these tests is _not_ to verify that `FindGuarantors` returns the correct sentinel error 
* the purpose is to test that `MutableState` correctly handles the error:
   * for _expected_ sentinel errors, `MutableState` wraps them as `InvalidExtensionError` 
   * `MutableState` does _not_ misinterpret unexpected exceptions as the input being invalid


unfortunately, we have to test these scenarios over and over again, because we can't mock the logic in `FindGuarantors`. I think adding a validator interface for collection guarantees might be valuable. The validator implementation could then call into `FindGuarantors`. Thereby, we can independently test the logic of detecting the byzantine cases (in validator), and correct handling of the expected lower-level sentinels (`MutableState`)

AlexHentschel (changes requested):
Thanks. Code looks good. Excellent documentation and good tests. 

Unfortunately, I think there are significant gaps where the code that we added related to signer indices (PR  https://github.com/onflow/flow-go/pull/2414 and this PR) does not handle byzantine cases to the level which we desire.

state/protocol/badger/mutator_test.go
```diff
@@ -1653,6 +1655,60 @@ func TestHeaderExtendHighestSeal(t *testing.T) {
...
+		// check Extend should accept this valid block
+		err = state.Extend(context.Background(), block)
+		require.NoError(t, err)
+
+		// now the guarantee has invalid signer indices: the checksum should have 4 bytes, but it only has 1
+		payload.Guarantees[0].SignerIndices = []byte{byte(1)}
+		err = state.Extend(context.Background(), block)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "invalid guarantors")
+		require.Contains(t, err.Error(), "split checksum") // check the error message should tell it's due to wrong checksum
```
AlexHentschel, line 1697: I personally prefer type checks of errors over checking error messages.
In this particular case, we expect that the `MutableState` wraps the `ErrInvalidChecksum` as a `InvalidExtensionError`. I think we can directly test with without having to inspect the error message:
```suggestion
		require.ErrorIs(t, err, signature.ErrInvalidChecksum)
```

state/protocol/badger/mutator_test.go
```diff
@@ -1653,6 +1655,60 @@ func TestHeaderExtendHighestSeal(t *testing.T) {
...
+		// let's test even if the checksum is correct, but signer indices is still wrong because the tailing are not 0,
+		// then the block should still be rejected.
+		invalidSignerIndices := validSignerIndices[:]
+		invalidSignerIndices[len(invalidSignerIndices)-1] = byte(255)
+
+		payload.Guarantees[0].SignerIndices = invalidSignerIndices
+		err = state.Extend(context.Background(), block)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "index vector padded with unexpected bit values")
+	})
```
AlexHentschel, line 1709: similarly here: we expect `ErrIllegallyPaddedBitVector` wrapped as `InvalidExtensionError`:
```suggestion
		require.ErrorIs(t, err, signature.ErrIllegallyPaddedBitVector)
		require.True(t, st.IsInvalidExtensionError(err))
	})
```

state/protocol/badger/mutator.go
zhangchiqing: Good point.

It is more and more common that we would like to distinguish expected error and unexpected exceptions. What you suggested works, but with so many checks in the if statement: `if errors.Is(er, sentinelError1) || errors.Is(err, sentinelError2) || ...)`, it's not easy to maintain. Adding one more sentinel error would have to update all places.

I have two options:
1. Add a `IsUnexpectedFindGuarantorsError(err error) bool` function 
2. Let FindGuarantors to return an extra bool to indicate whether the error is an exception. This could remind the caller to distinguish exception.

1 is simpler, but 2 is more explicit.

AlexHentschel: In my opinion, find `FindGuarantors` cannot decide in all cases, whether an error is expected or unexpected. A good example is the problem at hand:
* When extending the protocol state with a new block, we validate the collection guarantees. A collection guarantee including an unknown value for `guarantee.ChainID` is expected because the block could be proposed by a byzantine consensus node. 
* In contrast, when the execution node decodes the guarantors, the block was already approved by the consensus nodes. Finding an unknown value for `guarantee.ChainID` is fatal in this case.

zhangchiqing: Updated the error handling to handle each sentinel errors separately.

state/protocol/badger/mutator_test.go
zhangchiqing: Yeah, that's right. In the unit tests, I'm able to let FindGuarantees to return all the sentinel errors except `ErrEpochNotCommitted` error. That error isn't easy to create without creating a `guaranteeValidator` just like the `sealValidator` and `receiptValidator`. However, adding `guaranteeValidator` would create a lot of code changes. Will be addressed in a separate PR.

module/signature/errors.go
```diff
@@ -111,7 +111,7 @@ func IsInsufficientSignaturesError(err error) bool {
 
 // IsDecodeSignerIndicesError returns whether err is about decoding signer indices
 func IsDecodeSignerIndicesError(err error) bool {
```
AlexHentschel, line 113: It is really important to me to avoid this pattern of "functions that universally classify errors into expected and unexpected". I have included detailed explanations for my reasoning into [PR 2517](https://github.com/onflow/flow-go/pull/2517). [PR 2517](https://github.com/onflow/flow-go/pull/2517) consolidates the error handling and removes the function `IsDecodeSignerIndicesError`.

module/signature/errors.go
```diff
@@ -111,7 +111,7 @@ func IsInsufficientSignaturesError(err error) bool {
...
-	return errors.Is(err, ErrIllegallyPaddedBitVector) ||
-		errors.Is(err, ErrIncompatibleBitVectorLength) ||
-		errors.Is(err, ErrInvalidChecksum)
+	return errors.As(err, &ErrIllegallyPaddedBitVector) ||
+		errors.As(err, &ErrIncompatibleBitVectorLength) ||
+		errors.As(err, &ErrInvalidChecksum)
```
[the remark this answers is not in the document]
AlexHentschel: Please don't fix this, as I would like to avoid `IsDecodeSignerIndicesError` entirely. See next comment.

AlexHentschel (approved):
Thanks for the iteration. Very clean code. Great tests. Lots of added documentation! 👏 

Looks great. My only remaining two concerns are:
* the existence of method `IsDecodeSignerIndicesError` (see my comment [here](https://github.com/onflow/flow-go/pull/2462#discussion_r885110381))
* Usage of `errors.As` and `require.ErrorAs` (see my comment [here](https://github.com/onflow/flow-go/pull/2462#discussion_r885113742))

I have tried to address both points in my PR https://github.com/onflow/flow-go/pull/2517 (targeting this brach). 

Approving this PR under the assumption that you are happy with my suggestions in https://github.com/onflow/flow-go/pull/2517. Please only merge if you incorporated the suggestions in https://github.com/onflow/flow-go/pull/2517. Oterhwise, happy to discuss.

module/signature/signer_indices.go
```diff
@@ -211,80 +214,88 @@ func EncodeSignersToIndices(
...
+		}
+		return nil, fmt.Errorf("unexpected exception while checking signer indices: %w", err)
 	}
 
 	numberCanonicalNodes := len(canonicalIdentifiers)
 	err = validPadding(signerIndices, numberCanonicalNodes)
 	if err != nil {
-		return nil, fmt.Errorf("signerIndices are invalid: %w", err)
+		if errors.Is(err, ErrIncompatibleBitVectorLength) || errors.Is(err, ErrIllegallyPaddedBitVector) {
+			return nil, NewInvalidSignerIndicesErrorf("invalid padding of signerIndices: %w", err)
```
zhangchiqing, line 254: `DecodeSignerIndicesToIdentifiers` and `DecodeSignerIndicesToIdentities` share most of the logic, the code was duplicated. However, I met a bug that here it is supposed to return `NewInvalidSignerIndicesErrorf`, but was returning `NewInvalidSigTypeErrorf`. We fixed it in `DecodeSignerIndicesToIdentifiers`, but not in `DecodeSignerIndicesToIdentities`. 

To prevent this bug from happening, I extracted the common part of the two functions into `decodeSignerIndices`, so that most logic now is shared and reused.

zhangchiqing:
bors merge

PR #2467, opened 2022-05-23
Memory metrics

dsainati1:
Add histograms for some important metrics to scripts and transactions. In particular, we want to track the memory estimated, the actual memory usage, and the difference between the two.


engine/execution/computation/computer/computer.go
```diff
@@ -429,7 +429,14 @@ func (e *blockComputer) executeTransaction(
 		lg.Info().Msg("transaction executed successfully")
 	}
 
-	e.metrics.ExecutionTransactionExecuted(time.Since(startedAt), tx.ComputationUsed, len(tx.Events), tx.Err != nil)
+	e.metrics.ExecutionTransactionExecuted(
+		time.Since(startedAt),
+		tx.ComputationUsed,
+		memAllocAfter-memAllocBefore,
```
ramtinms, line 435: how expensive is to collect these two memAllocAfter and memAllocBefore, 
was wondering if we should set a flag for them to not collect them by default if they are expensive to collect ?

ramtinms (approved):
Looks good to me, just some minor comments

j1010001 (approved):

PR #2469, opened 2022-05-24
add go workspace files to .dockerignore and .gitignore

SaveTheRbtz:
I have the following [go workspace](https://go.dev/blog/get-familiar-with-workspaces) that simplifies my local development:

```golang
go 1.18

use (
        ./cmd/testclient
        ./crypto
        ./integration
)
```

Sadly, this adds additional entries to `git status` and breaks integration tests.  To prevent this, I need to constantly move these files in and out of the repo.  This PR fixes this workflow by adding go workspace files to both git and docker's ignore lists.


peterargue (approved):
awesome! thanks for adding

.dockerignore
```diff
@@ -4,3 +4,6 @@ integration/localnet/data/
 integration/localnet/trie/
 integration/localnet/docker-compose.nodes.yml
 integration/localnet/targets.nodes.json
+
+go.work
+go.work.sum
```
zhangchiqing, line 9: why not committing them?

SaveTheRbtz: go.work structure is the same as the go.mod, but the whole idea of a separate file is so that it is not accidentally committed to the repo, so that I would assume there won't ever be need for it in git: https://go.googlesource.com/proposal/+/master/design/45713-workspace.md

For docker, it is a bit trickier, since most of the overrides use local paths they would break the Docker build (esp. w/ relic) so we should likely ignore them in docker too.

zhangchiqing: Oh I see. It's meant to replace the `replace directive` which is mostly for local development

zhangchiqing (approved):

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

PR #2470, opened 2022-05-24
chore(tests): cleanup unused fields from loader

SaveTheRbtz:
`txStatsTracker` already dependency injected into the `txTracker` so there is no ned to keep it in an already way too bloated struct.
`supervisorClient` is not used used outside of dep. injection.

### DIFF
--- a/integration/utils/contLoadGenerator.go
+++ b/integration/utils/contLoadGenerator.go
@@ -42,7 +42,6 @@ type ContLoadGenerator struct {
 	numberOfAccounts     int
 	trackTxs             bool
 	flowClient           *client.Client
-	supervisorClient     *client.Client
 	serviceAccount       *flowAccount
 	flowTokenAddress     *flowsdk.Address
 	fungibleTokenAddress *flowsdk.Address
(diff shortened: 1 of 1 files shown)


integration/utils/contLoadGenerator.go
```diff
@@ -51,7 +51,6 @@ type ContLoadGenerator struct {
 	availableAccounts    chan *flowAccount                             // queue with accounts available for   workers
 	happeningAccounts    chan func() (*flowAccount, string, time.Time) // queue with accounts happening after worker processing
 	txTracker            *TxTracker
-	txStatsTracker       *TxStatsTracker
```
gomisha, line 54: have we tested that the loader runs without this field?

gomisha (approved):

integration/utils/contLoadGenerator.go
SaveTheRbtz: yep.  test, build, and manual `make load`

Kay-Zee (approved):

SaveTheRbtz:
bors merge

PR #2474, opened 2022-05-24
Read memory limit from state

robert-e-davidson3:
I branched off of https://github.com/onflow/flow-go/pull/2419 to continue work while Janez is OOO.

Description:

> This PR refactors some of the logic that reads settings from the state to look nicer.
> 
> It also adds the ability to set the memory limit in the service accounts storage, which will override the memory limit set by the context. This all works and is finished.


fvm/fvm_test.go
```diff
@@ -3696,6 +3702,66 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+				AddAuthorizer(chain.ServiceAddress()).
+				SetPayer(chain.ServiceAddress())
+
+			err := testutil.SignTransactionAsServiceAccount(txBody, 0, chain)
+			require.NoError(t, err)
+
+			tx := fvm.Transaction(txBody, 0)
+			err = vm.Run(ctx, tx, view, programs)
+			require.NoError(t, err)
+			require.Greater(t, tx.MemoryUsed, uint64(100_000_000))
```
robert-e-davidson3, line 3759: This fails because the service account is no longer metered if they're the payer and `testutil.SignTransactionAsServiceAccount` sets `txBody.Payer` to the service account. It also does so for the proposal key so the related calls to `txBody` above aren't doing anything.

fvm/fvm_test.go
```diff
@@ -3696,6 +3702,66 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+
+			var foo *errors.MemoryLimitExceededError
+			assert.ErrorAs(t, tx.Err, &foo)
```
robert-e-davidson3, line 3762: This will fail until this branch uses the code from cadence in https://github.com/onflow/cadence/pull/1660

fvm/transactionEnv.go
```diff
@@ -115,53 +115,96 @@ func NewTransactionEnvironment(
...
+				Msgf("error reading execution parameters")
+			return fatal
+		}
+		if err != nil {
+			e.ctx.Logger.
+				Info().
+				Err(err).
+				Msgf("could not get service account to read execution parameters")
+			return nil
+		}
```
ramtinms, line 144: do we need this split, considering if service account is not there its any way fatal

fvm/transactionEnv.go
```diff
@@ -115,53 +115,96 @@ func NewTransactionEnvironment(
...
+	}
+	if !ok {
+		return nil
+	}
```
ramtinms, line 148: not sure what this condition does

cmd/util/cmd/execution-state-extract/checkpoint_status.json
```diff
@@ -0,0 +1,3 @@
+{
+ "succeeded": true
+}
```
ramtinms, line 3: is this something added by this PR or accidentally here.

tools/golangci-lint/Makefile
```diff
@@ -0,0 +1,2 @@
+golangci-lint:
+	go build -o golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
```
ramtinms, line 2: do we need these golangci stuff in this PR ?

ledger/complete/checkpoint_status.json
```diff
@@ -0,0 +1,3 @@
+{
+ "succeeded": true
+}
```
ramtinms, line 3: same here

cmd/util/cmd/execution-state-extract/checkpoint_status.json
robert-e-davidson3: Oh that must have been generated by the tests.

tools/golangci-lint/Makefile
robert-e-davidson3: argh no we do not. I accidentally added it

fvm/bootstrap.go
```diff
@@ -44,6 +44,8 @@ type BootstrapProcedure struct {
 	transactionFees        BootstrapProcedureFeeParameters
 	executionEffortWeights weightedMeter.ExecutionEffortWeights
 	executionMemoryWeights weightedMeter.ExecutionMemoryWeights
+	// executionMemoryLimit of nil means that it won't be set in the state. The FVM will use the default value from the context.
+	executionMemoryLimit *uint64
```
m4ksio, line 48: I think it would be more go-like to use a value here and compare against default value - 0 rather than use pointer.

fvm/executionParameters.go
```diff
@@ -0,0 +1,137 @@
...
+	)
+	if err != nil {
+		// this might be fatal, return as is
+		return nil, err
+	}
+
+	computationWeightsRaw, ok := utils.CadenceValueToWeights(value)
+	if !ok {
+		// this is a non-fatal error. It is expected if the weights are not set up on the network yet.
+		return nil, errors.NewCouldNotGetExecutionParameterFromStateError(
```
m4ksio, line 38: If this is non-fatal - why even return error?

fvm/executionParameters.go
```diff
@@ -0,0 +1,137 @@
...
+			blueprints.TransactionExecutionParametersPathDomain,
+			blueprints.TransactionFeesExecutionEffortWeightsPathIdentifier)
+	}
+
+	// Merge the default weights with the weights from the state.
+	// This allows for weights that are not set in the state, to be set by default.
+	// In case the network is stuck because of a transaction using an FVM feature that has 0 weight
+	// (or is not metered at all), the defaults can be changed and the network restarted
+	// instead of trying to change the weights with a transaction.
+	computationWeights = make(weighted.ExecutionEffortWeights)
```
m4ksio, line 49: small optimisation - we can set a size here

fvm/fvm_test.go
```diff
@@ -241,7 +241,7 @@ func TestPrograms(t *testing.T) {
 	)
 
 	t.Run("script execution programs are not committed",
-		newVMTest().run(
+		newVMTest().withBootstrapProcedureOptions(fvm.WithExecutionMemoryLimit(math.MaxUint64)).run(
```
m4ksio, line 244: This shouldn't be needed in this unit test. 
We set this as an option, so max should be the default not to change the existing behaviour.

fvm/fvm_test.go
```diff
@@ -1276,7 +1278,7 @@ func TestBlockContext_ExecuteScript(t *testing.T) {
 
 	t.Run("storage ID allocation", func(t *testing.T) {
 
-		ledger := testutil.RootBootstrappedLedger(vm, ctx)
+		ledger := testutil.RootBootstrappedLedger(vm, ctx, fvm.WithExecutionMemoryLimit(math.MaxUint64))
```
m4ksio, line 1281: Same as previously

fvm/fvm_test.go
```diff
@@ -3696,6 +3704,70 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+	).run(
+		func(t *testing.T, vm *fvm.VirtualMachine, chain flow.Chain, ctx fvm.Context, view state.View, programs *programs.Programs) {
+			privateKeys, err := testutil.GenerateAccountPrivateKeys(1)
+			require.NoError(t, err)
+
+			accounts, err := testutil.CreateAccounts(vm, view, programs, privateKeys, chain)
+			require.NoError(t, err)
+
+			txBody := flow.NewTransactionBody().
+				SetScript([]byte(`
```
m4ksio, line 3731: We should probably move this to testutils where other specific txes already are

fvm/executionParameters.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/onflow/cadence/runtime/common"
+
+	"github.com/onflow/flow-go/fvm/blueprints"
+	"github.com/onflow/flow-go/fvm/errors"
+	"github.com/onflow/flow-go/fvm/meter/weighted"
+	"github.com/onflow/flow-go/fvm/utils"
+)
+
+// getExecutionWeights reads stored execution effort weights from the service account
+func getExecutionEffortWeights(
```
m4ksio, line 15: Do we have a test for this function? It's rather crucial so we should have it covered.

fvm/fvm_test.go
```diff
@@ -3696,6 +3704,70 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+
+			err = testutil.SignTransaction(txBody, accounts[0], privateKeys[0], 0)
+			require.NoError(t, err)
+
+			tx := fvm.Transaction(txBody, 0)
+			err = vm.Run(ctx, tx, view, programs)
+			require.NoError(t, err)
+			require.Greater(t, tx.MemoryUsed, uint64(100_000_000))
+
+			var memoryLimitExceededError *errors.MemoryLimitExceededError
```
m4ksio, line 3766: Why would this tx trigger memory error though? It doesn't really allocate anything

fvm/fvm_test.go
```diff
@@ -3696,6 +3704,70 @@ func TestSettingExecutionWeights(t *testing.T) {
...
+				}
+			`))
+
+			err = testutil.SignTransaction(txBody, accounts[0], privateKeys[0], 0)
+			require.NoError(t, err)
+
+			tx := fvm.Transaction(txBody, 0)
+			err = vm.Run(ctx, tx, view, programs)
+			require.NoError(t, err)
+			require.Greater(t, tx.MemoryUsed, uint64(100_000_000))
```
m4ksio, line 3764: Where does this constant comes from?

m4ksio (commented):
Holding approval to make sure we have all the tests we need.
Minor comments and nitpicking, great work overall!

ramtinms (approved):
Looks good to me, still think would be beneficial for @janezpodhostnik to take a look as well tomorrow when he is back.

fvm/transactionEnv.go
janezpodhostnik: The service account should exist, but there are cases when it doesn't. Besides tests, the one that comes to mind is bootstrapping. I think the first step during bootstrapping is to create the service account, I don't know how that work if this logic is hit before it exists.

fvm/blueprints/fees.go
```diff
@@ -259,3 +258,38 @@ const setExecutionWeightsScript = `
...
+		AddAuthorizer(service)
+
+	return tx, nil
+}
+
+const setExecutionMemoryLimit = `
+	transaction(newWeights: UInt64, path: StoragePath) {
+		prepare(signer: AuthAccount) {
+			signer.load<UInt64>(from: path)
+			signer.save(newWeights, to: path)
```
janezpodhostnik, line 292:
```suggestion
	transaction(newLimit: UInt64, path: StoragePath) {
		prepare(signer: AuthAccount) {
			signer.load<UInt64>(from: path)
			signer.save(newLimit, to: path)
```
Just noticed this.

fvm/bootstrap.go
janezpodhostnik: I think we could use 0 as the "not-set" value since it doesn't make sense to set it to 0

fvm/fvm_test.go
janezpodhostnik: no, but the weight say that it does: `memoryWeights[common.MemoryKindBreakStatement] = 1_000_000`

fvm/fvm_test.go
janezpodhostnik: It comes from this `memoryWeights[common.MemoryKindBreakStatement] = 1_000_000` and the fact that there is 100 break statements

fvm/handler/programs.go
```diff
@@ -94,7 +94,8 @@ func (h *ProgramsHandler) Get(location common.Location) (*interpreter.Program, b
 			if err != nil {
 				// ignore LedgerIntractionLimitExceededError errors
```
janezpodhostnik, line 95: The comment should be updated as well

fvm/fvm_test.go
janezpodhostnik: Not sure about this as this is a special transaction and does not make sense outside of this test case.

fvm/bootstrap.go
janezpodhostnik: [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/executionParameters.go
janezpodhostnik: [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/fvm_test.go
janezpodhostnik: This is not necessary: [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/fvm_test.go
janezpodhostnik: [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/fvm_test.go
janezpodhostnik: Added some comments [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/fvm_test.go
janezpodhostnik: Added some comments [389fb3c](https://github.com/onflow/flow-go/pull/2474/commits/389fb3cc0adbb7727024a2e00e337500c1994894)

fvm/executionParameters.go
janezpodhostnik: In this context fatal errors are those that should crash the node, because something is very wrong.

In this case this is expected if the value is not set on the network yet. So a `errors.Error` is returned instead of a `error.Fatal`.

This error is currently handled by using default settings instead of those in the state.

In the future all networks should have these values defined in the state and this can be changed to a fatal error and we can remove the fallback to defaults.

fvm/executionParameters.go
janezpodhostnik: yes we do https://github.com/onflow/flow-go/blob/389fb3cc0adbb7727024a2e00e337500c1994894/fvm/fvm_test.go#L3581

m4ksio: Does it test reading values? It seems that it check if values are respected, but given in context params

m4ksio: Does it test reading values? It seems that it check if values are respected, but given in context params

fvm/executionParameters.go
m4ksio: > In the future all networks should have these values defined in the state and this can be changed to a fatal error and we can remove the fallback to defaults.

Yes, thats my question - why use error instead of just default/empty values.

m4ksio: > In the future all networks should have these values defined in the state and this can be changed to a fatal error and we can remove the fallback to defaults.

Yes, thats my question - why use error instead of just default/empty values.

m4ksio (approved):
We should be good to merge, logic is sound.
Just need to add more tests in a follow up PR.

janezpodhostnik:
bors merge

fvm/executionParameters.go
j1010001: We agreed @janezpodhostnik  will add this test in new PR to unblock merge of this for the spork

PR #2476, opened 2022-05-24
Optimize Ledger.Get() by making Forest.Read() use ~5x fewer allocs/op and run faster

### LINKED ISSUE #2475: [Execution] Forest.Read() can use less memory
## Problem

`Forest.Read()` reads payloads from MTrie and makes a deep copy of both payload key and value.  `Forest.Read` is only called by `Ledger.Get` .  `Ledger.Get` creates a new slice of payload values and discards payload keys.  

Deep copying payload key is unnecessary and  it creates 4 heap allocations for each key.  

## The Proposed Solution

Change `Forest.Read` to only deep copy and return payload values.  This change is limited to `Forest` in ledger package.

fxamacker:
### Changes

Change `Forest.Read()` to return `[]ledger.Value` without deep copying payload keys.  This avoids 4 heap allocation per key.

This change doesn't affect `Ledger.Get()` (the caller) because it discards the payload keys.

### Benchmark comparison for batch read size of 100 keys
```
name        old time/op    new time/op    delta
TrieRead100-4     524µs ± 1%     420µs ± 1%  -19.77%  (p=0.000 n=10+10)

name        old alloc/op   new alloc/op   delta
TrieRead100-4     190kB ± 0%      95kB ± 0%  -50.04%  (p=0.000 n=10+10)

name        old allocs/op  new allocs/op  delta
TrieRead100-4     1.52k ± 0%     0.32k ± 0%  -79.17%  (p=0.000 n=10+10)
```

Closes #2475

### DIFF
--- a/cmd/util/cmd/read-execution-state/list-accounts/cmd.go
+++ b/cmd/util/cmd/read-execution-state/list-accounts/cmd.go
@@ -89,12 +89,12 @@ func run(*cobra.Command, []string) {
 			},
 		}
 
-		payload, err := forest.Read(read)
+		values, err := forest.Read(read)
 		if err != nil {
 			return nil, err
 		}
 
-		return payload[0].Value, nil
+		return values[0], nil
 	})
 
 	sth := state.NewStateHolder(state.NewState(ldg))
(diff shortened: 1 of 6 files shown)


fxamacker:
Closing this because it was based on master.  I'll open new PR based on PR #2473.

PR #2483, opened 2022-05-25
Fix script execution panic handling

smnzhu:
Fixes incorrect panic handling for execution node script execution


engine/execution/computation/manager.go
```diff
@@ -170,42 +213,15 @@ func (e *Manager) ExecuteScript(
...
-			}
-
-			elapsed := time.Since(start)
-
-			if r := recover(); r != nil {
-				prepareLog().
-					Interface("recovered", r).
-					Msg("script execution caused runtime panic")
-
-				err = fmt.Errorf("cadence runtime error: %s", r)
```
smnzhu, line 197: The bug is that setting the `err` variable on this line doesn't actually cause the error to be returned, because the parent function `ExecuteScript` does not declare `err` as a named return variable

PR #2484, opened 2022-05-25
Crypto: BLS ciphersuite updates

tarakby:
Restructure BLS ciphersuite and domain separation tags (DST) to comply with :
 - IETF hash to curve drat v14 (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14)
 - IETF BLS signature draft v4 (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04)
 
 The BLS ciphetrsuite of the crypto library is now updated to : 
 - `BLS_SIG_BLS12381G1_XOF:KMAC128_SSWU_RO_POP_` for signatures
 - `BLS_POP_BLS12381G1_XOF:KMAC128_SSWU_RO_POP_` for proofs of possession
 
 When using the crypto library in Flow for signatures with a certain domain tag `DST`, the full tag would be:
 `FLOW-DST-V00-CS00-with-BLS_SIG_BLS12381G1_XOF:KMAC128_SSWU_RO_POP_`
 
 Side changes:
  - expand_message in BLS (hasher in the Go interface) is limited to 128 bytes output only
  - domain tags are moved to `module/signature` package instead of `model/encoding`. 
  - add safety tests
  - README updates


crypto/bls_multisig.go
```diff
@@ -31,18 +31,15 @@ import (
...
+// used for signatures.
+var popKMAC = internalBLSKMAC(blsPOPCipherSuite)
 
 // BLSGeneratePOP returns a proof of possession (PoP) for the receiver private key.
 //
-// The KMAC hasher used in the function is guaranted to be orthogonal to all hashers used
+// The KMAC hasher used in the function is guaranteed to be orthogonal to all hashers used
 // for signatures or SPoCK proofs. This means a specific domain tag is used to generate PoP
 // and is not used by any other application.
 func BLSGeneratePOP(sk PrivateKey) (Signature, error) {
```
AlexHentschel, line 45: It would be great to add documentation around the expected sentinel errors:
* For the method `BLSGeneratePOP`, one could argue that no errors are expected, as this method is presumably called by the person who also generated the private key. 
* However,  for method https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/crypto/bls_multisig.go#L54-L57 I feel invalid inputs are certainly expected during normal operations (including an invalid public key). 

There are also other methods in this class where we don't document errors; while to me it seems that errors are potentially expected during normal operations: 
* https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/crypto/bls_multisig.go#L241
* https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/crypto/bls_multisig.go#L267
* https://github.com/onflow/flow-
* go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/crypto/bls_multisig.go#L405

Some methods [state](https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/crypto/bls_multisig.go#L404-L405) 
> // An error is returned if the key slice is empty.

However, I am not sure that this exhaustively describes all error cases. For example, I think an `invalidInputsError` is also returned if one of the input public keys is invalid. I think the protocol is quite diverse and I am not sure that we can always be absolutely sure invalid public keys are completely impossible for all usages of this method. 

Lastly, invalid public keys vs empty input seem different failure cases. For example, the consensus vote processing contains specific logic (preliminary check [here](https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/module/signature/aggregation.go#L177-L180), so `invalidInputsError` [in later call](https://github.com/onflow/flow-go/blob/dd9c8157e42d70bbaf0be8ad9839359d57f50860/module/signature/aggregation.go#L189-L197) can be attributed to specific failure case) to work around the crypto library lumping different failure cases into one sentinel.

AlexHentschel (approved):
Very nice. Revisions on the cryptographic code look good to me; but I am also not a domain expert. 

Overall, I feel there is room for improvement regarding the specificity of error returns in the crypto library and their documentation. So far, the crypto library largely returns `invalidInputsError` for a diverse set of failure conditions. This makes it hard for higher level code to handle failure paths, specifically in the presence of byzantine actors. 

I don't expect us to go back and refactor the error handling of all existing code. But I would like to encourage us that we revisit and improve error handling when we anyway work on some library. I feel it is the case here, and therefore I'd like to request a detailed review of the error handling and improvements. 
Please see my comment below for more details. 

Thanks

crypto/bls_multisig.go
tarakby: Thanks @AlexHentschel for the valuable comment. 

1. I totally agree the documentation should be improved by listing the error cases exhaustively in every public API 💯  I created https://github.com/dapperlabs/flow-go/issues/6264 to work on this next sprint.

2. The crypto module currently returns 2 main types of errors:
   - `invalidInputErrors` : errors returned when the API is misused by calling it with invalid inputs. 
   - other unexpected errors : errors not related to invalid inputs, but related to unexpected low level failures (memory issues or other).
From a pure crypto perspective, all `invalidInputErrors` are input-related, and the crypto module (as a generic library) doesn't know which of these errors is considered possible for normal operations. This depend closely on the calling business logic (in our case the Flow protocol). This is why these errors aren't sub-classified further at the time the code was written. That said, if the Flow protocol (or another calling logic) requires differentiating `invalidInputErrors` because they have different interpretations (implementation bug/ normal operations/ BFT behaviour or not..), I am happy to create specific sentinel errors. That happened already in DKG for a [specific error ](https://github.com/onflow/flow-go/blob/master/crypto/dkg.go#L60-L66).
 
Ideally the module would create one sentinel error for each different error reason, but that may create a high number of unnecessary sentinels that upper layer logic does not need. This is why I need to work with the protocol engineers to know when such sentinels are needed for Flow protocol (the signatures aggregation case you noted is one good example, but unfortunately I do not know all the examples). I will create sentinels for the ones I know (https://github.com/dapperlabs/flow-go/issues/6264), but feel free to ask me/create tickets for other cases 🙏🏼 


3. To answer your questions about `BLSGeneratePOP` and `BLSVerifyPOP`:
 - `BLSGeneratePOP` fails when it's called with a non-BLS private key, so it can fail from the module perspective. In Flow code, it can only fail because of an implementation bug. 
 - `BLSVerifyPOP` : In Flow, an invalid pubic key is also not expected in normal operation (a public key is decoded from the bytes provided by a staking nodes, and if the bytes are invalid, the decoding would fail). So an error can only happen because of an implementation bug.

AlexHentschel: regarding `BLSVerifyPOP`, lets assume that a byzantine node submitted an invalid public key with an invalid Proof of possession. I think this is a realistic scenario. 
For me as an engineer, it seems reasonable to just hand both of those inputs to the `BLSVerifyPOP`. From my perspective, we have to verify at some point the proof os possession and the public key. At least the code's documentation provides no information that this verification should not be done using `BLSVerifyPOP`.

tarakby: Sorry, my explanation above was very brief, but it's good you understood why the error isn't expected during normal operations in Flow.
The detailed error description will be added in the future ticket.

tarakby:
Closing this PR because separate PRs need to be submitted in the following order:
 - update the crypto module
 - tag the new crypto module
 - update flow-go and use the new crypto module tag

PR #2490, opened 2022-05-26
1717 restrict sending on channels a node should only receive messages on

kc1116:
This PR _flags_ AN's as receive only on the PushBlocks and PushReceipts channels

- add a map to track the node roles that subscribe to a channel to receive messages only
- use require.Regex to check correct warning logs / errs are logged when rejecting messages

https://github.com/dapperlabs/flow-internal/issues/1717


peterargue (commented):
Rather than adding a new layer to the channel role map, I think we should switch to mapping each individual message types to a list of roles that are allowed to send them. It's a bit more new configuration, but is reusable for the unicast case and guarantees that we're checking exactly what's intended within the authorized sender validator.

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
-	require.Equalf(t, uint64(1), hookCalls, "expected 1 warning to be logged")
+	// expecting ErrSenderEjected to be logged
+	require.Regexp(t, validator.ErrSenderEjected, hook.Logs())
+}
+
+// TestAuthorizedSenderValidator_Ejected tests that the authorized sender validator rejects messages from nodes that should only receive on the channel
+func TestAuthorizedSenderValidator_ReceiveOnly(t *testing.T) {
+	sporkId := unittest.IdentifierFixture()
+	identity1, privateKey1 := unittest.IdentityWithNetworkingKeyFixture(unittest.WithRole(flow.RoleConsensus))
+	sn1 := createNode(t, identity1.NodeID, privateKey1, sporkId)
```
yhassanzadeh13, line 372: `createNode` is similar to the already existing [`nodeFixture`](https://github.com/onflow/flow-go/blob/58d3927376643892c39138ce079b87a43d2f548d/network/p2p/fixture_test.go#L88), which resides in the same `p2p_test` package. The `nodeFixture` is a more realistic fixture as it configs the `libp2p` node closely to what we build on a Flow node. Please use that instead and remove the `createNode` one, as having redundant fixtures with similar functionalities degrade the integrity of our tests over the time and also makes the test fixture harder to maintain. Thanks! 

Note-1: most probably you need to add a [`withRole`](https://github.com/onflow/flow-go/blob/58d3927376643892c39138ce079b87a43d2f548d/network/p2p/fixture_test.go#L48) option to `nodeFixture` similar to the other options that we already have. This allows the test to instantiate the libp2p nodes with a specific role. 

Note-2: the `nodeID` parameter of `createNode` is unused.

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
 
 	unittest.RequireReturnsBefore(t, wg.Wait, 5*time.Second, "could not receive message on time")
 
-	// expecting 1 warn calls for each rejected message from ejected node
-	require.Equalf(t, uint64(1), hookCalls, "expected 1 warning to be logged")
+	// expecting ErrSenderEjected to be logged
+	require.Regexp(t, validator.ErrSenderEjected, hook.Logs())
+}
+
+// TestAuthorizedSenderValidator_Ejected tests that the authorized sender validator rejects messages from nodes that should only receive on the channel
```
yhassanzadeh13, line 368:
```suggestion
// TestAuthorizedSenderValidator_ReceiveOnly tests that the authorized sender validator rejects messages from nodes that should only receive on the channel
```

network/p2p/topic_validator_test.go
```diff
@@ -44,13 +42,11 @@ func TestAuthorizedSenderValidator_Unauthorized(t *testing.T) {
...
-		if level == zerolog.WarnLevel {
-			atomic.AddUint64(&hookCalls, 1)
-		}
-	})
-	logger := zerolog.New(os.Stdout).Level(zerolog.WarnLevel).Hook(hook)
+	var (
+		hook   unittest.LoggerHook
+		logger zerolog.Logger
+	)
+	logger, hook = unittest.HookedLogger()
```
yhassanzadeh13, line 49:
```suggestion
	logger, hook := unittest.HookedLogger()
```

network/p2p/topic_validator_test.go
```diff
@@ -223,13 +217,11 @@ func TestAuthorizedSenderValidator_Unstaked(t *testing.T) {
...
```
yhassanzadeh13, line 224:
```suggestion
	logger, hook := unittest.HookedLogger()
```

network/p2p/topic_validator_test.go
```diff
@@ -298,13 +290,11 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
```
yhassanzadeh13, line 297:
```suggestion
	logger, hook := unittest.HookedLogger()
```

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
+
+	authorizedSenderValidator := validator.AuthorizedSenderValidator(logger, channel, func(pid peer.ID) (*flow.Identity, bool) {
+		fid, err := translator.GetFlowID(pid)
+		if err != nil {
+			return &flow.Identity{}, false
```
yhassanzadeh13, line 395:
```suggestion
			return nil, false
```

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
+		}
+		return ids.ByNodeID(fid)
+	})
+
+	require.NoError(t, an1.AddPeer(context.TODO(), *host.InfoFromHost(sn1.Host())))
+
+	// sn1 subscribe to the topic with the topic validator, while an1 will subscribe without the topic validator to allow sn2 to publish unauthorized messages
```
yhassanzadeh13, line 402:
```suggestion
	// sn1 subscribe to the topic with the topic validator, while an1 will subscribe without the topic validator to allow an1 to publish unauthorized messages
```

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
+	_, err = an1.Subscribe(topic)
+	require.NoError(t, err)
+
+	// assert that the nodes are connected as expected
+	require.Eventually(t, func() bool {
+		return len(sn1.ListPeers(topic.String())) > 0 &&
+			len(an1.ListPeers(topic.String())) > 0
+	}, 3*time.Second, 100*time.Millisecond)
+
+	var wg sync.WaitGroup
```
yhassanzadeh13, line 414:
```suggestion

```

network/p2p/topic_validator_test.go
```diff
@@ -371,8 +361,73 @@ func TestAuthorizedSenderValidator_Ejected(t *testing.T) {
...
+	defer cancel2s()
+	err = an1.Publish(timedCtx, topic, data)
+	require.NoError(t, err)
+
+	// sn1 should not receive rejected message from ejected sn2
+	timedCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	checkReceive(timedCtx, t, nil, sub1, &wg, false)
+
+	unittest.RequireReturnsBefore(t, wg.Wait, 5*time.Second, "could not receive message on time")
```
yhassanzadeh13, line 427:
```suggestion
	checkReceive(timedCtx, t, nil, sub1, nil, false)
```

kc1116: We need to pass a pointer to the wait group to checkReceive because it creates a goroutine in the code path where we are not expecting to receive a message.

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2490/#discussion_r887962976

kc1116:
@peterargue In response to https://github.com/onflow/flow-go/pull/2490/#pullrequestreview-989705272 we initially went with the design you suggested (static mapping of message code => list of authorized roles ) @yhassanzadeh13 suggested we refactor and use the channel role mapping see this comment from the first authorized sender validator PR https://github.com/onflow/flow-go/pull/2317/files/8c91881c1e8c124ddb0e31d5abaf785aac2e1b85#r853465515 . @yhassanzadeh13 What are your thoughts on this now ?

kc1116:
closed by https://github.com/onflow/flow-go/pull/2491/

PR #2491, opened 2022-05-26
1706 refactor codec as argument

kc1116:
This PR refactors the network message codec as a parameter for middleware. Currently, the AuthorizedSenderValidator in the middleware creates an instance of cbor codec to decode message types from a message. If this codec is ever changed upstream during the node scaffolding process the validator would break. Adding the codec as a parameter to middleware and passing it to the validators as an argument eliminates this problem.


network/codec/json/decode.go
```diff
@@ -20,74 +22,74 @@ func switchenv2v(env Envelope) (interface{}, error) {
 	switch env.Code {
 
 	// consensus
-	case CodeBlockProposal:
+	case codec.CodeBlockProposal:
```
peterargue, line 25: This is great. I'm happy you're aligning on using a single set of consts for all codecs. 

The original json and cbor consts were offset by 1. Do you know if it's safe to change the value? i.e. is the data stored anywhere, or is it only used in transit so an update at a spork boundary would be safe?

network/codec/cbor/decode.go
```diff
@@ -16,74 +18,74 @@ func switchenv2v(code uint8) (interface{}, error) {
 	switch code {
 
 	// consensus
-	case CodeBlockProposal:
+	case codec.CodeBlockProposal:
```
peterargue, line 21: maybe you could take this a step further and define a new type like
```
type MessageType uint8
```

which has `String()` and `Interface()` methods you can use to get the `what` and `interface{}` values. Then all this mapping code could be defined once and we'd have a more consistent interface to work with instead of `switchenv2what`

network/codec/json/codec.go
```diff
@@ -75,3 +75,20 @@ func (c *Codec) Decode(data []byte) (interface{}, error) {
 
 	return v, nil
 }
+
+// DecodeMsgType is a helper func that returns the underlying message of the message
+func (c *Codec) DecodeMsgType(data []byte) (byte, string, error) {
```
yhassanzadeh13, line 80: The idea of decoupling the message type decoding is nice. Though, the `JSON` implementation of it requires the entire message to be decoded, right? Then by switching to `JSON`, we are silently incurring the overhead of additional decoding per message. 

Is it possible to also encode `JSON` similar to `CBOR` by appending the first byte as the message type (and potentially getting rid of the `Envelope`)?

network/codec/json/decode.go
yhassanzadeh13: A reasonable concern to consider in the next TPM.

network/p2p/topic_validator_test.go
```diff
@@ -52,7 +52,7 @@ func TestAuthorizedSenderValidator_Unauthorized(t *testing.T) {
 	})
 	logger := zerolog.New(os.Stdout).Level(zerolog.WarnLevel).Hook(hook)
 
-	authorizedSenderValidator := validator.AuthorizedSenderValidator(logger, channel, func(pid peer.ID) (*flow.Identity, bool) {
+	authorizedSenderValidator := validator.AuthorizedSenderValidator(logger, channel, cborcodec.NewCodec(), func(pid peer.ID) (*flow.Identity, bool) {
```
yhassanzadeh13, line 55: maybe we can have a `unittest.Codec()` similar to what we have as `unittest.Logger()` currently. This also gives us the luxury of running all relevant tests against different codecs if needed.

network/codec/json/decode.go
kc1116: https://github.com/onflow/flow-go/pull/2491#discussion_r886048382

network/codec/cbor/decode.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/55201f6ad6642613e551228ba154816baaec6790

network/p2p/topic_validator_test.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/e1940b49cd0e78d3eb9bee9ee2aeba9a9069da06

network/message_codes.go
```diff
@@ -0,0 +1,354 @@
...
+var messageCodeMap codeToMessageCodes
+
+// initializeMessageCodeMap initializes the messageCodeMap
+func initializeMessageCodeMap() {
+	messageCodeMap = make(codeToMessageCodes)
+
+	// consensus
+	messageCodeMap[codec.CodeBlockProposal] = MessageCode{
+		Code:               codec.CodeBlockProposal,
+		authorizedChannels: ChannelList{ConsensusCommittee, PushBlocks, ReceiveBlocks},
```
kc1116, line 46: authorizedChannels is kept for readability and completeness since some channels have aliases, we should explicitly list out the alias channels.

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -78,65 +79,37 @@ func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getI
...
+	if prefix, ok := network.ClusterChannelPrefix(channel); ok {
+		authorizedRolesByChannel = code.AuthorizedRolesByChannel(network.Channel(prefix))
+	} else {
+		authorizedRolesByChannel = code.AuthorizedRolesByChannel(channel)
 	}
 
-	// check if message type code is in list of codes corresponding to channel
-	if !containsCode(codes, msgTypeCode) {
-		return nil, fmt.Errorf("invalid message type being sent on channel")
+	// check if role is authorized to send message on channel
```
kc1116, line 98: Here we check the message code authorization. Check if the current channel is in the message code's authorizedChannelsMap (asserts that this messsage is allowed to be sent on the channel) , then check if the senders role is in the channels authorized role list (asserts the sender is allowed to send this message on the current channel)

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -78,65 +79,37 @@ func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getI
...
+	if !authorizedRolesByChannel.Contains(identity.Role) {
+		return fmt.Errorf("sender role is not authorized to send this message type on channel")
 	}
 
+	// check if sender is authorized to participate on the channel
 	// get authorized list of roles for channel
-	roles, ok := channels.RolesByChannel(channel)
+	roles, ok := network.RolesByChannel(channel)
```
kc1116, line 105: We still need to check if the sender is allowed to participate on the channel.

network/message_codes.go
```diff
@@ -0,0 +1,354 @@
...
+	default:
+		return MessageCode{}, fmt.Errorf("invalid encode type (%T)", v)
+	}
+
+	return messageCodeMap[code], nil
+}
+
+// MessageCodeFromByte helper func that performs a sanity check before returning a byte b as a Code
+func MessageCodeFromByte(b byte) (MessageCode, error) {
+	c := codec.Code(b)
```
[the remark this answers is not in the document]
kc1116: I would like to keep any all encoding/decoding logic inside the codec package. I like the fact that currently, you can use the `codec.Code` without converting to `network.MessageCode` .

ledger/complete/checkpoint_status.json
```diff
@@ -0,0 +1,3 @@
+{
+ "succeeded": true
```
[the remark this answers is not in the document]
kc1116: Deleted, not sure why it shows up from time to time on my local flow-go dir

module/metrics/example/consensus/main.go
```diff
@@ -63,8 +64,8 @@ func main() {
 				collector.FinishBlockToSeal(flow.HashToID(entityID))
 			}
 
-			collProvider := engine.TestNetwork.String()
-			collIngest := engine.TestMetrics.String()
+			collProvider := network.TestNetwork.String()
+			collIngest := network.TestMetrics.String()
```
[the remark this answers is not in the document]
kc1116: No it causes weird import cycles because we would still need to import them into the network package and add them to the authorized roles mapping for channel participation. I renamed them for more readability.

engine/ghost/engine/rpc.go
```diff
@@ -47,7 +48,7 @@ func New(net network.Network, log zerolog.Logger, me module.Local, state protoco
 	// create a channel to buffer messages in case the consumer is slow
 	messages := make(chan ghost.FlowMessage, 1000)
 
-	codec := jsoncodec.NewCodec()
+	codec := unittest.NetworkCodec()
```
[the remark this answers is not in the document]
kc1116: No it should return the interface type not the cbor Codec type directly. The cbor codec is the underlying type.

module/metrics/example/consensus/main.go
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/2d73cd4855599a1b12a5e65f1d615a50624021e5

ledger/complete/checkpoint_status.json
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/645b8c414732abdb87bf1e5ec88bed9701487e02

network/p2p/network.go
```diff
@@ -220,7 +219,7 @@ func (n *Network) runMiddleware(ctx irrecoverable.SignalerContext, ready compone
 }
 
 func (n *Network) handleRegisterEngineRequest(parent irrecoverable.SignalerContext, channel network.Channel, engine network.MessageProcessor) (network.Conduit, error) {
-	if !channels.Exists(channel) {
+	if !network.Exists(channel) {
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/2c6c855bf6e1abe8b3c9509959d2d98b6fb4c023

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -78,65 +79,37 @@ func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getI
...
-
-// getRoles returns list of authorized roles for the channel associated with the message code provided
-func getRoles(channel network.Channel, msgTypeCode uint8) (flow.RoleList, error) {
+// isAuthorizedSender checks if node is an authorized role.
+func isAuthorizedSender(identity *flow.Identity, channel network.Channel, code network.MessageCode) error {
 	// echo messages can be sent by anyone
-	if msgTypeCode == cborcodec.CodeEcho {
-		return flow.Roles(), nil
+	if code.Code == codec.CodeEcho {
+		return nil
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/7515e916d89f3331627c96723427d20a3d964d75

engine/ghost/engine/rpc.go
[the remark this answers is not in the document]
peterargue: I guess I mean that it seems strange to use the `unittest` module outside of a `_test.go` file

network/message_codes.go
```diff
@@ -0,0 +1,354 @@
...
+			RequestApprovalsByChunk:  {flow.RoleCollection, flow.RoleExecution},
+			RequestReceiptsByBlockID: {flow.RoleCollection, flow.RoleExecution},
+		},
+	}
+
+	// testing
+	messageCodeMap[codec.CodeEcho] = MessageCode{
+		Code:                  codec.CodeEcho,
+		authorizedChannels:    ChannelList{},
+		authorizedChannelsMap: map[Channel]flow.RoleList{},
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/commit/7515e916d89f3331627c96723427d20a3d964d75

network/validator/pubsub/authorized_sender_validator.go
kc1116: Thats right, I think we can just remove it altogether. Participation is open to anyone but sending is a privilege.

kc1116:
- Added the static configuration and checking based on go type here https://github.com/onflow/flow-go/pull/2491/commits/fe7d7303853038f7df6cefb8beca4b851b24533f . 
- Cleanup removing the DecodeMsgType func from the network codec interface and removing MessageCode wrapper and references inside each codec https://github.com/onflow/flow-go/pull/2491/commits/983cf63f01db274de7a2ead3ac56770f5a3ec65b

cc: @peterargue @smnzhu @yhassanzadeh13

cmd/collection/main.go
```diff
@@ -6,6 +6,8 @@ import (
 
 	"github.com/spf13/pflag"
 
+	"github.com/onflow/flow-go/network"
```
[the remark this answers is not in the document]
smnzhu: Peter is correct here, we should include it with the rest of the imports from flow-go.

cmd/consensus/main.go
```diff
@@ -12,6 +12,8 @@ import (
 
 	"github.com/spf13/pflag"
 
+	"github.com/onflow/flow-go/network"
```
smnzhu, line 15: same here

network/codec/cbor/codec.go
```diff
@@ -46,7 +47,7 @@ func (c *Codec) NewDecoder(r io.Reader) network.Decoder {
 func (c *Codec) Encode(v interface{}) ([]byte, error) {
 
 	// encode the value
-	what, code, err := v2envelopeCode(v)
+	code, what, err := codec.MessageCodeFromInteface(v)
```
smnzhu, line 50:
```suggestion
	code, what, err := codec.MessageCodeFromInterface(v)
```

network/codec/cbor/decoder.go
```diff
@@ -39,7 +38,7 @@ func (d *Decoder) Decode() (interface{}, error) {
 	err = cbor.Unmarshal(data[1:], v) // all but first byte
 	//binstat.Leave(bs2)
 	if err != nil {
-		return nil, fmt.Errorf("could not decode CBOR payload with envelop code %d AKA %s: %w", code, what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
+		return nil, fmt.Errorf("could not decode CBOR payload with envelop code %d AKA %s: %w", data[0], what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
```
smnzhu, line 41:
```suggestion
		return nil, fmt.Errorf("could not decode CBOR payload with message code %d AKA %s: %w", data[0], what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
```

network/codec/json/decode.go
```diff
@@ -6,192 +6,17 @@ import (
...
 	// create the desired message
-	v, err := switchenv2v(env)
-	if nil != err {
-		return nil, err
-	}
-
-	what, err := switchenv2what(env)
-	if nil != err {
-		return nil, err
+	v, what, err := codec.InterfaceFromMessageCode(env.Code)
```
smnzhu, line 17: Since the JSON codec isn't being used as you mentioned, we can consider just removing it

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -8,38 +10,35 @@ import (
...
-// 4. The node is not ejected
-func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getIdentity func(peer.ID) (*flow.Identity, bool)) MessageValidator {
+// to be authorized to send a message if the following are true.
+// 1. The node is staked.
+// 2. The node is not ejected.
+// 3. The message type is a known message type (can be decoded with network codec).
+// 4. The message is authorized to be sent on channel.
+// 4. The sender role is authorized to send message channel.
+// 5. The sender role is authorized to participate on channel.
+func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, c network.Codec, getIdentity func(peer.ID) (*flow.Identity, bool)) MessageValidator {
```
smnzhu, line 28: One important thing here:

Since we are now decoding the payload in the topic validator, I think we should update the signature of `MessageValidator` to take in the decoded payload, instead of taking in `message.Message`. The reason is because there may be multiple `MessageValidators` registered to the base Topic Validator: https://github.com/onflow/flow-go/blob/2569fa44b3ca08eec521281fb7c099cb1853664a/network/validator/pubsub/validator.go#L63

We should decode the payload a single time in the TopicValidator, and add the decoded payload to the `ValidatorData`: https://github.com/onflow/flow-go/blob/2569fa44b3ca08eec521281fb7c099cb1853664a/network/validator/pubsub/validator.go#L79-L82

The ValidatorData is later retrieved by the application layer: https://github.com/onflow/flow-go/blob/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7/network/p2p/readSubscription.go#L82-L83

From here, we can pass the already decoded payload directly to the callback: https://github.com/onflow/flow-go/blob/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7/network/p2p/readSubscription.go#L94

Which calls this function: https://github.com/onflow/flow-go/blob/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7/network/p2p/middleware.go#L541-L552

This way, we can avoid decoding the payload a second time at the application layer: https://github.com/onflow/flow-go/blob/6c9e6b908ce6aca7be57b3077ef2e9b95d04a8c7/network/p2p/network.go#L350-L372

smnzhu (changes requested):
One important comment that should be addressed, but otherwise looks good.

cmd/consensus/main.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/515d6d7816cd87a8b3f004f331a0748a4792fe2b & https://github.com/onflow/flow-go/pull/2491/commits/cd874b35cc83a67cecf2c3cf2895417a34700c32

network/codec/json/decode.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/fdb5c242aa87a585612dcb2ff7fa549027ff17b7

network/codec/cbor/decoder.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/6d1737ae82daa408b72060eab06738ddb25363bc

network/codec/cbor/codec.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/c157869f0c03b862ccc8fe9b94addb0d3e79b614

yhassanzadeh13 (approved):
Except comments already made by Simon, rest LGTM

network/validator/pubsub/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/489ee370357a3be808f998c6ae1f94776c601fab

network/codec/cbor/decoder.go
```diff
@@ -39,7 +38,7 @@ func (d *Decoder) Decode() (interface{}, error) {
 	err = cbor.Unmarshal(data[1:], v) // all but first byte
 	//binstat.Leave(bs2)
 	if err != nil {
-		return nil, fmt.Errorf("could not decode CBOR payload with envelop code %d AKA %s: %w", code, what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
+		return nil, fmt.Errorf("could not decode CBOR payload with message code %d AKA %s: %w", data[0], what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/7b9740335bad3046a371d0f7e4a6cef09cfb748a

network/codec/cbor/decoder.go
```diff
@@ -27,9 +28,7 @@ func (d *Decoder) Decode() (interface{}, error) {
 		return nil, fmt.Errorf("could not decode envelope; len(data)=%d: %w", len(data), err)
 	}
 
-	code := data[0] // only first byte
-
-	what, v, err := envelopeCode2v(code)
+	v, what, err := codec.InterfaceFromMessageCode(data[0])
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/7b9740335bad3046a371d0f7e4a6cef09cfb748a

network/codec/cbor/codec.go
```diff
@@ -98,22 +99,8 @@ func (c *Codec) Decode(data []byte) (interface{}, error) {
 	err = cbor.Unmarshal(data[1:], v) // all but first byte
 	//binstat.Leave(bs2)
 	if err != nil {
-		return nil, fmt.Errorf("could not decode CBOR payload with envelope code %d AKA %s: %w", code, what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
+		return nil, fmt.Errorf("could not decode CBOR payload with envelope code %d AKA %s: %w", data[0], what, err) // e.g. 2, "CodeBlockProposal", <CBOR error>
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/7b9740335bad3046a371d0f7e4a6cef09cfb748a

network/codec/cbor/codec.go
```diff
@@ -85,10 +86,10 @@ func (c *Codec) Decode(data []byte) (interface{}, error) {
 
 	// decode the envelope
 	//bs1 := binstat.EnterTime(binstat.BinNet + ":wire>3(cbor)payload2envelope")
-	code := data[0] // only first byte
+
 	//binstat.LeaveVal(bs1, int64(len(data)))
 
-	what, v, err := envelopeCode2v(code)
+	v, what, err := codec.InterfaceFromMessageCode(data[0])
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2491/commits/7b9740335bad3046a371d0f7e4a6cef09cfb748a

network/middleware.go
```diff
@@ -73,7 +73,7 @@ type Overlay interface {
 	// GetIdentity returns the Identity associated with the given peer ID, if it exists
 	Identity(peer.ID) (*flow.Identity, bool)
 
-	Receive(nodeID flow.Identifier, msg *message.Message) error
+	Receive(nodeID flow.Identifier, msg *message.Message, decodedMsgPayload interface{}) error
```
smnzhu, line 76: Do we actually need to keep the second parameter?

smnzhu (approved):

network/middleware.go
kc1116: Not necessarily, but we could possibly use it in the future. I think it's worth keeping, what do you think?

integration/tests/epochs/suite.go
```diff
@@ -130,8 +130,8 @@ func (s *Suite) Ghost() *client.GhostClient {
 
 func (s *Suite) TearDownTest() {
 	s.log.Info().Msg("================> Start TearDownTest")
-	s.net.Remove()
-	s.cancel()
+	//s.net.Remove()
+	//s.cancel()
```
peterargue, line 134: what's this?

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -8,38 +10,34 @@ import (
...
-
-	return func(ctx context.Context, from peer.ID, msg *message.Message) pubsub.ValidationResult {
+	return func(ctx context.Context, from peer.ID, msg interface{}) pubsub.ValidationResult {
 		identity, ok := getIdentity(from)
 		if !ok {
 			log.Warn().Str("peer_id", from.String()).Msg("could not verify identity of sender")
 			return pubsub.ValidationReject
 		}
 
+		// redundant check if the node is ejected so that we can fail fast before decoding
```
peterargue, line 40: is this redundant check still needed now that we pre-decode the message?

network/validator/pubsub/authorized_sender_validator.go
```diff
@@ -8,38 +10,34 @@ import (
...
-// 3. The authorized roles list for the channel contains the senders role.
-// 4. The node is not ejected
+// to be authorized to send a message if the following are true.
+// 1. The node is staked.
+// 2. The node is not ejected.
+// 3. The message type is a known message type (can be decoded with network codec).
+// 4. The message is authorized to be sent on channel.
+// 4. The sender role is authorized to send message channel.
+// 5. The sender role is authorized to participate on channel.
 func AuthorizedSenderValidator(log zerolog.Logger, channel network.Channel, getIdentity func(peer.ID) (*flow.Identity, bool)) MessageValidator {
```
peterargue, line 27: I think we should add direct unit test coverage for `AuthorizedSenderValidator`

network/message_authorization_config.go
```diff
@@ -0,0 +1,300 @@
...
+	// testing
+	case *message.TestMessage:
+		return echo, nil
+
+	// dkg
+	case *messages.DKGMessage:
+		return dkgMessage, nil
+
+	default:
+		return MsgAuthConfig{}, fmt.Errorf("could not get authorizatoin config for message with type (%T)", v)
```
peterargue, line 298:
```suggestion
		return MsgAuthConfig{}, fmt.Errorf("could not get authorization config for message with type (%T)", v)
```

network/message_authorization_config.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// execution state synchronization
+	executionStateSyncRequest = MsgAuthConfig{
+		String: "",
+		config: nil,
+	}
+	executionStateDelta = MsgAuthConfig{
+		String: "",
+		config: nil,
+	}
```
peterargue, line 149: that's this? If these are supposed to be undefined, can you add a comment explaining

peterargue (commented):
looks good. added a few small comments/questions. I think we should add some direct unit test coverage of the authorized sender

network/validator/pubsub/authorized_sender_validator.go
kc1116: There is an issue to add test that cover each message code https://github.com/dapperlabs/flow-internal/issues/1718 , this is separate from the work done in this PR which is refactoring network codec etc but not changing logic in authorized sender validator.

network/message_authorization_config.go
kc1116: Added [deprecated] to the comment https://github.com/onflow/flow-go/pull/2491/commits/d40fc606ec58e98592cc2fbf5553ed5b47bb0ae1

module/emulator/emulator_client.go
```diff
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"google.golang.org/grpc"
```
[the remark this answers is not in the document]
peterargue: ok. in the future, it's best to put unrelated changes into a different pr

PR #2493, opened 2022-05-26
upload log level for memory weights

zhangchiqing:
Execution node is producing lots of INFO logs like this, which it's better to use debug level log.
```
{"level":"info","node_role":"execution","node_id":"b5dd412056f6f0f2624550debb71d57e27f93c33d04b977ef024d11976a826bb","error":"[Error Code: 1112] could not get execution parameter from the state (address: f4527793ee68aede path: storage/executionMemoryWeights)","time":"2022-05-26T17:32:33Z","message":"could not set execution memory weights. Using defaults"}
```

If we do want to know whether the default weights was used, better to add it to an existing info log as an extra field.


janezpodhostnik (approved):

peterargue (approved):

zhangchiqing:
bors merge

PR #2496, opened 2022-05-26
memory threshold 1 GB for v0.26.1

j1010001:
Adding 1GB memory metering limit for pentest network build based on v0.26.1


Kay-Zee:
no longer needed AFAIK

PR #2498, opened 2022-05-26
Add Safety Threshold for EECC

jordanschalm:
* Adds `EpochCommitSafetyThreshold` to protocol state as a global parameter (defined by the root snapshot)
  * Implements defaults and validation for this value in the bootstrap CLI and integration tests. The present rule is to define a default buffer value for each network, then ensure that we have at least that many views both (1) between DKG end and the commitment deadline and (2) between the commitment deadline and the epoch end
* Adds storage methods for `EpochCommitSafetyThreshold` in the protocol state

#### TODOs
- [x] Set epoch commitment deadline to a reasonable value (and adjust validation) for integration tests + localnet

#### To be implemented in a subsequent PR:
* Modifying the condition for triggering epoch fallback:
  * The view threshold before which an `EpochCommit` event must be sealed will become `epoch_commitment_deadline`, rather than the final view of the epoch
  * The protocol state will inject a protocol-defined fallback epoch into the protocol state (rather than having no defined `Next` epoch)
    * If a `Next` epoch has been setup, but not committed, it would be overwritten


storage/badger/operation/spork.go
```diff
@@ -30,3 +30,11 @@ func InsertProtocolVersion(version uint) func(*badger.Txn) error {
 func RetrieveProtocolVersion(version *uint) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeProtocolVersion), version)
 }
+
+func InsertEpochCommitSafetyThreshold(threshold uint64) func(*badger.Txn) error {
+	return insert(makePrefix(0), threshold)
```
jordanschalm, line 35: todo prefix

cmd/bootstrap/cmd/finalize.go
```diff
@@ -634,3 +641,28 @@ func generateEmptyExecutionState(
...
+	defaultSafetyThreshold, err := protocol.DefaultEpochCommitSafetyThreshold(chainID)
+	if err != nil {
+		log.Fatal().Err(err).Msg("unable to validate epoch config")
+	}
+
+	// sanity check 1: the safety threshold is >= the default for the chain
+	if flagEpochCommitSafetyThreshold < defaultSafetyThreshold {
+		log.Fatal().Msgf("potentially unsafe epoch config: epoch commit safety threshold smaller than expected (%d < %d)", flagEpochCommitSafetyThreshold, defaultSafetyThreshold)
+	}
+	// sanity check 2: the difference between DKG end and safety threshold is also >= the default safety threshold
```
durkmurder, line 660: I think it's misplaced, should be moved to next if statement

cmd/bootstrap/cmd/finalize.go
```diff
@@ -634,3 +641,28 @@ func generateEmptyExecutionState(
 	log.Info().Msg("")
 	return
 }
+
+// validateEpochConfig validates configuration of the epoch commitment deadline.
+func validateEpochConfig() {
```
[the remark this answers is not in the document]
AlexHentschel: +1

state/protocol/badger/state.go
```diff
@@ -456,6 +456,15 @@ func (state *State) bootstrapSporkInfo(root protocol.Snapshot) func(*badger.Txn)
 			return fmt.Errorf("could not insert protocol version: %w", err)
 		}
 
+		threshold, err := params.EpochCommitSafetyThreshold()
+		if err != nil {
+			return fmt.Errorf("could not get epoch commit safety threshold: %w", err)
+		}
+		err = operation.InsertEpochCommitSafetyThreshold(threshold)(tx)
```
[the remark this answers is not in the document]
jordanschalm: I was thinking the same thing, but decided not to, because the values are inserted only once and the retrieval pattern is to read them individually.

state/protocol/params.go
```diff
@@ -45,4 +46,49 @@ type GlobalParams interface {
...
+	// Example: A service event is emitted in block A. The seal for A is included in C.
+	// A<-B(RA)<-C(SA)<-...<-R
+	//
+	// A service event S is considered sealed w.r.t. a reference block R if:
+	// * S was emitted during execution of some block A, s.t. A is an ancestor of R
+	// * The seal for block A was included in some block C, s.t C is an ancestor of R
+	//
+	// When we finalize the first block B with B.View >= d:
+	// HAPPY PATH: If an EpochCommit service event has been sealed w.r.t. B, no action is taken.
+	// FALLBACK PATH: If no EpochCommit service event has been sealed w.r.t. B, epoch fallback mode (EECC) is triggered.
```
AlexHentschel, line 69: nice differentiation between sealing and finality. Appreciate the precision 🖖 🙇

storage/badger/operation/spork.go
```diff
@@ -30,3 +30,11 @@ func InsertProtocolVersion(version uint) func(*badger.Txn) error {
 func RetrieveProtocolVersion(version *uint) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeProtocolVersion), version)
 }
+
+func InsertEpochCommitSafetyThreshold(threshold uint64) func(*badger.Txn) error {
```
AlexHentschel, line 34: some doc would be great 🙏

AlexHentschel (approved):
Thanks for the effort on the detailed doc. This in particular [the explanation about the `EpochCommitSafetyThreshold`](https://github.com/onflow/flow-go/blob/88c1146f2adb85c05d64f838baadcc2dc8e93f5b/state/protocol/params.go#L50-L93) 🤩 is great; helped me a lot to understand the operational implications of the parameters 💡.

durkmurder (approved):

cmd/bootstrap/cmd/finalize.go
jordanschalm: Yeah, I meant the two conditionals below taken together - but agree it's confusing. Updating

PR #2502, opened 2022-05-26
[FVM] reducing register touches 

ramtinms:
-


fvm/context.go
```diff
@@ -34,7 +34,7 @@ type Context struct {
 	CadenceLoggingEnabled         bool
 	EventCollectionEnabled        bool
 	ServiceEventCollectionEnabled bool
-	AccountFreezeAvailable        bool
+	AccountFreezeEnabled          bool
```
ramtinms, line 37: renamed it to match other ones.

fvm/context.go
```diff
@@ -87,13 +87,11 @@ func defaultContext(logger zerolog.Logger) Context {
 		CadenceLoggingEnabled:         false,
 		EventCollectionEnabled:        true,
 		ServiceEventCollectionEnabled: false,
-		AccountFreezeAvailable:        false,
+		AccountFreezeEnabled:          true,
```
janezpodhostnik, line 90: Why did the default change?

fvm/transactionVerifier.go
```diff
@@ -13,24 +13,31 @@ import (
...
 
-type TransactionSignatureVerifier struct {
+// TransactionVerifier verifies the content of the transaction by
+// checking accounts (authorizers, payer, proposer) are not frozen
+// checking there is no double signature
+// all signatures are valid
+// all accounts provides enoguh weights
+//
+// if KeyWeightThreshold is set to a negative number, signature verification is skipped
+type TransactionVerifier struct {
```
janezpodhostnik, line 23: Whats the benefit of merging all the checks into one processor?

I was thinking that the frozen checker could actually be run for all transactions in a block(chunk?) in parallel before any transaction invocation. This would only change the behaviour of the freeze so that it is valid only from the next block on, which I think would be acceptable.

janezpodhostnik (commented):
Looks good, I just have some questions.

fvm/state/accounts.go
```diff
@@ -17,22 +17,15 @@ import (
...
-	KeyCode               = "code"
-	KeyContractNames      = "contract_names"
-	KeyPublicKeyCount     = "public_key_count"
-	KeyStorageUsed        = "storage_used"
-	KeyAccountFrozen      = "frozen"
-	KeyStorageIndex       = "storage_index"
-	uint64StorageSize     = 8
-	AccountFrozenValue    = 1
-	AccountNotFrozenValue = 0
+	KeyAccountStatus  = "account_status"
```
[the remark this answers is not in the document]
ramtinms: actually it is part of what I'm trying to do but might need to do it in a separate pr and reduce key size for all account dependent registers to use an enum or something

fvm/state/accounts_status.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/hex"
+
+	"github.com/onflow/flow-go/fvm/errors"
+)
+
+type AccountStatus uint8
+
+const (
+	maskExist  byte = 0b0000_0001
+	maskFrozen byte = 0b1000_0000
```
ramtinms, line 13: Instead of using enum style I did it mask style, cause if something gets frozen and unfrozen we can maintain future states for accounts, while maintaining frozen state

fvm/context.go
ramtinms: I actually changed the semantic of this parameter, 
now we don't change this based on if an account is payer or not, we can use it as a high level flag to enable and disable the feature.

fvm/state/accounts.go
```diff
@@ -17,22 +17,16 @@ import (
...
+	KeyContractNames  = "contract_names"
+	KeyStorageUsed    = "storage_used"
+	KeyStorageIndex   = "storage_index"
+	uint64StorageSize = 8
 )
 
-func keyPublicKey(index uint64) string {
-	return fmt.Sprintf("public_key_%d", index)
-}
-
```
ramtinms, line 35: moved it to the end of file

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -63,19 +63,12 @@ func extractExecutionState(
...
-		orderedMapMigration := mgr.OrderedMapMigration{
-			Log:       log,
-			OutputDir: dir,
+		accountStatusMigration := mgr.AccountStatusMigration{
+			Logger: log,
 		}
 
 		migrations = []ledger.Migration{
-			orderedMapMigration.Migrate,
-			storageUsedUpdateMigration.Migrate,
```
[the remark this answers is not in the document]
ramtinms: yes, I'll add it back in

fvm/transactionVerifier.go
ramtinms: I think checking frozen accounts should happen at the time of  the register reads, for example, we do have extra checks when a contract is going to load from an account that is not the authorizer of the tx but should still block that call, etc ...

fvm/state/accounts.go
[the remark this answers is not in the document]
ramtinms: I did reduce this key for this PR but we should do it for others as well soon.

fxamacker (approved):
LGTM!  I focused on changes with `exists` and `frozen` registers, and have one minor comment.

fvm/transactionVerifier.go
```diff
@@ -230,3 +246,27 @@ func (v *TransactionSignatureVerifier) checkSignatureDuplications(tx *flow.Trans
...
+
+	err := accounts.CheckAccountNotFrozen(tx.ProposalKey.Address)
+	if err != nil {
+		return fmt.Errorf("checking frozen account failed: %w", err)
+	}
+
+	err = accounts.CheckAccountNotFrozen(tx.Payer)
+	if err != nil {
+		return fmt.Errorf("checking frozen account failed: %w", err)
+	}
```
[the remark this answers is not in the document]
ramtinms: nice idea, though it requires way more changes as I don't think we can read batch of registers yet from inside the fvm yet. going to create a follow up PR to batch read in places that are possible.

ramtinms:
bors merge

ramtinms:
bors cancel

PR #2503, opened 2022-05-26
chore(tests): cleanup worker and add tests

SaveTheRbtz:
Following changes:
* switch from channel close to a cancelation through context: this would allow context propagation in the future.
* start the first job immediately on Start() instead of waiting for the first Tick.
* add basic test to prevent accidental breakage.

### DIFF
--- a/integration/utils/worker.go
+++ b/integration/utils/worker.go
@@ -1,42 +1,50 @@
 package utils
 
 import (
+	"context"
 	"time"
 )
+
+type workFunc func(workerID int)
 
 type Worker struct {
 	workerID int
 	interval time.Duration
-	work     func(workerID int)
+	work     workFunc
 	ticker   *time.Ticker
-	quit     chan struct{}
+
+	ctx    context.Context
+	cancel context.CancelFunc
 }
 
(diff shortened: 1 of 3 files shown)


integration/utils/worker.go
```diff
@@ -1,42 +1,66 @@
...
 	}
 }
 
 func (w *Worker) Start() {
-	w.ticker = time.NewTicker(w.interval)
-	w.quit = make(chan struct{})
-
 	go func() {
-		for {
+		defer w.wg.Done()
```
Kay-Zee, line 42: why do we need this one? we don't even have a corresponding `w.wg.Add(1)` for this

SaveTheRbtz: it's a tricky one.  Initially, this wg was split into two: one for all the workers (to prove that they exited) and one for the loop (to prove that no new workers would be submitted).   Then I've simplified that by coalescing them into one, and doing `wg.Add(1)` at the creation of the worker.   Let me move it to the Start() function instead so that it'll be move obvious.

Kay-Zee (approved):

PR #2504, opened 2022-05-27
chore(tests): cleanup worker stats tracker and add tests

SaveTheRbtz:
Mostly cosmetic changes, except for the cleanup of the stale timestamp buckets that we were previously leaking.

Also while here added a basic test.

### DIFF
--- a/integration/utils/worker_stats_tracker.go
+++ b/integration/utils/worker_stats_tracker.go
@@ -10,11 +10,12 @@ import (
 
 // WorkerStatsTracker keeps track of worker stats
 type WorkerStatsTracker struct {
+	mux              sync.Mutex
 	workers          int
 	txsSent          int
 	txsSentPerSecond map[int64]int // tracks txs sent at the timestamp in seconds
-	mux              sync.Mutex
-	printer          *Worker
+
+	printer *Worker
 }
 
 // NewWorkerStatsTracker returns a new instance of WorkerStatsTracker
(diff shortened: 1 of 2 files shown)


integration/utils/worker_stats_tracker.go
```diff
@@ -40,37 +41,38 @@ func (st *WorkerStatsTracker) AddWorker() {
...
-	st.txsSent += 1
-
-	now := time.Now().Unix()
-	if c, ok := st.txsSentPerSecond[now]; ok {
-		st.txsSentPerSecond[now] = c + 1
-		return
-	}
-	st.txsSentPerSecond[now] = 1
+	st.txsSent++
+	st.txsSentPerSecond[now]++
```
Kay-Zee, line 54: nice:

https://go.dev/play/p/_bSytnoSM4K

integration/utils/worker_stats_tracker.go
```diff
@@ -40,37 +41,38 @@ func (st *WorkerStatsTracker) AddWorker() {
...
 	for timestamp, count := range st.txsSentPerSecond {
-		if time.Unix(timestamp, 0).Before(start) {
+		if timestamp < start.Unix() {
+			toDelete = append(toDelete, timestamp)
 			continue
 		}
 		sum += count
 	}
 
+	for _, timestamp := range toDelete {
```
Kay-Zee, line 72: seems we weren't deleting these before, i guess this was giving an incorrect average before then?

integration/utils/worker_stats_tracker_test.go
```diff
@@ -0,0 +1,23 @@
...
+	st := NewWorkerStatsTracker()
+	st.StartPrinting(time.Second)
+	defer st.StopPrinting()
+
+	st.AddWorker()
+
+	assert.Equal(t, st.AvgTPSBetween(time.Now().Add(-time.Hour), time.Now()), 0.)
+	st.AddTxSent()
+	st.AddTxSent()
+	assert.Greater(t, st.AvgTPSBetween(time.Now().Add(-time.Hour), time.Now()), 0.)
```
Kay-Zee, line 22: i forget how this is calculated, is it easy enough to calculate exactly what this would return? like 2/3600s or is it calculated differently than that?

Kay-Zee (approved):
minor nit picks, but overall looks good for clean up

integration/utils/worker_stats_tracker.go
SaveTheRbtz: It was mostly affecting the speed and memory usage, though given that there is only a single entry per second you need to run it for multiple days to notice any measurable effect.

integration/utils/worker_stats_tracker_test.go
SaveTheRbtz: Good point.  I generally try to make any `time.*`-dependent tests to be veeeeery permissive, but here I think I can be slightly more aggressive.

PR #2505, opened 2022-05-27
1714 add rate limiting interceptor to the LN & EN gRPC API's

kc1116:
This PR adds the rate limiter interceptor used in the AN gRPC API to the LN & EN gRPC API's .


engine/execution/rpc/engine.go
```diff
@@ -26,6 +28,16 @@ import (
...
+var (
+	DefaultAPIRatelimits  map[string]int
+	DefaultAPIBurstlimits map[string]int
+)
+
+func init() {
+	DefaultAPIRatelimits = make(map[string]int)
+	DefaultAPIBurstlimits = make(map[string]int)
+}
+
```
yhassanzadeh13, line 40: Not sure what the purpose of these is. It seem they are literally initiating empty `map`s, which are used as the default values when loading the flags, right? then why not initiating and passing them at the `execution_builder`? e.g., 

```
flags.StringToIntVar(&e.exeConf.apiRatelimits, "api-rate-limits", make(map[string]int), "per second rate limits for GRPC API methods e.g. Ping=300,ExecuteScriptAtBlockID=500 etc.")
```

kc1116: The empty maps were just placeholders while I determine the values to use for each endpoint. That is done in this commit https://github.com/onflow/flow-go/pull/2505/commits/91fa062ad2832ee5d958cec35cb333a32b424df8 . I took a look at the req/s handled by each endpoint and cross-referenced it with the consensus rate and execution rate over the past 6 months. I did not find any correlation between spikes in req/s and consensus/block rate. Thus I took the peak req/s value for each endpoint over the past 6 months and used double that value as the req/s config value for rate limiting.

kc1116: We have decided to not use any default values for rate limiting, and that it should be optional for the node operator. https://github.com/onflow/flow-go/pull/2505/commits/792e3338e29debfba777e410d61f272659389be4

engine/access/rpc/rate_limit_interceptor.go
```diff
@@ -53,8 +53,8 @@ func NewRateLimiterInterceptor(log zerolog.Logger, apiRateLimits map[string]int,
 	}
 }
 
-// unaryServerInterceptor rate limits the given request based on the limits defined when creating the rateLimiterInterceptor
-func (interceptor *rateLimiterInterceptor) unaryServerInterceptor(ctx context.Context,
+// UnaryServerInterceptor rate limits the given request based on the limits defined when creating the rateLimiterInterceptor
```
peterargue, line 56: If this is going to be used for all the gRPC interfaces, we should move this somewhere common. Maybe `engine/rpc/common`?

kc1116: https://github.com/onflow/flow-go/pull/2505/commits/6fc22e9e58ffd509b2d49d44b6b084f800e99b93

cmd/collection/main.go
```diff
@@ -146,6 +148,8 @@ func main() {
 		// epoch qc contract flags
 		flags.BoolVar(&insecureAccessAPI, "insecure-access-api", false, "required if insecure GRPC connection should be used")
 		flags.StringSliceVar(&accessNodeIDS, "access-node-ids", []string{}, fmt.Sprintf("array of access node IDs sorted in priority order where the first ID in this array will get the first connection attempt and each subsequent ID after serves as a fallback. Minimum length %d. Use '*' for all IDs in protocol state.", common.DefaultAccessNodeIDSMinimum))
+		flags.StringToIntVar(&apiRatelimits, "api-rate-limits", map[string]int{}, "per second rate limits for GRPC API methods e.g. Ping=300,SendTransaction=500 etc. note limits allow globally to all clients.")
+		flags.StringToIntVar(&apiBurstlimits, "api-burst-limits", map[string]int{}, "burst limits for gRPC API methods e.g. Ping=100,SendTransaction=100 etc. note limits allow globally to all clients.")
```
peterargue, line 152: sorry, this was my bad
```suggestion
		flags.StringToIntVar(&apiRatelimits, "api-rate-limits", map[string]int{}, "per second rate limits for GRPC API methods e.g. Ping=300,SendTransaction=500 etc. note limits apply globally to all clients.")
		flags.StringToIntVar(&apiBurstlimits, "api-burst-limits", map[string]int{}, "burst limits for gRPC API methods e.g. Ping=100,SendTransaction=100 etc. note limits apply globally to all clients.")
```

cmd/execution_builder.go
```diff
@@ -159,6 +161,8 @@ func (e *ExecutionNodeBuilder) LoadFlags() {
 			flags.StringVar(&e.exeConf.s3BucketName, "s3-bucket-name", "", "S3 Bucket name for block data uploader")
 			flags.DurationVar(&e.exeConf.edsDatastoreTTL, "execution-data-service-datastore-ttl", 0,
 				"TTL for new blobs added to the execution data service blobstore")
+			flags.StringToIntVar(&e.exeConf.apiRatelimits, "api-rate-limits", map[string]int{}, "per second rate limits for GRPC API methods e.g. Ping=300,ExecuteScriptAtBlockID=500 etc. note limits allow globally to all clients.")
+			flags.StringToIntVar(&e.exeConf.apiBurstlimits, "api-burst-limits", map[string]int{}, "burst limits for gRPC API methods e.g. Ping=100,ExecuteScriptAtBlockID=100 etc. note limits allow globally to all clients.")
```
peterargue, line 165:
```suggestion
			flags.StringToIntVar(&e.exeConf.apiRatelimits, "api-rate-limits", map[string]int{}, "per second rate limits for GRPC API methods e.g. Ping=300,ExecuteScriptAtBlockID=500 etc. note limits apply globally to all clients.")
			flags.StringToIntVar(&e.exeConf.apiBurstlimits, "api-burst-limits", map[string]int{}, "burst limits for gRPC API methods e.g. Ping=100,ExecuteScriptAtBlockID=100 etc. note limits apply globally to all clients.")
```

peterargue (approved):

smnzhu (approved):

PR #2506, opened 2022-05-27
[Execution] Delta getter and setter method refactor

sideninja:
Small code optimization and duplication removal. On setting delta data the register is created twice. 

*Disclaimer: I came across this during studying flow-go code and I know its a pretty small change so if you don't think it's worth it discard it*


ramtinms:
I think we are going to get rid of RegisterID and RegisterValue soon

PR #2509, opened 2022-05-28
[Crypto] improve flaky random test

tarakby:
A rare flakiness has been seen in a statistical tests under crypto/random 
 
 - use math rand and add logging for failing tests seeds
 - use higher sampling for a statistical test to avoid flakiness (the higher the sampling, the closer the result to theoretical distribution)
 
 cc @durkmurder


durkmurder (approved):

durkmurder:
@gomisha this test started failed a few times on feature branch for Active Pacemaker. Here one if the evidences: https://github.com/onflow/flow-go/runs/6622170032?check_suite_focus=true

tarakby:
The test is probabilistic and there is always a low chance of failure, failures should happen very rarely. I'm also surprised it happens "a few times" on the feature branch.

gomisha (approved):

crypto/random/rand_test.go
```diff
@@ -82,26 +82,33 @@ func TestChacha20Compliance(t *testing.T) {
...
 // to test another PRG implementation.
 
 // Simple unit testing of Uint using a very basic randomness test.
 // It doesn't evaluate randomness of the output and doesn't perform advanced statistical tests.
 func TestUint(t *testing.T) {
+	seedMathRand(t)
 
 	seed := make([]byte, Chacha20SeedLen)
-	crand.Read(seed)
+	rand.Read(seed)
```
zhangchiqing, line 101: why replacing crypto/rand with math/rand?

tarakby: `math/rand` produces deterministic randoms depending on the initializing seed, while `crypto/rand` doesn't allow that. This is useful when a test fails, and by logging the seed we can reproduce the fail. 

Note that `math/rand` shouldn't be used outside of tests when secure entropy is needed.

tarakby:
bors merge

PR #2511, opened 2022-05-29
chore(tests): do not go through the resolver for localhost

SaveTheRbtz:
Currently, we go through the resolver codepath each time we create a local connection.  This is not too expensive for the benchmark (though it can create unexpected spikes in latencies), but it can be expensive for the docker's resolver.  To avoid DDoS'ing it, let's use `127.0.0.1` for now.  If our nodes switch to IPv6-only we can switch this to `::1`.

### DIFF
--- a/integration/loader/main.go
+++ b/integration/loader/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/hex"
 	"flag"
+	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


Kay-Zee (approved):

zhangchiqing (approved):

SaveTheRbtz:
bors merge

PR #2512, opened 2022-05-29
chore(tests): mark localnet and benchmarks owned by the performance team

SaveTheRbtz:
Also, added @zhangchiqing to the localnet based on commit history.


Kay-Zee (approved):
We don't use the Codeowners file as much anymore, but still good to have as general documentation probably

zhangchiqing (approved):

SaveTheRbtz:
bors merge

PR #2517, opened 2022-05-30
Suggestion for PR #2462

AlexHentschel:
## Suggestion for PR https://github.com/onflow/flow-go/pull/

_Context_
I consider methods like the following as generally unsafe 
https://github.com/onflow/flow-go/blob/632d86098f05e025d52c27692a580d67cf49c9d4/module/signature/errors.go#L112-L117
reasoning:
* In my experience they incentivise engineers to so skip expected byzantine paths in the protocol. For example, the engineer extend method [`DecodeSignerIndicesToIdentifiers`](https://github.com/onflow/flow-go/blob/632d86098f05e025d52c27692a580d67cf49c9d4/module/signature/signer_indices.go#L217) and introduce another sentinel error `X`. For their particular use case, `X` is expected during normal operations, so they add it to `IsDecodeSignerIndicesError`. However, in other segments of the code, this error might not be expected. But by adding `X` to `IsDecodeSignerIndicesError`, the engineer has now broken the Byzantine resilience of other parts of the protocol implementation. In my experience, methods like `IsDecodeSignerIndicesError` create the false impression that certain errors are expected/unexpected irrespective of their content. 
  
  Btw, a while back _exactly_ the situation described happened. Newer team members commonly struggle with internalizing the different engineering requirement for Flow, specifically that they do give enough considerations to the recovery paths from expected byzantine failures and just treat them as errors. Therefore, it is important to me to avoid pattens entirely that of incentivise engineers to rush through error handling logic, including the pattern of "introduce a function that universally classifies errors into expected and unexpected". 

* Having said all this, there are cases, where some set of sentinel errors _universally_ indicate a byzantine input. I think this is the case for [`DecodeSignerIndicesToIdentifiers`](https://github.com/onflow/flow-go/blob/632d86098f05e025d52c27692a580d67cf49c9d4/module/signature/signer_indices.go#L217). In this case, I would suggest to have the implementation already consolidate the errors. 

_Suggested Solution_

In our current example, we have 3 errors that all signal that the input violates protocol rules: https://github.com/onflow/flow-go/blob/632d86098f05e025d52c27692a580d67cf49c9d4/module/signature/signer_indices.go#L214-L217 The higher-level code is essentially only interested whether this is a broken input or not (highlighted by the existence of function `IsDecodeSignerIndicesError`). So why not have `DecodeSignerIndicesToIdentifiers` return a higher-level error (called `InvalidSignerIndicesError`)? 
* Thereby we avoid adding patters to the code base that we know incentivise new engineers to unsafe designs. 
* The API of `DecodeSignerIndicesToIdentifiers` becomes thinner, because we reduce the number of cases that the higher-level implementations have to handle when calling into `DecodeSignerIndicesToIdentifiers`.

This PR implements the suggestion ☝️: for [`DecodeSigTypeToStakingAndBeaconSigners`](https://github.com/onflow/flow-go/blob/4fbfb368a5c7b6b776121762acb657aeeca0c5d2/module/signature/signer_indices.go#L123), [`DecodeSignerIndicesToIdentifiers`](https://github.com/onflow/flow-go/blob/4fbfb368a5c7b6b776121762acb657aeeca0c5d2/module/signature/signer_indices.go#L218) and [`DecodeSignerIndicesToIdentities`](https://github.com/onflow/flow-go/blob/4fbfb368a5c7b6b776121762acb657aeeca0c5d2/module/signature/signer_indices.go#L259) the error handling is integrated into the functions themselves. They now wrap the lower-level errors into `signature.InvalidSignerIndicesError` and `signature.IsInvalidSigTypesError` respectively. Higher-level code now has to check only for these errors, which eliminates the need for `IsDecodeSignerIndicesError`. Tests are updated.

### DIFF
--- a/module/signature/signer_indices.go
+++ b/module/signature/signer_indices.go
@@ -235,7 +235,7 @@ func DecodeSignerIndicesToIdentifiers(
 	err = validPadding(signerIndices, numberCanonicalNodes)
 	if err != nil {
 		if errors.Is(err, ErrIncompatibleBitVectorLength) || errors.Is(err, ErrIllegallyPaddedBitVector) {
-			return nil, NewInvalidSigTypesErrorf("invalid padding of signerIndices: %w", err)
+			return nil, NewInvalidSignerIndicesErrorf("invalid padding of signerIndices: %w", err)
 		}
 		return nil, fmt.Errorf("unexpected exception while checking padding of signer indices: %w", err)
 	}
@@ -304,7 +304,6 @@ func validPadding(bitVector []byte, numUsedBits int) error {
 	if l != bitutils.MinimalByteSliceLength(numUsedBits) {
 		e := fmt.Errorf("the bit vector contains a payload of %d used bits, so it should have %d bytes but has %d bytes: %w",
 			numUsedBits, bitutils.MinimalByteSliceLength(numUsedBits), l, ErrIncompatibleBitVectorLength)
-		fmt.Println(errors.Is(e, ErrIncompatibleBitVectorLength))
 		return e
 	}
 	// Condition 1 implies that the number of padded bits must be strictly smaller than 8. Otherwise, the vector
--- a/module/signature/signer_indices.go
+++ b/module/signature/signer_indices.go
@@ -1,6 +1,7 @@
 package signature
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/onflow/flow-go/ledger/common/bitutils"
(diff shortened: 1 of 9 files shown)


module/signature/signer_indices.go
```diff
@@ -224,13 +225,19 @@ func DecodeSignerIndicesToIdentifiers(
...
+		}
+		return nil, fmt.Errorf("unexpected exception while checking signer indices: %w", err)
 	}
 
 	numberCanonicalNodes := len(canonicalIdentifiers)
 	err = validPadding(signerIndices, numberCanonicalNodes)
 	if err != nil {
-		return nil, fmt.Errorf("signerIndices are invalid: %w", err)
+		if errors.Is(err, ErrIncompatibleBitVectorLength) || errors.Is(err, ErrIllegallyPaddedBitVector) {
+			return nil, NewInvalidSigTypesErrorf("invalid padding of signerIndices: %w", err)
```
zhangchiqing, line 238: Why is this `NewInvalidSigTypesErrorf` rather than `NewInvalidSignerIndicesErrorf`?

zhangchiqing (approved):
Nice. Thanks for the fix and refactor. I like the idea of embedding low level sentinel error into high  level sentinel error

module/signature/signer_indices.go
AlexHentschel: This is for the function `DecodeSigTypeToStakingAndBeaconSigners`. This function does _not_ decode the signer indices. To the best of my understanding, we call this function _after_ the signer indices were decoded to identities (or identifiers). In a second step, we decode the signature types (random beacon sig or staking sig) for each signer. Hence, I included a different sentinel so we can differentiate the root cause of errors.  But it would also be fine to consider 'InvalidSignerIndicesError' as a high-level error also covering failure cases for decoding the `sigType`. 🤷 either is fine by me.

zhangchiqing: Yeah, I think using `InvalidSignerIndicesError` is necessary, because the caller of `DecodeSignerIndicesToIdentities` always expect `InvalidSignerIndicesError`.

PR #2518, opened 2022-05-30
[AdminTool] Admin require approvals

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/6248

This PR adds a command `set-required-approvals-for-sealing` to admin tool in order to change the required number of approvals for sealing. This allows consensus nodes to transit to sealing with approvals without restarting the node.


module/validation/seal_validator.go
```diff
@@ -44,26 +44,27 @@ func NewSealValidator(
 	results storage.ExecutionResults,
 	seals storage.Seals,
 	assigner module.ChunkAssigner,
-	requiredApprovalsForSealConstruction uint,
+	requiredApprovalsForSealConstructionGetter module.RequiredApprovalsForSealConstructionInstanceGetter,
 	requiredApprovalsForSealVerification uint,
 	metrics module.ConsensusMetrics,
 ) (*sealValidator, error) {
+	requiredApprovalsForSealConstruction := requiredApprovalsForSealConstructionGetter.GetValue()
 	if requiredApprovalsForSealConstruction < requiredApprovalsForSealVerification {
```
durkmurder, line 52: I feel like this getter approach is a bit dangerous since we basically render this check useless.

module/updatable_configs/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,68 @@
...
+	return createRequiredApprovalsForSealConstruction()
+}
+
+// SetValue updates the requiredApprovalsForSealConstruction and return the old value
+// This assume the caller has validated the new value
+func (r *RequiredApprovalsForSealConstructionInstance) SetValue(requiredApprovalsForSealConstruction uint) (uint, error) {
+	r.Lock()
+	defer r.Unlock()
+
+	err := validation.ValidateRequireApprovals(requiredApprovalsForSealConstruction)
```
zhangchiqing, line 52: validation is done before updating the new value

cmd/consensus/main.go
```diff
@@ -180,16 +183,9 @@ func main() {
 		}).
 		Module("beacon keys", func(node *cmd.NodeConfig) error {
 			safeBeaconKeys = bstorage.NewSafeBeaconPrivateKeys(dkgState)
-			return err
+			return nil
```
zhangchiqing, line 186: this seem to be a bug. fixed.

module/validation/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	"github.com/onflow/flow-go/module/chunks"
+)
+
+// ValidateRequireApprovals validates the given value against the default of other values.
+func ValidateRequireApprovals(requiredApprovalsForSealConstruction uint) error {
+	// NOTE: validating against the default value. Even though we take in new values from flag, but
+	// in practice we never use this flag, because it would need a coordination of all nodes updating
+	// the flag, which will most likely done by updating a new image with a new default value.
```
zhangchiqing, line 13: explaining why validating against default values

module/validation/seal_validator.go
zhangchiqing: Validation is added to ValidateRequireApprovals, which got run in `SetValue`

module/updatable_configs/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,68 @@
...
+// which is created by the very first call
+func AcquireRequiredApprovalsForSealConstructionGetter() module.RequiredApprovalsForSealConstructionInstanceGetter {
+	return createRequiredApprovalsForSealConstruction()
+}
+
+// SetValue updates the requiredApprovalsForSealConstruction and return the old value
+// This assume the caller has validated the new value
+func (r *RequiredApprovalsForSealConstructionInstance) SetValue(requiredApprovalsForSealConstruction uint) (uint, error) {
+	r.Lock()
+	defer r.Unlock()
```
zhangchiqing, line 50: Mutex is better than `atomic`, because we want to performance validation while updating the value

zhangchiqing:
flakey integration caused by 
```
        container.go:219: testingdock: container start failure: Error response from daemon: driver failed programming external connectivity on endpoint access_2 (0098f55dca3add44102bfc254da815b7d7af279b754bb85a0b1fbbeab93184ce): Bind for 0.0.0.0:45221 failed: port is already allocated
```

jordanschalm (commented):
I agree with Yurii that we should use a more generic way to accomplish dynamically updating configs. I think we will want to do this for many different components, and the approach here would be a lot of overhead to implement every time. I also think we should:
* avoid the global singleton pattern
* avoid the assumption that modification of the config field is sufficient in general 

Here's another suggestion (alternative to requiring restartable components) for doing this in a way that is more easily scalable to many different updatable configs/components. 

1. The `NodeBuilder` provides:
* one generic admin command to update any any updatable config
* an API for components to register their updatable config fields 
```go
// function provided by the component which validates and updates the config
// and refreshes any parts of the component to handle the changes as needed
type UpdateConfigField func(any) error

type NodeBuilder struct {
	...
 
	// called in component initialization, when a component wants to make a
	// field configurable they register it here. this data is made available to the
	// admin command
	RegisterUpdatableConfig(configName string, updateConf UpdateConfigField) error
}

type UpdateConfigAdminCommand struct {}

func (cmd *UpdateConfigAdminCommand) Handler(ctx context.Context, req *admin.CommandRequest) (interface{}, error) {
	updateConf := cmd.getUpdatableConfig(req.Data.ConfigName)
	
	err = updateConf(req.Data.Value)
	// return an error to user
```

2. Individual components implement dynamic updating for specific config fields
```go
type Component struct {
	conf Config
	...
}

// the function this component registers with ConfigUpdater. This function is called to
// update the config whenever we receive an admin request to do so
func (c *Component) updateConfigX(val any) error {
	// make sure val is expected type
	// make sure val is valid new value for config
	// update config field internally
	// update any dependencies on config
}

// during initialization, the component can register the dynamic config
Component("some component", func(conf *NodeConfig)  {
	component := asdf.NewComponent(...)
	conf.RegisterUpdatableConfig("component.x", component.UpdateX)
	return component
```

* We could use an approach similar to our [Badger iteration function](https://github.com/onflow/flow-go/blob/a484a3e88d8bf88448731704758c67dd5a4cb889/storage/badger/operation/common.go#L207) to handle different types of config fields. 
* We could also have the NodeBuilder check for a component implementing a certain interface which enumerates updatable config fields (similar to how we detect when a component implements `Component` vs `ReadyDoneAware`) and automatically register them

engine/consensus/sealing/core.go
```diff
@@ -94,6 +86,7 @@ func NewCore(
 	sealsMempool mempool.IncorporatedResultSeals,
 	approvalConduit network.Conduit,
 	config Config,
+	getRequiredApprovalsForSealConstruction module.RequiredApprovalsForSealConstructionInstanceGetter,
```
zhangchiqing, line 89: There are a few reasons that I'd like to inject this `module.RequiredApprovalsForSealConstructionInstanceGetter` module into the sealing core.

1. the sealing core is reading the dynamic config, and I don't want to expose the setter interface, that's why I provided only the Getter interface, and [let NodeBuilder create the setter](https://github.com/onflow/flow-go/pull/2518/files#diff-7d960671cc56ce82783f5366bad31f7108cf00270c1bb39b0e00fa57ca96334dR199) and only [let the admin use the Setter interface.](https://github.com/onflow/flow-go/pull/2518/files#diff-cb7163a6cdf8bc3d94772f989f6fb80fe289118a752166146084d8b412dc4e40R23). 
2. The admin needs [access to a global singleton object](https://github.com/onflow/flow-go/pull/2518/files#diff-cb7163a6cdf8bc3d94772f989f6fb80fe289118a752166146084d8b412dc4e40R23). However, I don't want other modules like this sealing core to depend on a global single object, which would make testing much harder. Modules like this sealing core who are using the config, should just call the Getter's method, without knowing it's a stateful object or not. 
3. Why not a `ConfigReader.GetValue(KeyForRequiredApprovals)` ? because such function would return an interface in order to support different types of dynamic configs to be returned, and I would have to cast it into `uint`, which gives up type safety check.
4. I'd like to use different locks for different dynamic configs, so that accessing different dynamic configs will have no race condition. If I use a centralized config module to manage different dynamic configs, then there must be some shared map object to acquire a lock from first, then getting a value with different lock, which is introducing race condition overhead. 

I was also thinking about more generic way, but I didn't find one that satisfies the above items, so I ended up avoid over engineering, just started simple and we can refactor later if things are more clear. 

@jordanschalm @durkmurder

cmd/consensus/main.go
```diff
@@ -198,6 +195,25 @@ func main() {
 				return fmt.Errorf("invalid consensus parameters: requiredApprovalsForSealConstruction > chunkAlpha")
 			}
 
+			// create the singleton instance and set the value with the flag
+			_, err := updatable_configs.AcquireRequiredApprovalsForSealConstructionSetter().SetValue(requiredApprovalsForSealConstruction)
```
jordanschalm, line 199: Even without generalizing the admin command, I think we can remove the global singleton and instead create a variable for the config field in this field.
```go
Module({
  requiredApprovalsForSealConstructionConfig = updatable_configs. NewRequiredApprovalsForSealConstructionConfigField(flagRequiredApprovalsForSealConstruction)
}).
...
AdminCommand({
  return &admincommon.SetRequiredApprovalsForSealingCommand{
    requiredApprovalsForSealConstructionConfig: requiredApprovalsForSealConstructionConfig,
  }
}).
...
Module({
  // create sealing engine
}).
...
Module({
  // create seal validator
}).
```

engine/consensus/sealing/core.go
```diff
@@ -29,28 +29,20 @@ import (
...
 // DefaultEmergencySealingActive is a flag which indicates when emergency sealing is active, this is a temporary measure
 // to make fire fighting easier while seal & verification is under development.
 const DefaultEmergencySealingActive = false
 
 // Config is a structure of values that configure behavior of sealing engine
 type Config struct {
-	EmergencySealingActive               bool   // flag which indicates if emergency sealing is active or not. NOTE: this is temporary while sealing & verification is under development
-	RequiredApprovalsForSealConstruction uint   // min number of approvals required for constructing a candidate seal
-	ApprovalRequestsThreshold            uint64 // threshold for re-requesting approvals: min height difference between the latest finalized block and the block incorporating a result
+	EmergencySealingActive    bool   // flag which indicates if emergency sealing is active or not. NOTE: this is temporary while sealing & verification is under development
```
jordanschalm, line 38: Any reason not to keep the updatable field with the rest of the config?
```suggestion
	EmergencySealingActive    bool   // flag which indicates if emergency sealing is active or not. NOTE: this is temporary while sealing & verification is under development
	RequiredApprovalsForSealConstruction updatable.UintGetter
```

utils/unittest/fixtures.go
```diff
@@ -2067,3 +2068,12 @@ func TransactionResultsFixture(n int) []flow.TransactionResult {
 	}
 	return results
 }
+
+func NewRequiredApprovalsForSealConstructionInstance(val uint) *updatable_configs.RequiredApprovalsForSealConstructionInstance {
```
jordanschalm, line 2072: we can remove this when removing the global singleton

module/updatable_configs/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,63 @@
...
+// min number of approvals required for constructing a candidate seal
+type RequiredApprovalsForSealConstructionInstance struct {
+	sync.RWMutex
+	requiredApprovalsForSealConstruction uint
+}
+
+var createRequiredApprovalsForSealConstructionOnce sync.Once
+var instanceRequiredApprovalsForSealConstruction *RequiredApprovalsForSealConstructionInstance
+
+func createRequiredApprovalsForSealConstruction() *RequiredApprovalsForSealConstructionInstance {
```
AlexHentschel, line 20: I personally find that the method name is over-specific. It says it creates the instance, but that is only true for the first time. How about we rename it to `acquireRequiredApprovalsForSealConstruction` ?

module/updatable_configs/validation/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,35 @@
+package validation
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/module/defaults"
+)
+
+// ValidateRequireApprovals validates the given value against the default of other values.
+func ValidateRequireApprovals(requiredApprovalsForSealConstruction uint) error {
```
AlexHentschel, line 10: Please include documentation about error returns

module/updatable_configs/validation/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func validateRequireApprovals(requiredApprovalsForSealConstruction uint, requiredApprovalsForSealVerification uint, chunkAlpha uint) error {
+	// We need to ensure `requiredApprovalsForSealVerification <= requiredApprovalsForSealConstruction <= chunkAlpha`
+	if requiredApprovalsForSealVerification > requiredApprovalsForSealConstruction {
+		return fmt.Errorf("invalid consensus parameters, requiredApprovalsForSealVerification (%v)  > requiredApprovalsForSealConstruction (%v)",
+			requiredApprovalsForSealVerification, requiredApprovalsForSealConstruction)
+	}
+
+	if requiredApprovalsForSealConstruction > chunkAlpha {
```
AlexHentschel, line 29: technically, the largest valid threshold is `requiredApprovalsForSealConstruction <= 2/3 * chunkAlpha` Otherwise, the system is not robust against byzantine nodes. Imagine we set `requiredApprovalsForSealConstruction = chunkAlpha`. Then a single byzantine verifier could halt the entire protocol by not approving.

module/updatable_configs/validation/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,35 @@
...
+// ValidateRequireApprovals validates the given value against the default of other values.
+func ValidateRequireApprovals(requiredApprovalsForSealConstruction uint) error {
+	// NOTE: validating against the default value. Even though we take in new values from flag, but
+	// in practice we never use this flag, because it would need a coordination of all nodes updating
+	// the flag, which will most likely done by updating a new image with a new default value.
+	return validateRequireApprovals(
+		requiredApprovalsForSealConstruction,
+		defaults.DefaultRequiredApprovalsForSealValidation,
+		// chunks.DefaultChunkAssignmentAlpha,
+		3,
```
AlexHentschel, line 18: ## ⚠️ 
I think **we can not use default values**. This is because the chunk alpha can be changed as a command line parameter: 
https://github.com/onflow/flow-go/blob/3e92041711fa82d322ef803fb3c88fd42f599b18/cmd/consensus/main.go#L148-L150

Validation must consider the current values (and not the default values, which could be different). 

For example, assume that we set `chunkAlpha` to 10. Then, the admin too will refuse to set required `requiredApprovalsForSealConstruction` to 5, despite this being a perfectly fine value.

module/updatable_configs/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/module/defaults"
+	"github.com/onflow/flow-go/module/updatable_configs/validation"
+)
+
+// min number of approvals required for constructing a candidate seal
+type RequiredApprovalsForSealConstructionInstance struct {
+	sync.RWMutex
+	requiredApprovalsForSealConstruction uint
```
AlexHentschel, line 14: Agree with [Jordan's comment](https://github.com/onflow/flow-go/pull/2518#discussion_r889329071). An atomic will have much lower runtime penalties. We can verify the new value without lock.

module/validation/seal_validator.go
```diff
@@ -44,26 +36,27 @@ func NewSealValidator(
...
 		return nil, fmt.Errorf("required number of approvals for seal construction (%d) cannot be smaller than for seal verification (%d)",
 			requiredApprovalsForSealConstruction, requiredApprovalsForSealVerification)
 	}
```
AlexHentschel, line 47: we have now implemented the validation logic redundantly in different places. From my perspective, it would be beneficial to re-use the same implementation. Thereby we reduce the risk of inconsistent code updates.

AlexHentschel (changes requested):
Overall I am concerned with the validity checks (see comment here). Comparing with default values is not safe.

module/updatable_configs/validation/require_approvals_for_seal_construction.go
zhangchiqing: This is not easy to do, because we want the `DefaultRequiredApprovalsForSealConstruction` to be updatable by any module from any file.

Why?

Because the admin tool doesn't have access to the NodeBuilder instance to access this variable, therefore, we had to make it as a global singleton. 

But we don't have to, if we can create a stateful config module, and pass it to the admin tool. 

That way we can read the latest values, and create the stateful config module with it, and pass to the admin tool. 

This would be the most clean way.

module/validation/seal_validator.go
zhangchiqing: The above validation is now just a sanity check. The validation has been moved to module/updatable_configs/validation/require_approvals_for_seal_construction.go

AlexHentschel (approved):
Overall looks good to me. 
* most of my comments below are stylistic in nature
* Please make sure you improve the documentation of error returns (added specific comments)
* However, there are some comments from prior review rounds that I feel would good to address:
   * https://github.com/onflow/flow-go/pull/2518#discussion_r892725131
   * https://github.com/onflow/flow-go/pull/2518#discussion_r892727504 (code has now shifted to `module/updatable_configs/validation/require_approvals_for_seal_construction.go` but still no doc about error returns)
   * address https://github.com/onflow/flow-go/pull/2518#discussion_r885727829 with comment in code
* On a design level, the separation of parameters makes the code harder for me to follow. Specifically it makes it harder for me to verify that correctness criteria are enforced. I have added a comment to this discussion here

module/updatable_configs/require_approvals_for_seal_construction.go
```diff
@@ -0,0 +1,60 @@
...
+	return &requiredApprovalsForSealConstructionInstance{
+		requiredApprovalsForSealConstruction: atomic.NewUint32(uint32(requiredApprovalsForSealConstruction)),
+		requiredApprovalsForSealVerification: requiredApprovalsForSealVerification,
+		chunkAlpha:                           chunkAlpha,
+	}, nil
+}
+
+// SetValue updates the requiredApprovalsForSealConstruction and return the old value
+// This assume the caller has validated the new value
+func (r *requiredApprovalsForSealConstructionInstance) SetValue(requiredApprovalsForSealConstruction uint) (uint, error) {
```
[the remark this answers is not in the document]
zhangchiqing: If the value is invalid, yes, it won't crash.

To me, either to crash is determined by the caller, not this module itself. But the module is responsible to help caller distinguish between normal error and exceptional error. 

If this module involves a database call, then we need to use sentinel error type to distinguish between data not found, and database exceptional error. 

However, this method is doing a pure computation, meaning there is no side-effect, therefore no exceptional error. 

If there is no exception error to distinguish from, I usually don't bother to create a sentinel error, just create an error with `fmt.Errorf` and return it. 

If the method only returns one type of error, then I usually don't specifically mention about the error type, but just saying it will return error. 

I'm not sure how to apply the default convention "(any error leads to component or node crash unless explicitly specified otherwise)." Do you mean we should make a sentinel error even if there is only one type of error?

engine/consensus/sealing/core.go
jordanschalm: > The updatable field is managed by a stateful instance, which needs to be referenced as pointer, where as Config is passed as value.

I don't understand why this would prevent us from including the `SealingConfigsGetter` with the other config:
```go
type Config struct {
    ...
    sealingConfig SealingConfigsGetter
}
```
Config can be passed by value, `sealingConfig` is a pointer field, copies of Config point to the same instance of `sealingConfig`.

Alex's suggestion of putting all the config in the `SealingConfigsGetter` seems like another good approach.

module/updatable_configs/sealing_configs_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func TestRequiredApprovalsForSealingContruction(t *testing.T) {
+
+	instance, err := updatable_configs.NewSealingConfigs(
+		flow.DefaultRequiredApprovalsForSealConstruction, flow.DefaultRequiredApprovalsForSealValidation, flow.DefaultChunkAssignmentAlpha)
+	require.NoError(t, err)
+
+	// should get the default value
+	val := instance.RequireApprovalsForSealConstructionDynamicValue()
+	require.Equal(t, uint(1), val)
```
jordanschalm, line 20:
```suggestion
	require.Equal(t, flow.DefaultRequiredApprovalsForSealConstruction, val)
```

module/updatable_configs/validation/require_approvals_for_seal_construction.go
jordanschalm: Could we add a TODO for this, and a note for why it's acceptable to leave as-is for now?

module/validation/seal_validator_test.go
```diff
@@ -121,7 +112,12 @@ func (s *SealValidationSuite) TestSealInvalidAggregatedSigCount() {
 //   ... <- LatestSealedBlock <- B0 <- B1{ Result[B0], Receipt[B0] } <- B2 <- ░newBlock{ Seal[B0]}░
 // The gap of 1 block, i.e. B2, is required to avoid a sealing edge-case
 // (see test `TestSeal_EnforceGap` for more details)
-func (s *SealValidationSuite) TestSealEmergencySeal() {
+func (s *SealValidationSuite) TestSealEmergencySeal_VerificationRequire0ApprovalEmergencyNotTriggered() {
```
jordanschalm, line 115: I think the godoc is outdated for this function, these parts:
> and that the emergency-seal metric will be incremented.

>  seal which has 0 signatures for at least one chunk will be accepted

We're checking that emergency sealing is not triggered on line 128 and the seal we're testing has 2 approvals (line 124)

utils/unittest/fixtures.go
```diff
@@ -2080,3 +2082,15 @@ func TransactionResultsFixture(n int) []flow.TransactionResult {
 	}
 	return results
 }
+
+func NewSealingConfigs(val uint) module.SealingConfigsSetter {
```
jordanschalm, line 2086:
```suggestion
func NewSealingConfigs(requiredApprovalsForSealConstruction uint) module.SealingConfigsSetter {
```

module/validation/seal_validator_test.go
```diff
@@ -132,29 +128,46 @@ func (s *SealValidationSuite) TestSealEmergencySeal() {
...
 	}).Return()
 	s.sealValidator.metrics = metrics
-	//
-	_, err := s.sealValidator.Validate(newBlock)
+
+	_, err = s.sealValidator.Validate(newBlock)
 	s.Require().NoError(err)
+}
+
+func (s *SealValidationSuite) TestSealEmergencySeal_VerificationRequire1ApprovalReceive1Approval() {
```
jordanschalm, line 136: It would be great to have short godocs for these test cases.

jordanschalm (approved):
Looks good, final round of comments are re documentation and organization.

module/updatable_configs/require_approvals_for_seal_construction.go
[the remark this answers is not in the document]
AlexHentschel: > this method is doing a pure computation, meaning there is no side-effect, therefore no exceptional error.
> If the method only returns one type of error, then I usually don't specifically mention about the error type, but just saying it will return error.

I am not trying to convince you to use sentinel errors. I am requesting that you document exactly what you explained here in your comment. **Please clearly document that any error is equivalent to a no-op.** (so other engineers don't have to reverse-engineer this insight from the code)

zhangchiqing:
I added another command `get-required-approvals-for-sealing`, in order to get the latest value. Tested working with localnet:

The following logs show I got the default value, which is 1, and then set it to `0`, and get it again, it returns the updated value, which is `0`
```
~/flow-go/integration/localnet on (leo/admin-require-approvals)
▸ curl localhost:3706/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-required-approvals-for-sealing"}'
{"output":"1"}⏎                                                                                                                         
~/flow-go/integration/localnet on (leo/admin-require-approvals)
▸ curl localhost:3706/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-required-approvals-for-sealing", "data": 0}'
{"output":"ok"}⏎                                                                                                                        
~/flow-go/integration/localnet on (leo/admin-require-approvals)
▸ curl localhost:3706/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-required-approvals-for-sealing"}'
{"output":"0"}⏎ 
```

PR #2519, opened 2022-05-31
chore(tests): remove unused batchLoadGenerator

SaveTheRbtz:
This has been unused for quite a while and replaced with a contLoadGenerator


Kay-Zee:
@ramtinms you've had some recent commits touching this, though i think this was originally Phillips work. Any point keeping this? I feel like there isn't but figured it's worth checking with you as well

ramtinms (approved):

ramtinms:
don't think we need to maintain it, continuous loader is a good substitute in this case

SaveTheRbtz:
bors merge

PR #2520, opened 2022-05-31
chore(tests): cleanup TxTracker

SaveTheRbtz:
This diff refactors TxTracker.  Main changes are:
* Fix panic due to wg.Done counter underflow. This may happen if transaction was executed and then tracking timed out, hence calling multiple callbacks.
* Fix panic due to concurent close and send on a `txs` channel. 
* Encapsulate callbacks into the `txCallbacks` structure for self-description.
* Tracker now returns a channel that is closed after transaction is removed from tracking for any reason.
* Do not leak connections on tracker Stop.
* Properly cleanup inflight transactions on Stop.
* Misc logging cleanups.

Depends on: #2519


integration/utils/txTracker.go
```diff
@@ -44,168 +57,178 @@ func NewTxTracker(
...
-						tx.stat.isExpired = true
-						stats.AddTxStats(tx.stat)
-						log.Warn().Int("worker_id", workerID).Str("tx_id", tx.txID.String()).
-							Msgf("tx has been expired in %v seconds", time.Since(tx.createdAt).Seconds())
-						continue
+					tx.stat.TTS = time.Since(tx.createdAt)
+					txt.stats.AddTxStats(tx.stat)
+					close(tx.wait)
+					continue
+				case flowsdk.TransactionStatusUnknown:
```
Kay-Zee, line 204: It's also possible that we start in an "unknown" status, so not sure if it makes sense to throw it into expired right away.

though this is probably impossible for a localnet with only 1 access node

Kay-Zee (approved):
Just one comment about the "unknown" status

integration/utils/txTracker.go
SaveTheRbtz: Interesting, this will make a difference for benchnet.  I would not complicate this stat machine more though... let me retry the `TransactionStatusUnknown` state until either result or timeout -- this will remove one of the states and make the code cleaner.

PR #2523, opened 2022-05-31
[v0.25] Exit validation early for empty collections

jordanschalm:
For empty collections, the reference height range was not set correctly, which resulted in checking for duplicates in all cluster blocks. This PR exits validation early for empty collections.

### DIFF
--- a/state/cluster/badger/mutator.go
+++ b/state/cluster/badger/mutator.go
@@ -7,7 +7,6 @@ import (
 	"math"
 
 	"github.com/dgraph-io/badger/v2"
-
 	"github.com/onflow/flow-go/model/cluster"
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/module"
@@ -120,7 +119,14 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// no validation of transactions is necessary for empty transactions
+		if payload.Collection.Len() == 0 {
+			return nil
+		}
+
 		// check that all transactions within the collection are valid
+		// keep track of the min/max reference blocks - the collection must be non-empty
+		// at this point so these are guaranteed to be set correctly
 		minRefID := flow.ZeroID
 		minRefHeight := uint64(math.MaxUint64)
 		maxRefHeight := uint64(0)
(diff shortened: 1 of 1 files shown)


state/cluster/badger/mutator.go
```diff
@@ -120,7 +119,14 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// no validation of transactions is necessary for empty transactions
```
jordanschalm, line 122:
```suggestion
		// no validation of transactions is necessary for empty collections
```

state/cluster/badger/mutator.go
```diff
@@ -120,7 +120,14 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// no validation of transactions is necessary for empty collections
+		if payload.Collection.Len() == 0 {
+			return nil
```
zhangchiqing, line 125: Good catch!

But we still need to validate `payload.ReferenceBlockID` exist

jordanschalm: I think it is actually fine to skip that for empty collections, since they are not sent to consensus nodes. We have an issue to define and check a specific value for empty collections' reference block ID: https://github.com/dapperlabs/flow-go/issues/6245. But since the field is never checked currently, I think we can leave that to be addressed with #6245

engine/consensus/compliance/engine.go
```diff
@@ -179,8 +179,15 @@ func (e *Engine) Ready() <-chan struct{} {
 		e.unit.Launch(e.finalizationProcessingLoop)
 
 		ctx, cancel := context.WithCancel(context.Background())
-		signalerCtx, _ := irrecoverable.WithSignaler(ctx)
+		signalerCtx, hotstuffErrChan := irrecoverable.WithSignaler(ctx)
```
zhangchiqing, line 182: Nice catch. 

I wonder how did you catch this? Is it because the empty collection somehow causing hotstuff to fail?
```suggestion
		signalerCtx, hotstuffFatalErrChan := irrecoverable.WithSignaler(ctx)
```

jordanschalm: Exactly - for the reason you pointed out before: https://github.com/onflow/flow-go/pull/2523#discussion_r887161191

zhangchiqing: Oh I see.

zhangchiqing (approved):

state/cluster/badger/mutator.go
```diff
@@ -120,7 +120,25 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// a valid collection must reference a valid reference block
+		// NOTE: it is valid for a collection to be expired at this point,
+		// otherwise we would compromise liveness of the cluster.
+		refBlock, err := m.headers.ByBlockID(payload.ReferenceBlockID)
+		if errors.Is(err, storage.ErrNotFound) {
```
zhangchiqing, line 127: I wonder if makes sense to move it even earlier. 

Why would a node receive a payload with unknown `ReferenceBlockID`? It's probably because the collection is up to date with the cluster chain, but behind the main chain, right?

I wonder if it's a `InvalidExtension`, which is a bit misleading, since we can't distinguish between:
1. we are behind the mainchain (Unverifiable)
2. ReferenceBlockID is an unknown block.  (Invalid)

jordanschalm: Agreed, creating a new sentinel error for that case, like `UnverifiableInputError`, is also part of the other issue: https://github.com/dapperlabs/flow-go/issues/6245.

PR #2524, opened 2022-05-31
Exit validation early for empty collections

jordanschalm:
For empty collections, the reference height range was not set correctly, which resulted in checking for duplicates in all cluster blocks. This PR exits validation early for empty collections.

⚠️  This PR also adds a mechanism for irrecoverable HotStuff errors, which currently are silently dropped. An issue to implement a more robust solution to this problem is in https://github.com/dapperlabs/flow-go/issues/6263.

Notes: https://www.notion.so/dapperlabs/Transaction-De-Dupe-Deployment-Notes-8027bde5183a4e4c931bb1d42905952a


state/cluster/badger/mutator.go
```diff
@@ -120,7 +119,14 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// no validation of transactions is necessary for empty transactions
```
jordanschalm, line 122:
```suggestion
		// no validation of transactions is necessary for empty collections
```

zhangchiqing (approved):

durkmurder (approved):

zhangchiqing:
bors retry

state/cluster/badger/mutator.go
```diff
@@ -120,7 +120,25 @@ func (m *MutableState) Extend(block *cluster.Block) error {
 		checkTxsSpan, _ := m.tracer.StartSpanFromContext(ctx, trace.COLClusterStateMutatorExtendCheckTransactionsValid)
 		defer checkTxsSpan.Finish()
 
+		// a valid collection must reference a valid reference block
+		// NOTE: it is valid for a collection to be expired at this point,
+		// otherwise we would compromise liveness of the cluster.
+		refBlock, err := m.headers.ByBlockID(payload.ReferenceBlockID)
+		if errors.Is(err, storage.ErrNotFound) {
+			return state.NewInvalidExtensionErrorf("unknown reference block (id=%x)", payload.ReferenceBlockID)
```
AlexHentschel, line 128: In my opinion, this does not always mean that the block/collection is invalid. In my mind, invalid means that some protocol rule has been violated and the proposer of this collection should be slashed. 
However, this error could also occur if this collector is behind on following the main consensus. This is a very different error case. It only means that we can't validate correctness of the collection; as opposed to concluding the collection is invalid.

PR #2527, opened 2022-06-01
Amlandeep/fix an script logging implementation

koko1123:
Adding documentation and cleaning up implementation of the script execution path in AN


koko1123:
Already fixed

PR #228, opened 2020-12-09
Add flag to disable script executions on EN

psiemens:
This PR adds a `scripts-enabled` flag to the execution node that can be toggled to disabled calls to the `ExecuteScriptAtBlockID` and `GetAccountAtBlockID` RPC methods.

PR #2528, opened 2022-06-01
Fix implementation of script logging

koko1123:
Adds documentation for md5 hash use as well as eliminating custom structure for lookup
also eliminates the need for [this structure](https://github.com/onflow/flow-go/issues/228) in the AN codebase


AlexHentschel (approved):
Good documentation, clear code. Very nice. Thanks for the cleanup ✨.

peterargue (approved):

PR #2531, opened 2022-06-01
[FVM] Fix error handling on contract function invocations

janezpodhostnik:
This PR changes the following FVM environment functions to more properly handle errors that could happen during the execution of the direct contract function invocation:
- GetStorageCapacity
- GetAccountBalance
- GetAccountAvailableBalance

The comment on these functions was:

> // TODO: Figure out how to handle this error. Currently if a runtime error occurs, available balance will be 0.
	// 1. An error will occur if user has removed their FlowToken.Vault -- should this be allowed?
	// 2. Any other error indicates a bug in our implementation. How can we reliably check the Cadence error?

1. We could previously not return an error from these functions as the cadence interface did not allow it, this has since changed
2. This core contract change [Make methods that get balance safe in the case the account has no default vault](https://github.com/onflow/flow-core-contracts/pull/289) makes it so that the script wont panic but will return 0.0
3. yes removing the default vault is allowed, but the transaction will fail because of fee deduction failing. (You could perhaps remove the default vault and swap it with some other vault in the same transaction)
4. by removing the panic if the default vault is not present, any other error is unexpected and should be passed forward

There is another concern in the comments:

> 	// 2. There will also be an error in case the accounts balance times megabytesPerFlow constant overflows,
	//		which shouldn't happen unless the the price of storage is reduced at least 100 fold

This was fixed a while ago by using saturating arithmetics in that cadence function.


fvm/accounts_test.go
```diff
@@ -1248,7 +1248,30 @@ func TestAccountBalanceFields(t *testing.T) {
 
 				assert.NoError(t, err)
 
-				assert.Equal(t, cadence.UFix64(1_0000_0000), script.Value)
+				assert.Equal(t, cadence.UFix64(100_000_000), script.Value)
+			}),
+	)
+
+	t.Run("Get balance fails for accounts that dont exist",
```
m4ksio, line 1255: 👍

m4ksio (approved):

m4ksio:
Looks good!
However we need one more test - running a script which uses `.getBalance` on a nonexisting block/trie. This should check different path than tests for a missing account.

janezpodhostnik:
@m4ksio I wasn't sure how to make a test like that... do you know of a test that does something similar (uses a non-existent trie)

smnzhu (approved):

janezpodhostnik:
Converting to draft, since there is still a difference between account not existing and non-existing block

ramtinms (approved):
looks good to me

janezpodhostnik:
bors merge

PR #2535, opened 2022-06-01
Batch removal

m4ksio:
Used by rollback executed height cmd utility.
Adds ability to remove certain items in batch


zhangchiqing (commented):
Curious why bother adding batch remove? Removing items should be pretty fast already, no?

m4ksio:
> Curious why bother adding batch remove? Removing items should be pretty fast already, no?

Not really. Removing larger ranges takes hours.

AlexHentschel (commented):
I noticed that you are introducing several new methods, without documenting what error returns are expected during normal execution. Please see our [Guidelines for Flow Core Protocol](https://github.com/onflow/flow-go/blob/a9748ffe23d830fed2ea61912078337e56fd5a46/CodingConventions.md#error-handling) for best practices. Please include at least error documentation for the methods you are adding. It would be great if you could also invest a bit of time to reduce our technical debt by adding error documentation for some of the existing methods (not all, but please leave the code with less tech debt than you found it with). Thanks a lot.

m4ksio:
> I noticed that you are introducing several new methods, without documenting what error returns are expected during normal execution. Please see our [Guidelines for Flow Core Protocol](https://github.com/onflow/flow-go/blob/a9748ffe23d830fed2ea61912078337e56fd5a46/CodingConventions.md#error-handling) for best practices. Please include at least error documentation for the methods you are adding. It would be great if you could also invest a bit of time to reduce our technical debt by adding error documentation for some of the existing methods (not all, but please leave the code with less tech debt than you found it with). Thanks a lot.

Entirety of this PR adds extra methods to a storage layer. I can't find any logic here which wouldn't be a boilerplate really, and any errors which does more than just bubbling up badger ones. 
It would be helpful if you could provide the basic example of what kind of documentation you would expect for that.

storage/badger/operation/common.go
```diff
@@ -133,6 +133,16 @@ func remove(key []byte) func(*badger.Txn) error {
 	}
 }
 
+func batchRemove(key []byte) func(writeBatch *badger.WriteBatch) error {
+	return func(writeBatch *badger.WriteBatch) error {
+		err := writeBatch.Delete(key)
+		if err != nil {
```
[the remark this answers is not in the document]
m4ksio: I think it should be find, but I couldn't find a clear documentation confirming that.
I therefore enhanced the test to simply call removal twice, second case should test non-exiting data

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height.go
```diff
@@ -180,6 +188,7 @@ func removeExecutionResultsFromHeight(
 
 // removeForBlockID remove block execution related data for a given block.
```
zhangchiqing, line 189:
```suggestion
// removeForBlockID remove block execution related data for a given block. 
// All data to be removed will be removed in a batch write.
```

storage/badger/operation/events.go
```diff
@@ -52,10 +52,23 @@ func RemoveServiceEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) err
 	return removeByPrefix(makePrefix(codeServiceEvent, blockID))
 }
 
+func BatchRemoveServiceEventsByBlockID(blockID flow.Identifier, batch *badger.WriteBatch) func(*badger.Txn) error {
```
zhangchiqing, line 55:
```suggestion
// BatchRemoveServiceEventsByBlockID removes all service events for the given blockID.
// It returns nil, if all service events are removed
// It returns nil, if no service events was found for the given blockID
// It returns error, if it runs into any exception
func BatchRemoveServiceEventsByBlockID(blockID flow.Identifier, batch *badger.WriteBatch) func(*badger.Txn) error {
```

storage/badger/operation/events.go
```diff
@@ -52,10 +52,23 @@ func RemoveServiceEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) err
...
+	return func(txn *badger.Txn) error {
+		return batchRemoveByPrefix(makePrefix(codeServiceEvent, blockID))(txn, batch)
+	}
+}
+
 func RemoveEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) error {
 	return removeByPrefix(makePrefix(codeEvent, blockID))
 }
 
+func BatchRemoveEventsByBlockID(blockID flow.Identifier, batch *badger.WriteBatch) func(*badger.Txn) error {
```
zhangchiqing, line 65:
```suggestion
// BatchRemoveEventsByBlockID removes all events for the given blockID.
// It returns nil, if all events are removed
// It returns nil, if no event was found for the given blockID
// It returns error, if it runs into any exception
func BatchRemoveEventsByBlockID(blockID flow.Identifier, batch *badger.WriteBatch) func(*badger.Txn) error {
```

storage/badger/operation/common.go
```diff
@@ -133,6 +133,16 @@ func remove(key []byte) func(*badger.Txn) error {
 	}
 }
 
+func batchRemove(key []byte) func(writeBatch *badger.WriteBatch) error {
```
zhangchiqing, line 136:
```suggestion
// batchRemove removes the given key in a batch.
// It returns nil if the key is successfully removed
// It returns nil if the key does not exist
// It returns error if running into any exception
func batchRemove(key []byte) func(writeBatch *badger.WriteBatch) error {
```

Could you add a test case to verify the following case:
```
// It returns nil if the key does not exist
```

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height.go
```diff
@@ -214,7 +223,7 @@ func removeForBlockID(
 		}
 
 		// remove chunkID-blockID index
-		err = headers.RemoveChunkBlockIndexByChunkID(chunkID)
+		err = headers.BatchRemoveChunkBlockIndexByChunkID(chunkID, writeBatch)
```
zhangchiqing, line 226: Sometimes, an EN might not have the chunk block index for the given chunk ID. 

If we call `RemoveChunkBlockIndexByChunkID `, then it will return `storage.ErrNotFound` error, and since we want to continue removing other data associated with the block, we handle the `storage.ErrNotFound` specifically, and continue with the next chunk.

How does it work with the BatchWrite? would `BatchRemoveChunkBlockIndexByChunkID` also return NotFound?
It's better to have a test case to understand the batchRemove's behavior when no key is found. 

If `BatchRemoveChunkBlockIndexByChunkID` never returns NotFound error, we could remove the "if" case below.

zhangchiqing (commented):
I added some code comments in the newly added functions as example. 

We'd better to explicit about what error could possibly return, for instance, if it might return `storge.ErrNotFound` as expected normal error, then better to mention it in the comment.

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height.go
m4ksio: Yes, we will never return `storage.ErrNotFound` in this code, I'm removing the test case

storage/badger/operation/chunkDataPacks.go
```diff
@@ -17,6 +17,11 @@ func BatchInsertChunkDataPack(c *badgermodel.StoredChunkDataPack) func(batch *ba
 	return batchWrite(makePrefix(codeChunkDataPack, c.ChunkID), c)
 }
 
+// BatchRemoveChunkDataPack removes a chunk data pack keyed by chunk ID in a batch
+func BatchRemoveChunkDataPack(chunkID flow.Identifier) func(batch *badger.WriteBatch) error {
```
AlexHentschel, line 21: Would it be possible to add a comment on expected error returns? I know it is complicated because we return a function ... maybe we could write something like "the returned functions errors with xyz in case ..."

storage/badger/operation/commits.go
```diff
@@ -33,3 +33,8 @@ func LookupStateCommitment(blockID flow.Identifier, commit *flow.StateCommitment
 func RemoveStateCommitment(blockID flow.Identifier) func(*badger.Txn) error {
 	return remove(makePrefix(codeCommit, blockID))
 }
+
+// BatchRemoveStateCommitment batch removes the state commitment by block ID
+func BatchRemoveStateCommitment(blockID flow.Identifier) func(batch *badger.WriteBatch) error {
```
AlexHentschel, line 38: similarly here: some documentation when the returned function might error would be great!

storage/badger/operation/common.go
```diff
@@ -196,6 +209,26 @@ func removeByPrefix(prefix []byte) func(*badger.Txn) error {
 	}
 }
 
+func batchRemoveByPrefix(prefix []byte) func(tx *badger.Txn, writeBatch *badger.WriteBatch) error {
```
AlexHentschel, line 212: Leo [added a suggestion for goDoc](https://github.com/onflow/flow-go/pull/2535#discussion_r960112736)? The comment was resolved (just unresolved it) but still not documentation (?) 😬

storage/badger/operation/events.go
```diff
@@ -52,10 +52,29 @@ func RemoveServiceEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) err
 	return removeByPrefix(makePrefix(codeServiceEvent, blockID))
 }
 
+// BatchRemoveServiceEventsByBlockID removes all service events for the given blockID.
+// It returns nil, if all service events are removed (or no service events were found)
+// It returns error, if it runs into any exception
```
AlexHentschel, line 57: what errors specifically are expected? What happens if I hand in a block ID that does not exist?

storage/badger/operation/events.go
```diff
@@ -52,10 +52,29 @@ func RemoveServiceEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) err
...
+	}
+}
+
 func RemoveEventsByBlockID(blockID flow.Identifier) func(*badger.Txn) error {
 	return removeByPrefix(makePrefix(codeEvent, blockID))
 }
 
+// BatchRemoveEventsByBlockID removes all events for the given blockID.
+// It returns nil, if all events are removed (or no events were found)
+// It returns error, if it runs into any exception
```
AlexHentschel, line 70: what errors specifically are expected? What happens if I hand in a block ID that does not exist?

storage/badger/results.go
```diff
@@ -158,3 +158,8 @@ func (r *ExecutionResults) ByBlockID(blockID flow.Identifier) (*flow.ExecutionRe
 func (r *ExecutionResults) RemoveIndexByBlockID(blockID flow.Identifier) error {
 	return r.db.Update(operation.SkipNonExist(operation.RemoveExecutionResultIndex(blockID)))
 }
+
+func (r *ExecutionResults) BatchRemoveIndexByBlockID(blockID flow.Identifier, batch storage.BatchStorage) error {
```
AlexHentschel, line 162: As an exported function, I would appreciate some minimal goDoc.

storage/badger/chunkDataPacks.go
```diff
@@ -79,6 +79,14 @@ func (ch *ChunkDataPacks) BatchStore(c *flow.ChunkDataPack, batch storage.BatchS
 	return operation.BatchInsertChunkDataPack(sc)(writeBatch)
 }
 
+func (ch *ChunkDataPacks) BatchRemove(chunkID flow.Identifier, batch storage.BatchStorage) error {
```
AlexHentschel, line 82: As an exported function, I would appreciate some minimal goDoc.

storage/badger/commits.go
```diff
@@ -79,3 +79,8 @@ func (c *Commits) ByBlockID(blockID flow.Identifier) (flow.StateCommitment, erro
 func (c *Commits) RemoveByBlockID(blockID flow.Identifier) error {
 	return c.db.Update(operation.SkipNonExist(operation.RemoveStateCommitment(blockID)))
 }
+
+func (c *Commits) BatchRemoveByBlockID(blockID flow.Identifier, batch storage.BatchStorage) error {
```
AlexHentschel, line 83: As an exported function, I would appreciate some minimal goDoc. Including expected error returns.

storage/badger/headers.go
```diff
@@ -196,6 +196,11 @@ func (h *Headers) RemoveChunkBlockIndexByChunkID(chunkID flow.Identifier) error
 	return h.db.Update(operation.RemoveBlockIDByChunkID(chunkID))
 }
 
+func (h *Headers) BatchRemoveChunkBlockIndexByChunkID(chunkID flow.Identifier, batch storage.BatchStorage) error {
```
AlexHentschel, line 199: As an exported function, I would appreciate some minimal goDoc. Including expected error returns.

AlexHentschel (commented):
Nice clean code. 

I'd like to request the following:
1. Overall, I miss documentation of exported methods, in particular error returns. Would really appreciate if we could add such documentation. You [state](https://github.com/onflow/flow-go/pull/2535#issuecomment-1195243218): 
  > Entirety of this PR adds extra methods to a storage layer. I can't find any logic here which wouldn't be a boilerplate really, and any errors which does more than just bubbling up badger ones.

   I think Leo has a very good point. Because as it stands right now, any engineer working with your methods has to go through the implementation (multiple levels) and figure out what badger errors are potentially expected. This increases the hurdle for other engineers to do their error handling properly, incentivizing them to also skip errors. 
   * I think it would be _very beneficial_ to explicitly list the badger errors that would be returned for plausible inputs. For an example, please see [my comment below](https://github.com/onflow/flow-go/pull/2535#discussion_r1035442944).

2.  For many of the storage implementations, we export the `BatchRemove` methods. But they are not part of the interfaces. Is there a specific reason? (if so, documentation in the code would be great). 
4. the documentation of `BatchStorage` is below minimal: https://github.com/onflow/flow-go/blob/795d4b31145987c5bb7e6c977ed177824e3dfe73/storage/batch.go#L9-L10 Just came across this interface while looking at this PR, and the documentation is not sufficient for me to even get a rough idea what this interface is intended to be used for.

storage/badger/operation/events.go
m4ksio: Added clarification

storage/badger/operation/events.go
m4ksio: Added clarification

storage/badger/operation/common.go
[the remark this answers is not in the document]
m4ksio: https://github.com/onflow/flow-go/pull/2535/files#diff-74818b43e4000c8ac3a548b6dbbff230416a9dbd9622a7a95e7994072d72a84cR117

storage/badger/operation/common.go
m4ksio: Added

AlexHentschel (approved):
Thanks Maks for the great work. Your PR has evolved to an exemplary piece of clean, well-documented work 👏. For example, I really like this error documentation:  https://github.com/onflow/flow-go/blob/ac727a97e91c4c7d1d20cff1ea168530422dd9a4/storage/events.go#L32-L35 it concisely describes that errors are not expected and what will be returned in case of an unexpected failure. That makes it very straight forward for engineers to built on your code and hopefully encourages them to follow your example with proper documentation. Also appreciate you [adding the documentation](https://github.com/onflow/flow-go/blob/ac727a97e91c4c7d1d20cff1ea168530422dd9a4/storage/badger/events.go#L166-L169) to the implementation 💚.

There is one minor detail that is outside of the code touched by your PR, but it would be great if your could just fix here: 
https://github.com/onflow/flow-go/blob/ac727a97e91c4c7d1d20cff1ea168530422dd9a4/storage/badger/transaction_results.go#L204-L205
The godoc still used an outdated method name ~ByBlockIDIndex~ which should be updated to `ByBlockIDTransactionIndex`. Thanks.

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height_test.go
```diff
@@ -95,8 +95,27 @@ func TestReExecuteBlock(t *testing.T) {
...
+			chunkDataPacks,
+			myReceipts,
+			events,
+			serviceEvents,
+			header.ID(),
+		)
+
+		require.NoError(t, err)
+
+		// remove again, to make sure missing entires are handled properly
```
zhangchiqing, line 116: To ensure missing entries are handled properly, here we make the tests to remove the result twice and flush once. But I don't think the actual code would be run like this. More likely that we will run the command line twice to remove results for the same block, which means we will run remove and flush twice for the same block, instead of remove the result twice and flush once. 

So it's better adjust the tests accordingly:
```
batch := bstorage.NewBatch(db)
err := removeForBlockID(batch, ...)
require.NoError(t, err)

err = batch.Flush()
require.NoError(t, err)

// remove again for the same block, it should not return any error if the results
// have been removed already.
batch2 := bstorage.NewBatch(db)
err := removeForBlockID(batch2, ...)
require.NoError(t, err)

err = batch2.Flush()
require.NoError(t, err)
```

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height_test.go
```diff
@@ -218,7 +243,10 @@ func TestReExecuteBlockWithDifferentResult(t *testing.T) {
 			header.ID(),
 		)
 
+		err2 := batch.Flush()
```
zhangchiqing, line 246: same comment here, better to check remove and flush twice instead of remove twice and flush once.

zhangchiqing (approved):
Looks good. Just a small comment on the tests.

cmd/util/cmd/rollback-executed-height/cmd/rollback_executed_height_test.go
m4ksio: Interesting idea, I have made it test both - so remove twice, flush and than remove again. This should cover all cases.

PR #2539, opened 2022-06-02
[Fix] add explicit mapping of cluster prefix -> message codes

kc1116:
This PR fixes missing message codes sent on cluster prefixed channels  causing false-negative validations in the libP2P authorized sender validator.

### DIFF
--- a/network/codec/cbor/msg_channel_mapping.go
+++ b/network/codec/cbor/msg_channel_mapping.go
@@ -55,4 +55,8 @@ func initializeChannelToMsgCodesMap() {
 
 	// dkg
 	ChannelToMsgCodes[channels.DKGCommittee] = []uint8{CodeDKGMessage}
+
+	// cluster based channels
+	ChannelToMsgCodes[channels.SyncClusterPrefix] = []uint8{CodeSyncRequest, CodeSyncResponse, CodeRangeRequest, CodeBatchRequest, CodeBlockResponse}
+	ChannelToMsgCodes[channels.ConsensusClusterPrefix] = []uint8{CodeClusterBlockProposal, CodeClusterBlockVote, CodeClusterBlockResponse}
 }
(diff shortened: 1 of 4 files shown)


network/codec/cbor/msg_channel_mapping.go
```diff
@@ -55,4 +55,8 @@ func initializeChannelToMsgCodesMap() {
 
 	// dkg
 	ChannelToMsgCodes[channels.DKGCommittee] = []uint8{CodeDKGMessage}
+
+	// cluster based channels
+	ChannelToMsgCodes[channels.SyncClusterPrefix] = []uint8{CodeSyncRequest, CodeSyncResponse, CodeRangeRequest, CodeBatchRequest, CodeBlockResponse}
```
yhassanzadeh13, line 60: Is it fine to have `CodeRangeRequest` for more than one channel? same applies to other message codes as well. If it is should not be the case, please fix it.

yhassanzadeh13 (approved):
Except for the two comments made, the rest LGTM.

network/p2p/topic_validator_test.go
```diff
@@ -423,7 +423,7 @@ func TestAuthorizedSenderValidator_ClusterChannel(t *testing.T) {
 	timedCtx, cancel5s := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel5s()
 	// create a dummy sync request to publish from our LN node
-	data := getMsgFixtureBz(t, &messages.SyncRequest{})
+	data := getMsgFixtureBz(t, &messages.RangeRequest{})
```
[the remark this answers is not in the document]
kc1116: Yes that's what I was thinking as well , I will add the test after this is merged.

jordanschalm (approved):

network/codec/cbor/msg_channel_mapping.go
kc1116: It is okay because sync-committee and sync-cluster use the same message codes unlike the consensus-cluster which has it's own CodeClusterBlockVote etc message codes. I am not sure why we don't also have CodeClusterSyncRequest etc for sync-cluster.

Issue #2275, opened 2022-04-07
importing contracts is counted as read every time they are read not once


bjartek:
## 🐞 Bug Report

When reading an imported contract TotalBytesRead is incremented every time this is done, even if the same contract is read twice in the same transaction. 

So given the following scenario:

```
A imports B
A imports C
B imports C
```

Then C is counted as read twice.  

## What is the severity of this bug?

**Should have:** * It would be better if we do this before we ship, but it's OK if we don't (product functions without this, but it's a better user experience). Consider adding to a future sprint.

## Expected behaviour

I would expect C to be count as read once.

turbolent:
@janezpodhostnik @ramtinms

bjartek:
Today I removed the "B imports C" part in my code and then I am below the limit again. C in my case here is FIND and that contract is 50k.

j1010001:
Duoplicate of https://github.com/onflow/cadence/issues/1684

PR #2540, opened 2022-06-02
Don't double count contract imports towards execution limits

dsainati1:
Fixes https://github.com/onflow/flow-go/issues/2275 and https://github.com/onflow/cadence/issues/1684

When loading a contract that we have already loaded during a transaction, we still meter computation and interactions even though we are not actually loading anything or performing any additional computation. This should ensure that we only meter when loading a contract for the first time during a transaction or script.


fvm/fvm_test.go
```diff
@@ -4371,3 +4371,101 @@ func TestScriptAccountKeyMutationsFailure(t *testing.T) {
...
+					pub fun main() {}
+				`, account1.String(), account2.String(), account3.String())))
+
+				scriptCtx := fvm.NewContextFromParent(ctx)
+
+				err = vm.Run(scriptCtx, script, view, ps)
+				require.NoError(t, err)
+				require.NoError(t, script.Err)
+				// we should meter computation for 3 GetPrograms, one for each contract
+				assert.Equal(t, int(script.GasUsed), 900)
```
dsainati1, line 4467: @janezpodhostnik before adding the cache skipping, this is 1400 instead of 900. I am not sure however why it would be 900 here or what the value actually should be in the ideal situation.

dsainati1: After some investigating; the interpreter itself calls get program in addition to the checker to enforce that imports are properly loaded. However these should still be skipped by electing not to merge the state, unless we are adding cost to the state of B for the load of A, and then when we load B from C, for example, it counts the cost of both B and A when loading C?

dsainati1: Ok after some more thought about this over the weekend I think just skipping merge after the first load of each program is not sufficient if we want to only report 300 units for this test. Here is a slightly simpler example:
```
pub contract A {}
```
```
import A from 0x1
pub contract B {}
```
```
import A from 0x1
import B from 0x2
pub fun main() {}
```
Executing the script here with the changes in this PR does the following computation metering and meter merges:
* GetProgram called for check of `import A` in script, increasing the meter of the script to 100
* SetProgram for A is called, state of checking `A` is merged into state of checking script, increasing no metering because no calls to GetProgram were made while checking A. 
* GetProgram called for check of `import B` in script, increasing the metering of the script to 200
* GetProgram called for check of `import A` in `B`, increasing the metering of B to 100
* SetProgram for B is called, state of checking `B` is merged into state of checking script, increasing metering of the script to 300
* GetProgram called for interpret of `import A` in script, increasing the metering of the script to 400
* GetProgram called for interpret of `import B` in script , increasing the metering of the script to 500

Because the merging occurs during the calls to SetProgram, the import of `A` that gets metered during the check of `B` is added to its meter, which is then merged into the meter of the script, even though we have already merged `A` into the script already.

fvm/transactionEnv.go
```diff
@@ -648,7 +643,15 @@ func (e *TransactionEnv) GetProgram(location common.Location) (*interpreter.Prog
 		}
 	}
 
-	program, has := e.programs.Get(location)
+	program, cached, has := e.programs.Get(location)
+
+	if !cached {
```
dsainati1, line 648: This additional check is what we need to get the metered computation down to the 300 that the test expects. However, we need to also make sure to not double count memory metering as well as register touches; basically everything that gets merged. Is there a way to generalize this?

dsainati1: Relatedly, what is the best way to access data like the number of interactions from within the tests?

fvm/handler/programs.go
```diff
@@ -22,19 +22,21 @@ type stackEntry struct {
...
 // views must be merged in order to make sure they are recorded
 type ProgramsHandler struct {
-	masterState  *state.StateHolder
-	viewsStack   []stackEntry
-	Programs     *programs.Programs
-	initialState *state.State
+	masterState    *state.StateHolder
+	viewsStack     []stackEntry
+	Programs       *programs.Programs
+	LoadedPrograms map[common.LocationID]interface{}
```
ramtinms, line 28: ⚠️ this is not safe and might cause execution fork, this can only happen through the state holder

fvm/transactionEnv.go
ramtinms: This is a bit controversial topic, I think the design is if we do touch anything we should measure it, and we might need to fix it by some cache on higher levels, if cadence runtime ask for it, we should measure it, and maybe we can do a cache on cadence runtime to not request for what has been loaded ?

fvm/handler/programs.go
ramtinms: I need to think more about it

PR #2542, opened 2022-06-02
feat(localnet): pass through tracing, extensive tracing, and cadence tracing flags

SaveTheRbtz:
While here, also fix the profiler flags.

### DIFF
--- a/integration/localnet/Makefile
+++ b/integration/localnet/Makefile
@@ -8,9 +8,14 @@ NCLUSTERS=1
 EPOCHLEN=10000   # 0 means use default
 STAKINGLEN=2000 # 0 means use default
 DKGLEN=2000     # 0 means use default
-PROFILER=false
 CONSENSUS_DELAY=800ms
 COLLECTION_DELAY=950ms
+
+PROFILER=false
+TRACING=true
+EXTENSIVE_TRACING=false
+CADENCE_TRACING=false
+
 # The Git commit hash
 COMMIT=$(shell git rev-parse HEAD)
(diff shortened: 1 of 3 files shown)


integration/localnet/bootstrap.go
```diff
@@ -42,6 +42,9 @@ const (
 	DefaultObserverCount     = 0
 	DefaultNClusters         = 1
 	DefaultProfiler          = false
+	DefaultTracing           = true
```
zhangchiqing, line 45: I wonder if makes sense to have default tracing as false.

integration/localnet/Makefile
```diff
@@ -8,9 +8,14 @@ NCLUSTERS=1
 EPOCHLEN=10000   # 0 means use default
 STAKINGLEN=2000 # 0 means use default
 DKGLEN=2000     # 0 means use default
-PROFILER=false
 CONSENSUS_DELAY=800ms
 COLLECTION_DELAY=950ms
+
+PROFILER=false
+TRACING=true
```
zhangchiqing, line 15: same question here as well, better to make tracing false by default

integration/localnet/bootstrap.go
SaveTheRbtz: This is a tricky one: i'm of course is quite biassed given that my workflow always includes traces.  If people do not use them on localnet, then I'm happy to switch the default.

FWIW, currently, it is hardcoded to `true` for localnet.

zhangchiqing (approved):

PR #2370, opened 2022-05-03
[2066] Setup CI pipeline and tools to build Observer binary and docker containers

szegedim:
This is our candidate to be merged into master first. It ensures that any observer changes we have runs like the one in master. It checks whether the observer localnet container runs with the right settings. +@Pradeep Palanisamy the other author.
Once it is merged, we plan to give you bite sized pull requests every day. This should not overwhelm you. It will help to catch up with our branches quickly.

Authors: Miklos Szegedi, Pradeep Palanisamy

PR #2547, opened 2022-06-03
Draft - 2420 -changes for migrating BlockHeader api out of access API

koli19761:
Initial changes for migrating BlockHeader api out of access API


access/legacy/handler.go
```diff
@@ -15,12 +15,14 @@ import (
 
 type Handler struct {
 	api   access.API
+	psapi access.PROTOCOL_STATE_API
```
koko1123, line 18: You dont' need to change the legacy handler! this is only for the new "observer"! https://github.com/onflow/flow-go/pull/2370/files#diff-c396d8ee377761a43fca284545cb4ee535f868ea2b16db01db81a21741855ed8R1-R28

koko1123 (changes requested):
I'm not sure I've explained this correctly, these changes belong in the new engine!

engine/access/rest/accounts.go
```diff
@@ -7,15 +7,15 @@ import (
 )
 
 // GetAccount handler retrieves account by address and returns the response
-func GetAccount(r *request.Request, backend access.API, link models.LinkGenerator) (interface{}, error) {
+func GetAccount(r *request.Request, backend access.API, psapi access.PROTOCOL_STATE_API, link models.LinkGenerator) (interface{}, error) {
```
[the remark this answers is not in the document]
koli19761: Ok , i will modify that

access/legacy/handler.go
koli19761: Sure will make the change

PR #2548, opened 2022-06-03
sync engine optimization

danielholmes839:
## Optimizations

Implemented two ideas from: https://github.com/onflow/flow-go/blob/master/flips/sync-protocol.md

1. Added a height parameter to `RequestBlock` which allows for more efficient pruning
2. Added a `localFinalizedHeight` attribute which is used to skip pruning in `ScanPending`

## Metrics

Implemented metrics for the sync core [here](https://github.com/onflow/flow-go/blob/dan/5934-sync-core/module/metrics/chainsync.go). The metrics could not be put inside module/metrics.go and the module/metrics/ package because of a cyclical import.


module/synchronization/core.go
```diff
@@ -165,8 +169,11 @@ func (c *Core) ScanPending(final *flow.Header) ([]flow.Range, []flow.Batch) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	// prune before doing any work
-	c.prune(final)
+	// prune if the current height is less than the new height
+	if c.localFinalizedHeight < int(final.Height) {
+		c.prune(final)
+		c.localFinalizedHeight = int(final.Height)
+	}
```
zhangchiqing, line 176: is `prune` possibly being called by other method? If yes, it's better to move this check into the prune function.

module/synchronization/core.go
```diff
@@ -124,7 +128,7 @@ func (c *Core) HandleHeight(final *flow.Header, height uint64) {
 	}
 }
 
-func (c *Core) RequestBlock(blockID flow.Identifier) {
+func (c *Core) RequestBlock(blockID flow.Identifier, height uint64) {
```
zhangchiqing, line 131: better to add comment on both arguments

module/synchronization/core.go
```diff
@@ -206,15 +213,17 @@ func (c *Core) queueByHeight(height uint64) {
 
 // queueByBlockID queues a request for a block by block ID, only if no
 // equivalent request has been queued before.
-func (c *Core) queueByBlockID(blockID flow.Identifier) {
+func (c *Core) queueByBlockID(blockID flow.Identifier, height uint64) {
```
zhangchiqing, line 216: with `localFinalizedHeight`, we will only keep requesting blocks that are above the finalized height. So we could also drop blocks that is below or equal to `localFinalizedHeight`.

module/synchronization/core.go
```diff
@@ -51,19 +51,23 @@ func DefaultConfig() Config {
...
-		log:      log.With().Str("module", "synchronization").Logger(),
-		Config:   config,
-		heights:  make(map[uint64]*Status),
-		blockIDs: make(map[flow.Identifier]*Status),
+		log:                  log.With().Str("module", "synchronization").Logger(),
+		Config:               config,
+		heights:              make(map[uint64]*Status),
+		blockIDs:             make(map[flow.Identifier]*Status),
+		metrics:              metrics,
+		localFinalizedHeight: -1, // used in ScanPending
```
jordanschalm, line 70: Could we use `uint64` (same type as `Header.Height`) and initialize it to `0` instead?

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+
+func NewMetricsCollector() *MetricsCollector {
+	return &MetricsCollector{
+		timeToPruned: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:      "time_to_pruned",
+			Namespace: namespaceSynchronization,
+			Subsystem: subsystemSyncCore,
+			Help:      "the time between queueing and pruning a block in milliseconds",
+			Buckets:   []float64{100, 250, 500, 1000, 2500, 5000, 7500, 10000, 20000},
+		}, []string{"status", "requested_by"}),
```
jordanschalm, line 58: There are some conventions around Prometheus metric units and naming we try to follow which we can apply here:

>  metric names should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers)

> should have a suffix describing the unit, in plural form. Note that an accumulating count has total as a suffix, in addition to the unit if applicable.

The full doc is [here](https://prometheus.io/docs/practices/naming/)

Example of updating to the unit convention:
```suggestion
		timeToPruned: prometheus.NewHistogramVec(prometheus.HistogramOpts{
			Name:      "time_to_pruned_seconds",
			Namespace: namespaceSynchronization,
			Subsystem: subsystemSyncCore,
			Help:      "the time between queueing and pruning a block in seconds",
			Buckets:   []float64{.1, .25, .5, 1, 2.5, 5, 7.5, 10, 20},
		}, []string{"status", "requested_by"}),
```
(can apply to the other metrics as well)

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+		timeToReceived: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:      "time_to_received",
+			Namespace: namespaceSynchronization,
+			Subsystem: subsystemSyncCore,
+			Help:      "the time between queueing and receiving a block in milliseconds",
+			Buckets:   []float64{100, 250, 500, 1000, 2500, 5000, 7500, 10000, 20000},
+		}, []string{"requested_by"}),
+		totalPruned: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name:      "total_blocks_pruned",
```
jordanschalm, line 67: Example of updating to the unitless counter convention:
```suggestion
			Name:      "blocks_pruned_total",
```
(can apply to the other metrics as well)

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+			Namespace: namespaceSynchronization,
+			Subsystem: subsystemSyncCore,
+			Help:      "the total number of blocks pruned by 'id' or 'height'",
+		}, []string{"requested_by"}),
+		storedByHeight: prometheus.NewGauge(prometheus.GaugeOpts{
```
jordanschalm, line 72: Would it make sense to combine this with `storedById` and use a `requested_by` label, like we do for eg. `totalPruned`?

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+			Name:      "blocks_stored_by_id",
+			Namespace: namespaceSynchronization,
+			Subsystem: subsystemSyncCore,
+			Help:      "the number of blocks currently stored by id",
+		}),
+		totalHeightsRequested: prometheus.NewCounter(prometheus.CounterOpts{
+			Name:      "total_heights_requested",
+			Namespace: namespaceSynchronization,
+			Subsystem: subsystemSyncCore,
+			Help:      "the total number of blocks requested by height. range of 0-10 would increase the counter by 10",
```
jordanschalm, line 88:
```suggestion
			Help:      "the total number of blocks requested by height, including retried requests for the same heights. Eg: a range of 1-10 would increase the counter by 10",
```
Adding a note that this includes retries, and fixing an off-by-one (ranges are inclusive on both boundaries)

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	s.totalPruned.With(prometheus.Labels{"requested_by": "height"}).Add(float64(totalByHeight))
+
+	// update gauges
+	s.storedById.Set(float64(storedById))
+	s.storedByHeight.Set(float64(storedByHeight))
+
+}
+
+func (s *MetricsCollector) RangeRequested(ran flow.Range) {
+	s.totalHeightsRequested.Add(float64(ran.To - ran.From))
```
jordanschalm, line 133:
```suggestion
	s.totalHeightsRequested.Add(float64(ran.To - ran.From+1))
```
Ranges are inclusive of both `To` and `From`

module/synchronization/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+type SynchronizationMetrics interface {
+	// record pruned blocks. requested and received times might be zero values
+	PrunedBlockById(status *Status)
+
+	PrunedBlockByHeight(status *Status)
```
jordanschalm, line 16: My understanding of why this metrics collector can't live in the `module/metrics` package is because it is depending on `Status`, which is defined in `module/synchronization`, creating the circular dependency. 

I suggest that we move the `Status` model to a separate model package so that we can maintain a consistent organization of the metric collectors. For example:
* Create a package `model/chainsync` (for context, @smnzhu has a pending [PR](https://github.com/onflow/flow-go/pull/2292) which changes the name of this module to `chainsync` -- we can preemptively name the model package accordingly)
* Add `Range`, `Batch`, `Status` to `model/chainsync`
* Then, we can move `SynchronizationMetrics` (can also preemptively rename this to `ChainSyncMetrics`) to `module/metrics.go and the implementation to the `module/metrics` package

module/synchronization/core.go
```diff
@@ -206,15 +213,17 @@ func (c *Core) queueByHeight(height uint64) {
...
 
 	// only queue the request if have never queued it before
 	if c.blockIDs[blockID].WasQueued() {
 		return
 	}
 
 	// queue the request
-	c.blockIDs[blockID] = NewQueuedStatus()
+	status := NewQueuedStatus()
+	status.Header = &flow.Header{Height: height}
```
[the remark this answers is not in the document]
jordanschalm: Agreed, I'd suggest to instead add a `BlockHeight` field to `Status` which is required to create a new status with `NewQueuedStatus`

engine/collection/compliance/core_test.go
```diff
@@ -149,7 +149,7 @@ func (cs *ComplianceCoreSuite) SetupTest() {
 
 	// set up synchronization module mock
 	cs.sync = &module.BlockRequester{}
-	cs.sync.On("RequestBlock", mock.Anything).Return(nil)
+	cs.sync.On("RequestBlock", mock.Anything, mock.Anything).Return(nil)
```
smnzhu, line 152: nit: coinsider using `mock.AnythingOfType` here.

engine/collection/compliance/core_test.go
```diff
@@ -378,7 +378,7 @@ func (cs *ComplianceCoreSuite) TestProposalBufferingOrder() {
 	for _, block := range proposals {
 
 		// check that we request the ancestor block each time
-		cs.sync.On("RequestBlock", mock.Anything).Once().Run(
+		cs.sync.On("RequestBlock", mock.Anything, mock.Anything).Once().Run(
```
smnzhu, line 381: same here

module/synchronization/core.go
```diff
@@ -237,31 +254,39 @@ func (c *Core) getRequestStatus(height uint64, blockID flow.Identifier) *Status
...
-			header := status.Header
-
-			if header.Height <= final.Height {
-				delete(c.blockIDs, blockID)
-				continue
-			}
+		if status.BlockHeight <= final.Height {
+			delete(c.blockIDs, blockID)
+			c.metrics.PrunedBlockById(status)
+			continue
```
smnzhu, line 279: nit: this `continue` is not necessary here

smnzhu (approved):

module/synchronization/core.go
```diff
@@ -237,31 +254,39 @@ func (c *Core) getRequestStatus(height uint64, blockID flow.Identifier) *Status
...
+
+	c.localFinalizedHeight = final.Height
 
 	// track how many statuses we are pruning
 	initialHeights := len(c.heights)
 	initialBlockIDs := len(c.blockIDs)
 
-	for height := range c.heights {
+	for height, status := range c.heights {
 		if height <= final.Height {
```
zhangchiqing, line 268: Some idea for future optimization: this is a O(N) loop, and later there is another O(N) loop over c.blockIDs. 

Can we do better here?

module/synchronization/core.go
```diff
@@ -356,7 +382,8 @@ func (c *Core) RangeRequested(ran flow.Range) {
 
 // BatchRequested updates status state for a batch of block IDs that has been
 // successfully requested. Must be called when a batch request is submitted.
-func (c *Core) BatchRequested(batch flow.Batch) {
+func (c *Core) BatchRequested(batch chainsync.Batch) {
+	c.metrics.BatchRequested(batch)
 	c.mu.Lock()
 	defer c.mu.Unlock()
```
zhangchiqing, line 388: better to move after the lock
```suggestion
	c.mu.Lock()
	defer c.mu.Unlock()
	c.metrics.BatchRequested(batch)
```

module/synchronization/core.go
```diff
@@ -340,7 +365,8 @@ func (c *Core) getRequestableItems() ([]uint64, []flow.Identifier) {
 
 // RangeRequested updates status state for a range of block heights that has
 // been successfully requested. Must be called when a range request is submitted.
-func (c *Core) RangeRequested(ran flow.Range) {
+func (c *Core) RangeRequested(ran chainsync.Range) {
+	c.metrics.RangeRequested(ran)
```
zhangchiqing, line 369: same here

module/metrics/chainsync.go
```diff
@@ -0,0 +1,102 @@
...
+	c.totalPruned.With(prometheus.Labels{"requested_by": "id"}).Add(float64(totalById))
+	c.totalPruned.With(prometheus.Labels{"requested_by": "height"}).Add(float64(totalByHeight))
+
+	// update gauges
+	c.storedBlocks.With(prometheus.Labels{"requested_by": "id"}).Set(float64(storedById))
+	c.storedBlocks.With(prometheus.Labels{"requested_by": "height"}).Set(float64(storedByHeight))
+}
+
+func (c *ChainSyncCollector) RangeRequested(ran chainsync.Range) {
+	c.totalHeightsRequested.Add(float64(ran.To - ran.From + 1))
```
zhangchiqing, line 97: can we add a `Len` method to `Range` type ?

zhangchiqing (approved):
Looks good.

module/synchronization/core.go
danielholmes839: I think if we added a min priority queue and kept our map for looking up blocks by height/id in O(1). We could change these loops to O(mlogn) where m is the number of blocks we actually need to prune which should be small assuming we're slowly increasing the finalized height and n is the number of blocks we're storing. 

We should see the number of blocks we're actually storing with the new metrics and be able to figure out if it would be worth doing.

PR #2549, opened 2022-06-03
Fix underflowing of memory metrics difference values

dsainati1:
Given two `uint64` values `x` and `y`, `float64(x-y)` will underflow the value of `x-y` when `x` is smaller, meaning that this will never produce a negative float. In the case of the memory metrics, this meant when we would have reported an underestimate, we instead produce an absurdly large positive value. This changes the expression to be `float64(x) - float64(y)` so that we can properly produce a negative number.


robert-e-davidson3 (approved):

ramtinms (approved):

turbolent (approved):

Kay-Zee (approved):

Kay-Zee:
bors merge

PR #2551, opened 2022-06-03
[Protocol State] Index seal by sealed block id

zhangchiqing:
Problem:
 Given a sealed height, how to get the result of the sealed block?

This is not easy, because we are missing an index. 
However, there are scenarios that need this query:

Scenario 1: execution node verifies its own result against sealed result. 
When EN is behind executing blocks or re-executing blocks, it might execute sealed blocks. When finish executing them, it could verify the result against sealed result, so that if they don't match, it should just stop executing rather than continue building on its fork. 
However, this can't be done without the above query

Scenario 2: execution state sync downloads state for block as soon as it's sealed. However, it can't find the result for the sealed block. 

This PR adds an index for seals by the sealed block ID, so that we can query the index to find the seal, and using the result ID in the seal to find the sealed result.


utils/unittest/epoch_builder.go
```diff
@@ -376,10 +376,11 @@ func (builder *EpochBuilder) addBlock(block *flow.Block) {
 		err := state.Extend(context.Background(), block)
 		require.NoError(builder.t, err)
 
-		err = state.Finalize(context.Background(), blockID)
-		require.NoError(builder.t, err)
 		err = state.MarkValid(blockID)
 		require.NoError(builder.t, err)
+
+		err = state.Finalize(context.Background(), blockID)
```
zhangchiqing, line 382: We should `MakeValid` before `Finalize` a block

utils/unittest/epoch_builder.go
```diff
@@ -395,14 +396,13 @@ func (builder *EpochBuilder) AddBlocksWithSeals(n int, counter uint64) *EpochBui
...
 		// - execution result for B
-		a := builder.blocks[len(builder.blocks)-2]
 		b := builder.blocks[len(builder.blocks)-1]
 
 		receiptB := ReceiptForBlockFixture(b)
 
 		block := BlockWithParentFixture(b.Header)
 		seal := Seal.Fixture(
-			Seal.WithResult(a.Payload.Results[0]),
+			Seal.WithResult(b.Payload.Results[0]),
```
zhangchiqing, line 405: This was a bug, we don't need last 2 blocks, because the result in the last block is already pointing at the last second block. We should always seal the result in the last block.

model/flow/sealing_segment.go
```diff
@@ -274,46 +279,57 @@ func (builder *SealingSegmentBuilder) hasValidSeal() bool {
...
 // isValidRootSegment will check that the block in the root segment has a view of 0.
 func (builder *SealingSegmentBuilder) isValidRootSegment() bool {
-	return len(builder.blocks) == rootSegmentBlocksLen && builder.highest().Header.View == rootSegmentBlockView
+	return len(builder.blocks) == rootSegmentBlocksLen &&
+		builder.highest().Header.View == rootSegmentBlockView &&
+		len(builder.results) == rootSegmentBlocksLen &&
+		builder.firstSeal != nil &&
+		builder.results[0].BlockID == builder.blocks[0].ID() && // result matches the block
+		builder.results[0].ID() == builder.firstSeal.ResultID && // seal matches the result
+		builder.results[0].BlockID == builder.firstSeal.BlockID // seal seals the block
```
zhangchiqing, line 308: Adding a bit more checks

model/flow/sealing_segment.go
```diff
@@ -63,6 +63,9 @@ type SealingSegment struct {
 	// sealed state, when the first block contains no seal.
 	// If the first block in the segment contains seal, then this field is `nil`
 	FirstSeal *Seal
+
+	// Seal is the seal for the first block (the lowest height) in the segment
+	Seal *Seal
```
zhangchiqing, line 68: I had to add Seal to sealing segment, because we want to make sure the root block 's seal is also indexed. Since the root block' seal is created by a sealing segment, I had to ensure the sealing segment's first block is sealed, and has the seal. This can be guaranteed, since the `validateSegment` call will look for the seal. So once `validateSegment` finds the seal for the first block, we just add it here to guarantee a SealingSegment always have a Seal.

model/flow/sealing_segment.go
```diff
@@ -274,46 +279,57 @@ func (builder *SealingSegmentBuilder) hasValidSeal() bool {
 		for _, seal := range block.Payload.Seals {
 			// if we found the latest seal, confirm it seals lowest
 			if seal.ID() == latestSealID {
-				return seal.BlockID == lowestID
+				if seal.BlockID == lowestID {
+					return seal, true
```
zhangchiqing, line 283: validate not only validates the first block is sealed, but also return the seal, so that we can attach it to the created SealingSegment, for it to be used in building the index.

state/protocol/inmem/snapshot.go
```diff
@@ -49,6 +49,10 @@ func (s Snapshot) SealedResult() (*flow.ExecutionResult, *flow.Seal, error) {
 	return s.enc.LatestResult, s.enc.LatestSeal, nil
 }
 
+func (s Snapshot) Seal() (*flow.Seal, error) {
+	return s.enc.SealingSegment.Seal, nil
```
zhangchiqing, line 53: I'm not quite sure about this. 

What is the SealingSegment for a snapshot? Is the SealingSegment's first block equal to the snapshot's block? @jordanschalm 

If yes, then I'm right here, that if this snapshot is sealed, then the sealingsegment 's seal is the seal for the snapshot's block

AlexHentschel (commented):
The approach looks good. Thanks for adding tests and improving the goDoc for the existing code.

There are two high-level areas where I would like to request improvements:
1. While the naming of the `Seals.ByBlockID` method was previously not very descriptive, I feel this issue is now amplified by adding a second method for retrieving seals. I have similar concerns for some of the lower-level methods. 
   * I added detailed comments and suggest for renaming the methods and some fields (including already existing code).
   * Maybe it would be beneficial to keep the names of the existing methods as they are for this PR. And then we rename them in a subsequent PR. I think this might simplify reviews to separate functional changes (this PR, touching a limited number of files with functional changes) from renaming (follow-up PR, touching many files, but no functional changes).
2. When talking to external people, we often simplify "Block X has a finalized seal" to "Block X is sealed". However, to accurately describe the mechanics of the protocol at a detailed level, I feel we need to be more precise:
   * From my perspective, "a block being sealed" is a property of a particular fork:
Sealing Here, block D contains a seal for block A. In the fork of D, the sealed result for A is now locked in. Any execution results included in descendants of D _must_ be derived from the sealed result. All subsequent seals _must_ extend the execution that was sealed. 
    * In different forks, different results for A can be sealed.  There are protocol edge cases, where it is possible that different forks seal different results for the same block. Example:  in one fork, the block was computed with a missing collection attestation; in the other later fork, the collection showed up.
    *  Finalization is a subsequent step, which commits the finalized fork (including the seals in the finalized block).
    
  What you are introducing here is a mapping from `finalized block X` -> `finalized seal for block X`.  Precision in our internal terminology is important to accurately represent the protocol mechanics with the necessary details. Here, the important details is the _finality_:
    * The mapping from 1 block to 1 seal is only possible if the seal is finalized. 
    * Without finality of the seal, we would need a 1-block-to-many-seals mapping. Which is very different than the logic you are adding here. 

  I have added comments with concrete suggestions for renaming methods, variables, as well as suggestions for improving the goDoc. Hope this helps to clarify the importance of finality and hopefully prevent (reduce) accidental misuse by newer engineers 🤞

state/protocol/badger/state.go
```diff
@@ -259,6 +259,10 @@ func (state *State) bootstrapStatePointers(root protocol.Snapshot) func(*badger.
 		if err != nil {
 			return fmt.Errorf("could not insert sealed height: %w", err)
 		}
+		err = operation.IndexBySealedBlockID(segment.Seal.ID(), segment.Seal.BlockID)(tx)
```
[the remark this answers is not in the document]
zhangchiqing: Actually, I think what I did is right, and could work for both root snapshot and non-root snapshot.

Note: for both cases, we will finalize and seal exactly one block, rather than multiple blocks. In other words, we only need to index for the lowest block height, rather than each seal in the payload of the highest block.

Let's walk though each of them. 
1. for root snapshot like this [R], we only finalize and seal one block, and the same block. 
2. for non-root snapshot, let's look at a happy path first: `A <- B (seal_X) <- C (seal_Y)<- D (seal_A)`, in this case, we set `D` as [the finalized height as well as the root height](https://github.com/onflow/flow-go/blob/b02e05c7e2e5c4c2bf8ee2a8de00d9128ae3c645/state/protocol/badger/state.go#L250-L254). And then we set `A` as the sealed height. 
Note: we didn't finalize A, B, C, but only D. and we didn't seal B, C, D yet. So we only finalize and seal one block.
Note: the sealed height is lower than the root height.  (side question, this should be OK, right? no one is supposed to query state for a height that is lower than root height)

3. let's look at another case where the last block has multiple seals: `A <- B (seal_X) <- C (seal_Y) <- D(seal_Z, seal_A)`. in this case, we do exactly the same as case `2`, finalize D and seal A, which exactly finalize and seal one block.
Note: this case is not possible: `A <- B (seal_X) <- C (seal_Y) <- (seal_A, seal_B)`. because this is an invalid sealing segment. The highest seal which is seal_B does not match with the lowest block which is A. 

Since we only seal one block, and is the lowest block, then it makes sense to index seal for the lowest block. And since one of the future blocks will have a seal for block B (the next unsealed block), and once that block is finalized, then we will index seal_B by block B. 

I also [listed the cases in comments for valid and invalid segments](https://github.com/onflow/flow-go/pull/2551/files#diff-04d7b3c448f8446615a26c7bc5ebe1ace7a6d4f4105cd5ea42ae8832cdc61767R267-R272), can go through that to validate my assert.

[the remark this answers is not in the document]
AlexHentschel: Thanks Leo for the comment.
You are stating the the following 
A <- B (seal_X) <- C (seal_Y) <- (seal_A, seal_B)
would not be a valid sealing segment. To the best of my understanding, Supporting this case is necessary. But we can figure this out independently of this PR. 

Essentially, you are saying that the sealing segment can only seal one block. Ok. Then I think we should remove the extra 'Seal' field which you add in this PR. Because the sealing segment already specifies a single seal, which you argue is the only one. So no need to add another one.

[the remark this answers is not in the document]
AlexHentschel: > Thanks Leo for the comment.
> You are stating the the following 
> A <- B (seal_X) <- C (seal_Y) <- (seal_A, seal_B)
> would not be a valid sealing segment. To the best of my understanding, Supporting this case is necessary. But we can figure this out independently of this PR. 
> 
> Essentially, you are saying that the sealing segment can only seal one block. Ok. Then I think we should remove the extra 'Seal' field which you add in this PR. Because the sealing segment already specifies a single seal, which you argue is the only one. So no need to add another one.

model/flow/sealing_segment.go
AlexHentschel: I don't understand why this is necessary. To the best of my understanding, 'FirstSeal' should be sufficient. 
In your later comment, you argue that a sealing segment can only seal one block; I thing the seal for this one block is 'FirstSeal' right (?)

jordanschalm: I don't think the new `Seal` field is necessary either. We already are guaranteed to have the seal for the lowest block in the sealing segment:
* If we have a root sealing segment, then `FirstSeal` is set and it is the seal for the lowest (only) block in the sealing segment by definition
* If we have a non-root sealing segment, then by definition, the seal for the lowest block must be included in the highest block (or its ancestor within the sealing segment)

state/protocol/inmem/snapshot.go
AlexHentschel: I also have mixed feeling about this. Originally, the Snapshot was supposed to be immutable. If retrieving a `Snapshot` instance is successful, then it should never change. The problem with the return value for `Seal()` is that it is eventually consistent. I am hoping that we could use https://github.com/onflow/flow-go/blob/b69fa33a25ed981e2ba77a33d3329cf0bb4e6fc7/storage/seals.go#L10-L24 instead

utils/unittest/epoch_builder.go
zhangchiqing: The result is in the previous block's payload, the seal is in the current block's payload. The seal and receipt are not in the same block.

If we seal the result 2 blocks away, we are sealing a result that is already being sealed. 

This wasn't caught until I added the index when a block is sealed in this PR, the error says a block is sealed twice, after debugging I found this bug here.

zhangchiqing: It seems we create put a receipt and seal in the same payload, but the receipt's result is in the previous block. 

Why not putting receipt and result into the same block? The previous blocks were like this. For instance[ block G's payload has result and receipt for block F](https://github.com/onflow/flow-go/blob/3104f4a591072c9ed5a43fbd5767a1466ee89a93/utils/unittest/epoch_builder.go#L297-L298)

state/protocol/inmem/snapshot.go
zhangchiqing: >In general, I would prefer we avoid adding these kinds of methods to the Snapshot API 

Yeah, but the `Seal` method is just the same as the `QuorumCertificate` method, which is some data in future blocks.

Based on what Jordan mentioned, I think what I implemented is wrong. The `Seal` method should return the seal for the snapshot's reference block, but what I implemented is the last seal as of the snapshot's reference block.

```
A <- B <- C <- D (seal_A) <- E <- F <- G (seal_D)
```

If block D is the snapshot's reference block, then we should return `seal_D`, but my implementation is returning `seal_A`. 

As Alex suggested, we could choose to use seals storage interface if we don't like protocol's API and don't have a good way to add it.

state/protocol/badger/state.go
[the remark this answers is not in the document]
jordanschalm: A sealing segment for a reference block `D` is defined as a segment of blocks such that the latest sealed block w.r.t. `D` is the lowest block in the segment ([reference docs](https://github.com/onflow/flow-go/blob/3104f4a591072c9ed5a43fbd5767a1466ee89a93/model/flow/sealing_segment.go#L15-L18)).

```
A <- B (seal_X) <- C (seal_Y) <- D(seal_A, seal_B)
```
In this example, the latest sealed block w.r.t. `D` is `B`, not `A`, therefore:
* `A<-B<-C<-D` is not a valid sealing segment
* `B<-C<-D` is a valid sealing segment (the only valid sealing for reference block D)

utils/unittest/epoch_builder.go
AlexHentschel: ## ⚠️ 

> we cannot include a receipt and seal for the same block

Correct; including a seal in the same block that also includes the result is _impossible_. This is because:
*  We first need to publish a block incorporating the result. 
* Then, we need a child block in addition for the source of randomness.
* The source of randomness of the block that incorporates the result defines the verification node assignment

Hence, the block that incorporates the result must be published, otherwise, verifiers don't know which chunks to check.

https://www.notion.so/dapperlabs/How-ConsensusNode-choose-seals-and-receipts-to-include-in-the-next-block-payload-72c178bccd4a416d8cd4380af4cfb6be

state/protocol/inmem/snapshot.go
smnzhu: @zhangchiqing I think we should use the seals storage interface, because the usecase we have that originally motivated this PR is the one where given a block, we want to find its seal.

smnzhu: Adding the `Seal()` method to the snapshot interface doesn't help satisfy this usecase.

smnzhu (changes requested):

state/protocol/inmem/snapshot.go
jordanschalm: > Yeah, but the Seal method is just the same as the QuorumCertificate method, which is some data in future blocks.

You're right. The difference to me is that a QC is a necessary part of the snapshot which must be present in order to bootstrap a protocol state instance, whereas the seal is not. 

Some further thoughts:
* I think it would be reasonable to restrict the API so that you can only access a complete snapshot for a block with a valid QC (we need to do this for non-consensus nodes anyway, because we don't know that a block payload is valid until we observe the QC).
  * that way, we can include `QuorumCertificate` in the `Snapshot` API and maintain the property that a snapshot is immutable over time
  * however, there would need to be a carve-out for the HotStuff committee module to be able to query the `Identities` for a block without a validated QC
* Restricting the API in the same way to only sealed blocks is quite a bit more restrictive

admin/commands/storage/read_seals.go
```diff
@@ -71,7 +71,7 @@ func (r *ReadSealsCommand) Handler(ctx context.Context, req *admin.CommandReques
 		if err != nil {
 			return nil, fmt.Errorf("failed to retrieve index for block %#v: %w", blockID, err)
 		}
-		lastSeal, err := r.seals.ByBlockID(blockID)
+		lastSeal, err := r.seals.HighestInFork(blockID)
```
zhangchiqing, line 74: This PR took the chance to rename this function in order to make it more clear as per the review suggestion

model/flow/sealing_segment.go
```diff
@@ -73,6 +73,34 @@ func (segment *SealingSegment) Lowest() *Block {
 	return segment.Blocks[0]
 }
 
+// FinalizedSeal returns the seal that seals the lowest block
+// For a valid SealingSegment, it's guaranteed to be found
+func (segment *SealingSegment) FinalizedSeal() (*Seal, error) {
```
zhangchiqing, line 78: This method get the seal for the lowest block of the segment, it is needed because the bootstrapping process needs to add the index for the sealed block.

model/flow/sealing_segment.go
```diff
@@ -256,41 +284,63 @@ func (builder *SealingSegmentBuilder) isValidHeight(block *Block) bool {
...
+// A <- B <- C <- D(seal_A,seal_B)        ==> invalid, because latest seal is B, but lowest block is A
+// A <- B <- C <- D(seal_X,seal_A)        ==> valid, because it's OK for block X to be unknown
+// A <- B <- C <- D(seal_A) <- E(seal_B)  ==> invalid, because latest seal is B, but lowest block is A
+// A(seal_A)                              ==> invalid, because this is impossible for non-root sealing segments
+func (builder *SealingSegmentBuilder) hasValidSeal() error {
+	_, err := hasValidSeal(builder.lowest().ID(), builder.highest().ID(), builder.latestSeals, builder.blocks)
+	return err
+}
 
+func hasValidSeal(lowestID, highestID Identifier, latestSeals map[Identifier]Identifier, blocks []*Block) (*Seal, error) {
```
zhangchiqing, line 303: I had to extract this function in order to reuse the logic in `FinalizedSeal`.

utils/unittest/epoch_builder.go
zhangchiqing: I think my fix is correct here. 

What we have is a chain with each block contains a result for previous and a seal for previous of previous block:
```
A <- B (result_A) <- C (result_B, seal_A) <- D(resultC, seal_B)
```

And we want to add similar blocks like this. 

But the tests was creating a wrong seal. It is supposed to seal its grandparent block (height - 2), instead it seals the result in its grandparent block, which is its grand-grandparent block (height - 3), causing the grand-grantparent to be sealed again, which is causing the tests to fail due to inserting the seal index twice.

My fix is to say seal the result in parent block, which is the result for its grandparent block.

AlexHentschel: Thanks for the example, that really helped to understand why your fix is correct. 👍

state/protocol/badger/mutator.go
```diff
@@ -648,13 +648,13 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 			return fmt.Errorf("could not update sealed height: %w", err)
 		}
 
-		// when a block is finalized, each seal in it is the seal for its sealed block.
-		// indexing each seal for its sealed block so that they can be used to find the seal result
-		// by sealed height.
+		// When a block is finalized, we commit the result for each seal it contains. The sealing logic
+		// guarantees that only as single, continuous execution fork is sealed. Here, we index for
```
smnzhu, line 652:
```suggestion
		// guarantees that only a single, continuous execution fork is sealed. Here, we index for
```

model/flow/sealing_segment.go
```diff
@@ -256,41 +284,63 @@ func (builder *SealingSegmentBuilder) isValidHeight(block *Block) bool {
 	return block.Header.Height == builder.highest().Header.Height+1
 }
 
-// hasValidSeal returns true if the latest seal as of highest is for lowest.
+// hasValidSeal checks if the latest seal as of highest block is for lowest block
+// it returns the latest seal and nil if found
```
jordanschalm, line 288:
```suggestion
// it returns the latest seal and no error if found
```

smnzhu (approved):

state/protocol/badger/mutator_test.go
```diff
@@ -123,6 +124,128 @@ func TestExtendValid(t *testing.T) {
...
+		err = state.Extend(context.Background(), b1)
+		require.NoError(t, err)
+
+		// block 2(result B1)
+		b1Receipt := unittest.ReceiptForBlockFixture(b1)
+		b2 := unittest.BlockWithParentFixture(b1.Header)
+		b2.SetPayload(flow.Payload{
+			Receipts: []*flow.ExecutionReceiptMeta{b1Receipt.Meta()},
+			Results:  []*flow.ExecutionResult{&b1Receipt.ExecutionResult},
+		})
```
jordanschalm, line 151:
```suggestion
		b2.SetPayload(unittest.PayloadFixture(unittest.WithReceipts(b1Receipt)))
```
Noting the fixture helpers we have for this

state/protocol/badger/mutator_test.go
```diff
@@ -123,6 +124,128 @@ func TestExtendValid(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// block 7
+		b7 := unittest.BlockWithParentFixture(b6.Header)
+		b7.SetPayload(flow.EmptyPayload())
+		err = state.Extend(context.Background(), b7)
+		require.NoError(t, err)
+
+		// finalizing b1 - b3
+		// when B4 is finalized, can only find seal for G
```
jordanschalm, line 198:
```suggestion
		// finalizing b1 - b4
		// when B4 is finalized, can only find seal for G
```

state/protocol/badger/state.go
```diff
@@ -235,6 +235,11 @@ func (state *State) bootstrapStatePointers(root protocol.Snapshot) func(*badger.
 		}
 		highest := segment.Highest()
 		lowest := segment.Lowest()
+		// find the finalize seal that seals the lowest block, meaning seal.BlockID == lowest.ID()
```
jordanschalm, line 238:
```suggestion
		// find the finalized seal that seals the lowest block, meaning seal.BlockID == lowest.ID()
```

storage/badger/operation/seals.go
```diff
@@ -38,14 +38,30 @@ func LookupPayloadResults(blockID flow.Identifier, resultIDs *[]flow.Identifier)
...
 
-func LookupBlockSeal(blockID flow.Identifier, sealID *flow.Identifier) func(*badger.Txn) error {
+// LookupLatestSealAtBlock finds the highest seal that was included in the fork up to (and including) blockID.
+// In most cases, it is the highest seal included in this block's payload. However, if there are no
+// seals in this block, sealID should reference the highest seal in blockID's ancestor.
+func LookupLatestSealAtBlock(blockID flow.Identifier, sealID *flow.Identifier) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeBlockToSeal, blockID), &sealID)
 }
 
+// IndexFinalizedSealByBlockID index a the _finalized_ seal by the sealed block ID.
```
jordanschalm, line 55:
```suggestion
// IndexFinalizedSealByBlockID index a the _finalized_ seal by the sealed block ID.
// A <- B <- C(SealA)
// Example: when block C is finalized, we create the index `A.ID->SealA.ID`
```

storage/badger/operation/seals.go
```diff
@@ -38,14 +38,30 @@ func LookupPayloadResults(blockID flow.Identifier, resultIDs *[]flow.Identifier)
 	return retrieve(makePrefix(codePayloadResults, blockID), resultIDs)
 }
 
-func IndexBlockSeal(blockID flow.Identifier, sealID flow.Identifier) func(*badger.Txn) error {
+// IndexLatestSealAtBlock persists the highest seal that was included in the fork up to (and including) blockID.
+// In most cases, it is the highest seal included in this block's payload. However, if there are no
+// seals in this block, sealID should reference the highest seal in blockID's ancestor.
+func IndexLatestSealAtBlock(blockID flow.Identifier, sealID flow.Identifier) func(*badger.Txn) error {
 	return insert(makePrefix(codeBlockToSeal, blockID), sealID)
```
jordanschalm, line 45:
```suggestion
	return insert(makePrefix(codeBlockIDToLatestSealID, blockID), sealID)
```
Could we update the prefix name for the existing index as well?

storage/badger/seals.go
```diff
@@ -71,9 +71,22 @@ func (s *Seals) ByID(sealID flow.Identifier) (*flow.Seal, error) {
...
 	var sealID flow.Identifier
-	err := s.db.View(operation.LookupBlockSeal(blockID, &sealID))
+	err := s.db.View(operation.LookupLatestSealAtBlock(blockID, &sealID))
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve seal for fork with head %x: %w", blockID, err)
+	}
+	return s.ByID(sealID)
+}
+
+func (s *Seals) FinalizedSealForBlock(blockID flow.Identifier) (*flow.Seal, error) {
```
jordanschalm, line 87:
```suggestion
// FinalizedSealForBlock returns the seal for the given block, only if that seal
// has been included in a finalized block.
// Returns storage.ErrNotFound if the block is unknown or unsealed.
func (s *Seals) FinalizedSealForBlock(blockID flow.Identifier) (*flow.Seal, error) {
```

storage/badger/seals_test.go
```diff
@@ -63,11 +63,37 @@ func TestSealIndexAndRetrieve(t *testing.T) {
...
+
+		expectedSeal := unittest.Seal.Fixture()
+		blockID := unittest.IdentifierFixture()
+		expectedSeal.BlockID = blockID
+
+		// store the seal first
+		err := store.Store(expectedSeal)
+		require.NoError(t, err)
+
+		// index the seal ID for the heighest sealed block in this fork
```
jordanschalm, line 91:
```suggestion
		// index the seal ID for the highest sealed block in this fork
```

jordanschalm (approved):
Thank you for the iteration on this. I left a few minor comments, but no blockers. 🎸

zhangchiqing:
bors merge

PR #2553, opened 2022-06-04
Handle events panic

m4ksio:
Handle event encoding properly 
Add elaborate test to ensure the behaviour


fvm/context.go
```diff
@@ -62,9 +60,9 @@ func newContext(ctx Context, opts ...Option) Context {
 const AccountKeyWeightThreshold = 1000
 
 const (
-	DefaultComputationLimit             = 100_000        // 100K
-	DefaultMemoryLimit                  = math.MaxUint64 //
-	DefaultEventCollectionByteSizeLimit = 256_000        // 256KB
+	DefaultComputationLimit             = 100_000       // 100KB
+	DefaultMemoryLimit                  = 1_000_000_000 // 1GB
```
j1010001, line 64: This commit (changing the memory limit) should not go to v0.26 branch, this is strictly for the pentest network only ATM.

j1010001 (changes requested):

PR #2556, opened 2022-06-06
BFT - Integration Testing - Docker fix - no such host error

gomisha:
There is a docker issue where running the BFT integration tests in CI causes an error where the docker containers can't reach the host, resulting in this error:

`Error while dialing dial tcp: lookup host.docker.internal on 127.0.0.11:53: no such host`

The test works locally (outside of CI) but needs to be fixed so it runs in CI.


gomisha:
This PR is superseded by a workaround that bypassed the Docker issue entirely (https://github.com/dapperlabs/flow-go/issues/6276). So there was no need to fix the Docker issue anymore.

PR #2557, opened 2022-06-06
[Consensus] `SafetyRules` extra test cases 

durkmurder:
### Context

Following comments in https://github.com/onflow/flow-go/pull/2444, I have extended test cases for `ProduceVote` and `ProduceTimeout` to cover every possible edge case.

### DIFF
--- a/consensus/hotstuff/safetyrules/safety_rules_test.go
+++ b/consensus/hotstuff/safetyrules/safety_rules_test.go
@@ -359,6 +359,19 @@ func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnUnsafeProposal() {
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), vote)
 	})
+	s.Run("happy-path-includes-last-view-tc", func() {
+		proposal := helper.MakeProposal(
+			helper.WithBlock(
+				helper.MakeBlock(
+					helper.WithParentBlock(s.bootstrapBlock),
+					helper.WithBlockView(s.bootstrapBlock.View+1))),
+			helper.WithLastViewTC(helper.MakeTC()))
+		s.committee.On("IdentityByBlock", proposal.Block.BlockID, proposal.Block.ProposerID).Return(s.proposerIdentity, nil).Maybe()
+		vote, err := s.safety.ProduceVote(proposal, proposal.Block.View)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), vote)
+	})
 
 	s.signer.AssertNotCalled(s.T(), "CreateVote")
 	s.persister.AssertNotCalled(s.T(), "PutSafetyData")
@@ -447,29 +460,62 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
 // Hence, the cases tested here would be symptoms of an internal bugs, and therefore should not result in an NoVoteError.
 func (s *SafetyRulesTestSuite) TestProduceTimeout_NotSafeToTimeout() {
 
+	s.Run("newest-qc-nil", func() {
+		// newestQC cannot be nil
+		timeout, err := s.safety.ProduceTimeout(s.safetyData.LockedOneChainView, nil, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("happy-path-includes-last-view-tc", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+
+		// tc not needed but included
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, helper.MakeTC())
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("last-view-tc-nil", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+
+		// tc needed but not included
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("last-view-tc-for-wrong-view", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		// lastViewTC should be for newestQC.View+1
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, lastViewTC)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
 	s.Run("highest-qc-below-locked-round", func() {
-		view := s.proposal.Block.View
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView - 1))
 
-		timeout, err := s.safety.ProduceTimeout(view, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, nil)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), timeout)
 	})
 	s.Run("cur-view-below-highest-acknowledged-view", func() {
-		view := s.safetyData.HighestAcknowledgedView
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		s.safetyData.HighestAcknowledgedView = newestQC.View + 10
 
-		timeout, err := s.safety.ProduceTimeout(view-2, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, nil)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), timeout)
 	})
 	s.Run("cur-view-below-highest-QC", func() {
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
-		view := newestQC.View
 
-		timeout, err := s.safety.ProduceTimeout(view, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, nil)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), timeout)
--- a/consensus/hotstuff/safetyrules/safety_rules_test.go
+++ b/consensus/hotstuff/safetyrules/safety_rules_test.go
@@ -495,6 +495,26 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_NotSafeToTimeout() {
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), timeout)
 	})
+	s.Run("cur-view-below-highest-QC", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View - 2))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View-1, newestQC, lastViewTC)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("last-view-tc-is-newer", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		// newest QC included in TC cannot be higher than the newest QC known to replica
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View+1),
+			helper.WithTCHighestQC(helper.MakeQC(helper.WithQCView(newestQC.View+1))))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, lastViewTC)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
 	s.Run("highest-qc-below-locked-round", func() {
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView - 1))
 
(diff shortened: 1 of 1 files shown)


consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -286,31 +286,20 @@ func (s *SafetyRulesTestSuite) TestProduceVote_PersistStateException() {
 
 // TestProduceVote_VotingOnUnsafeProposal tests different scenarios where we try to vote on unsafe blocks
 func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnUnsafeProposal() {
-	s.Run("invalid-block-view", func() {
-		// create block with block.View == block.QC.View
+
+	s.Run("happy-path-includes-last-view-tc", func() {
```
jordanschalm, line 290:
```suggestion
	// a proposal which includes a QC for the previous round should not contain a TC
	s.Run("proposal-includes-last-view-qc-and-tc", func() {
```
It wasn't clear to me why this was named "happy path"

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
 func (s *SafetyRulesTestSuite) TestProduceTimeout_NotSafeToTimeout() {
 
-	s.Run("highest-qc-below-locked-round", func() {
-		view := s.proposal.Block.View
-		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView - 1))
+	s.Run("newest-qc-nil", func() {
+		// newestQC cannot be nil
+		timeout, err := s.safety.ProduceTimeout(s.safetyData.LockedOneChainView, nil, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 490:
```suggestion
```
We never expect any sentinel from `ProduceTimeout`, per the [godoc](https://github.com/onflow/flow-go/blob/5f361374705d4565ce68b5a0ee2120ed83ceb047/consensus/hotstuff/safetyrules/safety_rules.go#L122). (We definitely wouldn't expect `NoVoteError`). So I don't think we need to check for absence of specific sentinels here.

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("happy-path-includes-last-view-tc", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
 
-		timeout, err := s.safety.ProduceTimeout(view, newestQC, nil)
+		// tc not needed but included
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, helper.MakeTC())
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 499:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
 	})
-	s.Run("cur-view-below-highest-acknowledged-view", func() {
-		view := s.safetyData.HighestAcknowledgedView
+	s.Run("last-view-tc-nil", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+
+		// tc needed but not included
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 508:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+	})
+	s.Run("last-view-tc-for-wrong-view", func() {
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		// lastViewTC should be for newestQC.View+1
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View))
 
-		timeout, err := s.safety.ProduceTimeout(view-2, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, lastViewTC)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 518:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
 		require.Nil(s.T(), timeout)
 	})
 	s.Run("cur-view-below-highest-QC", func() {
 		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
-		view := newestQC.View
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View - 2))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View-1, newestQC, lastViewTC)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 527:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+	})
+	s.Run("last-view-tc-is-newer", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		// newest QC included in TC cannot be higher than the newest QC known to replica
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View+1),
+			helper.WithTCHighestQC(helper.MakeQC(helper.WithQCView(newestQC.View+1))))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, lastViewTC)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 538:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("highest-qc-below-locked-round", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView - 1))
+
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 546:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+	})
+	s.Run("cur-view-below-highest-acknowledged-view", func() {
+		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView))
+		// modify highest acknowledged view in a way that it's definitely bigger than the newest QC view
+		s.safetyData.HighestAcknowledgedView = newestQC.View + 10
 
-		timeout, err := s.safety.ProduceTimeout(view, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+1, newestQC, nil)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
```
jordanschalm, line 556:
```suggestion
```

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
-		view := s.proposal.Block.View
-		newestQC := helper.MakeQC(helper.WithQCView(s.safetyData.LockedOneChainView - 1))
+	s.Run("newest-qc-nil", func() {
+		// newestQC cannot be nil
+		timeout, err := s.safety.ProduceTimeout(s.safetyData.LockedOneChainView, nil, nil)
+		require.Error(s.T(), err)
+		require.False(s.T(), model.IsNoVoteError(err))
+		require.Nil(s.T(), timeout)
+	})
+	s.Run("happy-path-includes-last-view-tc", func() {
```
jordanschalm, line 493:
```suggestion
	// if a QC for the previous view is provided, a last view TC is unnecessary and must not be provided
	s.Run("includes-last-view-qc-and-tc", func() {
```
Also not sure why this is named happy path, since we're testing an error condition

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -447,29 +483,75 @@ func (s *SafetyRulesTestSuite) TestProduceTimeout_ShouldTimeout() {
...
+		// lastViewTC should be for newestQC.View+1
+		lastViewTC := helper.MakeTC(helper.WithTCView(newestQC.View))
 
-		timeout, err := s.safety.ProduceTimeout(view-2, newestQC, nil)
+		timeout, err := s.safety.ProduceTimeout(newestQC.View+2, newestQC, lastViewTC)
 		require.Error(s.T(), err)
 		require.False(s.T(), model.IsNoVoteError(err))
 		require.Nil(s.T(), timeout)
 	})
 	s.Run("cur-view-below-highest-QC", func() {
```
jordanschalm, line 521: The boundary condition is `curView==highestQC.view`, we could test that in addition, or instead.

consensus/hotstuff/safetyrules/safety_rules_test.go
AlexHentschel: I think the interface specification is a bit outdated https://github.com/onflow/flow-go/blob/5f361374705d4565ce68b5a0ee2120ed83ceb047/consensus/hotstuff/safety_rules.go#L37-L43 
`NoTimeoutError` is no longer in the code base. Maybe we should just update the API and remove these checks?

consensus/hotstuff/safetyrules/safety_rules_test.go
```diff
@@ -286,31 +286,20 @@ func (s *SafetyRulesTestSuite) TestProduceVote_PersistStateException() {
 
 // TestProduceVote_VotingOnUnsafeProposal tests different scenarios where we try to vote on unsafe blocks
 func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnUnsafeProposal() {
```
AlexHentschel, line 288: Would suggest a bit more documentation:
* for most test cases in `TestProduceVote_VotingOnUnsafeProposal`, we expect an exception. This is because all blocks are supposed to be pre-validated. And if any of the sanity checks fail, we interpret this as a symptom of an internal bug in the validation logic. 

### Suggestions:
1. Rename `TestProduceVote_VotingOnUnsafeProposal` to `TestProduceVote_VotingOnInvalidProposals` and include the following documentation
    ```suggestion
    // TestProduceVote_VotingOnInvalidProposals tests different scenarios where we try to vote on unsafe blocks
    // SafetyRules contain a variety of checks to confirm that QC and TC have the desired relationship to each other.
    // In particular, we test:
    //    (i) A TC should be included in a proposal, if and only of the QC is not the prior view.
    //   (ii) When the proposal includes a TC (i.e. the QC not being for the prior view), the TC must be for the prior view.
    //  (iii) The QC in the block must have a smaller view than the block.
    //   (iv) If the block contains a TC, the TC cannot contain a newer QC than the block itself.
    // Conditions (i) - (iv) are validity requirements for the block and all blocks that SafetyRules processes
    // are supposed to be pre-validated. Hence, failing any of those conditions means we have an internal bug.
    // Consequently, we expect SafetyRules to return exceptions but _not_ `NoVoteError`, because the latter 
    // indicates that the input block was valid but we didn't want to vote. 
    func (s *SafetyRulesTestSuite) TestProduceVote_VotingOnInvalidProposals() {
    ```
2 The only test scenarios in `TestProduceVote_VotingOnUnsafeProposal` that are different are the [last two cases](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L371-L397)
   * regarding [`view-below-acknowledged`](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L384-L397): I feel this test falls into the category of [`TestProduceVote_InvalidCurrentView`](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L194-L203). 
       * Essentially we test the scenario where the `EventHandler` is feeding `SafetyRules` with wrong values for current view. In this particular case, we are testing that `SafetyRules` throws an exception if it receives inputs where values for `curView` are not monotonously increasing. 
       * I would suggest to also update the goDoc of [`TestProduceVote_InvalidCurrentView`](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L194-L203) a bit (see my comment below)
   * regarding [`view-already-acknowledged`](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L371-L383): I feel this is conceptually a different test case, that probably deserves its own test function:
      * Here we are testing that _only_ the first block for a view receives a vote (provided it is safe to vote). Other valid blocks imply that the proposer equivocated and should lead to a `NotVoteError`. 
      * I think this case is conceptually different compared to the other things we are testing in `TestProduceVote_VotingOnUnsafeProposal` is: here the input block by itself is actually valid; in contrast, the blocks for all the other cases are actually broken.

AlexHentschel (approved):
very nice. Thanks for the excellent test coverage 🥇

AlexHentschel:
I think we could improve the clarity of the goDoc for `TestProduceVote_InvalidCurrentView`: https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules_test.go#L195-L197

I don't think we test that `SafetyRules` correctly handles invalid blocks here (?) From my perspective, we test that `SafetyRules` enforces consistency rules for the `curView` parameter in [`ProduceVote`](https://github.com/onflow/flow-go/blob/7ea4f828d37a5953bd4a4d1a01acbe422a8b6ebe/consensus/hotstuff/safetyrules/safety_rules.go#L65). Hence, I would suggest to update the goDoc to 
```golang
// TestProduceVote_InvalidCurrentView tests that no vote is created if `curView` has invalid values.
// In particular, `SafetyRules` requires that:
//  * the block's view matches `curView`
//  * that values for `curView` are monotonously increasing
// Failing any of these conditions is a symptom of an internal bug; hence `SafetyRules` should
// _not_ return a `NoVoteError`.
func (s *SafetyRulesTestSuite) TestProduceVote_InvalidCurrentView() {
```

consensus/hotstuff/safetyrules/safety_rules_test.go
durkmurder: This test checks curView > highestQC.View, I will add extra test for curView == highestQC.View

jordanschalm (approved):
🎸

durkmurder:
bors merge

PR #2558, opened 2022-06-06
Expose Headers in AN

lolpuddle:
Closes [5930](https://github.com/dapperlabs/flow-go/issues/5930)


peterargue:
Were you able to reproduce this issue and confirm this fixes it? 

My understanding is that `Access-Control-Expose-Headers` sets which http headers a  browser client can access within the cross-origin response, but in the original issue, the browser was complaining that the request itself wasn't allowed.

Issue #2248, opened 2022-04-04
[Execution State] Evaluate reducing number of memory allocs by changing Payload.Key data structure


fxamacker:
## Problem

High number of allocations involving `Payload.Key` was observed.

`Payload.Key` is a slice of `KeyPart`.  `KeyPart` consists of `uint16` and `[]byte`.

Updates epic https://github.com/onflow/flow-go/issues/1744

## The Proposed Solution

Determine if it's feasible to modify `Payload.Key` to reduce number of allocations.

fxamacker:
I'm closing this because it is addressed by PR #2930 and PR #2560.

PR #2560, opened 2022-06-06
(WIP) [EN Performance] Reduce memory use and speedup reading single register

### LINKED ISSUE #2559: [Execution Node] Reduce memory use and speedup reading single register
## Problem

Reading a single register is very frequent and it can be optimized to use less memory.

`ledger.Key` requires 4 allocs which can be avoided when reading registers and generating proofs because we only need register path and key is an intermediate object created to compute path.

## The Proposed Solution

To optimize reading single register, make the execution state:
- use a new type `ledger.KeyID` to compute `ledger.Path`.
- use `KeyID` which is a simplified `ledger.Key` with fewer allocs.
- create `ledger.TrieRead` and `ledger.TrieUpdate` without intermediate objects such as `ledger.Query` and `ledger.Update`.

With these changes, `ledger.Key` is only created when updating trie with new payloads.

Updates #1744

fxamacker:
### Changes

To reduce memory used when reading single register, make the execution state:
- use a new type `ledger.KeyID` to compute `ledger.Path`.
- use `KeyID` which is a simplified `ledger.Key` with fewer allocs.
- create `ledger.TrieRead` and `ledger.TrieUpdate` without intermediate objects such as `ledger.Query` and `ledger.Update`.

With these changes, `ledger.Key` is only created when updating trie with new payloads.

Updates https://github.com/onflow/flow-go/issues/1744
Updates #2248
Closes #2559 

### Benchmark Comparisons

```
name                     old time/op    new time/op    delta
ExecutionStateRead1-4    7.16µs ± 1%    6.66µs ± 1%   -6.96%  (p=0.000 n=9+10)

name                     old alloc/op   new alloc/op   delta
ExecutionStateRead1-4    1.36kB ± 0%    1.16kB ± 0%  -14.98%  (p=0.000 n=10+8)

name                     old allocs/op  new allocs/op  delta
ExecutionStateRead1-4      15.0 ± 0%      11.0 ± 0%  -26.67%  (p=0.000 n=10+10)

Benchmarks used Go 1.17 on linux_amd64 (Haswell)
```


Kay-Zee:
@fxamacker is this blocked by anything?

fxamacker:
@Kay-Zee  This PR was a POC (for discussion).  I think it's worthwhile to look into other approaches before proceeding with this PR's approach.  I'm closing this PR for now and can open new PR after wrapping up CCF specification and CCF codec.

cc @ramtinms

PR #2564, opened 2022-06-06
[Consensus] `ActivePaceMaker` implementation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6194

### Context

This PR replaces old passive pacemaker with active pacemaker. Main changes are around `hotstuff/consensus/pacemaker` module. 

- Replaced passive pacemaker with new implementation
- Extended consumer with new functionality to report views that were changed by observing TC
- Updated existing tests, added new tests
- Updated timeout control API to support stopping of timeout.


consensus/hotstuff/pacemaker/pacemaker.go
```diff
@@ -12,111 +12,152 @@ import (
...
+func (p *ActivePaceMaker) LastViewTC() *flow.TimeoutCertificate {
+	return p.livenessData.LastViewTC
 }
 
 // Start starts the pacemaker
-func (p *NitroPaceMaker) Start() {
+func (p *ActivePaceMaker) Start() {
 	if p.started.Swap(true) {
 		return
 	}
```
jordanschalm, line 155: This purpose of `started` is to ensure we only start the initial timer once. To me it would be clearer to use `sync.Once` for this instead, since it's made exactly for that purpose.

consensus/hotstuff/pacemaker/pacemaker.go
```diff
@@ -12,111 +12,152 @@ import (
...
 
 // LastViewTC returns TC for last view, this could be nil if previous round
 // has entered with a QC.
-func (p *NitroPaceMaker) LastViewTC() *flow.TimeoutCertificate {
-	panic("not yet implemented")
+func (p *ActivePaceMaker) LastViewTC() *flow.TimeoutCertificate {
+	return p.livenessData.LastViewTC
 }
 
 // Start starts the pacemaker
```
jordanschalm, line 151:
```suggestion
// Start starts the pacemaker by starting the initial timer for the current view.
// Start should only be called once - subsequent calls are a no-op.
```

consensus/hotstuff/pacemaker/pacemaker.go
```diff
@@ -12,111 +12,152 @@ import (
...
-func (p *NitroPaceMaker) Start() {
+func (p *ActivePaceMaker) Start() {
 	if p.started.Swap(true) {
 		return
 	}
-	timerInfo := p.timeoutControl.StartTimeout(model.ReplicaTimeout, p.currentView)
+	timerInfo := p.timeoutControl.StartTimeout(model.ReplicaTimeout, p.CurView())
```
jordanschalm, line 156: We always use `ReplicaTimeout` here. What if we happen to be the leader for the first view after we start up? 

I guess the impact if any is small, because restarts are infrequent and at most this would cause one additional timed out one view per restart. But would using the more liberal timeout following a restart make sense?

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
...
+	nve, err := s.paceMaker.ProcessTC(tc)
+	require.NoError(s.T(), err)
+	s.notifier.AssertExpectations(s.T())
+	require.Equal(s.T(), uint64(4), s.paceMaker.CurView())
+	require.True(s.T(), nve.View == 4)
+	require.Equal(s.T(), tc, s.paceMaker.LastViewTC())
+
+	tc = helper.MakeTC(helper.WithTCView(12),
+		helper.WithTCNewestQC(s.livenessData.NewestQC),
+		helper.WithTCNewestQC(QC(3)))
```
jordanschalm, line 140:
```suggestion
		helper.WithTCNewestQC(s.livenessData.NewestQC),
```
setting `NewestQC` twice

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
...
-//}
+// TestProcessQC_PersistException tests that ActivePaceMaker propagates exception
+// when processing QC
+func (s *ActivePaceMakerTestSuite) TestProcessQC_PersistException() {
+	exception := errors.New("persist-exception")
+	qc := QC(3)
+	s.persist.On("PutLivenessData", mock.Anything).Return(exception).Once()
+	nve, err := s.paceMaker.ProcessQC(qc)
+	require.Nil(s.T(), nve)
+	require.ErrorAs(s.T(), err, &exception)
```
jordanschalm, line 167:
```suggestion
	require.ErrorIs(s.T(), err, exception)
```
Better to use the more specific error equality assertion

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
...
-//}
+// TestProcessTC_PersistException tests that ActivePaceMaker propagates exception
+// when processing TC
+func (s *ActivePaceMakerTestSuite) TestProcessTC_PersistException() {
+	exception := errors.New("persist-exception")
+	tc := helper.MakeTC(helper.WithTCView(3))
+	s.persist.On("PutLivenessData", mock.Anything).Return(exception).Once()
+	nve, err := s.paceMaker.ProcessTC(tc)
+	require.Nil(s.T(), nve)
+	require.ErrorAs(s.T(), err, &exception)
```
jordanschalm, line 178:
```suggestion
	require.ErrorIs(s.T(), err, exception)
```

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
...
+}
 
-// Test_SkipIncreaseViewThroughQC tests that PaceMaker increases View when receiving QC,
+// TestProcessQC_SkipIncreaseViewThroughQC tests that ActivePaceMaker increases view when receiving QC,
 // if applicable, by skipping views
-func Test_SkipIncreaseViewThroughQC(t *testing.T) {
-	pm, notifier := initPaceMaker(t, 3)
-
+func (s *ActivePaceMakerTestSuite) TestProcessQC_SkipIncreaseViewThroughQC() {
 	qc := QC(3)
```
jordanschalm, line 92:
```suggestion
	qc := QC(s.livenessData.CurrentView)
```
There are a lot of numeric literals used here and in the other test cases - it would be nice if these values were computed based on the test suite configuration, instead of using constants:
* the test will work without changes if the initial state changes
* it's also clearer to a reader, `QC(curView)` tells me that the QC is for the current view - I don't have to refer back to the config

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
...
-//	notifier.On("OnQcTriggeredViewChange", block.QC, uint64(14)).Return().Once()
-//	notifier.On("OnStartingTimeout", expectedTimerInfo(14, model.ReplicaTimeout)).Return().Once()
-//	notifier.On("OnStartingTimeout", expectedTimerInfo(14, model.VoteCollectionTimeout)).Return().Once()
-//	nve, nveOccurred = pm.UpdateCurViewWithBlock(block, true)
-//	notifier.AssertExpectations(t)
-//	assert.Equal(t, uint64(14), pm.CurView())
-//	assert.True(t, nveOccurred && nve.View == 14)
-//}
+// TestOnPartialTC_TriggersTimeout tests that ActivePaceMaker ignores partial TCs and triggers
+// timeout for active view
```
jordanschalm, line 191:
```suggestion
// TestOnPartialTC_TriggersTimeout tests that ActivePaceMaker ignores partial TCs for previous
// views and triggers an immediate timeout upon observing a partial TC for the active view.
```

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -30,456 +32,188 @@ func expectedTimerInfo(view uint64, mode model.TimeoutMode) interface{} {
 		})
 }
 
-func expectedTimeoutInfo(view uint64, mode model.TimeoutMode) interface{} {
-	return mock.MatchedBy(
-		func(timerInfo *model.TimerInfo) bool {
-			return timerInfo.View == view && timerInfo.Mode == mode
-		})
+func TestActivePaceMaker(t *testing.T) {
```
jordanschalm, line 35: Suggesting some additional test cases:
* `TestProcessTC_IgnoreOldTC` - same as `TestProcessQC_IgnoreOldQC`, but for TCs
* `TestProcessTC_IgnoreNilTC` - test that a nil TC is an allowable input
* Process a TC, then a QC -> make sure `LastViewTC` is `nil`

jordanschalm (commented):
Nice work. The core implementation looks good to me. Left some suggestions mainly around doc updates and additional test cases.

consensus/hotstuff/pacemaker/pacemaker.go
```diff
@@ -12,111 +12,166 @@ import (
...
+		return nil, err
+	}
+
+	p.notifier.OnTcTriggeredViewChange(tc, newView)
+	return &model.NewViewEvent{View: newView}, nil
 }
 
-func (p *NitroPaceMaker) OnPartialTC(curView uint64) {
-	panic("not yet implemented")
+func (p *ActivePaceMaker) OnPartialTC(newView uint64) {
```
AlexHentschel, line 147: I am starting to believe that the PaceMaker does not need to know about the Bracha timeout. Reasoning:
* We are building an information-driven system (specifically, because it helps modularizing algorithms). Information is idempotent. In this particular case, it means that reaching a timeout condition repeatedly should leave the system state invariant. 

   Benefits:
   * The pace maker can be only concerned with its time-based timeouts. No need to cancel the timeout if a Bracha timeout occurs first. Repeated timeout triggers for the same view don't hurt. 
   * If a Bracha timeout occurred, the Event Hander can directly call into SafetyRules. SafetyRules gracefully handles repeated calls for `ProduceTimeout`. 
      
        No need to call into PaceMaker to trigger the timeout, which then loops again through the Event Loop to trigger `SafetyRules`

AlexHentschel (commented):
First batch of comments

consensus/hotstuff/pacemaker/pacemaker.go
AlexHentschel: +1

consensus/hotstuff/pacemaker/pacemaker.go
AlexHentschel: In my opinion, the two different timeouts (`ReplicaTimeout` vs `VoteCollectionTimeout`) is legacy from the passive pacemaker. We should probably remove it. 
With the active pacemaker, there are now strict requirements for a node to change view: they need proof as either a QC or TC. There is no difference anymore between the leader vs other replicas. They all follow the same rules. Except that the leader constructs the needed evidence for view change first (triggering itself to enter the view where it is primary); and then the primary broadcasts the proof (specifically the QC) as part of its block, which allows all other nodes to change the view.

consensus/hotstuff/pacemaker/pacemaker.go
```diff
@@ -12,111 +12,166 @@ import (
...
 	}
-	p.currentView = newView
+	p.livenessData.LastViewTC = tc
+	err := p.persist.PutLivenessData(p.livenessData)
+	if err != nil {
+		return fmt.Errorf("could not persist liveness data: %w", err)
+	}
+
 	timerInfo := p.timeoutControl.StartTimeout(model.ReplicaTimeout, newView)
 	p.notifier.OnStartingTimeout(timerInfo)
```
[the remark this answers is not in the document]
durkmurder: yeah, I agree

consensus/hotstuff/pacemaker/pacemaker.go
durkmurder: Yeah, my initial assumption was built on idea of canceling the timeout, but yeah, I will rollback changes for canceling timeout and just leave it as it is.

AlexHentschel:
quick request to fix goDoc: https://github.com/onflow/flow-go/blob/30741902dd6938291b82a42acf427fe939e88dd8/consensus/hotstuff/pacemaker/timeout/config.go#L10-L11
Please replace by 
```golang
// Config contains the configuration parameters for exponential timeout.Controller:
```

consensus/hotstuff/pacemaker/timeout/config_test.go
```diff
@@ -9,41 +9,34 @@ import (
...
 	require.Equal(t, float64(1200), c.MinReplicaTimeout)
-	require.Equal(t, float64(0.73), c.VoteAggregationTimeoutFraction)
 	require.Equal(t, float64(1.5), c.TimeoutIncrease)
 	require.Equal(t, float64(0.85), c.TimeoutDecrease)
 	require.Equal(t, float64(1000), c.BlockRateDelayMS)
 
 	// should not allow startReplicaTimeout < minReplicaTimeout
-	_, err = NewConfig(800*time.Millisecond, 1200*time.Millisecond, 0.73, 1.5, 0.85, time.Second)
+	_, err = NewConfig(800*time.Millisecond, 1200*time.Millisecond, 1.5, 0.85, time.Second)
 	require.Error(t, err)
```
AlexHentschel, line 22: could we please replace _all_ `require.Error(t, err)` in this test by 
```golang
require.True(t, model.IsConfigurationError(err))
```
thanks

consensus/hotstuff/pacemaker/timeout/controller.go
```diff
@@ -53,50 +53,27 @@ func (t *Controller) TimerInfo() *model.TimerInfo { return t.timerInfo }
...
 }
 
-// VoteCollectionTimeout returns the duration of Vote aggregation _after_ receiving a block
-// during which the primary tries to aggregate votes for the view where it is leader
-func (t *Controller) VoteCollectionTimeout() time.Duration {
-	// time.Duration expects an int64 as input which specifies the duration in units of nanoseconds (1E-9)
-	return time.Duration(t.cfg.ReplicaTimeout * 1e6 * t.cfg.VoteAggregationTimeoutFraction)
-}
-
 // OnTimeout indicates to the Controller that the timeout was reached
```
AlexHentschel, line 77:
```suggestion
// OnTimeout indicates to the Controller that a view change was triggered by a TC (unhappy path).
```

consensus/participant.go
```diff
@@ -84,8 +77,7 @@ func NewParticipant(
 	// initialize the pacemaker
 	controller := timeout.NewController(timeoutConfig)
```
AlexHentschel, line 79:
```suggestion
```

AlexHentschel (approved):
Rock solid work with excellent tests. 👏

consensus/hotstuff/pacemaker/timeout/config.go
```diff
@@ -18,9 +18,6 @@ type Config struct {
 	ReplicaTimeout float64
 	// MinReplicaTimeout is the minimum the timeout can decrease to [MILLISECONDS]
 	MinReplicaTimeout float64
```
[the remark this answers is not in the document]
durkmurder: It will be updated as part of round duration research: https://github.com/dapperlabs/flow-go/issues/6203
But I will add a cap as you have suggested in this PR.

consensus/hotstuff/pacemaker/pacemaker_test.go
```diff
@@ -18,468 +20,200 @@ import (
...
-//	case <-time.After(time.Duration(3) * time.Duration(startRepTimeout) * time.Millisecond):
-//	}
-//
-//	actualTimeout = float64(time.Since(start).Microseconds()) * 0.001 // in millisecond
-//	// the actual timeout should be startRepTimeout * 1.5 *1.5 * multiplicativeDecrease * multiplicativeDecrease
-//	// because it hits timeout twice and then received blocks for current view twice
-//	assert.GreaterOrEqual(t, actualTimeout, 1.5*1.5*startRepTimeout*multiplicativeDecrease*multiplicativeDecrease, "the actual timeout should be greater or equal to the timeout")
-//	assert.Less(t, actualTimeout, 1.5*1.5*startRepTimeout*multiplicativeDecrease, "the actual timeout is too long")
-//}
+// TestProcessQC_InvalidatesLastViewTC
```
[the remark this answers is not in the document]
durkmurder: :face_exhaling:

jordanschalm (approved):

Issue #2563, opened 2022-06-06
Port FVM event encoding error handling fix to flow-go master


j1010001:
Port  https://github.com/onflow/flow-go/pull/2562 to flow-go master.

m4ksio:
Closed by https://github.com/onflow/flow-go/pull/2565

PR #2565, opened 2022-06-06
Handle event encoding error gracefully (Port of 2562)

m4ksio:
Originally this was a simple port, but it was unmerged for too long and I had to start from scratch.
I


janezpodhostnik (approved):

Kay-Zee:
bors merge

m4ksio:
bors merge

fvm/environment/event_encoder.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewCadenceEventEncoder() *CadenceEventEncoder {
+	var buf bytes.Buffer
+	return &CadenceEventEncoder{
+		buffer:  &buf,
+		encoder: jsoncdc.NewEncoder(&buf),
+	}
+}
+
+func (e *CadenceEventEncoder) Encode(event cadence.Event) ([]byte, error) {
```
pattyshack, line 27: just call `return jsoncdc.Encode(event)` here?  Since you end up resetting the buffer / copying the payload on each Encode call, there no benefit in pre-allocating a buffer / encoder.

m4ksio: Reusing the buffer should give some speedeup, as resetting keeps underlaying bytes, but this seemed to be rather insignificant. So I did a benchmark and looks like there is quite a difference.
```BenchmarkEncodingBuffer-16               1326105               895.0 ns/op
BenchmarkEncodingBuffer-16               1306316               949.6 ns/op
BenchmarkEncodingBuffer-16               1230076               944.8 ns/op
BenchmarkEncodingBuffer-16               1282364               941.2 ns/op
BenchmarkEncodingBuffer-16               1311118               900.7 ns/op
BenchmarkEncodingBufferV2-16             1468462               807.6 ns/op
BenchmarkEncodingBufferV2-16             1501274               821.0 ns/op
BenchmarkEncodingBufferV2-16             1456159               812.9 ns/op
BenchmarkEncodingBufferV2-16             1464187               818.2 ns/op
BenchmarkEncodingBufferV2-16             1410309               829.4 ns/op
PASS
ok      github.com/onflow/flow-go/fvm/environment       20.927```
V2 is the new version. I think difference would be even more visible for larger events.

pattyshack (approved):

ramtinms (approved):
Looks good to me

m4ksio:
Closes https://github.com/onflow/flow-go/issues/2563

PR #2566, opened 2022-06-06
[State Sync] Lookup ExecutionResult directly instead of using index

peterargue:
Depends on: https://github.com/onflow/flow-go/pull/2551

The `ExecutionDataRequester` currently uses the `ExecutionResult` index `ByBlockID` to lookup the result's ID, but this index is not part of the protocol state, and is written asynchronously. Rather than using the index which may be unreliable, lookup the result directly by ID. 

This requires https://github.com/onflow/flow-go/pull/2551 which adds a new index to the protocol state which maps the blockID of the block whose result is included in a seal, to the actual seal. This enables us to Lookup the result from a block that is known to be sealed. In this PR, that `ExecutionDataID` from the result is used to fetch the execution data.


AlexHentschel:
[`Headers.ByHeight`](
https://github.com/onflow/flow-go/blob/b69fa33a25ed981e2ba77a33d3329cf0bb4e6fc7/storage/headers.go#L19-L21) also supports querying by height, and also offers a caching layer on top. In the previous implementation `SealedBlockHeaderReader` already took a `storage.Headers` implementation, which is removed here.

AlexHentschel:
State is the first layer abstraction above the plain database. There are higher-level storage abstractions which have active cashing to improve runtime. Please use those. In this particular case, usage of https://github.com/onflow/flow-go/blob/b69fa33a25ed981e2ba77a33d3329cf0bb4e6fc7/storage/seals.go#L19-L20 is strongly encouraged.

AlexHentschel:
The prior version here https://github.com/onflow/flow-go/blob/b69fa33a25ed981e2ba77a33d3329cf0bb4e6fc7/module/state_synchronization/requester/jobs/execution_data_reader.go#L93 used the storage-layer abstractions _with_ caching. I think keeping this would be beneficial (?)

peterargue:
@AlexHentschel thanks for reviewing. I've switched back to the `storage` interfaces.

module/state_synchronization/requester/unittest/unittest.go
```diff
@@ -116,12 +124,25 @@ func MockProtocolStateSnapshot(opts ...SnapshotMockOptions) *statemock.Snapshot
...
 type StateMockOptions func(*statemock.State)
 
-func WithSnapshot(snapshot *statemock.Snapshot) StateMockOptions {
+func WithSealed(snapshot *statemock.Snapshot) StateMockOptions {
 	return func(state *statemock.State) {
 		state.On("Sealed").Return(snapshot)
 	}
 }
 
+func WithAtHeight(snapshots map[uint64]*statemock.Snapshot) StateMockOptions {
```
smnzhu, line 133: Can we name this something more descriptive? e.g `WithSnapshotsByHeight`?

module/state_synchronization/requester/unittest/unittest.go
```diff
@@ -116,12 +124,25 @@ func MockProtocolStateSnapshot(opts ...SnapshotMockOptions) *statemock.Snapshot
 
 type StateMockOptions func(*statemock.State)
 
-func WithSnapshot(snapshot *statemock.Snapshot) StateMockOptions {
+func WithSealed(snapshot *statemock.Snapshot) StateMockOptions {
```
smnzhu, line 127:
```suggestion
func WithSealedSnapshot(snapshot *statemock.Snapshot) StateMockOptions {
```

module/state_synchronization/requester/unittest/unittest.go
```diff
@@ -203,6 +224,25 @@ func WithByBlockID(resultsByID map[flow.Identifier]*flow.ExecutionResult) Result
 	}
 }
 
+func WithByResultID(resultsByID map[flow.Identifier]*flow.ExecutionResult) ResultsMockOptions {
```
smnzhu, line 227:
```suggestion
func WithResultsByID(resultsByID map[flow.Identifier]*flow.ExecutionResult) ResultsMockOptions {
```

module/state_synchronization/requester/unittest/unittest.go
```diff
@@ -213,6 +253,37 @@ func MockResultsStorage(opts ...ResultsMockOptions) *storagemock.ExecutionResult
 	return results
 }
 
+type SealsMockOptions func(*storagemock.Seals)
+
+func WithBySealedBlockID(sealsByBlockID map[flow.Identifier]*flow.Seal) SealsMockOptions {
```
smnzhu, line 258:
```suggestion
func WithSealsByBlockID(sealsByBlockID map[flow.Identifier]*flow.Seal) SealsMockOptions {
```

smnzhu (approved):

module/state_synchronization/requester/jobs/execution_data_reader.go
```diff
@@ -94,7 +97,13 @@ func (r *ExecutionDataReader) getExecutionData(signalCtx irrecoverable.SignalerC
...
 	}
 
-	result, err := r.results.ByBlockID(header.ID())
+	// get the ExecutionResultID for the block from the block's seal
+	seal, err := r.seals.BySealedBlockID(header.ID())
+	if err != nil {
+		return nil, fmt.Errorf("failed to lookup seal for block %s: %w", header.ID(), err)
+	}
+
+	result, err := r.results.ByID(seal.ResultID)
```
zhangchiqing: nit: we could extract this into a function, since we don't care about the seal but only the result

module/jobqueue/sealed_header_reader.go
```diff
@@ -26,33 +25,23 @@ func NewSealedBlockHeaderReader(state protocol.State, headers storage.Headers) *
...
 	}
 
-	// the block at height index is sealed
-	return BlockHeaderToJob(header), nil
-}
-
-// blockByHeight returns the block at the given height.
-func (r SealedBlockHeaderReader) blockByHeight(height uint64) (*flow.Header, error) {
-	block, err := r.headers.ByHeight(height)
+	header, err := r.headers.ByHeight(index)
```
zhangchiqing, line 38: `SealedBlockHeaderReader` only need to depend on protocol state, which internally manages the headers storage module.

```
header, err := r.state.AtHeight(height).Head()
```

zhangchiqing (approved):

module/jobqueue/sealed_header_reader.go
peterargue: I originally used this, but @AlexHentschel asked me to switch it back to take advantage of caching in the headers storage module

peterargue:
bors merge

peterargue:
bors retry

AlexHentschel (commented):
Suggestions for improvement for discussion at `BFT Engineering and PR Open house` (June 10, 2022)

PR #2575, opened 2022-06-08
Documentation: Coding Conventions

AlexHentschel:
first iteration; I wrote this some time back but forgot to commit it to master 😅


CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+err := foo()
+if err != nil {
+   log.Error().Err(err).Msg("foo failed")
+   return
+}
+```
+
+There are _rare_ instances, where this might be acceptable. For example, when attempting to send a message via the
+network to another node and the networking layer errored. In this case, please include a comment with an explanation,
+why it is acceptable to keep going, even if the other component returned an error.
```
jordanschalm, line 86:
```suggestion
There are _rare_ instances, where this might be acceptable. For example, when attempting to send a message via the
network to another node and the networking layer errored. In this case, please include a comment with an explanation,
why it is acceptable to keep going, even if the other component returned an error. For this example, it is acceptable to handle the error by logging because:
* we expect transient errors from this component (networking layer) during normal operation
* the networking layer uses a 3rd party library, which does not expose specific and exhaustive sentinel errors for expected failure conditions
* either it is not critical that the message is successfully sent at all, or it will be retried later on
```

CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+
+### prioritize safety over liveness
```
jordanschalm, line 88:
```suggestion
### Prioritize Safety Over Liveness
```

CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+```golang
+err := foo()
+if err != nil {
+   if errors.Is(err, XFailedErr) {
+      // expected error
+      return
+   }
+   log.Fatal().Err(err).Msg("unexpected internal error")
+   return
+}
```
jordanschalm, line 105:
```suggestion
if errors.Is(err, XFailedErr) {
   // expected error
   return
}
if err != nil {
   log.Fatal().Err(err).Msg("unexpected internal error")
   return
}
```

Personally prefer this format for error checking, because it saves a level of conditions.

CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+There are _rare_ instances, where this might be acceptable. For example, when attempting to send a message via the
+network to another node and the networking layer errored. In this case, please include a comment with an explanation,
+why it is acceptable to keep going, even if the other component returned an error.
+
+### prioritize safety over liveness
+
+**Ideally, a component should restart** (from a known good state), **when it encounters an unexpected error.**
+**If this is out of scope, we _prioritize safety over liveness_.** This means that we rather crash the node than
+continue on a best-effort basis.
+
```
jordanschalm, line 93:
```suggestion
continue on a best-effort basis.

New engines should implement the [`Component` interface](https://github.com/onflow/flow-go/blob/57f89d4e96259f08fe84163c91ecd32484401b45/module/component/component.go#L22)
and throw unexpected errors using the related [irrecoverable context](https://github.com/onflow/flow-go/blob/277b6515add6136946913747efebd508f0419a25/module/irrecoverable/irrecoverable.go). 

```

CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+## Engines
+
+On a high level, a Flow node consists of various data-processing components that receive information from node-internal
+and external sources. We model the data flow inside a Flow node as a graph:
+a data-processing component forms a vertex; two components exchanging messages is represented as an edge. An `Engine` is
+the interface for a data-processing component.
+
+Generally, we consider node-internal components as _trusted_ and external nodes as untrusted data sources. The `Engine`
+API differentiates between trusted and untrusted inputs:
```
jordanschalm, line 129: Given that the engine API is deprecated in favour of `MessageProcessor`, maybe we could use different examples here. 

I created a PR adding deprecation flags with some suggested alternatives to the deprecated Engine methods: https://github.com/onflow/flow-go/pull/2600. If we have consensus on those suggested alternative, I can help update the examples here as well.

CodingConventions.md
```diff
@@ -0,0 +1,222 @@
...
+
+Furthermore, engines offer synchronous and asynchronous processing of their inputs:
+
+* Methods `engine.ProcessLocal` and `engine.Process` run their logic in the caller's routine. Therefore, the core
+  business logic can return errors, which we should capture and propagate.
+* Methods `engine.SubmitLocal` and `engine.Submit` run the business logic asynchronously. (no error return)
+    * Note: computationally very cheap and predominantly non-blocking computation should _still_ be executed in the
+      calling routine. This is beneficial to avoid a cascade of go-routine launches when stacking multiple engines.
+
+### guidelines
```
jordanschalm, line 160:
```suggestion
### Guidelines
```

jordanschalm (commented):
Thank you, it's great to have this better documented 👏

CodingConventions.md
```diff
@@ -0,0 +1,307 @@
...
+As a first step, you can crash the node when an unexpected error hits the `Engine` level (for more details, see `Engine`
+section of this doc). If you have followed the error handling guidelines, leaving the restart logic for the component as
+technical debt for later hopefully does not add much overhead. In contrast, if you log errors and continue on a
+best-effort basis, you also leave the _entire_ logic for differentiating errors as tech debt. When adding this logic
+later, you need to revisit the _entire_ business logic again.
+
+
+# Appendix
+
+## Engines (interface deprecated)
```
franklywatson, line 208: I think it would be helpful to explain why engines are deprecated since they are still so pervasive. Also, is there a plan to fix this - eg, a tracking GH issue?

CodingConventions.md
```diff
@@ -0,0 +1,307 @@
...
+*not* an option. There is a clear definition of the happy path (i.e. the block must be in storage). Any deviation of this
+happy path is either
+
+1. a protocol violation (from a different node), which itself is clearly defined
+2. is an uncovered edge case (which could be exploited in the worst case to compromise the system)
+3. a corrupted internal state of the node
+
+### Best Practice Guidelines
+
+1. **Errors are part of your API:** If there is an error return, there is a documentation what the error means. We conceptually differentiate between the following two classes of errors:
```
franklywatson, line 77: suggest: "If there is an error returned the meaning of the error is clearly documented in the code"

CodingConventions.md
```diff
@@ -0,0 +1,307 @@
...
+### Byzantine Inputs result in benign sentinel errors
+
+Any input that is coming from another node is untrusted and the vertex processing it should gracefully handle any arbitrarily malicious input.
+Under no circumstances should a byzantine input result in the vertex's internal state being corrupted.
+Per convention, failure case due to a byzantine inputs are represented by specifically-typed
+[sentinel errors](https://pkg.go.dev/errors#New) (for further details, see error handling section).
+
+Byzantine inputs are one particular case, where a function returns a 'benign error'. The critical property for an error
+to be benign is that the component returning it is still fully functional, despite encountering the error condition. All
+benign errors should be handled within the vertex.
```
franklywatson, line 60: I think this section could be more explicit about the implementor weighing if an error is truly benign and having a high confidence in that. When that applies the benign error ensures no panic or fault in program flow.

CodingConventions.md
```diff
@@ -0,0 +1,307 @@
...
+treats all inputs as untrusted. Furthermore, `VoteAggregator` consumes auxiliary inputs
+([`AddBlock(block *model.Proposal)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L45),
+[`InvalidBlock(block *model.Proposal)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L51),
+[`PruneUpToView(view uint64)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L57))
+from the node's main consensus logic, which trusted.
+
+
+### Byzantine Inputs result in benign sentinel errors
+
+Any input that is coming from another node is untrusted and the vertex processing it should gracefully handle any arbitrarily malicious input.
```
franklywatson, line 53: nit: "and the vertex processing should"

CodingConventions.md
```diff
@@ -0,0 +1,307 @@
...
+
+For example, [`VoteAggregator`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L29) is
+a data flow vertex within a consensus nodes, whose job it is to collect incoming votes from other consensus nodes.
+Hence, the corresponding function [`AddVote(vote *model.Vote)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L37)
+treats all inputs as untrusted. Furthermore, `VoteAggregator` consumes auxiliary inputs
+([`AddBlock(block *model.Proposal)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L45),
+[`InvalidBlock(block *model.Proposal)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L51),
+[`PruneUpToView(view uint64)`](https://github.com/onflow/flow-go/blob/3899d335d5a6ce3cbbb9fde4f7af780d6bec8c9a/consensus/hotstuff/vote_aggregator.go#L57))
+from the node's main consensus logic, which trusted.
```
franklywatson, line 48: nit: which is trusted

CodingConventions.md
jordanschalm: Linked a tracking issue with an explanation

CodingConventions.md
jordanschalm: Added a sentence to this effect in https://github.com/onflow/flow-go/pull/2575/commits/6a5d51f431e91d89dd2b9efd570493dff2156667, let me know if you had specific wording in mind.

jordanschalm: Added a sentence to this effect in https://github.com/onflow/flow-go/pull/2575/commits/6a5d51f431e91d89dd2b9efd570493dff2156667, let me know if you had specific wording in mind.

CodingConventions.md
jordanschalm: Alex added a section here with a different example demonstrating this pattern: https://github.com/onflow/flow-go/pull/2575/files#diff-3b52addb04dfea0ae48593e2aea4efa434d108d00bc9df92f8b51f170dc807fbR41-R48

jordanschalm (approved):

CodingConventions.md
```diff
@@ -0,0 +1,310 @@
...
+    * Note: computationally very cheap and predominantly non-blocking computation should _still_ be executed in the
+      calling routine. This is beneficial to avoid a cascade of go-routine launches when stacking multiple engines.
+
+### Guidelines
+
+* In the error handling within an `Engine`, differentiate between trusted and untrusted inputs. For example, most
+  engines expect specific input types. Inputs with incompatible type should result in a specific sentinel error (
+  e.g. `InvalidMessageType`).    
+  For messages that come from the network (methods `Submit` and `Process`), we expect invalid types. But if we get an
+  invalid message from a trusted local engine
```
zhangchiqing, line 254:
```suggestion
  invalid message type from a trusted local engine
```

Maybe we should be more specific about what is considered to be invalid. 

For instance, the follower engine could receive a block from the network or synchronization engine. 

If a block from sync engine to follower engine is considered to be trusted input, then it's still possible for the block to be invalid. Sync engine doesn't validate the the message (block), it's just responsible for delivering the message.

Also whether or not a message is "invalid" might depend on another module's state, for instance, a new block at the moment is not invalid, but after a while might become "invalid" because it's conflicting with the finalized block. 

Maybe we could restrict the message type here, saying if we get an invalid message _type_ from a trusted local engine, then something is broken.

zhangchiqing (commented):
Looks good. Nice write up!

Just had one comment.

CodingConventions.md
jordanschalm: 👍 Good call-out. I will update this to be more detailed. 

Personally though I don't think we should constrain the requirement to "invalid message _type_". Different components will have different definitions for what is a valid message. In all cases, an unexpected message type is invalid. But in many cases, an expected type with malformed data is also invalid. 

We also should be moving away from non-type-safe interfaces between components (engines), so the invalid type case will be less of a concern over time. 

In the case of the sync engine and follower engine, the sync engine is responsible for forwarding blocks, but not verifying them. So a `SyncedBlock` message containing an invalid `flow.Block` would be a valid internal message from the `sync` engine. This is a bit of a special case IMO where an external message is forwarded without validation.

But to take another example, suppose HotStuff notifies Engine X about some block B being finalized. HotStuff guarantees that finalized blocks can be queried from the database. If Engine X cannot read B from the database, then the notification from HotStuff could be considered an invalid message and a critical failure.

jordanschalm: Ah I see this section is within the `Engine`-specific appendix, so I'll take your suggestion directly for this section, and add some more detail above: https://github.com/onflow/flow-go/pull/2575/commits/1d1106f74c9f3e18f2a6eb18fa69ae1986f13c2f

franklywatson (approved):

jordanschalm:
bors merge

PR #2577, opened 2022-06-09
Cleanup remote debugger code

SaveTheRbtz:


utils/debug/registerCache.go
```diff
@@ -135,17 +135,5 @@ func (c *fileRegisterCache) Persist() error {
 	if err != nil {
 		return err
 	}
-	err = f.Sync()
-	if err != nil {
-		return err
```
ramtinms, line 140: should we keep one of the syncs here ?

ramtinms (approved):
I think these were introduced by some sort of merge issue.

utils/debug/registerCache.go
SaveTheRbtz: bufio.Writer does require calling `Flush`, everything else does that automagically on Close.

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #2578, opened 2022-06-09
Add state migration for Flow Core Contracts

turbolent:
With the update to Secure Cadence (0.24), which introduces breaking changes, the Core Contracts were updated in:
- https://github.com/onflow/flow-core-contracts/pull/287
- https://github.com/onflow/flow-core-contracts/pull/278

We cannot update these contracts on the current networks as they are incompatible with the currently deployed version of Cadence (deployment will fail with type checks), and we cannot deploy them after the network upgrade, as the core contracts are used during the startup of the Execution Node, before we could send contract update transactions.

Add a state migration which updates the core contracts directly in the execution state.

### DIFF
--- a/cmd/util/ledger/migrations/core_contracts_migration.go
+++ b/cmd/util/ledger/migrations/core_contracts_migration.go
@@ -1,6 +1,7 @@
 package migrations
 
 import (
+	"encoding/hex"
 	"fmt"
 	"strings"
 
@@ -225,13 +226,13 @@ func (m *CoreContractsMigration) Migrate(payloads []ledger.Payload) ([]ledger.Pa
 	addresses := coreContractAddresses(m.Chain)
 	codes := coreContractCodes(addresses)
 
-	for _, payload := range payloads {
+	for i, payload := range payloads {
 		key := string(payload.Key.KeyParts[2].Value)
 		if !strings.HasPrefix(key, codeKeyPrefix) {
 			continue
 		}
 
-		owner := string(payload.Key.KeyParts[0].Value)
+		owner := hex.EncodeToString(payload.Key.KeyParts[0].Value)
 
 		addressCodes, ok := codes[owner]
 		if !ok {
@@ -246,7 +247,7 @@ func (m *CoreContractsMigration) Migrate(payloads []ledger.Payload) ([]ledger.Pa
 			continue
 		}
 
-		payload.Value = newCode
+		payloads[i].Value = newCode
 
 		m.Log.Info().Msgf("Updated core contract: %s.%s", owner, contractName)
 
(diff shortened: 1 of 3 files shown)


cmd/util/ledger/migrations/core_contracts_migration.go
```diff
@@ -0,0 +1,265 @@
...
+		newCode, ok := addressCodes[contractName]
+		if !ok {
+			m.Log.Warn().Msgf("unknown contract: %s", contractName)
+			continue
+		}
+
+		payloads[i].Value = newCode
+
+		m.Log.Info().Msgf("Updated core contract: %s.%s", owner, contractName)
+
```
ramtinms, line 253: do we want to log the diff between old code and new code to a file or log in case we have a unexpected change?

turbolent: Good idea, I'll add that!

turbolent: Added in [dca8edb](https://github.com/onflow/flow-go/pull/2578/commits/dca8edbe3e192e5a30509ba76a190da8e1cde578)

cmd/util/ledger/migrations/core_contracts_migration.go
```diff
@@ -0,0 +1,325 @@
...
+		"FlowIDTableStaking",
+		coreContracts.FlowIDTableStaking(
+			// fungibleTokenAddress:
+			addresses.fungibleToken,
+			// flowTokenAddress:
+			addresses.flowToken,
+			// flowFeesAddress:
+			addresses.fees,
+			// latest:
+			true,
```
ramtinms, line 163: I don't know what this boolean is

ramtinms (approved):
Looks good to me, we should just remember to check the csv logs after migration.

cmd/util/ledger/migrations/core_contracts_migration.go
turbolent: https://github.com/onflow/flow-core-contracts/blob/842c892f75675133fa77720ac46ae0dfdd2ddceb/lib/go/contracts/contracts.go#L176-L177

jordanschalm (approved):
I confirmed the contracts to upgrade, the addresses, and constructor inputs ([this diff](https://github.com/onflow/flow-go/pull/2578/files#diff-23aca7d1117c460b8d68502a17e3eee9f97cac3ce955ec61dff0d1654c5196fbR47-R151)). 

Did not review the migration code.

SupunS (approved):
Mostly reviewed the migration code, and looks good!

PR #2581, opened 2022-06-09
Add larger buckets for metrics for memory estimates

dsainati1:
The histogram buckets we previously used for measuring memory estimates had too low of a cap, so this adds more buckets with a higher cap to allow us to see more granulator data.


j1010001 (approved):

ramtinms (approved):

ramtinms:
please just make sure to include the changes to the master as well, we might forget to do that.

PR #2585, opened 2022-06-09
[FVM] remove the use of the legacy controller by the FVM registers

ramtinms:
This PR  
- migrates the remaining FVM-controlled registers to not reuse owner value as controller. 
- logs warning if we find more cases of controller still being in used
- simplifies the code in the account and removes the need for `isContoller` flag

This is a prework to remove the controller from storage and keys.


janezpodhostnik (approved):
Nice cleanup!

fxamacker (approved):
Nice work!  :+1:   Just some minor comments.

fxamacker:
Maybe this PR needs to be labeled as "Breaking Change"?

cmd/util/ledger/migrations/legacy_controller_migration.go
```diff
@@ -0,0 +1,59 @@
...
+				}
+				// case - contract names
+				if string(key) == state.KeyContractNames {
+					p.Key.KeyParts[1].Value = []byte("")
+					continue
+				}
+				// case - contracts
+				if bytes.HasPrefix(key, []byte(state.KeyCode)) {
+					p.Key.KeyParts[1].Value = []byte("")
+					continue
```
turbolent, line 47: The code in all blocks is the same, maybe combine the cases.
Also, do the keys need to be allocated byte slices? Can they just be `nil`?

```suggestion
				if string(key) == state.KeyPublicKeyCount ||
					bytes.HasPrefix(key, []byte("public_key_")) ||
					string(key) == state.KeyContractNames ||
					bytes.HasPrefix(key, []byte(state.KeyCode)) {
					
					p.Key.KeyParts[1].Value = nil
					continue
```

turbolent (commented):
Thank you for cleaning this up!

Maybe we can also take this opportunity to get rid of the controller?

ramtinms:
Getting rid of the controller in storage is the next step that we are going to do in the following PRs.

cmd/util/ledger/migrations/legacy_controller_migration.go
ramtinms: I didn't know if we need allocation or nil would work but using an empty slice was the way FVM was setting those values, so I tried to prevent accidental issues in this PR, anyway we are going to kill that controller field soon.

ramtinms: updated the conditions to reuse the code.

PR #2587, opened 2022-06-10
[util] update cmd to get seal for sealed block

zhangchiqing:
With the new seal index added, when using util cmd to query seals by block ID, it's more useful to query the seal for the given block , instead of the highest in fork seal.


jordanschalm (approved):

peterargue (approved):

zhangchiqing:
bors merge

PR #2591, opened 2022-06-10
[State Sync] Lookup ExecutionResult directly instead of using index (v0.26)

peterargue:
This pulls https://github.com/onflow/flow-go/pull/2566 into the v0.26 branch

### DIFF
--- a/module/state_synchronization/requester/execution_data_requester.go
+++ b/module/state_synchronization/requester/execution_data_requester.go
@@ -116,6 +116,7 @@ type executionDataRequester struct {
 	// Local db objects
 	headers storage.Headers
 	results storage.ExecutionResults
+	seals   storage.Seals
 
 	executionDataReader *jobs.ExecutionDataReader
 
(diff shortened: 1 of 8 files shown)


smnzhu (approved):

module/state_synchronization/requester/execution_data_requester.go
```diff
@@ -395,7 +397,7 @@ func (e *executionDataRequester) processFetchRequest(ctx irrecoverable.SignalerC
 		ctx.Throw(err)
 	}
 
-	logger.Debug().Msg("Fetched execution data")
+	logger.Info().Msg("execution data fetched")
```
vishalchangrani, line 400: nit: do we really want info? the log message doesn't prints out any details.
Dont know how frequently will we see this, but if it is very frequent then may be put it back to Debug?

peterargue: it will happen for every block, but it does print the blockID and other info

vishalchangrani (approved):

PR #2595, opened 2022-06-10
Fix reversed numbers for script metrics

dsainati1:
We were passing these values to these logging functions in the wrong order


fvm/fvm_bench_test.go
```diff
@@ -270,7 +270,7 @@ type txWeights struct {
 	TXHash                string `json:"txHash"`
 	LedgerInteractionUsed uint64 `json:"ledgerInteractionUsed"`
 	ComputationUsed       uint   `json:"computationUsed"`
-	MemoryUsed            uint   `json:"memoryUsed"`
+	MemoryEstimate        uint   `json:"memoryUsed"`
```
janezpodhostnik, line 273: this relies on the log in `fvm/transactionInvoker.go` so it should now change to `MemoryEstimate        uint   `json:"memoryEstimate"`

j1010001 (approved):

janezpodhostnik (approved):

PR #2570, opened 2022-06-07
[Crypto] performance improvement of SSWU map to curve

tarakby:
The crypto library has been using Relic's map-to-curve implementation of [SSWU algorithm](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#section-6.6.2).

Specific optimizations could be implemented in the case of mapping to BLS12 381 (A == 0, and p mod 4 = 3) to improve performance. The optimizations are detailed in the IETF draft ([optimized SSWU](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#appendix-F.2) and [square root ratio](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#appendix-F.2.1.2)).

This PR implements the optimized SSWU for BLS12 381 by updating an older implementation of SSWU (map [coefficient z](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#appendix-H.2) updated from -1 to 11 as per the IETF draft) as well as optimizing the computation steps. 

Other updates:
- remove hardcoded constants and rely on Relic for SSWU and BLS12 381 constants (z, isogeny map, prime field).
- cache extra constants required for SSWU (isogeny map constants, Montgomery constants)
- separate bench function for scalar multiplication of generator and generic points
- add a new test vector from the IETF draft 

Bench results on my laptop show about 35% improvement:
```
name       old time/op  new time/op  delta
MapToG1-8   364µs ± 1%   232µs ± 1%  -36.14%  (p=0.000 n=30+29)
```

PR #2598, opened 2022-06-10
crypto: allocate on stack where possible

SaveTheRbtz:
This replaces some of the `malloc` calls w/ on-stack allocations.

As a side benefit this also fixes a couple of memory leaks:
* `bn_randZr_star` leaking `seed`.
* `ep_sum_vector_byte` leaking `sigs` on failure.

Also fixed a syntax error around `ep_free(sig[i])` but since this macro expands to nothing in our code base, it does not really matter. 

Followup for the improvements in #2570


tarakby (commented):
Thanks for the updates, and for catching the memory leak case!

I turned some comments into a suggestion PR, feel free to review it and/or merge it : https://github.com/onflow/flow-go/pull/2602

crypto/bls12381_utils.c
```diff
@@ -583,30 +573,35 @@ void ep_sum_vector(ep_t jointx, ep_st* x, const int len) {
...
 int ep_sum_vector_byte(byte* dest, const byte* sigs_bytes, const int len) {
+    int error;
+
     // temp variables
     ep_t acc;        
     ep_new(acc);
     ep_set_infty(acc);
-    ep_st* sigs = (ep_st*) malloc(len * sizeof(ep_st));
+    ep_st* sigs = (ep_st*) calloc(len, sizeof(ep_st));
+    for (int i=0; i < len; i++) ep_new(sigs[i]);
```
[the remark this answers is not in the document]
tarakby: I didn't add this to the suggestions PR, I let you decide what version you prefer.

crypto/bls12381_utils.c
```diff
@@ -181,7 +181,7 @@ void ep2_print_(char* s, ep2_st* p) {
 void bn_randZr_star(bn_t x) {
     // reduce the modular reduction bias
     int seed_len = BITS_TO_BYTES(Fr_BITS + SEC_BITS);
```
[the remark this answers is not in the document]
SaveTheRbtz: Moved it to the define in `utils.h`

crypto/bls12381_utils.h
```diff
@@ -25,10 +25,12 @@ typedef uint8_t byte;
 #define Fp_BITS   381
 #define Fr_BITS   255
 #define Fp_BYTES  BITS_TO_BYTES(Fp_BITS)
-#define Fp_BYTES2 BITS_TO_BYTES(Fp_BITS)*2
+#define Fp2_BYTES (2*Fp_BYTES)
 #define Fp_DIGITS BITS_TO_DIGITS(Fp_BITS)
 #define Fr_BYTES  BITS_TO_BYTES(Fr_BITS)
 
+#define SEED_BYTES BITS_TO_BYTES(Fr_BITS + SEC_BITS)
```
tarakby, line 32: There are many seeds in the library and this length is particular for that `bn_randZr_star` usage. I think it would be better to keep it as a local seed length to that function.

tarakby (commented):
There are more changes related to `Fp2_BYTES` in the suggestions PR https://github.com/onflow/flow-go/pull/2602. Would it possible to merge that PR into your branch/cherry-pick the changes?

crypto/bls12381_utils.c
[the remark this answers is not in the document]
tarakby: I agree this approach doesn't work well for all cases.

tarakby (approved):

crypto/bls12381_utils.c
```diff
@@ -580,33 +570,38 @@ void ep_sum_vector(ep_t jointx, ep_st* x, const int len) {
...
 // The function assumes sigs is correctly allocated with regards to len.
 int ep_sum_vector_byte(byte* dest, const byte* sigs_bytes, const int len) {
+    int error;
+
     // temp variables
     ep_t acc;        
     ep_new(acc);
     ep_set_infty(acc);
-    ep_st* sigs = (ep_st*) malloc(len * sizeof(ep_st));
+    ep_st* sigs = (ep_st*) calloc(len, sizeof(ep_st));
```
durkmurder, line 582: Not sure why are we using `calloc`. I don't see that we rely on 0 initialized memory, in this case malloc will be faster since calloc does a `memset(0)` which is usually slower.

SaveTheRbtz: The main reason for using `calloc` here is to protect against `len * sizeof(ep_st)` overflow.  If we consider that it won't practically happen in our setup (64 bit, limited `len`) then I can likely leave `malloc` here.

(there is a separate issue for verifying the `malloc` returns against `NULL` in the codebase)

tarakby: I actually missed this line change during my review. 
The results of my napkin-math says that if there is overflow, the input `sigs_bytes` should point to a block of data of 3 EiB (Exbibytes), which is more than 3,000,000 Terabytes. 

I would rather avoid `memset` for every small data, than support the feature for such unrealistic size. I think `malloc` seems fine in this case.

SaveTheRbtz: re: `calloc` it was designed to help with security of 32bit platforms since it does fail on multiplication overflow, `malloc(3)`:

> If the multiplication of nmemb and size would result in integer
> overflow, then calloc() returns an error.  By contrast, an
> integer overflow would not be detected in the following call to
> malloc(), with the result that an incorrectly sized block of
> memory would be allocated:
> malloc(nmemb * size);

As for the allocation on stack, i'm a bit scared about it, unless the upper bound is in the low kilobytes: even with a pretty large stack there is no way to guarantee that this function won't be called from a recursive function, which there are some in our codebase.

tarakby: Overflow in `malloc` happens if `len * sizeof(ep_st)` exceeds 64 bits while `len` is less than 64 bits. `calloc` takes `len` as a separate input without the need to multiply.

We do no have an upper bound for `len` though we could add one for safety. At the same time, values of `len` creating issues do not seem practical.

crypto/bls12381_utils.c
```diff
@@ -580,33 +570,38 @@ void ep_sum_vector(ep_t jointx, ep_st* x, const int len) {
...
 // The function assumes sigs is correctly allocated with regards to len.
 int ep_sum_vector_byte(byte* dest, const byte* sigs_bytes, const int len) {
+    int error;
+
     // temp variables
     ep_t acc;        
     ep_new(acc);
     ep_set_infty(acc);
     ep_st* sigs = (ep_st*) malloc(len * sizeof(ep_st));
+    for (int i=0; i < len; i++) ep_new(sigs[i]);
```
pattyshack, line 583: nit: wrap for body in `{}`

pattyshack (approved):

crypto/bls12381_utils.c
SaveTheRbtz: a very good suggestion: would do a mass-rewrite in a separate diff to programmatically apply this.  Not forcing `{}` on lone ifs and fors is a disaster waiting to happen.

SaveTheRbtz:
bors merge

crypto/bls12381_utils.c
tarakby: I had a quick look at recommended coding styles: 
- [GNU conventions](https://www.gnu.org/prep/standards/html_node/Syntactic-Conventions.html#Syntactic-Conventions) seems to allow not including `{}` in some cases. 
- [Indian hill](http://www.literateprogramming.com/indhill-cstyle.pdf) coding standard also seems to include one line statements without braces. 

Happy to hear your thoughts on why you prefer to enforce braces.

SaveTheRbtz:
bors retry

crypto/bls12381_utils.c
SaveTheRbtz: The main reasons for using braces are usually:
* less error prone: although quite small, but there is still non-zero probability of accidentally trying to add a line to a single-line if or for.  If we can avoid that with a clang-format rule -- imho we should do that.
* diffs are cleaner: if you need to add code to an existing single line diff you don't bloat the diff.

Also, given that our primary language (go) forces brackets, so forcing them here should make code more familiar.

PS. Kernel style guides (both linux and BSD's `style(9)` do allow single line control statements, but here I think it is bad to rely on them, since their coding styles were invented way before auto-code formatting with `go fmt`/`python-black`/`clang-format` became mainstream.)

PR #2603, opened 2022-06-10
Update dependency modules version and update BLS DSTs

tarakby:
(This PR is a replacement of the larger closed PR https://github.com/onflow/flow-go/pull/2603: the work had to be done over a few steps: (1) update the crypto module (2) update the core-contract module (3) update flow-go using the new versions of crypto and core-contract modules)

- Restructure BLS domain separation tags (other than ciphersuites) to comply with :
   - IETF hash to curve drat v14 (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14)
   - IETF BLS signature draft v4 (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04)
- update the crypto module version to use the new BLS ciphersuite.
- update the core-contract module to the use the new domain separation tags.

 From now on, when using BLS signatures in the Flow protocol with a certain domain tag `DST`, the full domain tag would be:
 `FLOW-DST-V00-CS00-with-BLS_SIG_BLS12381G1_XOF:KMAC128_SSWU_RO_POP_`
 
 Side changes:
  - domain tags are moved to `module/signature` package instead of `model/encoding`. 
  - `NewBLSKMAC` is renamed to `NewBLSHasher` to abstract the expand_message function details from the protocol. `NewBLSHasher` is a wrapper around the expand_message function provided by the crypto library.


engine/execution/state/bootstrap/bootstrap_test.go
```diff
@@ -47,7 +47,7 @@ func TestBootstrapLedger(t *testing.T) {
 }
 
 func TestBootstrapLedger_ZeroTokenSupply(t *testing.T) {
-	expectedStateCommitmentBytes, _ := hex.DecodeString("f04b063fa45655110ff10ee3153726366966efc8ebcdc730d27b9971eca5cedf")
+	expectedStateCommitmentBytes, _ := hex.DecodeString("e4b748da1215452fcaa60f5549d261bf3f76a8a6e9a8388a6d8ee73193b4e331")
```
ramtinms, line 50: hmm, is this expected? did the execution state changed because of the way keys are encoded?

AlexHentschel:
@ramtinms good question. My assumption would be the following: 
* in the old code, the Domain Separation Tag is constructed as follows: https://github.com/onflow/flow-go/blob/ec5aedb4b3e98cca74829d5780a37aeef547d1c2/model/encoding/signing_tags.go#L10-L15
* while in the new code we have: https://github.com/onflow/flow-go/blob/1e03eb5ca57abc1c14e7545878aeef48a3cf9dae/module/signature/signing_tags.go#L17-L34

As the domain separation tag is cryptographically also baked into the Hash, changing the tag would imply that the hash values should change.

AlexHentschel (approved):
very nice.

engine/execution/state/bootstrap/bootstrap_test.go
tarakby: Good question, that seems to be caused by a core-contract update (domain tag is updated on a contract and this PR uses the new core-contract repo version). My guess is that since the contract is deployed to the service account, the state commitment has changed. Do you think that's a valid reason?

utils/unittest/execution_state.go
```diff
@@ -24,7 +24,7 @@ const ServiceAccountPrivateKeySignAlgo = crypto.ECDSAP256
 const ServiceAccountPrivateKeyHashAlgo = hash.SHA2_256
 
 // Pre-calculated state commitment with root account with the above private key
-const GenesisStateCommitmentHex = "479a97c8c3b0056c68718a3a40aab67bc6cc9db0b2d4d2e19a771ca79c8e1b5a"
+const GenesisStateCommitmentHex = "3e84f2892fae21fc922c322b4a86c10c330012fdda678437baedad0c09e05e0b"
```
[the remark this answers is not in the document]
tarakby: Same answer as above. This test failed only when I updated the core-contract version.

tarakby:
bors merge

PR #2510, opened 2022-05-28
feat(localnet): use grafana, tempo, loki

SaveTheRbtz:
This diff switches localnet to the production-grade monitoring stack:
* Grafana (metrics)
* Tempo (Tracing)
* Loki (Logs)

Here is an example of plotting the rate of slow transactions in the logs and going to the traces for the log line:
Screen Shot 2022-05-28 at 12 36 46 AM

Also pre-provisioned a dashboard based on "Bechnet1 - General":
Screen Shot 2022-06-01 at 9 13 04 PM
Screen Shot 2022-06-01 at 9 13 40 PM

While here, removed logstash, kibana, elasticsearch, and jaeger.

Plan:
- [x] see whether promtail can be easily used instead of a [docker log driver](https://grafana.com/docs/loki/latest/clients/docker-driver/).
  - [x] if not, make driver installation automatic.
- [x] pre-create a localnet benchmark dashboard.
- [x] lock docker images versions so they won't accidentally break under us on upgrade.
- [x] fix documentation.

PR #2609, opened 2022-06-13
localnet: switch from docker log driver to promtail

SaveTheRbtz:
This should allow both perf automation and loki to use logs at the same time.

Fixes: #2510

### DIFF
--- a/integration/localnet/Makefile
+++ b/integration/localnet/Makefile
@@ -53,21 +53,14 @@ init-light:
 init-short-epochs:
 	$(MAKE) -e EPOCHLEN=200 STAKINGLEN=10 DKGLEN=50 init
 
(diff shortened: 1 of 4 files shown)


simonhf:
Tested this out with the existing load test automation scripts here [1] and it appears to work :-)

There is however another issue discovered but listed that here [2].

[1] https://www.notion.so/dapperlabs/localnet-and-loader-automation-on-June-13th-2022-7d6a6c54df3f4c36969d9134723340c6
[2] https://github.com/onflow/flow-go/pull/2510#issuecomment-1154371573

Kay-Zee (approved):

SaveTheRbtz:
bors merge

PR #2610, opened 2022-06-13
fix logging current statecommitment in execution state extraction

zhangchiqing:
The state extraction util will run the migration and then generate a export-state.json file which includes the statecommitment before and after the migration. 

However, there is a bug that the statecommitment in the exported json file is wrong, it is showing the same statecommitment as the one before the migration. See comment about why it's wrong.

This PR fixes it.

### DIFF
--- a/cmd/util/cmd/execution-state-extract/execution_state_extract.go
+++ b/cmd/util/cmd/execution-state-extract/execution_state_extract.go
@@ -86,7 +86,6 @@ func extractExecutionState(
 			extractionReportName: reporters.NewExportReporter(log,
 				chain,
 				func() flow.StateCommitment { return targetHash },
-				func() flow.StateCommitment { return flow.StateCommitment(newState) },
 			),
 			"account": &reporters.AccountReporter{
 				Log:   log,
@@ -101,7 +100,7 @@ func extractExecutionState(
 		}
 	}
 
-	newState, err = led.ExportCheckpointAt(
+	migratedState, err := led.ExportCheckpointAt(
 		newState,
 		migrations,
 		rs,
@@ -116,8 +115,8 @@ func extractExecutionState(
 
 	log.Info().Msgf(
 		"New state commitment for the exported state is: %s (base64: %s)",
-		newState.String(),
-		newState.Base64(),
+		migratedState.String(),
+		migratedState.Base64(),
 	)
 
 	return nil
--- a/cmd/util/ledger/reporters/export_reporter.go
+++ b/cmd/util/ledger/reporters/export_reporter.go
@@ -54,7 +54,7 @@ func (e *ExportReporter) Name() string {
 func (e *ExportReporter) Report(payload []ledger.Payload, commit ledger.State) error {
 	script, _, err := ExecuteCurrentEpochScript(e.chain, payload)
 	failedExportReport := ExportReport{
-		ReportSucceeded: true,
+		ReportSucceeded: false,
 	}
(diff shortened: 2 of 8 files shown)


cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -86,7 +86,6 @@ func extractExecutionState(
 			extractionReportName: reporters.NewExportReporter(log,
 				chain,
 				func() flow.StateCommitment { return targetHash },
-				func() flow.StateCommitment { return flow.StateCommitment(newState) },
```
zhangchiqing, line 89: This is the bug. 

This callback is supposed to read the newState after the migration unless the callback is called AFTER the migration and the newState is overwritten by L104.

The callback was called after the migration, but before the newState is returned to be used to overwrite thew `newState`. Therefore,the callback is still reading the old value.

To fix it, we simply pass the statecommitment to each reporter, so each reporter can decide whether to use it or not.

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -101,7 +100,7 @@ func extractExecutionState(
 		}
 	}
 
-	newState, err = led.ExportCheckpointAt(
+	migratedState, err := led.ExportCheckpointAt(
```
zhangchiqing, line 103: rename the variable to avoid overwritten.

cmd/util/ledger/reporters/export_reporter.go
```diff
@@ -92,7 +89,7 @@ func (e *ExportReporter) Report(payload []ledger.Payload) error {
 	report := ExportReport{
 		EpochCounter:            script.Value.ToGoValue().(uint64),
 		PreviousStateCommitment: e.getBeforeMigrationSCFunc(),
-		CurrentStateCommitment:  e.getAfterMigrationSCFunc(),
+		CurrentStateCommitment:  flow.StateCommitment(commit),
```
zhangchiqing, line 92: This takes the migrated statecommit, and convert it into statecommitment

cmd/util/ledger/reporters/export_reporter.go
```diff
@@ -33,28 +33,25 @@ type ExportReporter struct {
 	logger                   zerolog.Logger
 	chain                    flow.Chain
 	getBeforeMigrationSCFunc GetStateCommitmentFunc
-	getAfterMigrationSCFunc  GetStateCommitmentFunc
```
zhangchiqing, line 36: Since we removed the callback, we don't need it here either.

ledger/complete/ledger.go
```diff
@@ -461,13 +461,13 @@ func (l *Ledger) keepOnlyOneTrie(state ledger.State) error {
...
 
-func runReport(r ledger.Reporter, p []ledger.Payload, l zerolog.Logger) error {
+func runReport(r ledger.Reporter, p []ledger.Payload, commit ledger.State, l zerolog.Logger) error {
 	l.Info().
 		Str("name", r.Name()).
 		Msg("starting reporter")
 
 	start := time.Now()
-	err := r.Report(p)
+	err := r.Report(p, commit)
```
zhangchiqing, line 470: pass the root has to reporter

sjonpaulbrown (approved):

jordanschalm (approved):

zhangchiqing:
bors merge

vishalchangrani (approved):

zhangchiqing:
bors merge

PR #2618, opened 2022-06-14
Add addresses for Mainnet to Core Contract state migration

turbolent:
Just like we updated the Core Contracts during the Testnet spork, we also need to update them during the Mainnet spork.

I took the addresses from https://docs.onflow.org/core-contracts/. Please verify these are correct.

### DIFF
--- a/cmd/util/ledger/migrations/core_contracts_migration.go
+++ b/cmd/util/ledger/migrations/core_contracts_migration.go
@@ -63,8 +63,21 @@ func coreContractAddresses(chain flow.Chain) networkAddresses {
 			stakingProxy:      "7aad92e5a0715d21",
 		}
 	case flow.Mainnet:
-		// TODO:
-		panic("TODO")
+		return networkAddresses{
+			fungibleToken:     "f233dcee88fe0abe",
+			flowToken:         "1654653399040a61",
+			contractAudits:    "e467b9dd11fa00df",
+			serviceAccount:    "e467b9dd11fa00df",
+			storageFees:       "e467b9dd11fa00df",
+			fees:              "f919ee77447b7497",
+			qc:                "8624b52f9ddcd04a",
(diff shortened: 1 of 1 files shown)


cmd/util/ledger/migrations/core_contracts_migration.go
```diff
@@ -63,8 +63,21 @@ func coreContractAddresses(chain flow.Chain) networkAddresses {
...
+			serviceAccount:    "e467b9dd11fa00df",
+			storageFees:       "e467b9dd11fa00df",
+			fees:              "f919ee77447b7497",
+			qc:                "8624b52f9ddcd04a",
+			epoch:             "8624b52f9ddcd04a",
+			dkg:               "8624b52f9ddcd04a",
+			stakingTable:      "8624b52f9ddcd04a",
+			lockedTokens:      "8d0e87b65159ae63",
+			stakingCollection: "8d0e87b65159ae63",
+			stakingProxy:      "62430cf28c26d095",
```
zhangchiqing, line 79: wasn't sure about this, but else look good

zhangchiqing: [Actually found it here. ](https://github.com/onflow/flow-core-contracts/blob/master/lib/go/templates/cmd/manifest/main.go#L39)

zhangchiqing (approved):

cmd/util/ledger/migrations/core_contracts_migration.go
turbolent: It's also on the documentation page, just a bit hidden, it also took me a while to find it: 
https://docs.onflow.org/core-contracts/locked-tokens/#stakingproxy

janezpodhostnik (approved):

joshuahannan (approved):

PR #2620, opened 2022-06-14
[Consensus] Active Pacemaker: Timeout Collector implementation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6199

### Context

This PR implements `TimeoutCollector` and `TimeoutProcessor` which will be integrated into `VoteAggregator`.
Generally it will look in next way:

`TimeoutAggregator <- TimeoutCollectors <- TimeoutCollector <- TimeoutProcessor`

Design is similar to `VoteCollector` and `VoteProcessor`. 
Some key points:
- `TimeoutCollector` has single state comparing to `VoteCollector` it doesn't perform any caching before validation, all timeouts are processed immediately 
- `TimeoutProcessor` implements validation logic of timeout objects, it performs all compliance checks and signature checks.
- `TimeoutProcessor` uses callbacks to notify subscribers about valid TCs and QCs which were checked as part of timeout validation logic.
- `TimeoutProcessor` tracks two thresholds, `f+1` and `2f+1`, both are reported using callbacks.
- Same as vote collection, collector and processor will be used by `TimeoutAggregator` which implements work balancing and concurrency support.


consensus/hotstuff/timeout_collector.go
```diff
@@ -0,0 +1,50 @@
...
+type OnNewQCDiscovered func(*flow.QuorumCertificate)
+
+// OnNewTCDiscovered is a callback which will be called to notify about new validated TC.
+// This callback is called from processing logic.
+type OnNewTCDiscovered func(*flow.TimeoutCertificate)
+
+// TimeoutCollector collects all timeout objects for a specified view. On the happy path, it
+// generates a TimeoutCertificate when enough timeouts have been collected.
+type TimeoutCollector interface {
+	// AddTimeout adds a timeout object to the collector
```
AlexHentschel, line 28:
```suggestion
	// AddTimeout adds a Timeout Object [TO] to the collector.
```

consensus/hotstuff/timeout_collector.go
```diff
@@ -0,0 +1,50 @@
...
+	// When f+1 TOs will be collected then callback for partial TC will be triggered,
+	// after collecting 2f+1 TOs a TC will be created and passed to the EventLoop.
```
AlexHentschel, line 30: for all our implementations, we assume the max possible number of byzantine nodes. Using `f` to represent the number of malicious nodes is uncommon in our code base, in particular because nodes are weighted.

```suggestion
	// When TOs from strictly more than 1/3 of consensus participants (measured by weight)
	// were collected, the callback for partial TC will be triggered.
	// After collecting TOs from a supermajority, a TC will be created and passed to the EventLoop.
```

consensus/hotstuff/timeout_collector.go
```diff
@@ -0,0 +1,50 @@
...
+// This callback is called from processing logic.
+type OnNewQCDiscovered func(*flow.QuorumCertificate)
+
+// OnNewTCDiscovered is a callback which will be called to notify about new validated TC.
+// This callback is called from processing logic.
+type OnNewTCDiscovered func(*flow.TimeoutCertificate)
+
+// TimeoutCollector collects all timeout objects for a specified view. On the happy path, it
+// generates a TimeoutCertificate when enough timeouts have been collected.
+type TimeoutCollector interface {
```
AlexHentschel, line 27: Would suggest to include a comment that implementations of `TimeoutCollector` are concurrency safe.

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,298 @@
...
+		return true
+	}
+	return false
+}
+
+// NewestQCTracker is a helper structure which keeps track of the highest QC(by view)
+// in concurrency safe way.
+type NewestQCTracker struct {
+	lock     sync.RWMutex
+	newestQC *flow.QuorumCertificate
```
AlexHentschel, line 43: I am not sure if it will work (but it might be an interesting experiment nevertheless):
* I think we might be able to use [`atomic.Value`](https://pkg.go.dev/sync/atomic#Value) here . It implements `CompareAndSwap`, which could be sufficient for our algorithm. My hope is that an atomic would be more performant than a mutex and less prone to congestion.  
* Only downside of using `atomic.Value` is that it doesn't handle `nil`. So we would need to initialize with a QC. While we always have a QC, it might bloat the code too much to add it to all the constructors.

AlexHentschel (commented):
first set of comments

consensus/hotstuff/timeout_collector.go
```diff
@@ -0,0 +1,50 @@
...
+// This callback is called from processing logic.
+type OnPartialTCCreated func(view uint64)
+
+// OnNewQCDiscovered is a callback which will be called to notify about new validated QC.
+// This callback is called from processing logic.
+type OnNewQCDiscovered func(*flow.QuorumCertificate)
+
+// OnNewTCDiscovered is a callback which will be called to notify about new validated TC.
+// This callback is called from processing logic.
+type OnNewTCDiscovered func(*flow.TimeoutCertificate)
```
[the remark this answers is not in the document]
durkmurder: works for me

consensus/hotstuff/model/timeout.go
```diff
@@ -53,3 +54,7 @@ type TimeoutObject struct {
 func (t *TimeoutObject) ID() flow.Identifier {
 	return flow.MakeID(t)
```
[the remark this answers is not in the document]
AlexHentschel: thanks. Looks great!

AlexHentschel:
Just looked at [this code](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_cache.go#L56-L73) in the `TimeoutObjectsCache` and was wondering if we should optimize it:
* At the moment, we do the [error handling](https://github.com/onflow/flow-go/blob/94397095c12935a5f89067496272666395fd7a33/consensus/hotstuff/timeoutcollector/timeout_cache.go#L68-L71) inside the lock. Unfortunately, the `ID` computations for `TimeoutObject`s can be quite expensive, because they include a QC and optionally a TC.
* I am worried about lock congestion, specifically in attack scenarios. 

I think we could release the lock right away in the `if exists {` branch, which would reduce the attack surface slightly. I feel like you had it initially like this Yurii, and I might have requested to change it to the current version 😅. Sorry, if that was the case 🙈.

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,295 @@
...
+
+	signerIndices, err := p.signerIndicesFromIdentities(signers)
+	if err != nil {
+		return nil, fmt.Errorf("could not encode signer indices: %w", err)
+	}
+
+	return &flow.TimeoutCertificate{
+		View:          p.view,
+		NewestQCViews: highQCViews,
+		NewestQC:      p.NewestQCTracker.NewestQC(),
```
AlexHentschel, line 277: ## ❗ potential race condition ❓ 

I am not entirely sure, but to me this looks like a potential edge case, where a race condition could lead to constructing an invalid TC. Reasoning:
* In the `Validator`, we require that the QC included in a TC satisfies `QC.View` must be equal to `max(TC.NewestQCViews)`:  https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/validator/validator.go#L88-L97
* Specifically, this means that it is _illegal_ to include a newer QC or an older QC
* I think this code is vulnerable to a race condition: https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L137-L151
   * I managed to sketch an argument in my head that including a newer QC is impossible. This is because the newest QC known to the `NewestQCTracker` must previously have been added to the `sigAggregator`: https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L146-L151 
   * However, I don't think the reverse holds. Lets consider the following situation:
      * There is only one honest node (lets call that node 'Alice') that knows the highest QC for view 100. All other nodes only know the lower QC for view 99.
      * We receive TOs from enough nodes to construct a TC, without Alice's TO. 
      * `Thread 1` is in [line 145](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L145) with the _last_ TO (_not_ Alice's TO) needed to construct a TC.
      * `Thread 2` also sits in [line 145](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L145) with Alice's TO.
      * `Thread 1` gets to go first. It adds the last needed TO to `sigAggregator` and the returned `totalWeight` meets the minimal threshold for the first time. Therefore, `Thread 1` passes [line 160](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L160. It reaches [line 163](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L163) and pauses there. 
      * Now, `Thread 2` executes [line 146](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L146), adding Alice's TO to `sigAggregator`. But `Thread 2` pauses in [line 150](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L150).
      * At this point, `sigAggregator` knows that there is a TO for view 100 (Alice's TO). But the TO has not been added to the `NewestQCTracker`.
      * Finally, `Thread 1` enters [`p.buildTC()`](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L164), aggregates the signatures and includes view 100 in the `TC.NewestQCViews`. However, `p.NewestQCTracker.NewestQC()` only knows a QC for view 99, which we include in the TC. 

Hence, we have constructed an invalid TC ☠️

AlexHentschel (commented):
Next batch of comments. Mostly suggestions for polishing comments and other aesthetics. 
However, there is one algorithmic edge case, which might be non-trivial to fix.

consensus/hotstuff/timeoutcollector/timeout_processor.go
AlexHentschel: From my perspective, the root cause for this edge case is that tracking the views and the highest QC is implemented as a non-atomic operation: https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_processor.go#L146-L151

I can think of two different ways around this problem:
* In both proposals, we would switch the order of operations:
    ```golang
    p.NewestQCTracker.Track(timeout.NewestQC)

    totalWeight, err := p.sigAggregator.VerifyAndAdd(timeout.SignerID, timeout.SigData, timeout.NewestQC.View)
    if err != nil {
      return fmt.Errorf("could not process invalid signature: %w", err)
    }
    ```
   Then, `NewestQCTracker` could know of a newer QC than `sigAggregator`
* _Proposal 1_:  
   Based on my understanding, the HotStuff proofs require that a TC includes a QC that has view _at least as new_ as `max(TC.NewestQCViews)`. However, I think it _might_ be possible to allow a **newer** QC, with view strictly higher than any value in `TC.NewestQCViews`. Though, this is only a hypothesis; we would need to go over the proofs in detail to confirm.
* _Proposal 2_:  
   The other option would be that `NewestQCTracker` does not only track the QC with largest view, but stores _all_ QCs that were included in TOs for the respective view. Well, technically we only need one QC for each view. For example, `NewestQCTracker` could maintain a map: `map[uint64]QuorumCertificate`. In `buildTC()`, we would then retrieve the specific QC that matches `max(highQCViews)`.

consensus/hotstuff/consumer.go
```diff
@@ -180,3 +194,42 @@ type QCCreatedConsumer interface {
...
+
+	// OnPartialTcCreated notifications are produced by the TimeoutProcessor
+	// component, whenever it aggregates enough timeouts(1/3+1) to claim that TC eventually will be constructed
+	// Prerequisites:
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnPartialTcCreated(view uint64)
+
+	// OnNewQcDiscovered notifications are produced by the TimeoutCollector
+	// component, whenever it discovers new QC included in timeout object.
```
jordanschalm, line 223:
```suggestion
	// OnNewQcDiscovered notifications are produced by a TimeoutCollector for view v,
	// whenever it discovers a new QC embedded in a TimeoutObject for view v, iff the
	// the new QC has a view greater than any other QC observed in a TimeoutObject for view v.
	// 
```
This is my understanding of how this notification would behave, since each `TimeoutCollector` (view-scoped) will make independent assessments of their local newest QC.
* I'm assuming here that instances of `TimeoutCollectorConsumer` are intended to be global (not view-scoped), based on how the consumer for `QCCreatedConsumer` works

❓  If this is right, I'm wondering if we have an intended use for these notifications currently? They don't seem that useful to me right now, since we will not have QCs ordered by view (even less strict guarantees on ordering than the caveat Alex mentioned, I think), and we will have duplicate QCs in the notifications.

jordanschalm (commented):
First round of comments, still reviewing

consensus/hotstuff/consumer.go
AlexHentschel: ## Answer to question 1: 
> `TimeoutCollectorConsumer` is intended to be global (not view-scoped)

Correct.

## My thoughts regarding question 2:
> Do we have an intended use for these notifications?

From my perspective, we strongly desire these notifications. 

### Context:
* `TimeoutObject`s contain QCs and TCs. In the HotStuff liveness proofs, we require that these QCs and TCs -- specifically the newest TC and QC -- are visible to the HotStuff state machine (in our implementation the `EventHander`) to perform the necessary view changes. 
* The `TimeoutObject`s arrive at the [`TimeoutAggregator`](https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeout_aggregator.go), which is its own component, which runs concurrently to the `EventLoop`. 
   *  I think from a theoretical perspective, liveness is guaranteed if the newest QCs and TCs are transmitted to the `EventHandler` as part of a TC or a `partialTC`. 
   * However, waiting for a TC or a `partialTC` could have a large penalty in practise: 
       * Consider the case where node `Bob` is primary for view `100`.  View 99 has timed out. But Bob is behind still stuck in view 98. Maybe he just recovered from a crash and missed most of the TOs for view 99. 
       * Now, when the first node in view 100 times out and sends a TO, the information contained therein would allow Bob to transition to view 100, quickly construct his block and broadcast it. Hopefully in time that a supermajority of nodes still get the proposal, vote for it and the view completes successfully.
       * However,  if the `TimeoutAggregator` does not expose new information to the `EventHandler` asap, view 100 will likely time out. By the time Bob's `TimeoutAggregator` constructs a `partialTC`, too many nodes have timed out so no QC can be constructed for Bob's proposal anymore. 
* In a nutshell, `TimeoutAggregator` receives QCs and TCs that the `EventHandler` needs to transition to the newest view and contribute to consensus. Waiting for a `partialTC` to propagate this information adds latency. The extra latency can result extra views timing out. I think this might be a relatively likely scenario when the committee is operating close to its timeout threshold. 
* Hence, we desire a _fast_ channel over which the `TimeoutAggregator` can push the _newest_ QCs or TCs to the `EventHander`. This channel is implemented via `OnNewQcDiscovered` and `OnTcConstructedFromTimeouts`

### Completeness: newest QC and TC is eventually delivered and _never_ lost
Reasoning:
* based on the context above, the newest QC and TC must reach `EventHander` (skipping QCs or TCs with lower view is fine)
* It is straight to show that the current implementation will _always_ deliver the newest QC and TC to the `EventHander`. QCs and TCs for older views might be interjected rarely; however the `EventHander` can just drop older QCs or TCs by inspecting their view alone. 

### Responsiveness: newest QC and TC is _quickly_ delivered with negligible delay 
Reasoning:
* Our logic only only generates a `OnNewQcDiscovered ` notification if the QC's view is strictly larger than any QC that was sent before: https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/timeoutcollector/timeout_collector.go#L81-L87 In other words, for each view, at most a single `OnNewQcDiscovered ` is sent. Hence, it is extremely unlikely for the notification channel to become congested with notifications about old QCs while the newest QC is delayed. 
* similar argument applies to TC

Summary:
We have reasoned that `OnNewQcDiscovered` and `OnTcConstructedFromTimeouts` are desired for the `TimeoutAggregator` to quickly push new QCs and TCs to the `EventHander` to improve responsiveness of the consensus protocol. Furthermore, we have sketched proofs that the current implementation completely delivers the necessary information without noteworthy latency.

AlexHentschel: ## Follow-up comment
**Caution:** I think the current implementation of `OnPartialTcCreated` is potentially insufficient to guarantee liveness: https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/consumer.go#L220

@durkmurder

AlexHentschel: ## Follow-up comment 2

We also have to think about how we handle a QC for a block that we haven't received yet.

consensus/hotstuff/notifications/pubsub/timeout_collector_consumer.go
```diff
@@ -0,0 +1 @@
+package pubsub
```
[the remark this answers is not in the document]
durkmurder: yup, I will add impl for it.

model/flow/quorum_certificate.go
```diff
@@ -22,6 +22,22 @@ type QuorumCertificate struct {
...
+		SignerIndices []byte
+		SigData       []byte
+	}{
+		View:          t.View,
+		BlockID:       t.BlockID,
+		SignerIndices: t.SignerIndices,
+		SigData:       t.SigData,
+	}
+	return MakeID(body)
+}
```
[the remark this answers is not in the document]
durkmurder: yup, you are right, they are identical

consensus/hotstuff/timeoutcollector/timeout_processor.go
durkmurder: Thanks for spotting it, really good observation, very nice. I have a feeling proposal 1 is perfectly fine, for safety properties in all proofs we build proofs based on values of `TC.NewestQCViews`. we will vote for proposal if `QC.View >= max(TC.NewestQCViews)`. We already allow including newer QC into proposal. 
Let's consider a case where TC includes QC: `TC.NewestQC.View > max(Tc.NewestQCViews)`. 
1. Leader includes includes that QC into proposal, this satisfies our rule `QC.View >= max(TC.NewestQCViews)` -> it's safe to vote for this proposal.
2. Leader doesn't include that QC into proposal, then `QC.View >= max(TC.NewestQCViews)` holds, but this proposal is not safe to vote for since we fail this [check](https://github.com/onflow/flow-go/blob/feature/active-pacemaker/consensus/hotstuff/safetyrules/safety_rules.go#L257)

Based on proofs we require only that `QC.View >= max(TC.NewestQCViews)` but in our implementation we need to make sure that we are extending the newest QC available.

consensus/hotstuff/timeoutcollector/timeout_processor.go
durkmurder: https://github.com/onflow/flow-go/pull/2620/commits/e9d18cc4d891953b9a3b747d7b439707429dfac7

consensus/hotstuff/timeoutcollector/timeout_collector_test.go
```diff
@@ -0,0 +1,186 @@
...
+	require.NoError(s.T(), err)
+
+	otherTimeout := helper.TimeoutObjectFixture(helper.WithTimeoutObjectView(s.view),
+		helper.WithTimeoutObjectSignerID(timeout.SignerID))
+
+	s.notifier.On("OnDoubleTimeoutDetected", timeout, otherTimeout).Once()
+
+	err = s.collector.AddTimeout(otherTimeout)
+	require.NoError(s.T(), err)
+	s.notifier.AssertCalled(s.T(), "OnDoubleTimeoutDetected", timeout, otherTimeout)
```
jordanschalm, line 87:
```suggestion
	s.notifier.AssertExpectations(s.T())
```
nit: can use `AssertExpectations` to avoid duplicating the `On` call

consensus/hotstuff/consumer.go
jordanschalm: Thanks for the explanation Alex, it makes perfect sense now 👌 

> > We also have to think about how we handle a QC for a block that we haven't received yet.
> 
> Can you elaborate more on this topic?

Here is my understanding:
* Previously, the only way for a new QC to enter HotStuff was as part of a `BlockProposal`. Since The compliance layer guarantees that parent blocks are processed before children, we could be certain that we know about the block pointed to by any QC which is processed by HotStuff. 
* However `TimeoutObject`s are processed separately without a similar ordering constraint, so we may process a `TimeoutObject` containing a QC for block `B`, before we process block `B`
  * This would cause a (currently unexpected) error [when adding the QC to `Forks`](https://github.com/onflow/flow-go/blob/1e0a30e7a9118aad96a706639e37ec955f562542/consensus/hotstuff/forks/forkchoice/newest.go#L93)
* This is similar to the [discussion about processing blocks containing TCs on the Implementation doc](https://www.notion.so/Active-Pacemaker-Implementation-Draft-f654eaa990954508908304ba03428a24?d=3029169ae9b4468396226451bd63d02c#3888981b6b5646e6aed0a2a7df3a5bb4)

I think ultimately the problem is, we can process QCs for block `B` without having processed `B`. But this complicates our proposal and voting logic, because:
* we would like to build on top of our highest known QC, but we can't propose a block on top of a block `B` if we haven't validated the payload for block `B`
* we can't vote for a proposal building on top of block `B` if we haven't validated the payload for block `B` yet

I see two options:
1. we can avoid processing QCs (or messages containing QCs) referencing an unknown block, eg. by queueing messages at the 
2. we can optimistically process QCs, but voting and proposing logic (and internal HotStuff state) needs to be aware of the possibility of a QC for an unknown block

To me, Option 1 is preferable because it limits more edge cases within HotStuff and is consistent with the approach taken for TCs (linked discussion) of optimizing for happy-path of voting or proposing.

consensus/hotstuff/timeoutcollector/timeout_processor.go
jordanschalm: Another option: we could embed the functionality of `NewestQCTracker` within `TimeoutSignatureAggregator` and update both `NewestQC` and `NewestQCViews` atomically in the locked section here:
https://github.com/onflow/flow-go/blob/338166875960fde9396caeb670445a6f16df6eb2/consensus/hotstuff/timeoutcollector/aggregation.go#L126-L131

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,292 @@
...
+		return model.NewInvalidTimeoutErrorf(timeout, "TO's QC %d cannot be newer than the TO's view %d",
+			timeout.NewestQC.View, timeout.View)
+	}
+
+	// (b) If a TC is included, the TC must be for the past round, no matter whether a QC
+	//     for the last round is also included. In some edge cases, a node might observe
+	//     _both_ QC and TC for the previous round, in which case it can include both.
+	if timeout.LastViewTC != nil {
+		if timeout.View != timeout.LastViewTC.View+1 {
+			return model.NewInvalidTimeoutErrorf(timeout, "invalid TC for previous round")
```
jordanschalm, line 194:
```suggestion
			return model.NewInvalidTimeoutErrorf(timeout, "invalid TC for non-previous view, expected view %d, got view %d", timeout.View-1, timeout.LastViewTC.View)
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		if timeout.NewestQC.View < timeout.LastViewTC.NewestQC.View {
+			return model.NewInvalidTimeoutErrorf(timeout, "timeout.NewestQC has older view that the QC in timeout.LastViewTC")
```
jordanschalm, line 197:
```suggestion
			return model.NewInvalidTimeoutErrorf(timeout, "timeout.NewestQC is older (view=%d) than the QC in timeout.LastViewTC (view=%d)", timeout.NewestQC.View, timeout.LastViewTC.NewestQC.View)
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,292 @@
...
+	//     to round timeout.View is possible either by observing a QC or a TC for the previous round.
+	//     If no QC is included, we require a TC to be present, which by check (1b) must be for
+	//     the previous round.
+	lastViewSuccessful := timeout.View == timeout.NewestQC.View+1
+	if !lastViewSuccessful {
+		// The TO's sender did _not_ observe a QC for round timeout.View-1. Hence, it should
+		// include a TC for the previous round. Otherwise, the TO is invalid.
+		if timeout.LastViewTC == nil {
+			return model.NewInvalidTimeoutErrorf(timeout, "timeout must include TC")
+		}
```
jordanschalm, line 210: I would suggest to move the [check on lines 193-195](https://github.com/onflow/flow-go/pull/2620/files#diff-88341aa1d744505d6d69c70fb72b2f2ec2ceda5bd61c69de860e145c4dc6dbfdR193-R195) here:
* it's easier to verify `(c)` when all the relevant checks are together
* it's the same structure as we use for with [similar validation with proposals](https://github.com/onflow/flow-go/blob/9f9d03bdd0ef107ed565dfab00e1aae7879ff1fb/consensus/hotstuff/validator/validator.go#L225-L228)

I'm also wondering, why do we not define this validation function in `Validator` alongside eg `ValidateQC`? To me it makes more sense for timeout validation to live alongside validation of the other entities.

model/flow/timeout_certificate.go
```diff
@@ -21,22 +21,24 @@ type TimeoutCertificate struct {
...
 		NewestQCViews []uint64
-		NewestQC      QuorumCertificate
+		NewestQCID    Identifier
 		SignerIndices []byte
 		SigData       crypto.Signature
 	}{
 		View:          t.View,
 		NewestQCViews: t.NewestQCViews,
-		NewestQC:      *t.NewestQC,
+		NewestQCID:    t.NewestQC.ID(),
```
jordanschalm, line 39: ⚠️  `QuorumCertificate` doesn't handle computing the ID of a `nil` QC, like we implemented for `TimeoutCertificate`, so we will panic if we compute the `ID` of a `TimeoutCertificate`. 

This is equivalent to assuming that we will always validate a `TimeoutCertificate` before computing its ID. I'm not convinced this is safe in general. For example, suppose we don't know the block for `tc.NewestQC` so we decide to cache the TC. Then we receive a new TC and want to see if it already exists in our cache, so we compare its ID to the cached TC's ID - that would cause a crash.

jordanschalm (commented):
Next set of comments, still need to review `timeout_process_test.go`

consensus/hotstuff/consumer.go
AlexHentschel: > Lemma: Bracha broadcast is need to ensure that we produce a TC as fast as possible by immediately timing out current view, this ensures lowest possible time to enter round `r` within `2*Δ` after GST. 

This is precisely the argument we need to look at. Thanks.

I think the current implementation is _not_ sufficient to guarantee that lowest possible time to enter round `r` within `2*Δ` after GST. Reasoning:
* consider the following situation: 
   * a replica is collecting enough timeouts for view 100 such that `TimeoutAggregator` emits a `OnPartialTcCreated` notification
   * however, the replica's `EventHander` is still in view 99 (it hasn't seen the necessary evidence to transition to view 100 yet)
* `OnPartialTcCreated` for view 100 arrives at the `EventHander`, the event hander calls into `SafetyRules`  https://github.com/onflow/flow-go/blob/90b4f79a88d5bcfdbfa11f07d80e7144df59cf40/consensus/hotstuff/safetyrules/safety_rules.go#L123 
   * But neither the QC nor the TC allow for switching to view 100.
   * Hence, `SafetyRules` will not generate a `TimeoutMessage` and   `EventHander` will not enter the view in time.
  
I would conclude that our implementation does not guarantee validity of the Lemma.

AlexHentschel: > > We also have to think about how we handle a QC for a block that we haven't received yet.
>
> Can you elaborate more on this topic? I am not quite sure what do you exactly mean.

exactly what [@jordanschalm said](https://github.com/onflow/flow-go/pull/2620#discussion_r902007880).

>I see two options:
> 1. we can avoid processing QCs (or messages containing QCs) referencing an unknown block, eg. by queueing messages before they are processed by `TimeoutAggregator`.
> 2. we can optimistically process QCs, but voting and proposing logic (and internal HotStuff state) needs to be aware of the possibility of a QC for an unknown block.

I think both options generate undesired edge cases. My gut feeling is that option 2 might be better. With option 2, we might halt the generation of TCs locally, if only a few `TimeoutObject`s refer to unknown blocks (since they are cached and don't reach the `TimeoutAggregator`).

I still have the vague hope that we can adjusting the `EventHander` might be less complex. The node can't propose or vote if it doesn't have sufficiently new blocks, but it can at least participate in the TC production. Eventually a node that knows the highest block will become primary (existence of a QC implies that a supermajority of nodes know the block) and propose a proper extension. 
Cycling through leaders requires an intact timeout mechanism; hence my concern with withholding messages from the `TimeoutAggregator`

consensus/hotstuff/timeoutcollector/timeout_processor.go
AlexHentschel: this block of code is only executed if the QC was _not_ for the previous round: https://github.com/onflow/flow-go/blob/f8ca29d13a8a4d2829de244c89810450f1462568/consensus/hotstuff/timeoutcollector/timeout_processor.go#L204-L205
In case the the last view _was_ successful (i.e. `lastViewSuccessful == true`), we would still want to check the TC. But we would just ignore the TC in this case, which I am concerned about from a BFT perspective. 

I honestly feel the current logic is correct (though I admit not the most intuitive):
(a) if a TC is included, it must be for the last view (independently of the QC)
(b) if the QC is not for the prior view, the TC cannot be nil 

The important aspect is that condition (a) must be universally enforced.

model/flow/timeout_certificate.go
AlexHentschel: agree, I think it would be fine to also define ID for a nil QC.

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+			helper.MakeTC(helper.WithTCView(s.view-1),
+				helper.WithTCNewestQC(helper.MakeQC(helper.WithQCView(s.view-1))))),
+	)
+
+	s.Run("invalid-signer", func() {
+		*s.committee = *mocks.NewReplicas(s.T())
+		s.committee.On("IdentityByEpoch", mock.Anything, mock.Anything).Return(nil, model.NewInvalidSignerError(errors.New(""))).Once()
+		err := s.processor.Process(timeout)
+		require.True(s.T(), model.IsInvalidTimeoutError(err))
+	})
```
jordanschalm, line 162: I'd suggest to add a test case that, if `IdentityByEpoch` returns `ErrViewForUnknownEpoch`, we should:
* expect an error from `Process`
* the error should not be of type `ErrViewForUnknownEpoch` or `InvalidTimeoutError` (ie. not an expected error)

Note: the second point is not reflected in the implementation yet, though I think we should update the implementation to be consistent with this test description (see [this comment](https://github.com/onflow/flow-go/pull/2620#discussion_r902017497))

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+		s.validator.On("ValidateTC", timeout.LastViewTC).Return(model.ErrViewForUnknownEpoch).Once()
+
+		err := s.processor.Process(timeout)
+		require.False(s.T(), model.IsInvalidTimeoutError(err))
+		require.NotErrorIs(s.T(), err, model.ErrViewForUnknownEpoch)
+	})
+}
+
+// TestProcess_ValidTimeout tests that processing a valid timeout succeeds without error
+func (s *TimeoutProcessorTestSuite) TestProcess_ValidTimeout() {
```
jordanschalm, line 251: It would be great to test more cases of a valid TO:
* a TO with `LastViewTC` for last view and QC for older view
* a TO with QC for last view, and nil `LastViewTC`
* a TO with both QC for last view, and `LastViewTC` for last view

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+			startupWg.Wait()
+			err := s.processor.Process(timeout)
+			require.NoError(s.T(), err)
+		}(i, timeout)
+	}
+
+	startupWg.Done()
+
+	// wait for all routines to finish
+	shutdownWg.Wait()
```
jordanschalm, line 382:
```suggestion
	shutdownWg.Wait()
	
	s.notifier.AssertExpectations(s.T())
```
⚠️  I think we need this, otherwise we aren't sure whether the TC was actually created

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+		timeout := helper.TimeoutObjectFixture(
+			helper.WithTimeoutObjectView(s.view),
+			helper.WithTimeoutNewestQC(newestQC),
+			helper.WithTimeoutObjectSignerID(signer.NodeID),
+			helper.WithTimeoutLastViewTC(nil),
+		)
+		go func(i int, timeout *model.TimeoutObject) {
+			defer shutdownWg.Done()
+			startupWg.Wait()
+			err := s.processor.Process(timeout)
```
jordanschalm, line 374: ⚠️  I think we have a race condition here, if multiple goroutines modify `s.totalWeight` concurrently. `totalWeight` is not concurrency-safe, but `VerifyAndAdd` is used concurrently by `TimeoutProcessor`.

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+		require.NoError(t, err)
+		err = voteProcessor.Process(vote)
+		require.NoError(t, err)
+	}
+
+	require.NotNil(t, newestQC, "vote processor must create a valid QC at this point")
+
+	// create verifier that will do crypto checks of created TC
+	verifier := verification.NewStakingVerifier()
+	// create validator which will do compliance and crypto checked of created TC
```
jordanschalm, line 455:
```suggestion
	// create validator which will do compliance and crypto checks of created TC
```

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+	var lastViewTC *flow.TimeoutCertificate
+	onTCCreated := func(args mock.Arguments) {
+		tc := args.Get(0).(*flow.TimeoutCertificate)
+		// check if resulted TC is valid
+		err := validator.ValidateTC(tc)
+		require.NoError(t, err)
+		lastViewTC = tc
+	}
+
+	aggregator, err := NewTimeoutSignatureAggregator(view, stakingSigners, encoding.CollectorVoteTag)
```
jordanschalm, line 467: Should we have a separate tag for timeout objects compared to votes?

consensus/hotstuff/timeoutcollector/timeout_processor.go
jordanschalm: Just to clarify, I agree the current logic is correct - my suggestion to move the checks around is about readability.

AlexHentschel: 🙈 Sorry, probably misunderstood. I only wanted to highlight that moving [lines 193 - 195](https://github.com/onflow/flow-go/blob/f8ca29d13a8a4d2829de244c89810450f1462568/consensus/hotstuff/timeoutcollector/timeout_processor.go#L192-L199) _inside_ the body of the [`if !lastViewSuccessful` statement](https://github.com/onflow/flow-go/blob/f8ca29d13a8a4d2829de244c89810450f1462568/consensus/hotstuff/timeoutcollector/timeout_processor.go#L205-L211) would create an additional edge case that is now properly handled.

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
AlexHentschel: also great catch. Yes, we totally should.

consensus/hotstuff/timeoutcollector/timeout_processor.go
durkmurder: @jordanschalm I am pretty sure we have decided to keep QC tracking and sig aggregation as separate paths, maybe we need to revisit this

model/flow/timeout_certificate.go
durkmurder: Good catch, I've updated ID() to work with nil QC.

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
```diff
@@ -0,0 +1,502 @@
...
+// TestProcess_TimeoutNewerHighestQC tests that TimeoutProcessor fails with model.InvalidTimeoutError if
+// timeout contains a QC with QC.View > timeout.View, QC can be only with lower view than timeout.
+func (s *TimeoutProcessorTestSuite) TestProcess_TimeoutNewerHighestQC() {
+	err := s.processor.Process(helper.TimeoutObjectFixture(helper.WithTimeoutObjectView(s.view),
+		helper.WithTimeoutNewestQC(helper.MakeQC(helper.WithQCView(s.view)))))
+	require.True(s.T(), model.IsInvalidTimeoutError(err))
+}
+
+// TestProcess_LastViewTCRequiredButNotPresent tests that TimeoutProcessor fails with model.InvalidTimeoutError if
+// timeout contains a proof that sender legitimately entered timeout.View but it has wrong view meaning he used TC from previous rounds.
```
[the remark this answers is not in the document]
durkmurder: Yup, mistake, will fix.

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
durkmurder: I've added new tags for consensus and collector timeouts

consensus/hotstuff/timeoutcollector/timeout_processor_test.go
jordanschalm: That's handy - Didn't know about that feature 👌

jordanschalm (approved):
Nice work 🎸

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,300 @@
...
+	if p.tcTracker.Done() {
+		return nil
+	}
+
+	err := p.validateTimeout(timeout)
+	if err != nil {
+		return fmt.Errorf("received invalid timeout: %w", err)
+	}
+
+	p.newestQCTracker.Track(timeout.NewestQC)
```
AlexHentschel, line 143: Lets add some documentation to explain that it is critical for correctness that `newestQCTracker` is updated _before_ `sigAggregator`

```suggestion
	// CAUTION: for correctness it is critical that we update the `newestQCTracker` first, _before_ we add the
	// TO's signature to `sigAggregator`. Reasoning:
	//  * For a valid TC, we require that the TC includes a QC with view ≥ max{TC.NewestQCViews}.
	//  * The `NewestQCViews` is maintained by `sigAggregator`.
	//  * Hence, for any view `v ∈ NewestQCViews` that `sigAggregator` knows, a QC with equal or larger view is
	//    known to `newestQCTracker`. This is guaranteed if and only if `newestQCTracker` is updated first. 
	p.newestQCTracker.Track(timeout.NewestQC)
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,300 @@
...
+		return fmt.Errorf("received incompatible timeout, expected %d got %d: %w", p.view, timeout.View, ErrTimeoutForIncompatibleView)
+	}
+
+	if p.tcTracker.Done() {
+		return nil
+	}
+
+	err := p.validateTimeout(timeout)
+	if err != nil {
+		return fmt.Errorf("received invalid timeout: %w", err)
```
AlexHentschel, line 140: Not all errors imply that the timeout is invalid. For unexpected exceptions, I think this error message would be misleading

```suggestion
		return fmt.Errorf("validating timeout failed: %w", err)
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,300 @@
...
+	}
```
AlexHentschel, line 141: validating the timeout can be comparatively time intensive. Therefore, I think it might be beneficial to check `tcTracker.Done()` here again before we proceed:
```suggestion
	}
	if p.tcTracker.Done() {
		return nil
	}
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
AlexHentschel: Result of discussion: 
* we'll implement Proposal 1 (allowing `QC.View >= max(TC.NewestQCViews)` as opposed to requiring strict equality)

consensus/hotstuff/consumer.go
AlexHentschel: Result of discussion:
* we'll implement option 2 (QCs and TCs are forwarded to EventHander, which progresses through views but might not vote if parent block is unknown).

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,300 @@
...
+	p.notifier.OnTcConstructedFromTimeouts(tc)
+
+	return nil
+}
+
+// validateTimeout performs validation of timeout object, verifies if timeout is correctly structured
+// and included QC and TC is correctly structured and signed.
+// ATTENTION: this function doesn't check if timeout signature is valid, this check happens in signature aggregator
+// Expected error returns during normal operations:
+// * model.InvalidTimeoutError - submitted invalid timeout(invalid structure or invalid signature)
```
AlexHentschel, line 178: > invalid structure or **invalid signature**
we not verify the signature here. Following my earlier suggestion to neither check the signer ID nor the signature here:

```suggestion
// ATTENTION: this function does _not_ check whether the TO's `SignerID` is an authorized node nor if 
// the signature is valid. These checks happens in signature aggregator.
// Expected error returns during normal operations:
// * model.InvalidTimeoutError - submitted invalid timeout
```

consensus/hotstuff/timeoutcollector/timeout_collector.go
```diff
@@ -0,0 +1,114 @@
...
+		newestReportedQC:  counters.NewMonotonousCounter(0),
+		newestReportedTC:  counters.NewMonotonousCounter(0),
+	}
+}
+
+// AddTimeout adds a Timeout Object [TO] to the collector.
+// When TOs from strictly more than 1/3 of consensus participants (measured by weight)
+// were collected, the callback for partial TC will be triggered.
+// After collecting TOs from a supermajority, a TC will be created and passed to the EventLoop.
+func (c *TimeoutCollector) AddTimeout(timeout *model.TimeoutObject) error {
```
AlexHentschel, line 47:
```suggestion
// Expected error returns during normal operations:
//  * timeoutcollector.ErrTimeoutForIncompatibleView - submitted timeout for incompatible view
// All other exceptions are symptoms of potential state corruption.
func (c *TimeoutCollector) AddTimeout(timeout *model.TimeoutObject) error {
```

AlexHentschel (approved):
Very very nice. Thanks for the thorough tests and the great documentation. Solid design and clean implementation. ✨ 

My comments are mostly minor nitpicks regarding documentation. Only two points that require attention:
* There is one case, where I am not sure if all errors are handled accordingly ([comment here](url))
* And I think the validator does not allow TCs to include a QC with view larger than `max{TC.NewestQCViews}` ([comment here](url))

AlexHentschel:
Sorry, I just realized that [my earlier suggestion here](https://github.com/onflow/flow-go/pull/2620#discussion_r899778173) might not work:
* consider the situation where the network enters view 1
* the highest known QC is for view 0 (the root QC)
* what happens if the leader for view 1 is down? 

I think `NewestQCTracker` might then report the dummy QC 🤦 

I see two options:
1. if we added a method like the following to `hotstuff.Replicas`: https://github.com/onflow/flow-go/blob/dc815618f01e7155e857614d404f7dc34df1231a/state/protocol/cluster.go#L29-L30
then we could initialize the `NewestQCTracker` with a valid QC for a finalized block and avoid these problems. 

   @jordanschalm do you have thoughts how to best get the root QC from? 
2. The easiest might be to revert to your previous version of `NewestQCTracker` with a lock, where you can explicitly check for `nil`. We can hopefully assume that lock contention will not be a problem. 

there is some more discussion [here](https://stackoverflow.com/questions/21447463/is-assigning-a-pointer-atomic-in-go) but it involves using `unsafe.Pointer` 😬

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -0,0 +1,300 @@
...
+// function is implemented to be called by multiple goroutines at the same time.
+// Design of this function is event driven, as soon as we collect enough weight
+// to create a TC or a partial TC we will immediately do so and submit it
+// via callback for further processing.
+// Expected error returns during normal operations:
+// * ErrTimeoutForIncompatibleView - submitted timeout for incompatible view
+// * model.InvalidTimeoutError - submitted invalid timeout(invalid structure or invalid signature)
+// * model.ErrViewForUnknownEpoch if no epoch containing the given view is known
+// All other errors should be treated as exceptions.
+func (p *TimeoutProcessor) Process(timeout *model.TimeoutObject) error {
```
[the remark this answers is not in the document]
durkmurder: Yup, 3) is handled in `validateTimeout` and indeed it's not expected.
1) should be propagated to caller 
2) I will add docs and test for this case, good catch!

durkmurder:
> Sorry, I just realized that [my earlier suggestion here](https://github.com/onflow/flow-go/pull/2620#discussion_r899778173) might not work:
> 
> * consider the situation where the network enters view 1
> * the highest known QC is for view 0 (the root QC)
> * what happens if the leader for view 1 is down?
> 
> I think `NewestQCTracker` might then report the dummy QC facepalm
> 
> I see two options:
> 
> 1. if we added a method like the following to `hotstuff.Replicas`: https://github.com/onflow/flow-go/blob/dc815618f01e7155e857614d404f7dc34df1231a/state/protocol/cluster.go#L29-L30
>    
>    then we could initialize the `NewestQCTracker` with a valid QC for a finalized block and avoid these problems.
>    @jordanschalm do you have thoughts how to best get the root QC from?
> 2. The easiest might be to revert to your previous version of `NewestQCTracker` with a lock, where you can explicitly check for `nil`. We can hopefully assume that lock contention will not be a problem.
> 
> there is some more discussion [here](https://stackoverflow.com/questions/21447463/is-assigning-a-pointer-atomic-in-go) but it involves using `unsafe.Pointer` ([code example](https://www.geeksforgeeks.org/atomic-compareandswappointer-function-in-golang-with-examples/) for `CompareAndSwapPointer` for unsafe pointers.)grimacing

I've made it work with the `unsafe.Pointer`, we can address it later if there are some objections.

durkmurder:
bors merge

durkmurder:
bors merge

durkmurder:
bors retry

PR #2622, opened 2022-06-14
Implement a range limit 

danielholmes839:
Implemented a `selectRangeRequest` method that limits the size of ranges that are requested. Also implemented a check to `onBlockResponse` to limit the number of blocks that can be processed from a single block response


module/synchronization/core.go
```diff
@@ -480,13 +482,53 @@ func (c *Core) getBatches(blockIDs []flow.Identifier) []chainsync.Batch {
 // candidate range and batch requests. Range requests are given precedence and
 // the total number of requests does not exceed the configured request maximum.
 func (c *Core) selectRequests(ranges []chainsync.Range, batches []chainsync.Batch) ([]chainsync.Range, []chainsync.Batch) {
-	max := int(c.Config.MaxRequests)
+	maxRequests := int(c.Config.MaxRequests)
+	maxLength := int(c.Config.MaxRangeLength)
 
-	if len(ranges) >= max {
-		return ranges[:max], nil
+	ranges = c.selectRangeRequests(ranges, maxRequests, maxLength)
```
zhangchiqing, line 488: it's better not to overwrite the original value
```suggestion
	selectedRange := c.selectRangeRequests(ranges, maxRequests, maxLength)
```

zhangchiqing: Instead of selecting a sub range to request, we should probably limit at the "upper stream".

It's ourselves that created `ranges` in the first place, can we limit there?

PR #2625, opened 2022-06-15
Trigger Epoch Fallback Based on Epoch Commitment Deadline

jordanschalm:
This PR modifies the trigger condition for entering epoch fallback mode:

* **Previously:** _incorporating_ the _first block of the next epoch_, before the next epoch is committed, triggers epoch fallback mode.
* **With this PR:** _finalizing_ the _first block beyond the `EpochCommitmentDeadline`_, before the next epoch is committed, triggers epoch fallback mode.


state/protocol/badger/mutator.go
```diff
@@ -404,32 +399,26 @@ func (m *FollowerState) lastSealed(candidate *flow.Block) (*flow.Seal, error) {
...
-	//    (i) Determine protocol state for block's _current_ Epoch.
-	//        As we don't have slashing yet, the protocol state is fully
-	//        determined by the Epoch Preparation events.
-	//   (ii) Determine protocol state for block's _next_ Epoch.
-	//        In case any of the payload seals includes system events,
-	//        we need to check if they are valid and must apply them
-	//        to the protocol state as needed.
-	ops, err := m.handleServiceEvents(candidate)
+	// apply any state changes from service events sealed by this block's parent
+	dbUpdates, insertingBlockTriggersEpochFallback, err := m.handleEpochServiceEvents(candidate)
```
jordanschalm, line 414: Method 1 for triggering epoch fallback: we incorporate a block applying an invalid service event. Even if the block is not finalized, we trigger epoch fallback immediately.

state/protocol/badger/mutator.go
```diff
@@ -526,99 +527,55 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
-	//
-	// We skip updating these metrics when EECC has been triggered
-	if header.View > parentEpochFinalView && !epochFallbackTriggered {
-		events = append(events, func() { m.consumer.EpochTransition(currentEpochSetup.Counter, header) })
+	// Determine metric updates and protocol events related to epoch phase
+	// changes and epoch transitions.
+	// If epoch emergency fallback is triggered, the current epoch continues until
+	// the next spork - so skip these updates.
+	if !epochFallbackTriggered {
+		epochPhaseMetrics, epochPhaseEvents, err := m.epochPhaseMetricsAndEventsOnBlockFinalized(header, epochStatus)
```
jordanschalm, line 552: Refactored epoch-related metrics, events, and state changes into:
* `epochPhaseMetricsAndEventsOnBlockFinalized`
* `epochTransitionMetricsAndEventsOnBlockFinalized`
* `epochFallbackTriggeredByFinalizedBlock`

state/protocol/badger/mutator.go
```diff
@@ -809,49 +920,52 @@ SealLoop:
 				// validate the service event
 				err := isValidExtendingEpochSetup(ev, activeSetup, epochStatus)
 				if protocol.IsInvalidServiceEventError(err) {
-					// EECC - we have observed an invalid service event, which is
-					// an unrecoverable failure. Flag this in the DB and exit
-					ops = append(ops, transaction.WithTx(operation.SetEpochEmergencyFallbackTriggered(blockID)))
+					// we have observed an invalid service event, which triggers epoch fallback mode
+					insertingCandidateTriggersEpochFallback = true
```
jordanschalm, line 924: Rather than returning a database operation function regarding epoch fallback being triggered, now return a boolean (aiming to mimic the pattern of `epochFallbackTriggeredByFinalizedBlock`) and let the upper layer deal with the details of events/metrics/state changes associated with this.

state/protocol/badger/mutator_test.go
```diff
@@ -1207,100 +1207,94 @@ func TestExtendEpochTransitionWithoutCommit(t *testing.T) {
...
-			assertEpochEmergencyFallbackTriggered(t, db)
-
-			// epoch metrics should not be emitted
-			metricsMock.AssertNotCalled(t, "EpochTransition", epoch2Setup.Counter, mock.Anything)
-			metricsMock.AssertNotCalled(t, "CurrentEpochCounter", epoch2Setup.Counter)
+			// since EECC has been triggered, epoch transition metrics should not be updated
+			metricsMock.AssertNotCalled(t, "EpochTransition", mock.Anything, mock.Anything)
+			metricsMock.AssertNotCalled(t, "CurrentEpochCounter", epoch1Setup.Counter+1)
 			metricsMock.AssertExpectations(t)
+			protoEventsMock.AssertExpectations(t)
```
durkmurder, line 1271: I would suggest using [new mockery API](https://github.com/onflow/flow-go/blob/jordan/6227-eecc-trigger-condition/module/mock/compliance_metrics.go#L88) that automatically call `AssertExpectations`

jordanschalm: Done here https://github.com/onflow/flow-go/pull/2625/commits/38ff9cc69a0b0a8459b622fbf540afdfa424e0f2

state/protocol/badger/mutator.go
```diff
@@ -449,14 +438,27 @@ func (m *FollowerState) insert(ctx context.Context, candidate *flow.Block, last
...
 		// apply any optional DB operations from service events
-		for _, apply := range ops {
+		for _, apply := range dbUpdates {
 			err := apply(tx)
 			if err != nil {
 				return fmt.Errorf("could not apply operation: %w", err)
 			}
 		}
 
+		// emit protocol events
```
AlexHentschel, line 457: Follow-up on our discussion: 
* please include TODO's here and create an issue to update the implementation of the service events to be completely asynchronous (with recovery logic where needed)

state/protocol/badger/mutator.go
```diff
@@ -480,8 +483,14 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 	}
 	header := block.Header
 
-	// FIRST: verify that the parent block is the latest finalized block. This
-	// must be the case, as the `Finalize(..)` method only finalizes one block
+	// keep track of metrics updates and protocol events to emit:
+	// * metrics are updated after a successful database update
+	// * protocol events are emitted atomically with the database update
```
AlexHentschel, line 488: include toDo to move events out of the DB transaction

state/protocol/badger/mutator.go
```diff
@@ -508,16 +517,8 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
-	// THIRD: preparing Epoch-Phase-Change service notifications and metrics updates.
-	// Convention:
-	//                            .. <--- P <----- B
-	//                                    ↑        ↑
-	//             block sealing service event        first block of new
-	//           for epoch-phase transition        Epoch phase (e.g.
-	//              (e.g. EpochSetup event)        (EpochSetup phase)
-	// Per convention, service notifications for Epoch-Phase-Changes are emitted, when
-	// the first block of the new phase (EpochSetup phase) is _finalized_. Meaning
-	// that the new phase has started.
```
AlexHentschel, line 520: I feel this old comment https://github.com/onflow/flow-go/blob/4372a12958672917ff8e8b72874a46650163ceb8/state/protocol/badger/mutator.go#L526-L535 is actually quite helpful to retain because we are planning to uniformly wait for finalized seal for the service event

state/protocol/badger/mutator.go
```diff
@@ -526,99 +527,55 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+		}
+		metrics = append(metrics, epochTransitionMetrics...)
+		events = append(events, epochTransitionEvents...)
 	}
 
-	// if EECC is triggered, update metric
+	// if epoch emergency fallback is triggered, update metric
 	if epochFallbackTriggered {
-		m.metrics.EpochEmergencyFallbackTriggered()
+		metrics = append(metrics, m.metrics.EpochEmergencyFallbackTriggered)
```
AlexHentschel, line 569: I feel this logic should be consolidated with this one here: https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/mutator.go#L541-L544

in that way, we append events and metrics updates together, same as [for the code path with `!epochFallbackTriggered`](https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/mutator.go#L551-L565)

state/protocol/badger/mutator.go
```diff
@@ -658,45 +620,204 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+// No errors are expected during normal operation.
+func (m *FollowerState) epochFallbackTriggeredByFinalizedBlock(block *flow.Header, epochStatus *flow.EpochStatus, currentEpochSetup *flow.EpochSetup) (bool, error) {
+
+	// 1. determine whether block B is past the epoch commitment deadline
+	safetyThreshold, err := m.Params().EpochCommitSafetyThreshold()
+	if err != nil {
+		return false, fmt.Errorf("could not get epoch commit safety threshold: %w", err)
+	}
+	epochCommitmentDeadline := currentEpochSetup.FinalView - safetyThreshold
+	blockExceedsDeadline := block.View >= epochCommitmentDeadline
```
AlexHentschel, line 650: Subtractions of view-based quantities are generally prone to underflow. I think it would be completely fine to algebraically change the computation to use addition only, which circumvents the problem entirely: 
```suggestion
	blockExceedsDeadline := block.View+safetyThreshold >= currentEpochSetup.FinalView
```

state/protocol/badger/mutator.go
```diff
@@ -658,45 +620,204 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+) {
+
+	parentBlocksEpoch := m.AtBlockID(block.ParentID).Epochs().Current()
+	parentEpochFinalView, err := parentBlocksEpoch.FinalView()
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not get parent epoch final view: %w", err)
+	}
+
+	if block.View > parentEpochFinalView {
+		events = append(events, func() { m.consumer.EpochTransition(currentEpochSetup.Counter, block) })
```
AlexHentschel, line 688: Would it make sense to add a sanity check here to verify that `currentEpochSetup` is actually for epoch `parentBlocksEpoch.Counter()+1`? This check has marginal runtime overhead and would catch already a broad class of bugs.

AlexHentschel (approved):
looks good. Overall, I have three structural suggestion, which we already discussed in our face-to-face meeting today:
1. I think it would be beneficial to only act on EpochService events once they have a finalized seal. In particular, this would include invalid service events. See [my comment here ](url)for more details. 
2. In the file [`badger/validity.go`](https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/validity.go), we have mixed patterns of error handling. In particular
   * [isValidExtendingEpochSetup`](https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/validity.go#L25) returns sentinel error when validating the `extendingSetup` Epoch setup event. It also consumes auxiliary inputs, which we assume are correct but theoretically could trigger sanity checks to fail.
   * In comparison, [`extendingSetup`](https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/validity.go#L181) has the same method signature: it takes as input an Epoch Commit event whose validity is to be verified and a prior `EpochSetup` even that is assumed to be valid. However, here we only return no typed sentinel error and attribute all errors as the Epoch Commit event being invalid. 
3. We propagate the undesired pattern of emitting notifications within badger transactions: https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/mutator.go#L457-L460 We consider this pattern technical deb:
   * we decided to add TODO's in this PR for cleaning this up later
   * please also create an issue 🙏 🙇

state/protocol/badger/mutator.go
```diff
@@ -404,32 +399,26 @@ func (m *FollowerState) lastSealed(candidate *flow.Block) (*flow.Seal, error) {
...
+// insert stores the candidate block in the database.
+// The `candidate` block _must be valid_ (otherwise, the state will be corrupted).
+// dbUpdates contains other database operations which must be applied atomically
+// with inserting the block.
 func (m *FollowerState) insert(ctx context.Context, candidate *flow.Block, last *flow.Seal) error {
 
 	span, _ := m.tracer.StartSpanFromContext(ctx, trace.ProtoStateMutatorExtendDBInsert)
 	defer span.Finish()
 
 	blockID := candidate.ID()
```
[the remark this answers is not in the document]
jordanschalm: included in https://github.com/onflow/flow-go/pull/2747

state/protocol/badger/mutator.go
jordanschalm: Created issue: https://github.com/dapperlabs/flow-go/issues/6316

state/protocol/badger/mutator.go
jordanschalm: Created issue: https://github.com/dapperlabs/flow-go/issues/6317

state/protocol/badger/mutator.go
jordanschalm: Added TODOs: https://github.com/onflow/flow-go/pull/2625/commits/a562d1f08f0d4c1811887edab1ba309c9230d994

state/protocol/badger/mutator.go
jordanschalm: It's still there, just modified a bit and moved into the method where the notifications and metrics updates are generated: https://github.com/onflow/flow-go/blob/a562d1f08f0d4c1811887edab1ba309c9230d994/state/protocol/badger/mutator.go#L710-L719

state/protocol/badger/mutator.go
jordanschalm: Good point. 

The reason I had this case separated was to ensure the metric still gets set properly following a restart (where we would not enter the first block on line 541). However, this is already handled in the initialization code:
https://github.com/onflow/flow-go/blob/c60050541bb5eaef19cad610232d562bd80e39de/state/protocol/badger/state.go#L627-L635

Updated in https://github.com/onflow/flow-go/pull/2625/commits/0b38f8e9c28670420942a94e41e0805457cf4252

state/protocol/badger/mutator.go
jordanschalm: Addressed in https://github.com/onflow/flow-go/pull/2747

state/protocol/badger/mutator.go
jordanschalm: 👍 Good idea, added in https://github.com/onflow/flow-go/pull/2625/commits/b74d85fbc7b799d808b86770e275ae9b25e64657

jordanschalm:
> In the file [badger/validity.go](https://github.com/onflow/flow-go/blob/ca33e9e09c6b2bf2fe8edb71a3a5504d850fbc0a/state/protocol/badger/validity.go), we have mixed patterns of error handling.

Updated all validity functions which have trusted inputs to return sentinel error. https://github.com/onflow/flow-go/pull/2625/commits/fdb8c50243f46091b19bb53aa246e36b4fcff612

For `verifyEpochSetup` I left it returning generic errors, all of which must indicate the input is invalid.

PR #2629, opened 2022-06-15
Add exponential backoff for execution data download timeouts

peterargue:
When an ExecutionData request fails, bitswap does not return until the context times out. This means if we naively wait for the full max timeout, there is a large delay whenever there's an issue with a peer or the network. However, we can't just make the timeout short either in case there is a large blob that takes a while to download.

This PR adds an incremental backoff for the timeout, and configuration to control the settings.


smnzhu (approved):

SaveTheRbtz (approved):

peterargue:
bors merge

PR #2630, opened 2022-06-15
Revert "1706 refactor network codec as argument"

peterargue:
Reverts onflow/flow-go#2491

This PR was causing unexpected unauthorized errors on localnet. reverting so we can investigate further.


simonhf (approved):
LGTM :-)

This [1] is the thread where the errors were first reported by me and reproduced by Peter.

[1] https://axiomzen.slack.com/archives/C015G65HR2P/p1655333011804609

simonhf:
Looks like no need to revert because the following changes to localnet make the warnings and errors go away :-)

```
$ git diff bootstrap.go 
-               "--supports-unstaked-node=true",
+               "--supports-unstaked-node=false",
```

PR #2631, opened 2022-06-16
Handle race conditions around stopping

pattyshack:
- AddTx can crash if Stop succeed before AddTx is invoked
- Stop and statusWorker can livelock if number of transactions > txs channel's maxCap

### DIFF
--- a/integration/utils/tx_tracker.go
+++ b/integration/utils/tx_tracker.go
@@ -37,6 +37,7 @@ type txCallbacks struct {
 // transactions by txID, in case of transaction state change
 // it calls the provided callbacks
 type TxTracker struct {
+	ctx           context.Context
 	cancel        context.CancelFunc
 	log           zerolog.Logger
 	numbOfWorkers int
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
This is a good diff!  Thanks for fixing my bugs!  And alternative proposal would be just removing that code so that we have less stuff to support: #2632. (basically relying only on `TxFollower`).  WDYT?

integration/utils/tx_tracker.go
```diff
@@ -219,8 +227,12 @@ func (txt *TxTracker) statusWorker(ctx context.Context, workerID int, fclient *c
 			// don't put too much pressure on access node
 			time.Sleep(sleepAfterOp)
 
-			// put it back
-			txt.txs <- tx
+			select {
+			case txt.txs <- tx:
```
pattyshack, line 231: This can still livelock in general, when # txn > maxCap.   is there a strong reason why we need to use a limited worker pool instead of just using a goroutine per txn?  if it's only for rate limiting, it may be better to switch to a goroutine per txn, and use a semaphore for rate limiting.

pattyshack:
less is more

PR #2632, opened 2022-06-16
loader: remove TxTracker and TxStatsTracker

SaveTheRbtz:
Currently we have two ways of tracking transactions in the loader:
* TxTracker, which periodically calls `GetTransactionResult` **per transaction inflight** to get state of transactions until it reaches either error state or seals.
* TxFollower -- it tails blocks and notifies callers when their transactions appear in a collection.

Since the latter code is more performant I would propose consolidating everything there.

Also, while here, since `TxStatsTracker` is only used in `TxTracker` and we do not even emit that data anywhere.
Replaced it with a prometheus metric and a histogram.  Put them on the localnet's grafana dashboard as a PoC:

Screen Shot 2022-06-15 at 3 45 06 PM


integration/utils/follower.go
```diff
@@ -136,12 +142,16 @@ Loop:
 				if ch, loaded := f.txToChan.LoadAndDelete(tx.Hex()); loaded {
 					txi := ch.(txInfo)
 
+					duration := time.Since(txi.submisionTime)
 					f.logger.Trace().
-						Dur("duration", time.Since(txi.submisionTime)).
+						Dur("durationInMS", duration).
```
pattyshack, line 147: Why is this in milliseconds, and the prometheus stats in seconds?

SaveTheRbtz: Sadly, `Dur` uses the  `zerolog.DurationFieldUnit` unit for measuring latencies.  But the time to execution block time is about 9 seconds (this is the number of roundtrips between the block stages multiplied by the fixed length of the sage).  Therefore I felt that measuring value that will likely be always greater than a second in milliseconds was quite silly.  Can switch it back for consistency though.

module/metrics/loader.go
```diff
@@ -23,6 +28,17 @@ func NewLoaderCollector() *LoaderCollector {
 			Namespace: namespaceLoader,
 			Help:      "transactions per second that the loader should send",
 		}),
+		transactionsExecuted: promauto.NewCounter(prometheus.CounterOpts{
+			Name:      "transactions_executed",
+			Namespace: namespaceLoader,
+			Help:      "transaction successfully executed by the loader",
+		}),
+		tteInSeconds: promauto.NewHistogram(prometheus.HistogramOpts{
+			Name:      "transactions_tte_seconds",
```
pattyshack, line 37: s/transactions_tte_seconds/transactions_executed_in_seconds

integration/utils/follower.go
```diff
@@ -40,8 +41,13 @@ func WithInteval(interval time.Duration) followerOption {
 	return func(f *txFollowerImpl) { f.interval = interval }
 }
 
+func WithMetrics(m *metrics.LoaderCollector) followerOption {
+	return func(f *txFollowerImpl) { f.metrics = m }
+}
+
 type txFollowerImpl struct {
```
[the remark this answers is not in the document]
SaveTheRbtz: The main reason for using mutex for height and block is because they are updated as a pair.  But I get your point, let me rewrite it in a slower but robust way -- we do not need that code to be performant but we need it to be robust and easy to understand.

integration/utils/contLoadGenerator.go
```diff
@@ -40,15 +40,13 @@ type ContLoadGenerator struct {
...
-	trackTxs             bool
 	flowClient           *client.Client
 	serviceAccount       *flowAccount
 	flowTokenAddress     *flowsdk.Address
 	fungibleTokenAddress *flowsdk.Address
 	favContractAddress   *flowsdk.Address
 	accounts             []*flowAccount
-	availableAccounts    chan *flowAccount // queue with accounts available for   workers
-	txTracker            *TxTracker
+	availableAccounts    chan *flowAccount // queue with accounts available for workers
```
pattyshack, line 49: can you remove the unused fields?  e.g., initialized & feedbackEnabled are populated, but never used.

SaveTheRbtz: nice catch!  what static analyzer did you use to find these?

pattyshack: cmd+f on chrome =P

integration/utils/contLoadGenerator.go
```diff
@@ -556,40 +488,27 @@ func (lg *ContLoadGenerator) sendFavContractTx(workerID int) {
 		return
 	}
 
-	lg.sendTx(workerID, tx)
+	ch, err := lg.sendTx(workerID, tx)
+	if err != nil {
+		return
```
pattyshack, line 493: log error?

integration/utils/contLoadGenerator.go
```diff
@@ -304,76 +263,47 @@ func (lg *ContLoadGenerator) createAccounts(num int) error {
...
+				panic(err)
+			}
+
+			newAcc := newFlowAccount(i, &accountAddress, accountKey, signer)
+			i++
+
+			lg.accounts = append(lg.accounts, newAcc)
+			lg.availableAccounts <- newAcc
+
+			log.Trace().Hex("address", accountAddress.Bytes()).Msg("new account added")
```
pattyshack, line 304: return nil here (can the result contain multiple eventAccountCreated events)?

integration/utils/contLoadGenerator.go
```diff
@@ -304,76 +263,47 @@ func (lg *ContLoadGenerator) createAccounts(num int) error {
...
+		}
+	}
 	return nil
```
pattyshack, line 307: this should return error if no account created event was found.

integration/utils/contLoadGenerator.go
```diff
@@ -556,40 +492,27 @@ func (lg *ContLoadGenerator) sendFavContractTx(workerID int) {
...
+	<-ch
 }
 
-func (lg *ContLoadGenerator) sendTx(workerID int, tx *flowsdk.Transaction) {
+func (lg *ContLoadGenerator) sendTx(workerID int, tx *flowsdk.Transaction) (<-chan struct{}, error) {
 	log := lg.log.With().Int("workerID", workerID).Str("tx_id", tx.ID().String()).Logger()
 	log.Trace().Msg("sending transaction")
 
+	// Add watcher before sending the transaction to avoid race condition
+	ch := lg.follower.CompleteChanByID(tx.ID())
```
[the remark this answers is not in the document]
SaveTheRbtz: Maybe in the next diff when I improve the goroutine safety of the follwoer?

integration/utils/contLoadGenerator.go
SaveTheRbtz: sendTx already logs it

integration/utils/contLoadGenerator.go
SaveTheRbtz: yep, i think it will have exactly `i` accounts so returning here will break stuff.

pattyshack: in that case, we should verify num accounts actually created = num asked?

pattyshack (approved):

Kay-Zee (approved):

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors retry

PR #2641, opened 2022-06-16
Add a separate channel name for push-blocks on the public network

peterargue:
Access Nodes relay some messages from the staked libp2p network to the public network to support observers following consensus. The public network should use different channel names to avoid confusion and the potential for collisions if there were regressions/bugs. There is already a `public-sync-committee` channel. This PR adds a `public-push-blocks` channel as well.

### DIFF
--- a/engine/common/follower/engine.go
+++ b/engine/common/follower/engine.go
@@ -40,7 +40,7 @@ type Engine struct {
 	con            network.Conduit
 	sync           module.BlockRequester
 	tracer         module.Tracer
-	channel        network.Channel
+	channel        channels.Channel
 }
 
 type Option func(*Engine)
@@ -55,7 +55,7 @@ func WithComplianceOptions(opts ...compliance.Opt) Option {
 }
 
 // WithChannel sets the channel the follower engine will use to receive blocks.
-func WithChannel(channel network.Channel) Option {
+func WithChannel(channel channels.Channel) Option {
 	return func(e *Engine) {
 		e.channel = channel
 	}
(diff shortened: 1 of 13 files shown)


engine/common/follower/engine.go
```diff
@@ -80,7 +80,12 @@ func New(
 		tracer:         tracer,
 	}
 
-	con, err := net.Register(network.ReceiveBlocks, e)
+	channel := network.ReceiveBlocks
+	if !me.StakedNode() {
+		channel = network.PublicPushBlocks
+	}
+
+	con, err := net.Register(channel, e)
```
smnzhu, line 88: Could we maybe instead just pass the channel into the engine as an argument? This way we can get rid of the `StakedNode()` function

module/local.go
```diff
@@ -17,6 +17,9 @@ type Local interface {
 	// Address returns the (listen) address of the local node.
 	Address() string
 
+	// StakedNode returns true if the local node is a staked node. (i.e. not an observer)
+	StakedNode() bool
```
smnzhu, line 21: We can remove this if we do the above.

engine/common/follower/engine.go
smnzhu: An easier way to do it is to define a function option, which we can use to change this channel from the default. This way we won't need to modify all node `main.go` files, and only need to change the Observer file

smnzhu (approved):

kc1116 (approved):

peterargue:
bors merge

PR #2644, opened 2022-06-17
Don't leak signerLock outside of Account encapsulation

pattyshack:


SaveTheRbtz (approved):

integration/utils/accounts.go
```diff
@@ -21,6 +21,23 @@ type flowAccount struct {
 	signerLock sync.Mutex
 }
 
+func (acc *flowAccount) signCreateAccountTx(createAccountTx *flowsdk.Transaction) error {
```
Kay-Zee, line 24: not that i'm looking at this, why do we need one specifically for create account? can't we just use `signTx` and pass in `acc.accountKey.Index` as the keyID?

pattyshack: this increments a different sequence number than regular transaction (acc.accountKey.SequenceNumber vs acc.seqNumber).  I don't know if using two different seq num is intentional or not.  If they should be the same, I can can the behavior.

AndrewM-SDET (approved):

pattyshack:
bors merge

PR #2649, opened 2022-06-17
chore(localnet): force uid within the container to be root

SaveTheRbtz:
Previously, jaeger-all-in-one container was running implicitly under root (its Dockerfile does not have `USER` directive).  Make new stack also run under `root` so there are no permission problems under Linux.

Please note that this should not affect security in any way, since:
* this is the username _inside_ the container.
* these are `localnet`-only containers.
* no changes under macOS or on Linux where docker does not run under root user itself.

### DIFF
--- a/integration/localnet/docker-compose.metrics.yml
+++ b/integration/localnet/docker-compose.metrics.yml
@@ -2,6 +2,7 @@ version: '3.7'
 services:
   tempo:
     image: grafana/tempo:main-4d7e191
+    user: root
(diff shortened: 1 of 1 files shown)


zhangchiqing (approved):

simonhf (approved):
I tested this out on Linux and it works! :-)

Side-note: AFAIK the tempo `user: root` is not necessary but doesn't do any harm :-)

SaveTheRbtz:
bors merge

PR #2654, opened 2022-06-19
[Exec] improve tx processor tracing

SaveTheRbtz:
All other transaction processor invocations are already traced so lets trace this one too.


pattyshack (approved):

ramtinms:
We actually had this in place in the past but it was always very negligible in terms of time it took, so we removed it. putting it back is fine but I'm merging this processor into the verifier here at #2502, we could add extra sub spans for that processor.

PR #2658, opened 2022-06-20
[FVM] Move service account exceptions after payer signature has been verified

janezpodhostnik:
The call to `st.SetPayerIsServiceAccount()` enables some metering exceptions for the service account. The call to this should not happen before we verify that the payer is actually the service account. This verification happens in the `TransactionSignatureVerifier` processor.


tarakby (approved):

fvm/context.go
```diff
@@ -93,6 +93,7 @@ func defaultContext(logger zerolog.Logger) Context {
 		ExtensiveTracing:              false,
 		TransactionProcessors: []TransactionProcessor{
 			NewTransactionVerifier(AccountKeyWeightThreshold),
+			NewTransactionServiceAccountChecker(),
```
ramtinms, line 96: I think this should be part of the TransactionVerification, ideally we should avoid adding process for small operations, these simply adds a lot of unnecessary allocations.

janezpodhostnik: done [84ce8fb](https://github.com/onflow/flow-go/pull/2658/commits/84ce8fb676cb619a61af109098cbf789b5c5418f)

Kay-Zee: hmm, in terms of code modularity, i actually liked the previous setup. I guess extra processors are currently not cheap to add?

pattyshack:
This no longer make sense.  PayerIsServiceAccount is no longer a thing.  Also, janez' latest RunWithAllLimitsDisable will take care of these issues.

PR #2659, opened 2022-06-20
Size args slice correctly

pattyshack:


engine/execution/computation/manager.go
```diff
@@ -178,7 +178,7 @@ func (e *Manager) ExecuteScript(
 
 			prepareLog := func() *zerolog.Event {
 
-				args := make([]string, 0)
+				args := make([]string, 0, len(arguments))
 				for _, a := range arguments {
 					args = append(args, hex.EncodeToString(a))
```
SaveTheRbtz, line 183: does it matter from the performance perspective whether it is append or simply args[i] = ?

SaveTheRbtz: Also, what does the `args[:]` below do?  Is it needed?

SaveTheRbtz (approved):

engine/execution/computation/manager.go
pattyshack: if the slice is sized correctly, you probably won't notice a difference between the two

ramtinms (approved):

pattyshack:
bors merge

PR #2660, opened 2022-06-20
Simplify blockCommitter/eventHasher

pattyshack:
The callbacks make the code less readable

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -135,8 +135,10 @@ func (e *blockComputer) executeBlock(
 		Events:             make([]flow.EventsList, chunksSize),
 		ServiceEvents:      make(flow.EventsList, 0),
 		TransactionResults: make([]flow.TransactionResult, 0),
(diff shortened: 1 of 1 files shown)


engine/execution/computation/computer/computer.go
```diff
@@ -221,9 +207,6 @@ func (e *blockComputer) executeBlock(
 	close(eh.data)
```
SaveTheRbtz, line 207: nit: lets move these after the last commit/hash so that by the time Wait() is called goroutines are already gone.

pattyshack: this right now leaks gorountines on error.  i'll fix it in a different pull request.

pattyshack: #2661

SaveTheRbtz (approved):

ramtinms (approved):
Looks good to me

pattyshack:
bors merge

pattyshack:
bors merge

PR #2661, opened 2022-06-20
Don't leak blockCommitter/eventHasher gorountines on error

pattyshack:


engine/execution/computation/computer/computer.go
```diff
@@ -463,6 +468,14 @@ func (bc *blockCommitter) Commit(view state.View) {
 	bc.views <- view
 }
 
+func (bc *blockCommitter) Close() {
+	defer func() {
+		_ = recover() // allow double closing the channel
+	}()
+
+	close(bc.views)
```
SaveTheRbtz, line 476: maybe sync.Once instead, otherwise we might catch more than expected?

pattyshack: there's only a single line of code this could recover from ... but if it make you sleep better at night, sure.

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #2670, opened 2022-06-21
[Networking] Fixes flakey DNS cache expiry test

yhassanzadeh13:
This PR fixes the flakey test: `TestResolver_CacheExpiry`
The test was flakey basically because of two reasons: 
1. There was a race condition between [reading from the DNS cache and resolving an expired DNS entry](https://github.com/onflow/flow-go/blob/master/network/p2p/dns/resolver.go#L170-L177): Thread `a` would read the cache, find that a DNS record is expired, and would lock it for resolving. Thread `b` would also read the cache, find the same DNS record is expired and would check the lock for resolving it. However, by the time `b` would check for the lock, thread `a` would be done for updating the expired cache and hence would release the lock on that record, hence false positively setting up the requirements for thread `b` to obtain the lock for the same record and attempt on updating it. Since `TestResolver_CacheExpiry` would track the number of resolution attempts on a DNS record, this race condition would sometimes yield an extra attempt, which was failing the test. To fix it, this PR refactors the DNS cache maintenance by incorporating lock status of a DNS record to the cache itself. So the entire status of a DNS cache is retried by a [single query](https://github.com/onflow/flow-go/blob/yahya/6277-fix-flakey-dns-cache-test/network/p2p/dns/resolver.go#L177-L183), i.e., whether the cache exists or not? whether the cached record fresh or expired? and for an expired cache record, is there any ongoing resolution attempt?. By sending all these results through the same query both threads `a` and `b` in the earlier example will be notified whether there is an ongoing attempt, and [in case there is an ongoing attempt on resolving an expired cache, the other thread returns with the expired result (till update completes)](https://github.com/onflow/flow-go/blob/yahya/6277-fix-flakey-dns-cache-test/network/p2p/dns/resolver.go#L192). 

2. Using `HeroCache` as the underlying cache, the DNS records are identified with their domain. So as long as the domain was the same, the update would not take effect. We solved this issue by [implementing an atomic update method that removes the old entry first and writes the new one all in a single transaction to the `HeroCache`](https://github.com/onflow/flow-go/blob/yahya/6277-fix-flakey-dns-cache-test/module/mempool/herocache/dns_cache.go#L188-L210)). 

This PR also optimizes lock contentions by merging critical operations and adds extra logging to debug such cases easier later on. 

How to ensure `TestResolver_CacheExpiry` is no longer flakey? 

The following command runs the test 100 consecutive times till either all runs complete successfully or the test fails. And as it shows, the test is passing all 100 consecutive runs. 
```
go test -run TestResolver_CacheExpiry github.com/onflow/flow-go/network/p2p/dns/ -count 100 -failfast --tags=relic --vv
ok      github.com/onflow/flow-go/network/p2p/dns       401.391s
```


network/p2p/dns/resolver_test.go
```diff
@@ -76,33 +75,35 @@ func TestResolver_CacheExpiry(t *testing.T) {
...
 
 	cancelCtx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	ctx, _ := irrecoverable.WithSignaler(cancelCtx)
 	resolver.Start(ctx)
 	unittest.RequireCloseBefore(t, resolver.Ready(), 100*time.Millisecond, "could not start dns resolver on time")
 
-	size := 10 // we have 10 txt and 10 ip lookup test cases
-	times := 5 // each domain is queried for resolution 5 times
+	size := 5  // we have 5 txt and 5 ip lookup test cases
```
gomisha, line 86: why were these values reduced?

network/p2p/dns/resolver_test.go
```diff
@@ -76,33 +75,35 @@ func TestResolver_CacheExpiry(t *testing.T) {
...
+	times := 3 // each domain is queried for resolution 3 times
 	txtTestCases := testnetwork.TxtLookupFixture(size)
 	ipTestCase := testnetwork.IpLookupFixture(size)
 
 	// each domain gets resolved through underlying resolver twice: once initially, and once after expiry.
 	resolverWG := mockBasicResolverForDomains(t, &basicResolver, ipTestCase, txtTestCases, happyPath, 2)
 
+	// queries (5 + 5) cases * 3 = 100 queries.
```
gomisha, line 94: is this math correct? (5+5) * 3 = 30 queries

network/p2p/dns/resolver_test.go
```diff
@@ -76,33 +75,35 @@ func TestResolver_CacheExpiry(t *testing.T) {
...
 	queryWG := syncThenAsyncQuery(t, times, resolver, txtTestCases, ipTestCase, happyPath)
 	unittest.RequireReturnsBefore(t, queryWG.Wait, 1*time.Second, "could not perform all queries on time")
 
-	time.Sleep(2 * time.Second) // waits enough for cache to get invalidated
+	time.Sleep(4 * time.Second) // waits enough for cache to get invalidated
```
gomisha, line 98: was the test failing with the lower number? that's a long time to sleep

gomisha (approved):
LGTM, as per live walkthrough.
Left some minor comments and questions.

network/p2p/dns/resolver_test.go
LLsq2: 2nd this comment, if we can get away with a smaller timeout can we use a shorter time?

LLsq2 (approved):

network/p2p/dns/resolver_test.go
yhassanzadeh13: To reduce the test complexity for debugging and also reduce the run time. The goal is to test the concurrent queries, and 30 test cases ( (5+5) * 3) serve the same purpose as the original 100 test cases ((10 + 10) * 5), but with a smaller set of goroutines. Ultimately the desire is to even reduce it further to 5-10 concurrent goroutines for the test.

network/p2p/dns/resolver_test.go
yhassanzadeh13: Thanks, fixed it here: https://github.com/onflow/flow-go/pull/2670/commits/ef0832baecc78883ec4000be1f746dfa5b25b9cc

network/p2p/dns/resolver_test.go
yhassanzadeh13: Confirming that the test is passing with reduced timeouts: https://github.com/onflow/flow-go/pull/2670/commits/04ac8c71637161270be21fa8ad86f3e30fc33f09

PR #2673, opened 2022-06-21
optimize emergency-sealing check

zhangchiqing:
This PR optimize the emergency-sealing check:

1. when checking results for emergency sealing, we don't have an upper limit. In the event of sealing halt, the number of unsealed blocks will increase, and causing the CheckEmergencySealing to take longer and longer. This PR sets an upper limit to the number of unsealed blocks for checking emergency sealing.
2. The emergency sealing was only triggered when there are 100 blocks finalized _after_ the results have been incorporated. In order to make the emergency sealing more useful, this condition is breaking down into two parts:75 unsealed and finalized blocks, and 25 blocks to be finalized after the matching results is incorporated.


engine/consensus/sealing/core.go
```diff
@@ -448,30 +448,44 @@ func (c *Core) processApproval(approval *flow.ResultApproval) error {
...
 	}
-	return nil
+
+	// We iterate through the collectorTree to check if any collector is able to trigger emergency sealing.
+	// We can do some optimization instead of bindly iterate through all collectors.
+	// The optimization is to start from the collector for the first unsealed block, if it can trigger emergency sealing, we
+	// check for the next unsealed and finalized block, and stop as soon as we met with a height that cannot trigger emergency sealing.
+	// The reason this optimization could work is because the following facts:
+	// 1) emergency sealing is only for finalized blocks,
+	// 2) for finalized blocks, there is only one collector at each height
```
AlexHentschel, line 476: I am not sure why this should hold. Imagine we have an execution fork. Then there would be two different results (in the finalized blocks) both awaiting approvals from verifiers.

zhangchiqing: sorry, yeah, you are right. I realized later as well.

My optimization could work with execution fork on finalized blocks, just forgot to update the comments here. Will update

AlexHentschel: I think you would need to change the logic to 
```
if none of the results for this block height can be sealed, then we don't need to check descending results whether they can be emergency sealed. 
```

AlexHentschel (approved):
Very nice. 

My only conceptual comment is that `DefaultEmergencySealingThresholdForExecution` and `DefaultEmergencySealingThresholdForVerification` are _not_ added. 

Emergency_Sealing

We require that 
* Δh<sub>1</sub> > `DefaultEmergencySealingThresholdForExecution` ([code](https://github.com/onflow/flow-go/blob/0d4981d3598198ed2e016fc68f8b8e398283d132/engine/consensus/approvals/verifying_assignment_collector.go#L81))
* and Δh<sub>2</sub> > `DefaultEmergencySealingThresholdForVerification` ([code](https://github.com/onflow/flow-go/blob/0d4981d3598198ed2e016fc68f8b8e398283d132/engine/consensus/approvals/verifying_assignment_collector.go#L82))

For example, assume block `A` has height 100. Block `B` (incorporating `A`'s result) has height `150`. I think we would emergency seal `A`'s result at height `176`, maybe (?)

As a result, we would decrease the threshold for emergency sealing to kick in from 100 blocks (previous code version) to 75 blocks new code version. Just wanting to make sure that this reduction is intentional? 

P.S. I have put up a **PR https://github.com/onflow/flow-go/pull/2703 with lots of added documentation** targeting your branch. There are no functional changes, aside from an added sanity check and keeping the `DefaultEmergencySealingThresholdForExecution` at 100 (because we do _not_ add `DefaultEmergencySealingThresholdForVerification`).

AlexHentschel:
thanks for [your suggestion](https://github.com/onflow/flow-go/pull/2703#discussion_r907857233) to rename `DefaultEmergencySealingThresholdForExecution` to `DefaultEmergencySealingThresholdForFinalization`. Agree, that will make the code better to read.

engine/consensus/sealing/core.go
```diff
@@ -423,25 +423,49 @@ func (c *Core) processApproval(approval *flow.ResultApproval) error {
...
 }
 
-func (c *Core) checkEmergencySealing(observer consensus.SealingObservation, lastSealedHeight, lastFinalizedHeight uint64) error {
+// checkEmergencySealing triggers the AssignmentCollectors to check whether satisfy the conditions to
+// generate an emergency seal. To limit performance impact of these checks, we limit emergency sealing
+// to the 100 lowest finalized blocks that are still unsealed.
+// Inputs:
+//  * `observer` for tracking and reporting the current internal state of the local sealing logic
+//  * `lastFinalizedHeight` is the height of the latest block that is finalized
+//  * `lastHeightWithFinazliedSeal` is the height of the latest block that is finalized and in addition
```
jordanschalm, line 432:
```suggestion
//  * `lastHeightWithFinalizedSeal ` is the height of the latest block that is finalized and in addition
```
```suggestion
//  * `lastHeightWithFinazliedSeal` is the height of the latest block that is finalized and in addition
```

jordanschalm (approved):
Looks good, only comments are related to documentation and some typos.

zhangchiqing:
bors merge

PR #2678, opened 2022-06-22
[FVM] Don't read context from state on system chunk

janezpodhostnik:
Disable reading (mutating) the context from the state for system transaction and explicitly set a very high memory limit.

### DIFF
--- a/engine/execution/computation/computer/computer_test.go
+++ b/engine/execution/computation/computer/computer_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/onflow/flow-go/engine/execution/testutil"
 	"github.com/onflow/flow-go/fvm"
 	fvmErrors "github.com/onflow/flow-go/fvm/errors"
+	"github.com/onflow/flow-go/fvm/meter/weighted"
 	"github.com/onflow/flow-go/fvm/programs"
 	"github.com/onflow/flow-go/fvm/state"
 	"github.com/onflow/flow-go/fvm/systemcontracts"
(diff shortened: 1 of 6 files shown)


m4ksio (approved):

janezpodhostnik:
I tested manually to get the patch out quickly, but you are right, I will add a test.

fvm/context.go
```diff
@@ -13,10 +13,12 @@ import (
...
-	Chain                        flow.Chain
-	Blocks                       Blocks
-	Metrics                      handler.MetricsReporter
-	Tracer                       module.Tracer
+	Chain   flow.Chain
+	Blocks  Blocks
+	Metrics handler.MetricsReporter
+	Tracer  module.Tracer
+	// LoadContextFromState is a flag telling the fvm to load certain parts of the context from the state
+	LoadContextFromState         bool
```
turbolent, line 21: I don't know of a better name, but is "loading the context" the right term? 
Maybe it can be more specific to what it currently configures, i.e. if the the execution parameters are set?

ramtinms: Maybe we can name it `AllowContextOverrideByExecutionState`

fvm/scriptEnv.go
```diff
@@ -101,7 +101,9 @@ func NewScriptEnvironment(
 		env.seedRNG(fvmContext.BlockHeader)
 	}
 
-	env.setExecutionParameters()
+	if fvmContext.LoadContextFromState {
+		env.setExecutionParameters()
```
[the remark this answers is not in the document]
janezpodhostnik: the scrips have a different `setExecutionParameters` that doesn't return an error

janezpodhostnik:
@m4ksio added a test [c72999c](https://github.com/onflow/flow-go/pull/2678/commits/c72999c4914fe57335cb2a0587a8c409d85002c5). It passes, but the CI is having troubles with the private repos.

fvm/context.go
janezpodhostnik: Renamed [43aa0ae](https://github.com/onflow/flow-go/pull/2678/commits/43aa0aeb5a4b56a5ca191c4314edd75124e89650)

engine/execution/computation/computer/computer_test.go
```diff
@@ -362,6 +363,48 @@ func TestBlockExecutor_ExecuteBlock(t *testing.T) {
...
 
+	t.Run("system transaction does not read memory limit from state", func(t *testing.T) {
+
+		weighted.DefaultMemoryWeights = weighted.ExecutionMemoryWeights{
+			0: 1, // single weight set to 1
+		}
+		execCtx := fvm.NewContext(
+			zerolog.Nop(),
+			fvm.WithMemoryLimit(10), // the context memory limit is set to 10
+		)
```
ramtinms, line 374: this is a nice test, but we might add a test for the AllowContextOverrideByExecutionState functionality to the FVM as well.

PR #2684, opened 2022-06-23
[FVM] Code Cleanup

janezpodhostnik:
Some code cleanup 🧹 :

- [f440afa](https://github.com/onflow/flow-go/pull/2684/commits/f440afa1afa020b21d8ffe3e3cc253b3086a760c) errors cleanup: 
    - remove errors no longer in use
    - fix some typos
    - fix some inconsistent receivers
- [7f7f5a2](https://github.com/onflow/flow-go/pull/2684/commits/7f7f5a228a0a6c7ab9b02a6d2986af2dcaec972b) blueprints cleanup: 
    - extract cadence scripts into separate cdc files (this way we can late write tests for the cadence scripts or run some CI automation over them)
    - change some constants from a (domain string, path string) pair to a cadence path, so there is less code
- [51f789c](https://github.com/onflow/flow-go/pull/2684/commits/51f789cd3a1802eb73e76c868370ba16b24df34b) merge storage limiter with transaction invoker:
    - the limiter had one function and was only called from the invoker. Storage limiting is also contextually similar to fee deduction (an extra step to do after the transaction invocation) so it fits there.
- [0295c24](https://github.com/onflow/flow-go/pull/2684/commits/0295c2436ac7d95a18b715ab5deed568d0b8e8b4):
    - change cadence template to use arguments and the template replace function from core-contracts


fvm/blueprints/contracts.go
```diff
@@ -1,50 +1,48 @@
...
+	Identifier: "authorizedAddressesToRemoveContracts",
+}
+var IsContractDeploymentRestrictedPath = cadence.Path{
+	Domain:     common.PathDomainStorage.Identifier(),
+	Identifier: "isContractDeploymentRestricted",
+}
 
-const ContractRemovalAuthorizedAddressesPathDomain = "storage"
-const ContractRemovalAuthorizedAddressesPathIdentifier = "authorizedAddressesToRemoveContracts"
+//go:embed scripts/setContractOperationAuthorizersTransactionTemplate.cdc
```
Tonix517, line 29: thank you for doing this - it also has a potential benefit that when we have a good Cadence linter in place later, all these .cdc files can be easily linted

Tonix517 (approved):
LGTM

fvm/errors/codes.go
```diff
@@ -75,6 +77,23 @@ const (
...
+	ErrCodeInvalidScriptError ErrorCode = 1004
+	// Deprecated: ErrCodeInvalidGasLimitError
+	ErrCodeInvalidGasLimitError ErrorCode = 1005
+	// Deprecated: ErrCodeExpiredTransactionError
+	ErrCodeExpiredTransactionError ErrorCode = 1003
+	// Deprecated: ErrCodeEncodingUnsupportedValue
+	ErrCodeEncodingUnsupportedValue ErrorCode = 1102
+	// Deprecated: ErrCodeGasLimitExceededError
+	ErrCodeGasLimitExceededError ErrorCode = 1104
+	// Deprecated: ErrCodeContractNamesNotFoundError
```
ramtinms, line 97: keeping them deprecated in there might help not to allocate the same number, this way you might reuse and not see it.

fvm/blueprints/fees.go
```diff
@@ -1,29 +1,48 @@
...
+	Domain:     common.PathDomainStorage.Identifier(),
+	Identifier: "executionMemoryWeights",
 }
-`
+var TransactionFeesExecutionMemoryLimitPath = cadence.Path{
+	Domain:     common.PathDomainStorage.Identifier(),
+	Identifier: "executionMemoryLimit",
+}
+
+//go:embed scripts/deployTxFeesTransactionTemplate.cdc
```
ramtinms, line 29: what is the main benefit of using go:embed here, I found it making readability a bit worse now that you have to look up two places. 

I'm also worried we are using cdc files that are not a working version of cdc and are templates that requires replacement, maybe we could define a new file extension for cadence templates ?

fvm/errors/contracts.go
```diff
@@ -33,28 +33,3 @@ func (e *ContractNotFoundError) Error() string {
 func (e *ContractNotFoundError) Code() ErrorCode {
 	return ErrCodeContractNotFoundError
 }
-
-// ContractNamesNotFoundError is returned when fetching a list of contract names under an account
-type ContractNamesNotFoundError struct {
-	address flow.Address
-}
-
```
ramtinms, line 41: hmm, shall we return this when its happening? what we do right now in that case?

ramtinms (approved):
looks good to me, just added some questions.

fvm/errors/codes.go
ramtinms: thanks

janezpodhostnik:
This got stale. Splitting it up

PR #2691, opened 2022-06-23
Invalidate Pooling Logic in AN

lolpuddle:
- This change removes the use of the grpc closers from AN's connection factory

- This change allows connections to EN to be reused properly without closing the connection at the end of every call

- This change would also fix a race condition from when a request is made, and the deferred closer logic fails to close before another request is then made. For example if `ExecuteScriptAtBlockID` was called an Exec RPC Client would be cached. Before the first `ExecuteScriptAtBlockID` finishes and the deferred closer logic gets to execute, if a second ExecuteScriptAtBlockID was called, the original cached Exec RPC Client would then be reused, and before the second call gets a chance to use the client, the first call could execute the closer logic.

### DIFF
--- a/engine/access/rpc/backend/connection_factory.go
+++ b/engine/access/rpc/backend/connection_factory.go
@@ -3,7 +3,6 @@ package backend
 import (
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"time"
 
@@ -22,8 +21,10 @@ const defaultClientTimeout = 3 * time.Second
 
 // ConnectionFactory is used to create an access api client
 type ConnectionFactory interface {
-	GetAccessAPIClient(address string) (access.AccessAPIClient, io.Closer, error)
-	GetExecutionAPIClient(address string) (execution.ExecutionAPIClient, io.Closer, error)
+	GetAccessAPIClient(address string) (access.AccessAPIClient, error)
+	InvalidateAccessAPIClient(address string) bool
+	GetExecutionAPIClient(address string) (execution.ExecutionAPIClient, error)
+	InvalidateExecutionAPIClient(address string) bool
 }
 
 type ProxyConnectionFactory struct {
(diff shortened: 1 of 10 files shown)


vishalchangrani:
@lolpuddle - it will be good to have some metrics in the future around the connection pooling to see how many open connections does an AN has with the EN, how they grow and shrink over a period time etc.

engine/access/rpc/backend/connection_factory.go
```diff
@@ -93,38 +94,52 @@ func (cf *ConnectionFactoryImpl) retrieveConnection(grpcAddress string, timeout
...
+	return accessAPIClient, nil
 }
 
-func (cf *ConnectionFactoryImpl) GetExecutionAPIClient(address string) (execution.ExecutionAPIClient, io.Closer, error) {
+func (cf *ConnectionFactoryImpl) InvalidateAccessAPIClient(address string) bool {
+	grpcAddress, err := getGRPCAddress(address, cf.CollectionGRPCPort)
+	if err != nil {
+		return true
+	}
+	return cf.ConnectionsCache.Remove(grpcAddress)
```
Kay-Zee, line 118: we don't have to close the connection before removing it?

Kay-Zee (commented):
While i agree it makes sense to not close the connect after every call, i am somewhat surprised we've removed the closer entirely. But i guess a closer is mostly un-needed externally, as that's something the connection pool should take care of internally

engine/access/rpc/backend/connection_factory.go
lolpuddle: The logic from here should cover the closing of connections.
https://github.com/onflow/flow-go/blob/c9d7fddceb0ba58b8d537c974d58ace69eb67328/engine/access/rpc/engine.go#L143

This test (https://github.com/onflow/flow-go/blob/c9d7fddceb0ba58b8d537c974d58ace69eb67328/engine/access/rpc/backend/connection_factory_test.go#L387) also checks if our key removal logic would trigger the eviction policy code to close the connection

koko1123 (approved):

lolpuddle:
bors merge

PR #2692, opened 2022-06-23
Add latest height metrics for execution receipts and collections

danielholmes839:


koko1123 (approved):
This makes sense! will defer to @smnzhu and @peterargue if we need to lookup `maxheight` from local storage on initialization

engine/execution/ingestion/engine_test.go
```diff
@@ -768,6 +769,9 @@ func TestBlocksArentExecutedMultipleTimes_collectionArrival(t *testing.T) {
 		wgC := sync.WaitGroup{}
 		wgC.Add(1)
 
+		// stops .ByID() call from blocking in handleCollection
+		ctx.blocks.EXPECT().ByID(gomock.Any()).AnyTimes().Return(nil, errors.New("mock error"))
```
peterargue, line 773: Is this still needed?

engine/access/ingestion/engine.go
```diff
@@ -424,6 +426,15 @@ func (e *Engine) handleExecutionReceipt(originID flow.Identifier, r *flow.Execut
 		return fmt.Errorf("failed to store execution receipt: %w", err)
 	}
 
+	block, err := e.blocks.ByID(r.ExecutionResult.BlockID)
+	if err != nil {
+		return fmt.Errorf("failed to lookup block id while handling receipt: %w", err)
```
peterargue, line 431:
```suggestion
		return fmt.Errorf("failed to lookup block while handling receipt: %w", err)
```

engine/access/ingestion/engine.go
```diff
@@ -424,6 +426,15 @@ func (e *Engine) handleExecutionReceipt(originID flow.Identifier, r *flow.Execut
...
+	}
+
+	if block.Header.Height > e.maxReceiptHeight {
+		e.transactionMetrics.UpdateExecutionReceiptMaxHeight(block.Header.Height)
```
peterargue, line 435: I missed this before, but I think you want to also update `maxReceiptHeight` here with the new highest height. Same in the execution engine

danielholmes839: ahhhh I can't believe I forgot this.

peterargue (approved):

PR #2693, opened 2022-06-23
Fix copylocks in TestOnBlockProposal_RemoteOrigin

fxamacker:
`Suite` shouldn't be passed by value because it contains `sync.RWMutex`.

Use `*Suite` instead of `Suite` in `TestOnBlockProposal_RemoteOrigin` to pass `sync.RWMutex` lock by reference.


zhangchiqing (approved):

turbolent:
Thank you for fixing this @fxamacker! `make lint` reports this on `master`, how did this end up there?

turbolent (approved):

PR #2695, opened 2022-06-24
[FVM] Change restricted contract deployment env setting on testnet

janezpodhostnik:
This is a temporary change.

See comments in the code, and the doc here: https://www.notion.so/dapperlabs/Enable-Permissionless-Contract-Deployment-On-mainnet-3a9ff010423444c0975ed6052e028968

The state commitment change is due to the flag now being set in the state at bootstrapping.
The test change is due to the flag being false by default.


m4ksio (approved):

janezpodhostnik:
no longer needed, due to permissionless contract deployment now enabled everywhere.

PR #2697, opened 2022-06-24
Extend consensus follower config to support compliance settings

jordanschalm:
The consensus follower has its own `Config` struct which limits the scope of what can be configured by passing in configuration arguments to the constructor. This PR adds compliance config to the Follower's Config, so it is configurable when instantiating a Follower instance.


peterargue (approved):

jordanschalm:
bors merge

PR #2699, opened 2022-06-24
[Flakey Test] Handle registering node without machine account

jordanschalm:
Previously the epoch tests would pass in an argument representing a machine account key, even for nodes which had no machine account. This worked because the SDK accepted an empty account key and the argument was ignored.

An SDK API change caused this to no longer work, because the SDK will not accept empty account keys.

Instead, we handle the case of an empty machine account explicitly by passing in a nil optional for this parameter. This should address the 0% success rate for AN/VN tests, which do not use machine accounts.

### DIFF
--- a/integration/tests/epochs/suite.go
+++ b/integration/tests/epochs/suite.go
@@ -8,11 +8,12 @@ import (
 
 	"github.com/onflow/cadence"
 	"github.com/onflow/flow-core-contracts/lib/go/templates"
-	sdk "github.com/onflow/flow-go-sdk"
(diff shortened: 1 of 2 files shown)


gomisha:
Can you add the names of all tests that should be fixed by this in the 
a) PR title and 
b) PR description? 

This would help when we quickly scan the list of open PRs in our master Epic tracking all flaky test work: [[Test][Epic] Fix flaky tests #6255](https://github.com/dapperlabs/flow-go/issues/6255)

gomisha (approved):

peterargue (approved):

jordanschalm:
bors merge

PR #2701, opened 2022-06-24
Refactor Access API into Protocol API

szegedim:
Protocol API will allow Flow Api Service to get block and block header bypassing collection node and execution node.


peterargue:
these changes were pulled into https://github.com/onflow/flow-go/pull/2962

PR #2704, opened 2022-06-27
Remove header caching

SaveTheRbtz:
This 1 item caching introduces global lock.  Also code is quite brittle and can cause bugs if `Header` struct changes.

This diff is no-op on Skylake:
```
name                              old us/tx    new us/tx    delta
ComputeBlock/1/cols/16/txes-16     1.44k ± 0%   1.46k ± 3%    ~     (p=0.099 n=9+10)
ComputeBlock/1/cols/32/txes-16     1.37k ± 1%   1.38k ± 1%  +0.60%  (p=0.007 n=9+10)
ComputeBlock/1/cols/64/txes-16     1.36k ± 1%   1.35k ± 0%  -0.51%  (p=0.014 n=10+8)
ComputeBlock/1/cols/128/txes-16    1.34k ± 1%   1.34k ± 0%    ~     (p=0.673 n=10+9)
ComputeBlock/4/cols/16/txes-16     1.36k ± 2%   1.36k ± 1%    ~     (p=0.237 n=10+10)
ComputeBlock/4/cols/32/txes-16     1.34k ± 1%   1.34k ± 1%    ~     (p=0.329 n=10+10)
ComputeBlock/4/cols/64/txes-16     1.35k ± 1%   1.34k ± 2%    ~     (p=0.117 n=9+10)
ComputeBlock/4/cols/128/txes-16    1.36k ± 2%   1.35k ± 2%    ~     (p=0.287 n=9+10)
ComputeBlock/16/cols/16/txes-16    1.33k ± 1%   1.32k ± 1%    ~     (p=0.643 n=10+10)
ComputeBlock/16/cols/32/txes-16    1.32k ± 1%   1.32k ± 1%    ~     (p=0.202 n=9+10)
ComputeBlock/16/cols/64/txes-16    1.35k ± 1%   1.35k ± 1%    ~     (p=0.957 n=10+10)
ComputeBlock/16/cols/128/txes-16   1.32k ± 1%   1.32k ± 2%    ~     (p=0.643 n=10+10)
```

And a 3% improvement on M1 Pro:
```
$ go test -tags relic -bench='BenchmarkComputeBlock' -count 10 -run='^$' ./engine/execution/computation/.
...
$ benchstat old new
name                             old us/tx    new us/tx     delta
ComputeBlock/1/cols/16/txes-8       590 ± 3%      803 ±62%    ~     (p=0.404 n=10+10)
ComputeBlock/1/cols/32/txes-8       554 ± 2%      559 ± 2%  +0.93%  (p=0.043 n=9+8)
ComputeBlock/1/cols/64/txes-8       557 ± 4%      538 ± 2%  -3.46%  (p=0.000 n=10+10)
ComputeBlock/1/cols/128/txes-8      560 ± 1%      538 ± 2%  -3.92%  (p=0.000 n=7+9)
ComputeBlock/4/cols/16/txes-8       558 ± 3%      539 ± 1%  -3.51%  (p=0.000 n=10+10)
ComputeBlock/4/cols/32/txes-8       555 ± 1%      540 ± 1%  -2.81%  (p=0.000 n=10+10)
ComputeBlock/4/cols/64/txes-8       556 ± 2%      538 ± 2%  -3.13%  (p=0.000 n=10+10)
ComputeBlock/4/cols/128/txes-8      546 ± 1%      527 ± 2%  -3.50%  (p=0.000 n=10+10)
ComputeBlock/16/cols/16/txes-8      552 ± 3%      533 ± 2%  -3.44%  (p=0.000 n=10+10)
ComputeBlock/16/cols/32/txes-8      548 ± 1%      527 ± 2%  -3.68%  (p=0.000 n=9+10)
ComputeBlock/16/cols/64/txes-8      548 ± 1%      530 ± 1%  -3.34%  (p=0.000 n=10+10)
ComputeBlock/16/cols/128/txes-8     522 ± 1%      519 ± 2%    ~     (p=0.327 n=9+9)
```

cc: @zhangchiqing since he was also looking at it here https://github.com/onflow/flow-go/pull/1279#discussion_r712510891

### DIFF
--- a/model/flow/header.go
+++ b/model/flow/header.go
@@ -1,9 +1,7 @@
 package flow
 
 import (
-	"bytes"
 	"encoding/json"
-	"sync"
 	"time"
 
 	"github.com/fxamacker/cbor/v2"
@@ -72,54 +70,10 @@ func (h Header) Fingerprint() []byte {
 	return fingerprint.Fingerprint(h.Body())
 }
 
-var mutexHeader sync.Mutex
-var previdHeader Identifier
-var prevHeader Header
-
 // ID returns a unique ID to singularly identify the header and its block
 // within the flow system.
 func (h Header) ID() Identifier {
-	// NOTE: this is just a sanity check to make sure that we don't get
-	// different block IDs if someone forgets to use UTC timestamps
-	if h.Timestamp.Location() != time.UTC {
-		h.Timestamp = h.Timestamp.UTC()
-	}
-
-	mutexHeader.Lock()
-
-	// unlock at the return
-	defer mutexHeader.Unlock()
-
-	// compare these elements individually
-	if prevHeader.ParentVoterIndices != nil &&
-		prevHeader.ParentVoterSigData != nil &&
-		prevHeader.ProposerSigData != nil &&
-		len(h.ParentVoterIndices) == len(prevHeader.ParentVoterIndices) &&
-		len(h.ParentVoterSigData) == len(prevHeader.ParentVoterSigData) &&
-		len(h.ProposerSigData) == len(prevHeader.ProposerSigData) {
-
-		if h.ChainID == prevHeader.ChainID &&
-			h.Timestamp == prevHeader.Timestamp &&
-			h.Height == prevHeader.Height &&
-			h.ParentID == prevHeader.ParentID &&
-			h.View == prevHeader.View &&
-			h.PayloadHash == prevHeader.PayloadHash &&
-			bytes.Equal(h.ProposerSigData, prevHeader.ProposerSigData) &&
-			bytes.Equal(h.ParentVoterIndices, prevHeader.ParentVoterIndices) &&
-			bytes.Equal(h.ParentVoterSigData, prevHeader.ParentVoterSigData) &&
-			h.ProposerID == prevHeader.ProposerID {
-
-			// cache hit, return the previous identifier
-			return previdHeader
-		}
-	}
-
-	previdHeader = MakeID(h)
-
-	// store a reference to the Header entity data
-	prevHeader = h
-
-	return previdHeader
+	return MakeID(h)
 }
 
 // Checksum returns the checksum of the header.


bluesign:
Very nice idea (I was mostly against from the beginning), it would be nice to leave a comment to the next guy though.

model/flow/header.go
```diff
@@ -72,54 +70,10 @@ func (h Header) Fingerprint() []byte {
...
-		}
-	}
-
-	previdHeader = MakeID(h)
-
-	// store a reference to the Header entity data
-	prevHeader = h
-
-	return previdHeader
+	return MakeID(h)
```
pattyshack, line 76: how many ID() and Checksum() calls do we make on each Header object?  We can just precompute and store the hash in the struct?

pattyshack (approved):

zhangchiqing:
If golang had an immutable data structure that is guaranteed won't be updated that we can cache the ID. However, it doesn't have it.

So if we cache the ID in the struct, it's possible that the cache will become stale if the struct is later updated.

In fact, ID method is called a lot of times. 

The existing "optimization" is to say, let's cache the ID once we have computed it for a blocker header. If we are going to compute the ID for the same header again, then we check if all fields of the header are identical, if yes, then we return the cached ID. 

However, this approach only works if we are calling ID on the same header multiple times in a row. We evaluated that this is probably common. Since, in practice, when we receive a block, that block header will likely go through many layers (modules) and each layer will need to compute the ID of it. This means, we will compute the same header multiple times during the processing of the block, and the cache will hit.

So the trade off is that, we sacrify some concurrency by using a global lock, but we save time if the same header's ID is computed multiple times in a row. 

That being said, I'm not sure if the benchmark could reflect the real chance on mainnet about how often the same block header's ID method are being called.

SaveTheRbtz:
@zhangchiqing benchmark suggests that cost of taking and releasing lock + the cost of full byte-by-byte comparison (it is always the worst case for the identical headers since we need to compare all fields) is the same as the cost of reading all the fields and hashing them w/o taking/releasing the lock.

pattyshack:
you can work around the lack of immutable with some refactoring / classic oo pattern

e.g., you can make all fields private (lower case) and require all field updated via setters, which update the field and recompute the hash at the same time (you would need to maintain a second struct for marshalling / unmarshalling)

zhangchiqing:
we have some discussion around whether to cache the result of MakeID.

This was some benchmark did in the past.
https://www.notion.so/dapperlabs/MakeID-2823afa345544e4080964f1a86a08186

https://www.notion.so/dapperlabs/Top-CPU-users-most-call-makeID-1b9c5786afdc4833a3ce150b2f8b1fd7

https://www.notion.so/dapperlabs/Header-ID-analysis-366a90bb14b742d8b0b07ad33690a52b

I think @ramtinms and @AlexHentschel @simonhf might have more context.

SaveTheRbtz:
I went through the docs, I'm not sure there was any research about TPS before and after the change.

model/flow/header.go
SaveTheRbtz: computing the hash should have pretty much the same cost as comparing the fields, so the benefit unknown.  let me hack it with a sync.Once and run the benchmark to see if result is even measurable.

SaveTheRbtz: hacked in header caching. (even without locks) it does not seem to affect performance on `ComputeBlock` tests:
```
name                             old us/tx      new us/tx      delta
ComputeBlock/1/cols/16/txes-8         593 ± 1%       588 ± 2%  -0.73%  (p=0.007 n=9+8)
ComputeBlock/1/cols/32/txes-8         561 ± 3%       561 ± 6%    ~     (p=0.347 n=10+9)
ComputeBlock/1/cols/64/txes-8         537 ± 2%       542 ± 3%    ~     (p=0.163 n=9+10)
ComputeBlock/1/cols/128/txes-8        542 ± 2%       544 ± 2%    ~     (p=0.847 n=9+9)
ComputeBlock/4/cols/16/txes-8         538 ± 1%       542 ± 1%  +0.78%  (p=0.045 n=10+10)
ComputeBlock/4/cols/32/txes-8         546 ± 2%       545 ± 3%    ~     (p=0.703 n=9+10)
ComputeBlock/4/cols/64/txes-8         546 ± 4%       552 ± 4%    ~     (p=0.197 n=9+9)
ComputeBlock/4/cols/128/txes-8        536 ± 3%       538 ± 7%    ~     (p=0.794 n=10+9)
ComputeBlock/16/cols/16/txes-8        541 ± 4%       533 ± 1%    ~     (p=0.059 n=10+10)
ComputeBlock/16/cols/32/txes-8        532 ± 2%       531 ± 2%    ~     (p=0.734 n=10+9)
ComputeBlock/16/cols/64/txes-8        535 ± 1%       532 ± 2%    ~     (p=0.249 n=9+10)
ComputeBlock/16/cols/128/txes-8       527 ± 2%       520 ± 1%  -1.39%  (p=0.007 n=9+10)
```

SaveTheRbtz:
bors merge

AlexHentschel:
> We can just precompute and store the hash in the struct?

Yes, this would be the general direction of a useful optimization. Though, there are a few subtle but very important security considerations (very specific to blockchains) that we have to take into account. In case you are planning to implement some optimization later, please include me in the discussion during the design phase. Thanks.

PR #2705, opened 2022-06-27
[FVM] All service accounts should be bootstrapped with a key

janezpodhostnik:
All service accounts:
- service account
- flow token account
- fungible token account
- flow fees account
- default node accounts

Should have an associated key so that the contracts/resources on those accounts are available if anyone has the need to change them they can.

The key is the same for all accounts and is the service account key.

This only affects transient networks like benchnet, localnet, emulator,...


engine/execution/computation/execution_verification_test.go
```diff
@@ -160,7 +160,7 @@ func Test_ExecutionMatchesVerification(t *testing.T) {
 		err = testutil.SignTransaction(addKeyTx, accountAddress, accountPrivKey, 0)
 		require.NoError(t, err)
 
-		minimumStorage, err := cadence.NewUFix64("0.00008164")
+		minimumStorage, err := cadence.NewUFix64("0.00008312")
```
janezpodhostnik, line 163: The reason this needed a change is because `executeBlockAndVerify` also does bootstrapping and it is one of the bootstrapped accounts that dictates the minimum possible minimum account balance.

janezpodhostnik:
It is a full weight key.

The reason these are different accounts is that int the past there was other way to do this (one account one contract). We could change this by deploying all contracts to one account, but this would mean that new networks (which are usually transient networks and always for testing) would have a significantly different setup that mainnet, which might lead to some weird edge cases.

On the other hand you could also argue that the responsibilities of the different contracts should be separate...

I think the best solution is to add option for each key separately, and if not set, default to the service account key.

janezpodhostnik:
@tarakby Added option to set individual keys for all the accounts created during bootstrapping

tarakby:
Is this code used to bootstrap Mainnet, transient networks, or both?

- I like the idea of using a list of keys for each account, instead of a single full weight key. The service account on Mainnet for instance is controlled by multiple keys of partial weights. 
- It's also good that the bootstrapping software allows using different keys for each account. It's the "bootstrapper"'s responsibility to decide to re-use the same keys for all accounts (which I think is not a good practice, unless we deploy 5 contracts under the same account) or to use a different set for each account (only that would separate the responsibilities as intended).

fvm/bootstrap.go
```diff
@@ -55,6 +55,14 @@ type BootstrapProcedure struct {
 	identities flow.IdentityList
 }
 
+type BootstrapAccountKeys struct {
```
tarakby, line 58: This structure can be private right?

janezpodhostnik: I don't think so because it is used in the option
```
func WithBootstrapAccountKeys(keys BootstrapAccountKeys) BootstrapProcedureOption {
```

I could rewrite the option so you have to pass keys individually. Would that be preferred?

tarakby: No it is fine with the public struct. I just thought it's not used anywhere else :)

ramtinms (approved):
Looks good to me, but maybe adding this as a note to the PR description: 

this would only impact benchnet and localnets right? considering we never do the bootstrapping for older networks.

janezpodhostnik:
bors merge

PR #2719, opened 2022-06-28
Add detailed debug logging in executeBlock

Kay-Zee:


SaveTheRbtz (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -196,12 +204,25 @@ func (e *blockComputer) executeBlock(
 		if err != nil {
 			return nil, fmt.Errorf("failed to execute collection at txIndex %v: %w", txIndex, err)
 		}
+		collectionID := collection.Guarantee.CollectionID[:]
```
SaveTheRbtz, line 207: you may consider creating collectionLogger here.

engine/execution/computation/computer/computer.go
```diff
@@ -196,12 +204,25 @@ func (e *blockComputer) executeBlock(
...
+			Msg("collection events hashed")
 		err = stateView.MergeView(colView)
 		if err != nil {
 			return nil, fmt.Errorf("cannot merge view: %w", err)
 		}
+		blockLogger.Debug().
+			Hex("collection_id", collectionID).
+			Msg("collection view merged")
 		collectionIndex++
 	}
```
SaveTheRbtz, line 228: maybe also something after the execution of the system collection

SaveTheRbtz (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -196,12 +204,25 @@ func (e *blockComputer) executeBlock(
...
+			Hex("collection_id", collectionID).
+			Msg("collection executed")
 		bc.Commit(colView)
+		blockLogger.Debug().
+			Hex("collection_id", collectionID).
+			Msg("collection committed")
 		eh.Hash(res.Events[i])
+		blockLogger.Debug().
+			Hex("collection_id", collectionID).
+			Msg("collection events hashed")
```
SaveTheRbtz, line 218: these maybe a bit too much, given that both Commit and Hash are async.

smnzhu: I think the more accurate log line would be something like "collection events hashing initiated". In the callback, we can say "collection events hashed"

engine/execution/computation/computer/computer.go
```diff
@@ -175,17 +179,21 @@ func (e *blockComputer) executeBlock(
 				panic(err)
 			}
 			res.EventsHashes = append(res.EventsHashes, hash)
+			blockLogger.Debug().Msg("event hashed (callback)")
```
smnzhu, line 182: Would itbbe worth adding the collection ID to the callback and logging it so it's easier to correlate the two debug lines before and after hashing? Either that or the index

smnzhu (approved):

Kay-Zee:
going to close this, but feel free to still use it as an example if needed @simonhf

PR #2721, opened 2022-06-28
Fix time.Sleep

danielholmes839:
Updates `SubmitProposal` to return a done channel to notify the follower engine when a block has finished being processed by hotstuff.


consensus/hotstuff/follower_loop.go
```diff
@@ -25,7 +30,7 @@ func NewFollowerLoop(log zerolog.Logger, followerLogic FollowerLogic) (*Follower
 	return &FollowerLoop{
 		log:           log,
 		followerLogic: followerLogic,
-		proposals:     make(chan *model.Proposal),
+		proposals:     make(chan proposalTask),
```
zhangchiqing, line 33:
```suggestion
		proposals:     make(chan *proposalTask),
```

engine/common/follower/engine.go
```diff
@@ -368,8 +369,12 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
 	log.Info().Msg("forwarding block proposal to hotstuff")
 
 	// submit the model to follower for processing
-	e.follower.SubmitProposal(header, parent.View)
-
+	select {
+	case <-e.follower.SubmitProposal(header, parent.View):
```
zhangchiqing, line 373: this will wait for every single proposal, but we only want to wait if the block is one of the range response. Right?

consensus/hotstuff/follower_loop.go
```diff
@@ -63,7 +75,9 @@ func (fl *FollowerLoop) loop() {
 
 		select {
 		case p := <-fl.proposals:
-			err := fl.followerLogic.AddBlock(p)
+			err := fl.followerLogic.AddBlock(p.Proposal)
+			close(p.done)
```
zhangchiqing, line 79: we will need this in event_loop.go as well

zhangchiqing (commented):
Looks good. Added some comments. 

Better to add more comments on the modules and methods

engine/common/follower/engine.go
zhangchiqing: We don't want to for every single proposal, because if follower engine didn't receive the next proposal, it doesn't need to wait. 

We only want to wait if we have received a range response, because after processing a block, we know the next block has already received, and we want to wait until this block to be processed before processing the next.

In order to do that, you will need to send the entire BlockResponse to the follower engine, instead of [processing every single `SyncBlock`](https://github.com/onflow/flow-go/blob/master/engine/common/synchronization/engine.go#L305-L318)

When follower engine process a range of block, it could call onBlockProposal to process each proposal as usual, but it will also wait for the callback, whereas the original onBlockProposal will not wait for the callback.

consensus/hotstuff/follower_loop.go
danielholmes839: Writing the conclusion from our meeting here. This will be added once we confirm the changes to the follower engine are working as expected since they're almost identical changes.

engine/common/follower/engine.go
```diff
@@ -368,10 +391,24 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
 	log.Info().Msg("forwarding block proposal to hotstuff")
 
 	// submit the model to follower for processing
-	e.follower.SubmitProposal(header, parent.View)
+	if inRangeBlockResponse {
+		select {
+		case <-e.follower.SubmitProposal(header, parent.View):
+			// we wait if the block proposal was part of a range block response since
+			// processing block n+1 before block n has finished will fail
```
zhangchiqing, line 398:
```suggestion
			// after submitting proposal to hotstuff, then hotstuff will start processing block n, and follower
			// engine is concurrently processing block n + 1. 
			// however follower engine will fail to process block n + 1 if block n is not saved in protocol state.
			// Block n is only saved in protocol state when hotstuff finishes processing block n.
			// In order to ensure follower engine don't process block n + 1 too early, we wait until hotstuff finish
			// processing block n.
			// this wait is only needed when processing range block response, since blocks are processed in order.
```

engine/common/follower/engine.go
```diff
@@ -368,10 +391,24 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
...
+			break
+		case <-time.After(time.Millisecond * 200):
+			// this shouldn't happen very often. 99.8% of proposals are processed within 150ms
```
zhangchiqing, line 401:
```suggestion
			// this shouldn't happen very often. 99.8% of proposals are processed within 200ms
```

engine/common/follower/engine.go
```diff
@@ -175,11 +182,27 @@ func (e *Engine) onSyncedBlock(originID flow.Identifier, synced *events.SyncedBl
...
+			return err
+		}
+	}
+
+	return nil
 }
 
 // onBlockProposal handles incoming block proposals.
-func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.BlockProposal) error {
+func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.BlockProposal, inRangeBlockResponse bool) error {
```
zhangchiqing, line 205: can you add comment for `inRangeBlockResponse`?

engine/common/follower/engine.go
```diff
@@ -175,11 +182,27 @@ func (e *Engine) onSyncedBlock(originID flow.Identifier, synced *events.SyncedBl
...
+func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockResponse) error {
+	for _, block := range res.Blocks {
+		proposal := &messages.BlockProposal{
+			Header:  block.Header,
+			Payload: block.Payload,
+		}
+
+		// process block proposal with a wait
+		if err := e.onBlockProposal(originID, proposal, true); err != nil {
+			return err
```
zhangchiqing, line 197:
```suggestion
	for i, block := range res.Blocks {
		proposal := &messages.BlockProposal{
			Header:  block.Header,
			Payload: block.Payload,
		}

		// process block proposal with a wait
		if err := e.onBlockProposal(originID, proposal, true); err != nil {
			return fmt.Errorf("fail to process the block at index %v in a range block response that contains %v blocks: %w", i, len(res.Blocks), err)
```

zhangchiqing (approved):

engine/common/follower/engine.go
```diff
@@ -368,10 +391,29 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
...
+			// Block n is only saved in protocol state when hotstuff finishes processing block n.
+			// In order to ensure follower engine don't process block n + 1 too early, we wait until hotstuff finish
+			// processing block n.
+			// this wait is only needed when processing range block response, since blocks are processed in order.
+			break
+		case <-time.After(time.Millisecond * 200):
+			// this shouldn't happen very often. 99.8% of proposals are processed within 200ms
+			e.log.Warn().Msg("HotStuffFollower SubmitProposal timeout")
+			break
+		}
```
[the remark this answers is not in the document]
danielholmes839: We should be able to can remove the timeout case. The timeout isn't actually cancelling anything inside `SubmitProposal` and the channel always closes before another proposal can be processed.

smnzhu (approved):

engine/common/synchronization/engine.go
```diff
@@ -307,15 +306,9 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockRe
 			e.log.Debug().Uint64("height", block.Header.Height).Msg("block handler rejected")
 			continue
 		}
-		synced := &events.SyncedBlock{
```
zhangchiqing, line 310: With this change, we will be forwarding the `messages.BlockResponse` to compliance engine, instead of converting to `SyncedBlock`. 

We've changed the follower engine to handle `messages.BlockResponse`, but we haven't changed the [consensus compliance engine](https://github.com/onflow/flow-go/blob/master/engine/consensus/compliance/engine.go#L105) and [collection compliance engine](https://github.com/onflow/flow-go/blob/master/engine/collection/compliance/engine.go#L120) to process `messages.BlockResponse`, which means this will break forward syncing for them.

In order to fix it, we need to add a new FIFOQueue as `pendingRangeResponses` just like the [pendingBlocks](https://github.com/onflow/flow-go/blob/master/engine/collection/compliance/engine.go#L90), and add a [Pattern](https://github.com/onflow/flow-go/blob/master/engine/collection/compliance/engine.go#L108) to the message handler to handle the RangeResponse. 

Then when processing the RangeResponse, we will be able to pass the inRangeResponse parameter to the OnBlockProposal call.

PR #2729, opened 2022-06-30
Fix GetTransactionResultsByBlockID Panic

danielholmes839:
Fixes a panic [here](https://github.com/onflow/flow-go/blob/v0.26/engine/access/rpc/backend/backend_transactions.go#L326). This can only happen if `resp` and `err` returned from `getTransactionResultsByBlockIDFromAnyExeNode` [here](https://github.com/onflow/flow-go/blob/v0.26/engine/access/rpc/backend/backend_transactions.go#L317) are both nil. 

There are two bugs that caused this issue. The first is that `getTransactionResultsByBlockIDFromAnyExeNode` can return nil for `resp` and `err` when it is given zero execution nodes since no errors are appended to the `multierror` and that causes the call to [`ErrorOrNil`](https://github.com/onflow/flow-go/blob/3b7c152d081b61106667c46ff6b03eb625a4ac9f/engine/access/rpc/backend/backend_transactions.go#L744) to return nil. This was fixed by returning an error when the function receives 0 execution nodes.

The second bug is in `executionNodesForBlockID`. We call `executionNodesForBlockID` [here](https://github.com/onflow/flow-go/blob/v0.26/engine/access/rpc/backend/backend_transactions.go#L308-L317) to get the execution nodes for `getTransactionResultsByBlockIDFromAnyExeNode`. `executionNodesForBlockID` should return an error when it returns zero execution nodes but there's one case when the context times out where no error has been set and no execution are returned. This was fixed by returning the `ctx.Err()`


peterargue (commented):
this will stop the panic, but I think we should figure out why/how `getTransactionResultsByBlockIDFromAnyExeNode` is called with an empty list in the first place. I think there's a bug elsewhere in the code, so prevent a crash here will just change how the bug manifests

danielholmes839:
cleaned up `getTransactionResultsByBlockIDFromAnyExeNode` and fixed `executionNodesForBlockID`. `executionNodesForBlockID` would return any empty list of execution nodes when the context cancels and no error has been set.

engine/access/rpc/backend/backend_transactions.go
```diff
@@ -716,13 +716,7 @@ func (b *backendTransactions) getTransactionResultsByBlockIDFromAnyExeNode(
 	req execproto.GetTransactionsByBlockIDRequest,
 ) (*execproto.GetTransactionResultsResponse, error) {
 	var errs *multierror.Error
-	logAnyError := func() {
```
peterargue, line 719: I think we should keep this. It's intended to make sure we log errors from failing nodes even if the call succeeded

engine/access/rpc/backend/backend_transactions.go
```diff
@@ -737,6 +731,14 @@ func (b *backendTransactions) getTransactionResultsByBlockIDFromAnyExeNode(
 		}
 		errs = multierror.Append(errs, err)
 	}
+
+	// if we were passed 0 execution nodes add a specific error
+	if len(execNodes) == 0 {
```
peterargue, line 736: this should go above the loop and you can simplify by must returning the error directly instead of adding it to `errs`

engine/access/rpc/backend/backend_transactions.go
```diff
@@ -767,6 +769,11 @@ func (b *backendTransactions) getTransactionResultByIndexFromAnyExeNode(
 		}
 	}
 	defer logAnyError()
+
+	if len(execNodes) == 0 {
+		errs = multierror.Append(errs, errors.New("zero execution nodes provided"))
```
peterargue, line 774: this should just return the error directly

engine/access/rpc/backend/backend_transactions.go
danielholmes839: I can keep it but it only logs the errors that come from the `multierror`. Or should it log the not found error and the error when it's passed 0 execution nodes as well?

peterargue: it should just log the multierror. we just want to make sure we log errors from ENs even if we eventually succeed

peterargue (approved):
looks good. Thanks for the fixes @danielholmes839!

koko1123 (approved):

PR #2739, opened 2022-06-30
[Exec] Remove Stop The World pauses from transaction/script execution

SaveTheRbtz:
Currently we call [runtime.ReadMemStats](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/runtime/mstats.go;l=403) for each transaction and script execution.  This has a lot of overhead, esp. on high-core-count boxes we use.

```
name                             old us/tx    new us/tx    delta
ComputeBlock/1/cols/16/txes-8       586 ± 1%     533 ± 6%   -8.97%  (p=0.000 n=8+10)
ComputeBlock/1/cols/32/txes-8       557 ± 1%     490 ± 0%  -11.95%  (p=0.000 n=9+9)
ComputeBlock/1/cols/64/txes-8       542 ± 2%     496 ± 9%   -8.43%  (p=0.000 n=9+10)
ComputeBlock/1/cols/128/txes-8      538 ± 1%     475 ± 2%  -11.60%  (p=0.000 n=9+10)
ComputeBlock/4/cols/16/txes-8       551 ± 4%     497 ± 7%   -9.74%  (p=0.000 n=10+10)
ComputeBlock/4/cols/32/txes-8       551 ± 3%     471 ± 1%  -14.59%  (p=0.000 n=10+10)
ComputeBlock/4/cols/64/txes-8       543 ± 1%     468 ± 3%  -13.81%  (p=0.000 n=8+9)
ComputeBlock/4/cols/128/txes-8      528 ± 2%     460 ± 4%  -12.83%  (p=0.000 n=9+9)
ComputeBlock/16/cols/16/txes-8      538 ± 4%     472 ± 0%  -12.28%  (p=0.000 n=9+8)
ComputeBlock/16/cols/32/txes-8      528 ± 1%     470 ± 2%  -11.02%  (p=0.000 n=9+8)
ComputeBlock/16/cols/64/txes-8      530 ± 1%     460 ± 4%  -13.16%  (p=0.000 n=10+10)
ComputeBlock/16/cols/128/txes-8     521 ± 1%     455 ± 0%  -12.70%  (p=0.000 n=9+8)
```


engine/execution/computation/computer/computer.go
```diff
@@ -321,10 +321,13 @@ func (e *blockComputer) executeTransaction(
...
 
+	sample := []metrics.Sample{{Name: "/gc/heap/allocs:bytes"}}
+	metrics.Read(sample)
+
 	var memAllocBefore uint64
-	var m runtime.MemStats
-	runtime.ReadMemStats(&m)
-	memAllocBefore = m.TotalAlloc
+	if sample[0].Value.Kind() == metrics.KindUint64 {
+		memAllocBefore = sample[0].Value.Uint64()
```
pattyshack, line 329: too much copy/paste ... refactor this into a function.  maybe stick it into util/debug?

utils/debug/runtime.go
```diff
@@ -0,0 +1,16 @@
+package debug
+
+import (
+	"runtime/metrics"
+)
+
+// GetHeapAllocsBytes returns the value of /gc/heap/allocs:bytes.
+func GetHeapAllocsBytes() uint64 {
+	sample := []metrics.Sample{{Name: "/gc/heap/allocs:bytes"}}
```
pattyshack, line 9: maybe just alloc this once globally

pattyshack (approved):

utils/debug/runtime.go
SaveTheRbtz: since it is modified in place i think it is better to leave it like this for now since the locking cost will be higher. Also, hopefully the compiler is smart enough to not escape this one to the heap.

SaveTheRbtz:
I think it should not. [In the docs](https://pkg.go.dev/runtime/metrics) `/gc/heap/allocs:bytes` is defined as:

> /gc/heap/allocs:bytes
>	Cumulative sum of memory allocated to the heap by the application.

So I would assume that it matches the old `TotalAlloc`.

dsainati1:
Either way the total alloc measurements were just a sanity check on the memory metering, so if it changes slightly that shouldn't be an issue. Thanks for fixing this and sorry about the regression!

utils/debug/runtime.go
zhangchiqing, line 9:
```suggestion
	// https://pkg.go.dev/runtime/metrics#hdr-Supported_metrics
	sample := []metrics.Sample{{Name: "/gc/heap/allocs:bytes"}}
```

Better to add the reference for the "magic string"

zhangchiqing (approved):
Looks good. Good catch!

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors retry

PR #2676, opened 2022-06-22
[Consensus] `HasEpochForView` for consensus committee

durkmurder:
### Context

Extended consensus committee to support `HasEpochForView` which will be used by `TimeoutAggregator` to filter our timeout objects for views that are referring future or previous epochs.

PR #2740, opened 2022-07-01
[Consensus] Active Pacemaker: `TimeoutAggregator` implementation

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6195

### Context

This PR implements core logic for `TimeoutAggregator`. Design ideas and implementation is similar to `VoteAggregator`.
`TimeoutAggregator` implements concurrency logic for concurrent timeout processing, it uses queue to distribute incoming events to worker threads for further processing. It also handles `OnEnteringView` events to perform pruning of stale timeout collectors. 

TODO: replace `LeaderForView` call with `HasEpochForView` from https://github.com/onflow/flow-go/pull/2676.


consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+// TestAddTimeout_EpochUnknown tests if timeout objects targeting unknown epoch should be ignored
+func (s *TimeoutAggregatorTestSuite) TestAddTimeout_EpochUnknown() {
+	*s.committee = *mocks.NewReplicas(s.T())
+	timeout := helper.TimeoutObjectFixture(helper.WithTimeoutObjectView(s.lowestRetainedView))
+	s.committee.On("LeaderForView", timeout.View).Return(nil, model.ErrViewForUnknownEpoch).Once()
+	s.aggregator.AddTimeout(timeout)
+	require.Eventually(s.T(),
+		func() bool {
+			return s.committee.AssertCalled(s.T(), "LeaderForView", timeout.View)
```
jordanschalm, line 116: Same comment here about the potential race condition

consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+// TestOnEnteringView tests if entering view event gets processed when send through `TimeoutAggregator`.
+// Tests the whole processing pipeline.
+func (s *TimeoutAggregatorTestSuite) TestOnEnteringView() {
+	view := s.lowestRetainedView + 1
+	s.collectors.On("PruneUpToView", view).Once()
+	s.aggregator.OnEnteringView(view, unittest.IdentifierFixture())
+	require.Eventually(s.T(),
+		func() bool {
+			return s.collectors.AssertCalled(s.T(), "PruneUpToView", view)
```
jordanschalm, line 139: Same comment here about the potential race condition

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+// GetOrCreateCollector retrieves the hotstuff.TimeoutCollector for the specified
+// view or creates one if none exists.
+//  -  (collector, true, nil) if no collector can be found by the view, and a new collector was created.
+//  -  (collector, false, nil) if the collector can be found by the view
+//  -  (nil, false, error) if running into any exception creating the timeout collector state machine
+// Expected error returns during normal operations:
+//  * mempool.DecreasingPruningHeightError - in case view is lower than lowestRetainedView
```
jordanschalm, line 46: In this application, we're using `DecreasingPruningHeightError` for a slightly different purpose than its documentation and other uses.

The [Godoc](https://github.com/onflow/flow-go/blob/1f7ce61f6a6fbe1f0cbf708de2712ae416533442/module/mempool/errors.go#L37-L38) for `DecreasingPruningHeightError` says:
```go
// DecreasingPruningHeightError indicates that we are pruning a mempool by a height that
// is lower than existing height
```

But here (and in most usages of the error type), we return the error when we are querying, not when we are pruning. I'd suggest to change the error name and documentation to correspond better to the usage:
```go
// BelowPrunedThresholdError indicates that we are attempting to query or prune a mempool by a 
// key (typically block height or block view) which is lower than the lowest retained key threshold. 
// In other words, we have already pruned above the queried key.
type BelowPrunedThresholdError ...
```

consensus/hotstuff/timeoutaggregator/timeout_collectors_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+// TimeoutCollectorsTestSuite is a test suite for isolated testing of TimeoutCollectors.
+// Contains helper methods and mocked state which is used to verify correct behavior of TimeoutCollectors.
+type TimeoutCollectorsTestSuite struct {
+	suite.Suite
+
+	mockedCollectors map[uint64]*mocks.TimeoutCollector
+	factoryMethod    NewCollectorFactoryMethod
+	collectors       *TimeoutCollectors
+	lowestLevel      uint64
```
jordanschalm, line 34:
```suggestion
	lowestView      uint64
```
This is equivalent to lowest retained view right? It's clearer for me to use the view terminology

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+	_, err := t.committee.LeaderForView(timeoutObject.View)
+	if err != nil {
+		// ignore TO if we don't have information for epoch
+		if errors.Is(err, model.ErrViewForUnknownEpoch) {
+			return nil
+		}
+		return fmt.Errorf("unknown error when querying epoch state for view %d: %w", timeoutObject.View, err)
+	}
+
+	collector, created, err := t.collectors.GetOrCreateCollector(timeoutObject.View)
```
jordanschalm, line 151: I'm wondering if we should limit how far ahead of the current view we will allow TOs to be, and still cache them. Thinking "out loud" a bit here 👇 

Epochs last about ~500,000 views currently, and we create a collector object for a view before validating the first TO for that view. So an adversary could in the worst case cause us to create ~500,000 unnecessary TimeoutCollector objects in memory which we would prune very slowly. Some thoughts:
* If this happens, it would impact memory usage but not pruning, because of the [clever pruning optimization](https://github.com/onflow/flow-go/pull/2740/files#diff-edb27b0b5326b242a3b35e845cf264382ded9b3ae2f92fd6a50dc55531c4d051R110-R113)
* To accomplish this, an adversary would need to send invalid TCs, which is a slashable offense. In the mature protocol, the adversary should be ejected before they can send enough messages
* The attack is expensive (sending messages linear with the resulting adverse memory usage increase)

Overall, I don't think we need any code changes to deal with this spamming potential, but we could add a comment about why its safe:
```suggestion
	// We create a timeout collector before validating the first TO, so processing an invalid TO will
	// result in a collector being added, until the corresponding view is pruned.
	// Since epochs on Mainnet have 500,000-1,000,000 views, there is an opportunity for memory exhaustion here.
	// However, because only invalid TOs can result in creating collectors at an arbitrary view,
	// and invalid TOs are slashable, resulting eventually in ejection of the sender, this attack is impactical.
	collector, created, err := t.collectors.GetOrCreateCollector(timeoutObject.View)
```

consensus/hotstuff/timeoutaggregator/timeout_collectors_test.go
```diff
@@ -0,0 +1,158 @@
...
+			_, created, err := s.collectors.GetOrCreateCollector(view)
+			require.NoError(s.T(), err)
+			if created {
+				createdTimes.Add(1)
+			}
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
```
jordanschalm, line 117:
```suggestion
	unittest.AssertReturnsBefore(s.T(), wg.Wait, time.Second)
```

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// It's ok to silently drop timeouts in case our processing pipeline is full.
+	// It means that we are probably catching up.
+	if ok := t.queuedTimeouts.Push(timeoutObject); ok {
+		t.queuedTimeoutsNotifier.Notify()
+	}
+}
+
+// PruneUpToView deletes all timeouts _below_ to the given view, as well as
+// related indices. We only retain and process whose view is equal or larger
```
AlexHentschel, line 201: incomplete sentence (?)
```suggestion
// related indices. We only retain and process `TimeoutCollector`s, whose view is equal or larger
```

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+
+		return fmt.Errorf("could not process TO for view %d: %w",
+			timeoutObject.View, err)
+	}
+
+	return nil
+}
+
+// AddTimeout checks if TO is stale and appends TO into processing queue
+// actual processing will be called in other dispatching goroutine.
```
AlexHentschel, line 180:
```suggestion
// AddTimeout checks if TO is stale and appends TO to processing queue.
// The actual processing will be done asynchronously by one of the `TimeoutAggregator` internal worker routines.
```

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
AlexHentschel: good observation. Agree on no code change needed but adding documentation.

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+func (t *TimeoutAggregator) processQueuedTimeout(timeoutObject *model.TimeoutObject) error {
+	// TODO: replace this check by a specific function to query if there is epoch by view
+	_, err := t.committee.LeaderForView(timeoutObject.View)
+	if err != nil {
+		// ignore TO if we don't have information for epoch
+		if errors.Is(err, model.ErrViewForUnknownEpoch) {
+			return nil
+		}
+		return fmt.Errorf("unknown error when querying epoch state for view %d: %w", timeoutObject.View, err)
+	}
```
AlexHentschel, line 149: In my mind, it would be more intuitive to have this logic for handling unknown future views to be included in `TimeoutCollectors`. For `TimeoutCollectors` we should anyway specify how it would handle a call to `GetOrCreateCollector` with a view for a future unknown epoch.

Concrete suggestion:
* https://github.com/onflow/flow-go/blob/f40669d422f7f67b193c0dc8a35e95220559fd56/consensus/hotstuff/timeoutaggregator/timeout_collectors.go#L13-L14 could return an `model.ErrViewForUnknownEpoch`
* include handling of `ErrViewForUnknownEpoch` in [line 154](https://github.com/onflow/flow-go/blob/f40669d422f7f67b193c0dc8a35e95220559fd56/consensus/hotstuff/timeoutaggregator/timeout_aggregator.go#L154)

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+			timeoutObject.View, err)
+	}
+	if created {
+		t.log.Info().Uint64("view", timeoutObject.View).Msg("timeouts collector is created by processing TO")
+	}
+
+	err = collector.AddTimeout(timeoutObject)
+	if err != nil {
+		if model.IsDoubleTimeoutError(err) {
+			doubleTimeoutErr := err.(model.DoubleTimeoutError)
```
AlexHentschel, line 167: ## ⚠️ ❓ 
`DoubleTimeoutError` has already a convenience function `AsDoubleTimeoutError`, which includes the type cast. This is important, because `err` might be a _wrapped_ `DoubleTimeoutError` in which cast the case would panic (I think?)
```suggestion
		if doubleTimeoutErr, is := model.AsDoubleTimeoutError(err); is {
```

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("could not get collector for view %d: %w",
+			timeoutObject.View, err)
+	}
+	if created {
+		t.log.Info().Uint64("view", timeoutObject.View).Msg("timeouts collector is created by processing TO")
+	}
+
+	err = collector.AddTimeout(timeoutObject)
+	if err != nil {
+		if model.IsDoubleTimeoutError(err) {
```
AlexHentschel, line 166: ## ⚠️ 
There is no documentation in the API that `TimeoutCollector.AddTimeout( .. )` can return `DoubleTimeoutError`. To the best of my understanding, `TimeoutCollector` already handles duplicates internally (as well as invalid timeouts), so we don't have to do anything here. 

(-> [code](https://github.com/onflow/flow-go/blob/f40669d422f7f67b193c0dc8a35e95220559fd56/consensus/hotstuff/timeoutcollector/timeout_collector.go#L52-L62))

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+		// ignore if our routine is outdated and some other one has pruned collectors
+		if mempool.IsDecreasingPruningHeightError(err) {
+			return nil
+		}
+		return fmt.Errorf("could not get collector for view %d: %w",
+			timeoutObject.View, err)
+	}
+	if created {
+		t.log.Info().Uint64("view", timeoutObject.View).Msg("timeouts collector is created by processing TO")
+	}
```
AlexHentschel, line 162: To me, it would be more intuitive to include this log in `TimeoutCollectors` ([here](https://github.com/onflow/flow-go/blob/f40669d422f7f67b193c0dc8a35e95220559fd56/consensus/hotstuff/timeoutaggregator/timeout_collectors.go#L74))

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+
+// AddTimeout checks if TO is stale and appends TO into processing queue
+// actual processing will be called in other dispatching goroutine.
+func (t *TimeoutAggregator) AddTimeout(timeoutObject *model.TimeoutObject) {
+	// drop stale objects
+	if timeoutObject.View < t.lowestRetainedView.Value() {
+		t.log.Info().
+			Uint64("view", timeoutObject.View).
+			Hex("signer", timeoutObject.SignerID[:]).
+			Str("timeout_id", timeoutObject.ID().String()).
```
AlexHentschel, line 187: While the TO's `view` and `SignerID` are very relevant, I feel the `ID` is practically of no interest (because we don't store TOs in the data base, for which the ID could be useful to retrieve later)

```suggestion
```

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+			timeoutObject := msg.(*model.TimeoutObject)
+			err := t.processQueuedTimeout(timeoutObject)
+			if err != nil {
+				return fmt.Errorf("could not process pending TO %v: %w", timeoutObject.ID(), err)
+			}
+
+			t.log.Info().
+				Uint64("view", timeoutObject.View).
+				Str("timeout_id", timeoutObject.ID().String()).
+				Msg("TO has been processed successfully")
```
AlexHentschel, line 127: same here: TO's ID is probably of little interest but expensive to compute. On the other hand, we don't include the signer, which is probably relevant:
```suggestion
				Hex("signer", timeoutObject.SignerID[:]).
				Msg("TO has been processed successfully")
```

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+	"sync"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/consensus/hotstuff"
+	"github.com/onflow/flow-go/module/mempool"
+)
+
+// NewCollectorFactoryMethod is a factory method to generate a TimeoutCollector for concrete view
+type NewCollectorFactoryMethod = func(view uint64) (hotstuff.TimeoutCollector, error)
```
AlexHentschel, line 14: I think we should allow the factory method to return a `model.ErrViewForUnknownEpoch` as an expected error.

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
AlexHentschel, line 46: I feel it should be legal for `TimeoutCollectors` to return a  `model.ErrViewForUnknownEpoch`:

```suggestion
	//  * mempool.DecreasingPruningHeightError if view is below the pruning threshold
	//  * model.ErrViewForUnknownEpoch if view is not yet pruned but no epoch containing the given view is known
```

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+//  * mempool.DecreasingPruningHeightError - in case view is lower than lowestRetainedView
+func (t *TimeoutCollectors) getCollector(view uint64) (hotstuff.TimeoutCollector, bool, error) {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+	if view < t.lowestRetainedView {
+		return nil, false, mempool.NewDecreasingPruningHeightErrorf("cannot retrieve collector for pruned view %d (lowest retained view %d)", view, t.lowestRetainedView)
+	}
+
+	clr, found := t.collectors[view]
+
```
AlexHentschel, line 89:
```suggestion
```

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+//  -  (collector, false, nil) if the collector can be found by the view
+//  -  (nil, false, error) if running into any exception creating the timeout collector state machine
+// Expected error returns during normal operations:
+//  * mempool.DecreasingPruningHeightError - in case view is lower than lowestRetainedView
+func (t *TimeoutCollectors) GetOrCreateCollector(view uint64) (hotstuff.TimeoutCollector, bool, error) {
+	cachedCollector, hasCachedCollector, err := t.getCollector(view)
+	if err != nil {
+		return nil, false, err
+	}
+
```
AlexHentschel, line 52:
```suggestion
```

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+	defer t.lock.Unlock()
+	if t.lowestRetainedView >= lowestRetainedView {
+		return
+	}
+	if len(t.collectors) == 0 {
+		t.lowestRetainedView = lowestRetainedView
+		return
+	}
+
+	sizeBefore := len(t.collectors)
```
AlexHentschel, line 108: I would move the declaration of `sizeBefore` just before [line 103](https://github.com/onflow/flow-go/blob/f40669d422f7f67b193c0dc8a35e95220559fd56/consensus/hotstuff/timeoutaggregator/timeout_collectors.go#L103) and then use `sizeBefore` everywhere

AlexHentschel (commented):
Looks good. Overall mostly stylistic comments.

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
```diff
@@ -0,0 +1,231 @@
...
+
+// NewTimeoutAggregator creates an instance of timeout aggregator
+// No errors are expected during normal operations.
+func NewTimeoutAggregator(log zerolog.Logger,
+	notifier hotstuff.Consumer,
+	lowestRetainedView uint64,
+	committee hotstuff.Replicas,
+	collectors hotstuff.TimeoutCollectors,
+) (*TimeoutAggregator, error) {
+	queuedTimeouts, err := fifoqueue.NewFifoQueue(fifoqueue.WithCapacity(defaultTimeoutQueueCapacity))
```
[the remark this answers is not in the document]
durkmurder: I was planning to address all metrics in separate issue, I will add a TODO.

consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
```diff
@@ -0,0 +1,141 @@
...
+	collector := mocks.NewTimeoutCollector(s.T())
+	callCount := atomic.NewUint64(0)
+	collector.On("AddTimeout", mock.Anything).Run(func(mock.Arguments) {
+		callCount.Add(1)
+	}).Return(nil).Times(timeoutsCount)
+	s.collectors.On("GetOrCreateCollector", s.lowestRetainedView).Return(collector, true, nil).Times(timeoutsCount)
+
+	var wg, startWg sync.WaitGroup
+	wg.Add(timeoutsCount)
+	startWg.Add(1)
```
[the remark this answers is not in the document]
durkmurder: `wg.Wait` doesn't give guarantee that calls indeed were made since `AddTimeout` is async so we definitely need to use `callCount`. I've forgot to wait on `startWg`, the idea was to delay start of processing until all goroutines are spawned. Update it.

consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
```diff
@@ -0,0 +1,141 @@
...
+	timeout := helper.TimeoutObjectFixture(helper.WithTimeoutObjectView(s.lowestRetainedView))
+	collector.On("AddTimeout", timeout).Return(model.NewDoubleTimeoutErrorf(timeout, timeout, "")).Once()
+	s.collectors.On("GetOrCreateCollector", s.lowestRetainedView).Return(collector, true, nil).Once()
+	s.consumer.On("OnDoubleTimeoutDetected", mock.Anything, mock.Anything).Once()
+	s.aggregator.AddTimeout(timeout)
+
+	require.Eventually(s.T(),
+		func() bool {
+			return s.consumer.AssertCalled(s.T(), "OnDoubleTimeoutDetected", timeout, timeout)
+		}, time.Second, time.Millisecond*20)
```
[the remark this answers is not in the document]
durkmurder: Why do you think it's bad the way it is right now? That's exactly the case why I use `Eventually` to make sure that eventually this assertion will be satisfied.

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -0,0 +1,134 @@
...
+// TimeoutCollectors implements management of multiple timeout collectors indexed by view.
+// Implements hotstuff.TimeoutCollectors interface. Creating a TimeoutCollector for a
+// particular view is lazy (instances are created on demand).
+// This structure is concurrently safe.
+// TODO: once VoteCollectors gets updated to stop managing worker pool we can merge VoteCollectors and TimeoutCollectors using generics
+type TimeoutCollectors struct {
+	log                zerolog.Logger
+	lock               sync.RWMutex
+	lowestRetainedView uint64                               // lowest view, for which we still retain a TimeoutCollector and process timeouts
+	collectors         map[uint64]hotstuff.TimeoutCollector // view -> TimeoutCollector
```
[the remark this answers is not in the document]
durkmurder: Added TODO

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
durkmurder: I have moved check for epoch into `TimeoutCollectors` and now `GetOrCreateCollector` checks for both pruning height and unknown epoch error conditions. I am not really following why we need `model.ErrViewForUnknownEpoch` for factory as well?

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
durkmurder: Moved it

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
durkmurder: answered [here](https://github.com/onflow/flow-go/pull/2740#discussion_r915019387)

consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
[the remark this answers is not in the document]
durkmurder: Oh, that was painful, I didn't know it has side-effects, makes sense this way. I will change this. Thanks.

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
durkmurder: https://github.com/onflow/flow-go/pull/2740/commits/35d88f2fc769aa3d99586fba43d4830c6cf0c132

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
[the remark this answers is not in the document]
jordanschalm: Makes sense - is there an issue already? I took a quick look and couldn't find one. If not, might be nice to create that now, so we can add items to it as we implement/review.

consensus/hotstuff/timeoutaggregator/timeout_collectors.go
```diff
@@ -43,17 +46,23 @@ func NewTimeoutCollectors(log zerolog.Logger, lowestRetainedView uint64, createC
 //  -  (collector, false, nil) if the collector can be found by the view
 //  -  (nil, false, error) if running into any exception creating the timeout collector state machine
 // Expected error returns during normal operations:
-//  * mempool.DecreasingPruningHeightError - in case view is lower than lowestRetainedView
+//  * mempool.DecreasingPruningHeightError if view is below the pruning threshold
```
jordanschalm, line 49:
```suggestion
//  * model.BelowPrunedThresholdError if view is below the pruning threshold
```

jordanschalm (approved):

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
AlexHentschel: Sorry, I skipped this detail in my comment. Here is my reasoning why the factory method should be allowed to return a `model.ErrViewForUnknownEpoch`:
* During a factory method call, we need to query the  [`Replicas` API](https://github.com/onflow/flow-go/blob/35d88f2fc769aa3d99586fba43d4830c6cf0c132/consensus/hotstuff/committee.go#L31) to get the set of authorized consensus nodes, their keys, quorum thresholds. This all is needed to initialize the subcomponents like `TimeoutProcessor`, `TimeoutSignatureAggregator`, ...
* Obviously that is going to error if we feed a view for an unknown future epoch into the `NewCollectorFactoryMethod`. And the error source will be the `Replicas` interface. 
* From the perspective of the `NewCollectorFactoryMethod` and all the layers below, this is a completely benign error. The `ErrViewForUnknownEpoch` just says that we have asked about something `Replicas` didn't know about. 

I am saying that the code already nicely bubbles up the consistent error. We only have to document it properly. Then we don't need all these pre-checks just to make sure that the Epoch information for the view is available. For me that just seems cleaner 🤷 .



https://github.com/onflow/flow-go/blob/35d88f2fc769aa3d99586fba43d4830c6cf0c132/consensus/hotstuff/timeoutcollector/timeout_processor.go#L95-L96

AlexHentschel (approved):
Looks good. Only critical one is this sentinel mixup here.

consensus/hotstuff/timeoutaggregator/timeout_aggregator_test.go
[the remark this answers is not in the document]
durkmurder: Yeah, I think we can remove it.

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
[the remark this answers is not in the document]
durkmurder: https://github.com/dapperlabs/flow-go/issues/6319

consensus/hotstuff/model/errors.go
```diff
@@ -398,3 +398,30 @@ func IsInvalidTimeoutError(err error) bool {
...
+
+func (e BelowPrunedThresholdError) Error() string {
+	return e.err.Error()
+}
+
+// IsBelowPrunedThresholdError returns whether the given error is an BelowPrunedThresholdError error
+func IsBelowPrunedThresholdError(err error) bool {
+	var newIsBelowPrunedThresholdError BelowPrunedThresholdError
+	return errors.As(err, &newIsBelowPrunedThresholdError)
+}
```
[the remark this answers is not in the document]
durkmurder: Sounds good, https://github.com/dapperlabs/flow-go/issues/6318

consensus/hotstuff/timeoutaggregator/timeout_aggregator.go
durkmurder: Yeah, that works, I was thinking about a separate call to committee to make it explicit but that works as well.

durkmurder:
bors merge

PR #2742, opened 2022-07-01
[Exec] Remove global lock from script execution

SaveTheRbtz:
global random uses a global `lockedSource` that has implicit global synchronization.  Hence, introduce explicit locking and hide it behind the `e.log.Debug().Enabled()`.

### DIFF
--- a/engine/execution/computation/manager.go
+++ b/engine/execution/computation/manager.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"runtime"
 	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


pattyshack (approved):

tarakby (approved):
Nice catch, and thanks for adding the initialization too.

peterargue (approved):

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors retry

Issue #2273, opened 2022-04-07
[Uploader] Make block data upload robust against failures


smnzhu:
Currently if EN is restarted or crashes after a block gets executed but before the block data is successfully uploaded to GCP, then upon restart that block won’t get uploaded.

Currently there is no retry mechanism, because it would be rather hard to do - it requires trie updates which aren't easily obtained outside of execution window.

Perhaps we can persist the block data to disk before marking the execution of the block as complete. This way we have it saved in case we need to retry, but at the same time we don’t need to wait for the upload to complete, this part can still happen async.

j1010001:
Update after chat with Maks:
When EN restarts it tries to catch up and execute blocks. It loads them into a memory and then start executing
but if there are a lots of blocks, they won't fit into the memory. Ideally, we would have some form of a buffer or list - load blocks from one end, and execute from the other, so it makes progress. 
In a recent issue on testnet where we had to re0execute blocks we patched EN to load 100k blocks, execute and crash. We kept restarting the EN until it executed and uploaded all blocks.

Tonix517:
discussed with @m4ksio , the out of memory failure can be tracked separately: #2669

PR #2743, opened 2022-07-01
Block computation result upload Retry implementation

Tonix517:
#2273 - This PR is supposed to address the problem described in this issue, that sometimes ComputationResult upload failed but was not retried so it caused data loss in GCP.

### Design

According to the issue #2273, this problem usually occurred at abrupt crash or reboot of EN process before all computation result upload succeeded. To address this, we will need to store the upload status and let EN pick up for retry at boot. The basic idea of this PR is simple:

1. After after one ComputationResult is ready, we store one entry to mark the upload status as `false` with given ExecutionDataID into local BadgerDB
2. When upload is completed successfully, the corresponding item in BadgerDB will be replaced with `true`
3. If upload fails for certain ComputationResult items, EN process will pick them up and retry upload at reboot.

### Metrics

Two new metrics are added in to monitor upload behavior:
- Count of successful upload
- Count of upload retry

Examples of these metrics of different scenarios can be found in the Test Plan section below.

### Notes

- ComputationResult reconstruction from EDS and BadgerDB
We don't directly store whole ComputationResult instance into BadgerDB, because that'll take too much space. Actually all required fields in ComputationResult (the ones in `BlockData`) have already stored in EDS and BadgerDB, so we grab `Events` and `TrieUpdates` from EDS, and all other fields from BadgerDB to reconstruct ComputationResult for reload.

- Storage overhead
We don't expect it will take too much extra storage space in BadgerDB, since it will only be an extra boolean value per ComputationResult. We still keep these markers even after one ComputationResult upload is done for potential future devops uses.

- AWS uploader
Confirmed that we now only cares about GCP uploader. To support new uploaders we will need to introduce new key code for BadgerDB (more details in the comment of the PR)

- Duplicated upload
Confirmed that duplicated upload of the same ComputationResult record is fine and will not trigger any problems.

### Test Plan

- [X] **UT**: `go test --tags=relic` on all new unit test cases

- [X] **localnet**: no stored ComputationResult status in DB to retry -> all upload succeeds
retry_ no_prev _ upload_ok

- [X] **localnet**: no stored ComputationResult status in DB to retry -> all upload fails
retry_ no_prev _ upload_fail

- [X] **localnet**: with stored ComputationResult status in DB to retry -> retry fails 
retry_ prev_some _ retry_upload_fail

- [X] **localnet**: with stored ComputationResult status 
retry_ prev_some _ retry_upload_ok
in DB to retry -> retry succeeds + all upload succeeds

- [ ]  **devnet**


ramtinms (approved):
looks good to me

Tonix517:
Update: did a quick offline review with @m4ksio . Things to refine: instead of storing ComputationResult into BadgerDB, which may introduce extra storage overhead(could be hundreds of GB), we can simply utilize data we already stored in EDS to re-construct ComputationResult at upload retry, and only store upload status flag in BadgerDB.

storage/badger/computation_result.go
```diff
@@ -0,0 +1,48 @@
...
+	db *badger.DB
+}
+
+func NewComputationResultUploadStatus(db *badger.DB) *ComputationResultUploadStatus {
+	return &ComputationResultUploadStatus{
+		db: db,
+	}
+}
+
+func (c *ComputationResultUploadStatus) Store(computationResultID flow.Identifier,
```
m4ksio, line 20: Why don't key by blockID?

storage/badger/computation_result.go
```diff
@@ -0,0 +1,48 @@
...
+		return operation.GetComputationResultUploadStatus(computationResultID, &ret)(btx)
+	})
+	if err != nil {
+		return false, err
+	}
+
+	return ret, nil
+}
+
+func (c *ComputationResultUploadStatus) Remove(computationResultID flow.Identifier) error {
```
m4ksio, line 44: Would we ever remove the status? I think once uploaded, we set it to true, but not remove

storage/badger/operation/computation_result.go
```diff
@@ -0,0 +1,45 @@
...
+func GetComputationResultUploadStatus(computationResultID flow.Identifier,
+	wasUploadCompleted *bool) func(*badger.Txn) error {
+	return retrieve(makePrefix(codeComputationResults, computationResultID), wasUploadCompleted)
+}
+
+// GetAllComputationResultIDs returns all IDs of stored ComputationResult instances.
+func GetAllComputationResultIDs(computationResultIDs *[]flow.Identifier) func(*badger.Txn) error {
+	return traverse(makePrefix(codeComputationResults), func() (checkFunc, createFunc, handleFunc) {
+		check := func(key []byte) bool {
+			*computationResultIDs = append(*computationResultIDs, flow.HashToID(key[1:]))
```
m4ksio, line 31: I think we usually only just check the entry in this function, and collect data in `handle` func

storage/badger/operation/computation_result.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// GetComputationResult returns stored ComputationResult instance with given ID.
+func GetComputationResultUploadStatus(computationResultID flow.Identifier,
+	wasUploadCompleted *bool) func(*badger.Txn) error {
+	return retrieve(makePrefix(codeComputationResults, computationResultID), wasUploadCompleted)
+}
+
+// GetAllComputationResultIDs returns all IDs of stored ComputationResult instances.
+func GetAllComputationResultIDs(computationResultIDs *[]flow.Identifier) func(*badger.Txn) error {
```
m4ksio, line 28: Does it work as expected? it seems to collect all the entries where value is true?

engine/execution/ingestion/engine.go
```diff
@@ -141,6 +149,12 @@ func New(
 // successfully started.
 func (e *Engine) Ready() <-chan struct{} {
 	if !e.pauseExecution {
+		if uploadEnabled {
+			if err := e.computationManager.RetryUpload(); err != nil {
```
m4ksio, line 153: Is there a reason to keep `RetryUpload` as a part of `ComputationManager`? It doesn't seem to be related, and at this point `RetryUpload` doesn't really have anything to do with computation

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+
+			// Removing existing upload status from BadgerDB
+			if err = uploadStatusStore.Remove(computationResult.ExecutionDataID); err != nil {
+				log.Warn().Msg(fmt.Sprintf(
+					"ComputationResults with ID %s failed to be removed on local storage. ERR: %s ",
+					computationResult.ExecutionDataID.String(), err.Error()))
+			}
+			// Store upload status as Done(true) anyway
+			if err := uploadStatusStore.Store(computationResult.ExecutionDataID,
```
m4ksio, line 74: Why do we remove and insert instead of just updating?
Isn't that creating a potential race condition where uploaded entry can be tried because its not in the database?

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+	commits storage.Commits,
+	transactionResults storage.TransactionResults,
+	uploadStatusStore storage.ComputationResultUploadStatus,
+	execDataService state_synchronization.ExecutionDataService,
+	metrics module.ExecutionMetrics) *BadgerRetryableUploader {
+
+	// check storage\eds interface params
+	if blocks == nil || commits == nil || transactionResults == nil ||
+		uploadStatusStore == nil || execDataService == nil {
+		panic("Not all storage parameters are valid")
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	// NOTE: only AsyncUploader is supported for now.
+	switch uploader.(type) {
+	case *AsyncUploader:
+		// When Uploade() is successful, the stored ComputationResult in BadgerDB will be removed.
+		onCompleteCB := func(computationResult *execution.ComputationResult, err error) {
+			if err != nil {
+				log.Warn().Msg(fmt.Sprintf("ComputationResults upload failed with ID %s",
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+
+			if computationResult == nil {
+				log.Warn().Msg("nil ComputationResult parameter")
+				return
+			}
+
+			// Removing existing upload status from BadgerDB
+			if err = uploadStatusStore.Remove(computationResult.ExecutionDataID); err != nil {
+				log.Warn().Msg(fmt.Sprintf(
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+	case module.ReadyDoneAware:
+		readyDoneAwareUploader := b.uploader.(module.ReadyDoneAware)
+		return readyDoneAwareUploader.Done()
+	}
+	return b.unit.Done()
+}
+
+func (b *BadgerRetryableUploader) Upload(computationResult *execution.ComputationResult) error {
+	if computationResult == nil || computationResult.ExecutableBlock == nil {
+		return errors.New("ComputationResult or its ExecutableBlock is nil when Upload() is called.")
```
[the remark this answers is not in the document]
Tonix517: I've left all error msg with format params to `fmt.Errorf` and used `errors.New` for those without format params.

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+		blocks:             blocks,
+		commits:            commits,
+		transactionResults: transactionResults,
+		uploadStatusStore:  uploadStatusStore,
+	}
+}
+
+func (b *BadgerRetryableUploader) Ready() <-chan struct{} {
+	switch b.uploader.(type) {
+	case module.ReadyDoneAware:
```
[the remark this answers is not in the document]
Tonix517: good idea - done

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,262 @@
...
+	execDataService state_synchronization.ExecutionDataService,
+	metrics module.ExecutionMetrics) *BadgerRetryableUploader {
+
+	// check storage\eds interface params
+	if blocks == nil || commits == nil || transactionResults == nil ||
+		uploadStatusStore == nil || execDataService == nil {
+		panic("Not all storage parameters are valid")
+	}
+
+	// NOTE: only AsyncUploader is supported for now.
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/computation/computer/uploader/retryable_uploader_test.go
```diff
@@ -0,0 +1,207 @@
...
+		nil,
+		mockComputationResultStorage,
+		nil,
+		&metrics.NoopCollector{},
+	)
+	defer testRetryableUploader.Done()
+
+	// nil input - no call to Upload()
+	err := testRetryableUploader.Upload(nil)
+	assert.NotNil(t, err)
```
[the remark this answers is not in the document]
Tonix517: done

storage/badger/computation_result.go
Tonix517: i used to think about the same. the problem is on EDS, we are only able to use ExecutionDataID to retrieve stored data in which we can find BlockID, so I simply use ExecutionDataID(renamed).

storage/badger/computation_result.go
Tonix517: not in our use cases for now. but since it is in the storage implementation, i'm leaving `Remove` for completeness of storage functionalities, however I did add in `Upsert` which is used in this PR.

storage/badger/operation/computation_result.go
Tonix517: i grep our code base and only found that only the value part in the key-value pair is returned in `handle`, and the only place to get key is in `checkFunc`. But let me know if we can indeed get key from `handle`, it is indeed semantically better.

storage/badger/operation/computation_result.go
Tonix517: It simply returns all stored items, regardless of the actual values stored (true\false). Since it is in storage layer, I'd keep the implementation use case agnostic, leaving filtering to its callers.

engine/execution/ingestion/engine.go
Tonix517: the reason is all uploaders are being kept in `ComputationManager`. I will need to access all uploaders to make this call.

engine/execution/computation/computer/uploader/retryable_uploader.go
Tonix517: good point. I added an upsert implementation into storage layer so that we will not have remove-then-store issue here. However in our low-level Badger implementation, `update` code (https://github.com/onflow/flow-go/blob/master/storage/badger/operation/common.go#L88) will err out with non-existed keys, so in my storage layer code, I use both `retrieve` first to check existence of given key to implement `upsert`.

In terms of race condition, my question is how often\possible it is that the same block upload is being retried and computed (including original upload call)? if it can happen frequently, we may have to address that race-condition here.

storage/badger/operation/computation_result.go
m4ksio: ah yes, right!

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (b *BadgerRetryableUploader) Upload(computationResult *execution.ComputationResult) error {
+	if computationResult == nil || computationResult.ExecutableBlock == nil {
+		return errors.New("ComputationResult or its ExecutableBlock is nil when Upload() is called.")
+	}
+
+	// Before upload we store ComputationResult upload status to BadgerDB as false before upload is done.
+	// It will be marked as true when upload completes.
+	if err := b.uploadStatusStore.Upsert(computationResult.ExecutionDataID, false /*not completed*/); err != nil {
+		log.Warn().Msgf("failed to store ComputationResult into local DB with ID %s",
```
m4ksio, line 109: Shouldn't we bubble up this error instead of just logging?

storage/badger/computation_result.go
```diff
@@ -0,0 +1,59 @@
...
+	db *badger.DB
+}
+
+func NewComputationResultUploadStatus(db *badger.DB) *ComputationResultUploadStatus {
+	return &ComputationResultUploadStatus{
+		db: db,
+	}
+}
+
+func (c *ComputationResultUploadStatus) Upsert(executionDataId flow.Identifier,
```
m4ksio, line 23: I get that its easier to retry if we store executionDataID, but would be great to key it by blockID, so its more consistent with rest of a system.

storage/badger/computation_result.go
```diff
@@ -0,0 +1,59 @@
...
+	wasUploadCompleted bool) error {
+	_, err := c.ByID(executionDataId)
+	if errors.Is(err, storage.ErrNotFound) {
```
m4ksio, line 26: If we insist on having upset, we should create a function for this, along `insert` and `update`, since querying for value is not performant, and badger really works as upset by default.

engine/execution/ingestion/engine.go
m4ksio: Hmm, I don't think thats the best approach. ie. `ComputationManager` doesn't do retry, it can find (or not!) some uploaders and call retry.
Would be better to store a pointer to retriableUploader in the engine (or some other place?) and make a call explicitly, not involving Manager at all.

storage/badger/computation_result.go
Tonix517: done at PR #3062

storage/badger/computation_result.go
Tonix517: addressed. the data retrieval interface now parameterized with a target status to filter

engine/execution/computation/computer/uploader/retryable_uploader.go
Tonix517: the reason for this non-blocking logging is that even when there's sth going on with BadgerDB ops, upload can still be executed

engine/execution/computation/manager.go
```diff
@@ -358,3 +369,14 @@ func (e *Manager) GetAccount(address flow.Address, blockHeader *flow.Header, vie
 
 	return account, nil
 }
+
+func (e *Manager) RetryUpload() (err error) {
+	for _, u := range e.uploaders {
+		switch u.(type) {
+		case uploader.RetryableUploader:
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/computation/computer/uploader/retryable_uploader_test.go
```diff
@@ -0,0 +1,152 @@
...
+	mockBlocksStorage := new(storageMock.Blocks)
+	mockBlocksStorage.On("ByID", mock.Anything).Return(nil, nil)
+
+	mockCommitsStorage := new(storageMock.Commits)
+	mockCommitsStorage.On("ByBlockID", mock.Anything).Return(nil, nil)
+
+	mockTransactionResultsStorage := new(storageMock.TransactionResults)
+	mockTransactionResultsStorage.On("ByBlockID", mock.Anything).Return(nil, nil)
+
+	mockComputationResultStorage := new(storageMock.ComputationResultUploadStatus)
```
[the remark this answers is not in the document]
Tonix517: not checking number of times being called here for now, simply adding mocked implementation to make the mocked interface to return given objects.

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,245 @@
...
+	execDataService    state_synchronization.ExecutionDataService
+	unit               *engine.Unit
+	metrics            module.ExecutionMetrics
+	blocks             storage.Blocks
+	commits            storage.Commits
+	transactionResults storage.TransactionResults
+	uploadStatusStore  storage.ComputationResultUploadStatus
+}
+
+func NewBadgerRetryableUploader(
```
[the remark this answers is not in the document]
Tonix517: renamed

engine/execution/computation/computer/uploader/retryable_uploader.go
```diff
@@ -0,0 +1,245 @@
...
+
+			b.metrics.ExecutionComputationResultUploadRetried()
+		}
+	}
+
+	// return latest occurred error
+	return retErr
+}
+
+func (b *BadgerRetryableUploader) reconstructComputationResultFromEDS(
```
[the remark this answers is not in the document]
Tonix517: done

engine/execution/ingestion/engine.go
Tonix517: addressed. now keeping uploader references in engine. the `RetryUpload()` call now will not involve `ComputationManager`

storage/badger/computation_result.go
Tonix517: addressed - now this new status data is keyed by `blockID`.

admin/commands/uploader/toggle_uploader.go
```diff
@@ -16,6 +17,7 @@ type ToggleUploaderCommand struct{}
 func (t *ToggleUploaderCommand) Handler(ctx context.Context, req *admin.CommandRequest) (interface{}, error) {
 	enabled := req.ValidatorData.(bool)
 	computation.SetUploaderEnabled(enabled)
+	ingestion.SetUploaderEnabled(enabled)
```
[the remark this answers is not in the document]
Tonix517: addressed. removed the extra flag. now reusing this existing flag

storage/badger/operation/computation_result.go
Tonix517: ddressed. the data retrieval interface now parameterized with a target status to filter

m4ksio (approved):

Tonix517:
bors merge

PR #2744, opened 2022-07-04
change unittest fixture to use *flow.Header instead of flow.Header

pattyshack:
Make test code more consistent with regular code


SaveTheRbtz (approved):

pattyshack:
bors merge

PR #2745, opened 2022-07-04
remove disk size metrics

zhangchiqing:
I suspect this is related to the recently execution slowdown.


ramtinms (approved):
No harm doing it

PR #2746, opened 2022-07-05
remove disk size metrics

zhangchiqing:
This PR removes the disk size metrics which is a blocking call that might slow down the execution


SaveTheRbtz:
I have an alternative proposal https://github.com/onflow/flow-go/pull/2733 but I would vote for yours if we do not use these metrics.

ledger/complete/wal/wal.go
```diff
@@ -69,17 +69,6 @@ func (w *DiskWAL) RecordUpdate(update *ledger.TrieUpdate) error {
 		return fmt.Errorf("error while recording update in LedgerWAL: %w", err)
 	}
 
-	select {
-	case <-w.diskUpdateLimiter.C:
```
SaveTheRbtz, line 73: can you remove this field and a corresponding goroutine?

ledger/complete/wal/wal.go
```diff
@@ -69,17 +69,6 @@ func (w *DiskWAL) RecordUpdate(update *ledger.TrieUpdate) error {
...
-		diskSize, err := w.DiskSize()
-		if err != nil {
-			w.log.Warn().Err(err).Msg("error while checking forest disk size")
-		} else {
-			w.metrics.DiskSize(diskSize)
```
SaveTheRbtz, line 78: can you also remove this stat code?

ledger/complete/wal/wal.go
```diff
@@ -69,17 +69,6 @@ func (w *DiskWAL) RecordUpdate(update *ledger.TrieUpdate) error {
 		return fmt.Errorf("error while recording update in LedgerWAL: %w", err)
 	}
 
-	select {
-	case <-w.diskUpdateLimiter.C:
-		diskSize, err := w.DiskSize()
```
SaveTheRbtz, line 74: this function can also go away.

SaveTheRbtz (approved):

zhangchiqing:
>I have an alternative proposal https://github.com/onflow/flow-go/pull/2733 but I would vote for yours if we do not use these metrics. (less code is better.)

Oh nice. Didn't see that PR. But yeah, I'd like to remove them since I don't think we are using that metrics.

pattyshack (approved):

ledger/complete/wal/wal.go
```diff
@@ -37,14 +32,12 @@ func NewDiskWAL(logger zerolog.Logger, reg prometheus.Registerer, metrics module
 		return nil, err
 	}
 	return &DiskWAL{
```
fxamacker, line 34: I don't think `metrics` is used any more in `NewDiskWAL()`.  Maybe we can remove it, or replace it with `_`?

If we choose to remove `metrics`, maybe we can also remove `WALMetrics` because it's an empty interface and it isn't used anywhere else.

fxamacker (approved):
LGTM!  Just a minor comment about removing unused `metrics`.

ledger/complete/wal/wal.go
zhangchiqing: yeah, we could remove the metrics and the WALMetrics interface, but it breaks lots of tests. Fixing them will have to introduce lots of changes.

I'd like to keep the metrics as is, in case we need some metrics for WAL, then we don't have to add the metrics back and go through test cases fixing again.

zhangchiqing:
bors merge

PR #2748, opened 2022-07-05
[FVM] Benchstat CI improvements

janezpodhostnik:
1. Increased the number of bench runs to 10 (from 7) since other CI take longer so we have extra time to get some more accuracy on the benchmarks.
2. Disabled running regular tests during the benchmarks
3. There was a bug in one of the benchmarks causing it to error out
4. Added memstats to the benchmarking output

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -73,7 +73,7 @@ jobs:
 
             This branch with compared with the base branch ${{  github.event.pull_request.base.label }} commit ${{ github.event.pull_request.base.sha }}
 
-            The command `(for i in {1..N}; do go test ./fvm --bench . --tags relic -shuffle=on; done)` was used.
+            The command `(for i in {1..N}; do go test ./fvm --bench . --tags relic -shuffle=on --benchmem --run ^$; done)` was used.
 
             Bench tests were run a total of ${{ steps.settings.outputs.benchmark_repetitions }} times on each branch.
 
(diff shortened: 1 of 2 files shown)


.github/workflows/bench.yml
```diff
@@ -37,14 +37,14 @@ jobs:
 
       - name: Run benchmark on current branch
         run: |
-          (for i in {1..${{ steps.settings.outputs.benchmark_repetitions }}}; do go test ./fvm --bench . --tags relic -shuffle=on; done) | tee new.txt
+          (for i in {1..${{ steps.settings.outputs.benchmark_repetitions }}}; do go test ./fvm --bench . --tags relic -shuffle=on --benchmem --run ^$; done) | tee new.txt
```
janezpodhostnik, line 40: Bench memory and don't run tests (just benchmarks).

turbolent (approved):
Nice! 

Could you please also port these improvements to the Cadence repo?

SaveTheRbtz (approved):

.github/workflows/bench.yml
SaveTheRbtz: why not just `-count ${{ steps.settings.outputs.benchmark_repetitions }}`

SaveTheRbtz: (nit: you may also want to update the "The command `...` was used.")

janezpodhostnik: > why not just -count ${{ steps.settings.outputs.benchmark_repetitions }}?

Last time I was checking this, the problem was that using `count + shuffle` didn't do what I wanted. 

count 3 + shuffle of tests {A, B, C} does:
- BBB AAA CCC

I wanted something more spread out, so this does:
- BCA CBA ACB

Which I think is statistically better if the runner is noisy.

janezpodhostnik:
bors merge

PR #2749, opened 2022-07-05
tweak Programs locking + minor refactoring

pattyshack:
- reduce Get's critical section (don't hold lock when looking up parent's entry)
- guard HasChanges

cc: https://github.com/dapperlabs/flow-internal/issues/1722

### DIFF
--- a/fvm/programs/programs.go
+++ b/fvm/programs/programs.go
@@ -27,73 +27,57 @@ type ProgramEntry struct {
 	State    *state.State
 }
 
-type ProgramGetFunc func(location common.Location) (*ProgramEntry, bool)
-
-func emptyProgramGetFunc(_ common.Location) (*ProgramEntry, bool) {
-	return nil, false
-}
-
 // Programs is a cumulative cache-like storage for Programs helping speed up execution of Cadence
 // Programs don't evict elements at will, like a typical cache would, but it does it only
 // during a cleanup method, which must be called only when the Cadence execution has finished.
 // It it also fork-aware, support cheap creation of children capturing local changes.
 type Programs struct {
-	lock       sync.RWMutex
-	programs   map[common.LocationID]ProgramEntry
-	parentFunc ProgramGetFunc
-	cleaned    bool
+	lock     sync.RWMutex
+	programs map[common.LocationID]*ProgramEntry
+	parent   *Programs
+	cleaned  bool
 }
 
 func NewEmptyPrograms() *Programs {
-
 	return &Programs{
-		programs:   map[common.LocationID]ProgramEntry{},
-		parentFunc: emptyProgramGetFunc,
+		programs: map[common.LocationID]*ProgramEntry{},
 	}
 }
 
 func (p *Programs) ChildPrograms() *Programs {
 	return &Programs{
-		programs: map[common.LocationID]ProgramEntry{},
-		parentFunc: func(location common.Location) (*ProgramEntry, bool) {
-			return p.get(location)
-		},
+		programs: map[common.LocationID]*ProgramEntry{},
+		parent:   p,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


SaveTheRbtz (approved):

turbolent:
@pattyshack by creating PRs from branches in this repo, CI can run benchmarks

pattyshack:
@turbolent Is the standard procedure is to create branches directly in onflow/flow-go ?  I'm happy to comply if that's the standard, but this practice seems a bit surprising since it excludes external developers from contributing to the repo (or at least creates a barrier of entry).

Perhaps the right thing to do is to fix CI to run benchmarks regardless of which repo the PR originated from?

pattyshack:
@Kay-Zee looks like the ci benchstat lacks "pull-request"-write permission from forks due to GITHUB_TOKEN settings.  would it be possible to enable that by default (or is that too much of a security concern)?

fvm/programs/programs.go
```diff
@@ -27,73 +27,57 @@ type ProgramEntry struct {
...
-}
-
 // Programs is a cumulative cache-like storage for Programs helping speed up execution of Cadence
 // Programs don't evict elements at will, like a typical cache would, but it does it only
 // during a cleanup method, which must be called only when the Cadence execution has finished.
 // It it also fork-aware, support cheap creation of children capturing local changes.
 type Programs struct {
-	lock       sync.RWMutex
-	programs   map[common.LocationID]ProgramEntry
-	parentFunc ProgramGetFunc
```
turbolent, line 43: Was there a particular reason this was a function in the first place?

fvm/programs/programs.go
```diff
@@ -103,9 +87,23 @@ func (p *Programs) Set(location common.Location, program *interpreter.Program, s
 // HasChanges indicates if any changes has been introduced
 // essentially telling if this object is identical to its parent
 func (p *Programs) HasChanges() bool {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
 	return len(p.programs) > 0 || p.cleaned
 }
 
+func (p *Programs) unsafeForceCleanup() {
```
turbolent, line 96: Good idea to refactor this common code out, and nice job naming it well 👍

turbolent (approved):
LGTM! Maybe wait for a review from e.g. @ramtinms (@m4ksio is OOO)

janezpodhostnik (approved):
LGTM as well. Nice change!

pattyshack:
bors retry

fvm/programs/programs.go
ramtinms: I vaguely remember that there was a reason for this to be a func, but not in details, I'm okey we change it for now and when @m4ksio is back we would ask him.

ramtinms (approved):
Looks good to me

fvm/programs/programs.go
m4ksio: I think we wanted to prevent long chains of programs. Having a function gave us ability to copy entries from parents up, but I don't think we ever used it

fvm/programs/programs.go
```diff
@@ -27,73 +27,57 @@ type ProgramEntry struct {
...
-	return &programEntry, true
+	return p.programs[location.ID()], p.parent
 }
 
 func (p *Programs) Set(location common.Location, program *interpreter.Program, state *state.State) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
-	p.programs[location.ID()] = ProgramEntry{
+	p.programs[location.ID()] = &ProgramEntry{
```
m4ksio, line 80: Was there a reason to move to a pointer? Isn't this increasing allocations count, instead of keeping data inside a map?

PR #2750, opened 2022-07-05
[FVM] Debug specific Transaction

janezpodhostnik:


janezpodhostnik:
The code here is still a very good example, but there is no point in keeping a draft PR open.

Issue #2669, opened 2022-06-21
Out-of-memory failure when re-executing large number of blocks on EN nodes


Tonix517:
mentioned in #2273. It can be tracked in a different issue. Quoting @j1010001 

> Update after chat with Maks:
When EN restarts it tries to catch up and execute blocks. It loads them into a memory and then start executing
but if there are a lots of blocks, they won't fit into the memory. Ideally, we would have some form of a buffer or list - load blocks from one end, and execute from the other, so it makes progress.
In a recent issue on testnet where we had to re-execute blocks we patched EN to load 100k blocks, execute and crash. We kept restarting the EN until it executed and uploaded all blocks.

Tonix517:
According to current behavior of EN ingestion (https://github.com/onflow/flow-go/blob/master/engine/execution/ingestion/engine.go#L366), all unexecuted blocks will be directly queued at once, but one block will only be removed from queue when executed, which happens asynchronously, and may take some time considering the overhead of getting all collection and then execution. So block removal happens much slower, during which process all loaded collections may take a lot of memory.

One idea is to limit the number of blocks in worker queue. We only add in new unexecuted blocks when previous batch of blocks all executed. Also talking about observability, it may help if we can emit 1) # of active unexecuted blocks currently being processed; 2) # of currently active collection request; etc., and put these metrics on Grafana, which reflects more internal runtime mechanism of EN, may help us debug quicker in future.

PR #2756, opened 2022-07-07
[Execution] Putting limit to number of blocks being reloaded at EN Ready()

Tonix517:
#2669 

#### Summary
As mentioned #2669, when there are large number of unexecuted blocks to reload at EN node bootup, it can bloat the memory. The solution this PR introduces is to limit the number of blocks currently being processed in `reloadUnexecutedBlocks()` call, so as to correspondingly limit the memory consumption at that catching up step.

#### Note
This is mostly to limit the currently being process blocks at reloadUnexecutedBlocks() which load blocks in sequential orders, so limiting the number of blocks in work queue should effectively limit the memory usage too. However handleBlock() is called asynchronously this won't help limit the memory usage after we've done with reloadUnexecutedBlocks() and start the normal block processing use cases, because new blocks will keep staying in memory before they can be processed.

#### Test Plan
- [X] new unit test cases

- [X] load test with work queue size limit as **1_000_000** and monitor localnet General dashboards to make sure everything works: `go run --tags relic ../loader -log-level info -tps 70 -tps-durations 120s`
rate_limit_1m_size_#_in_q_20m

- [X] with work queue size limit as **10** and monitor localnet General dashboards to make sure everything works: `go run --tags relic ../loader -log-level info -tps 70 -tps-durations 120s`
rate_limit_10_size_#_in_q_20m


engine/execution/ingestion/engine.go
```diff
@@ -1220,6 +1240,37 @@ func (e *Engine) logExecutableBlock(eb *entity.ExecutableBlock) {
...
+}
+
+func (e *Engine) blockingMarkOneBlockAddedToWorkQueue() {
+	if e.workQueueLimitCh != nil {
+		// NOTE: since this could be blocking when chan is full, this function should
+		//		 not be called inside mempool.Run() which is globally locked itself.
+		e.workQueueLimitCh <- struct{}{}
+		e.metrics.ExecutionBlockAddedToWorkQueue()
+	}
+}
```
ramtinms, line 1264: is it possible that this causes a deadlock, all the blocks in the queue not be executable since they are all missing a parent block?

engine/execution/ingestion/engine.go
```diff
@@ -35,6 +35,14 @@ import (
 	"github.com/onflow/flow-go/utils/logging"
 )
 
+// This is mostly to limit the currently being process blocks at reloadUnexecutedBlocks()
+// which load blocks in sequential orders, so limiting the number of blocks in work queue
+// should effectively limit the memory usage too. However handleBlock() is called asynchronously
+// this won't help limit the memory usage after we've done with reloadUnexecutedBlocks() and start
+// the normal block processing use cases, because new blocks will keep staying in memory before
+// they can be processed.
+const DefaultMaxWorkQueueSize = 1_000_000
```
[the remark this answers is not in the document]
ramtinms: maybe we could have a flag as well to adjust the default value when needed?

ramtinms (approved):
looks good to me, just some minor suggestions and questions.

engine/execution/ingestion/engine.go
Tonix517: Good catch @ramtinms. Logically yes it is possible. When that happens, what could be the largest number of blocks being stuck in the waiting to be executed status? The limit size here is supposed to be very large though, but if a missing parent block can exist in arbitrary height, we will have to address this concern.

Also when we reload blocks, could that happen that certain parent block is missing? If not, we can be conservative that this limit will only be applied at block reloading

janezpodhostnik (commented):
This looks good to me, but I am not familiar with this part of the code enough.

Tonix517:
Found this code piece at https://gist.github.com/cyfdecyf/4562635 which resembles our use case of quite well, which is also a typical producer-consumer scenario. On my laptop I got:

> BenchmarkChan-4                6         195208715 ns/op
BenchmarkLock-4                2         751450229 ns/op

The buffered channel runs much faster than mutex lock\unlock logic. Also, the code using chan is much neater than the mutex based code with less state to manually maintain. @m4ksio

engine/execution/ingestion/engine.go
```diff
@@ -66,6 +74,8 @@ type Engine struct {
 	syncFast               bool                // sync fast allows execution node to skip fetching collection during state syncing, and rely on state syncing to catch up
 	checkAuthorizedAtBlock func(blockID flow.Identifier) (bool, error)
 	pauseExecution         bool
+	workQueueLimitCh       chan struct{} // a buffered channel of struct{} to guard max number of blocks currently being processed
+	workQueueLimitSize     uint32
```
SaveTheRbtz, line 78: nit: you may consider factoring this out into something more reusable like https://github.com/remeh/sizedwaitgroup (in case you feel bounded concurrency can be a useful pattern in the codebase.)

Tonix517: Nice idea. I will explore in that direction

PR #23, opened 2020-09-28
Yahya/4818 fix concurrent map access

yhassanzadeh13:
This PR addresses #4818: 

The `IdentifierMap` is a mempool that keeps data of type `IdMapEntity`:
```
type IdMapEntity struct {
	Key flow.Identifier
	IDs map[flow.Identifier]struct{}
}
```

Based on Golang convention of maps, every time `IdentifierMap.Get` is invoked and an instance of `IdMapEntity` is returned,  the `IdMapEntity.IDs` map part of the return object is returned by reference. Since it is returned to a non-critical section of the code, its access is not controlled by any lock. As part of the `IdentifierMap.Get` procedure, a copy of `IdMapEntity.IDs` is returned as a slice. Processing that the (by-referenced) map into slice is happening out of any concurrency control mechanism, which results in a `panic` on concurrent read-write if another goroutine is updating the same `IdMapEntity.IDs` _in the mempool_: 

```
fatal error: concurrent map iteration and map write
goroutine 1313005 [running]:
runtime.throw(0x145d835, 0x26)
/usr/local/go/src/runtime/panic.go:774 +0x72 fp=0xc00092b6d8 sp=0xc00092b6a8 pc=0x431682
runtime.mapiternext(0xc00092b830)
/usr/local/go/src/runtime/map.go:858 +0x579 fp=0xc00092b760 sp=0xc00092b6d8 pc=0x4115b9
github.com/dapperlabs/flow-go/module/mempool/stdmap.(*IdentifierMap).Get(0xc00013a7d8, 0x8e2a157eac22ee33, 0x8397b006c982dde1, 0xda1778635350785f, 0x416303044b7255c6, 0xc01d32aa01, 0xc9, 0x1f4, 0xc000070930)
```

To address issue, in this PR, we re-write the `IdentifierMap.Get` using the concurrency-safe method of `Backend.Run`.

PR #2757, opened 2022-07-07
Fixed docker-build-loader docker image build failure

Tonix517:
`make docker-build-loader` failed with the error below. Removed the hard-coded `GOOS` and `GOARCH` env var override from Dockerfile, then it builds again.

> [build-production 2/4] RUN --mount=type=cache,sharing=locked,target=/go/pkg/mod     --mount=type=cache,target=/root/.cache/go-build     --mount=type=ssh     cd integration &&     GO111MODULE=on CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build --tags relic -ldflags "-extldflags -static" -o ./app ././loader:
#23 0.465 # runtime/cgo
#23 0.465 gcc: error: unrecognized command line option '-m64'

### DIFF
--- a/integration/loader/Dockerfile
+++ b/integration/loader/Dockerfile
@@ -48,7 +48,7 @@ RUN --mount=type=cache,sharing=locked,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=ssh \
     cd integration && \
-    GO111MODULE=on CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build --tags relic -ldflags "-extldflags -static" -o ./app ./${TARGET}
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
You may consider making it customizable (to be consistent with the rest of the containers in the Makefile) in the file by:
* using `--build-arg GOARCH=$(GOARCH)` in the Makefile
* `ARG GOARCH=amd64`  in the dockerfile.

SaveTheRbtz (approved):

peterargue:
This was a hold over from when we added arm64 support a few months ago. I forgot to update the loader. Thanks for the fix!
https://github.com/onflow/flow-go/commit/c58ee3d43542d3711b67c605432f93411dc17fba#diff-c06f05e38e556217d7bdd4d1d715fad986024f661bffda19d750330660f469a5R64

PR #2760, opened 2022-07-07
[AN] - Trigger Cache Eviction on Overwrite

lolpuddle:
There currently exists an edge case where we are replacing connections in the connection pool without closing them properly, causing a memory leak. The closing logic is currently triggered on eviction from the connection pool.

Example:
If a cached connection we retrieve for `example_grpc_address` changes state and is [no longer in the ](https://github.com/onflow/flow-go/blob/3b7c152d081b61106667c46ff6b03eb625a4ac9f/engine/access/rpc/backend/connection_factory.go#L83)`Ready` state. This would cause us to enter the logic to replace the current value/connection for key `example_grpc_address` but not trigger the eviction logic since updating a value isn't a key eviction.

This PR ensures that we remove, and trigger the eviction logic to close connections before adding new connections to the pool if the connection is from the same grpc address


peterargue (approved):
please add a comment noting why that line is important

lolpuddle:
bors merge

vishalchangrani (approved):

PR #2761, opened 2022-07-08
Update static epoch test to use same testing flow

jordanschalm:
Simplify the static epoch transition test, use the same assertions and testing flow as the dynamic committee tests.


jordanschalm:
https://github.com/onflow/flow-go/pull/2786 should address the edge case requiring the 10s sleep at the beginning of this test

integration/tests/epochs/epoch_static_transition_test.go
```diff
@@ -33,107 +31,43 @@ func (s *StaticEpochTransitionSuite) SetupTest() {
...
-		}
-
-		for _, v := range phaseChecks {
-			item := lookup[v.finalView]
-			assert.Equal(s.T(), v.epoch, item.epochCounter, "wrong epoch at view %d", v.finalView)
-			assert.Equal(s.T(), v.phase, item.phase, "wrong phase at view %d", v.finalView)
-		}
-	}
+	// TODO investigate sealing segment error associated with retrieving snapshot near genesis
+	time.Sleep(10 * time.Second)
```
gomisha, line 42: why is so much sleep necessary?

gomisha (approved):

integration/tests/epochs/epoch_static_transition_test.go
jordanschalm: It's probably more than is necessary - I just put a relatively large value to avoid hitting this error condition while I fix it -> the fix is in https://github.com/onflow/flow-go/pull/2786

asdf
```diff
@@ -0,0 +1,104 @@
+
```
pattyshack, line 1: included by accident?

jordanschalm: Creating dummy commits by appending to this file, to trigger CI.

PR #2763, opened 2022-07-08
[Execution] check if state commitment exists before executing scripts

ramtinms:
An `state commitment` might exist inside the protocol state but missing from the execution state (trie being purged for older blocks). We receive a lot of scripts targeting old blocks on execution nodes and most of them fail inside the FVM when trying to load the very first register. This PR would add a check before executing scripts, returning early in these cases and preventing memory allocation, it also helps with interpretability of the error message, considering previously user would see an error about a random register (used to be the frozen account register, now its memory metering weights registers).


zhangchiqing (approved):
Nice!

SaveTheRbtz (approved):

PR #2764, opened 2022-07-08
chore(loader): switch from Client to Access API

SaveTheRbtz:
Client API is deprecated

### DIFF
--- a/integration/utils/contLoadGenerator.go
+++ b/integration/utils/contLoadGenerator.go
@@ -13,7 +13,7 @@ import (
 	"github.com/onflow/flow-go/module/metrics"
 
 	flowsdk "github.com/onflow/flow-go-sdk"
-	"github.com/onflow/flow-go-sdk/client"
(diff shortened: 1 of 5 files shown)


AndrewM-SDET (approved):

integration/utils/contLoadGenerator.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/onflow/flow-go/module/metrics"
 
 	flowsdk "github.com/onflow/flow-go-sdk"
-	"github.com/onflow/flow-go-sdk/client"
+ "github.com/onflow/flow-go-sdk/access"
```
pattyshack, line 16: run gofmt ...

pattyshack (approved):

integration/utils/contLoadGenerator.go
SaveTheRbtz: interesting... I have it automatic on save, looks like an IDE failure, should've noticed it.

PR #2766, opened 2022-07-08
chore(loader): fix Stop/Follow race

SaveTheRbtz:
This should prevent any races between Stop and Follow calls. Code now is slower but at least correct.

While here, address comments from #2632.


integration/utils/follower.go
```diff
@@ -55,13 +54,10 @@ type txFollowerImpl struct {
...
 
-	inprogress *atomic.Int64
-
-	mu      *sync.RWMutex
-	height  uint64
-	blockID flowsdk.Identifier
-
-	txToChan sync.Map
+	mu       *sync.RWMutex
+	height   uint64
```
pattyshack, line 58: nit: could you add `// guarded by mu` to each applicable fields in case the fields got moved around / newline added / etc.

integration/utils/follower.go
```diff
@@ -82,11 +78,11 @@ func NewTxFollower(ctx context.Context, client *client.Client, opts ...followerO
...
 		cancel: cancel,
 		logger: zerolog.Nop(),
-		mu:     &sync.RWMutex{},
 
-		inprogress: atomic.NewInt64(0),
+		mu:       &sync.RWMutex{},
+		txToChan: make(map[flowsdk.Identifier]txInfo),
 
-		stopped:  make(chan struct{}),
+		stopped:  make(chan struct{}, 1),
```
pattyshack, line 85: no need?  you never insert into the channel

pattyshack: o, could you remove `if f.height == 0 {` check below?  you've just initialized the follower, the height is always zero

integration/utils/follower.go
```diff
@@ -139,16 +135,21 @@ Loop:
 			}
 			for _, tx := range col.TransactionIDs {
 				blockTxs++
-				if ch, loaded := f.txToChan.LoadAndDelete(tx.Hex()); loaded {
-					txi := ch.(txInfo)
 
+				f.mu.Lock()
```
pattyshack, line 139: could you refactor the this into a method using the lock / defer unlock pattern

it's easy for future dev to shoot themselves in the foot with inline locking.

integration/utils/follower.go
```diff
@@ -158,6 +159,10 @@ Loop:
 			}
 		}
 
+		f.mu.RLock()
```
pattyshack, line 162: same here

integration/utils/follower.go
```diff
@@ -211,21 +231,17 @@ func (f *txFollowerImpl) Stop() {
...
-		},
-	)
-	for _, val := range toDelete {
-		f.txToChan.Delete(val)
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	for k, v := range f.txToChan {
+		close(v.C)
+		delete(f.txToChan, k)
```
pattyshack, line 239: don't delete in loop.  just initialize a new txToChan map after the loop?

integration/utils/follower.go
SaveTheRbtz:
```
	for _, opt := range opts {
		opt(f)
	}
```

modifies f inline

pattyshack: ah.  my bad

pattyshack (approved):

integration/utils/follower_test.go
```diff
@@ -28,6 +28,24 @@ func TestTxFollower(t *testing.T) {
 	f.Stop()
 }
 
+// TestTxFollowerFollowAfterStop creates new follower with a fixed block height and stops it.
+func TestTxFollowerFollowAfterStop(t *testing.T) {
+	// TODO(rbtz): test against a mock client, but for now we just expire
+	// the context so that the followere wont be able to progress.
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
```
Kay-Zee, line 36: this cancels the context right away, which i assume is on purpose, does moving this to line 44 just before the stop make the test more representative? or not really?

Kay-Zee (approved):

integration/utils/follower_test.go
SaveTheRbtz: sadly for now there is no mock available for the access.Client interface.  It is in my TODO to generate one.  Once it is available this hack can be removed.

SaveTheRbtz:
bors merge

PR #2771, opened 2022-07-08
Leon/add resolver channel

LLsq2:
Adds a test for overloading resolver channel worker and queue limit to ensure resolver does not fail.


yhassanzadeh13 (changes requested):

network/p2p/dns/resolver.go
```diff
@@ -76,32 +76,44 @@ const (
...
 
 	cm := component.NewComponentManagerBuilder()
 
-	for i := 0; i < numIPAddrLookupWorkers; i++ {
+	for i := 0; i < ipWorker; i++ {
 		cm.AddWorker(resolver.processIPAddrLookups)
 	}
 
-	for i := 0; i < numTxtLookupWorkers; i++ {
+	for i := 0; i < txtWorker; i++ {
```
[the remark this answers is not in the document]
LLsq2: Fixed this with original names

network/p2p/dns/resolver.go
```diff
@@ -76,32 +76,44 @@ const (
...
+
+func NewTestResolver(logger zerolog.Logger, collector module.ResolverMetrics, dnsCache mempool.DNSCache, opts ...optFunc) *Resolver {
+	resolver := newResolver(logger, collector, dnsCache, 1, 1, 2, 2)
+	for _, opt := range opts {
+		opt(resolver)
+	}
+	return resolver
+}
+
+func newResolver(logger zerolog.Logger, collector module.ResolverMetrics, dnsCache mempool.DNSCache, ipWorker int, txtWorker int, ipQueue int, txtQueue int) *Resolver {
```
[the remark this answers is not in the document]
LLsq2: Fixed

network/p2p/dns/resolver.go
```diff
@@ -67,6 +65,26 @@ func WithTTL(ttl time.Duration) optFunc {
 	}
 }
 
+// WithResolverSize is an option function for setting the number of resolver workers and queue size.
+func WithResolverSize(numIPAddrLookupWorkers int, numTxtLookupWorkers int, ipAddrLookupQueueSize int, txtLookupQueueSize int) optFunc {
```
yhassanzadeh13, line 69: Please have a separate method per each of these parameters, i.e., `WithIpWorkersCount(count int)`, `WithTxtWorkersCount(count int)`, `WithIpQueueSize(size int)` and `WithTxtQueueSize(size int)`. In general, we should not have high coupling in our option functions. Thanks.

network/p2p/dns/resolver.go
```diff
@@ -67,6 +65,26 @@ func WithTTL(ttl time.Duration) optFunc {
...
+			cm.AddWorker(resolver.processIPAddrLookups)
+		}
+
+		for i := 0; i < numTxtLookupWorkers; i++ {
+			cm.AddWorker(resolver.processTxtLookups)
+		}
+
+		resolver.cm = cm.Build()
+		resolver.Component = resolver.cm
+	}
```
yhassanzadeh13, line 85: Please note that an option function should not have any redundant logic other than just setting the member fields. These lines are a repetition of lines 109-115 which makes the code less DRY (Don't Repeat Yourself principle). We can fix this by moving lines 109-115 post applying the option function, so that we make sure that it adds workers with the new parameters. So the `NewResolver` would look like this: 

```
go

    for _, opt := range opts {
		opt(resolver)
	}

	for i := 0; i < numIPAddrLookupWorkers; i++ {
		cm.AddWorker(resolver.processIPAddrLookups)
	}

	for i := 0; i < numTxtLookupWorkers; i++ {
		cm.AddWorker(resolver.processTxtLookups)
	}

	resolver.cm = cm.Build()
	resolver.Component = resolver.cm
```
Moreover, there is a flaw with the current shape of the code, regardless of the values we pass for the number of workers in the option functions, it always adds workers with the default values in addition to the option function. Refactoring the code with the approach above will fix it. Thanks.

network/p2p/dns/resolver_test.go
```diff
@@ -201,6 +202,59 @@ func TestResolver_Expired_Invalidated(t *testing.T) {
...
+
+	time.Sleep(2 * time.Second) // waits enough for cache to get invalidated
+
+	queryWG = syncThenAsyncQuery(t, times, resolver, txtTestCases, ipTestCases, happyPath)
+
+	unittest.RequireReturnsBefore(t, queryWG.Wait, 1*time.Second, "could not perform all queries on time")
+	//Verifying
+	cancel()
+	unittest.RequireCloseBefore(t, resolver.Done(), 100*time.Millisecond, "could not stop dns resolver on time")
+	log.Printf("ipQueryCount: %d txtQueryCount: %d", ipQueryCount, txtQueryCount)
```
yhassanzadeh13, line 250: Please use `unittest.Logger()` for logging in the tests, it is more consistent with our codebase, also we can enable and disable it by adding or removing `-vv` flag, respectively. Thanks.

network/p2p/dns/resolver_test.go
```diff
@@ -201,6 +202,59 @@ func TestResolver_Expired_Invalidated(t *testing.T) {
...
+	resolver.Start(ctx)
+	unittest.RequireCloseBefore(t, resolver.Ready(), 100*time.Millisecond, "could not start dns resolver on time")
+
+	size := 10 // 10 text and 10 ip domains.
+	times := 1 // each domain is queried 5 times.
+	txtTestCases := testnetwork.TxtLookupFixture(size)
+	ipTestCases := testnetwork.IpLookupFixture(size)
+
+	// each domain is resolved only once through the underlying resolver, and then is cached for subsequent times.
+	mockBasicResolverForDomains(t, &basicResolver, ipTestCases, txtTestCases, happyPath, 2, LogResolveQueryCounts(&ipQueryCount, &txtQueryCount))
```
yhassanzadeh13, line 238: While this approach looks correct with passing the counters by parameters, I still recommend evaluating the correctness of the test through mocking the metrics and then asserting the expected calls have been done. 

Hence, please take the approach suggested in my previous comment: https://github.com/onflow/flow-go/pull/2771#discussion_r918085619

Compared to the current state of implementation, the suggested metrics-based approach adds the following values: 
1. It makes the test simpler, consistent with the existing ones and more readable, maintainable, and flexible. 
2. It does not need adding two more variables as well as dealing with their pointer-based lifecycle. 
3. It does not need to add more complexities to the test helpers, i.e., `mockBasicResolverForDomain`. 
4. It also tests the correctness of our metrics collection procedure for the cache.

LLsq2:
Closing for now and will restart when priority arises again

PR #2775, opened 2022-07-08
Fix Unit race condition on stopping.

pattyshack:
Race scenario:

goroutine 1 (calls unit.Launch)
    checked the context has not cancelled

context switch

goroutine 2 (calls unit.Done)
    cancels the context
    finished waiting for the wait group
    (unit is "Done" at this point)

context switch

goroutine 1 (remaining half of the original Launch call)
    add to wait group

### DIFF
--- a/engine/unit.go
+++ b/engine/unit.go
@@ -10,6 +10,8 @@ import (
 
 // Unit handles synchronization management, startup, and shutdown for engines.
 type Unit struct {
+	admitLock sync.Mutex // used for synchronizing context cancellation with work admittance
+
 	wg         sync.WaitGroup     // tracks in-progress functions
 	ctx        context.Context    // context that is cancelled when the unit is Done
 	cancel     context.CancelFunc // cancels the context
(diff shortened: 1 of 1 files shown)


engine/unit.go
```diff
@@ -27,29 +29,46 @@ func NewUnit() *Unit {
...
+		return false
+	default:
+	}
+
+	u.wg.Add(1)
+	return true
+}
+
+func (u *Unit) stopAdmitting() {
+	u.admitLock.Lock()
```
zhangchiqing, line 47: Does this change fix any existing test?
If not, could you add one that would fail without the synchronization here?

pattyshack: this does not fix any existing test and the race detector won't be able to detect this race since all variables involved are thread safe.

Adding a test for this makes the code look pretty unnatural.  We would need to artificially / deterministically inject a context switch inside `admit()` and check in test using sleep / timeouts.  I'm not sure if it's worth it

zhangchiqing (approved):

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors merge

PR #2777, opened 2022-07-08
Isolate TestEpochJoinAndLeaveVN

AndrewM-SDET:
Set up an isolated make target to run TestEpochJoinAndLeaveVN verbosely on its own. Update bft-tests.yml to check github actions and verify test in CI


AndrewM-SDET:
I was misunderstanding how github actions are set up, @gomisha has clarified, closing this PR.

PR #2778, opened 2022-07-08
chore(loader): wait for account creation results

SaveTheRbtz:
Flow API does not guarantee that `GetTransactionResult` is synchronized with the `GetBlockByHeight` results, hence we can get into a situation where after Follower reports successful execution of the transaction a subsequent `GetTransactionResult` still reports it being only in a `Finalized` state.  Hence, added `GetTransactionResult` that can gently spin on txID waiting for it to get into a terminal state.

### DIFF
--- a/integration/utils/tx_result.go
+++ b/integration/utils/tx_result.go
@@ -12,7 +12,7 @@ import (
 )
 
 // GetTransactionResult waits for the transaction to get into the terminal state and returns the result.
-func GetTransactionResult(ctx context.Context, client access.Client, txID flowsdk.Identifier) *flowsdk.TransactionResult {
+func GetTransactionResult(ctx context.Context, client access.Client, txID flowsdk.Identifier) (*flowsdk.TransactionResult, error) {
 	var b retry.Backoff
 	b = retry.NewFibonacci(100 * time.Millisecond)
 	b = retry.WithMaxDuration(60*time.Second, b)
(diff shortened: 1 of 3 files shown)


integration/utils/tx_result.go
```diff
@@ -0,0 +1,44 @@
...
+		case flowsdk.TransactionStatusExecuted, flowsdk.TransactionStatusSealed:
+			return nil
+		case flowsdk.TransactionStatusPending, flowsdk.TransactionStatusFinalized:
+			return retry.RetryableError(fmt.Errorf("waiting for transaction execution: %s", txID))
+		default:
+			return fmt.Errorf("unexpected transaction status: %s", result.Status)
+		}
+	})
+	if err != nil {
+		return &flowsdk.TransactionResult{Error: err}
```
pattyshack, line 41: just return `(*TransactionResult, error)` tuple instead of wrapping the error?

pattyshack (approved):

integration/utils/tx_result.go
SaveTheRbtz: Sadly the `TransactionResult` does have an error field, so downstream code usually checks it anyway.  That said, let me try duplicating the error in return statement -- this would likely be cleaner.

AndrewM-SDET (approved):

SaveTheRbtz:
bors merge

PR #2779, opened 2022-07-09
AN - Lock Updates to Conn Pool

lolpuddle:
Related to https://github.com/onflow/flow-go/pull/2760. There still exists a condition where an update can be made to the pool without properly expiring the previous key


engine/access/rpc/backend/connection_factory.go
```diff
@@ -81,14 +83,16 @@ func (cf *ConnectionFactoryImpl) retrieveConnection(grpcAddress string, timeout
...
 	}
 	if conn == nil || conn.GetState() != connectivity.Ready {
-		// This line ensures that when a connection is renewed, the previously cached connection is evicted and closed
-		cf.ConnectionsCache.Remove(grpcAddress)
 		var err error
 		conn, err = cf.createConnection(grpcAddress, timeout)
 		if err != nil {
 			return nil, err
 		}
+		cf.lock.Lock()
```
vishalchangrani, line 91: isn't the connection cache concurrency safe? Would this lock slow down ALL connection request?

lolpuddle: It's true that the connection cache is concurrency safe but that's only true for each independent operation. Our concern was that there may exist a race condition in updating the cache where it might look something like this:

#1 - cf.ConnectionsCache.Remove(grpcAddress)
#2 - cf.ConnectionsCache.Remove(grpcAddress)
#1 - cf.ConnectionsCache.Add(grpcAddress, conn)
#2 - cf.ConnectionsCache.Add(grpcAddress, conn)

peterargue: there's still a race condition here, except it will result in connection thrashing instead of connection leaking. 

1. Thread A get's the connection
2. Thread B get's the connection
3. Thread A determines the connection isn't good anymore and takes the lock
4. Thread B determines the connection isn't good anymore and waits for the lock
5. Thread A removes the old connection, adds a new one, then releases the lock
6. Thread B gets the lock, then removes the new connection and adds a different new one

In this case, Thread A's connection would be closed prematurely. I'm not 100% what happens here, but I'm guessing the connection gets reset and aborted early. We should find out.

To fix that we'll need to move the locking to before the initial get. As Vishal noted this will likely cause lock contention, so we may need to reduce the scope of the lock to a single address.

Kay-Zee: > To fix that we'll need to move the locking to before the initial get. 

Or, we think it's not very common for the connection to be "not good", we could add another `Get` inside the lock, so just below this line, and do another check. If it's good, use, if it's not, then remove and add the good connection.

It just depends how often connections need to get re-cycled.

engine/access/rpc/backend/connection_factory.go
```diff
@@ -74,24 +88,63 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
 
 func (cf *ConnectionFactoryImpl) retrieveConnection(grpcAddress string, timeout time.Duration) (*grpc.ClientConn, error) {
 	var conn *grpc.ClientConn
+	var mutex *sync.Mutex
```
peterargue, line 91: I think we could simplify this locking logic. there seem to be 2 main options:

1. Optimistic locking (Kan's suggestion) - optimize for infrequent client creation/refreshes
We use a single global lock, but only in the case where we need to add a new client. When adding a new client, take the lock, and check again if a client exists in the cache. If it does, return otherwise add it. When refreshing clients, we'll need to take the global lock. 

2. Per client locking - optimize for more frequent client creation/refreshes
We still need a global lock to ensure the get/add operations are atomic, but then switch to a per client lock. Essentially the flow would be:
* use `PeekOrAdd` to get the client object (or add a new one if it doesn't exist)
* take the client specific lock
* use `Get` to get the client again (this makes sure we update the access metadata)
* check if the connection is healthy. if it is return.
* otherwise setup the connection
* release the client lock

Looking over the grpc data on testnet/mainnet, the vast majority of api calls are for endpoints that use execution nodes, and there are a small number of nodes so they would always be in the cache. Send transaction is a relatively infrequent request, so the collection nodes would be the ones getting evicted.

If we increased the size of the cache so there were less frequent LN evictions, and found a fix for the frequent client invalidations, then I think 1 would be fine.

engine/access/rpc/backend/connection_factory.go
```diff
@@ -112,12 +165,17 @@ func (cf *ConnectionFactoryImpl) GetAccessAPIClient(address string) (access.Acce
...
+func (cf *ConnectionFactoryImpl) RefreshAccessAPIClient(address string) {
+	grpcAddress, _ := getGRPCAddress(address, cf.CollectionGRPCPort)
+	if res, ok := cf.ConnectionsCache.Get(grpcAddress); ok {
+		store := res.(ConnectionCacheStore)
+		store.mutex.Lock()
+		cf.lock.Lock()
+		defer store.mutex.Unlock()
+		defer cf.lock.Unlock()
+		store.ClientConn.Close()
+		_, _ = cf.addConnection(grpcAddress, cf.CollectionNodeGRPCTimeout, store.mutex)
```
peterargue, line 177: instead of adding it here, could we just close the connection and set it to nil? then the next time it's retrieved, we add the connection. This reduces the time spent with the lock
```
store := res.(ConnectionCacheStore)

store.mutex.Lock()

if store.conn != nil {
	conn := store.ClientConn
	store.ClientConn = nil
	defer conn.Close()
}

store.mutex.Unlock()
```

engine/access/rpc/backend/connection_factory.go
```diff
@@ -74,24 +88,63 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
...
-		// This line ensures that when a connection is renewed, the previously cached connection is evicted and closed
-		cf.ConnectionsCache.Remove(grpcAddress)
+		// this lock prevents a memory leak where a race condition may occur if 2 requests to a new connection at the
+		// same address occur. the second add would overwrite the first without closing the connection
+		cf.lock.Lock()
+		defer cf.lock.Unlock()
+		// Check if connection was created/refreshed by another thread
+		if res, ok := cf.ConnectionsCache.Get(grpcAddress); ok {
+			conn = res.(ConnectionCacheStore).ClientConn
+			if conn.GetState() == connectivity.Ready {
```
peterargue, line 113: are you sure `Ready` is the right measure here? it looks like anything but `Shutdown` could be ok

engine/access/rpc/backend/connection_factory.go
```diff
@@ -74,24 +89,53 @@ func (cf *ConnectionFactoryImpl) createConnection(address string, timeout time.D
...
+func (cf *ConnectionFactoryImpl) addConnection(grpcAddress string, timeout time.Duration, store *ConnectionCacheStore) (*grpc.ClientConn, error) {
+	conn, err := cf.createConnection(grpcAddress, timeout)
+	if err != nil {
+		return nil, err
+	}
+
+	store.ClientConn = conn
+
+	if cf.AccessMetrics != nil {
+		cf.AccessMetrics.TotalConnectionsInPool(uint(cf.ConnectionsCache.Len()), cf.CacheSize)
```
peterargue, line 138: since we're in here anyway. any other metrics we could add? Maybe a counter for new connections. it probably makes sense to add a debug level log for that as well as for evictions.

peterargue (commented):
any way you can think of to test the locking/concurrency?

engine/access/rpc/backend/connection_factory.go
```diff
@@ -136,12 +176,22 @@ func (cf *ConnectionFactoryImpl) GetExecutionAPIClient(address string) (executio
...
+}
+
+func (cf *ConnectionFactoryImpl) invalidateAPIClient(address string, port uint) {
+	grpcAddress, _ := getGRPCAddress(address, port)
+	if res, ok := cf.ConnectionsCache.Get(grpcAddress); ok {
+		store := res.(*ConnectionCacheStore)
+		store.mutex.Lock()
+		if store.ClientConn != nil {
+			conn := store.ClientConn
+			conn.Close()
```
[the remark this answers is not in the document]
lolpuddle: after looking at it some more, i don't particularly love the logic around how the invalidating of connections is handled. this is currently still a potential pain point where we're closing a connection without regard for whether there exists any threads that have acquired a client, and are still waiting to send a request out.

it would make more sense to add the logic to delay closing by the timeout time before actually attempting to close the connection

peterargue (approved):

vishalchangrani (approved):

PR #2780, opened 2022-07-11
[BFT Testing] Fixes wintermute orchestrator race condition

yhassanzadeh13:
This PR resolves a race condition in the Wintermute attack orchestrator that would cause a flaky behavior on the test [`TestTwoConcurrentExecutionReceipts_DistinctResult`](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator_test.go#L66). The root cause of the race condition is lack of isolation support for the [attack state](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator.go#L203-L206) updates considering that the Wintermute orchestrator is supposed to conduct only a single result corruption. So it is critical to make sure that at any time, there is at most one execution result corruption is ongoing, and out of all ongoing attempts on result corruption, only one is getting successful. 

Lack of isolation on the [attack state](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator.go#L203-L206) would cause a race condition between the threads A and B concurrently running [`handleExecutionReceipt`](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator.go#L126), e.g., while an attack was ongoing on [thread A](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator.go#L203-L206), [thread B](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator.go#L157) would still assume there is no attack conducted, hence would attempt on an attack. This would cause more than one result get corrupted, which would cause a flakey behavior on the exact test that was supposed to catch this edge case: [`TestTwoConcurrentExecutionReceipts_DistinctResult`](https://github.com/onflow/flow-go/blob/master/insecure/wintermute/attackOrchestrator_test.go#L66).

This PR fixes the aforementioned race condition by making sure that conducting the attack as well as reading and writing to the attack state are all done with [mutual exclusion provided](https://github.com/onflow/flow-go/blob/bbf2c0659256ddc98b2bc2d54ea8dc3ab2beb99e/insecure/wintermute/attackOrchestrator.go#L31-L32).


gomisha:
Tested locally multiple times and it's passing now:

`go test -v --tags relic ./insecure/wintermute/... -count 10`

`ok      github.com/onflow/flow-go/insecure/wintermute   28.338s`

gomisha (approved):

peterargue (approved):

PR #2783, opened 2022-07-11
Observer integration tests

danielholmes839:
Implemented Observer integration tests
- `TestObserverConnection` tests that the observer can be pinged successfully but returns an error when the upstream access node is stopped
- `TestObserverWithoutAccess` tests that the observer returns errors when the access node is stopped
-  `TestObserverCompareRPCs` tests that the observer and access node return the same errors


cmd/observer/node_builder/observer_builder.go
```diff
@@ -994,7 +992,33 @@ func (builder *ObserverServiceBuilder) enqueueRPCServer() {
...
+			return nil, fmt.Errorf("failed to get default TLS client config using public flow networking key %s %w", identity.NetworkPubKey.String(), err)
+		}
+
+		conn, _ := grpc.Dial(
+			identity.Address,
+			grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(grpcutils.DefaultMaxMsgSize)),
+			grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
+			backend.WithClientUnaryInterceptor(time.Minute))
+
+		client := accessproto.NewAccessAPIClient(conn)
```
danielholmes839, line 1018: Is it ok to do anything fancy right now when connecting to the upstream access node? I could add something like looping over all upstream-node-addresses and picking the first one that works. There was some more robust code in /apiproxy directory I removed. But I could add some of the features back

koko1123: you can add in the minimum feature of being able to connect to multiple staked ANs, and the retry strategy you described above. No need for the round robin lookup though, you can always loop over them in the order they were specified in the args

engine/access/rpc/engine_builder.go
```diff
@@ -68,15 +68,7 @@ func (builder *RPCEngineBuilder) WithMetrics() *RPCEngineBuilder {
...
-
-	if builder.router != nil {
-		builder.router.SetLocalAPI(localAPIServer)
-		localAPIServer = builder.router
-	}
-
-	accessproto.RegisterAccessAPIServer(builder.unsecureGrpcServer, localAPIServer)
-	accessproto.RegisterAccessAPIServer(builder.secureGrpcServer, localAPIServer)
-
+	accessproto.RegisterAccessAPIServer(builder.unsecureGrpcServer, builder.handler)
```
koko1123, line 71: nice

integration/tests/access/observer_test.go
```diff
@@ -0,0 +1,274 @@
...
+	assert.NoError(t, err)
+
+	// stop the upstream access container
+	err = suite.net.StopContainerByName(ctx, "access_1")
+	assert.NoError(t, err)
+
+	// ping the observer when access container is stopped
+	_, err = observer.Ping(ctx, &accessproto.PingRequest{})
+	assert.Error(t, err)
+}
```
koko1123, line 110: good idea

cmd/observer/node_builder/observer_builder.go
danielholmes839: Implemented

koko1123 (approved):

lolpuddle (approved):

peterargue (commented):
@danielholmes839 sorry I didn't review this earlier. I added some comments we should consider before calling this complete.

apiproxy/access_api_proxy.go
```diff
@@ -1,468 +0,0 @@
-package apiproxy
```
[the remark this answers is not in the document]
danielholmes839: I think it works but the handler wasn't set properly in the observer builder so I decided to simplify things. I will add back the round robin / multiple ANs in the PR for metrics and logs. I'm planning on writing something that will include metrics/logging but also let us configure per RPC how to handle it. I'm thinking: 
- forward immediately
- handle locally if there's an error then forward
- only handle locally

engine/access/rpc/forwarder.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/onflow/flow/protobuf/go/flow/access"
+	accessproto "github.com/onflow/flow/protobuf/go/flow/access"
+)
+
+type Forwarder struct {
+	UpstreamHandler accessproto.AccessAPIClient
+}
+
+func (f *Forwarder) Ping(context context.Context, req *access.PingRequest) (*access.PingResponse, error) {
```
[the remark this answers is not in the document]
danielholmes839: I'll have this in the next PR

apiproxy/access_api_proxy.go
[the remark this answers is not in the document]
peterargue: for the current observer grpc, you can keep the forward/local logic the same as what was there. There are certain endpoints that should always be served locally, and others that must be forwarded. I think if we're going to add smarts and better failover logic, that should probably go into the API service.

since the existing code went through several rounds of code review already, I think we should fix what's not working instead of a complete rewrite.

integration/testnet/network.go
```diff
@@ -750,6 +753,132 @@ func (net *FlowNetwork) addConsensusFollower(t *testing.T, rootProtocolSnapshotP
...
+			Binds: []string{
+				fmt.Sprintf("%s:%s:rw", flowDataDir, "/data"),
+				fmt.Sprintf("%s:%s:rw", flowProfilerDir, "/profiler"),
+				fmt.Sprintf("%s:%s:ro", nodeBootstrapDir, "/bootstrap"),
+			},
+			PortBindings: nat.PortMap{
+				"9000": []nat.PortBinding{{HostIP: "0.0.0.0", HostPort: "9000"}},
+				"9001": []nat.PortBinding{{HostIP: "0.0.0.0", HostPort: "9001"}},
+				"8000": []nat.PortBinding{{HostIP: "0.0.0.0", HostPort: "8000"}},
+			},
```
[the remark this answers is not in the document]
danielholmes839: Opened PR: https://github.com/onflow/flow-go/pull/2886

Issue #2759, opened 2022-07-07
[Execution] blocks are not sealed on localnet with only one EN node


Tonix517:
## 🐞 Bug Report

On localnet, when there are only one EN node instance, from the Grafana dashboards no blocks are sealed. Only after `EXECUTION=2` is set in `make init`, blocks can be sealed.

## What is the severity of this bug?

**Could have:** It really doesn't matter if we do this (product functions without this, impact to user is minimal). 
only impacting developer testing

## Reproduction steps

Steps to reproduce the behaviour:

1. `make -e EXECUTION=1 init`
2. `make start`
3. Open local Grafana and navigate to Max\Min Sealed blocks, and Unsealed Blocks panel to see all 0s.

- [X] I followed these steps and was able to reproduce the issue

## Expected behaviour

The number of sealed blocks should show green valid numbers on Grafana dashboards mentioned above.

## Screenshots

<img width="338" alt="Screen Shot 2022-07-07 at 2 16 10 PM" src="https://user-images.githubusercontent.com/5504323/177872732-cd4c967d-6f37-41df-99d1-ae8f7e3532e7.png">

## Specifications

- System: Mac M1 Pro, macOS 12.4
- Go: 1.18.3

Tonix517:
Found the PR that caused this behavior: https://github.com/onflow/flow-go/pull/632/files#diff-7fd319f5e35ff086e91814317aadbcb86eb3a89b904bf6a3cc38cb7a6ac40bf6R41

Tonix517:
confirmed with @durkmurder - it is required to run at least 2 EN nodes to avoid creating seal in case of execution forks.

PR #632, opened 2021-04-15
[Consensus] Yurii/5412 approval processing core

durkmurder:
implements [dapperlabs/flow-go/issues/5412](https://github.com/dapperlabs/flow-go/issues/5412)

### Overview: 

This PR implements new logic for processing approvals as described in [this](https://www.notion.so/dapperlabs/Concurrent-ResultApproval-Processing-7c6e1530534a44189708c9ed86002c6d) document.

- concept of `authorizedVerifiers` got removed since it duplicates chunk assignment
- fixed `ChunkAssigner` to filter out ejected nodes
- implemented complete result approvals processing pipeline
- emergency sealing & re-requesting approvals are part of this PR
- ATTENTION: contains an important change to leveled forest and `ChunkAssigner`

PR #2788, opened 2022-07-12
[localnet] Adding checking to EXECUTION configuration

Tonix517:
#2759 

#### Motivation
According to the change below, multiple execution nodes are required to prevent block seals to be generated in case of execution forks. Thus if there's only one EN node running, no blocks will be sealed. To avoid future confusion in localnet uses, now we put this checking into the Makefile.
https://github.com/onflow/flow-go/pull/632/files#diff-7fd319f5e35ff086e91814317aadbcb86eb3a89b904bf6a3cc38cb7a6ac40bf6R41-R42

Also applying the same check for consensus node.

#### Test
- [X] Run `make -e EXECUTION=1 init` and `make -e CONSENSUS=1 init` to see new error
- [X] Run all other EXECUTION\CONSENSUS values to see the make command to go through

### DIFF
--- a/integration/localnet/Makefile
+++ b/integration/localnet/Makefile
@@ -1,5 +1,6 @@
 COLLECTION = 3
 CONSENSUS = 3
+VALID_CONSENSUS := $(shell test $(CONSENSUS) -ge 2; echo $$?)
 EXECUTION = 2
 VALID_EXECUTION := $(shell test $(EXECUTION) -ge 2; echo $$?)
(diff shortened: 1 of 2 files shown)


SaveTheRbtz (approved):
You can consider adding the same for consensus.

pattyshack (approved):

Tonix517:
> You can consider adding the same for consensus.

addressed

PR #2790, opened 2022-07-12
dedup code in script/transaction environment (part 1 of many)

pattyshack:
A big chunk of the code in ScriptEnv and TransactionEnv are copy/paste
identical.  It's easy to forget to fix bugs in both place and hard to make
improvements.

Notes:
 1. I've decided to merge the shared code piecemeal since it's scary to
    merge everything at once.
 2. The environment variable e has been renamed to env since it's more
    searchable
 3. (A real change) I've presized the array allocation in RecordTrace

### DIFF
--- a/fvm/env.go
+++ b/fvm/env.go
@@ -1,7 +1,22 @@
 package fvm
 
 import (
+	"encoding/binary"
+	"math/rand"
+	"time"
+
 	"github.com/onflow/cadence/runtime"
+	"github.com/onflow/cadence/runtime/common"
+	"github.com/onflow/cadence/runtime/interpreter"
+	"github.com/opentracing/opentracing-go"
+	traceLog "github.com/opentracing/opentracing-go/log"
+
+	"github.com/onflow/flow-go/fvm/crypto"
+	"github.com/onflow/flow-go/fvm/handler"
+	"github.com/onflow/flow-go/fvm/programs"
+	"github.com/onflow/flow-go/fvm/state"
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/trace"
 )
 
 // Environment accepts a context and a virtual machine instance and provides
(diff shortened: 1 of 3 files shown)


fvm/env.go
```diff
@@ -11,3 +26,120 @@ type Environment interface {
...
+	return nil
+}
+
+func (env *commonEnv) RecordTrace(operation string, location common.Location, duration time.Duration, logs []opentracing.LogRecord) {
+	if !env.isTraceable() {
+		return
+	}
+	if location != nil {
+		if logs == nil {
+			logs = make([]opentracing.LogRecord, 0, 1)
```
pattyshack, line 79: The only line of code with actual change.  the original was `logs = make([]opentracing.LogRecord, 0)`

fvm/env.go
```diff
@@ -11,3 +26,120 @@ type Environment interface {
...
+func (env *commonEnv) VM() *VirtualMachine {
+	return env.vm
+}
+
+func (env *commonEnv) seedRNG(header *flow.Header) {
+	// Seed the random number generator with entropy created from the block
+	// header ID. The random number generator will be used by the UnsafeRandom
+	// function.
+	id := header.ID()
+	source := rand.NewSource(int64(binary.BigEndian.Uint64(id[:])))
```
SaveTheRbtz, line 60: this should not work since 50% of time it will panic on overflowing the int64.

fvm/env.go
```diff
@@ -11,3 +26,120 @@ type Environment interface {
...
+
+func (env *commonEnv) ValueDecoded(duration time.Duration) {
+	env.RecordTrace("decodeValue", nil, duration, nil)
+	env.metrics.ValueDecoded(duration)
+}
+
+// Commit commits changes and return a list of updated keys
+func (env *commonEnv) Commit() ([]programs.ContractUpdateKey, error) {
+	// commit changes and return a list of updated keys
+	err := env.programs.Cleanup()
```
SaveTheRbtz, line 117: I'm a bit curious why we would need to do that for script execution, but in the context of this refactoring this is a no-op.

fvm/env.go
```diff
@@ -11,3 +26,120 @@ type Environment interface {
...
+}
+
+func (env *commonEnv) RecordTrace(operation string, location common.Location, duration time.Duration, logs []opentracing.LogRecord) {
+	if !env.isTraceable() {
+		return
+	}
+	if location != nil {
+		if logs == nil {
+			logs = make([]opentracing.LogRecord, 0, 1)
+		}
```
SaveTheRbtz, line 80: why do we need that?  i thought that append takes care of this.

SaveTheRbtz (approved):

fvm/env.go
bluesign: does golang overflow / panic on this?

SaveTheRbtz: Good pint! Go does not panic in runtime on overflow but instead fails to compile when it can detect the overflow:

this fails to compile:
```
_ = int64(math.MaxUint64)
```

while following works fine:
```
x := uint64(math.MaxUint64)
_ = int64(x)
```

fvm/env.go
SaveTheRbtz: https://go.dev/play/p/Yr6u9JLEff6

pattyshack: as you've pointed out, we don't really need this.  i kept this line (but tweaked it to pre-size to 1) to minimize the number of changes.

pattyshack:
bors merge

Issue #2286, opened 2022-04-09
[Execution State] Avoid creating separate MTrie state during checkpoint creation to reduce peak RAM use by ~150GB and checkpoint duration by 24 minutes


fxamacker:
## Problem

Recent increase in transactions is causing WAL files to get created more frequently, causing checkpoints to happen more frequently, increasing checkpoint file size, and increasing ledger state size in memory.  These increases are causing checkpointing to consume too much RAM and take more than 2x longer than earlier this year.

|   | File Size | Checkpoint Frequency | 
| ------------- | ---: | --- |
| Early 2022     | 53 GB | 0-2 times per day |
| July 8, 2022  | 126 GB | every 2 hours  |

Without PR #1944 the system checkpointing would currently be:

- taking well over 20-30 hours each time, making it impossible to complete every 2 hours
- requiring more operational RAM, making OOM crashes very frequent
- creating billions more allocations and gc pressure, consuming CPU cycles and slowing down EN

After PR #1944 reduced Mtrie flattening and serialization phase to under 5 minutes (which sometimes took 17 hours on mainnet16), creating a separate MTrie state currently accounts for most of the duration and memory used by checkpointing.  This opens up new possibilities such as reusing ledger state to significantly reduce duration and operational RAM of checkpointing again.

Updates epic https://github.com/onflow/flow-go/issues/1744

## The Proposed Solution

We can avoid creating a separate MTrie state during checkpoint creation.  This can reduce peak RAM use by (very roughly) about 150GB and reduce checkpoint duration by 24 minutes (estimates based on snapshot of July 8, 2022).  Memory savings will increase over time.

Determine if it's feasible to avoid creating a separate MTrie state during checkpoint creation.  If the poof-of-concept doesn't reveal showstoppers then proceed with new PR.

- [x] Proof-of-concept #2770 - show it is feasible to avoid creating separate MTrie state.

- [x] PR #2792 This has fewer edge cases to handle than PR 2770 and looser coupling between layers but may require extra locks compared.

m4ksio:
Currently WAL/Checkpoints are disconnected from the mtrie progression - checkpoint is essentially a state after given complete WAL segment, and segments creation is an implementation detail for mForest and WAL.
If, however, we were able to signal the moment of new WAL creation, we should stop evicting mtries/keep a separate index and use it to create new checkpoint without reallocating memory.

fxamacker:
Closed by #2792

PR #2792, opened 2022-07-13
[Execution Node] Reuse ledger state in checkpoints for -152GB RAM and -24 minutes

fxamacker:
Avoid creating separate ledger state during checkpointing.

Updates #2286 
Updates #1744

### Impact

Based on EN2 logs (July 8, 2022), this will
- reduce operational memory (peak RAM) by at least 152GB
- reduce checkpoint duration by 24 mins (from 45 mins)
- reduce memory allocations (TBD)

Ledger state is continuously growing larger, so memory & duration savings will be better than listed as each day passes.

### Context

Recent increase in transactions is causing WAL files to get created more frequently, which causes checkpoints to happen more frequently, increases checkpoint file size, and increases ledger state size in memory.

|   | File Size | Checkpoint Frequency | 
| ------------- | ---: | --- |
| Early 2022     | 53 GB | 0-2 times per day |
| July 8, 2022  | 126 GB | every 2 hours  |

Without PR #1944 the system checkpointing would currently be:

- taking well over 20-30 hours each time, making it impossible to complete every 2 hours
- requiring more operational RAM, making OOM crashes very frequent
- creating billions more allocations and gc pressure, consuming CPU cycles and slowing down EN

After PR #1944 reduced Mtrie flattening and serialization phase to under 5 minutes (which sometimes took 17 hours on mainnet16), creating a separate MTrie state currently accounts for most of the duration and memory used by checkpointing.  This opens up new possibilities such as reusing ledger state to significantly reduce duration and operational RAM of checkpointing again.

### Design

Design goal is to reduce operational RAM, reduce allocations, and speed up checkpointing by not creating a separate ledger state.

To achieve these goals, this PR: 
- reuses tries from the main ledger state for checkpointing
- avoids blocking with time-consuming tasks such as creating checkpoint 

The `Compactor`:
- receives trie updates and new trie created from the update
- saves encoded updates in WAL segment 
- tracks active segment number returned by `RecordUpdate()` 
- starts checkpointing async when enough finalized segments are accumulated

NOTE: To reuse ledger tries for checkpointing, new tries must match its corresponding WAL update. 

### TODO:
- [x] add more tests
- [x] handle known edge cases 
- [x] handle errors
- [x] add more tests related to concurrency (longer duration test runs on benchnet)
- [x] preliminary review of draft PR by @zhangchiqing
- [x] remove draft status early to get @m4ksio and @ramtinms reviews
- [x] preliminary review and approval of PR by @m4ksio
- [x] preliminary review and approval of PR by @ramtinms 
- [x] incorporate feedback from preliminary reviews
- [x] add more tests
- [ ] awesome suggestion by @ramtinms  which should not be shortchanged "... run tests on benchnet by turning nodes off to make sure the logic is safe and no unknown case is there."
- [ ] continue looking for edge cases we may have missed
- [x] cleanup code and follow best practices (e.g. document errors returned from functions, etc.)
- [ ] request code reviews for non-urgent items (micro optimizations, improved design, etc.)

See TODOs listed in the source code for more details.


ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+// This ensures that WAL update and ledger state update are in sync.
+func processUpdateResult(update *WALTrieUpdate, updateResult error) {
+	// Send result of WAL update
+	update.ResultCh <- updateResult
+
+	// Wait for trie update to complete
+	<-update.DoneCh
+}
+
+func (c *Compactor) checkpoint(ctx context.Context, tries []*trie.MTrie, checkpointNum int) error {
```
zhangchiqing, line 238: checkpoint doesn't happen often. It worth to add logs about when checkpointing is started, and when it's finished, as well as when it's aborted or failed.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	default:
+	}
+
+	err = cleanupCheckpoints(c.checkpointer, int(c.checkpointsToKeep))
+	if err != nil {
+		return fmt.Errorf("cannot cleanup checkpoints: %w", err)
+	}
+
+	if checkpointNum > 0 {
+		for observer := range c.observers {
```
zhangchiqing, line 258: Do we need to place a lock before loop through the observers? 

Is it possible that Subscribe or Unsubscribe is called concurrently which will modify the observers?

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+		// if condition guarantees this never fails
+		checkpointsToRemove := checkpoints[:len(checkpoints)-int(checkpointsToKeep)]
+
+		for _, checkpoint := range checkpointsToRemove {
+			err := checkpointer.RemoveCheckpoint(checkpoint)
+			if err != nil {
+				return fmt.Errorf("cannot remove checkpoint %d: %w", checkpoint, err)
+			}
+		}
+	}
```
zhangchiqing, line 322: better to split into two functions, one to compute what files to remove, the other to remove by the specified files.

The first function would be easier to write unittest for.

ledger/complete/ledger.go
```diff
@@ -86,19 +89,64 @@ func NewLedger(
...
+	go func() {
+		defer close(done)
+		if l.trieUpdateCh != nil {
+			// Compactor is running in parallel.
+			// Ledger is responsible for closing trieUpdateCh channel,
+			// so Compactor can drain and process remaining updates.
+			close(l.trieUpdateCh)
+
+			// Wait for Compactor to finish all trie updates before closing WAL component.
+			<-l.trieUpdateDoneCh
```
zhangchiqing, line 145: Would this wait forever? What guarantees that eventually there will be a message pushed to this channel?

ledger/complete/ledger.go
```diff
@@ -229,10 +266,44 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
+func (l *Ledger) set(trieUpdate *ledger.TrieUpdate) (newState ledger.State, err error) {
+
+	resultCh := make(chan error)
+
+	if l.trieUpdateCh == nil {
+		go func() {
+			_, _, err := l.wal.RecordUpdate(trieUpdate)
+			resultCh <- err
+		}()
+	} else {
```
zhangchiqing, line 284: What's the use case that we create a Ledger without the trieUpdateCh? 

If the creator of a leader doesn't care about the trieUpdate, we can use a for-loop to ignore the updates. That would allow us to remove the if branch.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+			if checkpointResult.err != nil {
+				c.logger.Error().Err(checkpointResult.err).Msgf(
+					"compactor failed to checkpoint %d", checkpointResult.num,
+				)
+
+				// Retry checkpointing after active segment is finalized.
+				nextCheckpointNum = activeSegmentNum
+			}
+
+		case update, ok := <-c.trieUpdateCh:
```
zhangchiqing, line 173: I think it's better to move the checkpoint related logic to a separate module, rather than mixing together here.

Basically, there are two pipelines:
The first pipeline, subscribes from trieUpdateCh and decide whether a segment has been created. Once created, it passes the segment to the next pipeline
The second pipeline, will subscribe the segment events, and decide whether to create a checkpoint. Since checkpointing takes time, if won't block the first pipeline from processing the trieUpdates, or passing the segment even if the checkpointing hasn't been done yet. 

Correct me if wrong. The problem of the current implementation is that during checkpointing, we won't process trie update.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+				// - not enough segments for checkpointing (nextCheckpointNum >= activeSegmentNum), or
+				// - failed to get ledger state snapshop (nextCheckpointNum < activeSegmentNum)
+				if nextCheckpointNum < activeSegmentNum {
+					nextCheckpointNum = activeSegmentNum
+				}
+				continue
+			}
+
+			// Try to checkpoint
+			if checkpointSem.TryAcquire(1) {
```
zhangchiqing, line 196: If a second call is trying to acquire before the first call release it, would the second call get blocked and get resumed after the first call is done?

cmd/execution_builder.go
```diff
@@ -343,12 +343,6 @@ func (e *ExecutionNodeBuilder) LoadComponentsAndModules() {
 			}
 			return nil
 		}).
-		Component("Write-Ahead Log", func(node *NodeConfig) (module.ReadyDoneAware, error) {
```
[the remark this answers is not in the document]
fxamacker: `DiskWAL` needs to be a dependent component of `Ledger` because I found component shutdown is parallel (although it used to be in sequence).

ledger/complete/compactor.go
fxamacker: No, the second call would return without blocking.  If it fails to acquire semaphore, it retries checkpointing when active segment is finalized.

ledger/complete/compactor.go
fxamacker: > Correct me if wrong. The problem of the current implementation is that during checkpointing, we won't process trie update.

That is incorrect.  Checkpointing is done in a [new goroutine.](https://github.com/onflow/flow-go/blob/65f20620dc38ec35efce75e407a6ee241410c091/ledger/complete/compactor.go#L201-L205)

Compactor can process trie update while checkpointing is running (they execute in parallel).  Also, semaphore makes sure only one checkpointing is running at a time.

fxamacker: > Basically, there are two pipelines:
The first pipeline, subscribes from trieUpdateCh and decide whether a segment has been created. Once created, it passes the segment to the next pipeline
The second pipeline, will subscribe the segment events, and decide whether to create a checkpoint. Since checkpointing takes time, if won't block the first pipeline from processing the trieUpdates, or passing the segment even if the checkpointing hasn't been done yet.

Although I agree that there are two pipelines, I think they need to have different responsibilities from what you mentioned.  The goal is to not hold on to tries in memory that don't need to be checkpointed.

The first pipeline: subscribes from trieUpdateCh, decides whether a segment has been created, and decides whether a checkpoint needs to be created.  More importantly, it needs to take a snapshot of main ledger state at segment boundary when checkpointing is needed.  Taking a snapshot requires communication with `Ledger.Set` to ensure that WAL update and ledger state update are in sync.

The second pipeline is to create checkpoint file with ledger state snapshot taken from previous pipeline.  Checkpointing is done in a separate goroutine so it doesn't block the first pipeline.

Thoughts?

ledger/complete/ledger.go
fxamacker: > What's the use case that we create a Ledger without the trieUpdateCh?

It's easier to not require MockCompactor be running when testing Ledger and maybe others.  

> If the creator of a leader doesn't care about the trieUpdate, we can use a for-loop to ignore the updates. That would allow us to remove the if branch.

I'm not sure I understand.  Can you elaborate?   I'm not a fan of this `if` branch too.

ledger/complete/ledger.go
fxamacker: > Would this wait forever? What guarantees that eventually there will be a message pushed to this channel?

No because `Compactor` closes `trieUpdateDoneCh` [on shutdown](https://github.com/onflow/flow-go/blob/65f20620dc38ec35efce75e407a6ee241410c091/ledger/complete/compactor.go#L110-L111).  

Thanks for bringing this up :+1: , I think it would be safer to use defer to close `trieUpdateDoneCh`, so I pushed commit [b8797cc](https://github.com/onflow/flow-go/pull/2792/commits/b8797cc2e4564ffceae8cde1bb577cf2fd0863ba).

ledger/complete/compactor.go
fxamacker: The checkpoint logging is already done in [`createCheckpoint()`](https://github.com/onflow/flow-go/blob/b8797cc2e4564ffceae8cde1bb577cf2fd0863ba/ledger/complete/compactor.go#L274-L301).

ledger/complete/ledger.go
```diff
@@ -229,10 +266,44 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
 	l.logger.Info().Hex("from", state[:]).
-		Hex("to", newRootHash[:]).
+		Hex("to", newState[:]).
 		Int("update_size", update.Size()).
 		Msg("ledger updated")
-	return ledger.State(newRootHash), trieUpdate, nil
+	return newState, trieUpdate, nil
+}
+
+func (l *Ledger) set(trieUpdate *ledger.TrieUpdate) (newState ledger.State, err error) {
```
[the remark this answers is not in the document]
zhangchiqing: I feel there are quite many edge cases when allowing trie updates for different blocks to be written to WAL concurrently. Those edge cases would cause incomplete trie updates and duplications in the WAL files. 

I also wonder how does it work to load trie updates that are saved out of order. 
For instance, say block 10 has two trie updates, the second update is based on a new node that is created by the first trie update. However, if the second trie update is written before the first trie update, how can we add the second trie update before the first? Wouldn't it fail because it the new node created by the first trie update doesn't exist yet?

cc @m4ksio

ledger/complete/ledger.go
```diff
@@ -229,10 +266,44 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
+
+	resultCh := make(chan error)
+
+	if l.trieUpdateCh == nil {
+		go func() {
+			_, _, err := l.wal.RecordUpdate(trieUpdate)
+			resultCh <- err
+		}()
```
zhangchiqing, line 283: how does concurrent write to a disk WAL file work? Is it ultimately only one file writer writing at a time, and all other writers are blocked waiting?

If so, we might want to make WAL writing sequential, in order to ensure all trie updates of a block are written to a segment file before writing the next.

fxamacker: The code you highlighted is existing code currently running in mainnet.  With this PR, that block of code is only used in tests for `Ledger`, etc without having to run a MockCompactor.  Getting rid of it is on my mind but it can be done later because it doesn't impact the main goal of this PR.

> how does concurrent write to a disk WAL file work? Is it ultimately only one file writer writing at a time, and all other writers are blocked waiting?

Yes.  `wal` package uses write lock so only one writing at a time.

> If so, we might want to make WAL writing sequential, in order to ensure all trie updates of a block are written to a segment file before writing the next.

WAL writing is already sequential by sending `trieUpdate` over a channel so `Compactor` processes `trieUpdate` in sequence.  https://github.com/onflow/flow-go/blob/65f20620dc38ec35efce75e407a6ee241410c091/ledger/complete/ledger.go#L284-L289

ledger/complete/ledger.go
```diff
@@ -86,19 +89,64 @@ func NewLedger(
...
+}
+
+// TrieUpdateChan returns a channel which is used to receive trie updates that needs to be logged in WALs.
+// This channel is closed when ledger component shutdowns down.
+func (l *Ledger) TrieUpdateChan() <-chan *WALTrieUpdate {
+	return l.trieUpdateCh
+}
+
+// TrieUpdateDoneChan returns a channel which is closed when there are no more WAL updates.
+// This is used to signal that it is safe to shutdown WAL component (closing opened WAL segment).
```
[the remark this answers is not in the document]
fxamacker: > Why do we need extra channel? Can we just close the channel to indicate there is no more updates?

We use the extra channel because when trie update channel is closed, Compactor still needs to drain and process remaining trie update in the closed channel.  Ledger is responsible for closing opened WAL segment, so `TrieUpdateDoneChan` is used to indicate all WAL writing activities are done in `Compactor`.

On the other hand, if `Compactor` is responsible for closing opened WAL segment, this extra communication isn't needed.  Thanks for bringing this up :+1: , I'll look into it.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func (c *Compactor) run() {
+
+	// checkpointSem is used to limit checkpointing to one.
+	// If previous checkpointing isn't finished when enough segments
+	// are finalized for next checkpointing, retry checkpointing
+	// again when next segment is finalized.
+	// This avoids having more tries in memory than needed.
+	checkpointSem := semaphore.NewWeighted(1)
```
zhangchiqing, line 123: I don’t understand how this works. Every call will create a new semaphore object. Wouldn’t “TryAcquire” always return true? Don’t we need to share the semaphore object between different calls?

fxamacker: > I don’t understand how this works.  Every call will create a new semaphore object. 

No, that is incorrect.  `checkpointSem` is initialized outside of goroutine's `for` loop which waits for and processes messages from channels.  So it is created once and used inside `for` loop to limit one checkpointing at a time. 

> Wouldn’t “TryAcquire” always return true? Don’t we need to share the semaphore object between different calls?

No, `TryAquire` doesn't always return true.  All checkpointings need to acquire the same semaphore.  So this limits to one checkpointing at a time, and all other calls to `TryAcquire` returns false immediately (non-blocking).

ledger/complete/compactor.go
fxamacker: > Do we need to place a lock before loop through the observers?
>
> Is it possible that Subscribe or Unsubscribe is called concurrently which will modify the observers?

This code block is unchanged by this PR.  I can look into this later.

ledger/complete/ledger.go
```diff
@@ -229,10 +266,44 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
+
+	resultCh := make(chan error)
+
+	if l.trieUpdateCh == nil {
+		go func() {
+			_, _, err := l.wal.RecordUpdate(trieUpdate)
+			resultCh <- err
+		}()
+	} else {
+		doneCh := make(chan struct{})
```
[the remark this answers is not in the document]
fxamacker: > This means, if ledger.Set is called concurrently by two calls, one of the call will be blocked waiting until the first call is completed.

No, that's not exactly how it works.  

When `Ledger.Set` is called concurrently by two calls, both send `trieUpdate` message to `Compactor` non-blocking, and create new tries from updates concurrently.  What is blocking is `Compactor` can only process one `trieUpdate` at a time, so the second `Ledger.Set` waits for `Compactor` to process its message in turn.  

In other words, two `Ledger.Set` create new tries concurrently while updating WAL and adding new trie to ledger state are in sequence.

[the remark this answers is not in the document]
fxamacker: > IMO, the behavior of a RWMutex lock is easier than the doneCh to understand, even though both can be used to ensure concurrent-safety.

I understand what you're saying but adding RWMutex doesn't help because we need the extra channel based on discussions @m4ksio, you, and I had on Friday morning.

Based Friday's meeting, I replaced `doneCh` with `trieCh` to send new trie from `Ledger.Set` to `Compactor`.  It is safer to use a FIFO queue to store tries for checkpointing in `Compactor` than taking a snapshot of main ledger state.    

With this change, `ledger.Set` should communicate new trie to `Compactor` using the third channel.

One alternative approach to use 2 channels instead of 3 channels is:
- sending both `trieUpdate` and new trie from `Ledger.Set` to `Compactor` in first channel
- waiting for WAL update result in second channel

However, the alternative approach would take longer for updates so it isn't worthwhile.

ledger/complete/ledger.go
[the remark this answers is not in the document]
fxamacker: > I feel there are quite many edge cases when allowing trie updates for different blocks to be written to WAL concurrently.   Those edge cases would cause incomplete trie updates and duplications in the WAL files.

I hope the meeting we had on Friday with @m4ksio resolved your concerns.  

Checkpointing in this PR properly handles parallel trie updates.  Changing or improving the use of concurrent trie updates outside of checkpointing is outside the scope of this PR.

ledger/complete/ledger.go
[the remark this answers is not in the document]
zhangchiqing: Yes, you are right. Adding the trie update to forest could be concurrent. But the problem is it’s not safe. TrieUpdates could have dependency, and has to be applied sequentially if there is.  For instance, if a block has two trieUpdates, the second update might build on a node created by the first update, adding them concurrently would fail. 
Concurrent update could only work if the blocks of the updates are from different forks, and all the updates they depend on has been added to the forest.
We’ve been running this code for a long time, and never run into problem, is it because concurrent update actually worked? We actually concurrently adding updates to the forest? That’s why I checked the code and see whether Ledger.set could be called concurrently. Turns out, it’s always called sequentially, which means the concurrent update case never happens, and we never run into problem. 

In other words, the ledger set is not safe to be called concurrently with two updates. Ledger has no protocol level knowledge, has no visibility of the dependencies between two updates, so it’s only safe to be called sequentially for now, which is what we do currently.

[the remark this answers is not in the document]
fxamacker: > Adding the trie update to forest could be concurrent. But the problem is it’s not safe. 

No that is incorrect.

In this PR [creating new trie from update](https://github.com/onflow/flow-go/blob/fd43fd5b7c8af9dfbc10030d9634a5ca66e94de8/ledger/complete/ledger.go#L254) is concurrent.  However, [adding the new trie to forest](https://github.com/onflow/flow-go/blob/fd43fd5b7c8af9dfbc10030d9634a5ca66e94de8/ledger/complete/ledger.go#L264) is sequential (by use of channels).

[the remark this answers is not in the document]
fxamacker: >  For instance, if a block has two trieUpdates, the second update might build on a node created by the first update, adding them concurrently would fail.

No it won't fail because trie updates are in sequence for a block, so this case is handled properly.

> Concurrent update could only work if the blocks of the updates are from different forks, and all the updates they depend on has been added to the forest.

Yes, @m4ksio confirmed in #flow-execution that concurrent trie updates only happen if the block forks happened.

ledger/complete/compactor.go
zhangchiqing: I see. The lock is created outside of the for-loop. 

If the first call finishes before the second call, then the second call will trigger a new checkpoint.
If the first call hasn't finished before the second call, then the second call will not trigger a new checkpoint. After the first call finishes, it will also not trigger checkpointing. If the first call finishes before the third call, then the third call will trigger the second checkpoint. 

Right?

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	// Get tries from checkpoint queue.
+	// At this point, checkpoint queue contains tries up to
+	// last update (logged as last record in finalized segment)
+	// It doesn't include trie for this update
+	// until updated trie is received and added to checkpointQueue.
+	tries := checkpointQueue.Tries()
+
+	checkpointNum = nextCheckpointNum
+
+	return activeSegmentNum, checkpointNum, tries
```
zhangchiqing, line 392: Does this function need to return `checkpointNum`? It seems if it ever returns, it's always equal to the `nextCheckpointNum` from the input. The call could just use the input argument, instead of the output of this function

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	// Check new segment number is incremented by 1
+	if segmentNum != activeSegmentNum+1 {
+		c.logger.Error().Msg(fmt.Sprintf("compactor got unexpected new segment numer %d, want %d", segmentNum, activeSegmentNum+1))
+	}
+
+	// Update activeSegmentNum
+	prevSegmentNum := activeSegmentNum
+	activeSegmentNum = segmentNum
+
+	if nextCheckpointNum > prevSegmentNum {
```
zhangchiqing, line 374: The logic of when to trigger checkpoint is spread in many places. I would suggest to create a CheckpointTrigger to decouple the logic from Compactor.

Here is the responsibility between them:
1. Compactor processes trieUpdate, forwards them to WAL writer. When WAL writer returns the segment num that the update was written to, it pass the segment num to CheckpointTrigger.
2. CheckpointTrigger is a stateful module, it holds the `activeSegmentNum`. When called with a segment num, it knows whether a new segment has been created, if created, it compares with its local checkpoint distance config to decide whether to trigger checkpoint. 
3. CheckpointTrigger holds the checkpointer module, when checkpoint is triggered, it places a lock and calls the checkpointer's method.
4. Checkpoint knows whether checkpointing has been running already, and can skip if the new checkpoint is triggered before the previous is finished
5. Compactor could pass the checkpointQueue to the checkpointTrigger along with the segment num, so that when checkpoint should be triggered, CheckpointTrigger has everything needed to trigger checkpoint.
6. Compactor knows nothing about checkpointing. Checkpoint related logic is all moved to the CheckpointTrigger, and can be tested much easier. 

```
type CheckpointTrigger struct {
  checkpointer // for running checkpointing
  checkpointDistance // the total number of segment (write-ahead log) to trigger checkpointing
  checkpointToKeep // the max number of checkpoint files to keep on disk in order to prevent too many checkpoint files filling disk.
  activeSegmentNum // the num of the segment that the previous trie update was trigger to, used to detect whether trie update is written to a new segment.
  isRunning // an atomic.Bool for whether checkpointing is currently running.
}

func (t *CheckpointTrigger) NotifyTrieUpdateWrittenToWAL(segmentNum int, checkpointQueue *CheckpointQueue) error {
  ...
}
```

m4ksio (approved):
Really great work!
A few cosmetic comments.
And one a bit larger - I think we need a test for a case @zhangchiqing has described and that we discussed - namely multiple concurrent Sets on a ledger and a test ensuring the order is preserved. 
It seems like code would behave this way, but explicit test is always best

module/chunks/chunkVerifier_test.go
```diff
@@ -271,6 +271,14 @@ func GetBaselineVerifiableChunk(t *testing.T, script string, system bool) *verif
...
 	f, _ := completeLedger.NewLedger(&fixtures.NoopWAL{}, 1000, metricsCollector, zerolog.Nop(), completeLedger.DefaultPathFinderVersion)
 
+	compactor := fixtures.NewNoopCompactor(f)
+	<-compactor.Ready()
+
+	defer func() {
+		<-f.Done()
+		<-compactor.Done()
+	}()
+
```
ramtinms, line 281: do we need this for this test? I don't think we use the compactor in this case.

cmd/bootstrap/run/execution_state.go
```diff
@@ -41,15 +42,23 @@ func GenerateExecutionState(
...
-	defer func() {
-		<-diskWal.Done()
-	}()
 
 	ledgerStorage, err := ledger.NewLedger(diskWal, 100, metricsCollector, zerolog.Nop(), ledger.DefaultPathFinderVersion)
 	if err != nil {
 		return flow.DummyStateCommitment, err
 	}
 
+	compactor, err := complete.NewCompactor(ledgerStorage, diskWal, zerolog.Nop(), 100, 1_000_000, 1)
```
ramtinms, line 51: when cleaning up please add constants for these default values.

ramtinms: and also we should pass the logger here as well in my view.

ledger/complete/mtrie/forest.go
```diff
@@ -190,19 +190,36 @@ func (f *Forest) Read(r *ledger.TrieRead) ([]ledger.Value, error) {
...
+// and returns rootHash and error (if any).
+// In case there are multiple updates to the same register, Update will persist
+// the latest written value.
+func (f *Forest) Update(u *ledger.TrieUpdate) (ledger.RootHash, error) {
+	t, err := f.NewTrie(u)
+	if err != nil {
+		return ledger.RootHash(hash.DummyHash), err
+	}
+
+	err = f.AddTrie(t)
```
ramtinms, line 203: in the case that NewTrie is returning parent trie (len zero) we are going to add it.

ledger/complete/wal/wal.go
```diff
@@ -49,20 +49,23 @@ func (w *DiskWAL) UnpauseRecord() {
...
 
-	_, err := w.wal.Log(bytes)
+	locations, err := w.wal.Log(bytes)
 
 	if err != nil {
-		return fmt.Errorf("error while recording update in LedgerWAL: %w", err)
+		return 0, false, fmt.Errorf("error while recording update in LedgerWAL: %w", err)
+	}
+	if len(locations) != 1 {
+		return 0, false, fmt.Errorf("error while recording update in LedgerWAL: got %d location, expect 1 location", len(locations))
```
ramtinms, line 65: if the len is always 1, can we just make it a pointer to a location or something similar ?

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+		update.ResultCh <- updateResult
+
+		// Wait for updated trie
+		trie := <-update.TrieCh
+		if trie == nil {
+			c.logger.Error().Msg("compactor failed to get updated trie")
+			return
+		}
+
+		checkpointQueue.Push(trie)
```
zhangchiqing, line 349: ⚠️  we should check `updateErr` before adding the trie to checkpoint queue, otherwise might cause inconsistency.

And if there is `updateError`, we probably don't have to wait for `<-update.TrieCh`, because we anyway won't add it to the checkpoint queue

ledger/complete/checkpointqueue.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"github.com/onflow/flow-go/ledger/complete/mtrie/trie"
+)
+
+// CheckpointQueue is a fix-sized FIFO queue of MTrie.
+// CheckpointQueue is intentionally not threadsafe given its limited use case.
+// CheckpointQueue is not a general purpose queue to avoid incurring overhead
+// for features not needed for its limited use case.
+type CheckpointQueue struct {
```
zhangchiqing, line 11: CheckpointQueue sounds like a queue with checkpointing files. 

If we want to name it with behavior, maybe we could call it `TrieQueue`. 

If we want to name it with functionality, maybe `LatestWrittenTries` would be more accurate. It holds the latest trie whose trie update have been written to the WAL file.

ledger/partial/ledger_test.go
```diff
@@ -20,9 +20,19 @@ import (
...
 	require.NoError(t, err)
 
+	compactor := fixtures.NewNoopCompactor(l)
+	<-compactor.Ready()
+
+	defer func() {
+		<-l.Done()
+		<-compactor.Done()
+	}()
+
```
ramtinms, line 35: what would happen if someone just initial the newLedger without the compactor ?

ramtinms (approved):
the logic looks good to me, just added a few minor comments. 

I recommend we do eventually run a few tests on benchnet by turning nodes off to make sure the logic is safe and no unknown case is there.

ledger/complete/ledger.go
```diff
@@ -229,10 +236,39 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
-	return ledger.State(newRootHash), trieUpdate, nil
+	return newState, trieUpdate, nil
+}
+
+func (l *Ledger) set(trieUpdate *ledger.TrieUpdate) (newState ledger.State, err error) {
+
+	resultCh := make(chan error, 1)
+
+	trieCh := make(chan *trie.MTrie, 1)
+	defer close(trieCh)
```
zhangchiqing, line 250: better to close the resultCh as well with `defer`

ledger/complete/ledger.go
```diff
@@ -229,10 +236,39 @@ func (l *Ledger) Set(update *ledger.Update) (newState ledger.State, trieUpdate *
...
+
+	l.trieUpdateCh <- &WALTrieUpdate{Update: trieUpdate, ResultCh: resultCh, TrieCh: trieCh}
```
zhangchiqing, line 252:
```suggestion
	// the following two processes are running concurrently:     
	// 1. writing the trie update to WAL 
	// 2. creating a new root node for the trie update
	// since process 1 happens in a different thread, we use the resultCh to wait for it to finish.
	// even though these two processes happens concurrently, process 1 actually depends
	//  on the result of process 2.
	// since the compactor caches the root nodes for minimize memory foot-print while checkpointing,
	// process 1 needs the root trie node from the result of process 2, that's why we create a `trieCh` for
	// process 2 to pass the created trie node to process 1. 
	l.trieUpdateCh <- &WALTrieUpdate{Update: trieUpdate, ResultCh: resultCh, TrieCh: trieCh}
```

ledger/complete/checkpointqueue_test.go
```diff
@@ -0,0 +1,132 @@
...
+				savedTries = append(savedTries, trie)
+
+				tr := q.Tries()
+				require.Equal(t, savedTries, tr)
+				require.Equal(t, len(savedTries), q.Count())
+			}
+		}
+
+		// Push more tries to queue to overwrite older elements.
+		for i := 0; i < capacity; i++ {
```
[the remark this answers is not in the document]
m4ksio: I don't think we want to override all, just a few to test a wrap-around logic

ledger/complete/checkpointqueue.go
```diff
@@ -0,0 +1,87 @@
...
+	capacity int
+	tail     int // element index to write to
+	count    int // number of elements (count <= capacity)
+}
+
+// NewCheckpointQueue returns a new CheckpointQueue with given capacity.
+func NewCheckpointQueue(capacity uint) *CheckpointQueue {
+	return &CheckpointQueue{
+		ts:       make([]*trie.MTrie, capacity),
+		capacity: int(capacity),
```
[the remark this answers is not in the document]
fxamacker: > any reason to keep capacity as int?

Yes.

`capacity` function argument as `uint` is to avoid checking for negative value.  

`CheckpointQueue.capacity` as `int` is to prevent casting it to `int` when using it with `len(ts)` or slice index. 

Also, overflow when casting from `uint` to `int` is unlikely given we're only using 500 as default.

ledger/complete/checkpointqueue_test.go
```diff
@@ -0,0 +1,132 @@
...
+		trie, err := randomMTrie()
+		require.NoError(t, err)
+
+		q.Push(trie)
+
+		savedTries = append(savedTries, trie)
+
+		tr := q.Tries()
+		require.Equal(t, capacity, len(tr))
+		require.Equal(t, savedTries[len(savedTries)-capacity:], tr)
```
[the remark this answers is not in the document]
fxamacker: Added comment in 9b9d6f3bb96b3276d6442dcd4656ecfb3a392e0a.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	c.lm.OnStop(func() {
+		// Signal Compactor goroutine to stop
+		doneCh := make(chan struct{})
+		c.stopCh <- doneCh
+
+		// Wait for Compactor goroutine to stop
+		<-doneCh
+
+		// Shut down WAL component.
+		<-c.wal.Done()
```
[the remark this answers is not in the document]
fxamacker: > should Compactor stop WAL?

Excellent question which requires more than a simple yes/no.

Unless I'm mistaken, node components used to shutdown in sequence but that appears to have been changed to shutdown in parallel.

`WAL` used to be a separate component, at the same level as `Ledger` and `Compactor`.  Node components used to shutdown in reverse sequence so `WAL` shuts down after `Compactor` and `Ledger`.  However, I recently discovered node component shutdown are parallel (not sequential anymore).  Components receive signal and shuts down in parallel, which can cause WAL component to close segment file before all writes are done.

Given this, I removed `WAL` as an independent component, so it needs to be shutdown by either `Ledger` or `Compactor`.  It would be safer for `Compactor` to shutdown `WAL` after all writes are done because `Compactor` drains and processes remaining trie update (writing to WAL segment) in channel after receiving shutdown signal.

> Its supplied as a parameter so I think its lifecycle should be managed by the creator, not here.

Yeah, I was reluctant for the same reason and I used to have `Ledger` shutdown `WAL` in earlier commit.  But that requires communication between `Ledger` and `Compactor` during shutdown because `Compactor` may need to drain and process remaining trie update.  

Thoughts? @m4ksio @ramtinms @zhangchiqing

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	// Compute next checkpoint number.
+	// nextCheckpointNum is updated when
+	// - checkpointing starts, fails to start, or fails.
+	// - tries snapshot fails.
+	// NOTE: next checkpoint number must >= active segment num.
+	// We can't reuse mtrie state to checkpoint tries in older segments.
+	nextCheckpointNum := lastCheckpointNum + int(c.checkpointDistance)
+	if activeSegmentNum > nextCheckpointNum {
```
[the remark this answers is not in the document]
fxamacker: > isn't this block based on assumption that we always catch new segment creation, which might not always be right?

We always catch new segment creation (or the first update in the new segment) when Compactor is running (in the for loop).  But during shutdown, Compactor can write to segment and we don't track segment creation at that point.  

> I would expect to have some form of tracking previous and current last segment to find the trigger point.

The trigger point is:
- finalization of active segment, or 
- finalization of future segment (computed by last checkpoint number + checkpoint distance). 

In other words, the trigger point (next checkpoint number) must >= active segment number.

The earliest trigger point is the finalization of active segment so snapshot of tries in queue match records in finalized segments.

In the initialization block (highlighted code), if next checkpoint number is behind active segment number (maybe caused by EN shutdown before last checkpointing is completed), the next trigger point would be the finalization of active segment to ensure that tries in queue are in sync with finalized segments.

https://github.com/onflow/flow-go/blob/2127e44c0c86cf6af86ae967e6eba50ab99262df/ledger/complete/compactor.go#L154-L157

module/chunks/chunkVerifier_test.go
fxamacker: > do we need this for this test?

Yes.  We need a compactor running when `Ledger.Set` is called, even if we're not testing checkpointing or WAL segments.  `Ledger.Set` communicates with Compactor over channels.  If compactor isn't running, `Ledger.Set` would block.  

In old implementation, an anonymous goroutine is launched for each update, which calls `wal.RecordUpdate()`.  With `NoopWAL` used in tests, no WAL update is written.

In this PR, the anonymous goroutine is replaced by communicating with `Compactor` over channels.  `NoopCompactor` is used in tests for minimum communication.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+	// RecordUpdate returns the segment number the record was written to.
+	// Returned segment number (>= 0) can be
+	// - the same as previous segment number (same segment), or
+	// - incremented by 1 from previous segment number (new segment)
+	segmentNum, skipped, updateErr := c.wal.RecordUpdate(update.Update)
+
+	// This ensures that updated trie matches WAL update.
+	defer func(updateResult error) {
+		// Send result of WAL update
+		update.ResultCh <- updateResult
```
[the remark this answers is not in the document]
fxamacker: Good catch! :+1:  It can be called right away.  Done in commit f048d62d80f92f765a788822376009d686cd6e7d.

cmd/bootstrap/run/execution_state.go
fxamacker: Good catch!  Fixed in commit c5455bd07435c7664565e5c169f6e1d601ecf827.

ledger/complete/mtrie/forest.go
fxamacker: Yes because `AddTrie()` would be a noop for trie that is already in the forest:

https://github.com/onflow/flow-go/blob/2127e44c0c86cf6af86ae967e6eba50ab99262df/ledger/complete/mtrie/forest.go#L386-L388

Larger code block for more context:

https://github.com/onflow/flow-go/blob/2127e44c0c86cf6af86ae967e6eba50ab99262df/ledger/complete/mtrie/forest.go#L374-L395

ledger/partial/ledger_test.go
fxamacker: If ledger doesn't have update activities, we don't need a compactor running.  Otherwise, `Ledger.Set` would block without compactor running.

fxamacker:
> I think we need a test for a case @zhangchiqing has described and that we discussed - namely multiple concurrent Sets on a ledger and a test ensuring the order is preserved.  It seems like code would behave this way, but explicit test is always best

@m4ksio Thanks for focusing on missing tests.  :+1: 

Concurrent Sets are already tested in [`TestCompactorConcurrency`](https://github.com/onflow/flow-go/blob/c5455bd07435c7664565e5c169f6e1d601ecf827/ledger/complete/compactor_test.go#L441).  The test launches 10 concurrent Sets in 3 batches and tests tries replayed from generated checkpoints match tries replayed from generated segments only (ignoring checkpoint files).

https://github.com/onflow/flow-go/blob/c5455bd07435c7664565e5c169f6e1d601ecf827/ledger/complete/compactor_test.go#L507-L512

I'll make sure we also test the order is preserved as more tests are updated/added.

ledger/complete/wal/wal.go
fxamacker: Hmm, not sure I understand the question.  Can you rephrase or elaborate?

ledger/complete/mtrie/forest.go
```diff
@@ -190,19 +190,36 @@ func (f *Forest) Read(r *ledger.TrieRead) ([]ledger.Value, error) {
...
+	if err != nil {
+		return ledger.RootHash(hash.DummyHash), fmt.Errorf("adding updated trie to forest failed: %w", err)
+	}
+
+	return t.RootHash(), nil
+}
+
+// NewTrie updates the Values for the registers and returns updated trie and error (if any).
```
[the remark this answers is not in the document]
fxamacker: > better to update the comment in both Update and NewTrie to highlight the difference, which is that NewTrie doesn't add the root node to the cache.

Good catch!  :+1:  Yeah, I agree we should highlight the difference more in the comments.  
Added comment in commit 240520d97949a6aecf4faa8f8ae5b1324822d139.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,407 @@
...
+	// Update activeSegmentNum
+	prevSegmentNum := activeSegmentNum
+	activeSegmentNum = segmentNum
+
+	if nextCheckpointNum > prevSegmentNum {
+		// Not enough segments for checkpointing
+		return activeSegmentNum, -1, nil
+	}
+
+	// In the remaining code: nextCheckpointNum == prevSegmentNum
```
zhangchiqing, line 393: Is it possible that `prevSegmentNum > nextCheckpointNum`? Especially in the case when checkpoint triggers again before the previous checkpoint finishes? Just to double check.

If this is possible, we might want to use `prevSegmentNum` to trigger checkpoint. Either way, I think it's always safer to use `prevSegmentNum` than `nextCheckpointNum`.

ledger/complete/compactor.go
zhangchiqing: [Made this PR to explain the idea](https://github.com/onflow/flow-go/pull/2878/files)

ledger/complete/checkpointqueue_test.go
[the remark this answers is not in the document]
fxamacker: > To ensure all old elements are overwritten.

No, the change isn't needed because all old elements are overwritten.
 
The tests overrides all old elements because previous `for` loop fills up the queue and this `for` loop overrides all elements in the queue.

ledger/complete/compactor.go
fxamacker: > :warning:  we should check updateErr before adding the trie to checkpoint queue, otherwise might cause inconsistency.

No, it is not needed.

`Ledger.Set()` checks `updateErr` and closes `TrieCh` channel without sending trie if `updateErr` is not nil.  When `Compactor` receives nil trie from `TrieCh`, `Compactor` logs it and nil trie isn't pushed to queue.

https://github.com/onflow/flow-go/blob/2127e44c0c86cf6af86ae967e6eba50ab99262df/ledger/complete/compactor.go#L343-L347

> And if there is updateError, we probably don't have to wait for <-update.TrieCh, because we anyway won't add it to the checkpoint queue

I think we should wait for `<-update.TrieCh` because it is `Ledger.Set` responsibility to send a trie matching WAL update.  And there is another situation that `Ledger.Set` would send a nil trie.  So I don't think we should special case it.

ledger/complete/ledger.go
fxamacker: > better to close the resultCh as well with `defer`

No, `Ledger.Set` shouldn't close `resultCh` with `defer`.  

`Ledger.Set` receives from `resultCh` so it shouldn't close the channel.  It may not happen in this case, but only sender should close channel if needed to avoid `panic` :lady_beetle:   when sending to a closed channel.

On the other hand, `Ledger.Set` sends to `trieCh` so it can close the channel.

ledger/complete/compactor.go
fxamacker: @zhangchiqing  I took a look at your PR and it looks like a refactor that doesn't fix a bug or blocker.  At a glance, it looks like there are some aspects that would require time for closer examination (the devil is in the details and we don't want to introduce bugs that are missed by existing tests).

Given the priority and urgency of this PR, I prefer refactorings of this size are done after we merge this PR.  🙏 

I think some of your suggestions are good and we should give them the time they deserve without rushing.

ledger/complete/compactor.go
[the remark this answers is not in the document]
zhangchiqing: Agree with @fxamacker . 

Make sense to me to stop the WAL here. Better to add some comment to explain
```suggestion
		// only shut down wal after compactor has been shut down, in case there 
		// is still some goroutine writing to WAL files. 
		<-c.wal.Done()
```

ledger/complete/compactor.go
zhangchiqing: Oh, I see. the trie channel is closed when fails to write WAL.

ledger/complete/compactor.go
zhangchiqing: sounds good.

zhangchiqing (approved):
Great work! Looks good to me.

ledger/complete/checkpointqueue.go
fxamacker: Good idea!   Renamed `CheckpointQueue` to `TrieQueue` as suggested in https://github.com/onflow/flow-go/pull/2792/commits/853ff0545636870a21393258e81c986dccdbfc4a.  

Since the name TrieQueue is not specific to checkpointing, also moved it from `complete` package to `wal` package.

Putting it in the `wal` package makes it more clear `TrieQueue` is specific to checkpointing and can prevent it from being used for other purposes, getting modified, and introducing unwanted side-effects to checkpointing.

ledger/complete/compactor.go
```diff
@@ -0,0 +1,393 @@
...
+			} else {
+				// Failed to get semaphore because checkpointing is running.
+				// Try again when active segment is finalized.
+				c.logger.Info().Msgf("compactor delayed checkpoint %d because prior checkpointing is ongoing", nextCheckpointNum)
+				nextCheckpointNum = activeSegmentNum
+			}
+		}
+	}
+
+	// Drain and process remaining trie updates in channel.
```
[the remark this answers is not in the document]
fxamacker: Done in commit 7e29a67c6d7199d66dd0f29785f05868339d1f94.

ledger/complete/compactor.go
fxamacker: > Is it possible that prevSegmentNum > nextCheckpointNum? 

No.  It is an invariant that `nextCheckpointNum >= activeSegmentNum`.  Because `prevSegmentNum` is initialized to `activeSegmentNum` when active segment is finalized and becomes previous segment, therefore `nextCheckpointNum >= prevSegmentNum`.

> Especially in the case when checkpoint triggers again before the previous checkpoint finishes?

When checkpoint triggers again before the previous checkpoint finishes, we fail to acquire semaphore and this block is executed so that checkpointing is retried when active segment is finalized.
https://github.com/onflow/flow-go/blob/240520d97949a6aecf4faa8f8ae5b1324822d139/ledger/complete/compactor.go#L225-L226

ledger/complete/compactor.go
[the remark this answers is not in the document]
m4ksio: Ok, we should definitely have an ability to do synchronized shutdowns, but this PR certainly isn't a place to dwell over that, and this solution is fine by me

ledger/complete/compactor.go
fxamacker: > Does this function need to return checkpointNum?

Not exactly, but I return it anyway as `checkpointNum` along with `tries` to checkpoint because the variable is called `nextCheckpointNum` in the caller and it can be confusing.  Before doing this, I thought about renaming `nextCheckpointNum` in the caller but chose this way because it seemed easier to read this way.

PR #2799, opened 2022-07-13
[FVM] merge constant size account registers in account status 

### LINKED ISSUE #2702: [Execution Node] Reduce size and growth rate of mtrie by merging 5 kinds of FVM account payloads to reduce payloads per account
## Problem

About 5 FVM account payloads are created for every account which can cause mtrie to grow quickly when large number of accounts are created.

## Proposed Solution

Merge 5 kinds of FVM account payloads to reduce the number of payloads per account.  
- `exists`
- `storage_used`
- `storage_index`
- `frozen`
- `public_key_count`

Updates #1744 
Updates #2502 (This PR merges `exists` and `frozen` payloads among other improvements)

Thanks @ramtinms for making time to sync about this on Friday and confirming approach! :+1: 

## Misc

Breakdown of FVM account payloads:
- Every account has 3 FVM account payloads: `exists`, `storage_used`, `storage_index`.
- ~99% accounts have `public_key_count` payload, and each public key is stored in a separate payload under the key of `public_key_` +  index.
- Some accounts have `contract_names` payload, and each contract is stored in a separate payload under the key of `code.` + contract name.
- Frozen account has `frozen` payload.

ramtinms:
This PR 
- closes #2702 
- merges `storage_used`, `storage_index`, and `public_key_count` into the account status register, which reduces about 30M registers from the Mainnet and reduces proof sizes accordingly 
- updates the previous account status migration to cover these registers besides `frozen` and `exists` registers


fvm/accounts_test.go
```diff
@@ -1284,7 +1284,7 @@ func TestAccountBalanceFields(t *testing.T) {
 
 				assert.NoError(t, err)
 				assert.NoError(t, script.Err)
-				assert.Equal(t, cadence.UFix64(9999_2710), script.Value)
+				assert.Equal(t, cadence.UFix64(9999_3120), script.Value)
```
ramtinms, line 1287: storage is now taking less space, so more balance is expected.

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,97 @@
...
+// currently modeled as a byte slice with ondemand decoding
+// the first byte captures flags (e.g. frozen)
+// the next 8 bytes (big endian) captures storage used by an account
+// the next 8 bytes (big endian) captures storage index of an account
+// and the last 8 bytes (big endian) captures number of public keys stored on this account
+// if len of this byte slice is zero, account doesn't exist
+type AccountStatus []byte
 
 const (
-	maskExist  byte = 0b0000_0001
```
ramtinms, line 12: we don't need this flag any more, considering we use fixed size slices

fvm/state/accounts.go
```diff
@@ -255,15 +246,20 @@ func (a *StatefulAccounts) SetPublicKey(
 }
 
 func (a *StatefulAccounts) SetAllPublicKeys(address flow.Address, publicKeys []flow.AccountPublicKey) error {
+
+	count := uint64(len(publicKeys)) // len returns int and this will not exceed uint64
+
+	if count >= MaxPublicKeyCount {
+		return errors.NewAccountPublicKeyLimitError(address, count, MaxPublicKeyCount)
+	}
```
ramtinms, line 254: first checking the size before doing key addition.

fxamacker (approved):
Looks great! :+1:  Just some minor comments.

fxamacker:
After you mentioned slice vs array idea today, I looked into trade-off between `type AccountStatus []byte` and `type AccountStatus [AccountStatusSize]byte`.

The performance and allocations are similar if:
- `AccountStatusFromBytes(inp []byte)` converts slice `inp` to array pointer without make a copy
- `AccountStatus` functions use pointer receivers, such as `func (a *AccountStatus) AccountExists() bool`

The benefit of using array is to prevent creating `AccoutStatus` from byte slice of any length, such as
```
	b := []byte{1}
	acct := AccountStatus(b)
```

To be safer, I think we can use array as `AccountStatus`.

janezpodhostnik (approved):
Looks good to me (besides the things that Faye mentioned).

I am slightly concerned with the introduction of a `MaxPublicKeyCount`. Its definitely nice to have this control if we need it, but if we choose to use it we should be aware that keys cannot currently be removed, only revoked, which means that we need to know how we are going to help accounts that have Max-1 revoked keys, 1 good key and they want to switch that last key with a different one.

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,110 @@
...
+const (
+	// AccountStatusSize returns the byteslice size of an account status
+	AccountStatusSize = 1 + // flags
+		8 + // storage used
+		8 + // storage index
+		8 // public key counts
+
+	storageUsedStartIndex     = 1
+	storageIndexStartIndex    = 1 + 8
+	publicKeyCountsStartIndex = 1 + 8 + 8
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,110 @@
...
+}
+
+// StorageUsed returns the storage used by the account
+func (a AccountStatus) StorageUsed() uint64 {
+	return binary.BigEndian.Uint64(a[storageUsedStartIndex:storageIndexStartIndex])
+}
+
+// SetStorageIndex updates the storage index of the account
+func (a AccountStatus) SetStorageIndex(index atree.StorageIndex) {
+	copy(a[storageIndexStartIndex:publicKeyCountsStartIndex], index[:8])
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,110 @@
...
+}
+
+// StorageIndex returns the storage index of the account
+func (a AccountStatus) StorageIndex() atree.StorageIndex {
+	var index atree.StorageIndex
+	copy(index[:], a[storageIndexStartIndex:publicKeyCountsStartIndex])
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,110 @@
...
+// AccountExists returns true if account exists
 func (a AccountStatus) AccountExists() bool {
-	return a > 0
+	return len(a) > 0
 }
 
+// IsAccountFrozen returns true if account's frozen flag is set
 func (a AccountStatus) IsAccountFrozen() bool {
-	return uint8(a)&maskFrozen > 0
+	return a[0]&maskFrozen > 0
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,110 @@
...
 }
 
-func SetAccountStatusFrozenFlag(inp AccountStatus, frozen bool) AccountStatus {
+// SetFrozenFlag sets the frozen flag
+func (a AccountStatus) SetFrozenFlag(frozen bool) {
 	if frozen {
-		return AccountStatus(uint8(inp) | maskFrozen)
+		a[0] = a[0] | maskFrozen
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts_status_test.go
```diff
@@ -14,17 +16,45 @@ func TestAccountStatus(t *testing.T) {
...
+		returnedIndex := s.StorageIndex()
+		require.True(t, bytes.Equal(index[:], returnedIndex[:]))
+		require.Equal(t, uint64(34), s.PublicKeyCount())
+
+		// check no side effect on flags
+		require.True(t, s.IsAccountFrozen())
+	})
+
+	t.Run("test serialization", func(t *testing.T) {
+		clone, err := state.AccountStatusFromBytes(s.ToBytes())
```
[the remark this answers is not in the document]
ramtinms: updated at 2fc41bc

fvm/state/accounts.go
```diff
@@ -152,22 +146,19 @@ func (a *StatefulAccounts) Exists(address flow.Address) (bool, error) {
...
 	}
 
-	storageUsedByStorageUsed := uint64(RegisterSize(newAddress, KeyStorageUsed, make([]byte, uint64StorageSize)))
-	err = a.setStorageUsed(newAddress, storageUsedByStorageUsed)
-	if err != nil {
-		return err
-	}
+	accountStatus := NewAccountStatus()
+	accountStatus.SetStorageUsed(uint64(RegisterSize(newAddress, KeyAccountStatus, accountStatus.ToBytes())))
+	accountStatus.SetStorageIndex(atree.StorageIndex{0, 0, 0, 0, 0, 0, 0, 1})
```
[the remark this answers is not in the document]
ramtinms: I mostly ported this logic from the older version of create new accounts, even if its not doing anything new might be helpful to keep it here for more readability.

fvm/state/accounts.go
```diff
@@ -567,19 +568,21 @@ func (a *StatefulAccounts) DeleteContract(contractName string, address flow.Addr
...
+func (a *StatefulAccounts) setAccountStatus(address flow.Address, status AccountStatus) error {
+	err := a.SetValue(address, KeyAccountStatus, status.ToBytes())
+	if err != nil {
+		return fmt.Errorf("failed to store the account status for account (%s): %w", address.String(), err)
 	}
-	return binary.BigEndian.Uint64(input[:8]), input[8:], nil
+	return nil
 }
 
 func (a *StatefulAccounts) GetAccountFrozen(address flow.Address) (bool, error) {
```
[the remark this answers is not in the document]
ramtinms: good idea, updated at 150f1fe

ramtinms:
@janezpodhostnik I kept MaxPublicKeyCount and set it to maxUint64 to prevent overflowing for now and added to the TPM agenda if we want to have any lower value or concerns we have about it.

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,112 @@
...
+// AccountStatusFromBytes constructs an AccountStatus from the given byte slice
+func AccountStatusFromBytes(inp []byte) (*AccountStatus, error) {
+	var as AccountStatus
+	if len(inp) != accountStatusSize {
+		return &as, errors.NewValueErrorf(hex.EncodeToString(inp), "invalid account status size")
 	}
-	if len(inp) > 1 {
-		return 0, errors.NewValueErrorf(hex.EncodeToString(inp), "invalid account state")
+	copy(as[:], inp)
+	return &as, nil
```
fxamacker, line 65: Maybe we can avoid making a copy by using input's underlying array.  And we don't need `as` created earlier.

```suggestion
	return (*AccountStatusArray)(inp), nil
```

fvm/state/accounts_status.go
```diff
@@ -1,51 +1,112 @@
...
 )
 
-// NewAccountStatus sets exist flag and return an AccountStatus
-func NewAccountStatus() AccountStatus {
-	return AccountStatus(maskExist)
+// NewAccountStatus returns a new AccountStatus
+func NewAccountStatus(storageIndex atree.StorageIndex) *AccountStatus {
+	var as AccountStatus
+	as.SetStorageIndex(storageIndex)
+	return &as
```
fxamacker, line 46: Not sure, but if we only use default storage index for new account status, maybe we can just hard code the default index inside `NewAccountStatus`, instead of passing it as parameter.

fxamacker (approved):
Nice work!  Just a couple more minor comments.

fvm/state/accounts_status.go
ramtinms: I was not sure, I was worried we might update the registers directly in this case and if we get rid of deep copy feature of DB we might have some issues? what do you think ?

fvm/state/accounts_status.go
ramtinms: I thought about it but felt that's an account-level responsibility and not something account status should know, though from a performance perspective that might save use some copy time.

fvm/state/accounts_status.go
fxamacker: >I was worried we might update the registers directly in this case

I think it's OK if `AccountStatusFromBytes` is only called to create account status by reading:
- payload value from mtrie, or
- cached payload value, previously retrieved from mtrie.

Because [a copy of payload value is returned from mtrie](
https://github.com/onflow/flow-go/blob/01fae8ea3b6658715880aa014866e9b8b61233bb/ledger/complete/mtrie/forest.go#L135), I think we update the copy.

>if we get rid of deep copy feature of DB we might have some issues

Yes but getting rid of deep copy of the payload value can be problematic even without this change.  I think it would be safer to get rid of deep copy of the payload key.

fvm/state/accounts_status.go
ramtinms: updated at 58e5e9b - you're right if we never call it with different input it can have a default value built in.

fvm/state/accounts_status.go
ramtinms: Another case I still not sure, is the case we rollback, we drop the delta but the underlying read cache on view is invalid.

ramtinms: let's keep this optimization for another PR when we revisit all allocations and possible ways to not copy or deep copy.

PR #2800, opened 2022-07-13
backport execution context check from ScriptEnv to TransactionEnv

pattyshack:
prep work for deduping shared environment code

Note that I've inlined the checkContext method since golang doesn't do any
inlining optimization.

### DIFF
--- a/fvm/transactionEnv.go
+++ b/fvm/transactionEnv.go
@@ -1,6 +1,7 @@
 package fvm
 
 import (
+	"context"
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
@@ -53,6 +54,7 @@ type TransactionEnv struct {
 	txID             flow.Identifier
 	traceSpan        opentracing.Span
 	authorizers      []runtime.Address
+	executionCtx     context.Context
 }
 
 func NewTransactionEnvironment(
@@ -93,6 +95,9 @@ func NewTransactionEnvironment(
 		txIndex:          txIndex,
 		txID:             tx.ID(),
 		traceSpan:        traceSpan,
+		// NOTE: To prevent non-deterministic transaction execution,
+		// transactionEnv's executionCtx must never cancel / timeout.
+		executionCtx: context.Background(),
 	}
 
 	env.contracts = handler.NewContractHandler(accounts,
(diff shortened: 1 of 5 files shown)


fvm/transactionEnv.go
```diff
@@ -735,6 +738,24 @@ func (e *TransactionEnv) GenerateUUID() (uint64, error) {
...
+	// we might revisit this when optimizing script execution
+	// by only checking on specific kind of meterComputation calls.
+	//
+	// TODO: in the future this context check should be done inside the cadence
+	select {
+	case <-e.executionCtx.Done():
+		err := e.executionCtx.Err()
+		if errors.Is(err, context.DeadlineExceeded) {
+			return errors.NewExecutionTimedOutError()
+		}
```
ramtinms, line 753: TxEnv should **not** support deadline through context, as it could result in non-deterministic execution on different ENs and result in execution forks. And since EN would almost never cancel the execution of a transaction, it would impact the performance a bit as well.  Any execution limitation outside of computation, storage interaction and memory metering currently is almost impossible in the current architecture.

fvm/transactionEnv.go
```diff
@@ -93,6 +95,7 @@ func NewTransactionEnvironment(
 		txIndex:          txIndex,
 		txID:             tx.ID(),
 		traceSpan:        traceSpan,
+		executionCtx:     context.Background(),
```
pattyshack, line 98: setting executeCtx to context.Background() here enables us to reuse the same code for both ScriptEnv and TransactionEnv

SaveTheRbtz: You may want to put a comment about why it is important to use non-cancellable context here.  You may also consider factoring it into a global variable to not allocate it for each transaction.

SaveTheRbtz (approved):
looks like a no-op from code perspective.  I'll leave the final review to @ramtinms.

janezpodhostnik (commented):
Not sure if it makes sense doing this. Like Ramtin mentioned transactions should not be cancelable from the outside. 

I know that the context is `context.Background()`, but why even give this option to any future developers looking at this code.

From a performance perspective, doesn't this also add a very tiny, but unnecessary check to each transaction?

pattyshack:
as a future (or at least recent) developer who just joined, I can tell you that it's super confusing to have two almost two identical code path running in parallel.  If you can maintain all the variants in your head, you're much smarter than me.  I would gladly tradeoff a tiny bit of performance gain for code clarity.

fvm/transactionEnv.go
```diff
@@ -735,6 +740,24 @@ func (e *TransactionEnv) GenerateUUID() (uint64, error) {
 }
 
 func (e *TransactionEnv) meterComputation(kind common.ComputationKind, intensity uint) error {
```
janezpodhostnik, line 742: This is called very often.

I'm not certain of the time complexity of doing `select { case <-e.executionCtx.Done(): ...` every-time for transactions where this is never actually needed in a transaction...

Can I suggest packaging this into a separate function. Maybe something like:
```
func (e *TransactionEnv) meterComputation(kind common.ComputationKind, intensity uint) error {
  err := e.ExecutionTerminationError()
  if err != nil {
      return err
  }
  //...
}
```

The Transaction `ExecutionTerminationError` would just be a noop

pattyshack:
fine.  i'll do it another way

PR #2804, opened 2022-07-14
Invoke contracts directly instead of via currying

pattyshack:
the closures cost cpu/memory and doesn't provide any benefit


pattyshack:
go compiler does virtually zero optimization.  that's why it runs in seconds rather than minutes/hours (e.g., rust)

bluesign:
Oh this is strange, some weekend research for me, thanks @pattyshack

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #2807, opened 2022-07-15
[BFT Testing] Corrupting Ingress Traffic (part-1)

yhassanzadeh13:
This PR is part 1 of empowering our BFT testing framework to simulate also the corruption of the ingress traffic of the corrupted nodes. To do that, we implement a corruptible `Network` that is a wrapper around the original Flow network and is capable of relaying the ingress (and egress) traffic of the corrupted nodes to a remotely connected attacker through gRPC. The corruptible `Network` also wraps the registered engines (aka message processors) into corruptible `MessageProcessors`. At the current phase, a corruptible `MessageProcessor` just passes through the ingress traffic of the engine to the engine itself. However, in the next part of this development, the corruptible `MessageProcessor` will be furnished with the capability of relaying the ingress traffic of the node to the remote attacker.


network/p2p/network.go
```diff
@@ -99,44 +99,46 @@ type registerBlobServiceResp struct {
...
-	top network.Topology,
-	sm network.SubscriptionManager,
-	metrics module.NetworkMetrics,
-	identityProvider id.IdentityProvider,
-	receiveCache *netcache.ReceiveCache,
-	options ...NetworkOptFunction,
-) (*Network, error) {
-
-	mw, err := mwFactory()
+func NewNetwork(param *NetworkParameters) (*Network, error) {
```
yhassanzadeh13, line 119: Reducing the coupling of network constructor by introducing a parameter object.

network/p2p/conduit/conduit.go
```diff
@@ -5,32 +5,18 @@ import (
...
-	cm := component.NewComponentManagerBuilder().
-		AddWorker(func(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
-			ready()
-
-			<-ctx.Done()
-		}).Build()
-
-	d.ComponentManager = cm
-
-	return d
```
yhassanzadeh13, line 33: Conduit Factory no longer is going to run a `gRPC` server, we instead moved the server to `CorruptibleNetwork`, hence refactoring `ConduitFactory` from a `Component` to a `Module` (that no longer needs to be `ReadyDoneAware`).

insecure/corruptible/messageProcessor.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/onflow/flow-go/model/flow"
+	flownet "github.com/onflow/flow-go/network"
+)
+
+type MessageProcessor struct {
+	logger            zerolog.Logger
+	originalProcessor flownet.MessageProcessor // original message processor
+}
+
+var _ flownet.MessageProcessor = &MessageProcessor{}
```
gomisha, line 17: Why keep this if we don't use it? For ingress message processing?

insecure/corruptible/network.go
```diff
@@ -0,0 +1,370 @@
...
+	"github.com/onflow/flow-go/utils/logging"
+)
+
+// Network is a wrapper around the original flow network, that allows a remote attacker
+// to take control over its ingress and egress traffic flows.
+// A remote attacker can register itself to this corruptible network.
+// Whenever any corruptible conduit receives an event from its engine, it relays the event to this
+// network, which in turn is relayed to the register attacker.
+// The attacker can asynchronously dictate to the network to send messages on behalf of the node.
+type Network struct {
```
gomisha, line 39: we have `insecure/corruptible/network.go` and `insecure/attacknetwork/attackNetwork.go` - it's confusing to have 2 corruptible / insecure networks. Can we rename `insecure/attacknetwork/attackNetwork.go` so it's less confusing?

insecure/corruptible/network.go
```diff
@@ -0,0 +1,370 @@
...
+	// TODO: we can dissolve CCF and instead have a decorator pattern to turn a conduit into
+	// a corrupted one?
+	conduit, err := n.flowNetwork.Register(channel, corruptibleProcessor)
+	if err != nil {
+		return nil, fmt.Errorf("could not register corruptible message processor on channel: %s, %w", channel, err)
+	}
+	return conduit, nil
+}
+
+func (n *Network) RegisterBlobService(channel channels.Channel, store datastore.Batching, opts ...flownet.BlobServiceOption) (flownet.BlobService,
```
gomisha, line 120: can we add comments for all normal network methods we have to implement that we are NOT corrupting anything but just passing everything through to the flow network? So it's clear what we're changing and what we're leaving as-is

insecure/corruptible/network.go
```diff
@@ -0,0 +1,370 @@
...
+	error) {
+	return n.flowNetwork.RegisterBlobService(channel, store, opts...)
+}
+
+func (n *Network) RegisterPingService(pingProtocolID protocol.ID, pingInfoProvider flownet.PingInfoProvider) (flownet.PingService, error) {
+	return n.flowNetwork.RegisterPingService(pingProtocolID, pingInfoProvider)
+}
+
+func (n *Network) ProcessAttackerMessage(stream insecure.CorruptibleConduitFactory_ProcessAttackerMessageServer) error {
```
gomisha, line 129: Can we add a comment here about this being the main place messages are processed by the corrupted network and how it's called by gRPC?

insecure/corruptible/network.go
```diff
@@ -0,0 +1,370 @@
...
+
+	n.mu.Unlock()
+	n.logger.Info().Msg("attacker registered successfully")
+
+	// WARNING: this method call should not return through the entire lifetime of this
+	// corruptible conduit factory.
+	// This is a client streaming gRPC implementation, and the input stream's lifecycle
+	// is tightly coupled with the lifecycle of this function call.
+	// Once it returns, the client stream is closed forever.
+	// Hence, we block the call and wait till a component shutdown.
```
gomisha, line 325: can we add a comment to the top of the function definition that it needs to be called in a goroutine because it's a blocking call? Otherwise, it's not clear that's how it needs to be used.

Or is it better to have this function start the goroutine here so clients don't have to do that?

yhassanzadeh13: Added comment here: https://github.com/onflow/flow-go/pull/2807/commits/0fec3510d37b49e925a061dd2dc8ec09c1ae74e4

insecure/corruptible/network.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2807/commits/1df4164215782d40e304157a533259d36bc53236

insecure/corruptible/network.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2807/commits/cb1644fda89e97c3e29d4d77033a34aaa9757e79

insecure/corruptible/network.go
yhassanzadeh13: I think they are self-explanatory in the terms of interfaces they implement; the `Network` implements the `flownetwork.Network` (as we explicitly added `var _ flownet.Network = &Network{}`) which is the network layer of the node (godoc). On the other hand `AttackNetwork` implements `insecure.AttackNetwork`, which is the networking interface available to the attacker (godoc). We have several similar naming pairs across our codebase as well, and I think as long as we set the interface and godoc clear, it is easy to track them. Nevertheless, if you have any better naming suggestion please let me know, and we can consider renaming them. 

P.S: following your comment, I added `var _ insecure.AttackNetwork = &AttackNetwork{}` explicit interface conversion for attack network in https://github.com/onflow/flow-go/pull/2807/commits/fd0aa7a8e82b46bf4751eb7dddcbc336572d051b.

insecure/corruptible/messageProcessor.go
yhassanzadeh13: This is an engineering best practice in Golang, that we explicitly indicate what interfaces a struct is implementing. We do that through a dummy variable `_` that signals the compiler that it is intentionally of no usage, hence avoiding compile error. In this scenario, we explicitly tell whoever reads the code that this `MessageProcessor` `struct` is supposed to implement the `flownet.MessageProcessor` `interface`. Moreover, if any breaking changes happen that this interface implementation is violated, this line complains of a compile-time error, hence quickly identifying the root cause.

gomisha (approved):
LGTM - thank you for adding all the additional comments / clarifications.

peterargue (approved):
looks good. sorry for the delay

Issue #2797, opened 2022-07-13
[State Sync] Improve ExecutionDataRequester stability when peers become unavailable


peterargue:
## Problem Definition

The `ExecutionDataRequester` uses [bitswap](https://docs.ipfs.io/concepts/bitswap/) to download execution data from Execution nodes and other Access nodes. This is live on testnet and mainnet. However, there have been some issues when peers leave the network (e.g. reboots, networking issues, etc) where all nodes suddenly are no longer able to find certain execution data blobs. After restarting, the nodes will always catch up. Typically, only one AN needs to be restarted, then all the others become unstuck.

Talking with the bitswap devs, this is most likely because the bitswap client only detects peers that connect after the bitswap network was started on the node. So any nodes that connected before are never added creating a fragmented network.

## Proposed Solution

Update the bootstrapping logic to start the bitswap network before connecting to peers.

## Definition of Done

Bitswap is started before the node connects to any peers, allowing it to have a full view of the network.

franklywatson:
Duplicates: https://github.com/onflow/flow-go/issues/2718

Issue #2718, opened 2022-06-28
State Sync: fix bitswap block request getting stuck


franklywatson:
We need to investigate the reason why the requester gets stuck sometimes and isn't able to progress until the node is restarted.

One possibility is that the way we instantiate Bitswap is incorrect, in that we start it only after we've already connected to some peers. We should fix this in the code, and see if it solves the issue. However, if it doesn't, we need to investigate more to find out what the true cause is.

PR #2808, opened 2022-07-15
[Network] Refactor libp2pNode and peer manager into components

peterargue:
Closes: #2797 
Closes: #2718 

This is the first PR in a 2 part change.

## Background
Access and Execution nodes run bitswap to sync `ExecutionData` amongst themselves. Bitswap has a limitation where it will only discover peers that connect after it has been started. This means we need instantiate bitswap before connecting to any peers.

## Changes
This PR refactors `p2p.Node` and `p2p.PeerManager` so that their lifecycle can be controlled by the node builder. That's accomplished with the following changes
* Convert libp2pNode into a component
* Instantiate libp2pNode as a Component in scaffold, and inject into middleware
* Convert peer manager into a component
* Move peer manager from middleware into libp2pNode
* Expose peer manager’s component interface via libp2pNode, and make it a scaffold Component

It then adds a new `DependableComponent` type to scaffold that allows defining a component with an explicit list of `ReadyDoneAware` dependencies. `DependableComponent`s will wait for all of their dependencies to be ready before starting (instead of starting serially)


cmd/access/node_builder/access_node_builder.go
```diff
@@ -676,9 +677,13 @@ func (builder *FlowAccessNodeBuilder) Initialize() error {
...
 
-	builder.AdminCommand("get-transactions", func(conf *cmd.NodeConfig) commands.AdminCommand {
-		return storageCommands.NewGetTransactionsCommand(conf.State, conf.Storage.Payloads, conf.Storage.Collections)
-	})
+	builder.
+		AdminCommand("get-transactions", func(conf *cmd.NodeConfig) commands.AdminCommand {
+			return storageCommands.NewGetTransactionsCommand(conf.State, conf.Storage.Payloads, conf.Storage.Collections)
+		}).
+		AdminCommand("read-execution-data", func(conf *cmd.NodeConfig) commands.AdminCommand {
+			return stateSyncCommands.NewReadExecutionDataCommand(builder.ExecutionDataService)
```
smnzhu, line 685: what is this for?

network/p2p/middleware.go
```diff
@@ -311,23 +311,25 @@ func (m *Middleware) start(ctx context.Context) error {
...
+}
+
 // stop will end the execution of the middleware and wait for it to end.
 func (m *Middleware) stop() {
+	// TODO: does middleware need to stop the peer manager before the network?
+	// if not, we don't need to wait here since it will be stopped as a regular component
 	mgr, found := m.peerMgr()
 	if found {
-		// stops peer manager
+		// wait for peer manager to stop
```
smnzhu, line 332: Will this be a problem? This may cause the middle ware and network stopping to block forever, since it waits for peer manager to stop, but peer manager stoppage is never initiated?

network/p2p/peerManager.go
```diff
@@ -82,35 +101,45 @@ func PeerManagerFactory(peerManagerOptions []Option, connectorOptions ...Connect
...
-func (pm *PeerManager) updateLoop() {
 	for {
 		select {
-		case <-pm.peerRequestQ:
-			pm.updatePeers()
-		case <-pm.unit.Quit():
+		case <-ctx.Done():
+			return
+		default:
+		}
```
smnzhu, line 136: nit: I don't think this select is really necessary

network/p2p/middleware.go
peterargue: yea, this could cause a problem if the middleware was ever stopped independently of the node. @yhassanzadeh13 do you know if the peer manager needs to be stopped before the libp2p node or does it not matter?

cmd/access/node_builder/access_node_builder.go
peterargue: Right now that's only available on ENs. Adding it to ANs so we can get access to execution data there too. It came up as useful when attempting to backfill some missing data from gcp

network/p2p/middleware.go
yhassanzadeh13: @peterargue ideally the `PeerManager` should be stopped _before_ `LibP2PNode`. Otherwise, `PeerManager` may attempt on connecting to peers through libp2p's `Host` (which is encapsulated into `LibP2PNode`). If `LibP2PNode` has already been stopped, this connection attempt results in an error log. As this log is internal to libp2p, we don't have control over its level or content. So, except for the possibility of posting a misleading log, it seems there is no severe consequences. However, with the current state of architecture, it seems a safer approach to move `PeerManafger` as a member-field of the `LibP2PNode` instead of `Middleware` and let `LibP2PNode` handle the lifecycle of the `PeerManager` (i.e., `Start` and stop). It makes more sense as `PeerManager` does not seem as a dependency on `Middleware`. We can move running [`updateLoop`](https://github.com/onflow/flow-go/blob/master/network/p2p/peerManager.go#L107) into a separate `ComponentManager` worker, and then let `LibP2PNode` handle its lifecycle.

yhassanzadeh13: Revising my comment: 

In the continuation of the concern posed at the `network/middleware.go` and also the comment above, it makes more sense to treat the `PeerManager` as a member field of the `LibP2PNode` (than `Middleware`) but with a different lifecycle management, i.e., we can expose `PeerManager`'s `Component` behavior as an exportable method on `LibP2PNode`: 

```
func (n *Node) PeerManagerComponent() component.Component {
	return n.peerManager
}
```

In this way, we can still enqueue the `PeerManager` as a component independently in `scafold` but without introducing any anti-pattern. 

Moreover, by moving `PeerManager` from `Middleware` to the `LibP2PNode` we yield a cleaner architecture regarding Separation of Concerns, as most of the dependencies of `PeerManager` belong to the `LibP2PNode` and in the current state of affair, `Middleware` is only in charge of the lifecycle management of the `PeerManager`. 

Also, it is fine to let `PeerManager` stop independently of the `Middleware`. As far as I checked, it only causes an internal error log on the libp2p's side when it attempts to update the peers while the `LibP2PNode` is down. As we do not expose the internal logger of the libp2p, it should not cause concern. So the `Middleware` stop logic will no longer need to check for the termination of the `PeerManager`.

yhassanzadeh13 (changes requested):
Thanks for the proposed change, my concern is though regarding the long-term architectural flexibility and maintainability that this change introduces. Added the concerns and also a proposed solution in different comments. Would be happy to discuss further.

network/p2p/libp2pNode.go
```diff
@@ -283,6 +288,31 @@ func (n *Node) WithDefaultUnicastProtocol(defaultHandler libp2pnet.StreamHandler
 	return nil
 }
 
+func (n *Node) WithPeersProvider(peersProvider PeersProvider) error {
+	if n.peerManagerFactory == nil {
+		return nil
```
yhassanzadeh13, line 293: not an error case?

yhassanzadeh13 (approved):
Thanks for addressing the comments, looks good in general, added a few minor comments.

cmd/scaffold.go
```diff
@@ -1377,6 +1390,11 @@ func (fnb *FlowNodeBuilder) onStart() error {
 
 	fnb.EnqueueAdminServerInit()
 
+	// Peer Manager must be started after all other components to ensure the network does not start
+	// up before the components are initialized. This is critical for any nodes running bitswap since
+	// bitswap will only learn about peers that connect after it is initialized.
+	fnb.EnqueuePeerManager()
+
```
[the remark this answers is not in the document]
peterargue: this was removed in 55e4f1e, I think you may have reviewed using an old commit.

In this PR, I'm going to keep the existing startup order, and address it in the dependable component PR

network/p2p/middleware.go
```diff
@@ -177,7 +165,9 @@ func NewMiddleware(
 			ready()
 
 			<-ctx.Done()
-			mw.stop()
+
+			// wait for the readConnection and readSubscription routines to stop
+			mw.wg.Wait()
```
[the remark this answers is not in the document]
peterargue: I don't think we should to add a one-off timeout here since that's not a pattern we use anywhere else for shutdowns (afaik).

Components will all begin to shutdown in parallel when the node's context is cancelled, so worst case, the network hangs on shutdown while other components complete, and is eventually killed by the operator.

I'm not opposed to adding some shutdown timeout logic, but I think it should be added to the node builder framework.

cmd/access/node_builder/access_node_builder.go
```diff
@@ -938,41 +943,61 @@ func (builder *FlowAccessNodeBuilder) Build() (cmd.Node, error) {
...
+				return nil, err
+			}
+
+			builder.AccessNodeConfig.PublicNetworkConfig.Network = net
+
+			node.Logger.Info().Msgf("network will run on address: %s", builder.PublicNetworkConfig.BindAddress)
+			return net, nil
+		}).
+		Component("peer manager", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+			// TODO: make this component start after its dependencies are ready. This will require a dependable type of component
```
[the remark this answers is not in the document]
peterargue: this comment is just here to signal my intention in the follow up PR to make this dependable. I plan to merge both together

network/p2p/libp2pNode.go
peterargue: No, it's an expected case. In the previous implementation, not all middleware instances had a `peerManagerFactory`. This replicates that behavior.

That's currently used by observers/consensus followers, which don't use a peer manager.

smnzhu:
Otherwise looks okay

admin/command_runner_test.go
```diff
@@ -66,21 +76,17 @@ func (suite *CommandRunnerSuite) SetupCommandRunner(opts ...CommandRunnerOption)
...
 	suite.runner.Start(signalerCtx)
 	<-suite.runner.Ready()
-	go func() {
-		select {
-		case <-ctx.Done():
-			return
-		case err := <-errChan:
-			suite.Fail("encountered unexpected error", err)
-		}
-	}()
```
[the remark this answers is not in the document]
peterargue: I switched it to use `go unittest.NoIrrecoverableError()`

admin/command_runner_test.go
```diff
@@ -25,19 +26,22 @@ import (
...
 type CommandRunnerSuite struct {
 	suite.Suite
 
-	runner       *CommandRunner
-	bootstrapper *CommandRunnerBootstrapper
-	httpAddress  string
+	runner          *CommandRunner
+	bootstrapper    *CommandRunnerBootstrapper
+	httpAddress     string
+	grpcAddressSock string
```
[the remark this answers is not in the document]
peterargue: these tests were failing frequently during local testing. Seemed to be an issue with socket file collisions, possibly because the file wasn't being cleaned up.

engine/access/rpc/backend/connection_factory_test.go
```diff
@@ -429,10 +430,16 @@ func (n *node) setupNode(t *testing.T) {
...
+	// using a wait group here to ensure the goroutine has started before returning. Otherwise,
+	// there's a race condition where the server is sometimes stopped before it has started
+	wg := sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
+		wg.Done()
 		err := n.server.Serve(n.listener)
 		assert.NoError(t, err)
 	}()
+	wg.Wait()
```
yhassanzadeh13, line 442:
```suggestion
	unittest.RequireReturnsBefore(t, wg.Wait, 10 * time.Millisecond, "could not start goroutine on time")
```

cmd/scaffold.go
```diff
@@ -975,6 +1008,16 @@ func (fnb *FlowNodeBuilder) handleComponents() error {
 
 		parent = started
 	}
+
+	// Components with explicit dependencies are run asynchronously, which means dependencies in
+	// the dependency list must be initialized outside of the component factory.
+	for _, f := range asyncComponents {
+		err = fnb.handleComponent(f, util.AllReady(f.dependencies...), func() {})
+		if err != nil {
+			return err
```
yhassanzadeh13, line 1017:
```suggestion
			return fmt.Errorf("could not handle component %s: %w", f.name, err)
```

network/p2p/connection/peerManager.go
```diff
@@ -61,35 +71,45 @@ func PeerManagerFactory(connectionPruning bool, updateInterval time.Duration) p2
...
+		case <-ctx.Done():
+			return
+		default:
+		}
 
-	// makes sure that peer update request is invoked once before returning
-	pm.RequestPeerUpdate()
+		select {
+		case <-ctx.Done():
+			return
```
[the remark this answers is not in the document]
peterargue: since peer manager is managed as a top level component, it will inherit the initialization, startup, and shutdown log messages.

admin/command_runner_test.go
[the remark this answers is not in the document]
smnzhu: I still don't know what you mean

smnzhu (approved):
Cool! Honestly mostly looks good, although I haven't looked through the tests yet.

yhassanzadeh13 (approved):
Reviewed the recent changes, and they look good, thanks for applying my comments.

admin/command_runner_test.go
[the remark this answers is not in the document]
peterargue: I actually refactored since that last comment. Turns out using a separate goroutine is unreliable for capturing the error. Instead, there's a new mocked `SignallerContext` implementation that doesn't use a separate goroutine and instead fails the test directly when `Throw` is called. See `irrecoverable.NewMockSignalerContext`

peterargue:
bors merge

PR #2809, opened 2022-07-15
chore(comment): fix IsSampled description

SaveTheRbtz:
In future I'll likely rewrite that in terms of percents or percentage points so it is more intuitive.

### DIFF
--- a/model/flow/identifier.go
+++ b/model/flow/identifier.go
@@ -77,8 +77,11 @@ func (id Identifier) Format(state fmt.State, verb rune) {
 // IsSampled is a utility method to sample entities based on their ids
 // the range is from [0, 64].
 // 0 is 100% (all data will be collected)
-// 32 is ~50%
-// 64 is ~0% (no data will be collected)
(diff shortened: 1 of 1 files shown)


pattyshack (approved):

model/flow/identifier.go
```diff
@@ -77,7 +77,10 @@ func (id Identifier) Format(state fmt.State, verb rune) {
 // IsSampled is a utility method to sample entities based on their ids
 // the range is from [0, 64].
 // 0 is 100% (all data will be collected)
-// 32 is ~50%
+// 1 is ~50%
+// 2 is ~25%
+// 3 is ~12.5%
+// ...
 // 64 is ~0% (no data will be collected)
```
pattyshack, line 84: o, while you're at it.  change `64 is ~0%` to `> 64 is 0%`.  off by one error =P

ramtinms (approved):

SaveTheRbtz:
bors merge

PR #2814, opened 2022-07-18
upgrade btcd dependency

vikstrous2:
There was a breaking change in btcd and this PR upgrades that dependency so users of flow-go can also upgrade their dependency on btcd.

To make things work, I also had to add replace directives so that the current version of the crypto sub-module is used. I also had to upgrade github.com/libp2p/go-libp2p-core to v0.16.1


tarakby (commented):
Thanks for update! 
I suggested a different way to make the update below.

tarakby:
Hi @vikstrous2, we're planning to make a similar package upgrade to the repo. I wanted to check if you're planning to make changes to the PR sometime soon or if I should go ahead and update it. Thanks

vikstrous2:
Sorry, I totally forgot about this PR. I made these to like 10 repos before I ran out of steam. Feel free to just merge it in 2 parts yourselves if you have the bandwidth. I'll come back to this again later when all our dependencies are done upgrading.

tarakby:
Closing the current PR, dependencies update is done by https://github.com/onflow/flow-go/pull/3085 and another PR to follow up soon.

PR #2817, opened 2022-07-18
[Integration] Splitting Cadence files from go source

Tonix517:
Moving out all inlined Cadence source from .go file into separate files.

#### Test Plan
- [X] launch localnet and run loader with different load types. Monitor Grafana and tx traces to make sure everything works.

### DIFF
--- a/integration/utils/scripts.go
+++ b/integration/utils/scripts.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	_ "embed"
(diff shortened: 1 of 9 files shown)


pattyshack (approved):

SaveTheRbtz (approved):
thank you for refactoring this and making our syntax highlighter / autocomplete happy!

PR #2822, opened 2022-07-19
Improve error reporting for access api event endpoints

peterargue:
Improve status codes returned from `GetEvent*` Access API endpoints
* Return `NotFound` instead of `Internal` when the block wasn't found when searching for events
* Return `InvalidArgument` instead of `Internal` when block range size is too large


engine/access/rpc/backend/backend_transactions.go
```diff
@@ -66,6 +66,7 @@ func (b *backendTransactions) SendTransaction(
 	// store the transaction locally
 	err = b.transactions.Store(tx)
 	if err != nil {
+		// TODO: why would this be InvalidArgument?
 		return status.Error(codes.InvalidArgument, fmt.Sprintf("failed to store transaction: %v", err))
```
peterargue, line 70: Any idea why this is `InvalidArgument` instead of `Internal`?

koko1123: I thought this was to differentiate it on the client side? let me double check

koko1123 (approved):

engine/access/rpc/backend/backend_transactions.go
peterargue: It seems odd to return `InvalidArguement` if the tx is actually valid and the error was writing it to the db.

Thinking about this some more, I don't think we should return an error at all. As is the client may think the TX failed when it really succeeded, which could result in a double payment.

lolpuddle (approved):

peterargue:
bors merge

PR #2828, opened 2022-07-19
[WIP] [FVM] optimizing account status operation

ramtinms:


ramtinms:
closing it for now, re-opening later when we start optimizing deep copies of the ledger.

PR #2829, opened 2022-07-20
fixed decoding error for signer indices at epoch switchover

AlexHentschel:
* follow up to PR https://github.com/onflow/flow-go/pull/2789
* changes in this PR are identical to fix for `master`: https://github.com/onflow/flow-go/pull/2830

## Description of revisions:
For block `44f..a`, the access API fails to decode the signer indices:
```
could not decode signer indices for block 44ff570d66aeac087da261e7bac13247fa96bbb0792ee65cf8714e3632da034a: signer indices' checkum is invalid: decoder sees a canonical list [...], which has a different checksum eb94d50f than the encoder's checksum ef2c778c: index vector's checksum is invalid
```

I think it the culprit is this line in the code:
https://github.com/onflow/flow-go/blob/01fae8ea3b6658715880aa014866e9b8b61233bb/consensus/hotstuff/signature/block_signer_decoder.go#L38-L39
* because a block includes the signatures for the _parent_
* but we are retrieving the entities that are allowed to sign the block itself (not the parent)

This breaks exactly at epoch switchover

### DIFF
--- a/consensus/hotstuff/signature/block_signer_decoder.go
+++ b/consensus/hotstuff/signature/block_signer_decoder.go
@@ -35,15 +35,16 @@ func (b *BlockSignerDecoder) DecodeSignerIDs(header *flow.Header) (flow.Identifi
 		return []flow.Identifier{}, nil
 	}
 
-	id := header.ID()
-	members, err := b.Identities(id)
+	// The block header contains the signatures for the parents. Hence, we need to get the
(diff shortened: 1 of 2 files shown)


zhangchiqing (approved):

tarakby (approved):
Good catch!

peterargue (approved):

Issue #2834, opened 2022-07-20
[Access Node] Add support for disabling connection pooling


peterargue:
## Problem Definition

Access nodes have a connection pool, and a flag `--connection-pool-size` which controls how many cached connections to maintain. However, there's no way to disable the pool in the event that it's causing issues for the node.

## Proposed Solution

Allow setting `--connection-pool-size` to `0` to indicate that connections should not be cached. This should then open new connections each time one is retrieved from the connection factory.

PR #2836, opened 2022-07-20
AN - Option to Disable Conn Pooling

lolpuddle:
Closes https://github.com/onflow/flow-go/issues/2834


engine/access/rpc/backend/connection_factory.go
```diff
@@ -200,7 +214,7 @@ func (s *CachedClient) Close() {
 		return
 	}
 	// allow time for any existing requests to finish before closing the connection
-	time.Sleep(s.timeout)
+	time.Sleep(s.timeout + 1*time.Second)
```
koko1123, line 217: why do we need this additional second?

engine/access/rpc/engine.go
```diff
@@ -132,21 +132,21 @@ func NewBuilder(log zerolog.Logger,
 	// wrap the unsecured server with an HTTP proxy server to serve HTTP clients
 	httpServer := NewHTTPServer(unsecureGrpcServer, config.HTTPListenAddr)
 
-	// TODO: when cache size is set to 0, handle case where we do not use cache
+	var cache *lru.Cache
 	cacheSize := config.ConnectionPoolSize
-	if cacheSize == 0 {
-		cacheSize = backend.DefaultConnectionPoolSize
```
koko1123, line 138: you may need to change the value of `DefaultConnectionPoolSize` to `0` so we can now disable connection pooling by default

koko1123 (commented):
Looks good overall, couple of comments to address and we should be well on our way!

engine/access/rpc/backend/connection_factory.go
lolpuddle: Not the most elegant solution but I wanted to ensure that there was allotted time for existing requests to finish OR timeout before a connection is closed. It made sense for the delay in closing to be greater than the timeout time vs just being equal

engine/access/rpc/engine.go
```diff
@@ -132,21 +132,26 @@ func NewBuilder(log zerolog.Logger,
...
-		log.Debug().Str("grpc_conn_evicted", store.Address).Msg("closing grpc connection evicted from pool")
-		if accessMetrics != nil {
-			accessMetrics.ConnectionFromPoolEvicted()
+	if cacheSize > 0 {
+		// it was observed that eviction issues were occurring when the pool size was less than number of ENs + LNs
+		if cacheSize < 249 {
+			log.Debug().Msg("defaulting the connection pool size to 250")
+			cacheSize = 250
+		}
+		var err error
```
koko1123, line 143: I think the idea was to have `DefaultConnectionPoolSize` now be 250, and set the value to that

engine/access/rpc/engine.go
```diff
@@ -132,21 +132,26 @@ func NewBuilder(log zerolog.Logger,
...
-	cache, err := lru.NewWithEvict(int(cacheSize), func(_, evictedValue interface{}) {
-		store := evictedValue.(*backend.CachedClient)
-		store.Close()
-		log.Debug().Str("grpc_conn_evicted", store.Address).Msg("closing grpc connection evicted from pool")
-		if accessMetrics != nil {
-			accessMetrics.ConnectionFromPoolEvicted()
+	if cacheSize > 0 {
+		// it was observed that eviction issues were occurring when the pool size was less than number of ENs + LNs
+		if cacheSize < 249 {
+			log.Debug().Msg("defaulting the connection pool size to 250")
```
koko1123, line 140: might want to change this to something like `connection pool size below threshold, defaulting to 250`

cmd/access/node_builder/access_node_builder.go
```diff
@@ -591,9 +591,6 @@ func (builder *FlowAccessNodeBuilder) extraFlags() {
 				return errors.New("execution-data-max-search-ahead must be greater than 0")
 			}
 		}
-		if builder.rpcConf.ConnectionPoolSize == 0 {
```
peterargue, line 594: can you add a note in the flag description about the default value and meaning of 0?

koko1123 (approved):
looks fine to me, will have to incorporate the flag description change per @peterargue

engine/access/rpc/backend/backend_accounts.go
```diff
@@ -153,10 +153,14 @@ func (b *backendAccounts) getAccountFromAnyExeNode(ctx context.Context, execNode
...
 
 func (b *backendAccounts) tryGetAccount(ctx context.Context, execNode *flow.Identity, req execproto.GetAccountAtBlockIDRequest) (*execproto.GetAccountAtBlockIDResponse, error) {
-	execRPCClient, err := b.connFactory.GetExecutionAPIClient(execNode.Address)
+	execRPCClient, closer, err := b.connFactory.GetExecutionAPIClient(execNode.Address)
 	if err != nil {
 		return nil, err
 	}
+	if closer != nil {
+		defer closer.Close()
+	}
```
peterargue, line 162:
```suggestion
	defer closer.Close()
```

engine/access/rpc/backend/connection_factory.go
```diff
@@ -138,46 +139,70 @@ func (cf *ConnectionFactoryImpl) retrieveConnection(grpcAddress string, timeout
...
 	}
 
-	conn, err := cf.retrieveConnection(grpcAddress, cf.CollectionNodeGRPCTimeout)
+	var conn *grpc.ClientConn
+	if cf.ConnectionsCache != nil {
+		conn, err = cf.retrieveConnection(grpcAddress, cf.CollectionNodeGRPCTimeout)
+		if err != nil {
+			return nil, nil, err
+		}
+		return access.NewAccessAPIClient(conn), nil, err
```
peterargue, line 155: here you could return a noop closer:
```golang
type noopCloser struct{}
func (c *noopCloser) Close() {}
```

```suggestion
		return access.NewAccessAPIClient(conn), &noopCloser{}, err
```

that reduces that amount of boilerplate logic you need in all the callers since we always return a real closer implementation

engine/access/rpc/engine.go
```diff
@@ -132,21 +132,26 @@ func NewBuilder(log zerolog.Logger,
...
-		cacheSize = backend.DefaultConnectionPoolSize
-	}
-	cache, err := lru.NewWithEvict(int(cacheSize), func(_, evictedValue interface{}) {
-		store := evictedValue.(*backend.CachedClient)
-		store.Close()
-		log.Debug().Str("grpc_conn_evicted", store.Address).Msg("closing grpc connection evicted from pool")
-		if accessMetrics != nil {
-			accessMetrics.ConnectionFromPoolEvicted()
+	if cacheSize > 0 {
+		// it was observed that eviction issues were occurring when the pool size was less than number of ENs + LNs
```
peterargue, line 138:
```suggestion
		// TODO: remove this fallback after fixing issues with evictions
		// it was observed that evictions cause connection errors for in flight requests. This works
		// around the issue by forcing the pool size to be greater than the number of ENs + LNs
```

peterargue (commented):
looks good. added a few small comments. Is there test coverage of the case where caching is disabled?

peterargue (approved):
this looks good. my previous comments are nice to have, but don't need to block this going out

engine/access/rpc/backend/connection_factory_test.go
```diff
@@ -88,7 +79,7 @@ func TestProxyExecutionAPI(t *testing.T) {
 	}
 
 	// get an execution API client
-	client, err := proxyConnectionFactory.GetExecutionAPIClient("foo")
+	client, _, err := proxyConnectionFactory.GetExecutionAPIClient("foo")
```
peterargue, line 82: please close the noop closer in at least one of these tests to ensure it works as expected

peterargue (approved):
looks great. thanks @lolpuddle

PR #2837, opened 2022-07-20
[FVM] Storage check optimisation

janezpodhostnik:
Make one storage check for all account at once instead of one per account, reducing the number of ContractFunctionInvocations per transaction. 

ref: https://github.com/dapperlabs/flow-go/issues/6308

Depends on cadence runtime.Interface changes from https://github.com/onflow/cadence/pull/1811

# Results

The improvement can be seen below in the **FVM [Benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) comparison**.

All of the benchnet scenarios below check storage of 3 or 4 accounts which is also the most common case on Mainnet.

The touched accounts are:
- Payer
- FlowFees
- FlowToken
- in case it is a transfer, both the receiver and a sender (but in these bench tests the receiver is always the Payer)

The performance improvement on shorter transactions like `RuntimeTransaction/reference_tx-2` is about 15%. Tapering off on long running transactions, where a lot of other things also happens during the transaction. On `RuntimeTransaction/transfer_tokens-2` (which does 100 token transfers in 1 transaction between the same 2 users) the benefit is only ~2.5%.

There is one exception, which is `RuntimeTransaction/create_new_account` this runs about 8% faster even though it is a very long running transaction. The reason for this is that it creates 50 accounts, that all need to be checked for storage in addition to the 3 default ones.


pattyshack:
looks like you're doing two independent things in the same PR (bumping cadence version & the actual optimization).  could you split out the cadence change into a different pr?  thx

fvm/contractFunctionInvocations.go
```diff
@@ -150,6 +149,39 @@ func AccountStorageCapacityInvocation(
...
 }
 
+// AccountsStorageCapacityInvocation prepares a function that calls get storage capacity on the storage fees contract
+// for multiple accounts at once
+func AccountsStorageCapacityInvocation(
+	env Environment,
+	traceSpan opentracing.Span,
+) func(addresses []common.Address) (cadence.Value, error) {
+	return func(addresses []common.Address) (cadence.Value, error) {
+		arrayValues := make([]cadence.Value, len(addresses))
```
Tonix517: can we check nil or empty of `addresses` here to save one invocation if no valid address passed in?

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,89 @@ package fvm
...
+	resultArray, ok := result.(cadence.Array)
+	if !ok {
+		return fmt.Errorf("storage limit check failed: AccountsStorageCapacity did not return an array")
+	}
+
+	for i, value := range resultArray.Values {
+		capacity := storageMBUFixToBytesUInt(value)
 
-		usage, err := env.GetStorageUsed(commonAddress)
+		usage, err := env.GetStorageUsed(commonAddresses[i])
```
pattyshack, line 82: there's no point in invoking the account storage capacity contract if this fails.  maybe collect the usages first in a separate loop (and early exit on error), then invoke the contract and check capacity in this loop.

Tonix517:
Does existing Benchstat numbers include expected metrics we want? If not, could you publish the metrics before\after this optimization? thanks.

janezpodhostnik:
@Tonix517 Benchstat numbers do already include the performance comparison here. I added a section to the PR description

fvm/contractFunctionInvocations.go
janezpodhostnik: I am not sure if I want to put this logic here. This is just a wrapper to call the smart contract function. Any conditions like that should be either in the smart contract or better yet (performance wise) in the function calling this one. 

I will check that the function calling this one doesn't do it with an empty array.

What do you think?

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,100 @@ package fvm
...
+	}
+
+	commonAddresses := make([]common.Address, len(addresses))
+	for i, address := range addresses {
+		commonAddresses[i] = common.Address(address)
+	}
 
-		capacity, err := env.GetStorageCapacity(commonAddress)
+	usages := make([]uint64, len(commonAddresses))
+	for i, address := range commonAddresses {
```
pattyshack, line 46: combine this loop with the one above?

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,100 @@ package fvm
...
+		u, err := env.GetStorageUsed(address)
 		if err != nil {
 			return fmt.Errorf("storage limit check failed: %w", err)
 		}
+		usages[i] = u
+	}
+
+	n := len(addresses)/TransactionStorageLimiterScriptArgumentBatchSize + 1
+	for i := 0; i < n; i++ {
```
pattyshack, line 55: maybe simplify to
```
for start := 0; start < n; start += TransactionStorageLimiterScriptArgumentBatchSize {
  end := start + TransactionStorageLimiterScriptArgumentBatchSize
  if end > len(addresses) {
    end = len(addresses)
  }
  ...
}
```

pattyshack (approved):

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,96 @@ package fvm
...
+		if end > len(addresses) {
+			end = len(addresses)
+		}
+
+		err := d.batchCheckLimits(
+			env,
+			commonAddresses[start:end],
+			usages[start:end],
+			span,
+		)
```
ramtinms, line 63: any concern here that we break it into smaller batches ?

janezpodhostnik: I am somewhat concerned that sending too many arguments at once might inflate the memory costs. But the limit I set may be a bit low on second though... should be more around 10_000 or something

I just don't want it to be limitless really.

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,96 @@ package fvm
...
+	// This error only occurs in case of implementation errors. The InvokeAccountsStorageCapacity
+	// already handles cases where the default vault is missing.
+	if invokeErr != nil {
+		return invokeErr
+	}
+
+	// the resultArray elements are in the same order as the addresses and the addresses are deterministically sorted
+	resultArray, ok := result.(cadence.Array)
+	if !ok {
+		return fmt.Errorf("storage limit check failed: AccountsStorageCapacity did not return an array")
```
ramtinms, line 88: this would be a fatal error by default, otherwise we have to define a new error type and just do it that way to gracefully handle the case, thoughts?

ramtinms (approved):
Overall looks good had only two comments/concerns/questions.

fvm/transactionStorageLimiter.go
ramtinms: but isn't it anyway it might fail because of the memory ?

fvm/transactionStorageLimiter.go
janezpodhostnik: IMO, this is fatal. It means the contract function does not return an array. The only way that can happen is through implementation error.

SaveTheRbtz (approved):

fvm/transactionStorageLimiter.go
janezpodhostnik: I did some more thinking. 

My fear was that you could create an attack where you would touch so many accounts that getting the storage for all of them at once would OOM. 

But I think this problem might not be possible due to the fact that touching accounts requires computation, and even if it is, batching might not be the right approach to solve it. 

I have deleted the batching part to reduce complexity.

fvm/transactionStorageLimiter.go
```diff
@@ -3,42 +3,96 @@ package fvm
...
 
-func NewTransactionStorageLimiter() *TransactionStorageLimiter {
-	return &TransactionStorageLimiter{}
+func NewTransactionStorageLimiter() TransactionStorageLimiter {
+	return TransactionStorageLimiter{}
 }
 
-func (d *TransactionStorageLimiter) CheckLimits(
+// TransactionStorageLimiterScriptArgumentBatchSize is the number of accounts to check at once.
+// Using to many arguments at once might cause problems during InvokeAccountsStorageCapacity.
```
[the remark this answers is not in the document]
SaveTheRbtz: nvm, seems like you've removed batching.

SaveTheRbtz:
#2853 has landed.  You can rebase on top of master now.

bluesign:
maybe naive question, but why don't you filter this list with storage_used ? You should be able to skip a lot of checks where storage_used < N ( N = minimimAccountBalance granted storage ~ 100Kb)

janezpodhostnik:
@bluesign the minimum account balance is defined in the state, so getting it requires a call to GetValue.

A GetValue call is expensive, but not as expensive as a **Contract function invocation**. With the previous approach fetching minimum account balance once would definitely make sense, but with this approach (only one  **Contract function invocation**) I'm not sure it does. In some cases (the failure ones) this would mean that the check is faster, but the happy path would be a bit longer because of it.

janezpodhostnik:
bors merge

PR #2838, opened 2022-07-20
Fix loader unknown TX stat bug

simonhf:
This is a hacky fix to get the latest loader code to not report bogus unknown TXs. Although this fix "works"... am wondering if there is a better way? There is a small discussion here [1] and this PR was requested as a way of commenting on the hacky fix and/or possible better / alternative fix :-)

[1] https://axiomzen.slack.com/archives/C024LNQE5N1/p1658201790504759

### DIFF
--- a/integration/utils/follower.go
+++ b/integration/utils/follower.go
@@ -128,9 +128,20 @@ Loop:
 		getBlockByHeightDuration := time.Since(GetBlockByHeightTime)
 
 		var blockTxs, blockUnknownTxs uint64
+
+		for _, guaranteed := range block.CollectionGuarantees {
+			_, err := f.client.GetCollection(f.ctx, guaranteed.CollectionID)
+			if err != nil {
+				continue Loop // silently try again if not ready; can happen several dozen times
+			}
+		}
+
 		for _, guaranteed := range block.CollectionGuarantees {
 			col, err := f.client.GetCollection(f.ctx, guaranteed.CollectionID)
 			if err != nil {
+				f.logger.Error().
+					Uint64("blockHeight", block.Height).
+					Msg("GetCollection() error") // should never happen here
 				continue Loop
 			}
 			for _, tx := range col.TransactionIDs {
@@ -139,6 +150,7 @@ Loop:
 				if txi, loaded := f.loadAndDelete(tx); loaded {
 					duration := time.Since(txi.submisionTime)
 					f.logger.Trace().
+						Uint64("blockHeight", block.Height).
 						Dur("durationInMS", duration).
 						Hex("txID", tx.Bytes()).
 						Msg("returned account to the pool")
@@ -148,6 +160,10 @@ Loop:
 					}
 				} else {
 					blockUnknownTxs++
+					f.logger.Trace().
+						Uint64("blockHeight", block.Height).
+						Hex("txID", tx.Bytes()).
+						Msg("block unknown TX") // should never happen
 				}
 			}
 		}


integration/utils/follower.go
```diff
@@ -128,9 +128,20 @@ Loop:
 		getBlockByHeightDuration := time.Since(GetBlockByHeightTime)
 
 		var blockTxs, blockUnknownTxs uint64
+
+		for _, guaranteed := range block.CollectionGuarantees {
+			_, err := f.client.GetCollection(f.ctx, guaranteed.CollectionID)
+			if err != nil {
+				continue Loop // silently try again if not ready; can happen several dozen times
```
pattyshack, line 135: we don't need to go all the way back to `Loop` (in particular, GetBlockByHeight's result doesn't change)?  we only need to retry getting this specific collection (and maybe double check for f.ctx.Done()).

integration/utils/follower.go
```diff
@@ -128,9 +128,20 @@ Loop:
...
+			}
+		}
+
 		for _, guaranteed := range block.CollectionGuarantees {
 			col, err := f.client.GetCollection(f.ctx, guaranteed.CollectionID)
```
pattyshack, line 140: Instead of iterating over block.CollectionGuarantees and calling getCollection again here, we can just collect the results, in say `map[collectionId]*Collection`, in the above loop

integration/utils/follower.go
```diff
@@ -128,9 +128,20 @@ Loop:
...
 			if err != nil {
```
simonhf, line 141: This is where the unknown TX stat bug occurred before this PR. How? The entire for loop for the block would start dozen or two dozen times, maybe process a collection, but also get into this error branch and error out. However, because some TXs in some collections did get processed and deleted from the map, this means that if an error out occurs after being deleted from the map, then the next attempt to process the block will find the same TX (already deleted from the map) and wrongly count it as an 'unknown TX'...

integration/utils/follower.go
```diff
@@ -128,9 +128,20 @@ Loop:
...
+				f.logger.Error().
+					Uint64("blockHeight", block.Height).
+					Msg("GetCollection() error") // should never happen here
 				continue Loop
```
pattyshack, line 145: This should panic if it should never happen.  This comment may not be relevant.  See above comment.

pattyshack (commented):
waiting for all the collections for a block to become available before checking seems sensible.    a few nit

integration/utils/follower.go
```diff
@@ -148,6 +160,10 @@ Loop:
 					}
 				} else {
 					blockUnknownTxs++
+					f.logger.Trace().
+						Uint64("blockHeight", block.Height).
+						Hex("txID", tx.Bytes()).
+						Msg("block unknown TX") // should never happen
```
SaveTheRbtz, line 166: comment is misleading

integration/utils/follower.go
simonhf: Thanks for the comment! For my taste the loop is complicated enough already, which is probably why this bug crept in in the first place :-) So I'm not keen on doing the loops within loops as suggested, and potentially introducing further bugs. If anything I was hoping somebody might have a suggestion to simplify the entire loop so that my hacky fix wouldn't be necessary :-)

integration/utils/follower.go
simonhf: Thanks for the comment! For me, this suggestions seems to make the code unnecessarily more complicated? I have no problem to scrap this PR and feel free to re-write the code as you see fit. If anything I was hoping somebody might have a suggestion to simplify the entire loop so that my hacky fix wouldn't be necessary and the `continue Loop`s would also not be necessary :-)

integration/utils/follower.go
pattyshack: your comment about the complicated loop is fair.  In that case, I would suggest that we refactor the two inner loops into `GetAllCollections(CollectionGuarantees) ([]Collection, error)` and `CheckTransactions(Block, []Collection) error` to break down the complexity

integration/utils/follower.go
simonhf: > This should panic if it should never happen

Possibly, but in theory the loader can continue so why not continue? Seeing the frequency of this error and then investigating might be more useful than forcing panic upon the first time the error occurs? Presumably this error can only occur due to some kind of concurrent writing to the data structures, anyway? If this error would only happen for 1 in 10,000 blocks then it hasn't done that much harm in the big scheme of things, but a panic would be annoying? Yes, the panic would force examination of the issue, but there are plenty of `.Error()` calls in the code base and they don't need t be panics, or? IMO the loader should somehow monitor for any errors from any nodes and if too many of them occur then abort anyway since it's not a 'clean' load test anymore?

integration/utils/follower.go
simonhf: Thanks for the comment! Why is the comment misleading? Are you expecting `block unknown TX` to show up?

integration/utils/follower.go
Kay-Zee: we could consider just looping here instead:

```
	col, err := f.client.GetCollection(f.ctx, guaranteed.CollectionID)
	for ; err != nil; col, err = f.client.GetCollection(f.ctx, guaranteed.CollectionID) {
		time.Sleep(time.Millisecond)
	}
```

don't know if you have to sleep in the loop, but we'd basically assume we'd always get to a point where we stop erroring. Might have to add a stop loop break after certain number of tries.

Kay-Zee:
yeah, I agree, i think currently, it's possible for the access node to still be waiting for collections itself, so cannot yet serve up the collection to the client until it's received it, in that scenario, then i think this PR isn't that hacky.

PR #2839, opened 2022-07-21
[Execution] a fix to atree reporter

ramtinms:
Controller has been removed, this PR fixes atree reporter
Also adds an extra condition to getAccountStatus to handle non-existing accounts with proper error


janezpodhostnik (approved):
LGTM. 

Was this not covered by tests?

PR #2821, opened 2022-07-18
Remove GO111MODULE=on everywhere

SaveTheRbtz:
Since go 1.16 `GO111MODULE=on` is the default. Since 1.17 `go get` for binaries is deprecated. Since 1.18 it is disabled.

Is there any reason to keep it?

PR #2841, opened 2022-07-21
Execution data format update

smnzhu:
Execution data format update. Changes the structure to separate Chunk Execution Data per chunk, which are combined into a Block Execution Data that serves as a table of contents.

Also replaces Execution Data Service with Downloader (for Access Nodes), which implements the logic for downloading the new format, and Provider (for Execution Nodes), which implements the logic of storing the new format.


peterargue (commented):
I haven't finished reviewing, but here's my comments so far. Overall it looks pretty good. 

As a general note. this PR could use lots more comments. Particularly:
* godocs for exported methods, including notes about expected error conditions and sentinels
* comments within methods outlining the logic. Some of this code is fairly complex. It would make it much easier to review and maintain if the logic was described up front.

I've noted this in a few places, but not everywhere

cmd/execution_builder.go
```diff
@@ -158,8 +160,8 @@ func (e *ExecutionNodeBuilder) LoadFlags() {
 			flags.BoolVar(&e.exeConf.enableBlockDataUpload, "enable-blockdata-upload", false, "enable uploading block data to Cloud Bucket")
 			flags.StringVar(&e.exeConf.gcpBucketName, "gcp-bucket-name", "", "GCP Bucket name for block data uploader")
 			flags.StringVar(&e.exeConf.s3BucketName, "s3-bucket-name", "", "S3 Bucket name for block data uploader")
-			flags.DurationVar(&e.exeConf.edsDatastoreTTL, "execution-data-service-datastore-ttl", 0,
-				"TTL for new blobs added to the execution data service blobstore")
+			flags.Uint64Var(&e.exeConf.executionDataPrunerHeightRangeTarget, "execution-data-height-range-target", 65000, "target height range size used to limit the amount of Execution Data kept on disk")
```
peterargue, line 163: I think the default should be to keep everything (pruning disabled). Especially as we're rolling out execution state sync, we need to make sure that there are always seeders for all blobs. Eventually we should be able to rely on ANs to do that, but I don't think we can yet.

cmd/execution_builder.go
```diff
@@ -400,77 +421,68 @@ func (e *ExecutionNodeBuilder) LoadComponentsAndModules() {
 
 			return compactor, nil
 		}).
-		Component("execution data service", func(node *NodeConfig) (module.ReadyDoneAware, error) {
-			err := os.MkdirAll(e.exeConf.executionDataDir, 0700)
-
+		Component("execution data pruner", func(node *NodeConfig) (module.ReadyDoneAware, error) {
```
peterargue, line 424: is there a way to disable pruning if I wanted to keep everything?

engine/execution/computation/computer/computer.go
```diff
@@ -219,9 +226,46 @@ func (e *blockComputer) executeBlock(
...
+	executionData := &execution_data.BlockExecutionData{
+		BlockID:             res.ExecutableBlock.ID(),
+		ChunkExecutionDatas: make([]*execution_data.ChunkExecutionData, 0, len(collections)+1),
+	}
+
+	for i, collection := range collections {
+		col := collection.Collection()
+		executionData.ChunkExecutionDatas = append(executionData.ChunkExecutionDatas, &execution_data.ChunkExecutionData{
+			Collection: &col,
+			Events:     res.Events[i],
```
zhangchiqing, line 255: ⚠️ I don't think this is correct, a collection might contain multiple events.

engine/execution/computation/computer/computer.go
```diff
@@ -219,9 +226,46 @@ func (e *blockComputer) executeBlock(
...
+			TrieUpdate: res.TrieUpdates[i],
+		})
+	}
+
+	executionData.ChunkExecutionDatas = append(executionData.ChunkExecutionDatas, &execution_data.ChunkExecutionData{
+		Collection: systemCol,
+		Events:     res.Events[len(res.Events)-1],
```
zhangchiqing, line 262: ⚠️ same here, system chunk might produce multiple events.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+	if util.CheckClosed(p.cm.ShutdownSignal()) {
+		return
+	}
+
+	p.fulfilledHeightsIn <- height
+}
+
+// SetHeightRangeTarget updates the Pruner's height range target.
+// This may block for the duration of a pruning operation.
+func (p *Pruner) SetHeightRangeTarget(heightRangeTarget uint64) error {
```
zhangchiqing, line 141: When will we call SetHeightRangeTarget and what is the height target? Is it a certain sealed height or finalized height? I don't find code that calls this methods.

If we are pruning based on a distance from sealed height, have we thought about implementing it [using jobqueue](https://github.com/onflow/flow-go/blob/master/module/jobqueue/README.md)?

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (p *Pruner) checkPrune(ctx irrecoverable.SignalerContext) {
+	if p.lastFulfilledHeight-p.lastPrunedHeight > p.heightRangeTarget+p.threshold {
+		pruneHeight := p.lastFulfilledHeight - p.heightRangeTarget
+
+		p.logger.Info().Uint64("prune_height", pruneHeight).Msg("pruning storage")
+		start := time.Now()
+
+		if err := p.storage.Prune(pruneHeight); err != nil {
```
zhangchiqing, line 191: Does this prune a specific height or up to a height?

If we are pruning specific height, how do we ensure all height below the pruneHeight has been pruned? 
If we are pruning up to the given height, we'd better making it clear by renaming this method to `PruneUpToHeight`.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+			ctx.Throw(fmt.Errorf("failed to prune: %w", err))
+		}
+
+		if err := p.pruneCallback(ctx); err != nil {
```
zhangchiqing, line 195: what is the use case for prune callback given that we didn't pass cid or height to it.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/module/component"
+	"github.com/onflow/flow-go/module/executiondatasync/tracker"
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+)
+
+const (
+	defaultHeightRangeTarget = uint64(400_000)
+	defaultThreshold         = uint64(100_000)
```
zhangchiqing, line 19: These seem to be very large number? Any reasoning why we pick them as is?

Especially the default threshold, this would cause each time to prune 100_000 height, which is a very big number of keys to remove.

I'm thinking about `100`.

cmd/execution_builder.go
```diff
@@ -400,77 +421,68 @@ func (e *ExecutionNodeBuilder) LoadComponentsAndModules() {
...
-							}
-						},
-					),
-				))
+			var prunerMetrics module.ExecutionDataPrunerMetrics = metrics.NewNoopCollector()
+			if node.MetricsEnabled {
+				prunerMetrics = metrics.NewExecutionDataPrunerCollector()
+			}
 
+			executionDataPruner, err = pruner.NewPruner(
```
zhangchiqing, line 449: can we move pruner into a separate PR?

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+// and triggers pruning once the difference between the fulfilled
+// height and the last pruned height reaches the height range
+// target + threshold.
+type Pruner struct {
+	storage       tracker.Storage
+	pruneCallback func(ctx context.Context) error
+
+	// channels used to send new fulfilled heights and config changes to the worker thread
+	// fulfilledHeights uses an unbounded channel to ensure NotifyFulfilledHeight is not blocking
+	fulfilledHeightsIn    chan<- interface{}
```
zhangchiqing, line 41: I found the word `fulfilled` a bit confusing, I think `unpruned` might be easier to understand.

The reason I had confusion is because I thought there were three status: pruned, unpruned, fulfilled. 

I understand `fulfilled` means the height has been executed (fulfilled), but to the pruner's perspective, the only difference is whether a height is pruned or not. If it's not pruned, then it's unpruned. So there are only two status.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+// * Height range target: The target number of most recent blocks
+//   to store data for. This controls the total amount of data
+//   stored on disk.
+// * Threshold: The number of block heights that we can exceed
+//   the height range target by before pruning is triggered. This
+//   controls the frequency of pruning.
+// The Pruner consumes a stream of fulfilled height notifications,
+// and triggers pruning once the difference between the fulfilled
+// height and the last pruned height reaches the height range
+// target + threshold.
```
zhangchiqing, line 34:
```suggestion
// The Pruner tries to preserve at minimum `heightRangeTarget` number of height of data in storage.
// It's notified with unpruned notifications, when there are `heightRangeTarget + threshold` number 
// of unpruned height, it prunes the storage until there are `heightRangeTarget` unpruned height left.
```

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+
+// Pruner is a component responsible for pruning data from
+// execution data storage. It is configured with the following
+// parameters:
+// * Height range target: The target number of most recent blocks
+//   to store data for. This controls the total amount of data
+//   stored on disk.
+// * Threshold: The number of block heights that we can exceed
+//   the height range target by before pruning is triggered. This
+//   controls the frequency of pruning.
```
zhangchiqing, line 30:
```suggestion
//   reduce the frequency of pruning, since pruning is implemented by batch removal,
//   and is more efficient when pruning a big range of height at once.
```

zhangchiqing (commented):
Great work!

I had quite many questions about the pruner. I wonder if possible to move it into a separate PR, given this PR is huge already. 

Ideally, we could keep PR as small as possible. If this is a big feature, we might want to create a feature branch, and making multiple PRs for different parts. Small PR would be easier to review.

module/executiondatasync/execution_data/store.go
```diff
@@ -0,0 +1,248 @@
...
+	return s
+}
+
+func (s *store) AddExecutionData(ctx context.Context, executionData *BlockExecutionData) (flow.Identifier, error) {
+	executionDataRoot := &BlockExecutionDataRoot{
+		BlockID:               executionData.BlockID,
+		ChunkExecutionDataIDs: make([]cid.Cid, len(executionData.ChunkExecutionDatas)),
+	}
+
+	for i, chunkExecutionData := range executionData.ChunkExecutionDatas {
```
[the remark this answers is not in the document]
smnzhu: We could, but it isn't necesary because the Execution Data Store is only used for tests and admin command

module/executiondatasync/execution_data/store.go
```diff
@@ -0,0 +1,248 @@
...
+			blobLen = len(data)
+		}
+
+		blob := blobs.NewBlob(data[:blobLen])
+		data = data[blobLen:]
+		blbs = append(blbs, blob)
+		cids = append(cids, blob.Cid())
+	}
+
+	if err := s.blobstore.PutMany(ctx, blbs); err != nil {
```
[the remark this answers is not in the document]
smnzhu: @zhangchiqing there will be no error, because adding the same blob twice is idempotent.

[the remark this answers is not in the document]
smnzhu: @peterargue it shouldn't be possible for the output of `Serialize` to be empty

Makefile
```diff
@@ -118,46 +118,49 @@ generate-proto:
...
-	mockery --name 'ConnectionFactory' --dir="./engine/access/rpc/backend" --case=underscore --output="./engine/access/rpc/backend/mock" --outpkg="mock"
-	mockery --name 'IngestRPC' --dir="./engine/execution/ingestion" --case=underscore --tags relic --output="./engine/execution/ingestion/mock" --outpkg="mock"
-	mockery --name '.*' --dir=model/fingerprint --case=underscore --output="./model/fingerprint/mock" --outpkg="mock"
-	mockery --name 'ExecForkActor' --structname 'ExecForkActorMock' --dir=module/mempool/consensus/mock/ --case=underscore --output="./module/mempool/consensus/mock/" --outpkg="mock"
-	mockery --name '.*' --dir=engine/verification/fetcher/ --case=underscore --output="./engine/verification/fetcher/mock" --outpkg="mockfetcher"
-	mockery --name '.*' --dir=insecure/ --case=underscore --output="./insecure/mock"  --outpkg="mockinsecure"
-	mockery --name '.*' --dir=./cmd/util/ledger/reporters --case=underscore --output="./cmd/util/ledger/reporters/mock" --outpkg="mock"
+	GO111MODULE=on mockery --name '(Connector|PingInfoProvider)' --dir=network/p2p --case=underscore --output="./network/mocknetwork" --outpkg="mocknetwork"
```
SaveTheRbtz, line 121: can you please remove the `GO111MODULE=on` here?  It was removed in https://github.com/onflow/flow-go/pull/2821

engine/execution/computation/computer/computer.go
smnzhu: It's corrext, `res.Events` is a 2d array

engine/execution/computation/computer/computer.go
smnzhu: same here

module/executiondatasync/execution_data/downloader_test.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+			close(blobCh)
+			return blobCh
+		},
+	)
+
+	_, err = downloader.Download(context.Background(), edID)
+	var blobNotFoundError *execution_data.BlobNotFoundError
+	assert.ErrorAs(t, err, &blobNotFoundError)
+}
```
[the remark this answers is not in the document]
smnzhu: other error conditions should be covered via the handler tests, since they have almost identical code

module/executiondatasync/execution_data/store.go
```diff
@@ -0,0 +1,249 @@
...
+	"fmt"
+
+	"github.com/ipfs/go-cid"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/blobs"
+)
+
+// ExecutionDataStore handles adding / getting execution data to / from a blobstore
+type ExecutionDataStore interface {
```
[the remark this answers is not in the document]
smnzhu: That's right

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+	thresholdChan         chan uint64
+	heightRangeTargetChan chan uint64
+
+	lastFulfilledHeight uint64
+	lastPrunedHeight    uint64
+
+	// the height range is the range of heights between the last pruned and last fulfilled
+	// heightRangeTarget is the target minimum value for this range, so that after pruning
+	// the height range is equal to the target.
+	heightRangeTarget uint64
```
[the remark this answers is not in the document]
smnzhu: The V2 requester takes this into account, and the pruner will work properly with that

module/util/util.go
```diff
@@ -145,3 +145,42 @@ func WaitError(errChan <-chan error, done <-chan struct{}) error {
 		return nil
 	}
 }
+
+// UnboundedChannel returns a pair of channels that represent the two ends of
+// an unbounded channel.
+// The in channel is used to write to the unbounded channel, and never blocks.
+// The out channel is used to read from the unbounded channel.
+func UnboundedChannel() (chan<- interface{}, <-chan interface{}) {
```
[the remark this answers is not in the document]
smnzhu: We're returning a send channel, which can be closed.

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+	if err := txn.Set(latestHeightKey, latestHeightValue); err != nil {
+		return fmt.Errorf("failed to set latest height value: %w", err)
+	}
+
+	return nil
+}
+
+func (s *storage) trackBlobs(blockHeight uint64, cids ...cid.Cid) error {
+	cidsPerBatch := 16
+	maxCidsPerBatch := getBatchItemCountLimit(s.db, 2, blobRecordKeyLength+latestHeightKeyLength+8)
```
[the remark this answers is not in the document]
smnzhu: yes, in theory it could change because the max batch count / size depends on the vlog size, which is configurable during runtime

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+
+			latestHeightValue, err := latestHeightItem.ValueCopy(nil)
+			if err != nil {
+				return fmt.Errorf("failed to retrieve latest height value for Cid %s: %w", blobCid.String(), err)
+			}
+
+			// a blob is only removable if it is not referenced by any blob tree at a higher height
+			latestHeight := binary.LittleEndian.Uint64(latestHeightValue)
+			if latestHeight < blockHeight {
+				// this should never happen
```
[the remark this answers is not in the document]
smnzhu: I think this is self explanatory, by definition latest height >= all block heights

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+				return fmt.Errorf(
+					"inconsistency detected: latest height recorded for Cid %s is %d, but blob record exists at height %d",
+					blobCid.String(), latestHeight, blockHeight,
+				)
+			} else if latestHeight == blockHeight {
+				if err := s.pruneCallback(blobCid); err != nil {
+					return err
+				}
+				dInfo.deleteLatestHeightRecord = true
+			}
```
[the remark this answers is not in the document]
smnzhu: that's why we keep `deleteLatestheightRecord` as false for these, but we still need to delete the blob record for this specific height

module/blobs/blob_store.go
```diff
@@ -0,0 +1,82 @@
...
+
+var ErrNotFound = errors.New("blobstore: blob not found")
+
+type blobstoreImpl struct {
+	bs blockstore.Blockstore
+}
+
+func NewBlobstore(ds datastore.Batching) *blobstoreImpl {
+	return &blobstoreImpl{bs: blockstore.NewBlockstore(ds)}
+}
```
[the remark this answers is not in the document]
smnzhu: Because the terminology is confusing. We want to avoid using the word "block" since that already has a different meaning

cmd/execution_builder.go
smnzhu: No, you would just set the threshold very large

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+	var batch []*deleteInfo
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if err := s.setPrunedHeight(height); err != nil {
+		return err
+	}
+
+	if err := s.db.View(func(txn *badger.Txn) error {
```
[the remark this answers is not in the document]
smnzhu: From the perspective of badger, these will essentially be treated as separate transactions. It has no way to tell that the transactions are being nested.

read only transactions are based off of a snapshot that's taken at the time the transaction is started, so the updates won't impact the read transaction.

module/executiondatasync/pruner/pruner.go
smnzhu: it's useful if the application wants to perform some logic after every pruning call. In particular, we use it to trigger value log GC

module/executiondatasync/pruner/pruner.go
smnzhu: We can use a different number for threshold. The smaller it is, the more frequent the pruning. if we pick 100, that would mean pruning happens approximately every 2 minutes, which seems quite frequent

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+}
+
+func (p *Pruner) loop(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+	ready()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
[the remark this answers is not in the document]
smnzhu: The problem is that if we close it, then `NotifyFulfilledHeight` could panic.

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+			}
+		}
+
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// this is a good time to do garbage collection
+	if err := s.db.RunValueLogGC(0.5); err != nil {
```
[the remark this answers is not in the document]
smnzhu: some data is guaranteed to be removed, as long as the new height is >= the old one

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,510 @@
...
+
+func (s *storage) Prune(height uint64) error {
+	blobRecordPrefix := []byte{prefixBlobRecord}
+	itemsPerBatch := s.batchDeleteItemLimit()
+	var batch []*deleteInfo
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if err := s.setPrunedHeight(height); err != nil {
```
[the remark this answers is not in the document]
smnzhu: It's up to the caller to ensure that the existing pruned height is lower than the new one

module/executiondatasync/pruner/pruner.go
smnzhu: It's used by admin commands to update the height range target.

module/executiondatasync/pruner/pruner.go
smnzhu: up to a height

peterargue (commented):
looks pretty good. mostly requests for comments and tests.

I think we should increase the default `execution-data-height-range-target` to something around a week (800K blocks). that would only take up ~30GB disk space and buy more time in the event there's a bug we need to fix on the requester/bitswap side.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, fmt.Errorf("failed to get pruned height: %w", err)
+	}
+
+	fulfilledHeight, err := storage.GetFulfilledHeight()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get fulfilled height: %w", err)
+	}
+
+	fulfilledHeightsIn, fulfilledHeightsOut := util.UnboundedChannel()
+	fulfilledHeightsIn <- fulfilledHeight
```
jordanschalm, line 102: I have the same reservations about using unbounded channels here that I mentioned in the earlier iteration of this PR. In general I think unbounded channels are an antipattern that we should try to avoid. 

For this case, I think we can instead use a monotonous counter, since we only want to retain the largest fulfilled height we are notified of. 
https://github.com/onflow/flow-go/blob/3eb8166c369f335c0315e42569ccf2af66d9222f/engine/consensus/sealing/counters/monotonous_counter.go#L9

We use this in the sealing engine, and something similar in the Active Pacemaker. 
https://github.com/onflow/flow-go/blob/29c6cdefecc181c7726bac9ebcad9ee9ef3eea67/consensus/hotstuff/voteaggregator/vote_aggregator.go#L33

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,520 @@
...
+
+// PruneCallback is a function which can be provided by the user which
+// is called for each CID when the last height at which that CID appears
+// is pruned.
+// Any returned error will be returned from the surrounding call to Storage.Prune.
+// The prune callback can be used to delete the corresponding
+// blob data from the blob store.
+type PruneCallback func(cid.Cid) error
+
+type Storage interface {
```
jordanschalm, line 125: For each method here, could you note any expected error returns. If there are none, could you add the one-liner
```
No errors are expected during normal operation
```

☝️ this helps to differentiate between:
* the function doesn't have any expected error types
* the function hasn't been updated to document it's expected error types yet

jordanschalm (commented):
First round of comments

module/util/util.go
```diff
@@ -145,3 +145,42 @@ func WaitError(errChan <-chan error, done <-chan struct{}) error {
 		return nil
 	}
 }
+
+// UnboundedChannel returns a pair of channels that represent the two ends of
```
jordanschalm, line 149: I'm wary of making use of unbounded channels in general. For the specific case they were used in this PR, I suggested an alternative [here](https://github.com/onflow/flow-go/pull/2841#discussion_r939451895). 

In general, I think we should avoid using unbounded channels because:
* they can hide performance problems and bugs, especially when the underlying buffer size isn't visible in any way at runtime
* there are usually better options to manage the back-pressure:
  * we can filter/drop the incoming messages (in the use case here, we only ever care about the largest height value in the queue, everything else can be dropped)
  * we can block the message producer

👉  I'm personally inclined to remove this function altogether, especially if we have consensus on the `MonotonousCounter` suggestion to replace the use-case in this PR. Then `UnboundedChannel` would be unused.

engine/execution/ingestion/engine.go
```diff
@@ -641,6 +645,10 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
 	if err != nil {
 		e.log.Err(err).Msg("failed in process block's children")
 	}
+
+	if e.executionDataPruner != nil {
```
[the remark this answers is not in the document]
smnzhu: not really, but allowing it made updating the tests possible / easier. It would have been much more complicated to update the existing tests if it wasn't allowed to be nil.

module/executiondatasync/provider/provider_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	ds := getDatastore()
+	provider := getProvider(getBlobservice(ds))
+	store := getExecutionDataStore(ds)
+
+	test := func(numChunks int, minSerializedSizePerChunk uint64) {
+		expected := generateBlockExecutionData(t, numChunks, minSerializedSizePerChunk)
+		executionDataID, err := provider.Provide(context.Background(), 0, expected)
+		require.NoError(t, err)
+		actual, err := store.GetExecutionData(context.Background(), executionDataID)
```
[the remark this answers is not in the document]
smnzhu: the test doesn't hang, it's just that the comparison `assert.DeepEquals` takes a very very long time.

Also, it doesn't return before blobs are actually stored, because we always wait for the channel to close, which happens after blobs are stored

module/executiondatasync/provider/provider.go
```diff
@@ -0,0 +1,243 @@
...
+			if err := p.blobService.AddBlobs(ctx, blobs); err != nil {
+				return fmt.Errorf("failed to add blobs: %w", err)
+			}
+
+			return nil
+		})
+		duration := time.Since(start)
+
+		if err != nil {
+			ch <- err
```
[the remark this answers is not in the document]
smnzhu: Provide actually doesn't return until all blobs are stored, because it waits for `errCh` to close before returning.

engine/execution/ingestion/engine.go
```diff
@@ -641,6 +645,10 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
...
+		e.executionDataPruner.NotifyFulfilledHeight(executableBlock.Height())
```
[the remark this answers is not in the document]
smnzhu: > Since we persist execution data asynchronously

As mentioned above, execution data is actually persisted synchronously, in that the Provider doesn't return until it has been persisted or an error is encountered.

module/executiondatasync/pruner/pruner.go
smnzhu: I just ended up using a buffered channel for now, because if I used the counter I would still need to add an additional Notifier so that the worker routine could be notified every time there's an update

module/executiondatasync/pruner/pruner.go
smnzhu: I don't think this is clearer, it's just a different way of saying the same thing

cmd/execution_builder.go
```diff
@@ -160,8 +159,8 @@ func (e *ExecutionNodeBuilder) LoadFlags() {
 			flags.BoolVar(&e.exeConf.enableBlockDataUpload, "enable-blockdata-upload", false, "enable uploading block data to Cloud Bucket")
 			flags.StringVar(&e.exeConf.gcpBucketName, "gcp-bucket-name", "", "GCP Bucket name for block data uploader")
 			flags.StringVar(&e.exeConf.s3BucketName, "s3-bucket-name", "", "S3 Bucket name for block data uploader")
-			flags.Uint64Var(&e.exeConf.executionDataPrunerHeightRangeTarget, "execution-data-height-range-target", 65000, "target height range size used to limit the amount of Execution Data kept on disk")
-			flags.Uint64Var(&e.exeConf.executionDataPrunerThreshold, "execution-data-height-range-threshold", 65000, "height threshold used to trigger Execution Data pruning")
+			flags.Uint64Var(&e.exeConf.executionDataPrunerHeightRangeTarget, "execution-data-height-range-target", 800000, "target height range size used to limit the amount of Execution Data kept on disk")
+			flags.Uint64Var(&e.exeConf.executionDataPrunerThreshold, "execution-data-height-range-threshold", 100000, "height threshold used to trigger Execution Data pruning")
```
jordanschalm, line 163:
```suggestion
			flags.Uint64Var(&e.exeConf.executionDataPrunerHeightRangeTarget, "execution-data-height-range-target", 800_000, "target height range size used to limit the amount of Execution Data kept on disk")
			flags.Uint64Var(&e.exeConf.executionDataPrunerThreshold, "execution-data-height-range-threshold", 100_000, "height threshold used to trigger Execution Data pruning")
```

module/executiondatasync/provider/provider.go
```diff
@@ -115,6 +115,9 @@ func (p *Provider) storeBlobs(parent context.Context, blockHeight uint64, blobCh
 
 // Provide adds the block execution data for a newly executed (generally not sealed or finalized) block to the blob store for distribution using Bitswap.
 // It computes and returns the root CID of the execution data blob tree.
+// This function returns once the root CID has been computed, and all blobs are queued for storage
+// in the Bitswap Blobstore. Storage operations are performed asynchronously and may continue after
+// this function returns.
```
jordanschalm, line 120:
```suggestion
// This function returns once the root CID has been computed, and all blobs are successfully stored
// in the Bitswap Blobstore.
```
Adjusting my original suggestion here, because it was based on an incorrect understanding of the behaviour.

jordanschalm (approved):
Took a second pass and all of the notes from my review are addressed. (Added a couple additional minor/documentation suggestions in this review as well.) 

I'm approving based on my review comments being resolved, though it looks like some other reviewer's comments are outstanding - those should still be addressed before merging.

module/executiondatasync/pruner/pruner.go
```diff
@@ -0,0 +1,204 @@
...
+	return p, nil
+}
+
+// NotifyFulfilledHeight notifies the Pruner of the latest fulfilled height.
+func (p *Pruner) NotifyFulfilledHeight(height uint64) {
+	if util.CheckClosed(p.cm.ShutdownSignal()) {
+		return
+	}
+
+	p.fulfilledHeights <- height
```
peterargue, line 135: I think this block execution if the pruner falls behind. We can just skip the notification if the buffer is full since we only care about the highest height.

peterargue (approved):
looks good

module/executiondatasync/tracker/storage.go
```diff
@@ -0,0 +1,529 @@
...
+
+const blobRecordKeyLength = 1 + 8 + blobs.CidLength
+
+func makeBlobRecordKey(blockHeight uint64, c cid.Cid) []byte {
+	blobRecordKey := make([]byte, blobRecordKeyLength)
+	blobRecordKey[0] = prefixBlobRecord
+	binary.LittleEndian.PutUint64(blobRecordKey[1:], blockHeight)
+	copy(blobRecordKey[1+8:], c.Bytes())
+	return blobRecordKey
+}
```
zhangchiqing, line 58: There is this [makePrefix](https://github.com/onflow/flow-go/blob/master/storage/badger/operation/prefix.go#L99) function in badger operation package, that can be reused. 

Any reason we can't reuse the storage modules?

I understand we store them in different database folder, but I think there are low level functions could be reused, no?

PR #2842, opened 2022-07-21
Execution data requester v2

smnzhu:


janezpodhostnik:
Is this still relevant?

jordanschalm:
@janezpodhostnik This is a version of execution data sync which allows for proactively syncing unsealed execution data. It reduces the time for Access nodes to be able to answer queries about just-sealed blocks, compared to v1 which is currently on master. We omitted it from last spork, because execution data sync as a whole was a new addition, and we wanted to start with the simpler requester logic. Eventually we'll want to review, test, and deploy this, but I don't think there is a need to prioritize it right now.

Kay-Zee:
think this needs a new owner, @zhangchiqing @j1010001 @franklywatson

PR #2844, opened 2022-07-22
6298 localnet loader tps reporting - DRAFT

AndrewM-SDET:
Providing visibility to the localnet loader changes for TPS reporting.


integration/utils/contLoadGenerator.go
```diff
@@ -254,6 +257,14 @@ func (lg *ContLoadGenerator) Start() {
 	lg.workerStatsTracker.StartPrinting(1 * time.Second)
 }
 
+func (lg *ContLoadGenerator) StartTime() {
```
pattyshack, line 260: rename to SetStartTime.  use verb name to indicate action. same below

integration/utils/worker_stats_tracker.go
```diff
@@ -57,22 +57,16 @@ func (st *WorkerStatsTracker) AddTxSent() {
...
 func (st *WorkerStatsTracker) AvgTPSBetween(start, stop time.Time) float64 {
 	sum := 0
-	toDelete := make([]int64, 0)
 
 	st.mux.Lock()
 	defer st.mux.Unlock()
 	for timestamp, count := range st.txsSentPerSecond {
-		if timestamp < start.Unix() {
-			toDelete = append(toDelete, timestamp)
+		if timestamp < start.Unix() || timestamp > stop.Unix() {
```
pattyshack, line 64: rm txsSentPerSecond sliding window as well if we do this

integration/loader/main.go
```diff
@@ -187,17 +191,87 @@ func main() {
...
+	}
+
+	avgTps := float64(totalTxs) / float64(executionNodeCount)
+	return avgTps
+}
+
+func printPrometheusTPS(preCount float64, postCount float64, duration time.Duration) {
+	totalTxs := postCount - preCount
+	totalTPS := totalTxs / duration.Seconds()
+	f, err := os.OpenFile("avgTps.log", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
pattyshack, line 261: use defer close pattern

integration/loader/main.go
```diff
@@ -187,17 +191,87 @@ func main() {
...
+	if err == nil {
+		totalTPSstring := fmt.Sprintf(
+			"\nOutput TPS:%.4f, Time:%v, Txs:%v",
+			totalTPS,
+			duration,
+			totalTxs)
+		f.Write([]byte(totalTPSstring))
```
pattyshack, line 268: check for err

simonhf:
As a general comment, it might be good to have a 2nd way to determine output TPS. This second way could be optionally enabled to double check that the first and default way is working correctly. For example, I tried a second way manually here [1] (on an earlier version of the code) and wasn't convinced that the output TPS reported is accurate. Ideally both mechanisms -- assuming they are both accurate -- used to calculate output TPS should produce similar results?

My suggestion is to have the 2nd mechanism available for localnet load tests, and it could use the nodes logs (easily available in the localnet environment) to calculate the output TPS. In this way, then localnet load tests, and CI localnet load tests can check / enforce that output TPS is accurate and sanity checked.

[1] https://www.notion.so/dapperlabs/Trying-out-the-new-loader-2c98d5c5e42d4c3f823895de5b6e5b8b

PR #2846, opened 2022-07-22
V0.26 sync engine improvement time.sleep

danielholmes839:
Update v0.26 with changes from https://github.com/onflow/flow-go/pull/2721


engine/common/synchronization/engine.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/onflow/flow-go/engine"
 	"github.com/onflow/flow-go/engine/common/fifoqueue"
-	"github.com/onflow/flow-go/model/events"
+	"github.com/onflow/flow-go/model/chainsync"
```
zhangchiqing, line 15: what is this?

danielholmes839: Just removed it. It was something leftover from fixing merge conflicts.

module/mock/hot_stuff_follower.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.12.1. DO NOT EDIT.
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
zhangchiqing, line 1: I think you might need upgrade the mockery

PR #2855, opened 2022-07-25
Invalidate program cache entries when accounts are freezed.

pattyshack:
Currently, once a program is added to the program cache, we never check the program's code (and its import dependencies) again to verify if the account has been frozen.  This can lead to nondeterminism during execution.

Invalidating the cache when accounts are frozen forces the account check.


fvm/env.go
```diff
@@ -86,6 +86,8 @@ type commonEnv struct {
 	logs          []string
 	rng           *rand.Rand
 	traceSpan     opentracing.Span
+
+	hasFreezedAccounts bool
```
janezpodhostnik, line 90: `hasFrozenAccounts`?

fvm/programs/programs.go
```diff
@@ -93,38 +93,23 @@ func (p *Programs) HasChanges() bool {
...
+		p.cleaned = true
+
+		// Stop using parent's data to prevent
+		// infinite chaining of objects
+		p.parent = nil
+
+		// start with empty storage
+		p.programs = make(map[common.LocationID]*ProgramEntry)
 		return
 	}
```
pattyshack, line 115: btw, why do we need this?  can we just skip setting the program if the location is not a real address?

janezpodhostnik (commented):
Since this is a temporary measure, and since calling `SetAccountFrozen` is extremely rare. Never happened on mainnet before. It might be better to just purge the cache inside `SetAccountFrozen`.

pattyshack:
purging the cache inside SetAccountFrozen doesn't seem kosher since the state has not been committed yet.  Also, I prefer to keep all the invalidation logic in one place.

fvm/programs/programs.go
j1010001: @janezpodhostnik , @m4ksio ^^ thoughts ?

bluesign: Other non real address locations are, Transaction and Script context afaik. So caching them for sure has some value.

bluesign:
I don't understand the change ( or intention here ), can you please write a description @pattyshack  ?

fvm/programs/programs.go
```diff
@@ -93,38 +93,23 @@ func (p *Programs) HasChanges() bool {
...
-// It should be gone when the extra retry is gone
-func (p *Programs) ForceCleanup() {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-
-	p.unsafeForceCleanup()
-}
-
-func (p *Programs) Cleanup(changedContracts []ContractUpdateKey) {
+func (p *Programs) Cleanup(changedContracts []ContractUpdateKey, hasFreezedAccounts bool) {
```
m4ksio, line 96: I think we should keep consistent form and use 'frozen' everywhere

fvm/programs/programs.go
m4ksio: This had to do with Cadence requirement that once program is set in the cache, it must be available during the execution.
Not sure if that's still a thing - @turbolent might know

fvm/env.go
```diff
@@ -86,6 +86,8 @@ type commonEnv struct {
 	logs          []string
 	rng           *rand.Rand
 	traceSpan     opentracing.Span
+
+	hasFrozenAccounts bool
```
m4ksio, line 90: nitpicking, but I find the name misleading. Its only true if the freezing has occurred, not if there are frozen accounts. `hasAccountFrozen` ? `didFreeze` ?

fvm/programs/programs.go
```diff
@@ -93,38 +93,23 @@ func (p *Programs) HasChanges() bool {
...
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
 	// In mature system, we would track dependencies between contracts
 	// and invalidate only affected ones, possibly setting them to
 	// nil so they will override parent's data, but for now
 	// just throw everything away and use a special flag for this
-	if len(changedContracts) > 0 {
-		p.unsafeForceCleanup()
+	if len(changedContracts) > 0 || hasFreezedAccounts {
```
m4ksio, line 104: Here we use a list of contracts despite the fact that we only check the length. This is in an anticipation that we will use this information one day to properly invalidate affected entire only.
We should probably do the same for accounts, ie provide it with a list of frozen accounts rather than set a boolean flag.

pattyshack:
address all comments, in particular, keep track of the list of frozen accounts instead of a single bool

fvm/programs/programs.go
j1010001: @SupunS do you know ?

fvm/env.go
```diff
@@ -86,6 +86,8 @@ type commonEnv struct {
 	logs          []string
 	rng           *rand.Rand
 	traceSpan     opentracing.Span
+
+	frozenAccounts []common.Address
```
janezpodhostnik, line 90: Thinking about this a bit more, `frozenAccounts []common.Address` should be part of `accounts`, not part of the env.

janezpodhostnik (commented):
2 more comments, otherwise looks good.

fvm/env.go
pattyshack: it's part of env right now cuz SetAccountFrozen is part of env.  I completely agree that these should be refactored out of env, and intend to do so in future PRs.

fvm/transactionInvoker.go
```diff
@@ -134,7 +132,7 @@ func (i *TransactionInvoker) Process(
 	// this writes back the contract contents to accounts
 	// if any error occurs we fail the tx
 	// this needs to happen before checking limits, so that contract changes are committed to the state
-	updatedKeys, err := env.Commit()
+	updatedKeys, frozenAccounts, err := env.Commit()
```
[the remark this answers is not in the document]
pattyshack: sure.  i can do that (I don't feel strong about it)

fvm/programs/programs.go
SupunS: I'm not fully familiar with the program cache implementation tbh, but yeah, cadence expects the FVM to return the same program that was previously set by the cadence. There's a good explanation for this 'informal contract' here:
https://github.com/onflow/cadence/blob/443ea529bed196b0ce0dc138246ee000ad4250fe/runtime/interface.go#L37-L62

Not having this may cause non-determinism. e.g: https://github.com/onflow/cadence/pull/1642/

pattyshack: can these entries be imported?  if not, then they are just top level "main" functions, and i'm pretty sure it's safe to not cache the "main"

SupunS: a 'program' in this case is always a contract (an import)

pattyshack: can these contracts programs be reused?  if not, we're basically stuck with sequential parsing (or we need a per transaction "cache" for these)

SupunS: yea, program cache is shared across transactions (and blocks?). Say `Tx1` imports contract `A` for the first time, then it will load source code from storage, parse and type check, and caches it in the programs cache. If another transaction `Tx2` imports the same contract `A`, then it will be loaded from the program cache.

pattyshack: there may be some miscommunication here.  I understand what the program cache is suppose to do in general.  My question is why do we need to explicitly evict these contract programs after each transaction?  Do their ast program / elaboration change across transactions?

bluesign: as far as I know, script / tx programs are `Set` but never `Get` ( not sure why they are set )

pattyshack: it's used here https://github.com/onflow/cadence/blob/master/runtime/runtime.go#L1241.  if we never get from cache, we wouldn't need this fix =P

bluesign: I meant non-contract programs :) Transaction and Script

SupunS: ah right! So I checked the code again, and it seems the currently executing script/transaction is also stored in the cache, not only the imported contracts - which I feel is not necessary. Not sure if it was done intentionally, but looks like an oversight to me.

To me, it sounds like we wouldn't need to store the transaction/script in the first place (should be handled at the cadence side), and could simply remove this check/loop from FVM.

bluesign: @SupunS only possibility here is caching scripts, but then FVM is evicting them. Maybe initial idea was some cache with TTL.

SupunS: ah yes, good point! There were some talk/ideas about re-using scripts. i.e: Sending the changes as script arguments, instead of sending a new script with changes. Perhaps that was the idea. (Sorry I don't know the details)

bluesign:
Sorry in advance if this is super ridiculous, but nondeterminism blurred my brain a bit. 

Now we said freezeAccount caused nondeterminism, in my understanding:

2 EN had different cache. 
- One had hit for cache lookup on some contract importing frozen account contract
- Second had miss on that, so tried to parse/check program but failed on frozen account import 

Is this theory correct?

pattyshack:
yes, the ENs are never prefectly in sync.  e.g., they restart at different times, and hence their cache content may vary.

bluesign:
> yes, the ENs are never prefectly in sync. e.g., they restart at different times, and hence their cache content may vary.

Thanks @pattyshack, I have a bad feeling about this, I will try to setup a localnet today and tomorrow and will let you know about the results.

pattyshack:
it may be hard to replicate this on localnet.  you can deterministically produce this issue with the following sequence:
1. start 2 EN
2. make sure both EN process some txn with code from address X + some extra blocks for good measure
3. shutdown one of the EN, leave the other one on
4. freeze address X
5. let the alive EN process txn T with code from address X (this should succeed cuz it's cached)
6. turn back on the shutdown EN to let it catch up to txn T (this should fail cuz it's not cached)

bluesign:
@pattyshack  Technically, correct me if I am wrong, we should be able to survive any cache reset event on a single EN ?

janezpodhostnik (approved):
Looks ok from my side. A test for this would be great as well.

pattyshack:
added a test for cache eviction (I've also switch the other frozen account tests to use the fully setFrozenAccount transaction)

janezpodhostnik (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors hates me ... it keeps complaining about a file I didn't touch =...(

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors retry

PR #2857, opened 2022-07-25
Update Consensus Loop

danielholmes839:
[This PR](https://github.com/onflow/flow-go/pull/2721) updated the follower engine to [submit](https://github.com/onflow/flow-go/pull/2721/files#diff-9876a3f5e3a13c1fe35fc34b77aa7cc88f3c970efe12f9f3abbf5d5d0ebce3c2R311) and [process](https://github.com/onflow/flow-go/pull/2721/files#diff-08f436e5039dd8c83b975482d9c1126337bf36bce71b5535116129eb580ad8c4R148-R153) `messages.BlockResponse`s instead of submitting individual multiple `events.SyncedBlock`s. However, the consensus compliance engine still only processes `events.SyncedBlock`s. So we need to update the consensus compliance engine as well


engine/collection/compliance/core.go
```diff
@@ -301,7 +301,7 @@ func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) err
 
 	// submit the model to hotstuff for processing
 	log.Info().Msg("forwarding block proposal to hotstuff")
-	c.hotstuff.SubmitProposal(header, parent.View)
+	<-c.hotstuff.SubmitProposal(header, parent.View)
```
danielholmes839, line 304: We always wait here. There doesn't seem to be an easy way to check if the block is part of a range response. Maybe this isn't an issue but I just want to confirm.

Previous PR
- Identify range response: https://github.com/onflow/flow-go/pull/2721/files#diff-08f436e5039dd8c83b975482d9c1126337bf36bce71b5535116129eb580ad8c4R146
- Conditional wait: https://github.com/onflow/flow-go/pull/2721/files#diff-08f436e5039dd8c83b975482d9c1126337bf36bce71b5535116129eb580ad8c4R393

zhangchiqing: Please see my comment here
https://github.com/onflow/flow-go/pull/2721/files#r929442712

danielholmes839, line 304: I think I should remove this. Or we could keep it and always wait to process blocks in the compliance engine.

zhangchiqing: Yeah, let's replace it with a TODO:
```suggestion
	// TODO: wait for the returned callback channel if we are processing blocks from range response
	c.hotstuff.SubmitProposal(header, parent.View)
```

engine/consensus/compliance/core.go
```diff
@@ -343,7 +343,12 @@ func (c *Core) processBlockProposal(proposal *messages.BlockProposal) error {
 
 	// submit the model to hotstuff for processing
 	log.Info().Msg("forwarding block proposal to hotstuff")
-	c.hotstuff.SubmitProposal(header, parent.View)
+
+	if inRangeBlockResponse {
```
zhangchiqing, line 347:
```suggestion
	// when the block is in range response, we should wait for hotstuff to finish processing the block,
	// otherwise processing the next block might fail because the current block hasn't been added
	// to protocol state yet.
	if inRangeBlockResponse {
```

engine/consensus/compliance/engine.go
```diff
@@ -64,6 +65,19 @@ func NewEngine(
 	prov network.Engine,
 	core *Core) (*Engine, error) {
 
+	blockResponseQueue, err := fifoqueue.NewFifoQueue(
+		fifoqueue.WithCapacity(defaultBlockQueueCapacity),
```
zhangchiqing, line 69: This define a new capacity with maybe 100 as default. 
The capacity for block queue is for buffering all blocks, we don't want to buffer as many range responses. 
We should buffer less relatively.
```suggestion
		fifoqueue.WithCapacity(defaultRangeResponseQueueCapacity),
```

engine/consensus/compliance/engine.go
```diff
@@ -90,6 +105,16 @@ func NewEngine(
...
 		engine.NewNotifier(),
+		engine.Pattern{
+			Match: func(msg *engine.Message) bool {
+				_, ok := msg.Payload.(*messages.BlockResponse)
+				if ok {
+					core.metrics.MessageReceived(metrics.EngineCompliance, metrics.MessageBlockResponse)
+				}
+				return ok
+			},
+			Store: pendingBlockResponses,
```
zhangchiqing, line 116: Better rename to "range responses", because block responses sounds like responses of each individual block.
```suggestion
			Store: pendingRangeResponses,
```

engine/consensus/compliance/engine.go
```diff
@@ -266,9 +292,26 @@ func (e *Engine) processAvailableMessages() error {
 	for {
 		// TODO prioritization
 		// eg: msg := engine.SelectNextMessage()
-		msg, ok := e.pendingBlocks.Get()
+		msg, ok := e.pendingBlockResponses.Get()
+		if ok {
+			blockResponse := msg.Payload.(*messages.BlockResponse)
+			for _, block := range blockResponse.Blocks {
+				// process each block
```
zhangchiqing, line 299:
```suggestion
				// process each block and indicate it's from a range of blocks
```

consensus/hotstuff/eventloop/event_loop.go
```diff
@@ -153,7 +158,8 @@ func (el *EventLoop) loop(ctx context.Context) error {
 
 			processStart := time.Now()
 
-			err := el.eventHandler.OnReceiveProposal(p)
+			err := el.eventHandler.OnReceiveProposal(p.Proposal)
+			close(p.done) // done processing the proposal
```
zhangchiqing, line 162:
```suggestion
			// done processing the proposal, notify the caller (usually the compliance engine) that
			// this block has been processed. If the block is valid, protocol state should have it stored.
			// useful when the caller is processing a range of blocks, and waiting for the current block
			// to be processed by hotstuff before processing the next block.
			close(p.done)
```

zhangchiqing (approved):
Great work, added some minor suggestions.

smnzhu (approved):

PR #2861, opened 2022-07-26
[Crypto] new sentinel errors and doc improvement

tarakby:
Addresses https://github.com/dapperlabs/flow-go/issues/6264

Updates include:
 - doc: detail returned error cases in all public function returning errors.
 - add new specific types of invalid inputs (sentinel errors) : 
      - `dkgInvalidStateTransition` for disallowed transitions. 
      - `aggregateEmptyList` for calling BLS aggregate or mult-verification APIs where empty lists are not allowed
      - `invalidSignatureError` when signature non-verification functions are given an invalid signature (either an invalid serialization, or a serialization that does not correspond to any point on E1 of BLS12-381) - Note this doesn't include G1 membership checks. 
      - `notBLSKeyError` when calling an API that is exclusively defined for BLS keys, but is given a non BLS key. 
      - `nilHasherError` when a `Hasher` is needed but nil is given
      - `invalidHasherSizeError` when a hasher is non-nil but it's output size is not suitable with the cryptographic primitive
      - for all other input errors, the generic `invalidInputsError` is returned. Other specific sentinel suggestions can be added upon request. 
 - update tests for new sentinel errors. 
 - add new DKG tests for `dkgInvalidStateTransition`.
 
 Side changes:
  - add bench for ECDSA decode
  - tighten a check of expand message KMAC output
  - tighten checks for hasher's output size in ECDSA (although the hashers supported by the library are all suitable for the ECDSA curves supported, `Hasher` is a public interface)
  - ECDSA and BLS implementations of interfaces `PrivateKey` and `PublicKey` are private.
  - `common.go` contains all package-wide definitions while primitive-specific definitions are declared in each primitive file (to prepare for future packages refactor)


crypto/types.go
```diff
@@ -117,8 +117,17 @@ type Signature []byte
 // invalid inputs.
 type invalidInputsError struct {
 	error
+	subType invalidInputErrorSubType
 }
 
+type invalidInputErrorSubType int
+
+const (
+	Other invalidInputErrorSubType = 0
```
zhangchiqing, line 126: `Other` sounds like an exception, but I guess it's not, right? It's just other than `dkgInvalidStateTransition` and `aggregationEmptyList`?

Maybe we should call it `default`. If not specified, it's the default error, otherwise, it could be other specific error like `dkgInvalidStateTransition` or `aggregationEmptyList`.

crypto/types.go
```diff
@@ -117,8 +117,17 @@ type Signature []byte
...
+	dkgInvalidStateTransition
+	aggregationEmptyList
```
zhangchiqing, line 128: better to define a `Error` method, that include the error type name in the error message, so that we can see clear error type in the error message rather than 1, 2.

zhangchiqing (approved):

crypto/types.go
```diff
@@ -117,8 +117,17 @@ type Signature []byte
...
 type invalidInputsError struct {
 	error
+	subType invalidInputErrorSubType
 }
 
+type invalidInputErrorSubType int
+
+const (
+	Other invalidInputErrorSubType = 0
+	dkgInvalidStateTransition
```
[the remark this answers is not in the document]
tarakby: I added an extra concise comment for each line. 

There is a more detailed comment for in public APIs for the exported Go doc, for instance here: https://github.com/onflow/flow-go/blob/tarak/6264-crypto-errors/crypto/dkg.go#L91-L96

crypto/types.go
tarakby: > sounds like an exception, but I guess it's not, right?

Yes

`default` cannot be used because it is a Golang expression, I used `generic` instead, let me know if it's still ambiguous.

crypto/types.go
tarakby: That's how I started to implement it. However I wanted the new errors to be "subtypes" of the existing sentinel error `invalidInputsError` , i.e, I want this code to pass:
```
subTyepeError := aggregationEmptyListErrorf("some error")
assert.True(t, IsAggregationEmptyListError(subTyepeError))
assert.True(t, IsInvalidInputsError(subTyepeError))
```
If I declare errors as a separate error type, the last assertion would fail, because `errors.As` fails. I didn't find a better design to achieve that.
Regarding understanding the error message, it won't be necessary to look at the number because the text should be self-explanatory.

zhangchiqing: I see. Yeah, I like your solution. Simple.

But for your reference, this is how to declare errors as a separate error type, the benefit is we can get rid of the `generic` sub type.

https://go.dev/play/p/DOrPlVOEZ1m

tarakby: Ah sweet, that's exactly how I implemented it first, but forgot the `Unwrap` function and it didn't work, https://go.dev/play/p/N7xniNVRH3i. I like your solution better, I might switch to it!

AlexHentschel (changes requested):
Thanks for revising the error returns in the `crypto` package. Previously, the package used predominantly `invalidInputsError`. I assume that your reasoning for not _replacing_ `invalidInputsError` by more specific error returns is downwards compatibility (?) 
To the best of my understanding, you are attempting to introduce sub-types of `invalidInputsError` with the following imperfections:
1. While some error paths have their own dedicated sentinel error that is a sub-type of `invalidInputsError`, you are not consequently following this pattern (see for example my comment [here]()). As a result, higher-level code needs to intersect different sentinel errors to find out what precisely went wrong, which has the following drawbacks:
   * It adds unnecessary complexity for engineers working with the crypto package.
   * Checks for error returns become oder-dependent. In my experience, order-dependency introduces substantial surface for accidental bugs and should be avoided is at all possible (see my comment [here](url))
   * It makes the `crypto` package brittle (increases probability of accidentally introducing bugs during future modifications of the crypto package).
2. I feel that the implementation does not properly reflect that the newly introduced errors are supposed to be sub-types of `invalidInputsError`. The default assumption is that more general errors wrap more specific errors, which naturally happens when errors are bubbled up the call stack. However, your code actually inverts this implementation pattern. (see my comment [here](url))
3. We have no documentation convention to compactly denote that an error is a sub-type of another error. You use different conventions, which I feel are still too ambiguous https://github.com/onflow/flow-go/blob/5d8306237f2b8fa343e4d3f405e13674f1efe470/crypto/bls_multisig.go#L268-L271 https://github.com/onflow/flow-go/blob/5d8306237f2b8fa343e4d3f405e13674f1efe470/crypto/bls_multisig.go#L84-L87
4. I would generally discourage to hard-code semantic relationships between errors. In my experience, this leads to problems, where errors are over-specific and cannot be safely re-used in new code, where the errors might have a different semantic relationship (For example, I find it conceivable that a `DKGInvalidStateTransitionError` might be returned by a sanity check, as a result of an implementation bug as opposed to an invalid input error. Your current implementation entirely excludes the possibility to use `DKGInvalidStateTransitionError`  in such cases, because it is hard-coded to be an `invalidInputsError`). 

### Suggestions:
* Introduce dedicated sentinel errors for each error condition. Thereby, higher-level code doesn't have to intersect errors and worry about oder-dependency of checks. (addresses 1.). E.g. in the following case https://github.com/onflow/flow-go/blob/5d8306237f2b8fa343e4d3f405e13674f1efe470/crypto/bls_multisig.go#L84-L87 also add a dedicated sentinel for the deserialization failure case. Thereby, the calling code can either check for `invalidInputsError` as a whole or individually for the different sub-cases. 
* The more general error (here `invalidInputsError`) should wrap the more specific errors. (addresses 2.) Error wrapping should happen in the implementation, _not_ in the error definition. (addresses 4.) In my experience, the code that needs to be added is negligible (see [here](url)) 
* I feel it would benefit clarity to be explicit in the documentation about error sub-types. E.g. 
   ```golang
    //  - (nil, invalidInputsError), which wraps the following more-specific sentinel errors:
    //		- (nil, aggregationEmptyListError) : if no signatures are provided (input slice is empty)
   ``` 
   (addresses 3.)

crypto/bls_multisig.go
```diff
@@ -478,3 +526,27 @@ func BatchVerifyBLSSignaturesOneMessage(pks []PublicKey, sigs []Signature,
...
+	}
+}
+
+// IsAggregationEmptyListError checks if the input error is of a aggregationEmptyList error type.
+// aggregationEmptyList is a subtype of invalidInputsError, returned when a BLS
+// aggregation function is called with an empty list which is not allowed in some cases.
+func IsAggregationEmptyListError(err error) bool {
+	var target *aggregationEmptyListError
+	return errors.As(err, &target)
+}
```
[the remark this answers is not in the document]
tarakby: I am trying to take a different approach, which is to declare specific errors within specific files. For instance an error that can only be returned by DKG protocols is declared in `dkg.go`. This is to prepare for an upcoming refactoring to split the crypto module into different sub-packages. 
`types.go` is supposed to be the file with sharable types across the module. I'll clean it up now by splitting its types further, I'll also rename it to `common.go`.

crypto/bls_multisig.go
```diff
@@ -70,16 +79,22 @@ func BLSVerifyPOP(pk PublicKey, s Signature) (bool, error) {
...
-// Expected error returns during normal operations:
-//  - invalidInputsError if no signatures are provided (sigs is empty) or
-//    at least one signature fails to deserialize.
+//
+// The function retruns:
+//  - (nil, invalidInputsError):
+//		- (nil, aggregationEmptyListError) : if no signatures are provided (input slice is empty)
+//      - Or a deserialization of at least one signature fails (input is an invalid serialization of a
+// 		compressed G1 element following [zcash]
+//     https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-08.html#name-zcash-serialization-format-)
```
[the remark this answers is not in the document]
tarakby: From my perspective, I was thinking it is fine to check 
```
if generic_erro_type {
    if sub_error_type {} 
    else {}
}
```
If I flatten the error hierarchy and define two independent error types, we would have the same number of checks: 
```
if error_A {}
else if error_B {}
```
I agree that the second code is simpler and less prone to inserting bugs. 

Now the complicated case you described is when a caller wants to check only for the sub error type's complement,  this would look like:
```
if generic_erro_type and !sub_error_type {}
```
I agree this is not great, if we want to test for such case then we should simply add another sentinel error for these errors. I just didn't know that we needed a sentinel error case for "failing to deserialize to a BLS signature". 
I thought [you only wanted to differentiate](https://github.com/dapperlabs/flow-go/issues/6264#issuecomment-1145571653) the empty list case.

I can add a new sentinel error to cover the "failing to deserialize to a BLS signature" (I'm being careful with the name because the error doesn't only happen because of a serialization issue).

crypto/bls_multisig.go
```diff
@@ -478,3 +526,27 @@ func BatchVerifyBLSSignaturesOneMessage(pks []PublicKey, sigs []Signature,
...
+func (e aggregationEmptyListError) Unwrap() error {
+	return e.error
+}
+
+// aggregationEmptyListErrorf constructs a new invalidInputsError (API misuse)
+// for the case of empty lists passed into aggregation functions.
+func aggregationEmptyListErrorf(msg string, args ...interface{}) error {
+	return &aggregationEmptyListError{
+		error: invalidInputsErrorf(msg, args...),
+	}
```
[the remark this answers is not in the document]
tarakby: I gave this some thoughts and I think I will move away from the parent_type/sub_type design. Initially, I thought that it doesn't quite make sense to separate a few error types from the parent type, while keeping all the remaining errors in the same bag together. 
After your comments, and when I looked at the rest of the code (for consistency), it just makes it complicated to keep many sentinel errors as subtypes of a main sentinel error type. I will therefore cut all complexity and use a simpler design : there are no sub types, any error case that needs to be separated from the others gets its own type.

My understanding is that addresses most of your concerns, but please let me know if it doesn't.

crypto/bls_multisig.go
```diff
@@ -238,12 +263,17 @@ func RemoveBLSPublicKeys(aggKey PublicKey, keysToRemove []PublicKey) (PublicKey,
 //
 // This is a special case function of VerifyBLSSignatureManyMessages, using a single
 // message and hasher.
+//
+// The function returns:
+//  - (false, invalidInputsError) if:
+//      - Or at least one key is not of type PubKeyBLSBLS12381
+//      - input hasher is nil or its output size is not 128 bytes
```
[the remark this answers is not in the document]
tarakby: This is a case where the interface i implemented by more than one structure. Therefore the conditions of returning certain errors aren't the same (some errors may even not be returned in some implementations). 
The 2 implementing structures are public, and I think that the caller should read their doc for details instead of the interface's doc. 
( I know that we usually duplicate interface and struct docs in our repo, but I believe in most of these cases, the interface is implemented by one struct and one mock)

tarakby:
@AlexHentschel, this should be ready for a second round. 
It's better to check the delta files of the entire PR because some commits are not very modular. The PR [description](https://github.com/onflow/flow-go/pull/2861#issue-1318405026) is updated and it summarizes the new sentinel error types as well as the other updates.

AlexHentschel (approved):
Amazing. Thank you. 

All my comments are only of aesthetic nature. I have recently learned that with go1.19 there are [updated formatting guidelines](https://go.dev/doc/comment#syntax). [Alexey already updated a lots of packages](https://github.com/onflow/flow-go/pull/3140).

Tried to follow up here to and support adopting the godoc formatting, fixed some typos, compacted some code in **[PR #3247](https://github.com/onflow/flow-go/pull/3247)** targeting your brach (completely optional).

tarakby:
bors merge

PR #2864, opened 2022-07-26
[WIP] resource reporter

ramtinms:


cmd/util/ledger/reporters/resource_reporter.go
```diff
@@ -0,0 +1,181 @@
...
+		err = r.progress.Add(1)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	wg.Done()
+}
+
+func allFieldsAreConstantSize(logger zerolog.Logger, inter *interpreter.Interpreter, value *interpreter.CompositeValue) bool {
```
SupunS, line 146: I think this recursive check might have to be done on the 'type' (i.e: `CompositeType`) rather than the 'value' (`CompositeValue`).
The reason is, there could be optional fields (`T?`), and the stored value is `nil`. Even though the `nil` value is not expandable, the potential value space for that field is expandable.
e.g:
```cadence
struct Foo {
    var values: [Bar]?
}
```
Here if the value was constructed as `Foo {values: nil}` the reporter would return as if it is inline-able. But in reality, it may be not.

ramtinms: How do I get to the type using the iterator ?

Kay-Zee:
@ramtinms @SupunS what's the status on this PR? Still needed?

ramtinms:
Results of this code has been collected and I don't think we need to merge it any more, going to close for now until we need it again.

PR #2877, opened 2022-07-27
refactor program logging

pattyshack:
1. Move program logging related api from commonEnv to ProgramLogger
2. Combine MetricsHandler logic into ProgramLogger

NOTE: I plan to do the same thing with the other handlers, and then move all env
related code into a separate package.

### DIFF
--- a/fvm/env.go
+++ b/fvm/env.go
@@ -3,7 +3,6 @@ package fvm
 import (
 	"encoding/hex"
 	"fmt"
-	"time"
 
 	"github.com/onflow/atree"
 	"github.com/onflow/cadence"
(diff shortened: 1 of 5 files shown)


janezpodhostnik (approved):
LGTM. The only concern I have is with the name "program logger". Not sure its the best fit for what it does... maybe "cadence callback logger" or "cadence callback handler" or something completely different

pattyshack:
it's in the context of environment, so it's not so bad.  I want to eventually move all env related code into an "environment" subpackage, that will make this more obvious.

SaveTheRbtz (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

Tonix517 (approved):

PR #2878, opened 2022-07-27
Add Checkpoint Trigger and Runner

zhangchiqing:
This PR extracts the checkpointing triggering logic from the compactor module into two separate modules - Trigger and Runner in order to break down the complexity of compactor.

It reuses the existing compactor's tests to verify the correctness of the refactor.

One unit test is flakey until [Faye's fix is merged.](https://github.com/onflow/flow-go/pull/2926)


ledger/complete/checkpoint/trigger.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type TrieQueue interface {
+	Tries() []*trie.MTrie
+}
+
+type Trigger struct {
+	logger                  zerolog.Logger
+	activeSegmentNum        int                     // the num of the segment that the previous trie update was written to, is used to detect whether trie update is written to a new segment.
+	checkpointedSegumentNum atomic.Int32            // the highest segment num that has been checkpointed
```
zhangchiqing, line 23: This is one change I made. 

The previous solution is to save the next segment num in the future to trigger checkpointing. And my change is to save the last checkpointed segment num.

The difference is that we don't need to update this value when checkpointing retry is needed.

ledger/complete/trigger.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Trigger takes checkpointing related configs and decides when to trigger checkpointing,
+// if triggered, it will use the checkpointRunner to create checkpoint.
+// It ensures only one checkpoint is being created at any time.
+type Trigger struct {
+	logger                  zerolog.Logger
+	activeSegmentNum        int                     // the num of the segment that the previous trie update was written to, is used to detect whether trie update is written to a new segment.
+	checkpointedSegumentNum *atomic.Int32           // the highest segment num that has been checkpointed
+	checkpointDistance      uint                    // the total number of segment (write-ahead log) to trigger checkpointing
+	isRunning               *atomic.Bool            // an atomic.Bool for whether checkpointing is currently running.
```
ramtinms, line 27: Just out of curiosity, is there a benefit of using atomic.Bool instead `semaphore.NewWeighted(1)` ?

zhangchiqing: No big difference. However,I think semaphore is mostly useful in a scenario where Weight is above 1. 

In our specific scenario here, Weight is 1, as we only want at most a single thread to run checkpointing. So it makes more sense simply using an atomic.Bool. A semaphore is a little bit overkill. 

With the isRunning state as an atomic.Bool value, if we only allow one thread to change this value, and only the thread changed isRunning to true to run the checkpointing, and we set isRunning back to false after checkpointing is done, Then we can guarantee checkpointing is run by at most one thread at any time.

ledger/complete/compactor.go
```diff
@@ -45,16 +38,14 @@ type checkpointResult struct {
 // This will be resolved automaticaly after the forest LRU Cache
 // (code outside checkpointing) is replaced by something like a FIFO queue.
 type Compactor struct {
```
ramtinms, line 40: We need to update the go doc here if we are moving checkpointer to somewhere else

PR #2879, opened 2022-07-27
[SyncEngine] not request pending ancestors

zhangchiqing:
This PR removes a O(N^2) loop in processing block proposals. See reasoning in comments.

The loop would cause a unnecessary O(N^2) operation when processing a range of blocks from sync engine and the first block is pending.


jordanschalm (approved):
Looks good, I added only some documentation tweak suggestions.

kc1116 (approved):

zhangchiqing:
bors merge

PR #2880, opened 2022-07-27
[FVM] Deduct fees on all networks by default

SaveTheRbtz:
This diff makes localnet and benchnet perf tests more realistic by supporting fee deduction.  (Fee deduction currently accounts for ~30% of transactions' execution time.)

Issue: https://github.com/dapperlabs/flow-go/issues/6310

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -858,7 +858,7 @@ func (fnb *FlowNodeBuilder) initFvmOptions() {
 		fvm.WithBlocks(blockFinder),
 		fvm.WithAccountStorageLimit(true),
 	}
-	if fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Stagingnet || fnb.RootChainID == flow.Mainnet {
+	if fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Stagingnet || fnb.RootChainID == flow.Localnet || fnb.RootChainID == flow.Benchnet || fnb.RootChainID == flow.Mainnet {
 		vmOpts = append(vmOpts,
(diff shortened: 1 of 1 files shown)


cmd/scaffold.go
```diff
@@ -858,7 +858,7 @@ func (fnb *FlowNodeBuilder) initFvmOptions() {
 		fvm.WithBlocks(blockFinder),
 		fvm.WithAccountStorageLimit(true),
 	}
-	if fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Stagingnet || fnb.RootChainID == flow.Mainnet {
+	if fnb.RootChainID == flow.Testnet || fnb.RootChainID == flow.Stagingnet || fnb.RootChainID == flow.Localnet || fnb.RootChainID == flow.Benchnet || fnb.RootChainID == flow.Mainnet {
```
pattyshack, line 861: we can just remove this all together / enable fees everywhere?

SaveTheRbtz: @ramtinms @janezpodhostnik @j1010001 how do you feel about this suggestion?  Or should I just negate the condition to something like `!= Emulator`?

janezpodhostnik (approved):
I'm ok with this, but this might cause some scenarios to break for people working with benchnet and localnet, so we should probably announce this.

cmd/scaffold.go
janezpodhostnik: There is actually a task for enabling this by default in the emulator: https://github.com/onflow/flow-cli/pull/581

Maybe it is time to just enable them everywhere by default...

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #2885, opened 2022-07-28
Fix state to only count bytes written/read when enforceLimit is enabled

pattyshack:
TotalBytesRead/TotalBytesWritten are only used for interaction metering and
should not be updated when metering (aka enforceLimit) is disabled.

NOTE: I'll refactor the logic into meter in a follow up PR (looks like there's
already a TODO in the meter interface)

### DIFF
--- a/fvm/state/state.go
+++ b/fvm/state/state.go
@@ -35,8 +35,6 @@ type State struct {
 	maxKeySizeAllowed     uint64
 	maxValueSizeAllowed   uint64
 	maxInteractionAllowed uint64
-	ReadCounter           uint64
-	WriteCounter          uint64
 	TotalBytesRead        uint64
 	TotalBytesWritten     uint64
 }
@@ -119,13 +117,12 @@ func (s *State) Get(owner, key string, enforceLimit bool) (flow.RegisterValue, e
 		return nil, fmt.Errorf("failed to read key %s on account %s: %w", PrintableKey(key), hex.EncodeToString([]byte(owner)), getError)
 	}
 
-	// if not part of recent updates count them as read
-	if _, ok := s.updateSize[mapKey{owner, key}]; !ok {
-		s.ReadCounter++
-		s.TotalBytesRead += uint64(len(owner) + len(key) + len(value))
-	}
-
 	if enforceLimit {
+		// if not part of recent updates count them as read
+		if _, ok := s.updateSize[mapKey{owner, key}]; !ok {
+			s.TotalBytesRead += uint64(len(owner) + len(key) + len(value))
+		}
+
 		return value, s.checkMaxInteraction()
 	}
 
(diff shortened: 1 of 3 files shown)


SaveTheRbtz:
I think this might have broken the "interactions" stat in the FVM benchmark:
Screen Shot 2022-07-28 at 2 19 03 PM

pattyshack:
the benchmark ran transactions using the service account as payer.  Since interaction enforcement is disabled for the service account, we ended up with zeros.  I've updated the benchmark to run the non-create-account transactions using a different account.  the interaction numbers now more accurately reflects the actual cost of the transaction.

fvm/fvm_bench_test.go
```diff
@@ -399,91 +419,112 @@ func BenchmarkRuntimeTransaction(b *testing.B) {
...
+			templateTx(100, fmt.Sprintf(`
 				signer.load<String>(from: /storage/testpath)
 				signer.save("%s", to: /storage/testpath)
-			`, longString)))
+			`, longString)),
+			false)
 	})
 	b.Run("create new account", func(b *testing.B) {
-		benchTransaction(b, templateTx(50, `let acct = AuthAccount(payer: signer)`))
+		benchTransaction(b, templateTx(50, `let acct = AuthAccount(payer: signer)`), true)
```
pattyshack, line 495: Note: this bench test still uses service account as payer, hence interaction is "zero"

pattyshack:
hmm. you know what, let me split out the benchmark change so we can get a better picture

pattyshack:
benchmark changes got split out into #2894 2894

fvm/state/state.go
```diff
@@ -128,13 +126,12 @@ func (s *State) Get(owner, key string, enforceLimit bool) (flow.RegisterValue, e
...
-		s.ReadCounter++
-		s.TotalBytesRead += uint64(len(owner) + len(key) + len(value))
-	}
-
 	if enforceLimit {
+		// if not part of recent updates count them as read
+		if _, ok := s.updateSize[mapKey{owner, key}]; !ok {
+			s.TotalBytesRead += uint64(len(owner) + len(key) + len(value))
+		}
+
```
ramtinms, line 134: I'm not sure if thats what we want to do, ideally we want to be able to measure interactions but not enforce them.

pattyshack: if that's the case, shouldn't we also do the same for other metered metrics?  this is consistent with computation / memory metering.

pattyshack: we shouldn't confuse metering with observability metrics.  perhaps the right thing to do is to add a separate prometheus metric?

bluesign: > computation / memory metering (i.e., we don't count when limit is not enforced). 

I mean technically we are counting when weight is zero: https://github.com/onflow/flow-go/blob/67ed1ebcc41aa6b7cf39677d58500bbf7251c6be/fvm/meter/weighted_meter.go#L354

janezpodhostnik: Like Ramtin said, we still want to measure we just don't want there to be a limit with interactions.

Its actually computation and memory that should behave the same way as the interaction. But the `enforceMemoryLimits` and `enforceComputationLimits` were used to get around some bugs, so we need to write some tests for that before we change the behavior.

pattyshack: All I'm advocating for is consistency between how we measure and limit thing.  How that's achieved doesn't matter much to me.

With that say, changing interaction limit to behave like memory/computation limits will cause us less headache in the long run, since otherwise, we will need to readjust the limit threshold every time we make major change to the non-enforced portion of the execution.

As I've pointed out above, we shouldn't conflict observability metrics with limit enforcement.  e.g., we can just maintain two meters; a unlimited meter for measuring everything and a limited mete for limit enforcement.

pattyshack:
if #2922 is accepted, i'll refactor the changes i've made here into weighted meter, and we can just double measure using the yet-to-be-implemented meter variant (which will internally maintain two separate weighted meter as proposed in the above comment).

pattyshack:
This no longer makes sense in light of discussion in #2922

PR #2886, opened 2022-07-28
Fix observer integration ports

danielholmes839:
randomly generate exposed ports for observer containers


integration/testnet/network.go
```diff
@@ -769,7 +777,7 @@ type ObserverConfig struct {
 	AccessGRPCSecurePort    string // Does not change the access node
 }
 
-func (net *FlowNetwork) AddObserver(ctx context.Context, conf *ObserverConfig) (stop func(), err error) {
+func (net *FlowNetwork) AddObserver(ctx context.Context, t *testing.T, conf *ObserverConfig) (stop func(), err error) {
```
kc1116, line 780:
```suggestion
func (net *FlowNetwork) AddObserver(t *testing.T, ctx context.Context, conf *ObserverConfig) (stop func(), err error) {
```

Make testing.T first parameter is consistent with the rest of the file.

kc1116 (approved):
Left  a minor suggestion

lolpuddle (approved):

integration/testnet/network.go
danielholmes839: Updated in [this commit](https://github.com/onflow/flow-go/pull/2886/commits/b0e4c85323347ef13a3854dee3bbd61cf2c9a343#diff-7540be57ca1fb2714cfce630ad0eba7bfc97362d8e30b79ac2551dc68792e75bR780)

PR #2887, opened 2022-07-28
Fix linting

zhangchiqing:
Fixing the following linting error
```
network/p2p/dns/resolver.go:37:2: field `processingTXTs` is unused (unused)
        processingTXTs map[string]struct{} // ongoing txt lookups through underlying resolver
        ^
network/p2p/dns/resolver.go:36:2: field `processingIPs` is unused (unused)
        processingIPs  map[string]struct{} // ongoing ip lookups through underlying resolver
        ^

fvm/accounts_test.go:1490:86: SA4009: argument view is overwritten before first use (staticcheck)
                        run(func(t *testing.T, vm *fvm.VirtualMachine, chain flow.Chain, ctx fvm.Context, view state.View, programs *programs.Programs) {
                                                                                                          ^
fvm/accounts_test.go:1500:5: SA4009(related information): assignment to view (staticcheck)
                                view = delta.NewView(func(owner, key string) (flow.RegisterValue, error) {
                                ^
```

### DIFF
--- a/fvm/accounts_test.go
+++ b/fvm/accounts_test.go
@@ -1497,14 +1497,14 @@ func TestGetStorageCapacity(t *testing.T) {
 					}
 				`, address)))
 
-				view = delta.NewView(func(owner, key string) (flow.RegisterValue, error) {
+				newview := delta.NewView(func(owner, key string) (flow.RegisterValue, error) {
 					if key == state.KeyAccountStatus {
 						return nil, fmt.Errorf("error getting register %s, %s", flow.BytesToAddress([]byte(owner)).Hex(), key)
 					}
 					return nil, nil
 				})
 
-				err := vm.Run(ctx, script, view, programs)
+				err := vm.Run(ctx, script, newview, programs)
 				require.ErrorContains(t, err, fmt.Sprintf("error getting register %s, %s", address.Hex(), state.KeyAccountStatus))
 			}),
 	)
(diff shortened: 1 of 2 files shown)


SaveTheRbtz (approved):

Tonix517 (approved):

Tonix517:
thanks Leo. a related question: what do you think we add `make lint` into our git post-commit hooks? so that all potential lint errors can be eliminated before creating PR

zhangchiqing:
> a related question: what do you think we add make lint into our git post-commit hooks? so that all potential lint errors can be eliminated before creating PR

Good point. Just that `make lint` takes a while, if every commit will trigger linting, I feel it might affect productivity.

zhangchiqing:
bors merge

Tonix517:
> > a related question: what do you think we add make lint into our git post-commit hooks? so that all potential lint errors can be eliminated before creating PR
> 
> Good point. Just that `make lint` takes a while, if every commit will trigger linting, I feel it might affect productivity.

then how about adding it into `pre-push` hook?

zhangchiqing:
bors retry

SaveTheRbtz:
@Tonix517  IIUC our commit queue already has the "[CI/Lint](https://github.com/onflow/flow-go/runs/7569130531?check_suite_focus=true)" workflow and `bors` will fail on any lint failures.  Though it is not a blocker for manual merges but I'm not sure that it should be given manual merge is only for emergency situations.

Tonix517:
> @Tonix517 IIUC our commit queue already has the "[CI/Lint](https://github.com/onflow/flow-go/runs/7569130531?check_suite_focus=true)" workflow and `bors` will fail on any lint failures. Though it is not a blocker for manual merges but I'm not sure that it should be given manual merge is only for emergency situations.

the motivation of my previous proposal was to remove that workflow. but since that lint workflow is apparently not the slowest, so no strong preference here.

PR #2888, opened 2022-07-28
[Util] set/get hotstuff view

zhangchiqing:
close https://github.com/dapperlabs/flow-go/issues/6346

This PR add the following command
```
# to get current hotstuff view
./util hotstuff-view get --datadir ./protocol

# to set current hotstuff view
./util hotstuff-view set 100 --datadir ./protocol
```

### DIFF
--- a/cmd/util/cmd/hotstuff-view/cmd/reader.go
+++ b/cmd/util/cmd/hotstuff-view/cmd/reader.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/consensus/hotstuff/persister"
+)
+
+type Reader struct {
+	persister *persister.Persister
+}
+
+func NewReader(persister *persister.Persister) *Reader {
+	return &Reader{
+		persister: persister,
+	}
+}
+
+func (r *Reader) SetHotstuffView(view uint64) error {
+	if view == 0 {
+		return fmt.Errorf("hotstuff view is not allowed to set to 0, please specify --view")
+	}
+
+	err := r.persister.PutStarted(view)
+	if err != nil {
+		return fmt.Errorf("could not put hotstuff view %v: %w", view, err)
+	}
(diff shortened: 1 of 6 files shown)


cmd/util/cmd/hotstuff-view/cmd/setter.go
```diff
@@ -0,0 +1,54 @@
...
+
+var SetterCmd = &cobra.Command{
+	Use:   "set",
+	Short: "set hotstuff view",
+	Run:   runSet,
+}
+
+func init() {
+	rootCmd.AddCommand(SetterCmd)
+
```
durkmurder, line 23:
```suggestion
```

durkmurder (approved):
Looks good. Can we place a comment that this is temporary and it will be obsolete as soon as we update to active pacemaker?

cmd/util/cmd/hotstuff-view/cmd/reader.go
```diff
@@ -0,0 +1,39 @@
...
+	return &Reader{
+		persister: persister,
+	}
+}
+
+func (r *Reader) SetHotstuffView(view uint64) error {
+	if view == 0 {
+		return fmt.Errorf("hotstuff view is not allowed to set to 0, please specify --view")
+	}
+
```
jordanschalm, line 23:
```suggestion

	// ensure we don't set the view lower than the current view, because this may cause this node to violate protocol rules
	currentView, err := r.GetHotstuffView()
	if err != nil {
		return err
	}
	if view < currentView {
		return fmt.Error("hotstuff view cannot be lower than current view (%d)", currentView)
	}

```

zhangchiqing: Good catch

jordanschalm (approved):
Nice, thanks for adding this 👌

zhangchiqing:
bors merge

PR #2889, opened 2022-07-29
Implement make flow-go-wide

simonhf:
Maybe there is an easier way to make Golang source 'wider'?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -477,6 +477,46 @@ tool-remove-execution-fork: docker-build-remove-execution-fork
 check-go-version:
 	go version | grep '1.1[6-9]'
 
+# Hacky script to reverse multi-line Golang statements and strip comments.
+# Note: Is there an easier way? Maybe with gofmt somehow? Or another tool?
+#
+# Why? For easier grepping, e.g. grepping without flow-go-wide:
+# $ find . -type f | egrep "\.go" | xargs egrep "\.Msg\(.+?\)" | wc -l # e.g. 100s of matches missing
+#    1100
+# $ find . -type f | egrep "\.go" | xargs egrep -i "executed failed"   # e.g. statement context missing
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
This will be quite hard to support given that it should be impossible to parse a context free grammar with a regex.

If the purpose is to look at the context then -B/-C should be enough (`ag` does that by default ).  If this is a usability feature, then most IDEs have terminal integration and you can ctrl/cmd+click on filename:lineno to open it directly in the editor:

https://user-images.githubusercontent.com/169976/181661390-c4bb52a9-9801-4826-b116-ebe8483b3cbd.mov

simonhf:
Thanks for the comments!

> This will be quite hard to support given that it should be impossible to parse a context free grammar with a regex.

It does a slightly better job than -- and is just another alternative to -- regular grep or `git grep` and shouldn't need that much support since Golang grammar does not change very often?

> If the purpose is to look at the context then `-B` / `-C` should be enough

Yes, that's what I have been using up until now. But I tried it to search for `zerolog` lines and some lines are ~ 20 lines long! :-) So if I just want to browse types of `zerolog` lines then the output becomes a mess. This PR offers an alternative way to grep where the ~ 20 lines are already flattened onto a single line and therefore `-B` / `-C` grep options are not necessary :-)

However, I would be interested in an alternative way to achieve the same results, because I like the output but like you, I don't like the regexes... If there is already a tool to do this then that'd be great... :-)

> `ag` does that by default

What's `ag`?

> If the purpose of the diff is to simplify automated refactoring

The purpose of the diff is just to give more context when grepping on the command line.

yhassanzadeh13 (approved):

Kay-Zee:
Going to close this as it's not caught any traction on the team. If problems similar to this pop up again, will revisit

PR #2892, opened 2022-07-29
[Consensus] Conflicting seals mitigation strategy 

durkmurder:


cmd/consensus/main.go
```diff
@@ -340,9 +340,16 @@ func main() {
 			return nil
 		}).
 		Module("block seals mempool", func(node *cmd.NodeConfig) error {
-			// use a custom ejector so we don't eject seals that would break
+			// use a custom ejector, so we don't eject seals that would break
 			// the chain of seals
-			seals, err = consensusMempools.NewExecStateForkSuppressor(consensusMempools.LogForkAndCrash(node.Logger), node.DB, node.Logger, sealLimit)
+			rawMempool := stdmap.NewIncorporatedResultSeals(sealLimit)
+			multipleReceiptsFilterMempool := consensusMempools.NewIncorporatedResultSeals(rawMempool, node.Storage.Receipts)
+			seals, err = consensusMempools.NewExecStateForkSuppressor(
```
durkmurder, line 347: resulting mempool is combined from `stdmap` mempool wrapped with `IncorporatedResultSeals` and finally with `ExecStateForkSuppressor`.

module/mempool/consensus/exec_fork_suppressor.go
```diff
@@ -108,21 +108,6 @@ func (s *ExecForkSuppressor) Add(newSeal *flow.IncorporatedResultSeal) (bool, er
 	}
 	blockID := newSeal.Seal.BlockID
 
-	// STEP 2: enforce that newSeal's state transition does not conflict with other stored seals for the same block
```
durkmurder, line 111: we don't enforce consistent state when adding anymore.

module/mempool/consensus/exec_fork_suppressor.go
```diff
@@ -132,15 +117,6 @@ func (s *ExecForkSuppressor) Add(newSeal *flow.IncorporatedResultSeal) (bool, er
 		return false, nil
 	}
 
-	// STEP 4: check whether wrapped mempool ejected the newSeal right away;
```
durkmurder, line 135: I've simplified this checks since we don't support this scenario anymore + it doesn't make sense when underlying mempool has extra logic on querying(enforcing multiple receipts)

module/mempool/consensus/exec_fork_suppressor.go
```diff
@@ -171,15 +147,45 @@ func (s *ExecForkSuppressor) Add(newSeal *flow.IncorporatedResultSeal) (bool, er
...
-	return s.seals.All()
+	seals := s.seals.All()
+	s.mutex.RUnlock()
+	sealsByBlockID := make(map[flow.Identifier]sealsList, 0)
+	for _, seal := range seals {
+		sealsPerBlock := sealsByBlockID[seal.Seal.BlockID]
+		sealsByBlockID[seal.Seal.BlockID] = append(sealsPerBlock, seal)
+	}
+
+	return s.filterConflictingSeals(sealsByBlockID)
```
zhangchiqing, line 158: Who would call seal mempool's All method? If that's not builder, is it OK to just return all the seals even if there are conflictions?

module/mempool/consensus/exec_fork_suppressor.go
zhangchiqing: better to mention that it's possible the seal is conflict with other seals for the same block, we will check that when builder finds seals to add to the block.

module/mempool/consensus/exec_fork_suppressor.go
```diff
@@ -171,15 +147,45 @@ func (s *ExecForkSuppressor) Add(newSeal *flow.IncorporatedResultSeal) (bool, er
...
 }
 
 // ByID returns an IncorporatedResultSeal by its ID
```
zhangchiqing, line 161:
```suggestion
// ByID returns an IncorporatedResultSeal by its ID
// the IncorporatedResultSeal's ID is the same as IncorporatedResult's ID, 
// so this call essentially is to find the seal for the incorporated result in the mempool. 
// Note: This call might crash if the block of the seal has multiple seals in mempool for conflicting 
// incorporated results. Usually the builder will call this method to find a seal for an incorporated 
// result, so the builder might crash if multiple conflicting seals exist.
```

We'd better add similar comment to the builder where ByID is used

zhangchiqing (commented):
Thanks for the fix

module/mempool/consensus/exec_fork_suppressor.go
durkmurder: `All` is not used at all, but I believe it should perform needed checks since this mempool tracks execution forks. 
We could rebuild builder for using `All` instead of `ByID`, but definitely not now.

module/mempool/consensus/exec_fork_suppressor.go
```diff
@@ -132,30 +123,15 @@ func (s *ExecForkSuppressor) Add(newSeal *flow.IncorporatedResultSeal) (bool, er
...
-	// STEP 4: add newSeal to secondary index of this wrapper
-	// CAUTION: the following edge case needs to be considered:
-	//  * the mempool only holds a single other seal (denominated as `otherSeal`) for this block
-	//  * upon adding the new seal, the mempool might decide to eject otherSeal
-	//  * during the ejection, we will delete the entire set from the `sealsForBlock`
-	//    because at this time, it only held otherSeal, which was ejected
-	// Therefore, the value for `found` in the line below might
-	// be different than the value in the earlier call above.
+	// STEP 3: add newSeal to secondary index of this wrapper
+	// CAUTION: We expect that underlying mempool NEVER ejects seals because it breaks liveness.
```
jordanschalm, line 127: We should add this requirement to the godoc for `IncorporatedResultSeals`:
https://github.com/onflow/flow-go/blob/f7bdce3a8204e108dc05e28c0b75a3104cf81602/module/mempool/stdmap/incorporated_result_seals.go#L14

jordanschalm (approved):

zhangchiqing (approved):

module/mempool/consensus/exec_fork_suppressor.go
durkmurder: https://github.com/onflow/flow-go/commit/a9c1bf9c5d5e120bb0a5a3022d3e4a08ee82db7c

durkmurder:
bors merge

durkmurder:
bors retry

durkmurder:
bors retry

durkmurder:
bors cancel

durkmurder:
bors merge

jordanschalm:
bors merge

PR #2893, opened 2022-07-29
[Ledger] Replace LRU cache with a FIFO queue (circular buffer)

ramtinms:
This PR 
- replaces the LRU cache used by the forest with a FIFO queue implemented with a circular buffer. 
- implements an adjusted copy of the great buffer @fxamacker has written for the checkpointer. (added lookup and mutex functionality) 
- updates the ledger to still be able to read trie removal entries from WAL but don't do anything about them and don't capture these types of entries going forward (backward compatibility for old version WALs).

But, Why? 
The initial decision to use an LRU cache had roots in another type of design considered for the ledger. ideally, the ledger should only care about purging the oldest trie added to the buffer and get trie calls having any impact on which trie is going to be evicted. This would also make the checkpointer and forest to be in harmony with how they deal with trie updates.


fxamacker (commented):
Great work and amazing how insanely fast you created this PR, right after talking about it! 

Some suggestions:
1. Replace `sync.Mutex` with `sync.RWMutex` in buffer.
2. Add `forest.PurgeCacheExcept()` test in forest_test.go. 
3. Replace references of LRU cache to FIFO queue in docs in ledger.go and forest.go.

First item is in the code comments but 2 and 3 are just here because it is unmodified by this PR.

Also, maybe rename Buffer to be more specific if you'd like.  I don't have a better name for it :laughing:

ledger/complete/mtrie/forest.go
```diff
@@ -42,19 +39,9 @@ type Forest struct {
...
-			if !ok {
-				panic(fmt.Sprintf("cache contains item of type %T", value))
-			}
-			onTreeEvicted(trie)
-		})
-	} else {
-		cache, err = lru.New(forestCapacity)
-	}
+	// TODO(RAMTIN) update forestCapacity to uint
+	cache := NewBuffer(uint(forestCapacity), onTreeEvicted)
```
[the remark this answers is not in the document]
ramtinms: nice catch.

ledger/complete/mtrie/buffer.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Purge removes all mtries stored in the buffer
+func (b *Buffer) Purge() {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	b.tail = 0
+	b.count = 0
+	b.tries = make([]*trie.MTrie, b.capacity)
```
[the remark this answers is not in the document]
ramtinms: yeah update that one to emit as well

ledger/complete/mtrie/buffer.go
```diff
@@ -0,0 +1,130 @@
...
+	b.lookup[t.RootHash()] = b.tail
+	b.tail = (b.tail + 1) % b.capacity
+	if !b.isFull() {
+		b.count++
+	}
+}
+
+// LastAddedTrie returns the last trie added to the buffer
+func (b *Buffer) LastAddedTrie() *trie.MTrie {
+	if b.count == 0 {
```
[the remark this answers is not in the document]
ramtinms: done

ledger/complete/mtrie/buffer.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/onflow/flow-go/ledger"
+	"github.com/onflow/flow-go/ledger/complete/mtrie/trie"
+)
+
+type OnTreeEvictedFunc func(tree *trie.MTrie)
+
+type Buffer struct {
+	tries         []*trie.MTrie
+	lookup        map[ledger.RootHash]int // index to item
+	lock          sync.Mutex
```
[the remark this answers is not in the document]
ramtinms: done

ledger/complete/mtrie/buffer.go
```diff
@@ -0,0 +1,130 @@
...
+	idx, found := b.lookup[rootHash]
+	if !found {
+		return nil, false
+	}
+	return b.tries[idx], true
+}
+
+// Count returns number of items stored in the buffer
+func (b *Buffer) Count() int {
+	return b.count
```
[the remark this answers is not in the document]
ramtinms: done

ledger/complete/mtrie/buffer.go
```diff
@@ -0,0 +1,130 @@
...
+		count:         0,
+		onTreeEvicted: onTreeEvicted,
+	}
+}
+
+// Purge removes all mtries stored in the buffer
+func (b *Buffer) Purge() {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
```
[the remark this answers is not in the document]
ramtinms: done

ledger/complete/mtrie/trieCache.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Push pushes trie to queue.  If queue is full, it overwrites the oldest element.
+func (tc *TrieCache) Push(t *trie.MTrie) {
+	tc.lock.Lock()
+	defer tc.lock.Unlock()
+
+	if tc.isFull() {
+		oldtrie := tc.tries[tc.tail]
+		if tc.onTreeEvicted != nil {
+			tc.onTreeEvicted(oldtrie)
```
SaveTheRbtz, line 100: how heavy is this callback?  It will be run under the write lock now.

ledger/complete/mtrie/trieCache.go
```diff
@@ -0,0 +1,149 @@
...
+	tc.lock.RLock()
+	defer tc.lock.RUnlock()
+
+	if tc.count == 0 {
+		return nil
+	}
+
+	tries := make([]*trie.MTrie, tc.count)
+
+	if tc.isFull() {
```
SaveTheRbtz, line 71: just a random though: if we used a double-linked list, would the logic here be a bit simpler?

pattyshack: array is much link list

pattyshack, line 71: you don't need to special case isFull()?  the else branch will also work for full?

ledger/complete/mtrie/trieCache.go
```diff
@@ -0,0 +1,149 @@
...
+		if tc.onTreeEvicted != nil {
+			tc.onTreeEvicted(oldtrie)
+		}
+		delete(tc.lookup, oldtrie.RootHash())
+	}
+	tc.tries[tc.tail] = t
+	tc.lookup[t.RootHash()] = tc.tail
+	tc.tail = (tc.tail + 1) % tc.capacity
+	if !tc.isFull() {
+		tc.count++
```
pattyshack, line 108: do this unconditionally.  `tc.count--` in the above if statement instead

ledger/complete/checkpoint_benchmark_test.go
```diff
@@ -145,7 +145,6 @@ func BenchmarkLoadCheckpointAndWALs(b *testing.B) {
 			return err
 		},
 		func(rootHash ledger.RootHash) error {
-			forest.RemoveTrie(rootHash)
```
m4ksio, line 148: With removal of removal, maybe we can get out of it completely, remove `WALDelete` (and `WALUpdate` since now we have only one operation, and all related code

ledger/complete/mtrie/forest_test.go
```diff
@@ -1072,3 +1068,38 @@ func TestNow(t *testing.T) {
 	require.Equal(t, updatedRoot, updatedRoot2)
 	require.Equal(t, 2, size)
 }
+
+func TestPurgeCacheExcept(t *testing.T) {
```
m4ksio, line 1072: Can we also add test cases for purging empty cache, purging except the trie which does not exist inside, purge with only the purged trie in

ledger/complete/mtrie/trieCache.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Purge removes all mtries stored in the buffer
+func (tc *TrieCache) Purge() {
+	tc.lock.Lock()
+	defer tc.lock.Unlock()
+
+	tc.tail = 0
+	tc.count = 0
+	for i := 0; i < tc.capacity; i++ {
+		if tc.onTreeEvicted != nil {
```
m4ksio, line 49: Should we care about the order here? I think it would be consistent to keep eviction preserving order

ledger/complete/mtrie/trieCache.go
fxamacker: Currently `onTreeEvicted` is nil, so it is a placeholder in case we need it.  

If callback is heavy, we can take a snapshot of tries using `Tries()` and call `onTreeEvicted` outside write lock. However, the tradeoff is `Tries()` would use heap alloc.

ledger/complete/mtrie/trieCache.go
m4ksio: Yep, with a full cache `tc.capacity - tc.count` is zero and else branch becomes a one for `isFull`

ledger/complete/mtrie/trieCache_test.go
```diff
@@ -0,0 +1,134 @@
...
+	var randomPath ledger.Path
+	rand.Read(randomPath[:])
+
+	var randomHashValue hash.Hash
+	rand.Read(randomHashValue[:])
+
+	root := node.NewNode(256, nil, nil, randomPath, nil, randomHashValue)
+
+	return trie.NewMTrie(root, 1, 1)
+}
```
m4ksio, line 134: Just one simple concurrency test, this is mission-critical piece of code

m4ksio (approved):

ledger/complete/checkpoint_benchmark_test.go
ramtinms: I think we need it for reading old WALs when we spork, we could get rid of it after next spork.

ledger/complete/mtrie/trieCache.go
pattyshack: if you're worry about running the callback while locking, just convert this into a helper method that return the evicted trie (or nil), and run the callback afterward.  e.g.,

```
func (tc *TrieCache) Push(t *trie.MTrie) {
  evicted := tc.push(t)
  if evicted != nil && tc.onTreeEvicted != nil {
    tc.onTreeEvicted(evicted)
  }
}

func (tc *TrieCache) push(t *trie.MTrie) *trie.MTrie {
   ... the original Push method ...
}
```

ledger/complete/mtrie/trieCache.go
SaveTheRbtz: agree!  that said, given that we use map for O(1) access it should not affect perf too much (esp. given how infrequently we should update the forest.)

ledger/complete/mtrie/trieCache.go
ramtinms: yeah mostly kept for compatibility with upper layer, but in this version we pass nil for production

ledger/complete/mtrie/trieCache.go
ramtinms: we don't care that much since its mostly just used for exporting checkpoint, but anyway I fixed it and added a test at c36ddc3 to be safe for the future

ledger/complete/mtrie/trieCache.go
fxamacker: This branch was my code and it was more readable this way (for me).  Yes, it isn't needed. We can get rid of it.  Nice catch! :+1:

fxamacker (approved):

ledger/complete/mtrie/forest_test.go
ramtinms: added more tests at [c1dc91f](https://github.com/onflow/flow-go/pull/2893/commits/c1dc91f304eb6009b13757e256b90e3bc7470d3b).

ledger/complete/mtrie/trieCache.go
ramtinms: updated at c36ddc3

ledger/complete/mtrie/trieCache_test.go
ramtinms: added a simple on at [02fc84c](https://github.com/onflow/flow-go/pull/2893/commits/02fc84c2a4e380356b853eb396fe35cea68db981)

ledger/complete/mtrie/trieCache.go
ramtinms: updated at [a9efa31](https://github.com/onflow/flow-go/pull/2893/commits/a9efa315ea23a456603047ef3f05dfe8054f1906)

ledger/complete/ledger_test.go
```diff
@@ -581,10 +580,6 @@ func Test_WAL(t *testing.T) {
 			}
 		}
 
-		// test deletion
-		s := led2.ForestSize()
-		assert.Equal(t, s, size)
-
```
ramtinms, line 587: We don't do removals any more so this part is not relevant anymore

ledger/complete/mtrie/trieCache.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Tries returns elements in queue, starting from the oldest element
+// to the newest element.
+func (tc *TrieCache) Tries() []*trie.MTrie {
+	tc.lock.RLock()
+	defer tc.lock.RUnlock()
+
+	if tc.count == 0 {
+		return nil
```
taabodim, line 72: suggestion, if you return an empty slice, it would make the job of caller easier. they won't have to check for `nil` anymore

ramtinms: hey @taabodim we are going to have a follow up PR for improvements, going to include your suggestions in that PR.

ledger/complete/mtrie/trieCache.go
ramtinms: there is going to be a follow up PR for improvements, going to consider your suggestions in the PR.

PR #2894, opened 2022-07-29
Run fvm benchmark transactions under a regular account

pattyshack:
This is more accurate since service account bypasses a lot of checks.

### DIFF
--- a/fvm/fvm_bench_test.go
+++ b/fvm/fvm_bench_test.go
@@ -327,13 +327,19 @@ func BenchmarkRuntimeTransaction(b *testing.B) {
 		logger := zerolog.New(logE).Level(zerolog.DebugLevel)
 
 		blockExecutor := NewBasicBlockExecutor(b, chain, logger)
-		serviceAccount := blockExecutor.ServiceAccount(b)
 
 		// Create an account private key.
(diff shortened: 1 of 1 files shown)


janezpodhostnik:
Good catch. Is there a reason or a limitation for not also doing this for account creation?

fvm/fvm_bench_test.go
```diff
@@ -348,21 +363,26 @@ func BenchmarkRuntimeTransaction(b *testing.B) {
 			}
 			`)
 
-		serviceAccount.AddArrayToStorage(b, blockExecutor, []string{longString, longString, longString, longString, longString})
```
Tonix517, line 351: q: `serviceAccount` does not need this even when we are using service account to do the work?

pattyshack: looks like this is only used for setting up copy / borrowing arrary bench tests.  those now run under regular accounts

pattyshack:
ok.  i've funded all the accounts, and all bench tests now run under a regular account

SaveTheRbtz (approved):

Tonix517 (approved):

pattyshack:
bors merge

PR #2899, opened 2022-07-30
Move program logger / tracer / unsafe random generator into environment package

pattyshack:
1. Renamed EnvContext to Tracer
2. Reduce dependency on Context (Context, like Environment, is a grabbag of random
   parameters.  The goal is to eventually remove dependency on Context completely):
    a. Expose additional method in Environment interface
    b. Stop passing Context into program logger / tracer / rand gen.  Only pass in
       what's necessary
3. Move "fully" refactored code into environment package (The goal is to
   eventually have all environment code in the package.  This also replaces the
   handler package).


pattyshack:
hmm, this has too many changes.  let me split it up a bit ...

PR #2903, opened 2022-08-02
[Profiler] Automatic pprof profile uploads to Google Cloud Profiler.

SaveTheRbtz:
This diffs adds an ability to upload AutoProfiler pprof files to [Cloud Profiler](https://cloud.google.com/profiler/docs/about-profiler).

For this to work VM should be running in GCE, have access to `cloudprofiler.googleapis.com:443`, and have the `https://www.googleapis.com/auth/monitoring.write` scope.  Also, one should check their "Create offline requests per day" quota.

Note that, this functionality is currently hidden behind both `-profiler-enabled` and `-profile-uploader-enabled` feature flags.

While here, also:
* Removed Mutex profiling, since [Block profiling is better](https://github.com/DataDog/go-profiler-notes/blob/main/block.md) for our usecase. 
* Switched from heap to alloc profile.  These only differ in the default sample name but identical otherwise.

Issue: https://github.com/dapperlabs/flow-go/issues/6310

Followup diffs:
- [ ] Add metrics.
- [ ] Add localnet support for uploading pprof data.

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -20,6 +20,8 @@ import (
 	"github.com/rs/zerolog"
 	"github.com/spf13/pflag"
 
+	gcemd "cloud.google.com/go/compute/metadata"
+
 	"github.com/onflow/flow-go/admin"
 	"github.com/onflow/flow-go/admin/commands"
 	"github.com/onflow/flow-go/admin/commands/common"
@@ -530,12 +532,58 @@ func (fnb *FlowNodeBuilder) initMetrics() {
 	}
 }
 
+func (fnb *FlowNodeBuilder) createUploader(client *gcemd.Client) (profiler.Uploader, error) {
+	projectID, err := client.ProjectID()
+	if err != nil {
+		fnb.Logger.Warn().Err(err).Msg("could not get CGE project ID")
+		projectID = "unknown"
+	}
+
+	instance, err := client.InstanceID()
+	if err != nil {
+		fnb.Logger.Warn().Err(err).Msg("could not get CGE instance ID")
+		instance = "unknown"
+	}
+
+	chainID := fnb.RootChainID.String()
+	if chainID == "" {
+		chainID = "unknown"
+	}
+
+	params := profiler.Params{
+		ProjectID: projectID,
+		ChainID:   chainID,
+		Role:      fnb.NodeConfig.NodeRole,
+		Version:   build.Semver(),
+		Commit:    build.Commit(),
+		Instance:  instance,
+	}
+
+	fnb.Logger.Info().Msgf("creating pprof profile uploader with params: %+v", params)
+
+	return profiler.NewUploader(fnb.Logger, params)
+}
+
 func (fnb *FlowNodeBuilder) initProfiler() {
 	// note: by default the Golang heap profiling rate is on and can be set even if the profiler is NOT enabled
 	runtime.MemProfileRate = fnb.BaseConfig.profilerMemProfileRate
 
+	var err error
+	var uploader profiler.Uploader
+	if gcemd.OnGCE() {
+		gcemdClient := gcemd.NewClient(nil)
+		uploader, err = fnb.createUploader(gcemdClient)
+		if err != nil {
+			fnb.Logger.Warn().Err(err).Msg("failed to create pprof uploader, falling back to noop")
+		}
+	} else {
+		fnb.Logger.Info().Msg("not running on GCE, setting pprof uploader to noop")
+		uploader = &profiler.NoopUploader{}
+	}
+
 	profiler, err := profiler.New(
 		fnb.Logger,
+		uploader,
 		fnb.BaseConfig.profilerDir,
 		fnb.BaseConfig.profilerInterval,
 		fnb.BaseConfig.profilerDuration,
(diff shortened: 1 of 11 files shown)


cmd/scaffold.go
```diff
@@ -19,6 +19,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog"
 	"github.com/spf13/pflag"
+	"google.golang.org/api/option"
+
```
pattyshack, line 23: rm newline

cmd/scaffold.go
```diff
@@ -530,12 +539,58 @@ func (fnb *FlowNodeBuilder) initMetrics() {
 	}
 }
 
+func (fnb *FlowNodeBuilder) createUploader(client *gcemd.Client, opts ...option.ClientOption) (profiler.Uploader, error) {
```
pattyshack, line 542: rename to createGCEProfileUploader

cmd/scaffold.go
```diff
@@ -530,12 +539,58 @@ func (fnb *FlowNodeBuilder) initMetrics() {
...
+	fnb.Logger.Info().Msgf("creating pprof profile uploader with params: %+v", params)
+
+	return profiler.NewUploader(fnb.Logger, params, opts...)
+}
+
 func (fnb *FlowNodeBuilder) initProfiler() {
 	// note: by default the Golang heap profiling rate is on and can be set even if the profiler is NOT enabled
 	runtime.MemProfileRate = fnb.BaseConfig.profilerMemProfileRate
 
+	var err error
```
pattyshack, line 578: move this out of initProfiler into a separate function, say createProfileUploader (also see my above renaming comment)

cmd/scaffold.go
```diff
@@ -530,12 +539,58 @@ func (fnb *FlowNodeBuilder) initMetrics() {
...
+		Commit:    build.Commit(),
+		Instance:  instance,
+	}
+
+	fnb.Logger.Info().Msgf("creating pprof profile uploader with params: %+v", params)
+
+	return profiler.NewUploader(fnb.Logger, params, opts...)
+}
+
 func (fnb *FlowNodeBuilder) initProfiler() {
```
pattyshack, line 574: not specific to your PR, but ignore errors (only logging warning) seems kind of terrible.  we should propagate the error whenever possible ...

module/profiler/profiler.go
```diff
@@ -135,18 +157,6 @@ func (p *AutoProfiler) pprofBlock(w io.Writer) error {
 	}
 }
 
-func (p *AutoProfiler) pprofMutex(w io.Writer) error {
```
pattyshack, line 138: why was this removed?  if the gce uploader don't support this, maybe just don't upload it

module/profiler/profiler.go
```diff
@@ -116,11 +138,11 @@ func newProfileFunc(name string) profileFunc {
 	}
 }
 
-func (p *AutoProfiler) pprofHeap(w io.Writer) error {
+func (p *AutoProfiler) pprofAllocs(w io.Writer) error {
```
pattyshack, line 141: if allocs and heap profile different things, we should profile both.

module/profiler/uploader.go
```diff
@@ -0,0 +1,121 @@
...
+	log = log.With().Str("component", "profile_uploader").Logger()
+
+	defaultOpts := []option.ClientOption{
+		option.WithEndpoint(apiAddress),
+		option.WithScopes(scope),
+		option.WithUserAgent("OfflinePprofUploader"),
+		option.WithTelemetryDisabled(),
+		option.WithGRPCDialOption(
+			grpc.WithDefaultCallOptions(
+				grpc.MaxCallSendMsgSize(maxMsgSize),
```
pattyshack, line 59: you realize this potentially means a 1GB protobuf message, right?

module/profiler/profiler.go
SaveTheRbtz: Will mention this in the description: "mutex" is almost a subset of "block" but less useful (it tracks where locks are held instead of where code is waiting (blocked) on acquiring the lock).

module/profiler/profiler.go
SaveTheRbtz: will mention this in the description: alloc and heap are the same. the only difference is which sample is marked as "default".

module/profiler/uploader.go
SaveTheRbtz: yep. Our traces right now are about ~60MiB, so i've given it ~ an order of magnitude of potential growth.  That said, it is still an improvement to the default 2GiB: https://github.com/grpc/grpc-go/blob/13b378bc458535e6da1693fc67e09a4951cd3259/clientconn.go#L97

pattyshack: o.  interesting.  this default used to be 4MB a few years back (I had to bump it up to 6mb by default at dbx cuz of block size).  i guess everyone got tired of manually increasing it

pattyshack (approved):

AndrewM-SDET (approved):
Approving. There is one error message typo but code itself looks good.

SaveTheRbtz:
bors merge

PR #2904, opened 2022-08-02
Dan/2791 observer metrics and logs

danielholmes839:


danielholmes839:
Decided to use Miklos's code instead

PR #2911, opened 2022-08-02
Update crypto scripts for Go 1.18

tarakby:
- `go get` is deprecated for module builds in Go 1.18. It is replaced by `go install`. `go get` can still be used for module updates in `go.mod`. 
- use `go list` tool to get a module version instead of manually looking at `go.mod`. 

Thanks @SaveTheRbtz for the recommendations


crypto/Dockerfile
```diff
@@ -3,6 +3,6 @@
 FROM golang:1.18-buster
 RUN apt-get update
 RUN apt-get -y install cmake zip
-RUN go get github.com/axw/gocov/gocov
-RUN go get github.com/matm/gocov-html
+RUN go install github.com/axw/gocov/gocov
+RUN go install github.com/matm/gocov-html
```
SaveTheRbtz, line 7: `install` needs explicit versions (even if it's just `latest`):

```suggestion
RUN go install github.com/axw/gocov/gocov@latest
RUN go install github.com/matm/gocov-html@latest
```

SaveTheRbtz (approved):

SaveTheRbtz:
You may also want to bump go version in `crypto/go.mod` as it si currently set to 1.16

pattyshack (approved):

tarakby:
bors merge

PR #2912, opened 2022-08-02
implement trie cache using trie queue

zhangchiqing:
There are some common logic in TrieCache and TrieQueue.

This PR implements TrieCache using TrieQueue in order to remove the duplicated logic.

Since both compactor and TrieCache depend on TrieQueue, I moved (git mv) TrieQueue  to a separate package to avoid cycle reference.

Made as a draft to get some early feedback first before fixing all the tests.


ledger/complete/common/triequeue.go
```diff
@@ -43,13 +43,23 @@ func NewTrieQueueWithValues(capacity uint, tries []*trie.MTrie) *TrieQueue {
...
 }
 
-// Push pushes trie to queue.  If queue is full, it overwrites the oldest element.
-func (q *TrieQueue) Push(t *trie.MTrie) {
+// Push pushes trie to queue and return the overwritten element, and whether any old element.
+// was ejected.
+// If queue is full, it overwrites and ejects the oldest element.
+// It return (nil, false) if no item was ejected.
+// It returns(*trie.Mtrie, true) if an old trie was ejected by the pushing of the input trie
+func (q *TrieQueue) Push(t *trie.MTrie) (*trie.MTrie, bool) {
```
zhangchiqing, line 51: added more information  about the internal state during the state transition so that it can be reused by trieCache

ledger/complete/mtrie/trieCache.go
```diff
@@ -43,23 +44,21 @@ func (tc *TrieCache) Purge() {
...
+	toEvict := tc.tries.Tries()
+
+	// update the state first
+	capacity := tc.tries.Capacity()
+	tc.tries = common.NewTrieQueue(capacity, tc.onTreeEvicted)
+	tc.lookup = make(map[ledger.RootHash]*trie.MTrie, capacity)
+
+	// evicting all nodes after state is reset
+	for _, trie := range toEvict {
+		tc.onTreeEvicted(trie)
```
zhangchiqing, line 60: I don't like this onTrieEviceted function:
Is is called before the lock is released, if someone is trying to report some metrics by calling `TrieCache.Count()`, it will introduce a dead lock by mistake.

A better pattern would be just return the evicted items to the caller _after_ releasing the lock, that's the moment actually the item is actually evicted.

Same idea for the `Push` method, it could also return the evicted item in the return value after the lock is released. It avoid chance of introducing dead lock if misused.

Is any code using this callback? I didn't find any

ledger/complete/common/triequeue.go
```diff
@@ -81,3 +91,19 @@ func (q *TrieQueue) Count() int {
 func (q *TrieQueue) isFull() bool {
 	return q.count == q.capacity
 }
+
+func (q *TrieQueue) LastAddedTrie() (*trie.MTrie, bool) {
```
zhangchiqing, line 95: Added the method here so that TrieQueue can be reused by trieCache

PR #2918, opened 2022-08-03
Don't pollute global program caches with non-address programs

pattyshack:
It's not safe to populate non-address programs into the program cache once we
start doing parallel pre-processing since these programs are not reusable.
We'll keep track of these programs within the local environment instead.

NOTE: we'll also need to eventually keep track of address program locally as
well in order to invalidate outdated programs.


SaveTheRbtz (approved):
This does look logical.  What was the reason for keeping them in the programs cache in the first place?

janezpodhostnik (approved):

janezpodhostnik:
Maybe someone from the Cadence team can take a look as well @SupunS

pattyshack:
bors merge

SupunS (commented):
This should be fine for now since non-address programs (transactions/scripts) are not re-usable at the moment. But if we are adding support for reusing transactions/script in future (as a perf optimization), would this not return an out-dated transactions from the cache? (e.g: in a transaction, the imported contract was updated so the global cache is cleared, but the transaction it-self wasn't cleared from the cache). 

Or maybe have I misunderstood on how it works?

pattyshack:
since the transaction prog is not stored in the shared cache, it'll just get gc as part of the transaction environment

PR #2922, opened 2022-08-03
Change meter interface to separate out observable vs enforceable measurements

pattyshack:
Prep work for adding a new meter variant which keeps track of observable and
enforceable measurements separately.

Note that I've also modify MergeMeter to reject non-weighted-meter child.


janezpodhostnik:
Can you explain a bit more here? In which case should observable and meter-able be different? 

I though we wan to transition to always metering the same, but have different limits in different situations?

pattyshack:
Right now, we're enforcing/counting usage in two different ways:
1. for computation and memory, we only count when enforceLimits is true (the downside to this is we lack visibility into unenforced usage for this case)
2. for interaction, we count unconditionally, but enforce limit occasionally (the downside to this is that the interaction limit threshold is unstable; we will need to readjust the limit threshold every time we make major change to the unenforced portion of the execution)

The idea here is that we always enforce limits using computation/memory enforcement behavior (i.e., the Enforced usage), but at the same time count unconditionally (i.e., the Observed usage).

Computation and memory will gain a Observed metric, while interaction will have a new enforcement behavior

pattyshack:
btw, this is roughly what i have in mind (it's still work in progress) https://github.com/onflow/flow-go/blob/patrick/controllable-meter/fvm/meter/controllable_meter.go

janezpodhostnik:
Sorry for taking so long on this one. I'm trying to get through it mentally.

Here is what I'm thinking about:

Having concepts of both enforced and observable for computation/memory/interaction adds a lot of complexity.

Lets consider only 2 parts of the transaction:
- Invoke transaction
- Deduct fees

We want to meter *invoke transaction* and also enforce the limits. Users transaction should not go over computation/memory/interaction/event limits and everything it does should be charge appropriately. So this would be the Enforced usage.

We cannot charge the user for resources used during the *deduct fees* part , because the input to that is the measurements we took before this step (we could charge a flat static fee, but that is a different story). Does it still matter that we meter this part? IMO yes! because there could be an exploit in the cadence fee function that causes infinite (or disproportionate) resources to be consumed.

So what I think we really need is separate meters for different parts, each with its own limit.

```
invokeMeterResults = meter1(
  limits1,
  invoke_transaction(),
)

meter2(
   limits2, // predetermined known limits specially for transaction fees
   deduct_fees(invokeMeterResults)
)
```

I'm not sure we should be giving any (cadence) code the ability to run without a limit.

For info purposes there could be a root meter that collects all resource usage, and that one has no limits (observed usage), but I would perhaps argue that such a root meter is useless, and if we want to collect the aggregate metering results, then just have an array of metering results somewhere and log/report that...

In conclusion (sort of):

- I think that during the transaction there should be different meters for different parts, all with limits, and always metering  (i.e., the Enforced usage).
- The concept of observed usage is not necessary and should be solved by having explicit meter results for each metered part.

pattyshack:
This PR is based on the assumption that we want to maintain a single unified meter for the lifetime of the transaction.  But sounds like you're ok with abandon this altogether, which opens up a lot of options.

Your proposal may become a lot more doable once we standardize nested transactions (#2963).  We can just create different nested transactions (with different meter parameters) for each part.

PR #2925, opened 2022-08-03
[Localnet] do not uninstall loki plugin on start

SaveTheRbtz:
After we switched from loki to `promtail` in #2609 we added this cleanup.  Now that it's been there for a while, it should be safe to remove it.


pattyshack (approved):

kc1116 (approved):

SaveTheRbtz:
bors merge

PR #2930, opened 2022-08-04
[EN Performance] Reduce memory used for ledger.Payload by 32+ GB, eliminate 1+ billion allocs/op, speedup various ops

### LINKED ISSUE #2569: [Execution Node] Reduce memory used for ledger.Payload by replacing decoded Key with encoded key buffer
## Problem

`ledger.Payload` can use less memory.  `Payload.Key` requires 4 allocations for any key with 3 parts.

We don't use `Payload.Key` except in migration and reports so it's expensive to needlessly keep it in memory.  

Also, `Payload.Key` is needlessly decoded while rebuilding mtrie during EN restart as well as checkpointing.

Updates #1744 #2585

## The Proposed Solution

Replace decoded `Payload.Key` with encoded key buffer to reduce number of allocations and avoid needlessly decoding it during EN restarts and checkpointing.

fxamacker:
This PR replaces decoded payload key with encoded key buffer because decoded payload key is only used for migration and reports.  This change made ledger.Payload immutable.

Closes #2569
Updates #1744

### Goals

- (main goal) reduce memory required by EN (operational RAM).
- reduce number of allocations on the heap.
- have zero negative performance tradeoffs.

### Impact

- operational RAM should be reduced by dozens of GB (very roughly 25-50GB initially and more as data grows)
- eliminate over 1+ billion heap allocations for mtrie in memory (about 1+ billion allocs when mtrie is created and additional savings during activities that update mtrie)
- as positive side-effects, speedup
  - ledger update (see TrieUpdate benchstats)
  - EN startup
  - checkpoint (de)serialization
  - TrieProof (de)serialization, and etc.

### Example positive side-effect beyond operational RAM reduction

Benchstats are only for ledger update.  Other improvements are not benchmarked yet.

```
name          old time/op    new time/op    delta
TrieUpdate-4     439ms ± 2%     409ms ± 1%   -6.94%  (p=0.000 n=18+20)

name          old alloc/op   new alloc/op   delta
TrieUpdate-4    73.5MB ± 0%    34.1MB ± 0%  -53.60%  (p=0.000 n=20+20)

name          old allocs/op  new allocs/op  delta
TrieUpdate-4      187k ± 0%      147k ± 0%  -21.44%  (p=0.000 n=20+20)
```

### Caveats

- Updating benchmark results or adding missing benchmarks can be done at a later date.
- Custom functions for CBOR and JSON serialization was added because immutable fields are no longer exported, but the memory reduction and speedups in other ops outweigh this.

### Reviewers

This PR is fairly simple because the most important changes are in the small commit e35bd9449eae87af76ecbdb6422c2b5c5ed44ba9.

The large number of lines changed by other commit is to:
- commit b2517548083f005fb8321d6ba820c7ac885c957b - make ledger.Payload immutable
- commit 74ee21ddaacb4cdc200b7441ee9a6f49d53d0b9e - eliminate circular dependencies (moving ledger/common/encoding/encodiing.go to ledger/trie_encoder.go, split common/utils/testutils.go into 2 packages)


cmd/util/ledger/migrations/account_status_migration.go
```diff
@@ -70,8 +70,12 @@ func (as *AccountStatusMigration) Migrate(payload []ledger.Payload) ([]ledger.Pa
...
 
 	for _, p := range payload {
-		owner := p.Key.KeyParts[0].Value
-		key := p.Key.KeyParts[2].Value
+		k, err := p.Key()
+		if err != nil {
+			return nil, err
+		}
+		owner := k.KeyParts[0].Value
+		key := k.KeyParts[2].Value
```
SaveTheRbtz, line 78: Is there some inherent structure to KeyParts?  Maybe we can convert it to either a Struct or an Array to remove even more dynamic allocations?

fxamacker: @SaveTheRbtz  

>Is there some inherent structure to KeyParts? Maybe we can convert it to either a Struct or an Array to remove even more dynamic allocations?

Ledger key is designed to be flexible, so it can contain variable number of KeyPart.  But we have plans to limit uses of Ledger key to migration and reports so that we can reduce number of heap allocs.

This PR eliminates ledger.Key from mtrie leaf nodes' payload.

The next PR related to this would be to eliminate ledger.Key created outside of migration and reporting which would reduce number of heap allocs caused by ledger.Key.

ledger/complete/mtrie/forest_test.go
```diff
@@ -1038,37 +1038,6 @@ func TestValueSizesWithDuplicatedKeys(t *testing.T) {
 	}
 }
 
-func TestNow(t *testing.T) {
-
```
ramtinms, line 1042: do we know what was the original reason for this test ?

ledger/trie.go
```diff
@@ -207,44 +210,153 @@ func ToPath(pathBytes []byte) (Path, error) {
...
+
+// Value returns payload value.
+// CAUTION: do not modify returned value because it shares underlying data with payload value.
+func (p *Payload) Value() Value {
+	if p == nil {
+		return Value{}
+	}
+	return p.value
+}
+
```
ramtinms, line 330: this is great, this can be a good start for the future PRs to remove deep copies.

ramtinms (approved):
Looks good to me, nice work.

ledger/complete/mtrie/forest_test.go
fxamacker: Yes, `TestNow` was a throwaway test that wasn't meant to be pushed to the repo.  I use `TestNow` to verify my understanding locally and forgot to remove it in a prior PR.

zhangchiqing (approved):
Look good. Great works! Thanks

PR #2935, opened 2022-08-05
[Consensus] Update disabled integration tests

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6201

### Context

This PR enables previously disabled integration tests. Test suits were extended to support new functionality(support for timeout aggregator for instance). Also some changes were made to the actual logic to address some discovered issues.

Important changes:
- `EventHandler` now stores newly created proposal in forks and has extra logic to check if he has already proposed in this round. Due to async design it could happen that event handler tries to create proposal twice for same round. Maybe we will be able to update compliance engine in future to pass each proposal only once for but now it appears to me as a good solution. 
- Updated compliance engine to accept block ranges, this code is on master but I have made a mistake when merging so now it's fixed.


engine/consensus/compliance/engine.go
```diff
@@ -105,6 +123,16 @@ func NewEngine(
...
+		engine.Pattern{
+			Match: func(msg *engine.Message) bool {
+				_, ok := msg.Payload.(*messages.BlockResponse)
+				if ok {
+					core.metrics.MessageReceived(metrics.EngineCompliance, metrics.MessageBlockResponse)
+				}
+				return ok
+			},
+			Store: pendingRangeResponses,
+		},
```
jordanschalm, line 135: > Updated compliance engine to accept block ranges, this code is on master but I have made a mistake when merging so now it's fixed

I might be missing some context on this change, I'm unsure why we've added this 🤔. Have a few questions:
* Do we ever send range requests to the compliance engine (as opposed to the sync engine)? I can't see this anywhere. In general I'm not sure why we handle this message type in this engine now. 
* Should we make the same change to the collection node compliance engine? It doesn't have a queue for this message type currently
https://github.com/onflow/flow-go/blob/201deb6a6dd923e2d1bc4396ddc82081c5a70833/engine/collection/compliance/engine.go#L85-L104

jordanschalm (commented):
Very solid - nice work

consensus/integration/slow_test.go
```diff
@@ -1,98 +1,158 @@
...
+		case *messages.TimeoutObject:
+			// drop first timeout object for every sender for every view
+			blockedPerView, found := blockedTimeoutObjectsTracker[sender.id.NodeID]
+			if !found {
+				blockedPerView = make(map[uint64]uint64)
+				blockedTimeoutObjectsTracker[sender.id.NodeID] = blockedPerView
+			}
+			blocked := blockedPerView[m.View] + 1
+			blockedPerView[m.View] = blocked
+			return blocked%2 == 1, 0
```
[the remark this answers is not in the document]
durkmurder: oh right, that's a leftover

consensus/hotstuff/integration/instance_test.go
```diff
@@ -391,8 +413,51 @@ func NewInstance(t require.TestingT, options ...Option) *Instance {
...
+			aggregator.On("VerifyAndAdd", mock.Anything, mock.Anything, mock.Anything).Return(
+				func(signerID flow.Identifier, _ crypto.Signature, newestQCView uint64) uint64 {
+					for {
+						nv := newestView.Load()
+						if nv < newestQCView {
+							newestView.CAS(nv, newestQCView)
+						} else {
+							break
+						}
+					}
```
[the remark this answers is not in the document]
durkmurder: we actually can use `NewestQCTracker`

consensus/hotstuff/eventhandler/event_handler.go
```diff
@@ -334,12 +336,21 @@ func (e *EventHandler) proposeForNewViewIfPrimary() error {
...
 		Uint64("finalized_view", e.forks.FinalizedView()).
-		Msg("entering new view")
+		Msg("proposing for current view")
 	e.notifier.OnEnteringView(curView, currentLeader)
 
 	if e.committee.Self() != currentLeader {
 		return nil
 	}
+
+	// check if there is proposal for current view, if there is then nothing to do for us
```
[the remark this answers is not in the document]
durkmurder: https://github.com/onflow/flow-go/pull/2935/commits/f4c69ca5f0457012fe2da9ac7106e902c5a892a4

jordanschalm (approved):

engine/consensus/compliance/engine.go
AlexHentschel: Oh man, [PR 2857](https://github.com/onflow/flow-go/pull/2857)  added so many changes ... I wish they had talked to me before adding all the channels. They could have just listened to the `OnBlockIncorporated` event from HotStuff instead of adding an entirely new mechanism 😞

consensus/integration/network_test.go
```diff
@@ -124,7 +126,9 @@ func (n *Network) unicast(event interface{}, channel network.Channel, targetID f
 	}
 
 	sender, receiver := n.node, net.node
+	n.hub.lock.Lock()
 	block, delay := n.hub.filter(channel, event, sender, receiver)
+	n.hub.lock.Unlock()
```
AlexHentschel, line 131: instead of putting the locking in the higher-level code, wouldn't it make sense if the implementations of `BlockOrDelayFunc` were itself concurrency safe?

consensus/integration/blockordelay_test.go
```diff
@@ -9,68 +6,49 @@ import (
...
 	rand.Seed(time.Now().UnixNano())
-	return func(channelID string, event interface{}, sender, receiver *Node) (bool, time.Duration) {
+	return func(channel network.Channel, event interface{}, sender, receiver *Node) (bool, time.Duration) {
 		block := rand.Intn(100) <= percent
 		return block, 0
 	}
 }
 
 func delayReceiverMessagesByRange(low time.Duration, high time.Duration) BlockOrDelayFunc {
 	rand.Seed(time.Now().UnixNano())
```
AlexHentschel, line 50: generally, it is considered statistically not sound to split the stream of random numbers from one pseudo-random number generator [prng] into independent sub-streams. For example, if you have a prng and you use values for purpose A and purpose B alternating:
* From the perspective of process A, you want a steam of numbers with high randomness and independent of the randomness for B. 
* But exactly that quality of "independence" is broken by sub-sampling both streams A and B from the same prng. Furthermore, a sub-sample of a random stream generally does not preserve the quality of randomness and should be avoided. 

These are relevant for statistically sound experiments. While the quality of randomness is not that important here, it is still beneficial to follow best practises. Specifically, the extra engineering effort would be marginal here.

consensus/integration/network_test.go
jordanschalm: +1 I agree, and seems like most of the implementations are already concurrency safe

engine/consensus/compliance/engine.go
```diff
@@ -293,7 +322,25 @@ func (e *Engine) processAvailableMessages() error {
 	for {
 		// TODO prioritization
 		// eg: msg := engine.SelectNextMessage()
```
AlexHentschel, line 324: I think the following code already has the desired prioritization. 
But we don't check for a termination signal here.

consensus/integration/integration_test.go
```diff
@@ -72,7 +72,7 @@ func Test5Nodes(t *testing.T) {
 
 	runNodes(signalerCtx, nodes)
 
-	<-stopper.stopped
+	unittest.RequireCloseBefore(t, stopper.stopped, 30*time.Second, "expect to stop before timeout")
```
AlexHentschel, line 75: unfortunattely this test timed out on CI 😭 
https://github.com/onflow/flow-go/runs/8029489368?check_suite_focus=true

AlexHentschel:
file https://github.com/onflow/flow-go/blob/c1e478ed0e4366575be43a77d26cd34f5d7b1b80/integration/tests/epochs/suite.go seems to have multiple compiler errors. Didn't look into very much, just wanted to add a note for later.

AlexHentschel (approved):
very nice. Great tests. 

Last round of optional revisions is https://github.com/onflow/flow-go/pull/3100

integration/tests/epochs/suite.go
```diff
@@ -79,8 +82,8 @@ func (s *Suite) SetupTest() {
 	consensusConfigs := []func(config *testnet.NodeConfig){
 		testnet.WithAdditionalFlag("--hotstuff-timeout=12s"),
 		testnet.WithAdditionalFlag("--block-rate-delay=100ms"),
-		testnet.WithAdditionalFlag(fmt.Sprintf("--required-verification-seal-approvals=%d", 1)),
-		testnet.WithAdditionalFlag(fmt.Sprintf("--required-construction-seal-approvals=%d", 1)),
+		testnet.WithAdditionalFlag(fmt.Sprintf("--required-verification-seal-approvals=%d", 0)),
+		testnet.WithAdditionalFlag(fmt.Sprintf("--required-construction-seal-approvals=%d", 0)),
```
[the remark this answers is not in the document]
jordanschalm: Yes, Yurii noticed that with `require-approvals>0`, sealing temporarily stopped on some of the longer tests (and had a longer time-to-seal delay in general). The epoch tests in particular are pretty sensitive to this, because we have a pretty tight bound on the time to run the DKG, where we're waiting for blocks to be sealed at various points. 

We decided it was OK to turn it off for this test, since the MVP and verification tests do test the sealing flow requiring approvals. I'll add a comment explaining the decision.

engine/consensus/compliance/engine.go
```diff
@@ -27,6 +27,9 @@ import (
 	"github.com/onflow/flow-go/utils/logging"
 )
 
+// defaultRangeResponseQueueCapacity maximum capacity of block range responses queue
+const defaultRangeResponseQueueCapacity = 100
```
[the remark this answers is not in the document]
jordanschalm: Yurii answered in [this thread](https://github.com/onflow/flow-go/pull/2935#discussion_r940704382):

> I think that updating compliance engine will render this change as not needed since there won't be hard dependency for validation between compliance and hotstuff. Because of it I would wait with changing collection cluster.

jordanschalm:
> file https://github.com/onflow/flow-go/blob/c1e478ed0e4366575be43a77d26cd34f5d7b1b80/integration/tests/epochs/suite.go seems to have multiple compiler errors (maybe it is just a local problem in my IDE?). Didn't look into very much, just wanted to add a note for later.

Do they look like this? 
image

This has been a problem for me as well for a while. The code compiles fine, but GoLand flags it as an error. I also haven't looked into it much. Not new with this PR though (or even this feature branch).

consensus/integration/integration_test.go
jordanschalm: `Test2TimeoutOutof4Instances` seems to have as well :(

integration/tests/epochs/suite.go
```diff
@@ -79,8 +82,8 @@ func (s *Suite) SetupTest() {
 	consensusConfigs := []func(config *testnet.NodeConfig){
 		testnet.WithAdditionalFlag("--hotstuff-timeout=12s"),
 		testnet.WithAdditionalFlag("--block-rate-delay=100ms"),
-		testnet.WithAdditionalFlag(fmt.Sprintf("--required-verification-seal-approvals=%d", 1)),
-		testnet.WithAdditionalFlag(fmt.Sprintf("--required-construction-seal-approvals=%d", 1)),
+		testnet.WithAdditionalFlag(fmt.Sprintf("--required-verification-seal-approvals=%d", 0)),
```
jordanschalm, line 85:
```suggestion
		// we disable requiring approvals on epoch tests, because it increases time-to-seal
		// which can interferes with the time-sensitive DKG and cause test flakiness
		testnet.WithAdditionalFlag(fmt.Sprintf("--required-verification-seal-approvals=%d", 0)),
```

consensus/integration/integration_test.go
jordanschalm: Should be fixed by https://github.com/onflow/flow-go/pull/2935/commits/4c603db291f2fd0aa82cf8a0da17330b4e6e1b5c 🤞

integration/tests/epochs/suite.go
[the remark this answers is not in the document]
AlexHentschel: 👍  thanks for the detailed explanation

engine/consensus/compliance/engine.go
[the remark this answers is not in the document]
AlexHentschel: 👍💡  thanks for the reference

AlexHentschel:
@jordanschalm 
> [Alex] compiler errors in [`epochs/suite.go`](https://github.com/onflow/flow-go/blob/c1e478ed0e4366575be43a77d26cd34f5d7b1b80/integration/tests/epochs/suite.go)
> [Jordan] Do they look like [this](https://github.com/onflow/flow-go/pull/2935#issuecomment-1228951376)? 

yep exactly. You are saying it is only a local problem? 👌 Then we'll just leave it.

PR #2937, opened 2022-08-05
Make memory/interaction metering consistent for all accounts.

pattyshack:
Previously, the cached meter state differed depending on which payer account
loaded the program into the cache, resulting in non-deterministic execution.

Note: the cached usage limits are "incorrect", but they does not impact the
final execution result.

Non-deterministic execution due to frozen account, weight updates, service
account metering are all manifestation of the same underlying problem.  In the
long run, we should stop caching transaction state as part of program cache.

### DIFF
--- a/fvm/executionParameters.go
+++ b/fvm/executionParameters.go
@@ -11,6 +11,91 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
+func setMeterParameters(env *commonEnv) error {
+	// Don't enforce limits while setting up limits.
+	env.sth.DisableAllLimitEnforcements()
+	defer env.sth.EnableAllLimitEnforcements()
+
+	if env.ctx.AllowContextOverrideByExecutionState {
+		err := setExecutionParameters(env)
+		if err != nil {
+			return err
+		}
+	}
+
+	if env.ctx.DisableMemoryAndInteractionLimits {
+		env.sth.DisableMemoryAndInteractionLimits()
+	}
+
+	return nil
+}
+
+func setExecutionParameters(env *commonEnv) error {
+	// Check that the service account exists because all the settings are
+	// stored in it
+	serviceAddress := env.Context().Chain.ServiceAddress()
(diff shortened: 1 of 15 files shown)


fvm/state/state_holder.go
```diff
@@ -33,7 +36,8 @@ func (s *StateHolder) SetActiveState(st *State) {
 
 // SetPayerIsServiceAccount sets if the payer is the service account
 func (s *StateHolder) SetPayerIsServiceAccount() {
-	s.payerIsServiceAccount = true
+	s.activeState.SetTotalMemoryLimit(math.MaxUint64)
```
janezpodhostnik, line 39: The fact that this is only on active state worries me a bit.

If this is not done on the base state it will cause issues when merging the child state (with high limits) into the parent state with low limits. Maybe these limits should already be set when creating the root state.

This will also be invalidated [here](https://github.com/onflow/flow-go/blob/5119cb5/fvm/transactionEnv.go#L172) if the limit is set in the state.

janezpodhostnik (changes requested):

fvm/state/state_holder.go
pattyshack: good catch

pattyshack:
I've deduplicated setExecutionParameters and move most of the weights/limit setup to one place (Ideally, we would pass all of these arguments to NewMeter, but this will do for now).

fvm/executionParameters.go
```diff
@@ -11,6 +11,91 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
+func setMeterParameters(env *commonEnv) error {
+	// Don't enforce limits while setting up limits.
+	env.sth.DisableAllLimitEnforcements()
```
pattyshack, line 16: NOTE: This behaviour is new.  Previously, we were metering while setting up the meter ...

pattyshack:
hmm.  look like i did something stupid and all the tests broke.  i'll start a fresh pr =(

PR #2938, opened 2022-08-05
rm ServiceAccountEnabled.  It should always be on

pattyshack:


bluesign:
Isn't `service account` temporary workaround ? Eventually wouldn't we get rid of it?

janezpodhostnik:
This will break the playground API https://github.com/onflow/flow-playground-api/blob/1ad967055f31db8f1ce88e008960e5fc14a9fbd1/compute/computer.go#L76.

Not sure why the playground doesn't have the service account, but maybe it could have it....

pattyshack:
Thanks for pointing this out.  Seems like we can safely drop this from the playground as well

pattyshack:
we can't drop since service account contracts are not initialized in the playground.   instead of removing this option, i'll update the comment instead.

PR #2939, opened 2022-08-05
rm AccountFreezeEnabled.  It should always be on

pattyshack:


bluesign:
Same comment here, accountFreeze is temporary measure till we have BFT. I don't understand why do we remove those.

janezpodhostnik:
@bluesign yes, account freeze is a temporary measure. Its only possible to freeze an account through a service account multisig event, as the service account [MUST](https://github.com/onflow/flow-go/blob/5119cb5/fvm/transactionEnv.go#L462) be an authorizer on the transaction.

This PR only removes the unused setting weather this is supported or not (it was always on except for some tests). Enabling/Disabling required a code change and a redeploy... but if a code change and a redeploy is required, we might as well disable/enable it in the function itself.

fvm/transactionInvoker.go
```diff
@@ -256,50 +256,45 @@ var setAccountFrozenFunctionType = &sema.FunctionType{
 }
 
 func valueDeclarations(env Environment) []runtime.ValueDeclaration {
-	var predeclaredValues []runtime.ValueDeclaration
-
-	if env.AccountFreezeEnabled() {
```
janezpodhostnik, line 261: The initial Idea was to have `setAccountFrozen` available as a function only if the authorizer is the service account, but I think this got lost along the way.

With the current implementation `setAccountFrozen` is available as a function to everyone, but will panic if called for everyone except the service account. 

@turbolent @ramtinms do you see any problems with this.

janezpodhostnik (approved):
Just one general question. Looks good otherwise.

fvm/transactionInvoker.go
pattyshack: my personal preference is to move this definition onto cadence side, and define it as part of the standard lib (and expose a SetAccountFrozen method to the environment interface).  I think it's fine to expose this function to everyone with the understanding that calling will fail if you're not the service account.

bluesign: This is very bad idea in my opinion, I commented something related to this today: https://github.com/onflow/cadence/issues/331#issuecomment-1208339185

I think moving environment functions to Cadence side is very bad idea. Cadence should be the language, fvm / or other environment hosting Cadence should provide the functions depending on the environment.

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors merge

PR #2942, opened 2022-08-08
Dedup setExecutionParameters

pattyshack:
setExecutionParameter has been converted into a normal function

In the future, we may want to move this into engine, and only fetch the weights / limits once per block


pattyshack:
we should probably merge #2945 first, then move setMeterParameter out of NewTransactionEnvironment / into TransactionInvoker so that we only set the weight once per transaction

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors merge

PR #2944, opened 2022-08-08
[Dynamic Bootstrap] Update FullBlockHeight on startup

zhangchiqing:
This PR fixes an issue for dynamic bootstrapped node. 

When dynamic bootstrapped access node starts up from a certain block height, it will start fetching collections in finalized blocks. However, it's possible that the finalized block might have a collection which contains a transaction whose reference block height is a below the bootstrapped root block.

In order to prevent this case, we update the full block height after bootstrap so that dynamic bootstrapped access node will only  fetch missing collections for blocks starting from rootHeight + transactionExpiry.

### DIFF
--- a/cmd/dynamic_startup.go
+++ b/cmd/dynamic_startup.go
@@ -146,7 +146,7 @@ func ValidateDynamicStartupFlags(accessPublicKey, accessAddress string, startPha
 // 2. Target epoch == "current", wait until target phase == current phase before setting root snapshot
 // 3. Target epoch > current epoch (in future), wait until target epoch and target phase is reached before
 // setting root snapshot
-func DynamicStartPreInit(nodeConfig *NodeConfig) error {
+func DynamicStartPreInit(nodeConfig *NodeConfig, storage Storage) error {
 	ctx := context.Background()
 
 	log := nodeConfig.Logger.With().Str("component", "dynamic-startup").Logger()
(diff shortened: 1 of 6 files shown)


cmd/dynamic_startup.go
```diff
@@ -213,6 +213,24 @@ func DynamicStartPreInit(nodeConfig *NodeConfig) error {
 
 	// set the root snapshot in the config - we will use this later to bootstrap
 	nodeConfig.RootSnapshot = snapshot
+
+	// set the FullBlockHeight to be rootHeight+TransactionExpiry.
```
jordanschalm, line 217: This approach works. I'd be happy to merge this as-is to the relevant `v0.xx` branch to quickly unblock the current AN issues. For merging to master though, I think it would be better to adjust the responsibility for handling this edge case. Adding some suggestions below 👇 

In the current implementation in this PR, the DynamicStartup function implements mitigating this edge case by initializing the `last_full_block_height` index. I think this is non-ideal because:
* The `DynamicStartup` feature is not at all related to the `ingestion` engine which manages the `last_full_block_height` index
* The `DynamicStartup` feature is used by all node roles (not just ANs, which are the only role where this mitigation is relevant)

I think it would be better for the `ingestion` engine to internally handle the additional initialization logic for the `last_full_block_height` index. `ingestion.Engine` implements `Component`, so we can implement this in the engine's `Start` method:
```go 
func (e *Engine) Start(parent irrecoverable.SignalerContext) {
	rootBlock := e.state.Params().Root()
	rootSnapshot := e.state.AtBlockID(rootBlock.ID())
	if !protocol.IsSporkRootSnapshot(rootSnapshot) {
		// this is useful for dynamically bootstrapped access node, they will request missing collections. In order to ensure all txs
		// from the missing collections can be verified, we must ensure they are referening to known blocks.
		// That's why we set the full block height to be rootHeight + TransactionExpiry, so that we only request missing collections
		// in blocks above that height.
		firstFullHeight := rootBlock.Height + flow.DefaultTransactionExpiry
		e.blocks.UpdateLastFullBlockHeight(firstFullHeight)
	}
	
	e.Start(parent)	
}

jordanschalm (commented):
Thank you!

The approach looks functionally good to me. I'm happy to merge this to the `v0.26` branch to unblock AN operators. 

Before merging this change to `master` though, I think it is worthwhile to move the index initialization logic into the `ingestion` engine as mentioned in the comment .

PR #2945, opened 2022-08-09
Fix transaction invoker to not early exit once transaction started

pattyshack:
1. proc's fields were not correctly populated due to early exiting
2. also fixed meter tests (the tests were testing the wrong things)

(These issue was discovered when I tried to disable limiting while setting
meter weights/limits).


pattyshack:
i messed up rebasing.  i'll send out the same pr on a different branch

PR #646, opened 2021-04-21
[Consensus] Compliance Engine to use Message Queue

zhangchiqing:
For https://github.com/dapperlabs/flow-go/issues/5355

This PR introduces a generic message enqueue framework to build message queues for engines. It makes it easy to reuse code for engines to process different types of incoming messages concurrently.

Messages are stored in a bounded first-in-first-out queue in memory, so that the engine can process the same type of messages one after another without being overwhelmed. 

This PR is based on the [previous work](https://github.com/onflow/flow-go/pull/516)

PR #2951, opened 2022-08-09
[Execution] Worker-based Chunk Data Pack Requests

yhassanzadeh13:
This PR refactors the provider engine of the execution node to perform a worker-based chunk data pack request processing. Prior to this PR, each incoming chunk data pack request would fire up a separate `goroutine` on the execution node that could lead to resource exhaustion on the execution nodes under the large load of chunk data pack requests. 

This PR implements a HeroCache-based queue that queues up the incoming chunk requests at execution nodes to be picked by the concurrent workers. Hence rate-limiting the chunk data pack requests at the execution nodes. 

Additionally, this PR fixes false-positive linter errors and re-generates stale mocks.


engine/execution/provider/engine.go
```diff
@@ -24,150 +28,165 @@ import (
...
-		}
-	})
-}
+	engine.cm = cm.Build()
+	engine.Component = engine.cm
 
-func (e *Engine) ProcessLocal(event interface{}) error {
-	return e.unit.Do(func() error {
-		return e.process(e.me.NodeID(), event)
-	})
```
yhassanzadeh13, line 113: Upgraded `ProviderEngine` to `MessageProcessor` interface (from deprecated `Engine` interface), which no longer requires these methods.

module/mempool/queue/chunkDataPack.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+var _ mempool.ChunkDataPackRequestQueue = &ChunkDataPackRequestQueue{}
+
+func NewChunkDataPackRequestQueue(sizeLimit uint32, logger zerolog.Logger, collector module.HeroCacheMetrics) *ChunkDataPackRequestQueue {
+	return &ChunkDataPackRequestQueue{
+		cache: herocache.NewCache(
+			sizeLimit,
+			herocache.DefaultOversizeFactor,
+			heropool.LRUEjection,
```
SaveTheRbtz, line 30: given that this is a queue, should it be LIFO?  Or maybe, given that size is fixed, we can have a no-ejection mode?

module/mempool/queue/chunkDataPack.go
```diff
@@ -0,0 +1,120 @@
...
+		return false
+	}
+
+	req := chunkDataPackRequestEntity{
+		ChunkDataPackRequest: mempool.ChunkDataPackRequest{
+			ChunkId:     chunkId,
+			RequesterId: requesterId,
+		},
+		id: identifierOf(chunkId, requesterId),
+	}
```
SaveTheRbtz, line 54: nit: you can likely create this object outside of the lock

module/mempool/queue/chunkDataPack.go
```diff
@@ -0,0 +1,120 @@
...
+	defer c.mu.Unlock()
+
+	head, ok := c.cache.Head()
+	if !ok {
+		// cache is empty, and there is no head yet to pop.
+		return nil, false
+	}
+
+	c.cache.Remove(head.ID())
+	request := head.(chunkDataPackRequestEntity)
```
SaveTheRbtz, line 88: this looks like a good place for generics. You can consider creating a type-safe interface for the cache e.g. `NewCacheOf[T]`

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
+	return e.process(originID, event)
 }
 
 func (e *Engine) process(originID flow.Identifier, event interface{}) error {
-	ctx := context.Background()
 	switch v := event.(type) {
 	case *messages.ChunkDataRequest:
-		e.onChunkDataRequest(ctx, originID, v)
+		pushed := e.chdpRequestsQueue.Push(v.ChunkID, originID)
```
SaveTheRbtz, line 172: so if push fails here, who is responsible for re-pushing it?

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
 func (e *Engine) Done() <-chan struct{} {
-	return e.unit.Done()
+	return e.cm.Done()
+}
+
+// processChunkDataPackRequestWorker encapsulates the logic of a single (concurrent) worker that picks a chunk data pack
+// request from this engine's queue and processes it.
+func (e *Engine) processChunkDataPackRequestWorker(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+	ready()
+	ticker := time.NewTicker(e.chdpRequestProcessInterval)
```
SaveTheRbtz, line 139:
```suggestion
	ticker := time.NewTicker(e.chdpRequestProcessInterval)
	defer ticker.Stop()
```

engine/execution/provider/engine_test.go
```diff
@@ -88,16 +119,22 @@ func TestProviderEngine_onChunkDataRequest(t *testing.T) {
...
 		}
+		cancelCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		ctx, _ := irrecoverable.WithSignaler(cancelCtx)
+		e.Start(ctx)
 		// submit using origin ID with zero weight
 		unittest.RequireCloseBefore(t, e.Ready(), 100*time.Millisecond, "could not start engine")
-		e.onChunkDataRequest(context.Background(), originID, req)
+		require.NoError(t, e.Process(channels.RequestChunks, originID, req))
+		time.Sleep(1 * time.Second) // waits enough so that request is picked from queue.
```
jordanschalm, line 136: It would be nice to avoid needing to sleep in all these tests. I don't think we have a general approach for waiting until a particular message has been processed, but here we could drain in-progress requests before shutting down the engine. Suggestion below (using terminology from TODO link):
* The `RequestShoveler` thread listens for `context.Done()` and (1) closes the `RequestChannel` (2) exits
* The `RequestProcessor` threads range over the `RequestChannel` channel (do not listen for `context.Done()` and exit the range statement when the `RequestChannel` is emptied and closed

Then we will drain and process all the requests retrieved from the `MessageStore` before terminating the engine. This would be useful for testing, because we could:
* Call `Process` on some message
* Assert that `engine.MessageStore` is empty
* Call `cancel` to drain then terminate the engine - we are sure that the request will be processed before the engine shuts down, and we don't need to wait.

module/mempool/queue/chunkDataPack.go
yhassanzadeh13: Does not seem efficient: 

Moving `id` creation out of critical section like below. Then we generate `id` even if the cache is full and we literally never get to add the `req` to the cache. `id` creation involves encoding and hashing the `chunkId, requesterId` pair and by moving it out of critical section, we pay an extra encoding + hashing cost each time we _try_ adding request to the cache, even if the cache is full and the add transaction aborted. 
```
	id := identifierOf(chunkId, requesterId)
	
	c.mu.Lock()
	defer c.mu.Unlock()

	if c.cache.Size() >= c.sizeLimit {
		return false
	}

	req := chunkDataPackRequestEntity{
		ChunkDataPackRequest: mempool.ChunkDataPackRequest{
			ChunkId:     chunkId,
			RequesterId: requesterId,
		},
		id: id,
	}

	return c.cache.Add(req.id, req)
```

Please note that `cache` is not concurrency-safe, hence `c.cache.Add(req.id, req)` is a critical section that needs to be covered by a lock. The other alternative is to lock/unlock twice, which is more expensive, and also breaks the isolation of the `Push` operation as two duplicate push operations can reach the `point-2` line concurrently while the first one is succeeding and the other one is failing (due to the duplicate one just added). However, for passing from `point-1` to `point-2` in the code, we pay an extra lock acquisition as well as identifier computation for the duplicate one. 
```
        // point-1
	c.mu.Lock()
	if c.cache.Size() >= c.sizeLimit {
		return false
	}

	req := chunkDataPackRequestEntity{
		ChunkDataPackRequest: mempool.ChunkDataPackRequest{
			ChunkId:     chunkId,
			RequesterId: requesterId,
		},
		id: identifierOf(chunkId, requesterId),
	}
	defer c.mu.Unlock()
	
	// point-2
	c.mu.Lock()
	defer c.mu.Unlock()
	return c.cache.Add(req.id, req)
```

engine/execution/provider/engine.go
yhassanzadeh13: Push only fails on two circumstances: (1) queue is full (2) the pair `v.ChunkID` and `originID` already exists in the queue. 

In case (1), the request will be repeated by the verification node after a backoff. In other words, when execution node is dropping a chunk data pack request, it means the node has reached its capacity on processing such requests. So when the verification node does not hear back about its chunk data pack request, it realizes something is wrong with the execution node, so the verification node by its protocol is backing off exponentially and retry the request a few seconds (up to a minute) later. 

In case (2), the request already exists in the cache and will literally be processed by the execution node, so there is no need to add a duplicate one.

module/mempool/queue/chunkDataPack.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/887d541421ea5a7e7e5b493dbd2318058d72d61a
https://github.com/onflow/flow-go/pull/2951/commits/4bbeb862a6b17538d04c6091f138d27dbd11deac
https://github.com/onflow/flow-go/pull/2951/commits/bcb19eb60a7e9fd59f5712883ac35fe57c9a462c

cmd/observer/node_builder/observer_builder.go
```diff
@@ -249,7 +249,7 @@ func (builder *ObserverServiceBuilder) deriveUpstreamIdentities() error {
 }
 
 func (builder *ObserverServiceBuilder) buildFollowerState() *ObserverServiceBuilder {
-	builder.Module("mutable follower state", func(node *cmd.NodeConfig) error {
+	builder.FlowNodeBuilder.Module("mutable follower state", func(node *cmd.NodeConfig) error {
```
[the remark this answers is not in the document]
yhassanzadeh13: Reverted for all places.

engine/execution/provider/engine.go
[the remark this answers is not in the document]
yhassanzadeh13: Implemented `MessageHandler` for `Provider` engine here: https://github.com/onflow/flow-go/pull/2951/commits/37b80a33dfdf4e6db3e2edabc5e940fa80fd2e81
Implemented `MessageStore` for `HeroCache` here: https://github.com/onflow/flow-go/pull/2951/commits/456554ce2b4e052c61e6fc025404abc18aa70d55

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
+	receiptCon                 network.Conduit
+	state                      protocol.State
+	execState                  state.ReadOnlyExecutionState
+	chunksConduit              network.Conduit
+	metrics                    module.ExecutionMetrics
+	chdpRequestsQueue          mempool.ChunkDataPackRequestQueue
+	checkAuthorizedAtBlock     func(blockID flow.Identifier) (bool, error)
+	chdpQueryTimeout           time.Duration
+	chdpDeliveryTimeout        time.Duration
+	chdpRequestProcessInterval time.Duration
```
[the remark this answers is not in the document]
yhassanzadeh13: Removed `chdpRequestProcessInterval` in https://github.com/onflow/flow-go/pull/2951/commits/aa495cda11219f04313b1a199af4240e97ab2544, but we still need the `.*Timeout$` ones as they are used for observability to see how fast is an Execution Node processing chunk data pack requests. Please note that the `.*Timeout$` ones were there prior to this PR.

engine/execution/provider/engine_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/4b92f13885fa9eeb12a894b6d0f9773ab74f1959

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
+	engine.Component = engine.cm
 
-func (e *Engine) ProcessLocal(event interface{}) error {
-	return e.unit.Do(func() error {
-		return e.process(e.me.NodeID(), event)
-	})
+	return &engine, nil
 }
 
 // Ready returns a channel that will close when the engine has
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/cca0a619fa20ca0d1f24397dba474d5835f4e70e

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
-	chunkID := req.ChunkID
-
 	lg := e.log.With().
-		Hex("origin_id", logging.ID(originID)).
-		Hex("chunk_id", logging.ID(chunkID)).
+		Hex("origin_id", logging.ID(request.RequesterId)).
+		Hex("chunk_id", logging.ID(request.ChunkId)).
 		Logger()
-
 	lg.Info().Msg("received chunk data pack request")
```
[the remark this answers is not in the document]
yhassanzadeh13: outdated due to new refactoring

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,174 @@ import (
...
 
 	// increases collector metric
 	e.metrics.ChunkDataPackRequested()
+	chunkDataPack, err := e.execState.ChunkDataPackByChunkID(request.ChunkId)
 
-	chunkDataPack, err := e.execState.ChunkDataPackByChunkID(ctx, chunkID)
 	// we might be behind when we don't have the requested chunk.
 	// if this happen, log it and return nil
 	if errors.Is(err, storage.ErrNotFound) {
```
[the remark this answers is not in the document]
yhassanzadeh13: Refactored in https://github.com/onflow/flow-go/pull/2951/commits/9da2a9840aa14a56ba7776fff64224cd712f0a40. However, not pretty comfortable with this change, as one we go permissionless with VNs, it allows a malicious VN to conduct a DoS attack by sending invalid `OriginId`s.

engine/execution/provider/engine.go
```diff
@@ -202,39 +230,36 @@ func (e *Engine) onChunkDataRequest(
 	if sinceProcess > e.chdpQueryTimeout {
 		lg.Warn().Msgf("chunk data pack query takes longer than %v secs", e.chdpQueryTimeout.Seconds())
 	}
-
-	lg.Debug().Msg("chunk data pack response lunched to dispatch")
+	lg.Info().Msg("chunk data pack response lunched to dispatch")
```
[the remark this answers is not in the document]
yhassanzadeh13: Changed the log message here: https://github.com/onflow/flow-go/pull/2951/commits/72d8905dd3cdd12d6ddcd18286646fc9fcbe2089. But not agree with changing the log level, as the current `Info` one really helped so far debugging sealing-halt incidents.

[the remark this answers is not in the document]
jordanschalm: Sounds good - let's keep at info then 👍

engine/execution/provider/engine.go
```diff
@@ -202,39 +230,36 @@ func (e *Engine) onChunkDataRequest(
...
 
 	// sends requested chunk data pack to the requester
-	e.unit.Launch(func() {
-		deliveryStart := time.Now()
+	deliveryStart := time.Now()
 
-		err := e.chunksConduit.Unicast(response, originID)
+	err = e.chunksConduit.Unicast(response, request.RequesterId)
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/63617c466dfbc5898cbbe6c12a961e2981eb8770

engine/execution/provider/engine.go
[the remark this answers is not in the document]
jordanschalm: > one we go permissionless with VNs, it allows a malicious VN to conduct a DoS attack by sending invalid OriginIds

I agree - If we don't also differentiate between expected and unexpected error conditions, then DOS attacks are made easier by this suggestion.

From what I can see, the unexpected errors we want to propagate are:
* storage error reading chunk https://github.com/onflow/flow-go/commit/9da2a9840aa14a56ba7776fff64224cd712f0a40#diff-99567a576e6710d390c77cd207573be724e407c3fab2d99f3cbd4bc6334e7572R257
  * this is already handled correctly, and isn't subject to DOS attacks
* storage error checking if requester is authorized https://github.com/onflow/flow-go/commit/9da2a9840aa14a56ba7776fff64224cd712f0a40#diff-99567a576e6710d390c77cd207573be724e407c3fab2d99f3cbd4bc6334e7572R262
  * this could cause a crash for requests with either a non-existent or unknown chunk ID, or unknown Origin IDs
  * to fix this, we should document all expected error return types from `ensureAuthorized`, handle the expected cases within `onChunkDataRequest`, and bubble up the unexpected errors (eg. storage error besides `ErrNotFound`)

module/mempool/chunk_requests.go
```diff
@@ -96,3 +96,34 @@ type ChunkRequests interface {
 	// Size returns total number of chunk requests in the memory pool.
 	Size() uint
 }
+
+const DefaultChunkDataPackRequestQueueSize = 10e4
+
+// ChunkDataPackRequestQueue is a FIFO (first-in-first-out) size-bound queue for maintaining chunk data pack requests.
+// It is designed to be utilized at Execution Nodes to maintain and respond chunk data pack requests.
+type ChunkDataPackRequestQueue interface {
```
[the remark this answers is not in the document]
yhassanzadeh13: > Implement a version of this HeroCache-backed message queue which conforms to engine.MessageStore

https://github.com/onflow/flow-go/pull/2951/commits/86502928d294f5f897367b2f5e9378e48742cc02

> Use the MessageStore/MessageHandler pattern in provider/engine.go

Implemented MessageHandler for Provider engine here: https://github.com/onflow/flow-go/commit/37b80a33dfdf4e6db3e2edabc5e940fa80fd2e81
Implemented MessageStore for HeroCache here: https://github.com/onflow/flow-go/commit/456554ce2b4e052c61e6fc025404abc18aa70d55

engine/execution/provider/engine_test.go
```diff
@@ -243,24 +328,31 @@ func TestProviderEngine_onChunkDataRequest(t *testing.T) {
 	})
 
 	t.Run("reply to chunk data pack request only when authorized", func(t *testing.T) {
+		currentAuthorizedState := true
 
 		ps := new(mockprotocol.State)
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/98c42a9ced507f47aa8c84ff352b1beaa195990e

module/mempool/queue/chunkDataPack.go
yhassanzadeh13: Thanks for the suggestion, as it is out of the scope of this PR, I created this tracking issue https://github.com/dapperlabs/flow-go/issues/6362 to address it separately.

engine/execution/provider/engine.go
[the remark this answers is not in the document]
yhassanzadeh13: Per our conversation, created this https://github.com/dapperlabs/flow-go/issues/6363 to address proper error handing on the provider engine separately. Also, for now, reverted back error handling to the same state as `master` hence not causing crash incidents.

fvm/mock/account_interface.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.13.1. DO NOT EDIT.
+// Code generated by mockery v2.13.0. DO NOT EDIT.
```
zhangchiqing, line 1: [please upgrade your mockery](https://github.com/onflow/flow-go/blob/master/Makefile#L61)

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,222 @@ import (
...
+
+	e.log.Debug().Msg("process chunk data pack request shovller worker started")
+
+	for {
+		select {
+		case <-e.chdpRequestHandler.GetNotifier():
+			// there is at list a single chunk data pack request queued up.
+			e.shovelChunkDataPackRequests()
+		case <-ctx.Done():
+			close(e.chdpRequestChannel)
```
zhangchiqing, line 157: This means, if there are 500 request in the queue, and we shut down the engine, it won't stop until `shovelChunkDataPackRequests` finish processing all of the 500 requests in the queue. 

Why not dropping all the requests in the queue?

yhassanzadeh13: cc @jordanschalm

jordanschalm: > if there are 500 request in the queue, and we shut down the engine, it won't stop until shovelChunkDataPackRequests finish processing all of the 500 requests in the queue.

Let's differentiate between the "message queue" and the "internal channel":
* the message queue is a HeroCache instance populated by incoming messages from the networking layer, it usually has a large size limit
* the internal channel (`chdpRequestChannel`) is a Go channel populated by the `processQueuedChunkDataPackRequestsShovelerWorker` goroutine. It can be much smaller (eg. currently it's size limit is [equal to the number of worker goroutines](https://github.com/onflow/flow-go/pull/2951/files#diff-99567a576e6710d390c77cd207573be724e407c3fab2d99f3cbd4bc6334e7572R116))

With the current implementation, we will wait for the smaller internal channel to drain, not the larger message queue. 

For context, [here](https://github.com/onflow/flow-go/pull/2951#discussion_r943982008) is the original comment suggesting that we drain the internal channel before stopping the engine. The reasoning for this was to make testing more responsive. Previously we needed to introduce a 1-second sleep into the test cases, because the test did not have a way to determine when a message was processed.

I agree, it is not desirable for an engine to continue processing message after it is stopped. Though personally I think the relatively small added time to stop the engine is worth the testing simplification.

module/metrics/execution.go
```diff
@@ -375,11 +378,34 @@ func NewExecutionCollector(tracer module.Tracer) *ExecutionCollector {
...
+		Help:      "the total time spent on dispatching a chunk data pack response on the network milliseconds",
+		Buckets:   []float64{100, 500, 1000, 1500, 2000, 2500, 3000, 6000},
+	})
+
+	chunkDataPackQueryTime := promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: namespaceExecution,
+		Subsystem: subsystemProvider,
+		Name:      "chunk_data_pack_query_time_ms",
+		Help:      "the total time spent on querying a chunk data pack from database",
+		Buckets:   []float64{100, 500, 1000, 1500, 2000, 2500, 3000, 6000},
```
jordanschalm, line 408: Suggesting to use second units here, to align with [Prometheus best practices](https://prometheus.io/docs/practices/naming/#metric-names):
```suggestion
	chunkDataPackResponseDispatchTime := promauto.NewHistogram(prometheus.HistogramOpts{
		Namespace: namespaceExecution,
		Subsystem: subsystemProvider,
		Name:      "chunk_data_pack_response_dispatch_time_seconds",
		Help:      "the total time spent on dispatching a chunk data pack response on the network in seconds",
		Buckets:   []float64{.1, .5, 1, 1.5, 2, 2.5, 3, 6},
	})

	chunkDataPackQueryTime := promauto.NewHistogram(prometheus.HistogramOpts{
		Namespace: namespaceExecution,
		Subsystem: subsystemProvider,
		Name:      "chunk_data_pack_query_time_seconds",
		Help:      "the total time spent on querying a chunk data pack from database in seconds",
		Buckets:   []float64{.01, .05, .1, .5, 1, 2, 5},
```

Also suggesting to adjust down the database query buckets, as this should typically be faster than what was the smallest bucket of 100ms.

module/metrics/execution.go
```diff
@@ -375,11 +378,34 @@ func NewExecutionCollector(tracer module.Tracer) *ExecutionCollector {
...
+		Name:      "chunk_data_pack_requested_total",
+		Help:      "the total number of chunk data pack requests processed by provider engine",
+	})
+
+	chunkDataPackResponseDeliveredTotal := promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: namespaceExecution,
+		Subsystem: subsystemProvider,
+		Name:      "chunk_data_pack_delivered_total",
+		Help:      "the total number of chunk data pack responses delivered by provider engine",
+	})
```
jordanschalm, line 393:
```suggestion
```

The histogram below will count the number of observations as `execution_provider_ chunk_data_pack_response_dispatch_time_ms_count`, so this separate counter is unnecessary

module/metrics/execution.go
```diff
@@ -726,10 +755,24 @@ func (ec *ExecutionCollector) RuntimeTransactionInterpreted(dur time.Duration) {
...
+// It increases the request processed counter by one.
+func (ec *ExecutionCollector) ChunkDataPackRequestProcessed() {
+	ec.chunkDataPackRequestProcessedTotal.Inc()
+}
+
+// ChunkDataPackResponseDispatchedInNetwork is executed every time a chunk data pack request is successfully replied.
+// It records the time it took to dispatch the reply on the network.
+// It also increases the response counter by one.
+func (ec *ExecutionCollector) ChunkDataPackResponseDispatchedInNetwork(dispatchTime time.Duration) {
+	ec.chunkDataPackResponseDispatchTime.Observe(float64(dispatchTime.Milliseconds()))
```
jordanschalm, line 768:
```suggestion
	ec.chunkDataPackResponseDispatchTime.Observe(float64(dispatchTime.Seconds()))
```
Consistency with prometheus naming

module/metrics/execution.go
```diff
@@ -726,10 +755,24 @@ func (ec *ExecutionCollector) RuntimeTransactionInterpreted(dur time.Duration) {
...
+	ec.chunkDataPackResponseDeliveredTotal.Inc()
+}
+
+// ChunkDataPackRetrievedFromDatabase is executed every time a chunk data pack is queried from database.
+// It records the time it took to query the chunk data pack.
+func (ec *ExecutionCollector) ChunkDataPackRetrievedFromDatabase(queryTime time.Duration) {
+	ec.chunkDataPackQueryTime.Observe(float64(queryTime.Milliseconds()))
```
jordanschalm, line 775:
```suggestion
	ec.chunkDataPackQueryTime.Observe(float64(queryTime.Seconds()))
```

Consistency with prometheus naming

engine/execution/provider/engine.go
jordanschalm, line 157:
```suggestion
			// close the internal channel, the workers will drain the channel before exiting
			close(e.chdpRequestChannel)
```

engine/execution/provider/engine.go
```diff
@@ -24,150 +28,222 @@ import (
...
-
-	processStart := time.Now()
-
-	// extracts list of verifier nodes id
-	chunkID := req.ChunkID
+// onChunkDataRequest receives a request for a chunk data pack,
+// and if such a chunk data pack is available in the execution state, it is sent to the requester node.
+// It returns an error if either of the following situatations hit:
+// 1. The chunk data pack exists, but not retrievable from the storage (e.g., inconsistent storage state).
+// 2. The requester for chunk data pack is not authorized in this current epoch.
```
jordanschalm, line 233:
```suggestion
// TODO improve error handling https://github.com/dapperlabs/flow-go/issues/6363
```

engine/execution/provider/engine_test.go
```diff
@@ -269,38 +313,43 @@ func TestProviderEngine_onChunkDataRequest(t *testing.T) {
 		blockID := unittest.IdentifierFixture()
 
 		execState.On("GetBlockIDByChunkID", chunkID).Return(blockID, nil)
-		ps.On("AtBlockID", blockID).Return(ss)
-
+		ps.On("AtBlockID", blockID).Return(ss).Once()
 		ss.On("Identity", originIdentity.NodeID).Return(originIdentity, nil).Once()
+
+		// channel tracking for the first chunk data pack request responded.
+		responded := make(chan struct{})
```
jordanschalm, line 320: This will work, however the fact that this is necessary in this test case suggests to me that we have a race condition in the other test cases. 

My suggestion to split the "request_shoveler" goroutine from the "request_processor" goroutines was so that we could generally guarantee that a request is processed before we stop the engine, by:
* Call `Process` on some message
* Wait for `engine.MessageStore` to be empty
* Call `cancel` to drain then terminate the engine

I think we are missing the middle step in the testing code:
> Wait for `engine.MessageStore` to be empty

In this test case, the `responded` channel guarantees that the request is processed before the test ends. But we don't have that guarantee for the other test cases. 

If we add this middle step to all test cases (both success and failure), we guarantee the request is processed before the test ends in all cases, removing this race condition. (Adding a concrete suggestion in next comment).

engine/execution/provider/engine_test.go
```diff
@@ -269,38 +313,43 @@ func TestProviderEngine_onChunkDataRequest(t *testing.T) {
...
+		ctx, _ := irrecoverable.WithSignaler(cancelCtx)
+		e.Start(ctx)
+		// submit using non-existing origin ID
 		unittest.RequireCloseBefore(t, e.Ready(), 100*time.Millisecond, "could not start engine")
+		require.NoError(t, e.Process(channels.RequestChunks, originIdentity.NodeID, req))
 
-		// an authorized request followed by an unauthorized one
-		e.onChunkDataRequest(context.Background(), originIdentity.NodeID, req)
+		// waits till first chunk data pack is replied and then makes the requester unauthorized.
+		unittest.RequireCloseBefore(t, responded, 1*time.Second, "could not get first chunk data pack responded")
```
jordanschalm, line 348:
```suggestion
		// waits till request is ingested to a worker for processing
		assert.Eventually(t, func() bool { return chdpQueue.Size() == 0 }, time.Second, 10*time.Millisecond)
```
Suggesting to implement this synchronization in a way that will also prevent race conditions in the unhappy path cases. (concrete suggestion for previous comment)

* `chdpQueue` is the queue passed to the `provider.Engine` constructor

module/mempool/queue/chunkDataPack.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Get pops the queue, i.e., it returns the head of queue, and updates the head to the next element.
+// Boolean return value determines whether pop is successful, i.e., poping an empty queue returns false.
+func (c *ChunkDataPackRequestMessageStore) Get() (*engine.Message, bool) {
+	head, ok := c.pop()
+	if !ok {
+		return nil, false
+	}
+	return &engine.Message{
```
jordanschalm, line 42: From my perspective it would be ideal if the Herocache-backed implementation of `MessageStore` was generic to message type (like the corresponding [FifoQueue](https://github.com/onflow/flow-go/blob/e1099a88efe6ad90c0afb8e4679990e6f97857c4/engine/fifoqueue.go#L7-L22)). Also hopefully a fruitful place to apply generics, like Alexey mentioned. 

Out of scope for this PR, just noting for consideration when/if we make use of Herocache for a message queue in a different context, in the future.

module/metrics/execution.go
yhassanzadeh13: @jordanschalm I think you mean to refactor it either to `ec.chunkDataPackResponseDispatchTime.Observe(float64(dispatchTime.Seconds()))` or `ec.chunkDataPackResponseDispatchTime.Observe(float64(float64(dispatchTime.Milliseconds())/1000)`, right?

jordanschalm: Exactly, that's what I meant - thanks for catching!

pattyshack:
I'm not familiar with this part of the code base.  I have a few ignorant questions:
1. why do you need to implement the queue using herocache?  Implementing a queue using cache data structure (map list + link list) seems expensive.  Why can't this just be a regular go channel (or a slice guarded by mutex)?
2. why do you need both chdpRequestQueue and chdpRequestChannel?  They basically do the same thing?

module/metrics/execution.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/383805d7d94200444432dea05d205138884636ed

module/mempool/herocache/backdata/heropool/pool.go
```diff
@@ -29,6 +30,16 @@ type poolEntity struct {
 	node link
 }
 
+type PoolAddResult struct {
```
zhangchiqing, line 33: usually golang encourages to use multiple return values to than grouping into a struct, in order to avoid allocating on heap. 

Grouping data into a struct is ok for other module, but I though herocache is trying to avoiding allocating memory on heap, right?

engine/execution/provider/engine_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/f32db517ed868b7636ee1aa577c2d2889d2e6bc9

engine/execution/provider/engine_test.go
yhassanzadeh13: Added the synchronization in the middle https://github.com/onflow/flow-go/pull/2951/commits/f32db517ed868b7636ee1aa577c2d2889d2e6bc9, removed this sync here: https://github.com/onflow/flow-go/pull/2951/commits/927d5399ca690d77387b1397899feeff402ce3b4

module/mempool/queue/chunkDataPack.go
yhassanzadeh13: Noted here: https://github.com/dapperlabs/flow-go/issues/6362#issuecomment-1227880978

SaveTheRbtz:
Patrick has a good point.  IIUC this PR has two parts:
* limited concurrency
* limited buffering

Both of these can be accomplished by a buffered channel with a pool of worker-goroutines for processing.

yhassanzadeh13:
> Patrick has a good point. IIUC this PR has two parts:
> 
> * limited concurrency
> * limited buffering
> 
> Both of these can be accomplished by a buffered channel with a pool of worker-goroutines for processing.

A buffered channel does not de-duplicate the repeated chunk data pack requests coming from the same requester for the same chunk. In production, a requester keeps repeating its request till it gets the result. If the requesters are just buffered in a channel, and no deduplication is involved, duplicate requests may be buffered and occupy the entire capacity of the channel while fresh and new requests are discarded. Moreover, it is enough for one copy of a request to buffer as long as it has not yet been processed, and buffering duplicate requests also means a redundant workload for ENs. 

HeroCache serves as the point of deduplication by optimizing memory and CPU cost: 
- Compared to other alternatives, e.g., a `map`, HeroCache has the minimum footprint on the heap and optimizes GC time.
-  HeroCache also computes the unique identifier corresponding to a pair of requester id and chunk id only once, and caches that internally, so any read or write query does not involve re-computing the identifier, which involves an encoding to byte array and taking the hash of the byte representation.

module/mempool/herocache/backdata/heropool/pool.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/4892426aeace2fbeb876b5f51706e8ed8089d6bd

engine/execution/provider/engine.go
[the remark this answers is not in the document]
zhangchiqing: I feel the metrics for querying and delivering chdp is a bit too granular to be useful.

Just like in a web app, we probably care more about the total response time of a request, rather than the individule durations of saving the data to database and sending the message over network. If the total response time is not higher than a threshold, we might not look into those more granular metrics. In other words, we might only look into those granular metrics if the total response time is too high, but in that case, i think we are probably debugging problems, and logs would be more efficient in debugging, as we can associate a certain high response time to a chunk data pack id, and see whether it's due to saving to db takes to long, or delivering over network takes too long.

That being said, IMO, a metrics for total response time of a chunk data pack is enough, and we could log the db query and network delivery along with chdp ID for debugging purpose.

jordanschalm:
@pattyshack @SaveTheRbtz Will also link some more general context from a while back:
* [FLIP 343 - Handing Messages from Networking Layer to Engines (Core Protocol)](https://github.com/onflow/flow/blob/master/flips/20210201-consuming-messages-from-network-layer.md)
  * [Discussion](https://github.com/onflow/flow/pull/343)
* [PR #646 - First implements FLIP for compliance engine, defines `MessageStore` interface](https://github.com/onflow/flow-go/pull/646)

engine/execution/provider/engine_test.go
```diff
@@ -71,6 +72,11 @@ func TestProviderEngine_onChunkDataRequest(t *testing.T) {
 		// submit using origin ID with invalid role
 		unittest.RequireCloseBefore(t, e.Ready(), 100*time.Millisecond, "could not start engine")
 		require.NoError(t, e.Process(channels.RequestChunks, originID, req))
+
+		require.Eventually(t, func() bool {
+			return requestQueue.Size() == uint(0) // ensuring all requests have been picked up from the queue.
+		}, 1*time.Second, 100*time.Millisecond)
```
jordanschalm, line 78:
```suggestion
		}, 1*time.Second, 10*time.Millisecond)
```
Just guessing, but seems likely this will usually take <100ms

module/mempool/herocache/backdata/heropool/pool.go
```diff
@@ -156,37 +152,35 @@ func (p Pool) Head() (flow.Entity, bool) {
...
-	ejection := false
-
+//
+// The first boolean return value says whether pool has an available slot. Pool goes out of available slots if
+// it is full and no ejection is set.
+//
+// If the pool has an available slot (either empty or by ejection), then the second boolean returned value determines
+// whether an ejection happened to make one slot free or not. Ejection happens if there is no available
+// slot, and there is an ejection mode set.
+func (p *Pool) sliceIndexForEntity() (EIndex, bool, bool) {
```
jordanschalm, line 162:
```suggestion
// The first boolean return value (hasAvailableSlot) says whether pool has an available slot. Pool goes out of available slots if
// it is full and no ejection is set.
//
// If the pool has an available slot (either empty or by ejection), then the second boolean returned value (ejectionOccurred) determines
// whether an ejection happened to make one slot free or not. Ejection happens if there is no available
// slot, and there is an ejection mode set.
func (p *Pool) sliceIndexForEntity() (i EIndex, hasAvailableSlot bool, ejectionOccurred bool) {
```

module/mempool/herocache/backdata/heropool/pool.go
```diff
@@ -96,34 +86,40 @@ func (p *Pool) initFreeEntities() {
...
-		p.poolEntities[result.EntityIndex].node.next.setUndefined()
-		p.poolEntities[result.EntityIndex].node.prev.setUndefined()
+// Add writes given entity into a poolEntity on the underlying entities linked-list.
+// The first boolean return value says whether pool has an available slot. Pool goes out of available slots if
+// it is full and no ejection is set.
+//
+// If the pool has an available slot (either empty or by ejection), then the second boolean returned value determines
+// whether an ejection happened to make one slot free or not. Ejection happens if there is no available
+// slot, and there is an ejection mode set.
+func (p *Pool) Add(entityId flow.Identifier, entity flow.Entity, owner uint64) (EIndex, bool, bool) {
```
jordanschalm, line 96:
```suggestion
func (p *Pool) Add(entityId flow.Identifier, entity flow.Entity, owner uint64) (i EIndex, slotAvailable bool, ejectionOccurred bool) {
```

engine/execution/provider/engine.go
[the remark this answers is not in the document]
jordanschalm: 👍  Good point Leo. I agree, one metric for total response time seems sufficient. The logs contain more granular timing, and more info about the request, so they're better suited for investigating outliers, which is what the original "log when exceeding the _timeout_" behaviour was geared toward.

jordanschalm (approved):
Just some minor stylistic comments

engine/execution/provider/engine.go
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/0f10739e8ecdcfd85c424158a279f803239b2885

module/mempool/herocache/backdata/heropool/pool.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/7f09e896b12a25da2119e81aa20814891b5a4ff4

module/mempool/herocache/backdata/heropool/pool.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/7f09e896b12a25da2119e81aa20814891b5a4ff4

engine/execution/provider/engine_test.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/2951/commits/63d9163669da376b331e728c89c949c497e9b719

pattyshack:
IC.  Queue implies the same number of items entering/leaving the collection, hence I assumed the items are unique.  Perhaps a different name is more appropriate.

This code doesn't seem to do a very good job at de-duplication.  While a worker is processing a chunk, an equivalent chunk can reenter the queue.

You may be better off using a real/normal queue for buffering, then maintain a separate data structure shared by the workers which keeps track of an item's id for a fixed duration time range, starting from the initial processes time, and dedup based on that.

yhassanzadeh13:
> This code doesn't seem to do a very good job at de-duplication. While a worker is processing a chunk, an equivalent chunk can reenter the queue.
> 
> You may be better off using a real/normal queue for buffering, then maintain a separate data structure shared by the workers which keeps track of an item's id for a fixed duration time range, starting from the initial processes time, and dedup based on that.

 https://github.com/dapperlabs/flow-go/issues/6373

PR #2957, opened 2022-08-10
[Access] Fix race condition in AN execution receipt metrics collection

peterargue:
https://github.com/onflow/flow-go/pull/2692 introduced a new metric to track the highest block height for `ExecutionReceipt`s received by the AN's ingestion engine. This is intended to be use to compare with the latest finalized block height to detect cases where the the `ExecutionReceipt`s fall behind finalization.

There was an uncaught race condition where the receipt may be received before the block is finalized and processed through the sync engine, causing fatal errors on the AN. 

This PR fixes that by using the latest finalized height in the case where the block is not yet available in the db.


koko1123 (approved):

vishalchangrani (approved):

engine/access/ingestion/engine.go
```diff
@@ -427,33 +432,35 @@ func (e *Engine) handleExecutionReceipt(originID flow.Identifier, r *flow.Execut
...
 	// retrieve the block
 	b, err := e.blocks.ByID(r.ExecutionResult.BlockID)
+
 	if errors.Is(err, storage.ErrNotFound) {
 		e.blocksToMarkExecuted.Add(r.ExecutionResult.BlockID, now)
+
+		// execution receipts are received as they are sent by ENs, which may be before the block
+		// is finalized and processed through the sync engine. When that happens, just use the
+		// latest finalized height.
+		e.transactionMetrics.UpdateExecutionReceiptMaxHeight(e.highestFinalizedHeight)
```
zhangchiqing, line 452: do we want to report this "wrong" statement. 

A finalized block doesn't mean it's executed. A sealed block is executed, but not finalized.

peterargue: I updated it to just update the metric when the block is finalized

module/metrics/transaction.go
```diff
@@ -277,5 +280,8 @@ func (tc *TransactionCollector) TransactionExpired(txID flow.Identifier) {
 }
 
 func (tc *TransactionCollector) UpdateExecutionReceiptMaxHeight(height uint64) {
-	tc.maxReceiptHeight.Set(float64(height))
+	if height > tc.maxReceiptHeightValue {
```
zhangchiqing, line 283: would this be called concurrently?

Also consider using this concurrent safe module
https://github.com/onflow/flow-go/blob/master/engine/consensus/sealing/counters/monotonous_counter.go

zhangchiqing:
```
counter := NewMonotonousCounter(0)

...
stored := counter.Set(height)
if stored {
	tc.maxReceiptHeight.Set(float64(height))
}

```

zhangchiqing (approved):

module/metrics/transaction.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/rs/zerolog"
 
+	"github.com/onflow/flow-go/engine/consensus/sealing/counters"
```
zhangchiqing, line 10: cc @durkmurder 

Now I feel this module can be moved to a shared package, as its usage is quite common across the codebase. 

can be done in a separate PR

vishalchangrani (approved):

vishalchangrani:
bors merge

zhangchiqing:
bors merge

PR #2960, opened 2022-08-11
[v0.26] Hotfix: Skip requesting collections when below full height

jordanschalm:
Extends https://github.com/onflow/flow-go/pull/2946. This eliminates a case where the `ingestion` engine will request collections below the last full block height.

### DIFF
--- a/engine/access/ingestion/engine.go
+++ b/engine/access/ingestion/engine.go
@@ -384,6 +384,7 @@ func (e *Engine) processFinalizedBlock(blockID flow.Identifier) error {
 		return fmt.Errorf("could not get last full block height: %w", err)
 	}
 	if block.Header.Height <= lastFullBlockHeight {
+		e.log.Info().Msgf("skipping requesting collections for finalized block below last full block height (%d<=%d)", block.Header.Height, lastFullBlockHeight)
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


engine/access/ingestion/engine.go
```diff
@@ -376,6 +376,17 @@ func (e *Engine) processFinalizedBlock(blockID flow.Identifier) error {
...
 
+	// skip requesting collections, if this block is below the last full block height
+	// this means that either we have already received these collections, or the block
+	// may contain unverifiable guarantees (in case this node has just joined the network)
+	lastFullBlockHeight, err := e.blocks.GetLastFullBlockHeight()
+	if err != nil {
+		return fmt.Errorf("could not get last full block height: %w", err)
+	}
+	if block.Header.Height <= lastFullBlockHeight {
+		return nil
```
kc1116, line 387: Should we add a log message here, something like "skipping requesting collections for block height < last full block height"?

kc1116 (approved):

engine/access/ingestion/engine.go
jordanschalm, line 387:
```suggestion
		e.log.Warn().Msgf("skipping requesting collections for finalized block below last full block height (%d<=%d)", block.Header.Height, lastFullBlockHeight)
		return nil
```

zhangchiqing (approved):
Good catch

PR #2961, opened 2022-08-11
[Testing] End-to-End TPS benchmark

AndrewM-SDET:
Run CI tests for TPS performance. The test is set to run for 10 minutes taking slices of each 2 minutes to determine the average TPS during the slice. This will run on the same schedule as existing CI git actions.


SaveTheRbtz:
Please split this into logical commits (either within this PR or, ideally, in multiple smaller PRs)

Also please use "dashes" for make target names e.g. `tps-benchmark`

integration/loader/main.go
```diff
@@ -193,6 +192,8 @@ func main() {
 				lg.Stop()
 			}
 		}
+
+		cancel()
```
pattyshack, line 196: why do we need to run this big loop in a goroutine?  can you simplify this to just run directly in the main function?

I think we can completely remove the context.  it doesn't provide any value since it never times out, and is only cancel here.

integration/benchmark/main.go
```diff
@@ -0,0 +1,376 @@
...
+	StartTime           time.Time
+	EndTime             time.Time
+	InputTps            float64
+	OutputTps           float64
+	ProStartTransaction float64
+	ProEndTransaction   float64
+}
+
+func main() {
+	sleep := flag.Duration("sleep", 0, "duration to sleep before benchmarking starts")
```
pattyshack, line 49: a lot of these flags seems like duplicates from integration/loader/main.go.  Can we refactor the shared parameter into a struct and the shared flag setup into a function?

integration/benchmark/main.go
```diff
@@ -0,0 +1,376 @@
...
+	supervisorAccessAddr := accessNodeAddrs[0]
+	if len(accessNodeAddrs) > 1 {
+		supervisorAccessAddr = accessNodeAddrs[1]
+	}
+	supervisorClient, err := client.NewClient(supervisorAccessAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatal().Err(err).Msgf("unable to initialize Flow supervisor client")
+	}
+
+	go func() {
```
pattyshack, line 160: this also looks identical to the other main.  refactor plz

integration/benchmark/main.go
AndrewM-SDET: This file is intentionally a modified copy of loader/main.go that is used specifically for CI/benchmark related work. We can refactor them both to use some shared code but it is outside the scope of this issue. We can create a new issue and add it to the backlog.

integration/loader/main.go
AndrewM-SDET: If we remove the context we lose in-line logging and prometheus data. cancel() is called here to prevent a deadlock that existed previously.

integration/localnet/Makefile
```diff
@@ -75,3 +75,14 @@ stop:
...
+
+.PHONY: tps-test
+tpsTest:
+	go run --tags relic ../benchmark -log-level info -tps 25 -tps-durations 10m
+
+.PHONY: tps-test-ci
+end2endTest:
+	make init
+	make start
+	go run --tags relic ../benchmark -ci-run -gitSha $(COMMIT_SHA)
```
SaveTheRbtz, line 87: We already have it defined in this makefile.
```suggestion
	go run --tags relic ../benchmark -ci-run -gitSha ${COMMIT}
```

integration/benchmark/main.go
AndrewM-SDET: I have created a new issue to track this work, separate from this PR.

https://github.com/onflow/flow-go/issues/2965

integration/localnet/Makefile
```diff
@@ -75,3 +75,14 @@ stop:
 .PHONY: load
 load:
 	go run --tags relic ../loader -log-level info -tps 1,0,10,0,100 -tps-durations 30s,10s,30s,10s
+
+.PHONY: tps-test
+tpsTest:
+	go run --tags relic ../benchmark -log-level info -tps 25 -tps-durations 10m
+
+.PHONY: tps-test-ci
+end2endTest:
```
SaveTheRbtz, line 84:
```suggestion
tps-test-ci:
```

integration/localnet/Makefile
```diff
@@ -75,3 +75,14 @@ stop:
 .PHONY: load
 load:
 	go run --tags relic ../loader -log-level info -tps 1,0,10,0,100 -tps-durations 30s,10s,30s,10s
+
+.PHONY: tps-test
+tpsTest:
```
SaveTheRbtz, line 80:
```suggestion
tps-test:
```

integration/benchmark/main.go
pattyshack: deduplicating the code later on will take more effort than doing some simple refactoring now, especially once the code diverge.  let's just fix it now.

I don't mind leaving the flags as is, but as least refactor the big ass loop as the end of main

integration/benchmark/main.go
```diff
@@ -0,0 +1,372 @@
...
+	ProEndTransaction   float64
+}
+
+func main() {
+	// holdover flags from loader/main.go
+	sleep := flag.Duration("sleep", 0, "duration to sleep before benchmarking starts")
+	loadTypeFlag := flag.String("load-type", "token-transfer", "type of loads (\"token-transfer\", \"add-keys\", \"computation-heavy\", \"event-heavy\", \"ledger-heavy\", \"const-exec\")")
+	chainIDStr := flag.String("chain", string(flowsdk.Emulator), "chain ID")
+	access := flag.String("access", net.JoinHostPort("127.0.0.1", "3569"), "access node address")
+	serviceAccountPrivateKeyHex := flag.String("servPrivHex", unittest.ServiceAccountPrivateKeyHex, "service account private key hex")
```
SaveTheRbtz, line 56: these can be hardcoded for the CI test.

integration/benchmark/main.go
```diff
@@ -0,0 +1,372 @@
...
+	argSizeInByteInConstExecTx := flag.Uint("const-exec-arg-size", 100, "byte size of tx argument for each constant exec transaction to generate")
+	payerKeyCountInConstExecTx := flag.Uint("const-exec-payer-key-count", 2, "num of payer keys for each constant exec transaction to generate")
+
+	// CI relevant flags
+	tpsFlag := flag.String("tps", "300", "transactions per second (TPS) to send, accepts a comma separated list of values if used in conjunction with `tps-durations`")
+	tpsDurationsFlag := flag.String("tps-durations", "10m", "duration that each load test will run, accepts a comma separted list that will be applied to multiple values of the `tps` flag (defaults to infinite if not provided, meaning only the first tps case will be tested; additional values will be ignored)")
+	ciFlag := flag.Bool("ci-run", false, "whether or not the run is part of CI")
+	leadTime := flag.String("leadTime", "30s", "the amount of time before data slices are started")
+	sliceSize := flag.String("sliceSize", "2m", "the amount of time that each slice covers")
+	gitSha := flag.String("gitSha", "", "the git hash of the run, used for BigQuery result tracking")
```
SaveTheRbtz, line 74: commit and semver are injected during the build, so you can use them through cmd/build/version.go

integration/benchmark/main.go
AndrewM-SDET: The loop is actually one of the only places that actually is different between the files because of the additional data needed that is then passed forward into BigQuery. If we try to reconverge those we lose the ability to separate out our TPS results by test case. The earlier parts of the file are the ones that are more static and able to be deduped.

SaveTheRbtz (approved):

integration/benchmark/main.go
pattyshack: yes, i noticed that difference and the lack infinite sleep

you can can easily combine the two loop if you refactor calculateTpsSlices & prepareDataForBigQuery into an object, say TPSWriter, and just provide a different implementation in the other loop

integration/loader/main.go
AndrewM-SDET: Simplified loop down to single case that will be used in CI, the subroutine is still required to allow for inline logging and pushing of data to local Prometheus

integration/benchmark/main.go
AndrewM-SDET: Rather than try to force a common structure for the two files that will further diverge I have instead removed all extraneous code from the CI implementation that would never be used. We now only have a single load case instead of parsing for multiples which removes the need for looping at all. Additionally many of the flags have been moved to hardcoded values as in a CI environment they will never change. They are still held as variables rather than hardcoded at point of use for ease of maintainability and updating for possible future use cases (e.g. a CI test pointed at testnet rather than localnet).

integration/benchmark/main.go
AndrewM-SDET: CLI specific code has been removed from the CI version of the loader entrypoint, still keeping the issue I had previously created in order to track further refactor as there is still some shared code that could be further simplified in the future.

integration/benchmark/main.go
```diff
@@ -0,0 +1,376 @@
...
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+type LoadCase struct {
+	tps      int
+	duration time.Duration
+}
+
+type dataSlice struct {
+	GitSha              string
```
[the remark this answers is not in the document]
AndrewM-SDET: Added

integration/benchmark/main.go
```diff
@@ -0,0 +1,376 @@
...
+	tpsDurationsFlag := flag.String("tps-durations", "10m", "duration that each load test will run, accepts a comma separted list that will be applied to multiple values of the `tps` flag (defaults to infinite if not provided, meaning only the first tps case will be tested; additional values will be ignored)")
+	chainIDStr := flag.String("chain", string(flowsdk.Emulator), "chain ID")
+	access := flag.String("access", net.JoinHostPort("127.0.0.1", "3569"), "access node address")
+	serviceAccountPrivateKeyHex := flag.String("servPrivHex", unittest.ServiceAccountPrivateKeyHex, "service account private key hex")
+	logLvl := flag.String("log-level", "info", "set log level")
+	metricport := flag.Uint("metricport", 8080, "port for /metrics endpoint")
+	pushgateway := flag.String("pushgateway", "127.0.0.1:9091", "host:port for pushgateway")
+	profilerEnabled := flag.Bool("profiler-enabled", false, "whether to enable the auto-profiler")
+	accountMultiplierFlag := flag.Int("account-multiplier", 50, "number of accounts to create per load tps")
+	feedbackEnabled := flag.Bool("feedback-enabled", true, "wait for trannsaction execution before submitting new transaction")
```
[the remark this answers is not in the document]
AndrewM-SDET: Hardcoded this and several other values that will always be consistent in CI testing.

.github/workflows/performance-ci.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.COMMIT_SHA }}
+      - name: Get commit date
+        id: commit_date
+        run: echo "::set-output name=date::$(git show --no-patch --no-notes --pretty='%cI' $COMMIT_SHA)"
+      - name: Build relic and other tools
+          run: make install-tools
+      - name: Run Localnet TPS test
+          run: bash -c "cd integration/localnet/ && make tpsTestCI
```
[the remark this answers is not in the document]
AndrewM-SDET: fixed

integration/localnet/Makefile
AndrewM-SDET: fixed

integration/localnet/Makefile
AndrewM-SDET: fixed

integration/benchmark/main.go
AndrewM-SDET: This and several other flags have been moved to hardcoded variables as they will never change in CI testing.

integration/benchmark/main.go
AndrewM-SDET: gitSha, goVersion and osVersion are now pulled directly from version and runtime.

integration/benchmark/main.go
```diff
@@ -0,0 +1,353 @@
...
+	serviceAccountPrivateKeyHex := unittest.ServiceAccountPrivateKeyHex
+
+	// Version and Commit Info
+	gitSha := build.Commit()
+	goVersion := runtime.Version()
+	osVersion := runtime.GOOS + runtime.GOARCH
+
+	chainID := flowsdk.Emulator
+
+	ctx, cancel := context.WithCancel(context.Background())
```
pattyshack, line 86: my comment in the other file still applies.  this context does nothing since it's only cancel at the very end.  just remove it and just run the goroutine (line 173) body as part of main

AndrewM-SDET: Completed in both files.

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+	// CI relevant flags
+	tpsFlag := flag.String("tps", "300", "transactions per second (TPS) to send, accepts a comma separated list of values if used in conjunction with `tps-durations`")
+	tpsDurationsFlag := flag.String("tps-durations", "10m", "duration that each load test will run, accepts a comma separted list that will be applied to multiple values of the `tps` flag (defaults to infinite if not provided, meaning only the first tps case will be tested; additional values will be ignored)")
+	ciFlag := flag.Bool("ci-run", false, "whether or not the run is part of CI")
+	leadTime := flag.String("leadTime", "30s", "the amount of time before data slices are started")
+	sliceSize := flag.String("sliceSize", "2m", "the amount of time that each slice covers")
+	flag.Parse()
+
+	// Hardcoded CI values
+	loadType := "token-transfer"
```
pattyshack, line 70: define these as const, outside the main function

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		log.Fatal().Err(err).Msgf("unable to init loader")
+	}
+	lg.Start()
+
+	testStartTime := time.Now()
+	time.Sleep(loadCase.duration)
+	testEndTime := time.Now()
+
+	if lg != nil {
```
pattyshack, line 206: lg is never nil

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+			f.Close()
+		} else {
+			fmt.Println(err)
+		}
+	}
+
+	// output human-readable json blob
+	timestamp := time.Now()
+	fileName := fmt.Sprintf("tps-results-%v.json", timestamp)
+	f, err := os.Create(fileName)
```
pattyshack, line 284: just use os.WriteFile

pattyshack (commented):
a few more nitpicks.  i'll rubberstamp once alexey is happy with this

integration/loader/main.go
```diff
@@ -54,9 +53,6 @@ func main() {
 
 	chainID := flowsdk.ChainID([]byte(*chainIDStr))
 
-	ctx, cancel := context.WithCancel(context.Background())
```
SaveTheRbtz, line 57: you should likely leave the top level context -- we'll need it for multiple things. (for example it is still useful for the proper pusher shutdown.)

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	// if we are uploading to BigQuery then we need a copy of the json file
+	// that is newline delimited json
+	if ci {
+		jsonString := string(jsonText)
+		// remove the surrounding square brackets
+		jsonString = jsonString[1 : len(jsonString)-1]
+		// end of object commas will be replaced with newlines
+		jsonString = strings.ReplaceAll(jsonString, "},", "}\n")
```
SaveTheRbtz, line 269: you should likely just do:
```
for _, slice := range slices {
  jsonText, _ := json.Marshal(slice)
  ...
}
```

This will avoid the text mangling.

integration/loader/main.go
pattyshack: i disagree.  this does nothing right now.  we can always add it back later if pusher does eventually support proper shutdown

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+	loadedAccessAddr := accessNodeAddrs[0]
+	flowClient, err := client.NewClient(loadedAccessAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatal().Err(err).Msgf("unable to initialize Flow client")
+	}
+
+	supervisorAccessAddr := accessNodeAddrs[0]
+	if len(accessNodeAddrs) > 1 {
+		supervisorAccessAddr = accessNodeAddrs[1]
+	}
```
SaveTheRbtz, line 159: always false.

integration/loader/main.go
```diff
@@ -140,62 +131,58 @@ func main() {
...
+		}
 
-			time.Sleep(c.duration)
-
-			if lg != nil {
-				lg.Stop()
-			}
+		// if the duration is 0, we run this case forever
+		if c.duration.Nanoseconds() == 0 {
+			return
```
pattyshack, line 177: This should be an infinite sleep loop?

integration/benchmark/main.go
```diff
@@ -0,0 +1,338 @@
...
+
+		for {
+			err := pusher.Push()
+			if err != nil {
+				log.Warn().Err(err).Msg("failed to push metrics to pushgateway")
+			}
+		}
+	}()
+
+	accessNodeAddrs := strings.Split(access, ",")
```
SaveTheRbtz, line 110: always single

integration/benchmark/main.go
AndrewM-SDET: This would still give us square brackets that would need to be removed on each object and we would have to remarshal for the non-newline JSON file anyway. It ends up replacing just line 269 with a loop. Are we sure we want this change?

integration/loader/main.go
AndrewM-SDET: Good catch, this behavior changed without the context/subroutine.

integration/loader/main.go
AndrewM-SDET: I am fine with this either way, which would you both prefer? As the code is now without the context is slightly cleaner and we do have the history to reference when we want to readd it.

SaveTheRbtz (approved):
Please fix the "Performance CI" workflow, make sure it runs and uploads results successfully.
Then move `main.go` into `cmd/benchmark/main.go`

In the followup diffs:
- [ ] move prometheus functions into a separate file.
- [ ] add unit tests for the processing/formatting/parsing functions. 
- [ ] add an end-to-end test for the localnet/loader since currently we need to run a manual test after every change.

pattyshack (approved):

AndrewM-SDET:
bors merge

AndrewM-SDET:
bors merge

AndrewM-SDET:
bors merge

PR #2962, opened 2022-08-11
dan/2952 handle RPCs using the protocol.API

danielholmes839:
Based on https://github.com/onflow/flow-go/pull/2701/files. The observer is already receiving blocks and headers. It already was storing and responding to RPCs using the access backend. However, we want to stop using the access node backend and replace it with the `protocol.API`


engine/access/apiproxy/access_api_proxy.go
```diff
@@ -201,6 +199,18 @@ func (h *FlowAccessAPIRouter) GetTransactionResult(context context.Context, req
 	return res, err
 }
 
+func (h *FlowAccessAPIRouter) GetTransactionResultsByBlockID(context context.Context, req *access.GetTransactionsByBlockIDRequest) (*access.TransactionResultsResponse, error) {
```
danielholmes839, line 202: These were missing without anyone noticing because of type composition the router still implemented the method.

engine/access/apiproxy/access_api_proxy.go
```diff
@@ -256,14 +266,14 @@ func (h *FlowAccessAPIRouter) GetEventsForBlockIDs(context context.Context, req
 }
 
 func (h *FlowAccessAPIRouter) GetNetworkParameters(context context.Context, req *access.GetNetworkParametersRequest) (*access.GetNetworkParametersResponse, error) {
-	res, err := h.AccessAPIServer.GetNetworkParameters(context, req)
-	h.log("observer", "GetNetworkParameters", err)
+	res, err := h.Upstream.GetNetworkParameters(context, req)
+	h.log("upstream", "GetNetworkParameters", err)
```
danielholmes839, line 270: Certain methods were switched from local to upstream since the `protocol.API` interface does not implement them.

engine/access/apiproxy/access_api_proxy.go
```diff
@@ -136,50 +135,49 @@ func (h *FlowAccessAPIForwarder) faultTolerantClient() (access.AccessAPIClient,
 // Ping pings the service. It is special in the sense that it responds successful,
 // only if all underlying services are ready.
 func (h *FlowAccessAPIRouter) Ping(context context.Context, req *access.PingRequest) (*access.PingResponse, error) {
-	res, err := h.AccessAPIServer.Ping(context, req)
-	h.log("observer", "Ping", err)
-	return res, err
+	h.log("observer", "Ping", nil)
+	return &access.PingResponse{}, nil
 }
```
danielholmes839, line 140: `Ping` is now implemented here because it doesn't make sense to forward upstream and the `protocol.API` interface does not have a `Ping` method. I think adding `Ping` to the  `protocol.API` also makes sense

cmd/observer/node_builder/observer_builder.go
```diff
@@ -156,7 +157,7 @@ type ObserverServiceBuilder struct {
 
 	// components
 	LibP2PNode              *p2p.Node
-	FollowerState           protocol.MutableState
+	FollowerState           stateprotocol.MutableState
```
koko1123, line 160: you didn't have to change the naming here?

engine/access/apiproxy/access_api_proxy.go
koko1123: we don't need to have `Ping` in protocol either

koko1123 (commented):
Looks good to me! some nits but no major changes needed, will leave it up to @peterargue

cmd/observer/node_builder/observer_builder.go
danielholmes839: I did since I imported the package with `protocol.API` https://github.com/onflow/flow-go/pull/2962/files/5b08f9c1f0f51eafed149a86e572578868d29b49#diff-8e2393e4d4435f2229d72e4d978fcc6b45e44cc54e978011f3375709e0a143fcR37

engine/protocol/handler.go
```diff
@@ -0,0 +1,145 @@
...
+	signerIndicesDecoder hotstuff.BlockSignerDecoder
+}
+
+// HandlerOption is used to hand over optional constructor parameters
+type HandlerOption func(*Handler)
+
+func NewHandler(api API) *Handler {
+	h := &Handler{
+		api:                  api,
+		signerIndicesDecoder: &signature.NoopBlockSignerDecoder{},
```
danielholmes839, line 26: I don't fully understand how this is used in the access handler.

engine/access/apiproxy/access_api_proxy.go
danielholmes839: https://github.com/onflow/flow-go/pull/2962/files#diff-3505449f31926a810a7226adb4a630986055d7196c265e57ac8633cc703ccb53R16-R51

koko1123 (approved):
Thanks for this! looks good to me

cmd/observer/node_builder/observer_builder.go
```diff
@@ -991,10 +992,15 @@ func (builder *ObserverServiceBuilder) enqueueRPCServer() {
...
-			Upstream:        forwarder,
-			AccessAPIServer: engineBuilder.Handler(),
+			Logger:   builder.Logger,
+			Metrics:  metrics.NewObserverCollector(),
+			Upstream: forwarder,
+			Observer: protocol.NewHandler(protocol.New(
+				node.State,
+				node.Storage.Blocks,
+				node.Storage.Headers,
+				backend.NewNetworkAPI(node.State, node.RootChainID, backend.DefaultSnapshotHistoryLimit),
```
smnzhu, line 1002: Why did we replace AccessAPIServer with o\Observer?

engine/access/rpc/backend/backend_network.go
```diff
@@ -0,0 +1,121 @@
...
+var SnapshotHistoryLimitErr = fmt.Errorf("reached the snapshot history limit")
+
+type backendNetwork struct {
+	state                protocol.State
+	chainID              flow.ChainID
+	snapshotHistoryLimit int
+}
+
+// used by the observer
+func NewNetworkAPI(state protocol.State, chainID flow.ChainID, snapshotHistoryLimit int) *backendNetwork {
```
smnzhu, line 23: Can you explain / add comments what is the purpose of this component and why we split it out?

cmd/observer/node_builder/observer_builder.go
danielholmes839: The AccessAPIServer was what the Observer used to handle requests without using the upstream access node. But the Observer should not support the full access api.

engine/access/apiproxy/access_api_proxy.go
```diff
@@ -28,9 +29,7 @@ type FlowAccessAPIRouter struct {
 	Logger   zerolog.Logger
 	Metrics  *metrics.ObserverCollector
 	Upstream *FlowAccessAPIForwarder
-
-	// local observer
-	access.AccessAPIServer
+	Observer *protocol.Handler
```
smnzhu: I mean here, why did we change it to Observer?

smnzhu: If this file is only used by observers, then can we move the file somewhere else, because the current location is confusing.

On the other hand if this file is used by both access node and observer, then we should rename this field to something more generic

engine/access/rpc/backend/backend_network.go
danielholmes839: Added the comment. The observer needs to handle GetNetworkParameters, and GetLatestProtocolSnapshot the same way the access node does without having to setup the entire access backend.

engine/access/apiproxy/access_api_proxy.go
danielholmes839: The access_api_proxy.go file name could be changed. I don't think it should be part of the scope for this PR though. I changed `access.AccessAPIServer` to Observer since the Observer should not support all RPCs.

smnzhu: But is this file used by the observer only? or is it also used by Access node?

danielholmes839: Only the observer uses it

smnzhu (approved):

PR #2963, opened 2022-08-11
Standardize nested transaction creation/restart/commitment

pattyshack:
1. Refactor program handler's state stack into StateHolder.
2. Switch transaction invoker and seq num verifier to the new api

In a future PR, I'll restrict access to Environment methods (if cadence accesses
anything besides GetCode, we need to know about it, and invalidate the programs
cache accordingly).

### DIFF
--- a/fvm/handler/programs.go
+++ b/fvm/handler/programs.go
@@ -10,11 +10,6 @@ import (
 	"github.com/onflow/flow-go/fvm/state"
 )
 
-type stackEntry struct {
-	state    *state.State
-	location common.Location
-}
-
 // ProgramsHandler manages operations using Programs storage.
 // It's separation of concern for hostEnv
 // Cadence contract guarantees that Get/Set methods will be called in a LIFO manner,
@@ -22,10 +17,8 @@ type stackEntry struct {
 // naturally, making cleanup method essentially no-op. But if something goes wrong, all nested
 // views must be merged in order to make sure they are recorded
 type ProgramsHandler struct {
-	masterState  *state.StateHolder
-	viewsStack   []stackEntry
-	Programs     *programs.Programs
-	initialState *state.State
+	masterState *state.StateHolder
+	Programs    *programs.Programs
 
 	// NOTE: non-address programs are not reusable across transactions, hence
 	// they are kept out of the shared program cache.
@@ -36,9 +29,7 @@ type ProgramsHandler struct {
 func NewProgramsHandler(programs *programs.Programs, stateHolder *state.StateHolder) *ProgramsHandler {
 	return &ProgramsHandler{
 		masterState:        stateHolder,
-		viewsStack:         nil,
 		Programs:           programs,
-		initialState:       stateHolder.State(),
 		nonAddressPrograms: map[common.LocationID]*interpreter.Program{},
 	}
 }
@@ -58,34 +49,13 @@ func (h *ProgramsHandler) Set(location common.Location, program *interpreter.Pro
 		return nil
 	}
 
-	if len(h.viewsStack) == 0 {
-		return fmt.Errorf("views stack empty while set called, for location %s", location.String())
+	state, err := h.masterState.CommitParseRestricted(address)
+	if err != nil {
+		return err
 	}
 
-	// pop
-	last := h.viewsStack[len(h.viewsStack)-1]
-	h.viewsStack = h.viewsStack[0 : len(h.viewsStack)-1]
-
-	if last.location.ID() != location.ID() {
-		return fmt.Errorf("set called for type %s while last get was for %s", location.String(), last.location.String())
-	}
-
-	h.Programs.Set(address, program, last.state)
-
-	err := h.mergeState(last.state, h.masterState.EnforceLimits())
-
-	return err
-}
-
-func (h *ProgramsHandler) mergeState(state *state.State, enforceLimits bool) error {
-	if len(h.viewsStack) == 0 {
-		// if this was last item, merge to the master state
-		h.masterState.SetActiveState(h.initialState)
-	} else {
-		h.masterState.SetActiveState(h.viewsStack[len(h.viewsStack)-1].state)
-	}
-
-	return h.masterState.State().MergeState(state, enforceLimits)
+	h.Programs.Set(address, program, state)
+	return nil
 }
 
 func (h *ProgramsHandler) Get(location common.Location) (*interpreter.Program, bool) {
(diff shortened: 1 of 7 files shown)


fvm/state/state_holder.go
```diff
@@ -16,29 +30,226 @@ import (
...
+		return fmt.Errorf("nested transaction not found")
+	}
+
+	for s.currentState() != id.state {
+		err := s.mergeIntoParent(s.EnforceInteractionLimits())
+		if err != nil {
+			return fmt.Errorf("cannot restart nested transaction: %w", err)
+		}
+	}
+
```
janezpodhostnik, line 250: I don't think it's necessary to do a full search just to get a better error message of it not existing instead of it not being at the top. Just checking that its not the top one would be ok I think.

fvm/state/state_holder.go
```diff
@@ -16,29 +30,226 @@ import (
...
+	// NOTE: limit enforcement is disabled here because cadence environment's
+	// Get() does not support returning error.
+	return s.mergeIntoParent(false)
+}
+
+// RestartNestedTransaction merges all changes that belongs to the nested
+// transaction about to be restart (for spock/meter bookkeeping), then
+// wipes its view changes.
+func (s *StateHolder) RestartNestedTransaction(
```
janezpodhostnik, line 229: Is this for optimisation purposes? If not I think it would be cleaner/more explicit to not have this at all. so you have to do commit and new.

fvm/transactionSequenceNum.go
```diff
@@ -72,8 +75,11 @@ func (c *TransactionSequenceNumberChecker) checkAndIncrementSequenceNumber(
 
 	_, err = accounts.SetPublicKey(proposalKey.Address, proposalKey.KeyIndex, accountKey)
 	if err != nil {
-		childState.View().DropDelta()
+		sth.DisableAllLimitEnforcements()
+		_ = sth.RestartNestedTransaction(nestedTxnId)
```
janezpodhostnik, line 79: Why the extra DisableAllLimitEnforcements here?

Also... A restart is not needed here. Since this is the end of the transaction. Maybe just call commit.

fvm/handler/programs.go
```diff
@@ -102,58 +72,18 @@ func (h *ProgramsHandler) Get(location common.Location) (*interpreter.Program, b
...
-		err := h.viewsStack[i-1].state.MergeState(entry.state, h.masterState.EnforceInteractionLimits())
+		err := h.masterState.AttachAndCommitParseRestrictedToParent(state)
 		if err != nil {
-			return fmt.Errorf("cannot merge state while cleanup: %w", err)
+			panic(fmt.Sprintf("merge error while getting program, panic: %s", err))
 		}
+		return program, true
 	}
 
-	err := h.initialState.MergeState(h.viewsStack[0].state, h.masterState.EnforceInteractionLimits())
```
janezpodhostnik, line 150: I guess we didn't need to do this during cleanup...?

janezpodhostnik (changes requested):
1. ~It feels like the concept of `IsParseRestricted` nested transactions could be a separate PR, as it is an upgrade to the nested transactions.~ ok, nvm I get it. Another time I'm not sure of the naming. We wont be able to restrict to only parsing, I imagine this is more of a write/read restriction; also maybe like this? `BeginNestedTransaction(WithReadRestriction(address))`?

2. Calling this concept a nested transaction,when this transaction is not like a cadence transaction, confused me a bit. For a moment I thought we are nesting cadence transactions into the transaction. (Which is something that technically happens with meta invocations).  

Not sure what we can do to make this clearer... maybe name it a stack? StateStack? ... or abbreviate it to `NestTran` when used as part of a function name... 

3. `CommitToMain and CommitToParent` could just be a single commit function: Commit that also takes the `NestedTransactionId` of the child that is to be commited one level up, and verifies that the child Id is actually last on the stack.

fvm/handler/programs.go
pattyshack: This got folded/standardized into RestartNestedTransaction

pattyshack:
`BeginNestedTransaction(WithReadRestriction(address))` I'm anti optional arguments in general.  It's easy to accidentally misconfigure the call.  I prefer something more explicit.

I'm also anti-abbreviation in method names (and to a lesser degree, variable names; one or two letter variable names are horrible; the bigger the variable scope, the less abbreviated the variable name should be).  Abbreviation don't usually save you much in terms of typing, but impedes readability, especially if the abbreviation is company specific.

(In general, program should be written in such a way that it's easily followable/readable for a person not familiar with the code base.)

wrt 1.

By ParseRestricted, I meant "restricted to cadence runtime methods involved in the process of creating the parsed cadence program".  So far, I'm aware of the following operations: GetProgram, SetProgram, GetCode, and ResolveLocation.  (If cadence does any storage read besides these methods, then our caching assumption is wrong, and we need to update cache invalidation accordingly)

wrt 2.

The unfortunate part is that crypto space reinvented a lot of standard cs/database terminology.  The classic example is smart contract is really just a specific from of stored procedures.  I probably would have name cadence/fvm transaction as cadence/fvm interpretation instead.

In the long run, we'll probably end up with 3 layers of "transactions"
1. cadence transaction/intepretation, which only handles interpreting cadence programs
2. fvm state transaction/interpretation, which handles metering, storage check, spock, etc.
3. (the real) storage transaction, which handles storage concurrency control and actual commit (Right now this is only a "view" and don't not have proper concurrency access control)

wrt 3.

That's a good idea.  I'll make the recommended change

fvm/state/state_holder.go
pattyshack: This is more about not accidentally merging nested transaction than about error messaging.

Suppose the calling provide an invalid nest transaction id.  Not checking the id first would cause us to accidently merge everything into the main transaction.

pattyshack: I'll add a comment to clarify the intention here

fvm/state/state_holder.go
pattyshack: This is not an optimization.  This replaces/standardizes the old program handler's Cleanup method.

Restarting is needed to preserve the spock/meter bookkeeping.  (If we didn't have to worry about spock/meter, we can simply abort the nested transaction by dropping the stack frames, and create a new nested transaction)

fvm/transactionSequenceNum.go
pattyshack: If we don't disable the limit, restart may fail since restart merges aborted nested transactions internally.

I'll add a comment

pattyshack:
i've added some comments for clarification, and combined CommitToParent & CommitToMain into Commit

janezpodhostnik (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors merge

PR #2966, opened 2022-08-12
Fix transaction invoker to not early exit once transaction started

pattyshack:
1. proc's fields were not correctly populated due to early exiting
2. also fixed meter tests (the tests were testing the wrong things)

(These issue was discovered when I tried to disable limiting while setting
meter weights/limits).

### DIFF
--- a/fvm/transactionInvoker.go
+++ b/fvm/transactionInvoker.go
@@ -79,6 +79,7 @@ func (i *TransactionInvoker) Process(
 	}()
 
 	env := NewTransactionEnvironment(*ctx, vm, sth, programs, proc.Transaction, proc.TxIndex, span)
+
 	predeclaredValues := valueDeclarations(env)
 
 	location := common.TransactionLocation(proc.ID)
@@ -159,6 +160,7 @@ func (i *TransactionInvoker) Process(
 			Uint64("blockHeight", blockHeight).
 			Msg("transaction executed with error")
 
+		// TODO(patrick): make env reusable on error
 		// reset env
 		env = NewTransactionEnvironment(*ctx, vm, sth, programs, proc.Transaction, proc.TxIndex, span)
 
@@ -180,7 +182,7 @@ func (i *TransactionInvoker) Process(
 				Uint64("blockHeight", blockHeight).
 				Msg("transaction fee deduction executed with error")
 
-			return feesError
+			txError = feesError
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


fvm/transactionInvoker.go
```diff
@@ -78,10 +78,17 @@ func (i *TransactionInvoker) Process(
...
 	}()
 
-	env, err := NewTransactionEnvironment(*ctx, vm, sth, programs, proc.Transaction, proc.TxIndex, span)
+	env, err := NewTransactionEnvironment(*ctx, vm, sth, programs, proc.Transaction, proc.TxIndex, span, true)
 	if err != nil {
 		return fmt.Errorf("error creating new environment: %w", err)
 	}
+
+	// TODO(patrick): make env reusable on error
+	errEnv, err := NewTransactionEnvironment(*ctx, vm, sth, programs, proc.Transaction, proc.TxIndex, span, false)
```
janezpodhostnik, line 87: Does it make sense creating this even though we most of the time (assuming successful transactions are prevalent) don't need it?

Not loading config from state should be ok as it does not matter how much we meter during transaction fee deduction. We cannot charge for that part. It does might if we hit some sort of limit, but the deduct fees method on the service account should be trusted not to do any infinite loops and the limiting is disabled right now anyway.

I think this (not loading config from state) is ok, but it does deserve some comments in the code what and why this is the case.

janezpodhostnik (changes requested):
One comment. Looks good otherwise.

fvm/transactionInvoker.go
pattyshack: (note that this may not matter after #2971)

I'll add a comment to the file.

There's no need to re-initialize the meter weights/limits since the transaction state (sth) is shared.  Creating new env is cheap once we stop re-reading the environment's meter weights/limits.

pattyshack: since #2971 is already approved, i'll just merge that first and simplify this (and skip adding a new comment)

fvm/transactionInvoker.go
```diff
@@ -180,7 +182,7 @@ func (i *TransactionInvoker) Process(
 				Uint64("blockHeight", blockHeight).
 				Msg("transaction fee deduction executed with error")
 
-			return feesError
```
SaveTheRbtz, line 183: is there a reason to call programs.Cleanup with an empty set there?

janezpodhostnik (approved):

fvm/transactionInvoker.go
pattyshack: I've ready cleaned this up as part of #2963

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors retry

PR #2969, opened 2022-08-12
Handle unsealed invalid service events

jordanschalm:
This PR improves handling of unsealed invalid service events. Currently, observation of an invalid service event triggers epoch fallback upon insertion of the block where the service event takes effect, even if that block is not yet finalized. With this change, processing an invalid service event instead marks an un-finalized fork's `EpochStatus` as "tentatively triggering epoch fallback". Only when the fork is finalized, is epoch fallback globally triggered.

### Changes
* Adds `InvalidServiceEventIncorporated` flag to `EpochStatus`, which is set when an invalid service event is processed on this fork
* Modifies epoch fallback triggering logic
  * Removes all cases where block insertion can trigger epoch fallback
  * When finalizing a block, additionally checks whether the `InvalidServiceEventIncorporated` flag is set, and if so triggers epoch fallback
  * Modifies behaviour of out-of-order service events to be consistent. Previously processing an `EpochCommit` before an `EpochSetup` would return an error, rather than triggering epoch fallback, as desired. Now all cases trigger epoch fallback.
* Refactors invalid service event tests
  * All blocks are now finalized, so all tests use a separate state instance


state/protocol/badger/mutator.go
```diff
@@ -894,25 +878,21 @@ func (m *FollowerState) epochStatus(block *flow.Header) (*flow.EpochStatus, bool
 //    this method returns a slice of Badger operations to apply while storing the block.
 //    This includes an operation to index the epoch status for every block, and
 //    operations to insert service events for blocks that include them.
-//  * insertingCandidateTriggersEpochFallback - if inserting this block causes epoch
-//    fallback mode to be triggered, this is true. This causes flag to be set in the
-//    database and a protocol event to be emitted. This can only be true for at most
-//    one block for a given database instance.
 //
 // No errors are expected during normal operation.
```
durkmurder, line 882: I think comment section for this function needs to be revisited after implementing QC front-loading, lets place a TODO for it.

state/protocol/badger/mutator_test.go
```diff
@@ -1365,7 +1505,7 @@ func TestEmergencyEpochFallback(t *testing.T) {
 			err = state.Finalize(context.Background(), block1.ID())
 			require.NoError(t, err)
 
-			assertEpochEmergencyFallbackTriggered(t, state)
+			assertEpochEmergencyFallbackTriggered(t, state, true)
```
durkmurder, line 1508: I really liked how you have checked that it wasn't triggered before and after some action(finalizing block) it triggers one, can we also add an assert that it wasn't triggered before?

durkmurder (approved):
Very nice and clean, looks good to me. Great job.

state/protocol/badger/mutator_test.go
jordanschalm: https://github.com/onflow/flow-go/pull/2969/commits/7909a51d8e111a58d62f60346d8b6efaedc0cc2f 👍

state/protocol/badger/mutator.go
jordanschalm: https://github.com/onflow/flow-go/pull/2969/commits/317b9f2605660a45fd3a0e824bea1685ee1e029a

AlexHentschel (approved):
Looks great. Thanks for the solid tests and improvements on documentation. 

I think this PR can be merged as is. Most of my comments are stylistic, including the following tiny request: 
* could we add some small documentation to this function: https://github.com/onflow/flow-go/blob/59f857161b66f3fbf1fbcf800e6d071f565d8bc2/state/protocol/badger/state.go#L688

A more conceptual consideration is the [this comment](https://github.com/onflow/flow-go/pull/2969#discussion_r1003733713). I am totally happy if we don't address it as part of this PR, but wanted to flag it nevertheless.

utils/unittest/protocol_state.go
```diff
@@ -89,5 +90,13 @@ func SealBlock(t *testing.T, st protocol.MutableState, block *flow.Block, receip
 	err = st.Extend(context.Background(), block3)
 	require.NoError(t, err)
 
-	return block3.Header
+	return block2.Header, block3.Header
+}
+
+// InsertAndFinalize inserts, then finalizes, then input block.
```
jordanschalm, line 96:
```suggestion
// InsertAndFinalize inserts, then finalizes, the input block.
```

state/protocol/badger/mutator.go
```diff
@@ -938,12 +918,12 @@ func (m *FollowerState) handleEpochServiceEvents(candidate *flow.Block) (
 	// chain finalization should halt.
 	parent, err := m.blocks.ByID(candidate.Header.ParentID)
 	if err != nil {
-		return nil, false, fmt.Errorf("could not get parent (id=%x): %w", candidate.Header.ParentID, err)
+		return nil, fmt.Errorf("could not get parent (id=%x): %w", candidate.Header.ParentID, err)
 	}
 	for _, seal := range parent.Payload.Seals {
```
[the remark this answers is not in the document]
jordanschalm: Will address this as a follow-up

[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3461

PR #2974, opened 2022-08-15
Set last full block height in ingestion engine component startup

kc1116:
This pull requests updates the LastFullBlockHeight in the ingestion engine component startup so that we only request missing collections in blocks above that height. In the processFinalizedBlock func we skip requesting collections for blocks under the lastFullBlockHeight.


engine/access/ingestion/engine.go
```diff
@@ -187,6 +187,34 @@ func New(
...
+		parent.Throw(fmt.Errorf("could not check if root snapshot is a spork root snapshot: %w", err))
+	}
+
+	// This is useful for dynamically bootstrapped access node, they will request missing collections. In order to ensure all txs
+	// from the missing collections can be verified, we must ensure they are referencing to known blocks.
+	// That's why we set the full block height to be rootHeight + TransactionExpiry, so that we only request missing collections
+	// in blocks above that height.
+	if !isSporkRootSnapshot {
+		firstFullHeight := rootBlock.Height + flow.DefaultTransactionExpiry
+		err := e.blocks.UpdateLastFullBlockHeight(firstFullHeight)
```
zhangchiqing, line 209: We should only ever set it once. 

Would it overwrite the value on every restart?

engine/access/ingestion/engine_test.go
```diff
@@ -112,12 +112,22 @@ func (suite *Suite) SetupTest() {
...
 	require.NoError(suite.T(), err)
 
-	// stops requestMissingCollections from executing in processBackground worker
-	suite.blocks.On("GetLastFullBlockHeight").Once().Return(uint64(0), errors.New("do nothing"))
+	getLastFullBlockHeightCalls := 0
+	returnErr := func() error {
+		// stops requestMissingCollections from executing in processBackground worker during component startup
+		if getLastFullBlockHeightCalls < 2 {
+			return fmt.Errorf("do nothing")
+		}
```
jordanschalm, line 120: why do we need this now?

engine/access/ingestion/engine.go
jordanschalm: +1 It will overwrite on every restart, we need insert-if-not-exists here.

engine/access/ingestion/engine_test.go
kc1116: So that we return an error for the first call to [getLastFullBlockHeight](https://github.com/onflow/flow-go/blob/443e8038838bf488a951d249bf05fb4be4708144/engine/access/ingestion/engine.go#L591) which allows us to [skip requesting collections in the background](https://github.com/onflow/flow-go/blob/443e8038838bf488a951d249bf05fb4be4708144/engine/access/ingestion/engine.go#L218-L239). For all calls after the first call we want to instead return nil now that we are checking the last full block height [here](https://github.com/onflow/flow-go/blob/443e8038838bf488a951d249bf05fb4be4708144/engine/access/ingestion/engine.go#L411).

engine/access/ingestion/engine.go
kc1116: https://github.com/onflow/flow-go/pull/2974/commits/f5b1e1da1be9792dabe98246a86739761d1dbfb6

jordanschalm (approved):

engine/access/ingestion/engine.go
```diff
@@ -187,6 +187,34 @@ func New(
...
+		parent.Throw(fmt.Errorf("could not check if root snapshot is a spork root snapshot: %w", err))
+	}
+
+	// This is useful for dynamically bootstrapped access node, they will request missing collections. In order to ensure all txs
+	// from the missing collections can be verified, we must ensure they are referencing to known blocks.
+	// That's why we set the full block height to be rootHeight + TransactionExpiry, so that we only request missing collections
+	// in blocks above that height.
+	if !isSporkRootSnapshot {
+		firstFullHeight := rootBlock.Height + flow.DefaultTransactionExpiry
+		err := e.blocks.InsertLastFullBlockHeight(firstFullHeight)
```
zhangchiqing, line 209: This call will be called multiple times right?

Usually insert operation will fail when a key already exists, which will cause it to crash.

How does it work here? Does it return nil for the second operation? We'd better comment about the behavior or describe the behavior in function name, such as `InsertLastFullBlockHeightIfNotExist`

If we are renaming to `InsertLastFullBlockHeightIfNotExist`, then we should handle the `storage.ErrAlreadyExists` error with `SkipDuplicates`.

storage/badger/operation/heights.go
```diff
@@ -38,6 +38,12 @@ func RetrieveSealedHeight(height *uint64) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeSealedHeight), height)
 }
 
+// SetLastCompleteBlockHeight inserts the last full block height if it is not already set.
+// Calling this function multiple times is a no-op and returns no expected errors.
+func SetLastCompleteBlockHeight(height uint64) func(*badger.Txn) error {
+	return SkipDuplicates(update(makePrefix(codeLastCompleteBlockHeight), height))
```
zhangchiqing, line 44: I think update will fail for the first operation, [because the key does exist yet.](https://github.com/onflow/flow-go/blob/master/storage/badger/operation/common.go#L125)

storage/badger/blocks.go
```diff
@@ -108,6 +108,17 @@ func (b *Blocks) IndexBlockForCollections(blockID flow.Identifier, collIDs []flo
 	return nil
 }
 
+// InsertLastFullBlockHeight inserts the last full block height
+func (b *Blocks) InsertLastFullBlockHeight(height uint64) error {
+	return operation.RetryOnConflict(b.db.Update, func(tx *badger.Txn) error {
+		err := operation.SetLastCompleteBlockHeight(height)(tx)
```
zhangchiqing, line 114: Should we use `InsertLastCompleteBlockHeight` instead?

storage/badger/operation/heights.go
kc1116: good catch I updated it to insert instead https://github.com/onflow/flow-go/pull/2974/commits/9e396d8e6097b0f8ecf7cec97181aac118f6a530

engine/access/ingestion/engine.go
kc1116: This will be called multiple times if the engine crashes, but will only insert the last full block height once if it doesn't exist. [SkipDuplicates](https://github.com/onflow/flow-go/blob/khalil/1761-an-finalized-block-request/storage/badger/operation/heights.go#L44) is already handling the `storage.ErrAlreadyExists ` for us. Functionality is already documented in the [interface](https://github.com/onflow/flow-go/blob/khalil/1761-an-finalized-block-request/storage/blocks.go#L36)  . I renamed the func for more readability https://github.com/onflow/flow-go/pull/2974/commits/0e04259bcc147991a8f9cc7db6f10a98496acf25.

storage/badger/blocks.go
kc1116: No, I renamed SetLastCompletedBlockHeight -> InsertLastCompletedBlockHeightIfNotExists https://github.com/onflow/flow-go/pull/2974/commits/5afa8ef0dc376fd24ffc268351fc7b5b28676242

storage/badger/operation/heights.go
```diff
@@ -38,6 +38,12 @@ func RetrieveSealedHeight(height *uint64) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeSealedHeight), height)
 }
 
+// SetLastCompleteBlockHeight inserts the last full block height if it is not already set.
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/2974/commits/6b1098e113178c0eb8005240931d943c3b644657

storage/badger/operation/heights.go
```diff
@@ -38,6 +38,12 @@ func RetrieveSealedHeight(height *uint64) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeSealedHeight), height)
 }
 
+// InsertLastCompleteBlockHeightIfNotExists inserts the last full block height if it is not already set.
+// Calling this function multiple times is a no-op and returns no expected errors.
+func InsertLastCompleteBlockHeightIfNotExists(height uint64) func(*badger.Txn) error {
+	return SkipDuplicates(insert(makePrefix(codeLastCompleteBlockHeight), height))
```
zhangchiqing, line 44:
```suggestion
	return SkipDuplicates(InsertLastCompleteBlockHeight(height))
```

zhangchiqing (approved):
Looks good. Thanks for the fix

storage/badger/operation/heights.go
kc1116: https://github.com/onflow/flow-go/pull/2974/commits/8a7a387e36cf7150ca7d9cc3bb8772e96d198838

kc1116:
bors merge

kc1116:
bors merge

kc1116:
bors try

kc1116:
bors try

kc1116:
bors try

kc1116:
bors try

kc1116:
bors try

kc1116:
bors try

PR #2979, opened 2022-08-15
Dedup more environment code

pattyshack:
NOTE:
 1. I've move meterMemory's logic into MeterMemory.
 2. GetStorageCapacity: we're using the script env version, which has additional
    comments.
 3. GetAccountAvailableBalance: we're using the script env verion, which returns
    0 instead of "value" (which is also zero) on error.
 4. GetAccountKey: we're using the transaction env version.  accountKeys is
    never nil.

### DIFF
--- a/fvm/env.go
+++ b/fvm/env.go
@@ -8,6 +8,7 @@ import (
 	"github.com/onflow/cadence"
 	jsoncdc "github.com/onflow/cadence/encoding/json"
 	"github.com/onflow/cadence/runtime"
+	"github.com/onflow/cadence/runtime/ast"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/cadence/runtime/interpreter"
 	"go.opentelemetry.io/otel/attribute"
@@ -42,8 +43,6 @@ type Environment interface {
 // TODO(patrick): rename the method names to match meter.Meter interface.
 type MeterInterface interface {
 	Meter(common.ComputationKind, uint) error
-
-	meterMemory(kind common.MemoryKind, intensity uint) error
 }
 
 // TODO(patrick): refactor this into an object
@@ -108,9 +107,11 @@ func (env *commonEnv) ComputationUsed() uint64 {
 	return uint64(env.sth.TotalComputationUsed())
 }
 
-// TODO(patrick): rm once Meter object has been refactored
 func (env *commonEnv) MeterMemory(usage common.MemoryUsage) error {
-	return env.meterMemory(usage.Kind, uint(usage.Amount))
+	if env.sth.EnforceMemoryLimits() {
+		return env.sth.MeterMemory(usage.Kind, uint(usage.Amount))
+	}
+	return nil
 }
 
 // TODO(patrick): rm once Meter object has been refactored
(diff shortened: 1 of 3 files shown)


fvm/env.go
```diff
@@ -283,6 +284,75 @@ func storageMBUFixToBytesUInt(result cadence.Value) uint64 {
...
+	address common.Address,
+) (
+	value uint64,
+	err error,
+) {
+	defer env.StartSpanFromRoot(trace.FVMEnvGetAccountBalance).End()
+
+	err = env.Meter(meter.ComputationKindGetAccountBalance, 1)
+	if err != nil {
+		return value, fmt.Errorf("get account balance failed: %w", err)
```
SaveTheRbtz, line 323:
```suggestion
		return 0, fmt.Errorf("get account balance failed: %w", err)
```

SaveTheRbtz (approved):

janezpodhostnik (approved):

fvm/env.go
```diff
@@ -283,6 +284,75 @@ func storageMBUFixToBytesUInt(result cadence.Value) uint64 {
...
+	address common.Address,
+) (
+	value uint64,
+	err error,
+) {
+	defer env.StartSpanFromRoot(trace.FVMEnvGetStorageCapacity).End()
+
+	err = env.Meter(meter.ComputationKindGetStorageCapacity, 1)
+	if err != nil {
+		return 0, fmt.Errorf("get storage capacity failed: %w", err)
```
zhangchiqing, line 297: Does `Meter` actually get the storage capacity or just increment the meter by 1?

If is later, it's better to mention that it's the metering failed not the actual data query failed

```suggestion
		return 0, fmt.Errorf("meter get storage capacity failed: %w", err)
```

fvm/env.go
zhangchiqing, line 323:
```suggestion
		return value, fmt.Errorf("meter get account balance failed: %w", err)
```

fvm/env.go
```diff
@@ -301,6 +371,80 @@ func (env *commonEnv) GetAccountContractNames(address runtime.Address) ([]string
...
 
+func (env *commonEnv) ResolveLocation(
+	identifiers []runtime.Identifier,
+	location runtime.Location,
+) ([]runtime.ResolvedLocation, error) {
+	defer env.StartExtensiveTracingSpanFromRoot(trace.FVMEnvResolveLocation).End()
+
+	err := env.Meter(meter.ComputationKindResolveLocation, 1)
+	if err != nil {
+		return nil, fmt.Errorf("resolve location failed: %w", err)
```
zhangchiqing, line 382:
```suggestion
		return nil, fmt.Errorf("meter resolve location failed: %w", err)
```

fvm/env.go
```diff
@@ -301,6 +371,80 @@ func (env *commonEnv) GetAccountContractNames(address runtime.Address) ([]string
...
+
+	// if the location is an address,
+	// and no specific identifiers where requested in the import statement,
+	// then fetch all identifiers at this address
+	if len(identifiers) == 0 {
+		address := flow.Address(addressLocation.Address)
+
+		err := env.accounts.CheckAccountNotFrozen(address)
+		if err != nil {
+			return nil, fmt.Errorf("resolving location failed: %w", err)
```
zhangchiqing, line 407:
```suggestion
			return nil, fmt.Errorf("checking account not frozen failed: %w", err)
```

fvm/env.go
```diff
@@ -436,6 +580,49 @@ func (env *commonEnv) Commit() (programs.ModifiedSets, error) {
...
+	hashAlgorithm runtime.HashAlgorithm,
+) (
+	bool,
+	error,
+) {
+	defer env.StartSpanFromRoot(trace.FVMEnvVerifySignature).End()
+
+	err := env.Meter(meter.ComputationKindVerifySignature, 1)
+	if err != nil {
+		return false, fmt.Errorf("verify signature failed: %w", err)
```
zhangchiqing, line 598:
```suggestion
		return false, fmt.Errorf("meter verify signature failed: %w", err)
```

fvm/env.go
```diff
@@ -436,6 +580,49 @@ func (env *commonEnv) Commit() (programs.ModifiedSets, error) {
...
+	}
+
+	return valid, nil
+}
+
+func (env *commonEnv) ValidatePublicKey(pk *runtime.PublicKey) error {
+	err := env.Meter(meter.ComputationKindValidatePublicKey, 1)
+	if err != nil {
+		return fmt.Errorf("validate public key failed: %w", err)
```
zhangchiqing, line 620:
```suggestion
		return fmt.Errorf("meter validate public key failed: %w", err)
```

fvm/env.go
```diff
@@ -472,3 +659,29 @@ func (env *commonEnv) AllocateStorageIndex(owner []byte) (atree.StorageIndex, er
...
+	keyIndex int,
+) (
+	*runtime.AccountKey,
+	error,
+) {
+	defer env.StartSpanFromRoot(trace.FVMEnvGetAccountKey).End()
+
+	err := env.Meter(meter.ComputationKindGetAccountKey, 1)
+	if err != nil {
+		return nil, fmt.Errorf("get account key failed: %w", err)
```
zhangchiqing, line 679:
```suggestion
		return nil, fmt.Errorf("meter get account key failed: %w", err)
```

zhangchiqing (approved):
Nice refactor. 

I just added a few comments for better error messages.

fvm/env.go
pattyshack: the error's detail is specify at the end.  for metering error, the specific string is "computation limit exceed ...".  It should be sufficiently clear.

fvm/env.go
pattyshack: updated this error string since it's less clear.

pattyshack:
bors merge

PR #2988, opened 2022-08-16
Go 1.19 support (libp2p upgrade)

SaveTheRbtz:
* https://github.com/libp2p/go-libp2p-swarm is deprecated.
* https://github.com/libp2p/go-libp2p-discovery is deprecated.
* https://github.com/libp2p/go-tcp-transport is deprecated.
* `blockstore.ErrNotFound` is no longer exported exported and replaced with `ipld.IsNotFound(err)`.
* `go vet` [now complains about wrong usage](https://tip.golang.org/doc/go1.19#vet) of `errors.As` which we have plenty 

As a tiny bonus, a 3% perf improvement based on the ComputeBlock benchmark:
```
$ benchstat old new   
name                             old time/op    new time/op    delta
ComputeBlock/16/cols/128/txes-8     2.09s ± 1%     2.03s ± 2%  -2.99%  (p=0.000 n=9+10)

name                             old us/tx      new us/tx      delta
ComputeBlock/16/cols/128/txes-8     1.02k ± 1%     0.99k ± 2%  -2.98%  (p=0.000 n=9+10)

name                             old alloc/op   new alloc/op   delta
ComputeBlock/16/cols/128/txes-8    1.88GB ± 1%    1.87GB ± 0%    ~     (p=0.370 n=9+8)

name                             old allocs/op  new allocs/op  delta
ComputeBlock/16/cols/128/txes-8     29.3M ± 0%     29.2M ± 0%  -0.13%  (p=0.000 n=8+8)
```


SaveTheRbtz:
@tarakby seems like `github.com/btcsuite/btcd` is no longer developed and crypto would need to migrate to `github.com/btcsuite/btcd/btcec/v2` at some point.

tarakby:
>  seems like github.com/btcsuite/btcd is no longer developed and ./crypto/ would need to migrate to github.com/btcsuite/btcd/btcec/v2 at some point.

There is a community PR with the upgrade, I'll follow up on that.

network/p2p/dht_test.go
```diff
@@ -146,6 +146,8 @@ func TestPubSubWithDHTDiscovery(t *testing.T) {
 	require.NoError(t, err)
 
 	for _, n := range nodes {
+		n := n
```
pattyshack, line 149: undo?

pattyshack (approved):

network/p2p/dht_test.go
SaveTheRbtz: `n` is passed into the goroutine down the line so we need a copy of it.  It was always a bug, but it started racing more consistently with the new versions of libp2p libraries.

network/p2p/keyutils/keyTranslator_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"math"
 	"testing"
 
-	"github.com/btcsuite/btcd/btcec"
+	btcec "github.com/btcsuite/btcd/btcec/v2"
```
SaveTheRbtz, line 9: @tarakby this should be safe, but just in case tagging you.

network/p2p/dht_test.go
pattyshack: you're passing s into the goroutine.  s != n

SaveTheRbtz: n is used inside the function as:
```
ch <- n.Host().ID()
```

AlexHentschel: nice catch!

From my perspective, it would be cleaner to avoid variable shadowing and write 
```golang
subReader := func(s *pubsub.Subscription, node *p2p.Node) {
  msg, err := s.Next(ctx)
  require.NoError(t, err)
  require.NotNil(t, msg)
  assert.Equal(t, data, msg.Data)
  ch <- node.Host().ID()
}
```
and then write in [line 165](https://github.com/onflow/flow-go/blob/b22e0a7c052b75b68659028e32c253fd2dd4cca3/network/p2p/dht_test.go#L165) 
```golang
go subReader(s, n)
```

network/p2p/compressed/compressedStream.go
```diff
@@ -74,6 +74,7 @@ func (c *compressedStream) Read(b []byte) (int, error) {
 func (c *compressedStream) Close() error {
 	c.writeLock.Lock()
 	defer c.writeLock.Unlock()
+	defer c.w.Close()
```
AlexHentschel, line 77: Just to confirm, this means that we are swallowing any errors the this call might return. Are we sure that such errors are not of any interest (even for debugging purposes)? 
The `multierr` seemed a generally applicable solution, just wondering what your reasoning is for using a less-general approach.

AlexHentschel (approved):
Looks good. Thanks for cleaning up the error handling. 
Don't know about your comment [here](https://github.com/onflow/flow-go/pull/2988#discussion_r948201994), so we might want to wait for Tarak's response before merging.

network/p2p/compressed/compressedStream.go
SaveTheRbtz: cc: @yhassanzadeh13

yhassanzadeh13: @AlexHentschel as far as I investigated, closing the libp2p stream also closes the writer on it. However, here the `gzip` writer is a wrapper around the original writer of the libp2p stream. This poses a cyclic causality:
1. We close `gzip` writer (i.e., `c.w.Close()`) _prior_ to closing the stream (this was what the previous `multierr` was doing: `multierr.Combine(c.w.Close(), c.Stream.Close())`) . Then, with the new upgrade, once the `gzip` writer is closing, it also closes the underlying writer of the libp2p stream, which leaves the stream on a half-open status. Then when the original stream is closing (i.e., `c.Stream.Close()`) it fails to close as closing a half-open stream results in a `stream rest`, so it returns an error. 
2. We close `gzip` writer _after_ closing the original stream. Then `gzip` also tries closing the underlying writer of the libp2p stream. But since it is already closed it results in the closing an already closed stream, which again returns a `stream rest` error.

Hence, what matters is to first safely and gracefully close the libp2p stream (which also tells the remote node to free up its resource), and then invoke closing the `gzip` writer that was established on top of the libp2p stream's writer. Closing this `gzip` writer is with the goal of just cleaning up the resource locally making sure that there is no orphan resources lingering post stream closing. However, we are aware that closing that stream will cause a s `stream reset` error. 

To clarify this further, I added an error checking to check the returned error of closing `gzip` writer, and return it if it is not a `stream reset` error: https://github.com/onflow/flow-go/pull/3008

network/p2p/keyutils/keyTranslator_test.go
tarakby: Thanks for the flag. I had a look and all looks good 👌🏼

SaveTheRbtz:
bors merge

SaveTheRbtz:
> @tarakby seems like `github.com/btcsuite/btcd` is no longer developed and `./crypto/` would need to migrate to `github.com/btcsuite/btcd/btcec/v2` at some point.

Got an ack from Tarak on slack for the `btcd` changes.

SaveTheRbtz:
bors retry

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors cancel

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors try

SaveTheRbtz:
bors merge p=-1

PR #2993, opened 2022-08-16
Standardize memory and interaction limit disabling

pattyshack:
This removes PayerIsServiceAcount and AllowContextOverrideByExecutionState
special casing.

Previously, the cached meter state differed depending on which payer account
loaded the program into the cache, resulting in non-deterministic execution.

Note: the cached usage limits are "incorrect", but they does not impact the
final execution result.

Non-deterministic execution due to frozen account, weight updates, service
account metering are all manifestation of the same underlying problem. In the
long run, we should stop caching transaction state as part of program cache.


fvm/fvm.go
```diff
@@ -45,25 +47,34 @@ func NewVirtualMachine(rt runtime.Runtime) *VirtualMachine {
 
 // Run runs a procedure against a ledger in the given context.
 func (vm *VirtualMachine) Run(ctx Context, proc Procedure, v state.View, programs *programs.Programs) (err error) {
-	meterParams, err := getEnvironmentMeterParameters(
+	meterParams := meter.DefaultParameters().
+		WithComputationLimit(uint(proc.ComputationLimit(ctx))).
```
zhangchiqing, line 51: Question, for service account, do we also need to disable computation limit?

pattyshack: I've asked the same question (I personally would have prefer to disable the computation limit).

The answer I got is, since we can specify gas limit (but not memory/interaction limits) as part of the transaction, we should maintain the computation limit and just set the gas limit to max.  This enables other cadence users to use service account transactions as example when they develop their own transaction.

zhangchiqing (approved):

SaveTheRbtz (approved):

janezpodhostnik (approved):
Nice change!

Can you double check that there is a test for the system transaction using up a lot of interaction?

moving event and interaction metering into the meter would make this even cleaner.

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

PR #2994, opened 2022-08-16
Emit Protocol Events after database transaction

jordanschalm:
This PR modifies protocol events:
* protocol events are emitted after the corresponding database transaction completes successfully
  * protocol events may not be delivered, only due to a crash/restart, and only if the corresponding state change (eg. block 100 is finalized) is readable in the Protocol State when the process restarts
* documentation is updated to describe delivery guarantees of protocol events
* adds documentation to all protocol event consumers describing how dropped protocol events due to crashes are handled
* updates `ReactorEngine` to handle missed `EpochCommittedPhaseStarted` events (this was the only case where the kind of possible dropped events were not already handled)

More detail in https://github.com/dapperlabs/flow-go/issues/6317


state/protocol/badger/mutator.go
```diff
@@ -1048,16 +1044,17 @@ func (m *FollowerState) MarkValid(blockID flow.Identifier) error {
...
 	if err != nil {
 		return fmt.Errorf("could not mark block as valid (%x): %w", blockID, err)
 	}
 
+	// skip emitting BlockProcessable for root block, as it is always processable
+	if parent.Height > rootHeight {
+		// emit protocol events after database transaction succeeds
+		m.consumer.BlockProcessable(parent)
+	}
+
```
zhangchiqing, line 1058: ⚠️ I think this might break the guarantee that the order of events to be the consistent with the order of these finalized blocks being saved to database.

Consider `Extend` method is called concurrently for 3 blocks: `A, B, C`, and because of database transaction, they are stored into database in this order `A, B, C`.

After they are stored, `BlockProcessable` will be called outside of the db transaction and in it's own goroutine concurrently, it's possible that a consumer will receive `BlockProcessable` event in `B, A, C`, which is different from the order they are saved in db. 

Should we add a mutex lock to the `Extend` method? But that would disallow concurrent processing of blocks on different fork.

jordanschalm: I don't think we had this guarantee before, either.

Badger guarantees that database state changes are persisted in an ACID-compliant way, it doesn't guarantee anything about the execution order of code which operates on a `*badger.Transaction` object. 

In general, we haven't used or tested concurrent block insertion/validation with the `MutatableState` that I'm aware of. Since this PR is not changing the behaviour, I'm inclined to leave any other changes to `Extend` until we make changes to the `compliance` engine to make use of concurrent block validation.

state/protocol/badger/mutator.go
```diff
@@ -623,19 +620,18 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
-		for _, emit := range events {
-			emit()
-		}
 		return nil
 	})
 	if err != nil {
 		return fmt.Errorf("could not persist finalization operations for block (%x): %w", blockID, err)
 	}
 
+	// emit protocol events after database transaction succeeds
```
zhangchiqing, line 629:
```suggestion
	// emit protocol events after database transaction succeeds
	// event delivery is not guaranteed, consumer needs to fetch missing events.
```

zhangchiqing (approved):

engine/consensus/dkg/reactor_engine.go
```diff
@@ -281,7 +284,9 @@ func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uin
 	if errors.Is(err, storage.ErrNotFound) {
 		log.Warn().Msg("checking beacon key consistency: no key found")
 		err := e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateNoKey)
-		if err != nil {
+		if errors.Is(err, storage.ErrAlreadyExists) {
+			e.log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```
durkmurder, line 288:
```suggestion
			log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```

engine/consensus/dkg/reactor_engine.go
```diff
@@ -305,19 +310,24 @@ func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uin
 			Str("canonical_beacon_pub_key", nextDKGPubKey.String()).
 			Msg("checking beacon key consistency: locally computed beacon public key does not match beacon public key for next epoch")
 		err := e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateInconsistentKey)
-		if err != nil {
+		if errors.Is(err, storage.ErrAlreadyExists) {
+			e.log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```
durkmurder, line 314:
```suggestion
			log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```

engine/consensus/dkg/reactor_engine.go
durkmurder, line 288:
```suggestion
			log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```

engine/consensus/dkg/reactor_engine.go
```diff
@@ -305,19 +310,24 @@ func (e *ReactorEngine) handleEpochCommittedPhaseStarted(currentEpochCounter uin
...
+		} else if err != nil {
 			log.Fatal().Err(err).Msg("failed to set dkg end state")
 		}
 		return
 	}
 
 	err = e.dkgState.SetDKGEndState(nextEpochCounter, flow.DKGEndStateSuccess)
-	if err != nil {
+	if errors.Is(err, storage.ErrAlreadyExists) {
+		e.log.Warn().Err(err).Msg("failed to set dkg end state - already set")
```
durkmurder, line 323: can this happen during normal operations? Isn't it a symptom of bug?

jordanschalm: Good point - with at-least-once delivery we can possibly get here during normal operations, but that isn't possible any more.

jordanschalm: https://github.com/onflow/flow-go/pull/2994/commits/0170a66c5d8d82ebc0b1fdae67c96fe1abb18226

AlexHentschel (commented):
While reviewing, I had a hard time with the [`epochmgr.Engine`](https://github.com/onflow/flow-go/blob/0170a66c5d8d82ebc0b1fdae67c96fe1abb18226/engine/collection/epochmgr/engine.go). Added a corresponding [comment in issue #5659](https://github.com/dapperlabs/flow-go/issues/5659#issuecomment-1292933491) with more details. I think this could all be addressed in one follow-up PR.

For this PR, I have some minor revisions as [PR 3460](https://github.com/onflow/flow-go/pull/3460) (targeting your branch). Please merge, if you are onboard. 

Still reviewing ...

AlexHentschel (approved):
great and clean 👏

jordanschalm:
bors merge

PR #2997, opened 2022-08-16
Add DependableComponent to node builder

peterargue:


cmd/scaffold_test.go
```diff
@@ -527,113 +527,86 @@ func testErrorHandler(logger *testLog, expected error) component.OnError {
...
-type testReadyDone struct {
-	name   string
-	logger *testLog
-
-	readyFn func(string)
-	doneFn  func(string)
-
-	ready chan struct{}
-	done  chan struct{}
+func TestDependableComponentWaitForDependencies(t *testing.T) {
```
yhassanzadeh13, line 530: It adds more value to this test if the same scenario is repeated a couple of times (e.g., 10) and the initialization order follows the expected pattern across all repetitions. Otherwise, a failure may be treated as a flake, while it indicates an actual malfunction, i.e., the desire is the order of initialization to be deterministic.

yhassanzadeh13 (approved):

PR #2998, opened 2022-08-16
add admintool for triggering checkpoint

zhangchiqing:


ledger/complete/compactor.go
```diff
@@ -247,6 +250,24 @@ Loop:
 				c.logger.Info().Msgf("compactor delayed checkpoint %d because prior checkpointing is ongoing", nextCheckpointNum)
 				nextCheckpointNum = activeSegmentNum
 			}
+
+		case _, ok := <-c.triggerCheckpointOnNextSegmentFinishCh:
```
zhangchiqing, line 254: This is the main change to allow compactor to listen to signals from admin tool

m4ksio, line 254: I think it would be better to move it to line 227? Would be clearer than having a nextCheckpoint setup here and wait for another loop

ledger/complete/compactor.go
```diff
@@ -45,16 +45,17 @@ type checkpointResult struct {
...
+	wal                                    realWAL.LedgerWAL
+	trieQueue                              *realWAL.TrieQueue
+	logger                                 zerolog.Logger
+	lm                                     *lifecycle.LifecycleManager
+	observers                              map[observable.Observer]struct{}
+	checkpointDistance                     uint
+	checkpointsToKeep                      uint
+	stopCh                                 chan chan struct{}
+	trieUpdateCh                           <-chan *WALTrieUpdate
+	triggerCheckpointOnNextSegmentFinishCh <-chan interface{} // to trigger checkpoint manually
```
m4ksio, line 58: Would be better to use boolean - so it can be easily turned off and status checked, I don't think there is any benefit to use channel.

zhangchiqing: Good suggestion, close in favor of this https://github.com/onflow/flow-go/pull/3010

PR #2999, opened 2022-08-17
Refactor environment meter into separate objects

pattyshack:


fvm/environment/meter.go
```diff
@@ -0,0 +1,93 @@
...
+) error {
+	// this method is called on every unit of operation, so
+	// checking the context here is the most likely would capture
+	// timeouts or cancellation as soon as they happen, though
+	// we might revisit this when optimizing script execution
+	// by only checking on specific kind of Meter calls.
+	//
+	// in the future this context check should be done inside the cadence
+	select {
+	case <-meter.ctx.Done():
```
zhangchiqing, line 82: Why MeterComputation needs to check meter.ctx.Done(), but not MeterMemory?

zhangchiqing (commented):
Looks good. Just have one question

fvm/environment/meter.go
pattyshack: based on my understanding, this is a hack.  we just it need returning error "quickly enough"

zhangchiqing (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #3010, opened 2022-08-18
add admin tool for triggering checkpoint

zhangchiqing:
This PR adds an admintool to force checkpoint to be triggered as soon as finish writing the current segment WAL file.

### DIFF
--- a/ledger/complete/compactor.go
+++ b/ledger/complete/compactor.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/rs/zerolog"
+	"go.uber.org/atomic"
 	"golang.org/x/sync/semaphore"
 
 	"github.com/onflow/flow-go/ledger"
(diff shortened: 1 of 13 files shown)


ledger/complete/compactor_test.go
```diff
@@ -479,6 +481,95 @@ func TestCompactorAccuracy(t *testing.T) {
...
+		fromBound := 5
+
+		co := CompactorObserver{fromBound: fromBound, done: make(chan struct{})}
+		compactor.Subscribe(&co)
+
+		// Run Compactor in background.
+		<-compactor.Ready()
+
+		fmt.Println("generate the tree and create WAL")
+		fmt.Println("13 segments will trigger checkpoint 5")
```
fxamacker, line 529:
```suggestion
		fmt.Println("6 segments will trigger checkpoint 5")
```

At least 6 segments (from segment 0 to 5) would be needed to trigger checkpoint 5 but there probably wouldn't be 13 segments created from the `for` loop because I think there would be around 1-2 records per segment using the random ranges specified in the test.

Random ranges specified in the test:
```
minPayloadByteSize = 2<<11 - 256 // 3840 bytes
maxPayloadByteSize = 2 << 11     // 4096 bytes
```

ledger/complete/compactor_test.go
```diff
@@ -479,6 +481,95 @@ func TestCompactorAccuracy(t *testing.T) {
...
+		// Shutdown ledger and compactor
+		<-l.Done()
+		<-compactor.Done()
+
+		checkpointer, err := wal.NewCheckpointer()
+		require.NoError(t, err)
+
+		nums, err := checkpointer.Checkpoints()
+		require.NoError(t, err)
+		fmt.Println("checkpoint", nums)
```
fxamacker, line 569: Maybe also test length and elements of `nums`.

fxamacker (commented):
Nice work!  I left some comments, mainly about possibility of updating uber's atomic package to v1.10 and replacing calls to a deprecated function.

ledger/complete/compactor.go
```diff
@@ -220,6 +224,19 @@ Loop:
 				continue
 			}
 
+			// listen to signals from admin tool in order to trigger a checkpoint when the current segment file is finished
+			if c.triggerCheckpointOnNextSegmentFinish.CAS(true, false) {
```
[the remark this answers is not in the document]
zhangchiqing: Good point. 

I will update it in other PRs, as there are many places using this method.

fxamacker (approved):

m4ksio (approved):

zhangchiqing:
bors merge

PR #2971, opened 2022-08-12
Fully initialize meter parameters before transaction execution

pattyshack:
This ensures that all meter parameters are set in one place, which enables us
to safely remove PayerIsServiceAccount special casing (in future PR).

PR #3013, opened 2022-08-18
[FVM] Fix transactions with low gas

janezpodhostnik:
# Context

There was an error occurring on testnet with the new v0.27 ([slack link](https://axiomzen.slack.com/archives/C015G65HR2P/p1660235921269599))

```json
{
  "level": "error",
  "node_role": "execution",
  "node_id": "48601e8a1568f18ca4a574a656710342857f939f598679280a6941cc7b800b6f",
  "error": "[Failure Code: 2000] unknown failure: Execution failed:\nerror: fatal error: storage error: get value failed: [Error Code: 1110] computation exceeds limit (10)\n--> \n",
  "time": "2022-08-11T14:17:02Z",
  "message": "error getting execution memory limit"
}
```

Whats happening here is that the gas limit runs out as the state config (memory weights, execution weights and memory limit) is being loaded, before the transaction really even started execution.

There is a few issues here:
- This error should have not occurred
- This error should not have been fatal
- Because it did occur and was a `Failure` it should have crashed the node

# Severity

- Transactions with low gas limit (<~ 10) will fail like this. 
- This does not crash the node
- This **is** deterministic

We can safely assume that this is not critical. 

If not patched, the downside is that some transactions with very low case limit might fail that shouldn't (but transactions like this should be practically empty), and that the grafana logs will contain error messages like this.

# Solution

1.  Error did not crash the node because it was overridden by a non `Failure` in the defer of the transaction invoker. The simple and dirty fix is to not override if it is a  `Failure`. Fixed with [940c95f](https://github.com/onflow/flow-go/pull/3013/commits/940c95fcef97ea0dd48d6b3f43e74f3f64df06e2). A more proper solution would be to not have so much logic in the defer here.

2. Error was fatal because cadence wrapped the error in an external error which cannot be unwrapped, so the underlying cause of running out of computation was not seen by the split error types. To get around this we already have a method called `HandleRuntimeError` which unwraps the external error. Fixed with [b3565ad](https://github.com/onflow/flow-go/pull/3013/commits/b3565ad0210fa4363ddc9af36caf82aacea9d3f5). This would not be necessary if cadence external error was unwrappable.
 
3. Error happened because loading the execution weights sets them on the current meter immediately. After that loading the memory weights and limit already causes the meter to go over the computation limit. This should not happen as we should read all the settings and then set all the settings. This is already fixed in master with https://github.com/onflow/flow-go/pull/2971. Here I just changed the load order, so execution weights get loaded last [d2129df](https://github.com/onflow/flow-go/pull/3013/commits/d2129dfe003e7c83adf46a529ba3fc0cb3d81606).

4. To get through all of this, I changed an existing test, and added some fuzz tests, just to be more sure no similar edge cases exist.  [9d7d224](https://github.com/onflow/flow-go/pull/3013/commits/9d7d224259297cce715402c7181b7cdf1c4e6523)

5. The test found another bug in the HandleRuntimeError function. So that is fixed now too [90f2679](https://github.com/onflow/flow-go/pull/3013/commits/90f26794b181eee66de76d5923b553a808dbcfef)

# Merging

Everything but [d2129df](https://github.com/onflow/flow-go/pull/3013/commits/d2129dfe003e7c83adf46a529ba3fc0cb3d81606) should get merged to master. [d2129df](https://github.com/onflow/flow-go/pull/3013/commits/d2129dfe003e7c83adf46a529ba3fc0cb3d81606) is already fixed properly.

### DIFF
--- a/fvm/transactionInvoker.go
+++ b/fvm/transactionInvoker.go
@@ -75,8 +75,27 @@ func (i *TransactionInvoker) Process(
 			proc.Events = make([]flow.Event, 0)
 			proc.ServiceEvents = make([]flow.Event, 0)
 		}
+
 		if mergeError := parentState.MergeState(childState, sth.EnforceInteractionLimits()); mergeError != nil {
-			processErr = fmt.Errorf("transaction invocation failed when merging state: %w", mergeError)
+			if processErr != nil {
+				// An error happened while an error was already being handled!
+				// Don't hide the processErr.
+				// But do take the one that is a Failure if one of them is a Failure.
+				var fatal errors.Failure
+				if errors.Is(mergeError, fatal) && !errors.Is(processErr, fatal) {
+					i.logger.Warn().
+						Err(processErr).
+						Msg("Hiding this error because a fatal error occurred during merging state")
+					processErr = fmt.Errorf("transaction invocation failed when merging state: %w", mergeError)
+				} else {
+					i.logger.
+						Warn().
+						Err(mergeError).
+						Msg("Hiding merging state error because an error occurred during transaction processing")
+				}
+			} else {
+				processErr = fmt.Errorf("transaction invocation failed when merging state: %w", mergeError)
+			}
 		}
 		sth.SetActiveState(parentState)
 	}()
--- a/engine/execution/testutil/fixtures.go
+++ b/engine/execution/testutil/fixtures.go
@@ -326,7 +326,7 @@ func BytesToCadenceArray(l []byte) cadence.Array {
 // CreateAccountCreationTransaction creates a transaction which will create a new account.
 //
 // This function returns a randomly generated private key and the transaction.
-func CreateAccountCreationTransaction(t *testing.T, chain flow.Chain) (flow.AccountPrivateKey, *flow.TransactionBody) {
+func CreateAccountCreationTransaction(t testing.TB, chain flow.Chain) (flow.AccountPrivateKey, *flow.TransactionBody) {
 	accountKey, err := GenerateAccountPrivateKey()
 	require.NoError(t, err)
 	encPublicKey := accountKey.PublicKey(1000).PublicKey.Encode()
--- a/fvm/errors/errors.go
+++ b/fvm/errors/errors.go
@@ -69,7 +69,7 @@ func HandleRuntimeError(err error) error {
 	// if is not a runtime error return as vm error
 	// this should never happen unless a bug in the code
 	if runErr, ok = err.(runtime.Error); !ok {
-		return NewUnknownFailure(runErr)
+		return NewUnknownFailure(err)
 	}
 
 	// External errors are reported by the runtime but originate from the VM.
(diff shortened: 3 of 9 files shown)


fvm/transactionInvoker.go
```diff
@@ -75,8 +75,27 @@ func (i *TransactionInvoker) Process(
 			proc.Events = make([]flow.Event, 0)
 			proc.ServiceEvents = make([]flow.Event, 0)
 		}
+
 		if mergeError := parentState.MergeState(childState, sth.EnforceInteractionLimits()); mergeError != nil {
-			processErr = fmt.Errorf("transaction invocation failed when merging state: %w", mergeError)
+			if processErr != nil {
```
zhangchiqing, line 80: I don't quite get who creates this processErr?

Seems like it's always nil, no?

bluesign: https://github.com/onflow/flow-go/blob/90f26794b181eee66de76d5923b553a808dbcfef/fvm/transactionInvoker.go#L228

zhangchiqing: Thanks for pointing out it, @bluesign. I see how it works now. 

I'm a big concern that this type of logic flow. 

It's better to avoid using defer here, we can do this instead:

```
function (i *TransactionInvoker) Process(...) error {
  err := i.process(...)
  if err != nil {
    ...
    sth.SetActiveState(parentState)
  }
}
```

fvm/transactionInvoker.go
```diff
@@ -75,8 +75,27 @@ func (i *TransactionInvoker) Process(
...
+				// An error happened while an error was already being handled!
+				// Don't hide the processErr.
+				// But do take the one that is a Failure if one of them is a Failure.
+				var fatal errors.Failure
+				if errors.Is(mergeError, fatal) && !errors.Is(processErr, fatal) {
```
zhangchiqing, line 85: This logic seems to me, we will still merge state even if processError is fatal? Why?

fvm/transactionInvoker.go
bluesign: @zhangchiqing I don't know this area of code much, but only option I feel makes sense is if we handle some panic not here but on upper layer but also want to merge state here in any case.

fvm/transactionInvoker.go
bluesign: Probably fees

pattyshack: we have to always merge because of spock

fvm/transactionInvoker.go
pattyshack, line 80: This seems unnecessary.  Why not just report format both the merge error string and the original process error string into the new processErr?

zhangchiqing (approved):

pattyshack (approved):
Pre-approve to unblock.  But my comment "This seems unnecessary. Why not just format both the merge error string and the original process error string into the new processErr?" in transaction invoker is still relevant

janezpodhostnik:
This was reverted from v0.27 since it requires a full spork to deploy.

PR #3028, opened 2022-08-19
[Testing] Minor improvements to crash test

SaveTheRbtz:
In this PR:
* Remove the need to pass test name and rely on `t.Name()` instead.
* Add ability to catch error codes > 1. (This may happen in libraries that conform to  BSD's [sysexits.h](https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+14.0-current&arch=default&format=html) or command-line utils, that frequently use `os.Exit(2)`)
* Add ability to catch explicit set of error codes.
* Switch from text parsing to using `cmd.ProcessState.ExitCode()`.

Ref: #2941

### DIFF
--- a/utils/unittest/unittest.go
+++ b/utils/unittest/unittest.go
@@ -414,7 +414,7 @@ func CrashTest(
 
 	if len(expectedStatus) == 0 {
 		// any non-zero status is expected
-		require.Greater(t, cmd.ProcessState.ExitCode(), 0)
+		require.NotEqual(t, cmd.ProcessState.ExitCode(), 0)
 	} else {
 		// expect specific status
 		require.Contains(t, expectedStatus, cmd.ProcessState.ExitCode())
--- a/utils/unittest/unittest.go
+++ b/utils/unittest/unittest.go
@@ -406,7 +406,7 @@ func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string)
 	outBytes, err := cmd.Output()
 	// expect error from run
 	require.Error(t, err)
-	require.Contains(t, "exit status 1", err.Error())
+	require.Contains(t, err.Error(), "exit status ")
 
 	// expect logger.Fatal() message to be pushed to stdout
 	outStr := string(outBytes)
--- a/utils/unittest/unittest.go
+++ b/utils/unittest/unittest.go
@@ -405,6 +405,9 @@ func CrashTestWithExpectedStatus(
 	expectedErrorMsg string,
 	expectedStatus ...int,
 ) {
+	require.NotNil(t, scenario)
+	require.NotEmpty(t, expectedStatus)
+
 	if os.Getenv("CRASH_TEST") == "1" {
 		scenario(t)
 		return
@@ -417,13 +420,8 @@ func CrashTestWithExpectedStatus(
 	// expect error from run
 	require.Error(t, err)
 
-	if len(expectedStatus) == 0 {
-		// any non-zero status is expected
-		require.NotEqual(t, cmd.ProcessState.ExitCode(), 0)
-	} else {
-		// expect specific status
-		require.Contains(t, expectedStatus, cmd.ProcessState.ExitCode())
-	}
+	// expect specific status codes
+	require.Contains(t, expectedStatus, cmd.ProcessState.ExitCode())
 
 	// expect logger.Fatal() message to be pushed to stdout
 	outStr := string(outBytes)
--- a/utils/unittest/unittest.go
+++ b/utils/unittest/unittest.go
@@ -394,13 +394,13 @@ func NetworkTopology() network.Topology {
 
 // CrashTest safely tests functions that crash (as the expected behavior) by checking that running the function creates an error and
 // an expected error message.
-func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string, testName string) {
+func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string) {
 	if os.Getenv("CRASH_TEST") == "1" {
 		scenario(t)
 		return
 	}
 
-	cmd := exec.Command(os.Args[0], "-test.run="+testName)
+	cmd := exec.Command(os.Args[0], "-test.run="+t.Name())
 	cmd.Env = append(os.Environ(), "CRASH_TEST=1")
 
 	outBytes, err := cmd.Output()
--- a/utils/unittest/unittest.go
+++ b/utils/unittest/unittest.go
@@ -417,14 +417,7 @@ func CrashTest(
 		require.Greater(t, cmd.ProcessState.ExitCode(), 0)
 	} else {
 		// expect specific status
-		anyStatus := false
-		for _, status := range expectedStatus {
-			if cmd.ProcessState.ExitCode() == status {
-				anyStatus = true
-				break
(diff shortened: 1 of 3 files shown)


utils/unittest/unittest.go
```diff
@@ -394,19 +394,19 @@ func NetworkTopology() network.Topology {
...
 
-	cmd := exec.Command(os.Args[0], "-test.run="+testName)
+	cmd := exec.Command(os.Args[0], "-test.run="+t.Name())
 	cmd.Env = append(os.Environ(), "CRASH_TEST=1")
 
 	outBytes, err := cmd.Output()
 	// expect error from run
 	require.Error(t, err)
-	require.Contains(t, "exit status 1", err.Error())
+	require.Contains(t, err.Error(), "exit status ")
```
pattyshack, line 409: the "1" trailing "exit status " got accidentally removed?

If it's intentional, then we should pass in the exit status an argument and verify the actual status

utils/unittest/unittest.go
```diff
@@ -394,19 +394,19 @@ func NetworkTopology() network.Topology {
 
 // CrashTest safely tests functions that crash (as the expected behavior) by checking that running the function creates an error and
 // an expected error message.
-func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string, testName string) {
+func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string) {
```
pattyshack, line 397: since we're no longer checking for exit 1, we should pass in the expected set of exit status as well?

i.e., `func CrashTest(t *testing.T, scenario func(*testing.T), expectedErrorMsg string, expectedStatus int, additionalExpectedStatus ...int)`

pattyshack: alternatively, 

```
func CrashTest(
    t *testing.T, scenario func(*testing.T),
    expectedErrorMsg string,
    expectedStatus ...int)
// exit code must be in expectedStatus if it's non-empty.  otherwise, it must be non-zero
```

utils/unittest/unittest.go
```diff
@@ -394,19 +394,38 @@ func NetworkTopology() network.Topology {
...
 	cmd.Env = append(os.Environ(), "CRASH_TEST=1")
 
 	outBytes, err := cmd.Output()
 	// expect error from run
 	require.Error(t, err)
-	require.Contains(t, "exit status 1", err.Error())
+
+	if len(expectedStatus) == 0 {
+		// any non-zero status is expected
+		require.Greater(t, cmd.ProcessState.ExitCode(), 0)
```
pattyshack, line 417: NotEqual rather than Greater ?

pattyshack (approved):

utils/unittest/unittest.go
SaveTheRbtz: Good point, given that it is `int` it can in theory be negative.

utils/unittest/unittest.go
```diff
@@ -394,19 +394,31 @@ func NetworkTopology() network.Topology {
...
-	cmd := exec.Command(os.Args[0], "-test.run="+testName)
+	cmd := exec.Command(os.Args[0], "-test.run="+t.Name())
 	cmd.Env = append(os.Environ(), "CRASH_TEST=1")
 
 	outBytes, err := cmd.Output()
 	// expect error from run
 	require.Error(t, err)
-	require.Contains(t, "exit status 1", err.Error())
+
+	if len(expectedStatus) == 0 {
```
gomisha, line 415: I feel it's too lenient to allow expecting ANY non-zero status. The function under test could crash for some other reason and the unit test could pass, giving us a false sense of security.

utils/unittest/unittest_test.go
```diff
@@ -41,6 +57,11 @@ func crash_ErrorMessage() {
 	os.Exit(1)
 }
 
+func crash_ExitCode() {
```
gomisha, line 60:
```suggestion
func crash_ExitCodeNot1() {
```

utils/unittest/unittest.go
pattyshack: that's a fair comment.  in that case, maybe provide a `CrashTestWithExpectedStatus` method that requires non-empty list, and convert `CrashTest` into a wrapper calling `CrashTestWithExpectedStatus` with `[]int{1}`

utils/unittest/unittest_test.go
```diff
@@ -12,15 +12,31 @@ func TestCrashTest_NoMessage(t *testing.T) {
...
+	}
+	CrashTest(t, f, "exit code != 1")
+}
+
+// TestCrashTest_ExitCodeExplicit tests that CrashTest() catches exit codes other than 1 w/ an explicit exit code.
+func TestCrashTest_ExitCodeExplicit(t *testing.T) {
+	f := func(t *testing.T) {
+		crash_ExitCode()
+	}
+	CrashTest(t, f, "exit code != 1", 4, 3, 2)
```
gomisha, line 39: Do we need the ability to test for multiple exit code codes at the same time? Why can't we just pass a single expected exit code, instead of having the ability to pass a list of exit codes?
That way, if an exit code changed, our unit test would fail and we would be forced to decide if we should expect the new exit code.

utils/unittest/unittest.go
```diff
@@ -394,19 +394,31 @@ func NetworkTopology() network.Topology {
...
+	expectedErrorMsg string,
+	expectedStatus ...int,
+) {
 	if os.Getenv("CRASH_TEST") == "1" {
 		scenario(t)
 		return
 	}
 
-	cmd := exec.Command(os.Args[0], "-test.run="+testName)
+	cmd := exec.Command(os.Args[0], "-test.run="+t.Name())
```
gomisha, line 408: much better to do is this way, then passing in a test name string 👍

utils/unittest/unittest.go
gomisha: I just your comment @pattyshack and I already commented on this in the PR, but why would we pass in a set of expected exit statuses, instead of just a single expected status? Also, why would we make it possible to expect ANY non-zero exit code? That seems too permissive to me because we could miss an error that we weren't expecting.

utils/unittest/unittest.go
gomisha: Why complicate it with a wrapper when we can just change `CrashTest` to take a single `int` of the expected exit code:

```
func CrashTest(
	t *testing.T,
	scenario func(*testing.T),
	expectedErrorMsg string,
	expectedExitCode int,
)
```

I don't see a case where we would want to pass in more than a single expected exit code.

utils/unittest/unittest.go
SaveTheRbtz: Good point.  Let me remove this.

utils/unittest/unittest_test.go
SaveTheRbtz: This is mostly for the cases where you expect a signal but do not know which specific one you'll get.  Most common example is `SIGSEGV` or `SIGBUS`.  It definitely won't hurt to have an ability to specify multiple codes IMO.

gomisha: OK, sounds good

gomisha (approved):

SaveTheRbtz:
bors merge

PR #3034, opened 2022-08-19
[Testing] Enable race detector by default

SaveTheRbtz:
Before we proceed with concurrent execution we first need to be sure that we do not have obvious data races in our codebase.

Depends on: #3092

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,8 @@ jobs:
     - name: Run tests
       if: github.actor != 'bors[bot]'
       run: make ci
+      env:
+        RACE_DETECTOR: 1
     - name: Run tests (Bors)
       if: github.actor == 'bors[bot]'
       uses: nick-invision/retry@v2
@@ -127,6 +129,8 @@ jobs:
     - name: Run tests
       if: github.actor != 'bors[bot]'
       run: ./tools/test_monitor/run-tests.sh
+      env:
+        RACE_DETECTOR: 1
     - name: Run tests (Bors)
       if: github.actor == 'bors[bot]'
       uses: nick-invision/retry@v2
--- a/.github/workflows/flaky-test-monitor-integration.yml
+++ b/.github/workflows/flaky-test-monitor-integration.yml
@@ -64,6 +64,7 @@ jobs:
       env:
         TEST_FLAKY: true
         JSON_OUTPUT: true
+        RACE_DETECTOR: 1
     - name: Process test results
       run: cat test-output | go run tools/test_monitor/level1/process_summary1_results.go
       env:
--- a/.github/workflows/flaky-test-monitor.yml
+++ b/.github/workflows/flaky-test-monitor.yml
@@ -69,6 +69,7 @@ jobs:
       env:
         TEST_FLAKY: true
         JSON_OUTPUT: true
+        RACE_DETECTOR: 1
     - name: Process test results
       run: cat test-output | go run tools/test_monitor/level1/process_summary1_results.go
       env:
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ cmd/util/util:
 .PHONY: unittest-main
 unittest-main:
 	# test all packages with Relic library enabled
-	go test -coverprofile=$(COVER_PROFILE) -covermode=atomic $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic ./...
+	go test -coverprofile=$(COVER_PROFILE) -covermode=atomic $(if $(RACE_DETECTOR),-race,) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic ./...
 
 .PHONY: install-mock-generators
 install-mock-generators:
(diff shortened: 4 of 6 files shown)


SaveTheRbtz:
bors try

SaveTheRbtz:
bors try

SaveTheRbtz:
bors try

module/jobqueue/component_consumer_test.go
```diff
@@ -125,17 +125,19 @@ func (suite *ComponentConsumerSuite) TestHappyPath() {
...
 		defer wg.Done()
 
 		index, err := JobIDToIndex(jobID)
 		assert.NoError(suite.T(), err)
+		suite.T().Logf("job %d finished", index)
 
+		mu.Lock()
+		defer mu.Unlock()
 		finishedJobs[index] = true
-
```
gomisha, line 137: why remove this?

module/jobqueue/component_consumer_test.go
```diff
@@ -125,17 +125,19 @@ func (suite *ComponentConsumerSuite) TestHappyPath() {
...
-		suite.T().Logf("job %d finished", index)
```
gomisha, line 138: why move the message that job is finished to before the `finishedJobs` map is updated?

module/jobqueue/component_consumer_test.go
```diff
@@ -182,6 +184,7 @@ func (suite *ComponentConsumerSuite) TestProgressesOnComplete() {
 	jobData := generateTestData(testJobsCount)
 	finishedJobs := make(map[uint64]bool, testJobsCount)
 
+	mu := sync.Mutex{}
```
pattyshack, line 187: This mutex is pointless since you're no guarding your read with it ... this is still not threadsafe.

network/stub/network.go
```diff
@@ -104,6 +104,17 @@ func (n *Network) Register(channel channels.Channel, engine network.MessageProce
 	return c, nil
 }
 
+func (n *Network) GetEngine(channel channels.Channel) (network.MessageProcessor, bool) {
+	n.Lock()
+	defer n.Unlock()
+	return n.getEngineLocked(channel)
+}
+
+func (n *Network) getEngineLocked(channel channels.Channel) (network.MessageProcessor, bool) {
```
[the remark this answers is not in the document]
pattyshack: i would probably rename this as unsafeGetEngine instead

network/stub/network.go
```diff
@@ -268,7 +279,13 @@ func (n *Network) sendToAllTargets(m *PendingMessage, syncOnProcess bool) error
 		receiverNetwork.seen(key)
 
 		// finds the engine of the targeted Network
-		receiverEngine, ok := receiverNetwork.engines[m.Channel]
+		var ok bool
+		var receiverEngine network.MessageProcessor
+		if receiverNetwork == n {
+			receiverEngine, ok = n.getEngineLocked(m.Channel)
+		} else {
+			receiverEngine, ok = receiverNetwork.GetEngine(m.Channel)
```
pattyshack, line 287: This is a thread safe data race ....

1.  GetEngine returns a valid engine
2. context switch & the engine got unregister
3. context switch back to current goroutine and process

pattyshack: you probably want to replace GetEngine with ProcessWithEngine (The lock should be held while Process() is using the engine).

module/jobqueue/component_consumer_test.go
SaveTheRbtz: read is synchronized by the `done` channel beloww

pattyshack: you're relying on an accidentally memory fence that only happens on test passes. when test fail, you're reading unguarded data

SaveTheRbtz: The test will stop and hence not progress beyond this fence.  That said you have a point: explicit guard is better, let me add it.

network/stub/network.go
SaveTheRbtz: done!  I'm a bit concerned about deadlocks given that we now hold multiple locks but it tis no worse than it was.

SaveTheRbtz:
bors try

module/jobqueue/component_consumer_test.go
```diff
@@ -133,6 +134,8 @@ func (suite *ComponentConsumerSuite) TestHappyPath() {
 		index, err := JobIDToIndex(jobID)
 		assert.NoError(suite.T(), err)
 
+		mu.Lock()
```
pattyshack, line 137: you forgot to fix line 170 ....

network/stub/network.go
```diff
@@ -268,23 +268,35 @@ func (n *Network) sendToAllTargets(m *PendingMessage, syncOnProcess bool) error
 		receiverNetwork.seen(key)
 
 		// finds the engine of the targeted Network
-		receiverEngine, ok := receiverNetwork.engines[m.Channel]
-		if !ok {
-			return fmt.Errorf("could find engine ID: %v for node: %v", m.Channel, nodeID)
+		err := receiverNetwork.processWithEngine(receiverNetwork == n, syncOnProcess, m)
```
pattyshack, line 271: there are multiple issues here.

1. haveSeen & seen should be part of processWithEngine (this currently is a threadsafe data race)
2. this will deadlock.  you need to acquire the receiverNetworks locks in some consistent order or independently to avoid the deadlocking.

I would make probably Hub threadsafe with it's own internal lock, remove top level lock in this function, and consistently grab lock inside processWithEngine instead (also switching to RWMutex).

Also split this out into a different PR since it's not simple

SaveTheRbtz: Ok. Factored jobqueue into https://github.com/onflow/flow-go/pull/3092. Will work on reworking the mocked network.

pattyshack (approved):
pre-approve to unblock (assuming you fix the known races first of course).

gomisha (approved):

SaveTheRbtz:
bors try

SaveTheRbtz:
bors try

SaveTheRbtz:
bors merge

PR #3041, opened 2022-08-21
[Testing] Fix races in BlockState

SaveTheRbtz:
Ref: #3034

### DIFF
--- a/integration/tests/lib/block_state.go
+++ b/integration/tests/lib/block_state.go
@@ -56,7 +56,7 @@ func (bs *BlockState) WaitForBlockById(t *testing.T, blockId flow.Identifier) *m
 	var blockProposal *messages.BlockProposal
 
 	require.Eventually(t, func() bool {
-		bs.RLock() // avoiding concurrent map access
+		bs.RLock()
 		defer bs.RUnlock()
 
 		if block, ok := bs.blocksByID[blockId]; !ok {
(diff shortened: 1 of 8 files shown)


zhangchiqing (approved):

integration/tests/lib/block_state.go
```diff
@@ -136,7 +136,9 @@ func (bs *BlockState) processAncestors(t *testing.T, b *messages.BlockProposal,
 // latest finalized block is 10, this will wait until any block height higher than 10 is finalized, and
 // returns that block.
 func (bs *BlockState) WaitForHighestFinalizedProgress(t *testing.T) *messages.BlockProposal {
```
pattyshack, line 138: This function seems like a flaky test generator.  In particular, currentFinalized could be have been updated even before WaitForHighestFinalizedProgress got executed.  We should pass in the currentFinalized explicitly.

integration/tests/lib/block_state.go
```diff
@@ -136,7 +136,9 @@ func (bs *BlockState) processAncestors(t *testing.T, b *messages.BlockProposal,
...
+	bs.RLock()
```
pattyshack, line 139: per usual, use defer unlock pattern plz (refactor into a help method if necessary)

integration/tests/lib/block_state.go
SaveTheRbtz: Yeah, it seemed quite bad for me at first glance usages are quite legit: it is used as a healthcheck for the network and bootstraps the `currentFinalized` block that is used thereafter in tests.

SaveTheRbtz: Renamed to `WaitForInitialization` to highlight its intended usage.

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #3043, opened 2022-08-22
[Profiler] Add heap allocation diffs

SaveTheRbtz:
Switch `ALLOC_HEAP` to upload "allocs" diff within the profiling duration, while `HEAP` only uploads the "inuse" objects.

Aside from being more consistent with both `pprof` http endpoint and `cloudprofiler` reference library, it also workarounds some of GCP API limits since it greatly reduces profile sizes:

```
Failed to create offline profile: rpc error: code = ResourceExhausted desc = Received message larger than max (46449079 vs. 16777216)
```

### DIFF
--- a/module/profiler/profiler.go
+++ b/module/profiler/profiler.go
@@ -167,7 +167,6 @@ func (p *AutoProfiler) goHeapProfile(sampleTypes ...string) (*profile.Profile, e
 	}
 	prof.TimeNanos = time.Now().UnixNano()
 
-	// TODO(rbtz): add tests.
 	selectedSampleTypes := make([]int, 0, len(sampleTypes))
 	for _, name := range sampleTypes {
 		for i, sampleType := range prof.SampleType {
--- a/module/profiler/profiler.go
+++ b/module/profiler/profiler.go
@@ -146,7 +146,11 @@ func newProfileFunc(name string) profileFunc {
 	}
 }
 
-func (p *AutoProfiler) goHeapProfile() (*profile.Profile, error) {
+func (p *AutoProfiler) goHeapProfile(sampleTypes ...string) (*profile.Profile, error) {
+	if len(sampleTypes) == 0 {
+		return nil, fmt.Errorf("no sample types specified")
+	}
+
 	// Forces the GC before taking each of the heap profiles and improves the profile accuracy.
 	// Autoprofiler runs very infrequently so performance impact is minimal.
 	runtime.GC()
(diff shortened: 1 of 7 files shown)


module/profiler/profiler.go
```diff
@@ -143,11 +146,96 @@ func newProfileFunc(name string) profileFunc {
...
+		return nil, fmt.Errorf("failed to generate allocs profile: %w", err)
+	}
+
+	prof, err := profile.Parse(buf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse allocs profile: %w", err)
+	}
+	prof.TimeNanos = time.Now().UnixNano()
+
+	if got := len(prof.SampleType); got != 4 {
```
pattyshack, line 166: These checks seems very brittle.

1. what happens if the profile returns more types than the expected ones?
2. this assumes a specific sample type ordering.  what happens if the profile returns the samples in different order?

module/profiler/profiler.go
```diff
@@ -143,11 +146,96 @@ func newProfileFunc(name string) profileFunc {
 	}
 }
 
-func (p *AutoProfiler) pprofAllocs(w io.Writer) error {
+func (p *AutoProfiler) goHeapProfile() (*profile.Profile, error) {
```
pattyshack, line 149: pass in the wanted type list instead of using the hardcode list below.  looks like allocs don't care about "inuse_objects" and "inuse_space"

module/profiler/profiler.go
SaveTheRbtz: Your coding standards are [higher than Google's](https://github.com/googleapis/google-cloud-go/blob/main/profiler/heap.go#L107), but fair enough, this code is indeed brittle.

Let me rewrite it in a more robust way, but for that I'll need to improve testing quite a bit.

pattyshack: real googler backend people write c++ =P

module/profiler/profiler.go
```diff
@@ -143,11 +146,99 @@ func newProfileFunc(name string) profileFunc {
...
+				selectedSampleTypes = append(selectedSampleTypes, i)
+			}
+		}
+	}
+	if len(selectedSampleTypes) != len(sampleTypes) {
+		return nil, fmt.Errorf("failed to find all sample types: want: %+v, got: %+v", sampleTypes, prof.SampleType)
+	}
+
+	for i, j := range selectedSampleTypes {
+		prof.SampleType[i] = prof.SampleType[j]
```
pattyshack, line 184: This is wrong.  you need to allocate a new slice (same below).  it accidentally works right now cuz you're passing in the type string argument in the correct/increasing order

SaveTheRbtz: I've purposely did it inplace, but I would assume this indeed looks scary.  Let me re-create the slice instead.

pattyshack: as i've pointed out, you did inline shifting incorrectly.  it works cuz the provided sampleTypes is in the correct order

you want something like this instead:

```
shiftIdx := 0
for i, sampleType := range prof.SampleType {
  include := false
  for _, name := range sampleTypes {
    if sampleType.Type == name {
      include = true
      break
    }
  }
  if include {
    prof.SampleType[shiftIdx] = sampleType
    shiftIdx++
  }
}
prof.SampleType = prof.SampleType[:shiftIdx]
```

pattyshack (approved):

SaveTheRbtz:
Verified on https://console.cloud.google.com/profiler;timespan=12h/flow-localnet-execution;type=HEAP_ALLOC/alloc_objects?project=flow-multi-region

sjonpaulbrown (approved):

SaveTheRbtz:
bors merge

PR #3052, opened 2022-08-22
Khalil/1762 check codec sentinels in topic validator

kc1116:
This PR adds error handling for sentinel errors returned by the network codec during decoding. It also adds slashing data collection for those sentinel errors and replaces hooked logger in the TopicValidator test with a mock slashing violations consumer instead.


network/validator/pubsub/topic_validator.go
```diff
@@ -92,10 +96,24 @@ func TopicValidator(log zerolog.Logger, codec network.Codec, peerFilter func(pee
 		}
 
 		// Convert message payload to a known message type
-		decodedMsgPayload, err := codec.Decode(msg.Payload)
+		decodedMsgPayload, err := c.Decode(msg.Payload)
+		if codec.IsErrUnknownMsgCode(err) {
+			// slash peer if message contains unknown message code byte
+			slashingViolationsConsumer.OnUnknownMsgTypeError(violation(from, msg, err))
```
smnzhu, line 102: Do we still want to keep the logging here in addition to passing it to the slashing violations consumer?

smnzhu (approved):

network/validator/pubsub/topic_validator.go
kc1116: https://github.com/onflow/flow-go/pull/3052/commits/bedd20952d052726576d12579d0f14dfa7bb58d6

peterargue (approved):

PR #3054, opened 2022-08-22
add admintool for triggering checkpoint for v0.26

zhangchiqing:
Add admin tool to trigger checkpoint manually for v0.26.

### DIFF
--- a/ledger/complete/wal/compactor.go
+++ b/ledger/complete/wal/compactor.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/rs/zerolog"
+	"go.uber.org/atomic"
 
 	"github.com/onflow/flow-go/module/lifecycle"
 	"github.com/onflow/flow-go/module/observable"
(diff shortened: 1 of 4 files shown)


ledger/complete/wal/compactor.go
```diff
@@ -111,26 +114,35 @@ func (c *Compactor) createCheckpoints() (int, error) {
 		return -1, fmt.Errorf("cannot get latest checkpoint: %w", err)
 	}
 
+	// if admin tool was used to trigger checkpoint, then it's equivilent to setting checkpoint distance
+	// as 1 for once.
+	if to-from > 1 && c.triggerCheckpointOnNextSegmentFinish.CAS(true, false) {
+		c.logger.Info().Msgf("manually trigger checkpointing (from: %v, to: %v)", from, to)
```
fxamacker, line 120: I think `to` segment isn't checkpointed because this code assumes last segment is being written to and not finalized.

```suggestion
		c.logger.Info().Msgf("manually trigger checkpointing (from: %v, to: %v)", from, to-1)
```

fxamacker (approved):
LGTM!  Just one minor comment about log message.

ledger/complete/wal/compactor.go
```diff
@@ -111,26 +114,35 @@ func (c *Compactor) createCheckpoints() (int, error) {
 		return -1, fmt.Errorf("cannot get latest checkpoint: %w", err)
 	}
 
+	// if admin tool was used to trigger checkpoint, then it's equivilent to setting checkpoint distance
+	// as 1 for once.
+	if to-from > 1 && c.triggerCheckpointOnNextSegmentFinish.CAS(true, false) {
```
m4ksio, line 119: I think we should reset the flag in any case - if it was used to trigger a checkpoint or it just happen than next segment would trigger a checkpoint anyway.
In other words - any checkpoint creation should reset this flag

m4ksio (approved):

PR #3057, opened 2022-08-22
Refactor UUIDGenerator into environment

pattyshack:
1. move UUIDGenerator from state/ into environment/
2. rename (u *UUIDGenerator) to (generator *UUIDGenerator)
3. rename stateHolder to stTxn
4. combine the tracer/meter logic from commonEnv into GenerateUUID

(also renamed meter to meterImpl to deal with name collision)


fvm/environment/uuids.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("generate uuid failed: %w", err)
+	}
+
+	uuid, err := generator.GetUUID()
+	if err != nil {
+		return 0, fmt.Errorf("cannot generate UUID: %w", err)
+	}
+
+	err = generator.SetUUID(uuid + 1)
```
SaveTheRbtz, line 80: this will be a fun one when we get to the concurrent execution =)

SaveTheRbtz (approved):

fvm/environment/uuids.go
pattyshack: there are many ways we can cheat here.  e.g., can switch to a stab allocator style allocator and maintain a freelist.  another easy hack is use the upper bits for sharding.

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3062, opened 2022-08-22
Adding Upsert for BadgerDB operations

Tonix517:
Adding `upsert` operation to BadgerDB operation list. 

Existing `insert()` errs out when key already exists, `update()` errs out when there's no given key already existed. `upsert` can be useful in certain use cases so that we will not have to `get()->insert()` which is unnecessarily slow.

Ref discussion: https://github.com/onflow/flow-go/pull/2743#discussion_r949695662

**Test**
- [X] new unit test cases

### DIFF
--- a/storage/badger/operation/common.go
+++ b/storage/badger/operation/common.go
@@ -115,6 +115,35 @@ func update(key []byte, entity interface{}) func(*badger.Txn) error {
 	}
 }
 
+// upsert will encode the given entity with MsgPack and upsert the binary data
+// under the given key in the badger DB.
+func upsert(key []byte, entity interface{}) func(*badger.Txn) error {
+	return func(tx *badger.Txn) error {
+		// update the maximum key size if the inserted key is bigger
+		if uint32(len(key)) > max {
+			max = uint32(len(key))
+			err := SetMax(tx)
+			if err != nil {
+				return fmt.Errorf("could not update max tracker: %w", err)
+			}
+		}
+
+		// serialize the entity data
+		val, err := msgpack.Marshal(entity)
+		if err != nil {
+			return fmt.Errorf("could not encode entity: %w", err)
+		}
+
+		// persist the entity data into the DB
+		err = tx.Set(key, val)
+		if err != nil {
+			return fmt.Errorf("could not upsert data: %w", err)
+		}
+
+		return nil
+	}
+}
+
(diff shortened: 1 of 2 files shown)


storage/badger/operation/common.go
```diff
@@ -115,6 +115,26 @@ func update(key []byte, entity interface{}) func(*badger.Txn) error {
...
+func upsert(key []byte, entity interface{}) func(*badger.Txn) error {
+	return func(tx *badger.Txn) error {
+		// serialize the entity data
+		val, err := msgpack.Marshal(entity)
+		if err != nil {
+			return fmt.Errorf("could not encode entity: %w", err)
+		}
+
+		// persist the entity data into the DB
+		err = tx.Set(key, val)
```
Tonix517, line 129: this BadgerDB API itself behaves as an `upsert`

m4ksio (approved):

storage/badger/operation/common_test.go
```diff
@@ -188,6 +188,67 @@ func TestUpdateEncodingError(t *testing.T) {
...
+		require.NoError(t, err)
+
+		var act []byte
+		_ = db.View(func(tx *badger.Txn) error {
+			item, err := tx.Get(key)
+			require.NoError(t, err)
+			act, err = item.ValueCopy(nil)
+			require.NoError(t, err)
+			return nil
+		})
```
SaveTheRbtz, line 230: nit: you can probably combine these tests if you upsert data in both cases.

storage/badger/operation/common.go
```diff
@@ -115,6 +115,26 @@ func update(key []byte, entity interface{}) func(*badger.Txn) error {
 	}
 }
 
+// upsert will encode the given entity with MsgPack and upsert the binary data
+// under the given key in the badger DB.
+func upsert(key []byte, entity interface{}) func(*badger.Txn) error {
+	return func(tx *badger.Txn) error {
+		// serialize the entity data
```
SaveTheRbtz, line 122: `insert` code has some cargo-culting around ``len(key)` and `SetMax`.  Do we want to keep it?

SaveTheRbtz (approved):
I have very little knowledge around our badger codebase, but here are my 2cents.

storage/badger/operation/common.go
Tonix517: good catch - yes we do need it when the key being upserted is new

PR #3065, opened 2022-08-23
disable pruner

smnzhu:


engine/testutil/nodes.go
```diff
@@ -575,6 +575,7 @@ func ExecutionNode(t *testing.T, hub *stub.Hub, identity *flow.Identity, identit
 	trackerStorage.On("Update", mock.Anything).Return(func(fn exedatatracker.UpdateFn) error {
 		return fn(func(uint64, ...cid.Cid) error { return nil })
 	})
+	trackerStorage.On("SetFulfilledHeight", mock.Anything).Return(nil)
```
Kay-Zee, line 578: this means that calls to this mock will always succeed, is that the desired behaviour?

Kay-Zee (approved):

zhangchiqing (approved):

PR #3070, opened 2022-08-23
Fix error logging in script execution

zhangchiqing:


janezpodhostnik (approved):

m4ksio:
Isn't that port of https://github.com/onflow/flow-go/pull/3071/files ? Would be great to sync, as the wording in log has changed

tarakby (approved):

SaveTheRbtz:
nit: it's generally better to use typed functions like .Hex() and .Uint64(): it's a bit faster, does not increase GC load and has a benefit of being easily searchable through [LogQL](https://grafana.com/docs/loki/latest/logql/)

PR #3073, opened 2022-08-23
[Testing] Add LoggerForTest to avoid hardcoding Caller

SaveTheRbtz:
Noticed during access integration test troubleshooting.  Most of these hardcoded callers are wrong.

### DIFF
--- a/utils/unittest/logging.go
+++ b/utils/unittest/logging.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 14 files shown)


pattyshack (approved):

Tonix517 (approved):

utils/unittest/logging.go
```diff
@@ -39,6 +42,16 @@ func LoggerWithWriterAndLevel(writer io.Writer, level zerolog.Level) zerolog.Log
 	return log
 }
 
+// go:noinline
+func LoggerForTest(t *testing.T, level zerolog.Level) zerolog.Logger {
+	_, file, _, ok := runtime.Caller(1)
```
janezpodhostnik, line 47: Should `t.Helper()` be called in here?

janezpodhostnik (approved):

utils/unittest/logging.go
SaveTheRbtz: TIL: `t.Helper()` =)

I think we are fine here, since we are not (and would not be) calling any flow-control test functions (e.g. t.Skip/t.Fail/etc) from this function, so user won't ever see this function in the testing output.

SaveTheRbtz:
bors merge

PR #3074, opened 2022-08-23
[Benchmark] Follow sealed blocks

SaveTheRbtz:
Following only sealed blocks is beneficial for the load test since it would verify the end-to-end pipeline and could find bottlenecks in verification nodes.

Sealed state is defined as:
> The verification nodes have verified the transaction (the block in which the transaction is) and the seal is included in the latest block

This also matches the flow-cli behavior: https://github.com/onflow/flow-cli/blob/7401e49afd2d7b89e15dc8f42c75fcb279f4e60f/pkg/flowkit/gateway/grpc.go#L148-L151

While here, also bump the account multiplier, since now we have more transactions inflight.

### DIFF
--- a/integration/benchmark/cmd/manual/main.go
+++ b/integration/benchmark/cmd/manual/main.go
@@ -43,7 +43,7 @@ func main() {
 	pushgateway := flag.String("pushgateway", "127.0.0.1:9091", "host:port for pushgateway")
 	profilerEnabled := flag.Bool("profiler-enabled", false, "whether to enable the auto-profiler")
 	_ = flag.Bool("track-txs", false, "deprecated")
-	accountMultiplierFlag := flag.Int("account-multiplier", 50, "number of accounts to create per load tps")
+	accountMultiplierFlag := flag.Int("account-multiplier", 100, "number of accounts to create per load tps")
 	feedbackEnabled := flag.Bool("feedback-enabled", true, "wait for trannsaction execution before submitting new transaction")
 	maxConstExecTxSizeInBytes := flag.Uint("const-exec-max-tx-size", flow.DefaultMaxTransactionByteSize/10, "max byte size of constant exec transaction size to generate")
 	authAccNumInConstExecTx := flag.Uint("const-exec-num-authorizer", 1, "num of authorizer for each constant exec transaction to generate")
--- a/integration/benchmark/follower.go
+++ b/integration/benchmark/follower.go
@@ -88,7 +88,7 @@ func NewTxFollower(ctx context.Context, client access.Client, opts ...followerOp
 		opt(f)
 	}
 
-	hdr, err := client.GetLatestBlockHeader(newCtx, false)
+	hdr, err := client.GetLatestBlockHeader(newCtx, true)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


integration/benchmark/follower.go
```diff
@@ -114,8 +114,21 @@ Loop:
...
+		}
+		if hdr.Height < nextHeight {
+			f.logger.Trace().Uint64("want", nextHeight).Uint64("got", hdr.Height).
+				Msg("expected block is not yet sealed")
+			continue
+		}
+
 		GetBlockByHeightTime := time.Now()
-		block, err := f.client.GetBlockByHeight(f.ctx, f.height+1)
+		block, err := f.client.GetBlockByHeight(f.ctx, nextHeight)
```
pattyshack, line 131: I thought you fix this in some PR, but it may be good to wait for all the collections to become available before processing the collections (at line 138).  otherwise, you could be reprocessing the same collection multiple times.

integration/benchmark/follower.go
```diff
@@ -114,8 +114,21 @@ Loop:
 		default:
 		}
 
+		nextHeight := f.Height() + 1
```
pattyshack, line 117: could you added a comment the txFollowerImpl struct explaining the threading model?  This code is only correct for single writer / multiple reader

integration/benchmark/follower.go
pattyshack: (fixed in #3078)

pattyshack (approved):

integration/benchmark/follower.go
AndrewM-SDET: Seconding, in general we have a lack of safety related comments.

integration/benchmark/follower.go
SaveTheRbtz: rebased.

AndrewM-SDET (approved):

integration/benchmark/cmd/manual/main.go
```diff
@@ -43,7 +43,7 @@ func main() {
 	pushgateway := flag.String("pushgateway", "127.0.0.1:9091", "host:port for pushgateway")
 	profilerEnabled := flag.Bool("profiler-enabled", false, "whether to enable the auto-profiler")
 	_ = flag.Bool("track-txs", false, "deprecated")
-	accountMultiplierFlag := flag.Int("account-multiplier", 50, "number of accounts to create per load tps")
+	accountMultiplierFlag := flag.Int("account-multiplier", 100, "number of accounts to create per load tps")
```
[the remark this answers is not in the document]
SaveTheRbtz: This one a one-time increase since we now wait around twice as long (verification takes time.) We can speed it up by:
* reusing setup state between runs. (market place simulator already does that).  (Can start factoring it out after #3091 lands.)
* not waiting for seals of account creating is also an option: this would at least make this phase faster.

As for two loaders pointing to the same benchmark cluster-- this may be outside of the scope right now: we are quite far from needing multiple concurrent benchmarks.

SaveTheRbtz:
bors merge

PR #3076, opened 2022-08-23
[Benchmark] Remove race in Digest() print

SaveTheRbtz:
Synchronize stat reads in `Digest()` with a mutex. While here, also group stats into a structure so we don't need to lock/unlock separately for each stat.

### DIFF
--- a/integration/benchmark/worker_stats_tracker.go
+++ b/integration/benchmark/worker_stats_tracker.go
@@ -8,11 +8,15 @@ import (
 	"github.com/jedib0t/go-pretty/table"
 )
 
+type WorkerStats struct {
+	workers int
+	txsSent int
+}
+
 // WorkerStatsTracker keeps track of worker stats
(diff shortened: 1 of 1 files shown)


integration/benchmark/worker_stats_tracker.go
```diff
@@ -44,6 +44,13 @@ func (st *WorkerStatsTracker) AddWorker() {
 	st.workers++
 }
 
+func (st *WorkerStatsTracker) GetWorkres() int {
```
pattyshack, line 47: typo?

pattyshack (approved):

AndrewM-SDET (approved):
Before merge please add a comment explaining how the race is fixed. It is handled by the print getting a mutex locked struct instead of pulling each variable separately, correct?

SaveTheRbtz:
bors merge

PR #3079, opened 2022-08-24
[Benchmark] Faster stopping of workers

SaveTheRbtz:
Do not block on channel at the start of the loop.  This removes delay on `Stop()`.  While here, also add a test case that would verify that.


pattyshack (approved):

AndrewM-SDET (approved):

SaveTheRbtz:
bors merge

PR #3082, opened 2022-08-24
[FVM] Fix account reporter memory limit

janezpodhostnik:
The account reporter was hitting the memory limit. This is because the memory limit was read from the state, instead of the config provided.

This disables the memory limit being read from the state, because it makes no sense here.

### DIFF
--- a/cmd/util/ledger/reporters/account_reporter.go
+++ b/cmd/util/ledger/reporters/account_reporter.go
@@ -139,7 +139,10 @@ type balanceProcessor struct {
 
 func NewBalanceReporter(chain flow.Chain, view state.View) *balanceProcessor {
 	vm := fvm.NewVirtualMachine(fvm.NewInterpreterRuntime())
-	ctx := fvm.NewContext(zerolog.Nop(), fvm.WithChain(chain))
+	ctx := fvm.NewContext(zerolog.Nop(),
+		fvm.WithChain(chain),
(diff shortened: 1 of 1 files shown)


cmd/util/ledger/reporters/account_reporter.go
```diff
@@ -139,7 +139,10 @@ type balanceProcessor struct {
 
 func NewBalanceReporter(chain flow.Chain, view state.View) *balanceProcessor {
 	vm := fvm.NewVirtualMachine(fvm.NewInterpreterRuntime())
-	ctx := fvm.NewContext(zerolog.Nop(), fvm.WithChain(chain))
+	ctx := fvm.NewContext(zerolog.Nop(),
+		fvm.WithChain(chain),
+		// otherwise the memory limit gets overridden
+		fvm.WithAllowContextOverrideByExecutionState(false))
```
pattyshack, line 145: can you send out a pr for the master branch as well, using the WithMemoryAndInteractionLimitsDisabled option instead WithAllowContextOverrideByExecutionState?

pattyshack (approved):

cmd/util/ledger/reporters/account_reporter.go
janezpodhostnik: https://github.com/onflow/flow-go/pull/3105

PR #3090, opened 2022-08-25
[Benchmark] Remove noisy benchmarks

SaveTheRbtz:
Leave only the slowest benchmark since it is the least noisy one.

As a side effect this also speeds up FVM benchmarks by quite a bit and makes output a bit less chaotic.

### DIFF
--- a/engine/execution/computation/manager_benchmark_test.go
+++ b/engine/execution/computation/manager_benchmark_test.go
@@ -146,39 +146,39 @@ func BenchmarkComputeBlock(b *testing.B) {
 		Header:  &flow.Header{},
 		Payload: &flow.Payload{},
 	}
-	for _, cols := range []int{1, 4, 16} {
-		for _, txes := range []int{16, 32, 64, 128} {
-			cols := cols
-			txes := txes
-			b.Run(fmt.Sprintf("%d/cols/%d/txes", cols, txes), func(b *testing.B) {
-				b.StopTimer()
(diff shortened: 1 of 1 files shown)


engine/execution/computation/manager_benchmark_test.go
```diff
@@ -146,8 +146,8 @@ func BenchmarkComputeBlock(b *testing.B) {
 		Header:  &flow.Header{},
 		Payload: &flow.Payload{},
 	}
-	for _, cols := range []int{1, 4, 16} {
-		for _, txes := range []int{16, 32, 64, 128} {
+	for _, cols := range []int{16} {
+		for _, txes := range []int{128} {
```
janezpodhostnik, line 150: What was the reason for having smaller benchmarks in the first place? Was there some scaling issue we were trying to see?

Does it still make sense to keep this as loops?

pattyshack (approved):
clean up the loops as janez suggested plz

engine/execution/computation/manager_benchmark_test.go
SaveTheRbtz: it was useful for estimating per-collection/per-block overhead.  Until we run these tests on a dedicated hardware it produces more noise in FVM benchmark report than it actually worth.

ramtinms (approved):

SaveTheRbtz:
bors merge

Issue #3098, opened 2022-08-25
Enable Cadence performance optimization by reusing environment


turbolent:
## Problem Definition

Cadence v0.26.0 provides a significant performance optimization: It is possible to prepare much of the data structures required for checking and interpretation (such as standard library definitions) once, and reuse them for multiple transaction and script executions.

## Proposed Solution

Modify FVM so that `runtime.Environment` is created once and reused for multiple transactions.

## Definition of Done

Optimization is enabled.

PR #3101, opened 2022-08-26
Reuse cadence runtime environment

pattyshack:
Issue: #3098

Note that I'm not reusing the runtime environment in script execution since that
breaks a bunch of tests.


fvm/contractFunctionInvoker.go
```diff
@@ -46,13 +46,17 @@ func (i *ContractFunctionInvoker) Invoke(env Environment) (cadence.Value, error)
 	span.SetAttributes(i.logSpanAttrs...)
 	defer span.End()
 
+	runtimeEnv := env.BorrowCadenceRuntime()
```
janezpodhostnik, line 49: Should we just one per tx/script and reuse it here? I imagine we would need much less of them that way. 

Not sure if it really matters....

- It might save some time since you don't have to get it and return it to/from a channel each time
- It might save some memory since less `ReusableCadenceRuntimePool`s need to exist at a given time

engine/execution/computation/computer/computer.go
```diff
@@ -46,14 +49,15 @@ type BlockComputer interface {
...
-	executionDataProvider *provider.Provider
+	vm                         VirtualMachine
+	vmCtx                      fvm.Context
+	metrics                    module.ExecutionMetrics
+	tracer                     module.Tracer
+	log                        zerolog.Logger
+	systemChunkCtx             fvm.Context
+	committer                  ViewCommitter
+	executionDataProvider      *provider.Provider
+	reusableCadenceRuntimePool fvm.ReusableCadenceRuntimePool
```
janezpodhostnik, line 60: Should the VM be responsible for holding this, since its already holding the runtime?

fvm/transactionEnv.go
```diff
@@ -190,13 +196,19 @@ func (e *TransactionEnv) GetIsContractDeploymentRestricted() (restricted bool, d
 	restricted, defined = false, false
 	service := runtime.Address(e.ctx.Chain.ServiceAddress())
 
+	runtimeEnv := e.BorrowCadenceRuntime()
+	defer e.ReturnCadenceRuntime(runtimeEnv)
```
janezpodhostnik, line 200: could we just reuse the one from above?

fvm/reusable_cadence_runtime.go
```diff
@@ -0,0 +1,131 @@
...
+	reusable.Declare(setAccountFrozen)
+	return reusable
+}
+
+type ReusableCadenceRuntimePool struct {
+	pool chan *ReusableCadenceRuntime
+}
+
+func NewReusableCadenceRuntimePool(poolSize int) ReusableCadenceRuntimePool {
+	var pool chan *ReusableCadenceRuntime
```
janezpodhostnik, line 96: Might be useful having a debug log entry when a `ReusableCadenceRuntimePool` is created.

janezpodhostnik (commented):
Nice!

Some questions/comments:

1. Should the VM be responsible for holding the `reusableCadenceRuntime` pool?
2. The pool (and related methods) should probably be named `reusableCadenceRuntimeEnvPool` because it doesn't really hold run-times...
3. It might still be better to only request a runtimeEnv from the pool only once per tx/script

pattyshack:
wrt renaming to reusableCadenceRuntimeEnvPool.  I'm planning to move the actual runtime into the reusableCadenceRuntime in a future PR.  the naming will probably make more sense then.

fvm/contractFunctionInvoker.go
janezpodhostnik: This is not an issue, but just a note:

during a transaction:
- you take one from the pool for the tx
- you take another one from the pool for fees (or any other operation)
- then you return both

This is not really an issue right now because the pool size is 1000 but if the pool size was 0 we would be creating and dropping 1 new runtime env each call.

fvm/transactionEnv.go
janezpodhostnik: my bad!

janezpodhostnik:
Would there be any problems moving the pool creation even higher up?
So we would keep the entire pool across blocks?

fvm/contractFunctionInvoker.go
pattyshack: that's no worst than today.  so this doesn't seem like an issue

pattyshack:
o, i thought the manager / computer is reused across blocks.  I'll take a second look

pattyshack:
I've double checked.  there's no need to move the pool out of BlockComputer.  The manager / computer is reused across blocks.

engine/execution/computation/computer/computer.go
```diff
@@ -27,7 +27,10 @@ import (
 	"github.com/onflow/flow-go/utils/logging"
 )
 
-const SystemChunkEventCollectionMaxSize = 256_000_000 // ~256MB
+const (
+	ReusableCadenceRuntimePoolSize    = 1000
```
turbolent, line 31: Maybe improve naming here and everywhere else: the pool is not of runtimes, but *environments*

fvm/reusable_cadence_runtime.go
```diff
@@ -0,0 +1,131 @@
...
+			setAccountFrozenFunctionType,
+		),
+	}
+
+	reusable.Declare(setAccountFrozen)
+	return reusable
+}
+
+type ReusableCadenceRuntimePool struct {
+	pool chan *ReusableCadenceRuntime
```
turbolent, line 92: Is there any reason for using a channel here instead of just `sync.Pool`?

turbolent (commented):
Nice!

engine/execution/computation/computer/computer.go
pattyshack: As I mentioned previously, I'm going to the runtime into ReusableCadenceRuntime object.  The naming will make more sense then

fvm/reusable_cadence_runtime.go
pattyshack: Either one would work.  sync.Pool doesn't save me much typing.  I still need to cast it back to *ReusableCadenceRuntime afterward

turbolent: Isn't the benefit of using a `sync.Pool` that there won't be any contention? If there are more items requested than available, it will create more; whereas the channel is bound to a fixed size

engine/execution/computation/computer/computer.go
turbolent: Ah, I missed the comment above, makes sense. Maybe add a TODO

janezpodhostnik (approved):

fvm/reusable_cadence_runtime.go
pattyshack: since transaction time is so slow, contention here really doesn't matter.  It would matter if we're constantly getting/returning items from the pool, which we aren't.

I'm being very generous with the pool size.  Realistically, I expect maybe 10s of items in the pool (100s if we're amazing)

SaveTheRbtz:
Niiice!

5% perf improvement:
ComputeBlock/16/cols/128/txes-2 | 2.41k ± 1% | 2.31k ± 2% | −4.24% | (p=0.000 n=10+10)
-- | -- | -- | -- | --

And 7% memory reduction:
ComputeBlock/16/cols/128/txes-2 | 22.5M ± 0% | 20.9M ± 0% | −7.06% | (p=0.000 n=10+10)
-- | -- | -- | -- | --

pattyshack:
bors merge

PR #3108, opened 2022-08-26
Add no op send reputation func to network adapter to

kc1116:
This PR adds a no op func SendReputationFeedback that will be used by engines via network conduits to report good/bad messages received.


network/p2p/network.go
```diff
@@ -318,6 +318,18 @@ func (n *Network) UnRegisterChannel(channel channels.Channel) error {
...
+// will report feedBack as true if a message was consumed in some happy path and false if a message was not useful. Currently this
+// fun is a no-op and will log a message.
+func (n *Network) SendReputationFeedback(id peer.ID, messageID string, topic channels.Topic, feedBack bool) {
+	n.logger.Debug().
+		Str("peer_id", id.Pretty()).
+		Str("message_id", messageID).
+		Str("topic", topic.String()).
+		Bool("feedback", feedBack).
+		Msg("sending libP2P reputation feedback for message")
+}
```
yhassanzadeh13, line 331: If the application layer feedback is _negative_, it makes more sense to log at a more severe level, e.g., warn. Also, it adds more observability to have metrics counting the number of negative feedbacks per channel.

kc1116:
Closing this because gossip scoring is complete we can go ahead and do the actual implementation on the conduit side. 
https://github.com/dapperlabs/flow-internal/issues/1841

PR #3110, opened 2022-08-26
[Networking] Fix network stub concurrency

SaveTheRbtz:
This diff fixes concurrency issues in stub networking:
* Protect Hub with it's own lock
* More robust defer unlock pattern in Buffer
* Protects `receiverNetwork` during send instead of sending one.

Ref: #3034

### DIFF
--- a/network/stub/hub.go
+++ b/network/stub/hub.go
@@ -1,6 +1,7 @@
 package stub
 
 import (
+	"sync"
 	"testing"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 // It maintains a set of network instances and enables them to directly exchange message
 // over the memory.
 type Hub struct {
+	sync.Mutex
 	networks map[flow.Identifier]*Network
 	Buffer   *Buffer
 }
@@ -33,7 +35,6 @@ func (h *Hub) DeliverAll() {
 	for _, network := range h.networks {
 		network.DeliverAll(false)
 	}
-
 }
 
 // DeliverAllEventually attempts on delivery of all the buffered messages in the Network instances
@@ -62,6 +63,9 @@ func (h *Hub) DeliverAllEventuallyUntil(t *testing.T, condition func() bool, wai
 
 // GetNetwork returns the Network instance attached to the node ID.
 func (h *Hub) GetNetwork(nodeID flow.Identifier) (*Network, bool) {
+	h.Lock()
+	defer h.Unlock()
+
 	net, ok := h.networks[nodeID]
 	return net, ok
 }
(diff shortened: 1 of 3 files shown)


network/stub/hub.go
```diff
@@ -13,6 +14,7 @@ import (
 // It maintains a set of network instances and enables them to directly exchange message
 // over the memory.
 type Hub struct {
+	sync.Mutex
```
pattyshack, line 17: is buffer threadsafe?  if not that should be guarded as well.

network/stub/hub.go
```diff
@@ -62,12 +63,18 @@ func (h *Hub) DeliverAllEventuallyUntil(t *testing.T, condition func() bool, wai
 
 // GetNetwork returns the Network instance attached to the node ID.
 func (h *Hub) GetNetwork(nodeID flow.Identifier) (*Network, bool) {
+	h.Lock()
```
pattyshack, line 66: maybe use RWMutex

network/stub/network.go
```diff
@@ -258,33 +232,41 @@ func (n *Network) sendToAllTargets(m *PendingMessage, syncOnProcess bool) error
...
-			}
-		} else {
-			// sender and receiver are synced over delivery of message
-			go func() {
-				_ = receiverEngine.Process(m.Channel, m.From, m.Event)
-			}()
-		}
+	receiverEngine, ok := n.engines[m.Channel]
+	if !ok {
+		return fmt.Errorf("could find engine ID: %v", m.Channel)
```
pattyshack, line 257: should still mention which node didn't have an engine?

pattyshack (approved):

network/stub/hub.go
SaveTheRbtz: Yep, already threadsafe.

network/stub/network.go
SaveTheRbtz: it is added up the stack since this error will be wrapped with:
```
		err := receiverNetwork.processWithEngine(syncOnProcess, key, m)
		if err != nil {
			return fmt.Errorf("could not process message for nodeID: %v, %w", nodeID, err)
		}
```

network/stub/network.go
```diff
@@ -258,33 +232,41 @@ func (n *Network) sendToAllTargets(m *PendingMessage, syncOnProcess bool) error
...
 
+	if syncOnProcess {
+		// sender and receiver are synced over processing the message
+		if err := receiverEngine.Process(m.Channel, m.From, m.Event); err != nil {
+			return fmt.Errorf("receiver engine failed to process event (%v): %w", m.Event, err)
+		}
+	} else {
+		// sender and receiver are synced over delivery of message
+		go func() {
+			_ = receiverEngine.Process(m.Channel, m.From, m.Event)
```
yhassanzadeh13, line 268: Not an issue for this PR, but as an attempt to fix the race condition we'd better also consider that `StopConDev` must return only when all buffered messages are delivered. Currently, a race condition may happen if the continuous delivery is stopped prior to the next tick on the `select` statement in `StartContinousDelivery`.

yhassanzadeh13 (approved):

SaveTheRbtz:
bors merge

PR #3112, opened 2022-08-29
[FVM] Basic fuzzing and minor test change

janezpodhostnik:
Add some basic fuzzing to the FVM. 

For now the only way to run this is manually.

I had it running for a while, and nothing was found :tada:.


fvm/fvm_test.go
```diff
@@ -85,6 +95,60 @@ func (vmt vmTest) run(
 	}
 }
 
+// bootstrapWith executes the bootstrap procedure and the custom bootstrap function
```
pattyshack, line 98: maybe move this to fuzz test if you're not going to reuse it elsewhere

fvm/fvm_fuzz_test.go
```diff
@@ -0,0 +1,259 @@
...
+)
+
+func FuzzTransactionComputationLimit(f *testing.F) {
+	// setup initial state
+	vmt, tctx := bootstrapFuzzStateAndTxContext(f)
+
+	f.Add(uint64(0), uint64(0), uint(0))
+	f.Add(uint64(5), uint64(0), uint(0))
+	f.Fuzz(func(t *testing.T, computationLimit uint64, memoryLimit uint64, transactionType uint) {
+		computationLimit %= flow.DefaultMaxTransactionGasLimit
```
pattyshack, line 30: why %=?  should be min?

fvm/fvm_fuzz_test.go
```diff
@@ -0,0 +1,259 @@
...
+		// ==== Create an account ====
+		var txBody *flow.TransactionBody
+		privateKey, txBody = testutil.CreateAccountCreationTransaction(tb, chain)
+
+		err := testutil.SignTransactionAsServiceAccount(txBody, 0, chain)
+		if err != nil {
+			return err
+		}
+
+		tx := fvm.Transaction(txBody, 0)
```
pattyshack, line 212: can you use incrementing tx index?  you can use `programs.NextTxIndexForTesting()` if you don't want to keep track of the tx index manually (same elsewhere)

pattyshack (approved):

fvm/fvm_test.go
janezpodhostnik: Using this, the fvm tests, especially those that repeat the same setup, could be much faster. I think I will be using this in the future.

fvm/fvm_fuzz_test.go
bluesign: it is something like: `random(0, flow.DefaultMaxTransactionGasLimit)`

janezpodhostnik: I did some searching.

I do want randomness here. that is the whole point of the fuzzing tests. `%` preserves the randomness of whatever the fuzzing framework provides. Another option would be skipping if out of bounds. not sure I like that.

fvm/fvm_fuzz_test.go
Tonix517: +1

Tonix517 (approved):

janezpodhostnik:
bors merge

zhangchiqing (approved):

fvm/fvm_fuzz_test.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+type transactionType struct {
+	createTxBody func(t *testing.T, tctx transactionTypeContext) *flow.TransactionBody
+	require      func(t *testing.T, tctx transactionTypeContext, results fuzzResults)
+}
+
+var fuzzTransactionTypes = []transactionType{
+	{
+		createTxBody: func(t *testing.T, tctx transactionTypeContext) *flow.TransactionBody {
```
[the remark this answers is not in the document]
janezpodhostnik: This does more then just verify fee deduction, It also verifies there is no fatal errors and that only the expected error codes are returned. But making it clearer which transaction type we are using in which example is a good idea.

janezpodhostnik:
bors merge

PR #3113, opened 2022-08-29
[FVM] Improve error handling in transaction invoker

janezpodhostnik:
We were quietly hiding some errors in the transaction invoker, if another error occurred during merging state.

### DIFF
--- a/fvm/errors/errors.go
+++ b/fvm/errors/errors.go
@@ -2,6 +2,7 @@ package errors
 
 import (
 	stdErrors "errors"
+	"github.com/rs/zerolog"
 
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/errors"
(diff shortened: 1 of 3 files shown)


fvm/errors/errors.go
```diff
@@ -85,3 +86,44 @@ func HandleRuntimeError(err error) error {
...
+
+	// both are nil return nil
+	if err1 == nil {
+		return nil
+	}
+
+	// if first error is a failure or second is not a failure, return first error, hide the second
+	var failure Failure
+	if As(err1, &failure) || !As(err2, &failure) {
+		logger.
```
pattyshack, line 114: Instead of just logging the masked error, we should return both errors.

looks like we already use "github.com/hashicorp/go-multierror" elsewhere in the codebase

janezpodhostnik: Not sure how that would work with error codes and unwrapping... let me check.

janezpodhostnik: I'm still not entirely sure how returning a multi-error from this location would complicate downstream error handling.
This error eventually reaches the user and is added as the transaction result, which means users would have to know this could happen and handle it appropriately. 

From a users perspective, it might not be easy to know what to do if you get multiple errors...

any thoughts here: @zhangchiqing @ramtinms @m4ksio

pattyshack: The user should handle the error by error code, rather than by the error message.  When I said use multi-err, I meant, keep both error messages, but provide the scarier of the two error codes.

The user would already need to handle both cases since you can return either one today

janezpodhostnik: Error code is currently not a separate field on the transaction result. So users have to check the error string for the code.

pattyshack: that's unfortunate.  we still return both error messages, but strip the less useful error code?

fvm/transactionInvoker.go
```diff
@@ -69,28 +74,33 @@ func (i *TransactionInvoker) Process(
 		if sth.NumNestedTransactions() > 1 {
 			err := sth.RestartNestedTransaction(nestedTxnId)
 			if err != nil {
-				processErr = fmt.Errorf(
-					"cannot restart nested transaction: %w",
-					err,
+				processErr = errors.FirstOrFailure(
```
zhangchiqing, line 77: It's not clear what we do when hitting failure.

It seems `processErr` could be a failure, in that case, do we still want to run `RestartNestedTransaction` in this defer call? What does `RestartNestedTransaction` do?

Shouldn't we stop everything if running into failure (fatal error)?

fvm/transactionInvoker.go
```diff
@@ -69,28 +74,33 @@ func (i *TransactionInvoker) Process(
...
 			proc.Logs = make([]string, 0)
 			proc.Events = make([]flow.Event, 0)
 			proc.ServiceEvents = make([]flow.Event, 0)
-
 		}
 
 		err := sth.Commit(nestedTxnId)
 		if err != nil {
-			processErr = fmt.Errorf("transaction invocation failed when merging state: %w", err)
+			processErr = errors.FirstOrFailure(
```
zhangchiqing, line 99: Same question here, do we still want to commit, if processErr is a failure already?

If we don't run commit when processErr is failure, then we won't have the `err` to handle.

fvm/transactionInvoker.go
pattyshack: restart nested transaction drops the nested transaction's deltas, but preserve the spock (and meter) bookkeeping.  Normally, if we didn't care about spock, we could just abort/discard the nested transaction.

Tonix517:
is this PR still applicable to the repo given our recent other code refactor?

janezpodhostnik:
no

PR #3114, opened 2022-08-29
[FVM] Add interaction limit to FVM fuzzing

janezpodhostnik:
I had to disable limits for some parts of verification.

TODO: Verifications are covered by inclusion fees, not execution fees. They should not be metered towards execution fees. During verification we should only make sure that there are no exploits (making the node work really hard to verify the transaction).

Future TODO: To make the absolute best use of the execution nodes time, the best order of verification/execution would be:
- is the payer frozen
- is the payers signature correct?
- is the proposers sequence number correct?
- does the payer have funds?
- (we are now sure that the payer can pay for this transaction even if it fails. Failing before this point means no fee deduction, and a possible slashing offence for whoever collected this transaction. Failing after this point should always result in some fee deduction)
- check the rest of the signatures and other things
- execute tx


fvm/transactionVerifier.go
```diff
@@ -86,8 +95,18 @@ func (v *TransactionVerifier) verifyTransaction(
 		return nil
 	}
 
+	// Skip checking limits when getting the public keys
+	getPublicKey := func(address flow.Address, keyIndex uint64) (pub flow.AccountPublicKey, err error) {
```
pattyshack, line 99: this should just be a regular TransactionVerifier method

fvm/state/state_holder.go
```diff
@@ -331,6 +331,14 @@ func (s *StateHolder) DisableAllLimitEnforcements() {
 	s.enforceLimits = false
 }
 
+// WithAllLimitsDisabled runs f with limits disabled
+func (s *StateHolder) WithAllLimitsDisabled(f func()) {
+	current := s.enforceLimits
```
Tonix517, line 336: nit: check `nil` for `f`?

fvm/fvm_fuzz_test.go
```diff
@@ -147,9 +152,36 @@ var fuzzTransactionTypes = []transactionType{
 			codes := []errors.ErrorCode{
 				errors.ErrCodeComputationLimitExceededError,
 				errors.ErrCodeCadenceRunTimeError, // because of the panic
+				errors.ErrCodeLedgerInteractionLimitExceededError,
```
zhangchiqing, line 155: This fuzzy test says if there is an error, then the error must be one of them. 

However, if interaction limit exceeded error was hit where it shouldn't, this fuzzy test would still pass. 

I wonder do we have another non-fuzzy test case that verifies if the interaction limit is exceeded, it will report interaction limit exceeded error?

zhangchiqing (approved):

fvm/fvm_fuzz_test.go
janezpodhostnik: yes, there are regular tests for this https://github.com/onflow/flow-go/blob/82493852fa223726a11a3beca11e44f6fc1105af/fvm/fvm_blockcontext_test.go#L999. 

We can improve the tests later to also predict which error is expected with which parameters.

Tonix517 (approved):

janezpodhostnik:
bors merge

PR #3117, opened 2022-08-29
[FVM] Cleanup fvm logging

janezpodhostnik:
Setup the context log in the computer to already have transaction fields on it.

Use only the logger from the FVM context in the FVM.

This is a breaking change for the emulator https://github.com/onflow/flow-emulator/pull/195
This might also break any code dependent on the logs as we were previously not consistent and used both "tx_id"  and "txHash" to log the tx id. Grafana dashboards come to mind.


pattyshack (approved):
nice (I want to do the same thing with Programs.  In most places, i.e., tests, we don't care about it)

Tonix517 (approved):

cmd/execution_builder.go
```diff
@@ -503,11 +503,11 @@ func (e *ExecutionNodeBuilder) LoadComponentsAndModules() {
 
 			vm := fvm.NewVirtualMachine(rt)
 
-			fvmOptions := append([]fvm.Option{}, node.FvmOptions...)
+			fvmOptions := append([]fvm.Option{fvm.WithLogger(node.Logger)}, node.FvmOptions...)
```
m4ksio, line 506: Wouldn't it be better to keep a logger as required parameter rather than an option?

m4ksio (approved):

cmd/execution_builder.go
janezpodhostnik: It is't not a required part for FVM to function, so I think having it as optional might be ok. It was mostly used as noop anyway.

m4ksio: Yes, but that makes it easy to miss. FVM seems crucial enough that we should always have a logger, even if its noop.

janezpodhostnik: This would cause an inconsistency with
- `NewContext(logger)`
- `context.Child().WithLogger(logger)`

I wanted to avoid that, so its always `.WithLogger(logger)`

engine/execution/computation/computer/computer.go
```diff
@@ -399,7 +399,16 @@ func (e *blockComputer) executeTransaction(
...
-	err := e.vm.Run(ctx, tx, txView, programs)
+	childCtx := fvm.NewContextFromParent(ctx,
+		fvm.WithLogger(ctx.Logger.With().
+			Str("tx_id", txID.String()).
+			Uint32("tx_index", txIndex).
+			Str("block_id", res.ExecutableBlock.ID().String()).
+			Uint64("height", res.ExecutableBlock.Block.Header.Height).
+			Bool("system_chunk", isSystemChunk).
+			Logger()),
+	)
```
bluesign, line 410: Doesn't this duplicate fields?

janezpodhostnik: no, In case of a collision, only the last ones are taken.

janezpodhostnik:
bors merge

janezpodhostnik:
bors retry

PR #3121, opened 2022-08-29
Consensus Compliance Engine updated to conform to `Component` interface

jordanschalm:
This PR updates `consensus/compliance` engine to implement `component.Component`. 

The `epochmgr` and `collection/compliance` engine will be updated in a subsequent PR.

## Other Changes
* Replaces type-unsafe communication between `compliance`, `sync`, and `provider` engines with specific interfaces
* Changes `provider` engine from a deprecated `Engine` to a `MessageProcessor`
* Refactors `Stopper` slightly to work with context cancellation


engine/consensus/compliance/engine.go
```diff
@@ -509,7 +511,16 @@ func (e *Engine) BroadcastProposalWithDelay(header *flow.Header, delay time.Dura
...
+	// spawn a goroutine to asynchronously broadcast the proposal - we do this
+	// to introduce a pre-proposal delay without blocking the Hotstuff EventLoop thread
+	go func() {
+		select {
+		case <-time.After(delay):
+		case <-e.cm.ShutdownSignal():
+			return
+		}
+
+		// TODO move into goroutine
```
jordanschalm, line 523:
```suggestion
```

engine/consensus/provider/engine.go
```diff
@@ -77,75 +80,49 @@ func (e *Engine) Done() <-chan struct{} {
...
-// given origin ID.
-func (e *Engine) process(originID flow.Identifier, event interface{}) error {
-	switch ev := event.(type) {
-	case *messages.BlockProposal:
-		return e.onBlockProposal(originID, ev)
-	default:
-		return fmt.Errorf("invalid event type (%T)", event)
+// broadcastProposal is used when we want to broadcast a local block to the rest  of the
+// network (non-consensus nodes). We broadcast to consensus nodes in compliance engine.
+// TODO error handline - differentiate expected errors in Snapshot
```
jordanschalm, line 106:
```suggestion
// TODO error handling - differentiate expected errors in Snapshot
```

consensus/integration/integration_test.go
```diff
@@ -40,21 +42,20 @@ func Test3Nodes(t *testing.T) {
...
-	runNodes(signalerCtx, nodes)
-
-	unittest.RequireCloseBefore(t, stopper.stopped, 30*time.Second, "expect to stop before timeout")
+	unittest.AssertClosesBefore(t, stopper.stopped, 30*time.Second, "expect to stop before timeout")
 
 	allViews := allFinalizedViews(t, nodes)
+	for i, views := range allViews {
+		fmt.Println(i, views)
+	}
 	assertSafety(t, allViews)
```
jordanschalm, line 57:
```suggestion
	assertSafety(t, allViews)
```

consensus/integration/nodes_test.go
```diff
@@ -146,11 +148,7 @@ type Node struct {
 	state             *bprotocol.MutableState
 	headers           *storage.Headers
 	net               *Network
-}
-
-func (n *Node) Shutdown() {
-	<-n.sync.Done()
-	<-n.compliance.Done()
+	cancel            context.CancelFunc
```
jordanschalm, line 151:
```suggestion
```

consensus/integration/stopper_test.go
```diff
@@ -47,11 +50,17 @@ func (s *Stopper) AddNode(n *Node) *StopperConsumer {
 	s.Lock()
 	defer s.Unlock()
 	s.running[n.id.ID()] = struct{}{}
-	s.nodes = append(s.nodes, n)
 	stopConsumer := &StopperConsumer{}
 	return stopConsumer
 }
 
+// WithStopFunc adds a function to use to stop
```
jordanschalm, line 57:
```suggestion
// WithStopFunc adds a function to use to stop all nodes (typically the cancel function of the context used to start them).
```

consensus/integration/stopper_test.go
```diff
@@ -70,25 +79,17 @@ func (s *Stopper) onFinalizedTotal(id flow.Identifier, total uint) {
 	}
 }
 
+// stopAll signals that all nodes should be stopped be closing the `stopped` channel.
+// NOTE: it is the responsibility of the testing code listening on the channel to
+// actually stop the nodes.
```
jordanschalm, line 84:
```suggestion
// stopAll stops all registered nodes, using the provided stopFunc.
```

consensus/integration/stopper_test.go
```diff
@@ -70,25 +79,17 @@ func (s *Stopper) onFinalizedTotal(id flow.Identifier, total uint) {
...
-		go func(node *Node) {
-			node.Shutdown()
-			wg.Done()
-		}(node)
+	if s.stopFunc == nil {
+		panic("Stopper used without a stopFunc - use WithStopFunc to specify how to stop nodes once stop condition is reached")
 	}
-	wg.Wait()
-
+	s.stopFunc()
```
jordanschalm, line 93:
```suggestion
	fmt.Println("stopping all nodes...")
	s.stopFunc()
```

engine/consensus/compliance/engine.go
```diff
@@ -224,71 +236,56 @@ func (e *Engine) WithConsensus(hot module.HotStuff) *Engine {
...
-}
-
-// SubmitLocal submits an event originating on the local node.
-func (e *Engine) SubmitLocal(event interface{}) {
-	err := e.ProcessLocal(event)
-	if err != nil {
-		e.log.Fatal().Err(err).Msg("internal error processing event")
-	}
+	// NOTE: this will create long-lived goroutines each time Ready is called
+	// Since Ready is called infrequently, that is OK. If the call frequency changes, change this code.
```
jordanschalm, line 266:
```suggestion
	// NOTE: this will create long-lived goroutines each time Done is called
	// Since Done is called infrequently, that is OK. If the call frequency changes, change this code.
```

engine/consensus/compliance.go
```diff
@@ -0,0 +1,20 @@
...
+// ProposalProvider provides proposals created by this node to non-consensus nodes.
+type ProposalProvider interface {
+	// ProvideProposal asynchronously submits our proposal to all non-consensus nodes.
+	ProvideProposal(proposal *messages.BlockProposal)
+}
+
+// ComplianceProcessor accepts blocks for ingestion and processing by the compliance layer.
+// It is used to ingest blocks received through other means than the HotStuff happy path,
+// like the block synchronization engine.
+type ComplianceProcessor interface {
```
durkmurder, line 17: Conceptually our engines deal with two types of inputs: inputs from other replicas and locally generated data. While locally generated data is trusted, inputs from other nodes aren't, they are subject to byzantine behavior. 
Based on specification of: [MessageProcessor](https://github.com/onflow/flow-go/blob/feature/active-pacemaker/network/engine.go#L47) this interface is responsible for handling such inputs. I would like to keep that convention and submit everything from other replicas using `Process` and use custom ingestion logic for dealing with locally generated events. Currently we are violating this "contract" by submitting unchecked inputs using other means and not `Process`.

engine/consensus/compliance/engine.go
```diff
@@ -44,39 +45,45 @@ const defaultTimeoutObjectsQueueCapacity = 1000
...
+	headers  storage.Headers
+	payloads storage.Payloads
+	tracer   module.Tracer
+	state    protocol.State
+	prov     consensus.ProposalProvider
+	core     *Core
+	// queues for inbound messsages
+	pendingBlocks         engine.MessageStore
+	pendingRangeResponses engine.MessageStore
+	pendingVotes          engine.MessageStore
```
durkmurder, line 61: we also need to get rid of `pendingVotes` and `pendingTimeouts`, shall we add a TODO at least?

engine/consensus/compliance/engine.go
```diff
@@ -303,31 +300,29 @@ func (e *Engine) Process(channel network.Channel, originID flow.Identifier, even
...
 			}
 		}
 	}
 }
 
-// processAvailableMessages processes any available messages from the inbound queues.
+// processAvailableMessages processes any available messages until the message queue is empty.
 // Only returns when all inbound queues are empty (or the engine is terminated).
-// No errors expected during normal operations.
+// TODO error docs
```
durkmurder, line 323: has something changed? Previously all sentinels were handled in this function, any error returned from this function is an exception and not expected in normal ops.

engine/consensus/provider/engine_test.go
```diff
@@ -53,25 +53,38 @@ func (suite *Suite) SetupTest() {
...
+
+	err := suite.engine.broadcastProposal(proposal)
+	suite.Assert().NoError(err)
 }
 
-func (suite *Suite) OnBlockProposal_Success() {
+func (suite *Suite) TestBroadcastProposal_Invalid() {
 
 	proposal := unittest.ProposalFixture()
+	proposal.Header.ProposerID = unittest.IdentifierFixture() // set a random proposer
```
durkmurder, line 87: It's already random from `unittest.ProposalFixture` isn't it?

engine/consensus/compliance/engine_test.go
```diff
@@ -14,74 +16,88 @@ import (
...
-	cs.hotstuff.On("Start", mock.Anything)
-	cs.hotstuff.On("Ready", mock.Anything).Return(ready)
-	<-cs.engine.Ready()
+	unittest.AssertClosesBefore(cs.T(), cs.engine.Ready(), time.Second)
+}
+
+// TearDownTest stops the engine and checks there are no errors thrown to the SignallerContext.
+func (cs *EngineSuite) TearDownTest() {
+	cs.cancel()
+	<-cs.engine.Done()
```
durkmurder, line 61:
```suggestion
	unittest.AssertClosesBefore(cs.T(), cs.engine.Done(), time.Second)
```

engine/consensus/compliance/engine_test.go
```diff
@@ -14,74 +16,88 @@ import (
...
 
-type ComplianceSuite struct {
-	ComplianceCoreSuite
+// EngineSuite tests the compliance engine.
+type EngineSuite struct {
+	CommonSuite
 
+	ctx    irrecoverable.SignalerContext
+	cancel context.CancelFunc
+	errs   <-chan error
```
durkmurder, line 37: Hm, this seems a bit dangerous, this channel is not read while test is running, should we spin a up a separate goroutine to read from this channel to provide a receiver?

durkmurder: Using `unittest.FailOnIrrecoverableError` for instance

engine/consensus/compliance/engine_test.go
```diff
@@ -187,35 +213,29 @@ func (cs *ComplianceSuite) TestSubmittingMultipleEntries() {
...
-	// in case of local processing error cannot be consumed since all inputs are trusted
-	err = cs.engine.ProcessLocal(invalidEvent)
-	require.Error(cs.T(), err)
-	require.True(cs.T(), engine.IsIncompatibleInputTypeError(err))
-}
-
 // TestOnFinalizedBlock tests if finalized block gets processed when send through `Engine`.
 // Tests the whole processing pipeline.
-func (cs *ComplianceSuite) TestOnFinalizedBlock() {
+func (cs *EngineSuite) TestOnFinalizedBlock() {
```
durkmurder, line 238: We should also fix this test, it uses `require.Eventually` with `AssertCalled`

engine/consensus/compliance/engine_test.go
```diff
@@ -14,74 +16,88 @@ import (
...
-		close(channel)
-		return channel
-	}()
-
-	cs.hotstuff.On("Done", mock.Anything).Return(done)
-
-	// The vote is transmitted asynchronously. We allow 10ms for the vote to be received:
-	<-time.After(10 * time.Millisecond)
+	// wait for vote to be sent
+	<-done
```
durkmurder, line 94:
```suggestion
	unittest.AssertClosesBefore(cs.T(), done, time.Second)
```

engine/consensus/compliance/engine_test.go
```diff
@@ -14,74 +16,88 @@ import (
...
+	cs.cancel()
```
durkmurder, line 95: any reason for cancelling manually? We already do this in `TearDown`

engine/consensus/compliance/engine_test.go
```diff
@@ -97,62 +113,71 @@ func (cs *ComplianceSuite) TestBroadcastProposalWithDelay() {
...
+	// submit to broadcast proposal
+	err := cs.engine.BroadcastProposalWithDelay(&headerFromHotstuff, 0)
+	require.NoError(cs.T(), err, "header broadcast should pass")
 
-	// should fail with wrong block ID (payload unavailable)
-	header.View++
-	err = cs.engine.BroadcastProposalWithDelay(header, 0)
-	require.Error(cs.T(), err, "should fail with missing payload")
-	header.View--
+	<-util.AllClosed(broadcasted, submitted)
```
durkmurder, line 173:
```suggestion
        unittest.AssertClosesBefore(cs.T(), util.AllClosed(broadcasted, submitted), time.Second)
```

engine/consensus/compliance/engine_test.go
```diff
@@ -97,62 +113,71 @@ func (cs *ComplianceSuite) TestBroadcastProposalWithDelay() {
...
+	cs.cancel()
```
durkmurder, line 174: Same question here, I don't think we need to stop manually

engine/consensus/compliance/engine_test.go
```diff
@@ -187,35 +213,29 @@ func (cs *ComplianceSuite) TestSubmittingMultipleEntries() {
...
+	// wait for all messages to be delivered to the engine message queue
 	wg.Wait()
-
-	time.Sleep(time.Second)
+	// wait for the votes queue to drain
+	assert.Eventually(cs.T(), func() bool {
+		return cs.engine.pendingVotes.(*engine.FifoMessageStore).Len() == 0
+	}, time.Second, time.Millisecond*10)
+	// stop the engine
+	cs.cancel()
```
durkmurder, line 228: Do we need to stop manually?

consensus/integration/integration_test.go
```diff
@@ -40,21 +42,20 @@ func Test3Nodes(t *testing.T) {
...
-	signalerCtx, _ := irrecoverable.WithSignaler(ctx)
+	start()
 
-	runNodes(signalerCtx, nodes)
-
-	unittest.RequireCloseBefore(t, stopper.stopped, 30*time.Second, "expect to stop before timeout")
+	unittest.AssertClosesBefore(t, stopper.stopped, 30*time.Second, "expect to stop before timeout")
 
 	allViews := allFinalizedViews(t, nodes)
+	for i, views := range allViews {
```
durkmurder, line 54: is this for debugging ?

engine/consensus/compliance.go
jordanschalm: Good point, for this case I'll change to `MessageProcessor`

engine/consensus/compliance/engine.go
jordanschalm: I changed this because errors are not properly documented or handled yet in calls to `compliance.Core`. The way I have been updating these is to only update error docs for a function once its dependencies are also updated.

engine/consensus/provider/engine_test.go
jordanschalm: It is - this is only for test readability

consensus/integration/integration_test.go
jordanschalm: Yes - will remove this...

engine/consensus/compliance/engine_test.go
jordanschalm: Nice catch 👍

engine/consensus/compliance/engine_test.go
jordanschalm: You're right, no need to cancel within the test. Removing all these...

engine/consensus/compliance/engine_test.go
jordanschalm: 👍 Good point, adding

engine/consensus/compliance/engine_test.go
```diff
@@ -97,62 +112,69 @@ func (cs *ComplianceSuite) TestBroadcastProposalWithDelay() {
...
+		header.View++
+		err := cs.engine.BroadcastProposalWithDelay(&header, 0)
+		require.Error(cs.T(), err, "should fail with missing payload")
+		header.View--
+	})
 
-	// submit to broadcast proposal
-	err := cs.engine.BroadcastProposalWithDelay(block.Header, 0)
-	require.NoError(cs.T(), err, "header broadcast should pass")
+	// unset chain and height to make sure they are correctly reconstructed
```
AlexHentschel, line 141: Could we put [lines 141 - 172](https://github.com/onflow/flow-go/blob/f7b85fbb7c67df694a8b92aa4dc7d96860e54702/engine/consensus/compliance/engine_test.go#L141-L172) into their own `cs.Run`  method to increase the structure of the test?

engine/consensus/compliance/engine_test.go
AlexHentschel: For my learning: is the old implementation (below) actually buggy or just unnecessarily verbose?
https://github.com/onflow/flow-go/blob/f7c9481e2d93a63a82160de645d2d4affc0b584e/engine/consensus/compliance/engine_test.go#L227-L230

engine/consensus/provider/engine.go
```diff
@@ -77,75 +80,49 @@ func (e *Engine) Done() <-chan struct{} {
...
-	default:
-		return fmt.Errorf("invalid event type (%T)", event)
+// broadcastProposal is used when we want to broadcast a local block to the rest  of the
+// network (non-consensus nodes). We broadcast to consensus nodes in compliance engine.
+// TODO error handling - differentiate expected errors in Snapshot
+func (e *Engine) broadcastProposal(proposal *messages.BlockProposal) error {
+
+	// sanity check: we should only broadcast proposals that we proposed here
+	if e.me.NodeID() != proposal.Header.ProposerID {
+		return fmt.Errorf("sanity check failed: attempted to provide another node's proposal (%x!=%x)", e.me.NodeID(), proposal.Header.ProposerID)
```
AlexHentschel, line 111: this sanity check is also in the compliance Engine: https://github.com/onflow/flow-go/blob/f7b85fbb7c67df694a8b92aa4dc7d96860e54702/engine/consensus/compliance/engine.go#L454-L457
do we need it in both places? 
... the check is probably cheap enough that it makes no practical difference, so why not 🤷

AlexHentschel (approved):
Thank you for the nice refactoring, the great documentation and the test revisions. 

In response to [Yurii's comment](https://github.com/onflow/flow-go/pull/3121#discussion_r961677875), I have **created PR https://github.com/onflow/flow-go/pull/3243** (targeting this branch) that adds the missing error documentation.

engine/consensus/compliance/engine.go
AlexHentschel: created PR https://github.com/onflow/flow-go/pull/3243 (targeting this branch) that adds the missing error documentation.

engine/consensus/compliance/engine_test.go
durkmurder: @AlexHentschel `AssertCalled` has  side-effects after first call that doesn't match assert it records a failure and reports it even if subsequent calls asserted to true.

engine/consensus/compliance/engine.go
```diff
@@ -44,39 +45,45 @@ const defaultTimeoutObjectsQueueCapacity = 1000
...
 	log zerolog.Logger,
 	net network.Network,
 	me module.Local,
-	prov network.Engine,
+	prov consensus.ProposalProvider,
 	core *Core,
 ) (*Engine, error) {
+
 	// Inbound FIFO queue for `messages.BlockResponse`s
+	// TODO can be removed along with https://github.com/dapperlabs/flow-go/issues/6254
```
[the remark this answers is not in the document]
durkmurder: This whole saga with `messages.BlockResponse` started as workaround to mitigate timing issue between compliance engine and hotstuff, instead of processing `messages.BlockResponse` we could feed block proposals in loop as it was before.

durkmurder (approved):
LGTM, great effort.

engine/consensus/compliance/engine.go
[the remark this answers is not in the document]
jordanschalm: To add to Yurii's comment, `BlockResponse`s specifically come from the sync engine. On `master` we added logic to deal with those ranges of blocks sequentially (because of the race condition around HotStuff not marking a block as valid before its child is inserted to the state). The race condition isn't a problem with our refactoring for Active Pacemaker, so we don't need to deal with blocks from the sync engine differently from any other block we receive.

consensus/integration/stopper_test.go
```diff
@@ -47,11 +51,17 @@ func (s *Stopper) AddNode(n *Node) *StopperConsumer {
 	s.Lock()
```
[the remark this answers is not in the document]
jordanschalm: Good point - as far as I can tell the `StopperConsumer` does nothing at all. Going to remove it.

engine/consensus/provider/engine.go
jordanschalm: Yeah I'm more inclined to just leave it

jordanschalm:
bors merge

PR #3127, opened 2022-08-30
Move accounts* from state to environment

pattyshack:


fvm/state/state_test.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
+func createByteArray(size int) []byte {
```
Tonix517, line 15: what is this function for?

Tonix517 (approved):

fvm/state/state_test.go
pattyshack: it's copied from accounts_test.go.  we need to duplicated it here since it's also used by these tests

SaveTheRbtz (approved):

pattyshack:
bors merge

ramtinms:
I would recommend adding more details to the PRs, specially in this PR its not clear to me why we are moving accounts to environment, previously we defined every stateful provider to be inside the state package, but this is kinda changing that

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

PR #3132, opened 2022-08-31
Create virtual machine / cadence runtime inside computation.Manager

pattyshack:
Instead of passing the virtual machine (and implicitly the cadence runtime)
into computation.Manager's New function, this creates the virtual machine and
runtime inside the New function itself.

This is prep work for moving the cadence runtime into ReusableCadenceRuntime.

### DIFF
--- a/engine/execution/computation/manager.go
+++ b/engine/execution/computation/manager.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	jsoncdc "github.com/onflow/cadence/encoding/json"
+	"github.com/onflow/cadence/runtime"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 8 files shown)


engine/execution/computation/manager.go
```diff
@@ -53,10 +63,20 @@ type ComputationManager interface {
...
+	ExtensiveTracing         bool
+	ProgramsCacheSize        uint
+	ScriptLogThreshold       time.Duration
+	ScriptExecutionTimeLimit time.Duration
+
+	// When NewCustomVirtualMachine is nil, the manager will create a standard
+	// fvm virtual machine via fvm.NewVirtualMachine.  Otherwise, the manager
+	// will create a virtual machine using this function.
+	//
+	// Note that this is primarily used for testing.
```
zhangchiqing, line 77: Maybe we could create a separate internal function `newWithVirtualMachine`. So that `New` will create a VM using the standard VM, and for test case, which also has access to `newWithVirtualMachine` will create a different VM  for testing purpose

zhangchiqing (approved):

engine/execution/computation/manager.go
pattyshack: The tests completely stub out the vm and has nothing in common with the real vm (see manager_test.go in this PR).  I'm not sure if newWithVirtualMachine will be useful.

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

PR #3135, opened 2022-08-31
[All] Go 1.19

SaveTheRbtz:
Aside from the [new features in 1.19](https://tip.golang.org/doc/go1.19) it will also give us 3-5% performance boost.
(FVM benchmark does not capture this because it is using the same go version on both checkouts.)

Interesting features for us are:
* `GOMEMLIMIT` so that we can run GC more rare than once in two minutes.
* New runtime metrics including: `/cgo/go-to-c-calls:calls` and `/gc/*`.
* Race detector is 2-3x faster.  Also `-asan` is available on Linux.

Depends on: #3136 #3137 #3140 #3141 #2988

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v3
         with:
-          go-version: "1.18"
+          go-version: "1.19"
           cache: true
 
       - name: Build relic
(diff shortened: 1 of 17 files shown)


SaveTheRbtz:
bors try

SaveTheRbtz:
bors try

pattyshack (approved):

gomisha (approved):

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

PR #3139, opened 2022-08-31
Update comments in checkpointer

zhangchiqing:
This PR updates comments in the checkpointer


ledger/complete/mtrie/flattener/iterator.go
```diff
@@ -76,13 +76,7 @@ type NodeIterator struct {
 // NodeIterator created by NewNodeIterator is safe for concurrent use
 // because visitedNodes is always nil in this case.
```
zhangchiqing, line 77: refactor by reusing the logic, and match what is described here in the comment

ledger/complete/wal/checkpointer.go
```diff
@@ -430,7 +465,7 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
 		return fmt.Errorf("cannot write checkpoint footer: %w", err)
 	}
 
-	// Write CRC32 sum
+	// Write CRC32 sum for validation
```
zhangchiqing, line 468: I wonder how to use this self-included checksum, because adding a checksum changes the checksum of the original file. Is there any tool that can automatically read the checksum from footer and compute the checksum of the remaining file and compare them?

ledger/complete/wal/checkpointer.go
```diff
@@ -378,9 +401,13 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
 			if root == nil {
 				continue
 			}
```
zhangchiqing, line 403: hi @fxamacker , could you explain why this `root` variable could possibly be `nil`?

fxamacker: @zhangchiqing

> could you explain why this root variable could possibly be nil?

Empty trie is [always added to forest](https://github.com/onflow/flow-go/blob/c62e3ad9a198fd546c90cb2173720910fad620ff/ledger/complete/mtrie/forest.go#L47-L48) as starting point and empty trie's root is nil. It remains in the forest until evicted by trie queue exceeding capacity.

Maybe add this comment:

```suggestion
                        // Empty trie is always added to forest as starting point and 
                        // empty trie's root is nil. It remains in the forest until evicted
                        // by trie queue exceeding capacity.
			if root == nil {
				continue
			}
```

zhangchiqing: Yeah, I think it makes sense [for root trie nodes](https://github.com/onflow/flow-go/pull/3139/files/77477eecee8ea809f31a8e8715bfe80088a209d4#diff-4d387a87aa55ecd3590e6cfeaa400f14d6aa9823fcbf4cded30e23c0117b767fR425-R427), but for subtrie roots, is it possible for them to be `nil`?

fxamacker: You are right, the comment I suggested is for nil trie root (in a code block in different location).

Checking `root == nil` for subtries is useful for tests but it isn't likely to happen in non-test env.  Generated test data can be more sparse than mainnet's data.

Tonix517 (approved):
thank you for the detailed comments. It will make it easier to understand this part of logic.

ledger/complete/wal/checkpointer.go
pattyshack: why do we need to write the checksum to the same file?  We can create the checkpoint file in a scratch directory, then atomic move the scratch directory to finalize the checkpoint.

pattyshack (approved):

zhangchiqing:
bors merge

ledger/complete/wal/checkpointer.go
Tonix517: Found this which is close to what @pattyshack described, just not sure if there's a more idiomatic solution. https://en.wikipedia.org/wiki/Simple_file_verification#:~:text=Simple%20file%20verification%20(SFV)%20is,usually%20used%20for%20SFV%20files.

pattyshack: The simplest solution is to maintain a manifest file, which keeps track of all the files that are part of this checkpoint and their corresponding checksum. (you will need to checksum the manifest file separately)

Another common approach is to divide the file into fixed size chunks, and checksum each chunk individually (e.g.,
https://github.com/google/leveldb/blob/main/doc/log_format.md)

zhangchiqing: The current approach is to use a CRCWriter and CRCReader to generate and verify the checksum. 

When data bytes are passed to CRCWriter, it will compute the new checksum in memory, and then write the bytes to the file. When all bytes have been written to the file, it appends the checksum to the file. Since the checksum is appended at the end, when the CRCReader reads the bytes, it also computes the checksum, when it reaches the end before reading the checksum at the end, it will first compute the checksum and then compare with the checksum which is read by reading the file again. 

So the self-included checksum is mainly to be used by the CRCReader for validating, rather than external tool to validate.

pattyshack: again, just because we've done something in certain ways in the past, doesn't mean we are stuck with them forever.

e.g., we can easily implement a different crc writer / reader that writes/reads the checksum to different file. i.e., suppose the crc writer is a io.WriteCloser, and we want to write file content into <blah>, the crc writer can create <blah>.crc32 when Close is called (and also closes <blah>).

zhangchiqing: That's good point. I moved this idea into this issue for further discussion. 

https://github.com/dapperlabs/flow-go/issues/6402

PR #3142, opened 2022-09-01
Backport Bitswap stability fix

peterargue:
Backports: https://github.com/onflow/flow-go/pull/2808

### DIFF
--- a/cmd/access/node_builder/access_node_builder.go
+++ b/cmd/access/node_builder/access_node_builder.go
@@ -21,6 +21,7 @@ import (
 	"github.com/onflow/flow/protobuf/go/flow/access"
 
 	"github.com/onflow/flow-go/admin/commands"
+	stateSyncCommands "github.com/onflow/flow-go/admin/commands/state_synchronization"
 	storageCommands "github.com/onflow/flow-go/admin/commands/storage"
 	"github.com/onflow/flow-go/cmd"
 	"github.com/onflow/flow-go/consensus"
@@ -663,9 +664,13 @@ func (builder *FlowAccessNodeBuilder) Initialize() error {
 	// enqueue the regular network
 	builder.EnqueueNetworkInit()
 
-	builder.AdminCommand("get-transactions", func(conf *cmd.NodeConfig) commands.AdminCommand {
-		return storageCommands.NewGetTransactionsCommand(conf.State, conf.Storage.Payloads, conf.Storage.Collections)
-	})
+	builder.
+		AdminCommand("get-transactions", func(conf *cmd.NodeConfig) commands.AdminCommand {
+			return storageCommands.NewGetTransactionsCommand(conf.State, conf.Storage.Payloads, conf.Storage.Collections)
+		}).
+		AdminCommand("read-execution-data", func(conf *cmd.NodeConfig) commands.AdminCommand {
+			return stateSyncCommands.NewReadExecutionDataCommand(builder.ExecutionDataService)
+		})
 
 	// if this is an access node that supports public followers, enqueue the public network
 	if builder.supportsObserver {
--- a/cmd/access/node_builder/access_node_builder.go
+++ b/cmd/access/node_builder/access_node_builder.go
@@ -21,7 +21,6 @@ import (
 	"github.com/onflow/flow/protobuf/go/flow/access"
 
 	"github.com/onflow/flow-go/admin/commands"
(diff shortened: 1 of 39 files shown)


cmd/access/node_builder/access_node_builder.go
```diff
@@ -923,55 +934,79 @@ func (builder *FlowAccessNodeBuilder) Build() (cmd.Node, error) {
...
+				return nil, err
+			}
+
+			builder.AccessNodeConfig.PublicNetworkConfig.Network = net
+
+			node.Logger.Info().Msgf("network will run on address: %s", builder.PublicNetworkConfig.BindAddress)
+			return net, nil
+		}).
+		Component("peer manager", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
+			// TODO: make this component start after its dependencies are ready. This will require a dependable type of component
```
peterargue, line 985: this is for the public network, and isn't necessary until we enable bitswap for observers

yhassanzadeh13 (approved):

network/p2p/middleware.go
```diff
@@ -179,7 +167,9 @@ func NewMiddleware(
 			ready()
 
 			<-ctx.Done()
-			mw.stop()
+
+			// wait for the readConnection and readSubscription routines to stop
+			mw.wg.Wait()
```
zhangchiqing, line 172: Could we add some log how long this wait will take?

I'm a bit concerned this Wait will take too long.

cmd/scaffold.go
```diff
@@ -263,44 +266,65 @@ func (fnb *FlowNodeBuilder) EnqueueNetworkInit() {
...
+
+	fnb.Module("middleware dependency", func(node *NodeConfig) error {
+		fnb.middlewareDependable = module.NewProxiedReadyDoneAware()
+		fnb.PeerManagerDependencies = append(fnb.PeerManagerDependencies, fnb.middlewareDependable)
+		return nil
+	})
+
+	fnb.DependableComponent("peer manager", func(node *NodeConfig) (module.ReadyDoneAware, error) {
+		return fnb.LibP2PNode.PeerManagerComponent(), nil
+	}, fnb.PeerManagerDependencies)
```
zhangchiqing, line 310:
```suggestion
	// peer manager won't be created until all PeerManagerDependencies are ready.
	fnb.DependableComponent("peer manager", func(node *NodeConfig) (module.ReadyDoneAware, error) {
		return fnb.LibP2PNode.PeerManagerComponent(), nil
	}, fnb.PeerManagerDependencies)
```

cmd/access/node_builder/access_node_builder.go
```diff
@@ -428,13 +429,22 @@ func (builder *FlowAccessNodeBuilder) BuildExecutionDataRequester() *FlowAccessN
...
+		}).
+		Component("execution data downloader", func(node *cmd.NodeConfig) (module.ReadyDoneAware, error) {
 			var err error
 			bs, err = node.Network.RegisterBlobService(channels.ExecutionDataService, ds)
 			if err != nil {
 				return nil, fmt.Errorf("could not register blob service: %w", err)
 			}
 
+			// add blobservice into ReadyDoneAware dependency passed to peer manager
+			// this configures peer manager to wait for the blobservice to be ready before starting
```
zhangchiqing, line 445:
```suggestion
			// this starts the blob service and configures peer manager to wait for the blobservice 
			// to be ready before starting
```

zhangchiqing (commented):
Looks good. Just had one question about the Wait

network/p2p/middleware.go
peterargue: sure. this was taken directly from the previous version though, so there's no change in behavior

zhangchiqing: Can you point out where is the code from the pervious version?

Github shows the change as changing from `mv.stop()` to `mv.wg.Wait()`.  So I thought the logic here is changed, the mv is now stopped elsewhere, and here, we are simply waiting for the result of the `stop`

peterargue: the original call was here:
https://github.com/onflow/flow-go/pull/3142/files#diff-95120943dbf1b8334da6c35dc229d380cea1f66bb6751a652aab627b9082ffa8L352

This PR converts the peer manager and libp2p node to components, so middleware no longer needs to manage their shutdown. `stop()` became just that single line so I removed the method

zhangchiqing (approved):
Looks good

PR #3144, opened 2022-09-01
Merge meter (and interaction) unconditionally

pattyshack:
Nested transaction commits no longer fails due to merging meter.

The next metering call (if any) will notice that the limit is reach, so it's
generally safe to not check the limit during merging.


zhangchiqing (approved):

Tonix517 (approved):

pattyshack:
bors merge

PR #3147, opened 2022-09-01
Fix flaky TopicValidatorTest

kc1116:
This PR fixes flaky TopicValidatorTest. There are 2 parts to this fix. 
1. Remove the require.Eventually that waited until each peer's peer list is updated. It is not necessary because peers may be connected before the peer list gets updated.
2. Generate peer info from identity struct instead of using  _*host.InfoFromHost_  . When getting the peer info to the add a peer directly from the host object it will generate a peer info with 2 multi address' like the following. 

     `[/ip4/192.168.1.163/tcp/63944 /ip4/127.0.0.1/tcp/63944]`

     Using the IP address' `192.168.1.163 or 127.0.0.1` cause flakiness when the host can not be resolved. Instead use 
     `p2p.PeerAddressInfo(identity)` so that the `0.0.0.0 ` is used as the IP in the multiaddress. 

     `[/ip4/0.0.0.0/tcp/63944]`


network/p2p/topic_validator_test.go
```diff
@@ -60,9 +57,12 @@ func TestTopicValidator_Unstaked(t *testing.T) {
 		return nil
 	}
 
+	pInfo2, err := p2p.PeerAddressInfo(identity2)
+	require.NoError(t, err)
+
 	// node1 is connected to node2
 	// sn1 <-> sn2
-	require.NoError(t, sn1.AddPeer(context.TODO(), *host.InfoFromHost(sn2.Host())))
+	require.NoError(t, sn1.AddPeer(context.TODO(), pInfo2))
```
peterargue, line 65: so the fix was to use `p2p.PeerAddressInfo(identity2)` instead of `host.InfoFromHost(sn2.Host())` to get the peer's `PeerInfo`?

network/p2p/topic_validator_test.go
```diff
@@ -73,11 +73,8 @@ func TestTopicValidator_Unstaked(t *testing.T) {
 	_, err = sn2.Subscribe(topic, unittest.NetworkCodec(), unittest.AllowAllPeerFilter(), slashingViolationsConsumer)
 	require.NoError(t, err)
 
-	// assert that the nodes are connected as expected
-	require.Eventually(t, func() bool {
-		return len(sn1.ListPeers(topic.String())) > 0 &&
-			len(sn2.ListPeers(topic.String())) > 0
-	}, 3*time.Second, 100*time.Millisecond)
```
peterargue, line 80: Since this was checking just that the peer list was non-empty, wouldn't that account for peers that connected early?

network/p2p/topic_validator_test.go
kc1116: Updated the description.

network/p2p/topic_validator_test.go
kc1116: No in general the 2 processes of subscribing and discovering peers interested in the topic are async. This discovery and connect process may finish before this peer list is updated or the peer list may not be updated at all in some cases. Thus checking this peer list is unnecessary and causes flakiness.

peterargue (approved):

kc1116:
bors merge

PR #3148, opened 2022-09-01
[FVM] Use embed for cadence scripts in blueprints

janezpodhostnik:
originally from:  https://github.com/onflow/flow-go/pull/2684

- extract cadence scripts into separate cdc files (this way we can late write tests for the cadence scripts or run some CI automation over them)
- change some constants from a (domain string, path string) pair to a cadence path, so there is less code
- change cadence template to use arguments and the template replace function from core-contracts


Tonix517 (approved):
LGTM

janezpodhostnik:
converting to draft since I forgot to fix the integration tests.

new.txt
```diff
@@ -0,0 +1,135 @@
+-test.shuffle 1657722833025907000
```
pattyshack, line 1: Remove?  Looks like this got accidentally included

pattyshack (approved):
in the future, could you split this into multiple PRs?  it's easier than reviewing individual commits

m4ksio (approved):

janezpodhostnik:
bors merge

PR #3163, opened 2022-09-01
Update to use Loader transaction tracking

AndrewM-SDET:
Swap from using Prometheus data to polling the completed transactions that are held directly in the loader via contLoadGenerator.go


AndrewM-SDET:
This work is still in progress as this implementation is causing an issue where the accounts are being exhausted before the test is actually starting somehow. Swapping over from using thread.Sleep() to something more akin to the pushgateway using time.Ticker

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,56 @@ func main() {
...
 		log.Fatal().Err(err).Msgf("unable to init loader")
 	}
+
+	// run load
 	lg.Start()
 
-	testStartTime := time.Now()
+	// prepare data slices
+	dataSlices := make([]dataSlice, 0)
+	go func() {
```
SaveTheRbtz, line 179: you may want to factor it out into a function (it'll be easier to both read and unnittest it).

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,56 @@ func main() {
 	if err != nil {
 		log.Fatal().Err(err).Msgf("unable to init loader")
 	}
+
+	// run load
 	lg.Start()
```
SaveTheRbtz, line 175: you may want to split this into 1) `Setup()` and 2) `Start()`.  That way you won't need the `leadDuration` delay.

PS. Given that there is already `Init()` you can move all the pre-benchmark code there and rename it to `Setup`.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,56 @@ func main() {
...
+
+		// wait through lead duration.
+		time.Sleep(leadDuration)
+
+		// grab time into start time
+		startTime := time.Now()
+		// grab total executed transactions into start transactions
+		startExecutedTransactions := lg.GetTxExecuted()
+
+		for !lg.Stopped {
```
SaveTheRbtz, line 191: This is a data race.  You may consider adding a method to the loader that would indicate that it finished.  It may even return a channel so that you can `select` on it.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,56 @@ func main() {
...
+				RunStartTime: runStartTime}
+
+			startExecutedTransactions = endExecutedTransaction
+			startTime = endTime
+
+			dataSlices = append(dataSlices, slice)
+		}
+	}()
+
 	time.Sleep(loadCase.duration)
```
SaveTheRbtz, line 219: Ideally you do not want to sleep here, but instead select on two channels: 1) test timeout 2) some indication from the loader that it has finished.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,56 @@ func main() {
...
 	lg.Stop()
```
SaveTheRbtz, line 220: this should be in defer.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +170,59 @@ func main() {
...
+
+		// wait through lead duration.
+		time.Sleep(leadDuration)
+
+		// grab time into start time
+		startTime := time.Now()
+		// grab total executed transactions into start transactions
+		startExecutedTransactions := lg.GetTxExecuted()
+
+		t := time.NewTicker(sliceDuration)
```
pattyshack, line 191: need to defer cleanup the ticker

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +168,76 @@ func main() {
...
-	//remove the lead time on both start and end, this should remove spin-up and spin-down times
-	leadDuration, _ := time.ParseDuration(leadTime)
-	endTime := end.Add(-1 * leadDuration)
-	sliceDuration, _ := time.ParseDuration(sliceTime)
+	// run load
+	lg.Start()
 
-	slices := make([]dataSlice, 0)
+	// prepare data slices
+	dataSlices := make([]dataSlice, 0)
```
pattyshack, line 176: This seems wrong.  `slice = append(slices, slice)` in line 236 won't modify this copy of the dataSlices variable.  As a result, your prepareDataForBigQuery call in the next line will always end up with an empty slice.

you probably want something more like this

```
var dataSlices []dataSlice
defer prepareDataForBigQuery(dataSlices, *ciFlag)
...
go func() {
  dataSlices = recordTransactionData(
    lg,
    sliceDuration,
    runStartTime,
    gitSha,
    goVersion,
    osVersion)
```

and return the slices from recordTransactionData when it exits

pattyshack: or better yet, use a channel to pass the dataSlices back (my above sample code is not threadsafe)

AndrewM-SDET: clever to use the channels to pass them back. I'll update to do that. Thanks.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,57 +168,76 @@ func main() {
...
-		sliceEndTime := currentTime.Add(sliceDuration)
+	defer lg.Stop()
 
-		inputTps := lg.AvgTpsBetween(currentTime, sliceEndTime)
-		proStart := getPrometheusTransactionAtTime(currentTime)
-		proEnd := getPrometheusTransactionAtTime(sliceEndTime)
+	sliceDuration, _ := time.ParseDuration(*sliceSize)
 
-		outputTps := (proEnd - proStart) / (sliceDuration).Seconds()
+	go recordTransactionData(
```
pattyshack, line 183: btw, I don't think you need  to use a goroutine here.  seems like you just want recordTransactionData to exit after some time.  you can pass a context (configured with `context.WithTimeout(...)`) into recordTransactionData, and check if the context is cancelled as part of select

this would simplify things a bit, and you don't have to worry about thread safety

AndrewM-SDET: I was thinking of using a `context.WithTimeout(...)` for removing the time.Sleep() at line 169 so this aligns well with that. Will do.

integration/benchmark/contLoadGenerator.go
```diff
@@ -60,6 +60,7 @@ type ContLoadGenerator struct {
 	workerStatsTracker  *WorkerStatsTracker
 	workers             []*Worker
 	stopped             bool
+	stoppedChannel      chan bool
```
SaveTheRbtz, line 63: you may want to remove stopped and just use this channel instead.

Also you might want to use `chan struct{}` instead.

integration/benchmark/contLoadGenerator.go
```diff
@@ -286,6 +288,11 @@ func (lg *ContLoadGenerator) Stop() {
 	lg.workerStatsTracker.StopPrinting()
 	lg.log.Debug().Msg("stopping follower")
 	lg.follower.Stop()
+	lg.stoppedChannel <- true
```
SaveTheRbtz, line 291: you want to just `close` the channel here.

integration/benchmark/contLoadGenerator.go
```diff
@@ -286,6 +288,11 @@ func (lg *ContLoadGenerator) Stop() {
...
+}
+
+func (lg *ContLoadGenerator) GetStoppedChannel() chan bool {
```
SaveTheRbtz, line 294: you might consider to make the signature compatible with the `context`'s Done, e.g.: 
```Done() <-chan struct{}```

File: integration/benchmark/contLoadGenerator.go (hunk 9 above)
Comment (AndrewM-SDET) [reply integration/benchmark/contLoadGenerator.go:63]: Swapping to `chan struct{}` but leaving stopped as a separate var for now as replacing it would complicate the check in `Init()` Will leave a TODO in the struct definition for later cleanup.

File: integration/benchmark/cmd/ci/main.go (hunk 2 above)
Comment (AndrewM-SDET) [reply integration/benchmark/cmd/ci/main.go:173]: I have removed `leadDuration` entirely as it wasn't providing much context or added data. If we find that the first and last slices are not providing useful data it will be easier to discard them after calculation rather than try to split the code up.

File: integration/benchmark/cmd/ci/main.go (hunk 8 above)
Comment (AndrewM-SDET) [reply integration/benchmark/cmd/ci/main.go:183]: With Alexey's suggestion to swap to using channels to check for the timeout and cancelations to remove our `Sleep()` calls this no longer fits cleanly and from what I can tell we will still require the subroutine.

File: integration/benchmark/cmd/ci/main.go [hunk 12]
```diff
@@ -163,60 +167,93 @@ func main() {
...
 	}
+
+	// run load
 	lg.Start()
 
-	testStartTime := time.Now()
-	time.Sleep(loadCase.duration)
-	lg.Stop()
-	testEndTime := time.Now()
+	testRunDuration := time.NewTicker(loadCase.duration)
```
AndrewM-SDET, line 174: Realized as I was porting this change over to manual/main.go that I had been conflating `time.Ticker` and `time.Timer` this should be swapped out with `time.Timer` since it is a one off event. Will be in my next commit.

AndrewM-SDET:
We have updated all relevant files but a new bug has been discovered in benchmark/cmd/ci.go where even though the timers are set up and we are entering `recordTransactionData()` no slices are being generated at any point. Investigating this issue now.

.gitignore
```diff
@@ -13,6 +13,9 @@ crypto/relic/
 # Test binary, build with `go test -c`
 *.test
 
+# BigQuery result files
+tps*.json
+
```
SaveTheRbtz, line 18: should not be needed after the #3241

integration/benchmark/cmd/ci/main.go
SaveTheRbtz: You do not need a goroutine here: you can jut return `dataSlices` from this function.

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,60 +167,95 @@ func main() {
...
 
-	testStartTime := time.Now()
-	time.Sleep(loadCase.duration)
-	lg.Stop()
-	testEndTime := time.Now()
+	testRunDuration := time.NewTimer(loadCase.duration)
+	defer testRunDuration.Stop()
+
+	// prepare data slices
+	dataSlices := make(chan dataSlice)
```
SaveTheRbtz, line 178: you need to buffer this channel, otherwise `recordTransactionData` will stuck on first send.

integration/benchmark/cmd/manual/main.go
```diff
@@ -138,17 +139,26 @@ func main() {
 			lg.Start()
 		}
 
+		var testRunDuration *timer.Timer
```
SaveTheRbtz, line 142: you can consider just using `time.After` instead of a ticker.

integration/benchmark/cmd/manual/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"flag"
+	"github.com/desertbit/timer"
```
SaveTheRbtz, line 6: please try remove this decency if possible.

AndrewM-SDET:
With the most recent commit we have full functionality for pulling TPS data from the loader itself. I have run several tests on my local machine that show data in expected ranges. We are good to officially open this PR and it is ready for formal review.

integration/benchmark/contLoadGenerator.go
```diff
@@ -82,7 +83,7 @@ type LoadParams struct {
 
 // New returns a new load generator
 func New(
-	// TODO(rbtz): use context
+// TODO(rbtz): use context
```
pattyshack, line 86: need to run the go format?

integration/benchmark/contLoadGenerator.go
```diff
@@ -286,6 +288,11 @@ func (lg *ContLoadGenerator) Stop() {
 	lg.workerStatsTracker.StopPrinting()
 	lg.log.Debug().Msg("stopping follower")
 	lg.follower.Stop()
+	close(lg.stoppedChannel)
```
pattyshack, line 291: do we need to worry about double close?

integration/benchmark/cmd/manual/main.go
```diff
@@ -138,17 +138,26 @@ func main() {
...
 		}
 
-		// if the duration is 0, we run this case forever
+		var runDuration time.Duration
+		// if the duration is 0, we run this case with a 24-hour timeout
 		if c.duration.Nanoseconds() == 0 {
-			for {
-				time.Sleep(time.Minute)
-			}
+			runDuration, _ = time.ParseDuration("24h")
```
pattyshack, line 144: just create a constant duration instead of parsing it from a string

integration/benchmark/cmd/ci/main.go
```diff
@@ -163,60 +170,89 @@ func main() {
...
-			OsVersion:           osVersion,
-			StartTime:           currentTime,
-			EndTime:             currentTime.Add(sliceDuration),
-			InputTps:            inputTps,
-			OutputTps:           outputTps,
-			ProStartTransaction: proStart,
-			ProEndTransaction:   proEnd}
-
-		slices = append(slices, slice)
+	for len(slicesChannel) != 0 {
```
pattyshack, line 252: I think you can just iterate over the channel

```
for slice := slicesChannel {
  slices = append(slices, slice)
}
```

integration/benchmark/contLoadGenerator.go
pattyshack: go formatter will probably complain if you add the space back.  maybe move this comment above `func New(` to make the indentation less jarring

integration/benchmark/cmd/ci/main.go
AndrewM-SDET: works with `for slice := range slicesChannel` 

will update

integration/benchmark/contLoadGenerator.go
AndrewM-SDET: Adding a check to the beginning of stop to make sure we are safely stopping. We would have to make changes to Start() if we want a single load generator to be able to be stopped and started multiple times, right now the load generators are one time use (one init, one start, one stop, and then if we need it again we need a whole new load generator).  Not sure if there is a use case for using the same load generator multiple times but worth considering for the future.

integration/benchmark/cmd/manual/main.go
```diff
@@ -138,17 +142,26 @@ func main() {
 			lg.Start()
 		}
 
-		// if the duration is 0, we run this case forever
+		var runDuration time.Duration
```
SaveTheRbtz, line 145: to avoid the need to have magical constants you can initialize the channel to something that never closes, and re-assign it only if duration is positive:
```suggestion
		waitC := make(<-chan time.Time)
		if c.duration != 0 {
			waitC = time.After(c.duration)
		}
```

integration/benchmark/cmd/manual/main.go
```diff
@@ -138,17 +142,26 @@ func main() {
...
+		} else {
+			runDuration = c.duration
 		}
 
-		time.Sleep(c.duration)
-
-		if lg != nil {
-			lg.Stop()
+		select {
+		case <-time.After(runDuration):
```
SaveTheRbtz, line 154:
```suggestion
		case <-waitC:
```

integration/benchmark/cmd/manual/main.go
```diff
@@ -29,6 +29,10 @@ type LoadCase struct {
 	duration time.Duration
 }
 
+const (
+	MaxRuntime = time.Hour * 24
+)
+
```
SaveTheRbtz, line 35:
```suggestion
```

.gitignore
SaveTheRbtz, line 18:
```suggestion
```

integration/benchmark/cmd/manual/main.go
AndrewM-SDET: I wasn't sure if it was an anti-pattern to make a channel that never closes. Thanks for the info, this is a cool technique.

SaveTheRbtz (approved):

pattyshack (approved):

AndrewM-SDET:
bors merge

PR #3168, opened 2022-09-02
Stop exposing Context() as part of environment api

pattyshack:
It creates a circular dependency which prevents us from moving the Environment
interface into the environment package.

This also removed AccountFreezeEnabled mock from tests (dead code).

NOTE: We will also need to stop exposing VM().  This will be done as part of
the ReusableCadenceRuntime changes.


zhangchiqing (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

pattyshack:
bors merge

PR #3170, opened 2022-09-02
Refactor value store into environment/

pattyshack:


fvm/environment/value_store.go
```diff
@@ -0,0 +1,110 @@
...
+			attribute.String("key", string(key)),
+		)
+	}
+	defer span.End()
+
+	err := store.meter.MeterComputation(
+		ComputationKindSetValue,
+		uint(len(value)))
+	if err != nil {
+		return fmt.Errorf("set value failed: %w", err)
```
Tonix517, line 76: nit: i know it is not the change from this PR - the msg here can be more specific for meter computation, and ideally we can also log the key which failed. same comments for `GetValue` above.

Tonix517 (approved):
LGTM with minor nit to original code

zhangchiqing (approved):

fvm/environment/value_store.go
pattyshack: i believe the meter error msg itself contains more info about which computation kind failed.  knowing the key in this case doesn't help too much since the user don't really have control over how cadence execute the code; the only thing the user can do is pay for more gas

pattyshack:
bors merge

pattyshack:
bors merge

SaveTheRbtz:
bors merge

Issue #2813, opened 2022-07-17
Allow emitEvent in script environment


bluesign:
## Problem Definition

Currently script environment does not allow emit event. I understand it is kind of pointless to emit the events but I think it is too harsh to panic. 

Context: I am coding a Safe Transaction Guard project ( using getAuthAccount and some cadence rewrite to basically detect changes in account state ) 

## Proposed Solution

It would be nice to allow emit to run as a no-op instead of panic
https://github.com/onflow/flow-go/blob/9652b94c903adc3839f0c11c31acd4c302aa5cef/fvm/scriptEnv.go#L534

cc: @janezpodhostnik

janezpodhostnik:
Without this you are prevented from calling contract functions which emit events, so I totally agree.

Is there some way we can make sure that the developer knows that events are ignored in scripts? Is documentation around this enough?

bluesign:
Oh this is very hard question to answer to be honest, but in general script environment doesn't mutate any state, but for sure it can be explained in more detail.

bluesign:
Is there any chance to implement this before next spork ?

PR #3183, opened 2022-09-06
[FVM] Events in scripts do nothing instead of causing errors

janezpodhostnik:
ref: https://github.com/onflow/flow-go/issues/2813

# Important Note

This is not a breaking change. Scripts can now call emit event, but nothing will happen instead of causing an error. This allows for scripts to call contract functions that would normally emit events.

### DIFF
--- a/fvm/environment/event_emitter.go
+++ b/fvm/environment/event_emitter.go
@@ -29,11 +29,11 @@ type EventEmitter interface {
 var _ EventEmitter = NoEventEmitter{}
 
 // NoEventEmitter is usually used in the environment for script execution,
-// where no event should be emitted.
+// where emitting an event does nothing.
 type NoEventEmitter struct{}
 
 func (NoEventEmitter) EmitEvent(event cadence.Event) error {
-	return errors.NewOperationNotSupportedError("EmitEvent")
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


fvm/environment/event_emitter.go
```diff
@@ -29,11 +29,11 @@ type EventEmitter interface {
 var _ EventEmitter = NoEventEmitter{}
 
 // NoEventEmitter is usually used in the environment for script execution,
-// where no event should be emitted.
+// where emitting an event does nothing.
 type NoEventEmitter struct{}
 
 func (NoEventEmitter) EmitEvent(event cadence.Event) error {
-	return errors.NewOperationNotSupportedError("EmitEvent")
+	return nil
```
Tonix517, line 36: is it useful if we log the event here?

Tonix517 (approved):
LGTM except a minor question

fvm/environment/event_emitter.go
janezpodhostnik: I was considering it, but there might be so many events happening, that it might completely fill the logs. Even just doing it in debug might be to much.

pattyshack (approved):

m4ksio (approved):

janezpodhostnik:
bors merge

PR #3184, opened 2022-09-06
Create one cadence runtime per ReusableCadenceRuntime object

pattyshack:
Since cadence runtime is not reentrant safe, each parallel runtime invocation
must have its own runtime.

### DIFF
--- a/fvm/env.go
+++ b/fvm/env.go
@@ -23,7 +23,6 @@ import (
 // Environment accepts a context and a virtual machine instance and provides
 // cadence runtime interface methods to the runtime.
 type Environment interface {
-	VM() *VirtualMachine
 	runtime.Interface
 
 	Chain() flow.Chain
@@ -143,10 +142,6 @@ func (env *commonEnv) Chain() flow.Chain {
 
 func (env *commonEnv) LimitAccountStorage() bool {
 	return env.ctx.LimitAccountStorage
-}
-
-func (env *commonEnv) VM() *VirtualMachine {
-	return env.vm
 }
 
 func (env *commonEnv) BorrowCadenceRuntime() *ReusableCadenceRuntime {
--- a/fvm/fvm.go
+++ b/fvm/fvm.go
@@ -27,11 +27,21 @@ func NewInterpreterRuntime(config runtime.Config) runtime.Runtime {
 
 // A VirtualMachine augments the Cadence runtime with Flow host functionality.
 type VirtualMachine struct {
-	// TODO(patrick): move this into ReusableCadenceRuntime
-	Runtime runtime.Runtime
+	Runtime runtime.Runtime // DEPRECATED.  DO NOT USE.
 }
 
-// NewVirtualMachine creates a new virtual machine instance with the provided runtime.
+func NewVM() *VirtualMachine {
+	return &VirtualMachine{}
+}
+
+// DEPRECATED.  DO NOT USE.
+//
+// TODO(patrick): remove after emulator is updated.
+//
+// Emulator is a special snowflake which prevents fvm from every changing its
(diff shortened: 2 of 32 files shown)


fvm/env.go
```diff
@@ -109,10 +108,6 @@ func (env *commonEnv) LimitAccountStorage() bool {
 	return env.ctx.LimitAccountStorage
 }
 
-func (env *commonEnv) VM() *VirtualMachine {
-	return env.vm
-}
-
 func (env *commonEnv) BorrowCadenceRuntime() *ReusableCadenceRuntime {
 	return env.ctx.ReusableCadenceRuntimePool.Borrow(env.fullEnv)
```
Tonix517, line 112: maybe we can emit some metrics in `Borrow` and `Return` func for monitoring purposes on Grafana? sth as simple as current active borrowed Cadence runtime etc. WDYT?

fvm/reusable_cadence_runtime.go
```diff
@@ -88,40 +83,151 @@ func NewReusableCadenceRuntime() *ReusableCadenceRuntime {
...
+	return runtime.NewInterpreterRuntime(pool.config)
 }
 
 func (pool ReusableCadenceRuntimePool) Borrow(
-	freezer AccountFreezer,
+	fvmEnv Environment,
 ) *ReusableCadenceRuntime {
 	var reusable *ReusableCadenceRuntime
 	select {
 	case reusable = <-pool.pool:
```
Tonix517, line 217: are we restricting the size of the pool?

fvm/env.go
pattyshack: I'm planning to refactor the borrowing to only happen once per transaction/script.  Instead of passing in the vm to Run (https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L18), we will pass in the ReusableCadenceRuntime.  We won't need the extra metric in that case.

fvm/reusable_cadence_runtime.go
pattyshack: the pool size is bounded, but we'll basically never hit that limit

Tonix517: how large is that limit, is that default or are we explicitly setting it? is it possible that this part of logic is being exploited and our memory grows more than the size we expect?

pattyshack: it's cap at 1000 right now.

Today, the real max number of item in the pool is 1 since everything is sequentially executed.  I imagine we would have knobs to control the maximum parallelism once we support parallel preprocessing/execution.  The number of reusable cadence runtime would be bounded by that when that happens.

Tonix517: good to know. that's what i was curious about.

Tonix517 (approved):

SaveTheRbtz (approved):

fvm/fvm.go
```diff
@@ -27,11 +27,21 @@ func NewInterpreterRuntime(config runtime.Config) runtime.Runtime {
 
 // A VirtualMachine augments the Cadence runtime with Flow host functionality.
 type VirtualMachine struct {
-	// TODO(patrick): move this into ReusableCadenceRuntime
-	Runtime runtime.Runtime
+	Runtime runtime.Runtime // DEPRECATED.  DO NOT USE.
 }
 
-// NewVirtualMachine creates a new virtual machine instance with the provided runtime.
+func NewVM() *VirtualMachine {
```
pattyshack, line 33: NOTE: I had to create this second function because of our ridiculous integration test setup.  We basically need 2-3 sporks in order to deprecate this function

pattyshack:
bors merge

PR #3188, opened 2022-09-06
Stop hiding BlockPrograms behind the old Programs interface.

pattyshack:
Note: We still need to preserve some of the old Programs' api until we refactor the lru cache in the block computer.


fvm/executionParameters.go
```diff
@@ -34,7 +36,23 @@ func getEnvironmentMeterParameters(
...
-	env := NewScriptEnvironment(context.Background(), ctx, vm, sth, programs)
+	txnProgs, err := blockProgs.NewSnapshotReadTransactionPrograms(
+		snapshotTime,
+		executionTime)
+	if err != nil {
+		return params, err
+	}
+
+	defer func() {
+		commitErr := txnProgs.Commit()
```
Tonix517, line 47: q: do we also need the `ShouldCommitPrograms()` check here?

fvm/fvm.go
```diff
@@ -88,7 +125,23 @@ func (vm *VirtualMachine) GetAccount(ctx Context, address flow.Address, v state.
...
-	account, err := getAccount(vm, ctx, stTxn, programs, address)
+	txnProgs, err := blockProgs.NewSnapshotReadTransactionPrograms(
+		programs.EndOfBlockExecutionTime,
+		programs.EndOfBlockExecutionTime)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		commitErr := txnProgs.Commit()
```
Tonix517, line 136: same question above

fvm/script.go
```diff
@@ -113,6 +113,22 @@ func (proc *ScriptProcedure) ShouldDisableMemoryAndInteractionLimits(
 	return ctx.DisableMemoryAndInteractionLimits
 }
 
+func (proc *ScriptProcedure) IsSnapshotReadTransaction() bool {
+	return true
+}
+
+func (proc *ScriptProcedure) ShouldCommitPrograms() bool {
+	return true
```
Tonix517, line 121: q: iirc Script is read-only? do we commit programs for Scripts for meters?

fvm/executionParameters.go
pattyshack: I've added ShouldCommitPrograms to deal with bootstrapping.  I'll rename `ShouldCommitPrograms` to `IsBootstrapping` to make it less confusing.

fvm/script.go
pattyshack: scripts are read-only and don't modify source code.

Given the same source code, the parsed program will be identical regardless whether it was used by a script or a transaction.  Hence it's safe to commit the program here.

We're committing the program to speed up future transaction/script which uses the same program

fvm/fvm.go
pattyshack: see above response

Tonix517 (approved):

SaveTheRbtz (approved):

fvm/programs/programs.go
```diff
@@ -1,105 +1,29 @@
...
 
-func (p *Programs) Cleanup(modifiedSets ModifiedSets) {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-
-	p.currentTxn.AddInvalidator(modifiedSets)
-
-	var err error
-	err = p.currentTxn.Commit()
+// DEPRECATED.  DO NOT USE
```
pattyshack, line 19: NOTE: I had to add this crap to deal with emulator circular dependency

pattyshack:
urgh ...  this needs major refactoring to deal with emulator circular dependency bullshit.  will close for now and tackle the refactoring piecemeal ...

PR #3192, opened 2022-09-07
[Execution] state migration clean up

ramtinms:
This PR remove the migration codes that has already been a applied on all networks.


cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -78,19 +77,7 @@ func extractExecutionState(
...
-			OutputDir: outputDir,
-		}
-		accountStatusMigration := mgr.NewAccountStatusMigration(log)
-		legacyControllerMigration := mgr.LegacyControllerMigration{Logger: log}
-
-		migrations = []ledger.Migration{
-			accountStatusMigration.Migrate,
-			legacyControllerMigration.Migrate,
-			storageUsedUpdateMigration.Migrate,
-			mgr.PruneMigration,
```
ramtinms, line 92: We don't need these two if there is no other migration applied

zhangchiqing (approved):

cmd/util/cmd/execution-state-extract/execution_state_extract.go
janezpodhostnik: Do we no longer need the `PruneMigration` If I remember correctly we always ran that.

janezpodhostnik (approved):

ramtinms:
bors merge

PR #3193, opened 2022-09-07
Refactor contract updater

pattyshack:
1. added ContractUpdater interface & NoContractUpdater for script
2. renamed (h *ContractHandler) to (updater *contractUpdater)
3. moved contract update methods from env to updater
4. moved ContractUpdateKeyList and ContractUpdateList into programs/
5. simplified Commit a bit

Note: we can't move the contract updater into environment/ yet since there are still some circular dependencies.

### DIFF
--- a/fvm/handler/contract.go
+++ b/fvm/handler/contract.go
@@ -12,20 +12,91 @@ import (
 	"github.com/onflow/flow-go/fvm/errors"
 	"github.com/onflow/flow-go/fvm/programs"
 	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module/trace"
 )
+
+type sortableContractUpdates struct {
+	keys    []programs.ContractUpdateKey
+	updates []programs.ContractUpdate
+}
+
+func (lists *sortableContractUpdates) Len() int {
+	return len(lists.keys)
+}
+
+func (lists *sortableContractUpdates) Swap(i, j int) {
+	lists.keys[i], lists.keys[j] = lists.keys[j], lists.keys[i]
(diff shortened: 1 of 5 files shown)


fvm/handler/contract.go
```diff
@@ -169,58 +266,59 @@ func (h *ContractHandler) Commit() ([]programs.ContractUpdateKey, error) {
...
+	keys := make(programs.ContractUpdateKeyList, 0, len(updater.draftUpdates))
+	updates := make(programs.ContractUpdateList, 0, len(updater.draftUpdates))
+	for key, update := range updater.draftUpdates {
+		keys = append(keys, key)
+		updates = append(updates, update)
 	}
 
 	sort.Sort(keys)
-	return keys
+	sort.Sort(updates)
```
SaveTheRbtz, line 292: You may consider using: [golang.org/x/exp/slices.SortFunc](https://pkg.go.dev/golang.org/x/exp/slices#SortFunc) since it has a couple of benefits:
* it only requires passing a sort function, which would allow us to dedup the `Less` method (and likely remove `Len` and `Swap`)
* it is generic

SaveTheRbtz (approved):

fvm/handler/contract.go
pattyshack: that's a reasonable ask.  i was just being lazy.  will make the recommended change

pattyshack: i've moved the sorter back into handler/contract.go since it's not safe to assume we can sort the key/update pairs together in general.

generic is overrated in this case.  your closure function will cost you whatever you've saved.

PR #3194, opened 2022-09-08
Convert programs.Programs into an optional argument for fvm.Run(V2)

pattyshack:
Prep work for fully integrating programs.BlockPrograms

Note: Because we're in a circular dependency hell (flow-go depends on emulator, which in turn depends on an older version of flow-go), we cannot simply change the Run method's signature.  Instead, we need to create a new RunV2 method to deal with backwards compatibility.  (Full migration to RunV2 / cleanup will requires at least 3 sporks ...)


Tonix517 (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #3200, opened 2022-09-09
Optimize GetEventsByBlockIDs

zhangchiqing:


engine/execution/rpc/engine.go
```diff
@@ -468,14 +468,14 @@ func (h *handler) eventResult(blockID flow.Identifier,
 	events := convert.EventsToMessages(flowEvents)
 
 	// lookup block
-	block, err := h.blocks.ByID(blockID)
+	header, err := h.headers.ByBlockID(blockID)
```
zhangchiqing, line 471: This is the main optimization.

pattyshack (approved):
I'm not familar with this part of the code base.  for my education purpose, could you explain why looking up the headers data structure is quicker than looking up the blocks data structure (and why we can't use make blocks equally quick?)

janezpodhostnik (approved):

zhangchiqing:
>why looking up the headers data structure is quicker than looking up the blocks data structure (and why we can't use make blocks equally quick?)

1. block body contains both the header and payload, when looking up the blocks, in additional to query header, it also [queries the payload](https://github.com/onflow/flow-go/blob/master/storage/badger/blocks.go#L50-L54)

2. block payload is much bigger than block header. 
Since payload contains guarantees, seals, receipts and results, which are stored under separate keys. [It would be way more database round-trip to query them](https://github.com/onflow/flow-go/blob/v0.27/storage/badger/payloads.go#L100), whereas looking up header is just a single db query to a single key.

janezpodhostnik:
Unit tests are failing. The failure is not from this PR, but from the previous one. It is also just about an error format changing.

I'm merging this anyway, so we can work on the deploy.

PR #3202, opened 2022-09-09
[FVM] Make failure more explicit

janezpodhostnik:
Small refactor to start making the difference between failure/non-failure cases more explicit.

### DIFF
--- a/fvm/transaction.go
+++ b/fvm/transaction.go
@@ -18,7 +18,10 @@ func Transaction(tx *flow.TransactionBody, txIndex uint32) *TransactionProcedure
 }
 
 type TransactionProcessor interface {
-	Process(*VirtualMachine, Context, *TransactionProcedure, *state.StateHolder, *programs.Programs) error
+	// Process processes a transaction.
+	// Process should expect that the transaction procedure might already contain an error.
+	// Process can only return a Failure in which case execution will halt.
+	Process(*VirtualMachine, Context, *TransactionProcedure, *state.StateHolder, *programs.Programs) errors.Failure
 }
 
 type TransactionProcedure struct {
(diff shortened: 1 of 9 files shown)


m4ksio (approved):
Good ideas!
I just think that maybe we should use more distinctive names - `error` vs `failure` is still rather generic and easy to mix-up.
Maybe `FVMFailure` vs `TransactionError` or something?

Tonix517 (approved):
LGTM and second @m4ksio 's suggestion.

fvm/transaction.go
```diff
@@ -58,19 +63,14 @@ func (proc *TransactionProcedure) Run(vm *VirtualMachine, ctx Context, st *state
 	}()
 
 	for _, p := range ctx.TransactionProcessors {
-		err := p.Process(vm, &ctx, proc, st, programs)
-		txErr, failure := errors.SplitErrorTypes(err)
+		failure := p.Process(vm, &ctx, proc, st, programs)
```
pattyshack, line 66: we should move the errors.SplitErrorType check here instead, instead of adding them separately to each of the processors

pattyshack: (I meant to request changes here)

pattyshack (changes requested):

fvm/transaction.go
zhangchiqing: i think what Janez did here makes sense. 

Processing a transaction should always succeeded without error. Any error would just cause the transaction to fail, and should not run into any fatal error. 

So if `Process` ever returns an error, then it must be a fatal error, in which case, we stop execution other txs, and exit right away. 

The `SplitErrorType` is to identify the error type and distinguish whether it's fatal or actually non-fatal. I think, a wrapper is needed to wrap the cadence Runtime, and distinguish fatal and non-fatal error from the error returned by cadence. With that wrapper, the caller would only need to handle a list of sentinel errors and fatal error.

zhangchiqing (approved):

fvm/transaction.go
pattyshack: Since Process is an internal api, it doesn't matter if the fatal error classification is handle inside or outside of Process, as long as it happens "somewhere" within TransactionProcedure.Run.

The main advantage of doing split error here instead of inside each Process method is less duplidated code, and in case we ever add a new processor to the list, we won't accidentally miss handling the error for the new processor.

pattyshack (approved):
(approve to unblock) i don't want to an ahole about this.  janez can decide whichever way he prefers.

fvm/transaction.go
janezpodhostnik: I was actually thinking about Patrik's comment

My next move was moving the errors.SplitErrorType check even deeper into each branch, but actually trying that, I didn't like what happened, because of the three separate code paths happening at once (success, normal error, failure).

Whats more is that normal transaction error path actually becomes more implicit wit the `proc.Err = err` moving into each function.

I still think this needs error handling here needs to change to become more explicit, but I'm not sure I'm heading down the right path.

I will convert this to a draft and try another approach I had in mind to see if that makes it better.

janezpodhostnik:
Since coded error this is much less needed, and would also need to be rewritten

PR #3205, opened 2022-09-09
Refactor ProgramsCache into ChainPrograms

pattyshack:
The main difference between the two is in how we create BlockPrograms for block execution.  The HasChanges optimization no longer makes sense since cloning BlockPrograms now deep copied the map in order to support fine grain eviction.


fvm/programs/chain_programs.go
```diff
@@ -0,0 +1,104 @@
...
+
+const DefaultProgramsCacheSize = 1000
+
+// ChainPrograms is a cache of BlockPrograms databases used for speeding up
+// cadence execution.
+//
+// Since programs are derived from external source, the BlockPrograms databases
+// need not be durable and can be recreated on the fly.
+type ChainPrograms struct {
+	mutex sync.Mutex
```
Tonix517, line 20: how about `RWMutex`?

fvm/programs/chain_programs.go
```diff
@@ -0,0 +1,104 @@
...
+	if ok {
+		return entry.(*Programs)
+	}
+
+	return nil
+}
+
+func (chain *ChainPrograms) NewBlockProgramsForScript(
+	currentBlockId flow.Identifier,
+	parentBlockId flow.Identifier,
```
Tonix517, line 96: nit: i assume the invariant here is that `parentBlockId` should be actually the id of `currentBlockId` 's parent node. But if we expose these params, there's a slight chance that caller may mistakenly call it with this invariant broken. not blocking to me, just sth to think about.

fvm/programs/chain_programs.go
pattyshack: It's not safe to use RWMutex in this case since lru will modify the data structure on Get

Tonix517: nice catch.

Tonix517 (approved):

fvm/programs/chain_programs.go
```diff
@@ -0,0 +1,85 @@
...
+) *Programs {
+	chain.mutex.Lock()
+	defer chain.mutex.Unlock()
+
+	currentEntry, ok := chain.lru.Get(currentBlockId)
+	if ok {
+		return currentEntry.(*Programs)
+	}
+
```
janezpodhostnik, line 63: this block could be a method `(chain *ChainPrograms) get(blockId flow.Identifier) *Programs`.

janezpodhostnik (approved):
LGTM.

1 minor comment.

fvm/programs/chain_programs.go
pattyshack: my bad here.  we can just use Get here directly.  I've made some minor changes and forgot to clean this up

pattyshack:
bors merge

Issue #3045, opened 2022-08-22
Stop/Crash EN at given height


m4ksio:
We need to have an ability to stop or crash EN at given height.
Initially, this is to be triggered by an admin command.
This command can specify if the node is to be stopped (or rather stop executing blocks) or crash - it depends on the operator preference, really.

Stop/Crash happens at given height, but there is extra problem associated with this.
As in this step we only support version switch, we must be certain there won't be any blocks to process at lower height. Since EN processed blocks as soon as possible, we must only stop when the height becomes finalized. For this, the node must stay online, accept blocks up until the height is finalized but not execute any blocks above the threshold height.

PR #3213, opened 2022-09-12
#3045 stop EN and VN at height

m4ksio:
This PR adds ability to issue admin command to EN to stop executing blocks above certain height, and optionally crash.
Also, VN now have additional command line parameter for the same procedure

Closes https://github.com/onflow/flow-go/issues/3045


admin/commands/execution/stop_at_height.go
```diff
@@ -0,0 +1,82 @@
...
+type StopAtHeightReq struct {
+	height uint64
+	crash  bool
+}
+
+func (s *StopAtHeightCommand) Handler(ctx context.Context, req *admin.CommandRequest) (interface{}, error) {
+	sah := req.ValidatorData.(StopAtHeightReq)
+
+	oldHeight := s.height.Swap(sah.height)
+	oldCrash := s.crash.Swap(sah.crash)
```
Tonix517, line 39: wondering if these two lines should be in the same critical section, if we care about atomicity here? like if A sends "1, true", then B sends "2, false", it may still happen that the received value is "1, false" etc. but if we don't expect concurrent command issuing, then it is fine doing so here.

admin/commands/execution/stop_at_height.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if !ok {
+		return errInvalidHeightValue
+	}
+
+	height, ok := result.(int)
+	if !ok || height <= 0 {
+		return errInvalidHeightValue
+	}
+	sahr.height = uint64(height)
```
Tonix517, line 65: nit: using generics may shorten this block, but no strong preference here.

engine/execution/ingestion/engine.go
```diff
@@ -424,6 +431,13 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
 		return
 	}
 
+	// skips blocks at or above requested stop height
+	stopHeight := e.stopAtHeight.Load()
+	if stopHeight > 0 && b.Height >= stopHeight {
```
Tonix517, line 436: q: does this equal to checking `pauseExecution`? if so, should it precede the previous "pauseExecution"?

engine/execution/ingestion/engine.go
```diff
@@ -440,6 +454,31 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
...
+	if stopHeight == 0 {
+		return
+	}
+
+	// once finalization reached stop height we can be sure no other fork will be valid at this height,
+	// so it is safe to stop or crash now
+	if h.Height >= stopHeight {
+
+		if e.stopAtHeightCrash.Load() {
+			e.log.Fatal().Msgf("Crashing as finalization reached requested stop height %d", stopHeight)
```
Tonix517, line 474: do we need to complete some finalization (like close\flush Badger etc.) before crashing here, to make sure data can still be intact, or is it really a don't care?

admin/commands/execution/stop_at_height.go
m4ksio: Its a good question, I considered it, but I doubt we need to care about race condition here

admin/commands/execution/stop_at_height.go
m4ksio: Hmm, can you show an example? We parse unknown object here and want to handle types mismatch, so I'm not sure how generics can help, but I don't have much experience with them

engine/execution/ingestion/engine.go
m4ksio: checking pauseExecution is faster so left it as a first check

engine/execution/ingestion/engine.go
m4ksio: Yes, we need finalization.
The thing is - we need to be sure that we executed all blocks up to height X, because we plan to use next version of EN/VN after restart, so it won't be possible to execute blocks below this height.
And we never know if there can be fork coming at any given height, unless this height is finalized.
Thats why we simply don't execute any blocks at or above X, but wait for X-1 to be finalized so we can be sure there are no possible forks.

admin/commands/execution/stop_at_height.go
Tonix517: sth in my mind like this:

```
func CheckAndGetValue[T any](input map[string]interface{}, key string) (*T, error) {
	result, ok := input[key]
	if !ok {
		return nil, errors.New("no such key in map")
	}

	retVal, ok := result.(T)
	if !ok {
		fmt.Println("not able to convert")
		return nil, errors.New("test")
	}

	return &retVal, nil
}
```

just to exchange some thoughts, not blocking for me

engine/execution/ingestion/engine.go
Tonix517: so do we still see the warning msg below when `stopHeight > 0 && b.Height >= stopHeight` is true?

admin/commands/execution/stop_at_height.go
m4ksio: Hmm, I experimented a bit with that, but it didn't increase clarity that much (for example we do check if height > 0, which mean with need either extra check anyway - and constructing another error string.
We should either use some proper marshalling/validating framework for mapping whole objects, but that's far outside the scope.
But thanks for the suggestion!

engine/execution/ingestion/engine.go
m4ksio: yes, until the stopHeight gets finalized

admin/commands/execution/stop_at_height.go
```diff
@@ -0,0 +1,82 @@
...
+	errInvalidCrashValue := fmt.Errorf("invalid value for \"crash\": expected a boolean, but got: %v", result)
+	if !ok {
+		return errInvalidCrashValue
+	}
+	crash, ok := result.(bool)
+	if !ok {
+		return errInvalidCrashValue
+	}
+
+	sahr.crash = crash
```
zhangchiqing, line 77:
```suggestion
	sahr :=  StopAtHeightReq{ height: uint64(height), crash: crash   }
```

engine/execution/ingestion/engine.go
zhangchiqing: A block get finalized doesn't mean it has been executed. This could happen if an EN is falling behind.
We need to check whether the block at height `stopAtHeight - 1` has been executed, and only crash if it has been executed. 
If it has not executed yet, we should not change `e.pauseExecution = true`, but wait until the last block (`stopAtHeight - 1`) has been executed.

zhangchiqing (changes requested):
Nice work! The admin tool changes look good. 

However, there is one case I think need to handle differently.

admin/commands/execution/stop_at_height.go
m4ksio: I don't think that matters in this case, but a good habit!

engine/execution/ingestion/engine.go
```diff
@@ -659,6 +709,14 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
 	if e.executionDataPruner != nil {
 		e.executionDataPruner.NotifyFulfilledHeight(executableBlock.Height())
 	}
+
+	if e.stopAfterExecuting == executableBlock.ID() {
+		// double check. Even if requested stop height has been changed multiple times,
+		// as long as it matches this block we are safe to terminate
```
zhangchiqing, line 715: This is for the case when stop height is changed before the old stop height has been reached.

This leads to think what happens when stop height is changed *after* the old stop height has been reached. 

If EN has stopped at height 100, and then we change the stop height to be 200, would it resume the execution from 100 to 199? It would be complex to get that to work. 

So it seems changing stop height only makes sense if the old stop height has _not_ been reached yet.

If so, how to stop EN from changing stop height if the stop height has been reached?

engine/execution/ingestion/engine.go
```diff
@@ -440,6 +453,43 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
...
+		if err != nil {
+			e.log.Error().Err(err).Str("block_id", h.ID().String()).Msg("failed to check if the block has been executed")
+			return
+		}
+
+		if executed {
+			e.stopExecution(stopCrash, stopHeight)
+		} else {
+			e.stopAfterExecuting = h.ParentID
+		}
```
zhangchiqing, line 488: I'm little concerned about the race condition here. 

`stopAfterExecuting` is checked by block execution. However block execution happens concurrently with block finalization, therefore there will be race condition here. 

Consider this case:

Let's say stop height is 100.
The two events: block 100 is finalized and block 99 is executed happen at the same time. 
Here, before `e.stopAfterExecuting = h.ParentID` in `BlockFinalized` is executed, the check `if e.stopAfterExecuting == executableBlock.ID() {` in `executeBlock ` still return `false`, so it won't crash the node. However, it just missed the last change to crash the EN. After `e.stopAfterExecuting = h.ParentID` is executed, the `e.stopAfterExecuting` is now block 99's ID, but since 99 has been executed, the check `if e.stopAfterExecuting == executableBlock.ID() ` will never return true anymore. EN will never crash.

Right now `stopHeight`, `stopEnabled`, `stopAfterExecuting`, `stopCrash` are read and written by different logic concurrently, I think we might need some synchronization here.

ramtinms (commented):
One concern I have is if there is a fork and we stop after executing the first block with the given height but the other branch that left unexecuted is actually the one that has to be executed.

engine/execution/ingestion/engine.go
m4ksio: Hmmm, ok, so the case when 99 has already been executed when 100 is finalized is covered in line 478.
The case where 99 has not finished execution before 100 is finalized, is covered
100 will never be executed, because we skip based on height.

So the problem is when finalization of 100 happens precisely before execution of 99 finishes but after it has started executing.

I think best approach would be to simply make sure both functions don't run at the same time - do you think simply using mutex to sync `OnBlockFinalized`  and final part of `OnBlockExecuted` when then check happens? 
Since its once per block and finalization handler is rather small I doubt there would be a lot of contention. Agree?

Regarding synchronization - Yes, I was considering that, but `stopEnabled`, `stopHeight`, `stopCrash` are synchronized, and `stopAfterExecuting` and `pauseExecution` are write-once really, so I don't think there would be a race there

engine/execution/ingestion/engine.go
m4ksio: Another great comment, thanks Leo!

You are right that adding any kind of resume would be a lot of work, and can branch into many cases when node it not up to date.

I think best solution would be to disallow change of stop height after any related event happen - so if we skip execution or crash/pause - we should simply lock the change and refuse any change.

m4ksio:
> One concern I have is if there is a fork and we stop after executing the first block with the given height but the other branch that left unexecuted is actually the one that has to be executed.

Hmm, I think we handle this case by waiting for finalization - so we execute every block up to height N-1, but only crash/stop once the N-1 gets finalized and this finalized block is executed.

engine/execution/ingestion/engine.go
zhangchiqing: Yeah, your solution makes sense. 

I wonder what's the use case for pausing the execution when stop height has been reached. If we always just crash, then we eliminate that case, because after stop height has been reached, the node has crashed already, stop height can't be changed again.

m4ksio: > I wonder what's the use case for pausing the execution when stop height has been reached. If we always just crash, then we eliminate that case, because after stop height has been reached, the node has crashed already, stop height can't be changed again.

I assume it might be beneficial to keep node running as it serves queries, kind of similar to how we do after spork, when old EN runs this way

engine/execution/ingestion/engine.go
zhangchiqing: I think one simple solution to solve the race condition is to check twice. 

```
	// we want to stop at stopHeight, meaning block at stopHeight will not be executed.
	lastHeightToExecute := stopHeight - 1
	if h.Height == lastHeightToExecute {
	        
			executed, err := checkAndStopExecution(stopCrash, lastHeightToExecute)

		        // there is a race condition in checkAndStopExecution, after setting stopAfterExecuting,
		        // the block might just have been executed, meaning the previous check was stale. 
		        // we can avoid this race condition by checking again.

			checkExecutedAgain, err := checkAndStopExecution(stopCrash, lastHeightToExecute)
	}

```

m4ksio: Yes, this seems to be great fix.
I have added a test to confirm that!

m4ksio:
> Nice work! The admin tool changes look good.
> 
> However,[ there is one case](https://github.com/onflow/flow-go/pull/3213#discussion_r973491032) I think need to handle differently.

I think this case is now properly handled

zhangchiqing (commented):
Had a few more questions

engine/execution/ingestion/engine.go
```diff
@@ -713,9 +729,13 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
 	if e.stopAfterExecuting == executableBlock.ID() {
 		// double check. Even if requested stop height has been changed multiple times,
 		// as long as it matches this block we are safe to terminate
-		if set, height, crash := e.stopAtHeight.Get(); set && height == executableBlock.Height()-1 {
-			e.stopExecution(crash, height)
-		}
+		e.stopAtHeight.Try(func(stopHeight uint64, crash bool) bool {
+			if executableBlock.Height()-1 == stopHeight {
```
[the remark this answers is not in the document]
m4ksio: Nice catch! Given that integration tests checking for crash works fine I guess it doesn't test this very case,

engine/execution/ingestion/engine.go
```diff
@@ -713,9 +729,13 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
...
+				e.stopExecution(crash, stopHeight)
+				return true
+			}
+			return false
```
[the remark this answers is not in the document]
m4ksio: It should get updated once new height is finalized, I don't think we have a knowledge to change this here

engine/execution/ingestion/engine.go
m4ksio: https://github.com/onflow/flow-go/pull/3213/files#diff-952a99ab2586822131b6723c10c40e350f122df570de319a2b572829ebc34875R62

In StopAtHeight object, that why we have those `Try` functions - they provide synchronisation and set the flag is stopping process has started.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	s.set = true
+	s.height = height
+	s.crash = crash
+
+	return oldSet, oldHeight, oldCrash, nil
+}
+
+func (s *StopAtHeight) Unset() {
```
zhangchiqing, line 73: What's the use case of this method? It seems we can Unset even if `stopping` is true. Maybe we should check `s.stopping` here as well

cmd/execution_builder.go
```diff
@@ -158,6 +160,9 @@ func (builder *ExecutionNodeBuilder) LoadComponentsAndModules() {
 		AdminCommand("trigger-checkpoint", func(config *NodeConfig) commands.AdminCommand {
 			return executionCommands.NewTriggerCheckpointCommand(exeNode.toTriggerCheckpoint)
 		}).
+		AdminCommand("stop-at-height", func(config *NodeConfig) commands.AdminCommand {
+			return executionCommands.NewStopAtHeightCommand(exeNode.stopAtHeight)
+		}).
```
zhangchiqing, line 165: Do we need to add this to VN as well?

engine/execution/ingestion/engine.go
```diff
@@ -424,6 +429,19 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
 		return
 	}
 
+	// skips blocks at or above requested stop height
+	stopping := e.stopAtHeight.Try(func(stopHeight uint64, stopCrash bool) bool {
+		if b.Height >= stopHeight {
+			e.log.Warn().Msgf("Skipping execution of %s at height %d because stop has been requested at height %d", b.ID(), b.Height, stopHeight)
+			return true
+		}
```
ramtinms, line 437: I think this should be `>` instead of `>=` so while we wait for the finalization signal to stop the node, we could executed forks

m4ksio: Hmm, stop height is the height new version should take over, so we don't want to executed the block at this height. 
Might be a bit counterintuitive now, but in next step we plan to have a contract with versions table, in a 

200 v0.23
100 v0.22

which makes sense to denote first block to use a version

ramtinms (approved):
Thanks for adding the comments to the code, makes it very readable, the logic makes sense to me, I added a comment about part of the logic and another question here, 
if a node is stopped but gets restarted by any reason it seems it would forget the fact that has to stop at the given height (maybe we could persist it somewhere)  

Thanks for adding integration tests, I think is still valuable to test this code on Benchnet few times in case we are missing an unknown situation.

engine/verification/fetcher/engine.go
```diff
@@ -190,6 +194,13 @@ func (e *Engine) processAssignedChunk(chunk *flow.Chunk, result *flow.ExecutionR
 		return false, blockHeight, nil
 	}
 
+	// skip chunk if it verifies a block at or above stop height
+	if e.stopAtHeight > 0 && blockHeight >= e.stopAtHeight {
+		e.log.Warn().Msgf("Skipping chunk %s - height  %d at or above stop height requested (%d)", chunkID, blockHeight, e.stopAtHeight)
+		e.chunkConsumerNotifier.Notify(chunkLocatorID) // tells consumer that we are done with this chunk.
```
zhangchiqing, line 200: a bit concerned here, because if we notify consumer this chunk is done. Then after restart with a new version, this chunk will not be processed again. 

Technically, if all ENs stop at the same height, then no result above e.stopAtHeight will be received by VN. 

However, if some EN didn't set the stop height, and still send result to the VN, then VN will mark it has been processed, which create a state where the VN is assigned with a chunk, and has verified this chunk without sending approvals. And after VN is restarted, it will not process the chunks from the result that is received before the crash.

engine/execution/ingestion/stop_at_height.go
m4ksio: No real use, I will remove it

cmd/execution_builder.go
m4ksio: VN doesn't have admin commands. You can specify stop height by command line, since it restarts fast.

engine/verification/fetcher/engine.go
m4ksio: Hmmm, I might miss something, but I think this state is not persisted, so restart will clean it

zhangchiqing: This is actually persisted. 

`chunkConsumerNotify.Notify` will call `chunkconsumer.Worker.Notify`, which will call `consumer.NotifyJobIsDone`. 
And jobqueue's `NotifyJobIsDone`, might [call `SetProcessedIndex ` ](https://github.com/onflow/flow-go/blob/master/module/jobqueue/consumer.go#L239), which will [update the processed index in database](https://github.com/onflow/flow-go/blob/master/storage/badger/consumer_progress.go#L43), because VN uses a [consumer progress module to store processed index](https://github.com/onflow/flow-go/blob/master/cmd/verification_builder.go#L151) in database.

m4ksio: Hmmm, does it simply looks like the solution is to omit this call to `e.chunkConsumerNotifier.Notify(chunkLocatorID)`?

zhangchiqing: If we don't Notify the job is done, then this worker will never be recycled to work on a new job. VN has 5 workers in total, so if we got 5 chunks for block 100, and not notify they are done, then when chunk block 99 is received, we will have no worker to process it, and we will get stuck here. 

What we need is a way to only not notify the next chunk after block 99 is done, and notify other chunks after that chunk to be done .

This is a bit tricky, because the next chunk after block 99 is not always the first chunk in block 100, because block 100 might not have any chunk assigned to this VN.

AlexHentschel (commented):
Some misc suggestions regarding comments. Completely aesthetical, feel free to ignore.   
* In a comment, paragraph consistent punctuation and capitalization of sentence beginnings increase readability. Nothing important.
* I had a hard time understanding one of the comments. Suggested different wording. Maybe it helps.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if started {
+		s.stopping = true
+	}
+
+	return started
+}
+
+// Set sets new values and return old ones
+func (s *StopAtHeight) Set(height uint64, crash bool) (bool, uint64, bool, error) {
```
AlexHentschel, line 54: please document error returns -> [guidelines](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/stop_at_height.go#L54)

admin/commands/execution/stop_at_height.go
```diff
@@ -0,0 +1,81 @@
...
+		stopAtHeight: sah,
+	}
+}
+
+type StopAtHeightReq struct {
+	height uint64
+	crash  bool
+}
+
+func (s *StopAtHeightCommand) Handler(ctx context.Context, req *admin.CommandRequest) (interface{}, error) {
```
AlexHentschel, line 33: once you have the error returns of https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/stop_at_height.go#L54 documented, you can just carry the doc over here.

engine/verification/fetcher/engine.go
zhangchiqing: I think we can keep your change here as is. 

Instead, we could change the assigner engine to not process height above stopHeight. 

Assigner engine process each finalized block, and check whether there is chunk assigned to me. It [stores the finalized height as processed index](https://github.com/onflow/flow-go/blob/master/cmd/verification_builder.go#L293). In other words, we can learn what's the last processed height from the processed index. 

Since assigner process finalized block height by height in a single thread manner, if assigner engine is processing height 100, it means it must have finish processing all height below 100. In other words, if assigner is processing stopHeight, it means it have processed all height below that. If we don't notify 100 is done, it is ok, because all heights below 100 have been processed. 

But when to crash the VN? 
We want to stop VN when all chunks below height 100 that are assigned to this VN have been processed (verified). 
It is not straightforward to know because the VN might not get assigned with a chunk for every height, there might be no chunk assigned for block 99. 

If we know what's the chunk locator for the last chunk I got assigned before block 100, then we could just wait until the processed index reaches the last chunk to stop VN. 

One solution would be to let assigner engine tell fetcher engine what's the last chunk, and fetcher engine will stop when the processed index has reached the last chunk.

zhangchiqing (approved):
The change for stopping EN at height look good to me now. However, to stop VN at a height might require more work. 

For now, I approve this PR for the EN changes, and we can work on VN in a separate PR.

admin/commands/execution/stop_at_height.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info().Msgf("admintool: EN will stop at height %d and crash: %t, previous values: %d %t", sah.height, sah.crash, oldHeight, oldCrash)
+
+	return "ok", nil
+}
+
+func (s *StopAtHeightCommand) Validator(req *admin.CommandRequest) error {
```
AlexHentschel, line 47: A brief doc would be great. Even an admin command might discover an internal state corruption (like reading a value from the DB that should be present surprisingly errors). Therefore, the default assumption is also here that only error returns explicitly documented as such can be considered benign (-> [guidelines](https://github.com/onflow/flow-go/blob/master/CodingConventions.md#best-practice-guidelines))

AlexHentschel (commented):
## Documentation of exported methods including error returns

For all **exported methods**
*  We generally would like to  a brief high-level documentation as to what this specific implementation does. 
* If the method has an error return, it should be documented what errors are benign. (-> [coding conventions](https://github.com/onflow/flow-go/blob/master/CodingConventions.md))

For exported methods, a minimal documentation is mandatory. For unexported methods, it is explicitly encouraged. 

It would be great if you could add goDoc (including error returns please) to the following methods (none present so far):
* https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L33
* https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L47

While the following methods have some goDoc, the error returns are not documented:
* https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/stop_at_height.go#L54
* https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L47

Please check also other parts of your implementation. Contributing to reducing the existing technical debt is also very much appreciated: if you find methods with insufficient documentation (incl. insufficient error doc), please add goDoc if possible with a reasonable time investment. We rely on every team member to contribute reducing the tech debt.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if !s.set {
+		return false
+	}
+
+	started := f(s.height, s.crash)
+
+	if started {
+		s.stopping = true
+	}
```
AlexHentschel, line 48: The naming of `started` is very confusing to me. I think it would be much more intuitive, if this was called `commencedStop`
```suggestion
	commencedStop := f(s.height, s.crash)
	if commencedStop {
		s.stopping = true
	}
```

AlexHentschel (commented):
### Suggestion for Error handling

I am using [admin/commands/execution/stop_at_height.go](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go) to outline my suggested best practises for error handling:
* you already define an error and return it at more than one place: https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L54
   * I would suggest to move this out of the function and define it as a sentinel:
      ```golang
      var InvalidAdminParametersError = errors.New("invalid admin parameters")
      ```
       I like to keep sentinels quite general so they are easily re-usable
   * you can then replace https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L54-L58 with 
      ```golang
      if !ok {
        return fmt.Errorf("missing value for `height`: %w", InvalidAdminParametersError)
      }
      ```
   * and replace https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L60-L63 by 
      ```golang
      if !ok || height <= 0 {
        return fmt.Errorf("value for `height` parameter must be a positive integer, but got %v: %w", result, InvalidAdminParametersError)
      }
      ```

  This approach also allows you to customize the error messages a bit more (if you desire so)
* The goDoc for the [`Validator` function](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/admin/commands/execution/stop_at_height.go#L47) could then read
  ```golang
  // Validator validates the configuration for the `StopAtHeightCommand`.
  // The following sentinel errors are expected during normal operations:
  //  * `commands.ErrValidatorReqDataFormat` if `req` is not a key-value map
  //  * `InvalidAdminParametersError` if necessary parameters are missing or any of the specified values are invalid
  func (s *StopAtHeightCommand) Validator(req *admin.CommandRequest) error {
  ```

Note: this implementation would be the most consistent with the remaining code base (except for tech debt). I feel the additional implementation overhead is marginal. If you prefer to spend even less time, you could leave the error handling as is and write something like:
```golang
// Validator validates the configuration for the `StopAtHeightCommand`.
// This function errors if `req` is invalid or incomplete. All returned errors are 
// benign, as the function is side-effect free and failures are simply a no-op.  
func (s *StopAtHeightCommand) Validator(req *admin.CommandRequest) error {
```

engine/execution/ingestion/engine.go
```diff
@@ -440,6 +458,54 @@ func (e *Engine) BlockProcessable(b *flow.Header) {
 	}
 }
 
+func (e *Engine) checkAndStopExecution(header *flow.Header, stopHeight uint64, stopCrash bool) {
+	executed, err := state.IsBlockExecuted(e.unit.Ctx(), e.execState, header.ParentID)
+	if err != nil {
+		e.log.Error().Err(err).Str("block_id", header.ID().String()).Msg("failed to check if the block has been executed")
+		return
+	}
```
AlexHentschel, line 466: ### ❓❗ 
To me, this looks like a case of "log error and continue with best effort", which is an anti-pattern (-> [conventions](https://github.com/onflow/flow-go/blob/master/CodingConventions.md#anti-pattern)). In case it is actually possible to continue here and the error is _not_ and indicator of internal state corruption, please add a comment explaining why this is the case.

AlexHentschel (commented):
## technical debt in the [`ingestion.Engine`](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go)

While I certainly lack context as to how the [`ingestion.Engine`](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go) is designed, it would still be important that the code was written in such a way that a reviewer could understand. There is a component of technical debt here, but this PR adds complexity to the engine without addressing the existing technical debt (lack of documentation as to how the engine works) and it further adds to the technical debt by adding control variables whose usage is hardly documented and concurrent access is not clear. This is a case, where it would be important to at least incrementally address the existing technical debt. 

A detailed comment in the struct would already help a lot: https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L41-L42
because receiving and saving incoming blocks is certainly not the only thing this engine does. It also processes `BlockFinalized` events and `BlockProcessable` from the `protocol.State`. A description how  these events are processes and how they interact (with some design-level explanations) would really help a lot. A description not only helps other engineers to understand the design, but it also helps the engineers extending the engine to potentially discover design flaws in their own code, such as (potentially) insufficient concurrency handling, which I talk about below. 

## ❗ Concurrency [`ingestion.Engine`](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go)

I have really tried to understand the concurrency handling in the [`ingestion.Engine`](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go) but could not convince myself that this is safe. Spend at least 30min but eventually had to give up. 
* Engines are inherently concurrent. This engine processes `BlockFinalized` and 
`BlockProcessable` events, which are emitted by other components within the Execution node. 
   * When processing a `BlockFinalized` event, the engine calls into https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L491 which writes to the engine's `stopAfterExecuting` variable: https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L471 
   * The engine's `stopAfterExecuting` variable is read  in the `executeBlock` method https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L875-L877 without any concurrency protection ?! At least i don't understand how this could be potentially concurrency safe and there is no documentation
* Similarly, I have not managed to understand why access to [`pauseExecution`](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L71) is concurrency safe 
  * It is read in while processing a `BlockProcessable` notification:  https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L426-L430
  * the same flag is written while processing a `BlockFinalized` event https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L469 and during block execution https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L734

**Suggestion**

The logic for updating the stopping parameters, reacting to block finalization, and reacting to block execution is the core logic of this PR. Essentially, it is a little state machine, whose inputs (parameter update through admin command, notification that block was finalized, notification that block was executed) are concurrent. 
In my opinion, a major source of complexity and potential bugs is the fact that this state machine is distributed over the `ingestion.Engine` and the `ingestion.StopAtHeight` struct with different (or no) concurrency control and no atomic state update. 

The probability that there is still a bug or uncovered edge case in the current implementation is quite large imho. Therefore, I do not consider the code ready for production.

I would strongly suggest to revise this as follows
* `ingestion.StopAtHeight` holds the entire logic for pausing or crashing execution, including the logic around `stopAfterExecuting` and `pauseExecution` (potentially in a simplified manner
* `ingestion.StopAtHeight` implements a little state machine, with **atomic state transition** (this will eliminate the headaches and workarounds like [this](https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L491-L502)) 
* the `ingestion.Engine` interfaces with the `StopAtHeight` for each block it wants to execute:
   * `StopAtHeight`  has a method `ShouldProcessBlock(block flow.Header) bool` with this logic: https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L426-L443
   * `StopAtHeight`  has a method `BlockExecuted(block flow.Header)`, which contains the logic https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L729-L739
   * `StopAtHeight`  has a method `BlockFinalized(ctx context.Context, block flow.Header)` , with the logic https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L478-L506 You can probably significantly simplify this, once you execute this logic atomically within the `StopAtHeight`. If it is too complicated to inject an instance of `ReadOnlyExecutionState` at construction time of `StopAtHeight`, you could just add it as a function parameter: `BlockFinalized(ctx context.Context, block flow.Header, state ReadOnlyExecutionState)` 
   * Not sure if you would still need it, but if you do, `StopAtHeight` could also expose a method `IsExecutionPaused() bool` which is analogous to https://github.com/onflow/flow-go/blob/3c6f7bdfd479cd12b209e48011196ddccdde8704/engine/execution/ingestion/engine.go#L71

In this design, `StopAtHeight` implements the state machine for starting and stopping. It could even track the height of the most recently executed block and refuse admin commands to stop below this height. Either way, this would remove a major source of complexity from the ingestion engine. It makes the concurrent nature of  `StopAtHeight` very apparent and avoids pitfalls of missing concurrency control.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,176 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/engine/execution/state"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+type StopControl struct {
```
zhangchiqing, line 14: since the logic has been extracted into this module, can we add more test cases?

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,176 @@
...
+	return s.set, s.height, s.crash
+}
+
+func (s *StopControl) IsPaused() bool {
+	s.RLock()
+	defer s.RUnlock()
+	return s.paused
+}
+
+// Try runs function f with current values of height and crash if the values are set
```
zhangchiqing, line 49: is Try being used?

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/onflow/flow-go/engine/execution/state"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+type StopControl struct {
+	sync.RWMutex
+	height             uint64
+	crash              bool
+	commenced          bool // if stopping process has started. We disallow any changes then.
```
zhangchiqing, line 18: Can we get rid of this state by:

```
isCommenced := s.stopAfterExecuting != flow.ZeroID
```

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,176 @@
...
+
+		executed, err := state.IsBlockExecuted(ctx, execState, h.ParentID)
+		if err != nil {
+			// any error here would indicate unexpected storage error, so we crash the node
+			s.log.Fatal().Err(err).Str("block_id", h.ID().String()).Msg("failed to check if the block has been executed")
+			return
+		}
+
+		if executed {
+			s.stopExecution()
```
zhangchiqing, line 142: can we set stopAfterExecuting regardless?

engine/execution/ingestion/stop_at_height.go
m4ksio: Nope, this should be set to true if we skip any block, which include skipping blocks above the height, not necessary setting the final block ID.

m4ksio:
@AlexHentschel I have refactored the code according to your directions. 
While the logic should stay the same, I think it warrant a re-review.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,167 @@
...
+		} else {
+			s.stopAfterExecuting = h.ParentID
+			s.set = true
```
zhangchiqing, line 133: why updating the `set` here?

If was not set, then s.height should not have a value for `h.Height == s.height` to be true, right?

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,167 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	if !s.set {
+		return false
+	}
+
+	if s.paused {
+		return true
+	}
```
zhangchiqing, line 93: if `set == true and paused == true` then BlockProcessable would return false, meaning we should not skip the block, therefore should process it, however, isn't `paused == true` means we should not process it? It's a bit confusing here.

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// skips blocks at or above requested stop height
+	if b.Height >= s.height {
+		s.log.Warn().Msgf("Skipping execution of %s at height %d because stop has been requested at height %d", b.ID(), b.Height, s.height)
+		s.commenced = true
```
zhangchiqing, line 98: This leads me to think, what ensures that before setting the stop height, there isn't a block above the stop height has been executed?

If block 101 has been executed, and I set stop executing at height 100, then when switching version after restart, block 101 will be skipped as it's already executed by old version. 

We have a Highest executed height saved in database, should we use that?

engine/execution/ingestion/stop_at_height.go
```diff
@@ -0,0 +1,167 @@
...
+	paused             bool // if execution is paused
+	stopAfterExecuting flow.Identifier
+	log                zerolog.Logger
+}
+
+// NewStopControl creates new empty NewStopControl
+func NewStopControl(log zerolog.Logger, paused bool) *StopControl {
+	return &StopControl{
+		paused:    paused,
+		commenced: paused,
```
zhangchiqing, line 31: why `commenced` is initialized as paused?

engine/execution/ingestion/stop_at_height.go
m4ksio: we change internal state here (by setting stopAfterExecuting) so we mark it as set regardless

engine/execution/ingestion/stop_at_height.go
m4ksio: > if set == true and paused == true then BlockProcessable would return false,

It would return true, from line 92.
I added the docs clarifying that this function returns bool indicating is block should be skipped

engine/execution/ingestion/stop_at_height.go
zhangchiqing: This is a bit confusing. because set is actually `setStopHeightAndCrashed` as indicated in the comments, rather than `setStopAfterExecuting`. 

At this point `set` must have been true, if not.

engine/execution/ingestion/stop_at_height.go
m4ksio: if we start in paused state its the same state as we started the stopping process. 
This state if also checked while starting engine (it skips block upload retry if its paused IIRC) so its changed the state, hence we set commenced to true

engine/execution/ingestion/stop_at_height.go
m4ksio: Thats a good question. 
In the version contract we explicitly disallow changes to blocks within 1000 blocks from current one.
Maybe that would work here as well - so disallow setting within X blocks of latest executed (I think there can be some edge cases if we disallow on last executed height directly, forks, etc)

engine/execution/ingestion/stop_control_test.go
```diff
@@ -0,0 +1,73 @@
...
+		require.True(t, oldSet)
+
+		// block at stop height, it should be skipped
+		header = unittest.BlockHeaderFixture(unittest.WithHeaderHeight(37))
+		sah.BlockFinalized(context.TODO(), execState, header)
+
+		_, _, _, err = sah.SetStopHeight(2137, false)
+		require.Error(t, err)
+	})
+
```
zhangchiqing, line 72: Let's add more tests to cover different combination of state variables. 
One of the tests should call `BlockExecuted`, since there is state changes in that method.
```
- it won't stop if stop height was not set
- it will stop if stop height was reached first then the stop height get finalized
- it will stop if stop height was finalized first, then stop height was reached
- if stop height was finalized first, it will disallow changing stop height, and it will stop at original stop height
- if stop height was reached first, and stop height didn't get finalized, it should allow changing stop height, and it will stop at the new height.
- etc
```

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,167 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	if !s.set {
+		return false
+	}
+
+	if s.paused {
+		return true
+	}
```
zhangchiqing, line 93: we should handle paused first, if it's true, then we must skip. if is false, then we check set

```suggestion
	if s.paused {
		return true
	}
	
	if !s.set {
		return false
	}


```

m4ksio: if `set` is false, `pause` value doesn't matter (as it has not been explicitly set)

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,167 @@
...
+
+	oldSet := s.set
+	oldHeight := s.height
+	oldCrash := s.crash
+
+	if s.commenced {
+		return oldSet, oldHeight, oldCrash, fmt.Errorf("cannot update stop height, stopping already in progress for height %d with crash=%t", oldHeight, oldCrash)
+	}
+
+	s.set = true
```
zhangchiqing, line 61: please log every state changes, in case we need to debug, it's better to have all logs for how we arrive at a certain state. 

we could log all state variables, if state transition happens

engine/execution/ingestion/stop_control_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+// If stopping mechanism has caused any changes to execution flow (skipping execution of blocks)
+// we disallow setting new values
+func TestCannotSetNewValuesAfterStoppingCommenced(t *testing.T) {
+
+	t.Run("when processing block at stop height", func(t *testing.T) {
+		sah := NewStopControl(zerolog.Nop(), false)
```
zhangchiqing, line 20: useful to also have log in tests in case we need to debug
```suggestion
		sah := NewStopControl(unittest.Logger(), false)
```

engine/execution/ingestion/stop_control.go
m4ksio: The idea is that `set` determines if this whole module is even active. Maybe it would be better to rename this to `active` or something along those lines. 
Maybe it would avoid confusion if I turn this into a proper state machine - lifecycle like 
`off` -> `set` -> `skipped` -> `paused` -> `stopped`  should be cleaner (not sure if thats fully cycle here, just an example).

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,167 @@
...
+// StopControl is a specialized component used by ingestion.Engine to encapsulate
+// control of pausing/stopping blocks execution.
+type StopControl struct {
+	sync.RWMutex
+	height             uint64
+	crash              bool
+	commenced          bool // if stopping process has started. We disallow any changes then.
+	set                bool // whether stop at height has been set at all, it its envisioned most of the time it won't
+	paused             bool // if execution is paused
+	stopAfterExecuting flow.Identifier
```
zhangchiqing, line 23: I think we can get rid of the `paused` state variable. Here is why.

I think the minimum state set is:

```golang
// on startup, this is initialized as highest executed height
// after startup, this is updated by a Processable block with higher height 
// stop height can not be set below or equal to this height, otherwise we can't guarantee 
// no block above stopHeight has been executed. 
highestProcessableHeight uint64

// this is true means (height, crash) are set
stopHeightSet bool 

// the height above (or equal to) which won't be executed
height uint64 

// whether to crash when (height - 1) is finalized and executed. 
crash bool 

// indicate if any block above lastHeight has been seen. lastHeight is (stopHeight - 1).
// this is set to true when BlockProcessable is called with a block whose height is above lastHeight
// if this happens, stop height can not be changed, because we might have stopped some block
// from executing.
hasBlockAboveLastHeight bool 

// whether block at (height - 1) is finalized. lastHeight is (stopHeight - 1).
lastHeightFinalized bool 

// the ID for the finalized block at (height - 1), only updated when `lastHeightFinalized` is true.
lastHeightID flow.Identifier 

// indicate whether the last height (the block with lastHeightID) has been executed,
// note must be the finalized block at last height, not a different block at the same height.
lastHeightExecuted bool
```

Now with the above state variable, the most important value is the return value of `BlockProcessable`, which decides whether to execute that block. 

Is suggest this function to return true for "processable", instead of "skip". As it's easy to read:
```
// Notify the stop control about a processable block and return whether the block is processable.
func (s *StopControl) NotifyBlockProcessable(h *flow.Header) bool

// in ingestion engine
func (e *IngestionEngine) BlockProcessable(h *flow.Header) {
  processable := s.NotifyBlockProcessable(h) 
  if !processable {
    // queue the block to process 
  }
}
```

And NotifyBlockProcessable can be implemented as:
```
func (s *StopControl) NotifyBlockProcessable(h *flow.Header) bool {
  s.Lock()
  defer s.Unlock()

  // if paused already, then we won't execute any block
  if s.paused() {
    return false
  }
  
  // always update the highest height
  if h.Height > s.highestProcessableHeight {
    s.highestProcessableHeight = h.Height
  }

  // if stop height not set, any block are executable
  if !set {
    return true
  }

  lastHeight := s.height - 1
  if h.Height > lastHeight {
    s.hasBlockAboveLastHeight = true 
  }
}

func (s *StopControl) paused() bool {
  return s.lastHeightFinalized && s.lastHeightExecuted
}
```

zhangchiqing: Now you might wonder what if user want to pause the execution to begin with?

That's actually what I wanted to distinguish. 

The `pause` state means two things:
1. user wants to pause the execution to begin with
2. user wants to not crash after reaching stop height.

For 2, we can use the StopControl module without a pause method, rather than a state
For 1, we can use a AlwaysSkip module which implements the same interface as StopControl, but the NotifyBlockProcessable will always return false (not processable). 

This way, we get rid of the `pause` state from `StopControl` module.

m4ksio: I like the idea of tracking latest height, and preventing setting new one if thats below or close to it.
However, I think path with interfaces and separate modules for rather simple distinction is a bit of an overkill.

zhangchiqing: The pause flag is a like a main control above all other controls. 

If you don't like using interface, consider wrapping StopControl with an addition pause flag:
```
type StopController struct {
  pause bool
  stopControl *stopControl
}
```

zhangchiqing (changes requested):
Please refer to my [review comments and suggestion here. ](https://github.com/onflow/flow-go/pull/3213/files#r990563498)

Also I don't think the existing test cases are enough to cover all the combination of state changes. 

I've listed a few that haven't been covered, please consider adding more.

m4ksio:
> Please refer to my [review comments and suggestion here. ](https://github.com/onflow/flow-go/pull/3213/files#r990563498)
> 
> Also I don't think the existing test cases are enough to cover all the combination of state changes.
> 
> I've listed a few that haven't been covered, please consider adding more.

Heh, you were few minutes faster :)
I have implemented most of your suggestions, refactored the whole mechanism to behave like a state machine. Added more tests which should now cover all the test cases you mentioned.

engine/execution/ingestion/stop_at_height.go
m4ksio: I have added height tracking now, which should prevent this problem.

engine/execution/ingestion/stop_control.go
m4ksio: Good idea, added

engine/execution/ingestion/stop_control_test.go
m4ksio: 👍

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+func (s *StopControl) GetStopHeight() (uint64, bool) {
+	s.RLock()
+	defer s.RUnlock()
+
+	return s.height, s.crash
+}
+
+// BlockProcessable should be called when new block is processable.
+// It returns boolean indicating if the block should be processed.
+func (s *StopControl) BlockProcessable(b *flow.Header) bool {
```
AlexHentschel, line 128: At the moment, we are tracking the last executed block:
https://github.com/onflow/flow-go/blob/93680d1b7ad29944d05f85f3f36e85289cff2a0b/engine/execution/ingestion/stop_control.go#L25

But I am wondering if it wouldn't make sense to rather track the latest Block _scheduled_ for execution (`latestScheduledForExecution`). We would increase this variable here. 

The reasoning is that this is a completely asynchronous process. I think it makes sense to require for the admin command that only heights are specified that are not already scheduled for execution. Otherwise, we get into the complicated realm of "what happens if I set a height where the block execution is already in flight?"

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	if s.state == StopControlPaused {
+		return
+	}
+
+	if h.Height > s.lastExecutedHeight {
+		s.lastExecutedHeight = h.Height
+	}
```
AlexHentschel, line 197: would suggest to replace this by tracking the latest Block _scheduled_ for execution (`latestScheduledForExecution`) and moving it to method `BlockProcessable`

admin/errors.go
```diff
@@ -0,0 +1,34 @@
+package admin
+
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrValidatorReqDataFormat error = errors.New("wrong input format: expected JSON")
+
+type InvalidAdminParameterError struct {
```
AlexHentschel, line 10: for custom error types, we usually implement this convenience method: 
https://github.com/onflow/flow-go/blob/93680d1b7ad29944d05f85f3f36e85289cff2a0b/consensus/hotstuff/committees/leader/leader_selection.go#L28-L31

engine/verification/assigner/engine.go
```diff
@@ -165,6 +170,11 @@ func (e *Engine) ProcessFinalizedBlock(block *flow.Block) {
 // processFinalizedBlock indexes the execution receipts included in the block, performs chunk assignment on its result, and
 // processes the chunks assigned to this verification node by pushing them to the chunks consumer.
 func (e *Engine) processFinalizedBlock(ctx context.Context, block *flow.Block) {
+
+	if e.stopAtHeight > 0 && block.Header.Height == e.stopAtHeight {
+		e.stopAtBlockID = block.ID()
```
AlexHentschel, line 175: I think this method is called by an external thread. Hence you might need concurrency handling here 🤔 . [Atomic Value](https://pkg.go.dev/sync/atomic#Value) might work nicely here, since you have a default initialization value (`flow.ZeroID`)

AlexHentschel (approved):
Thanks really love this PR, specifically:
* nice error handling in [`StopAtHeightCommand`](https://github.com/onflow/flow-go/blob/93680d1b7ad29944d05f85f3f36e85289cff2a0b/admin/commands/execution/stop_at_height.go) and thanks for testing the error returns for correct type
* thank you very much for improving error handling in the other admin command implementations too 
   🎖️addressing tech debt 🎖️ 
* solid documentation
* Very clean state machine logic in [`ingestion. StopControl`](https://github.com/onflow/flow-go/blob/93680d1b7ad29944d05f85f3f36e85289cff2a0b/engine/execution/ingestion/stop_control.go#L25) 💡 great to follow.

The revisions on this PR really payed off. Nice work. 

Mostly stylistic comments. There are two questions w.r.t to concurrency handling ([here](https://github.com/onflow/flow-go/pull/3213#discussion_r992996828) and [here](https://github.com/onflow/flow-go/pull/3213#discussion_r993020899)), which might affect correctness.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+
+// BlockProcessable should be called when new block is processable.
+// It returns boolean indicating if the block should be processed.
+func (s *StopControl) BlockProcessable(b *flow.Header) bool {
+
+	s.Lock()
+	defer s.Unlock()
+
+	if s.state == StopControlOff {
+		return true
```
zhangchiqing, line 134: I think there is a unhappy path missed here, please refer to my comment for `highestProcessableHeight` and `NotifyBlockProcessable`. 

The idea of `highestProcessableHeight` is the same as @AlexHentschel 's idea for using `latestScheduledForExecution`. Let me call it `highestProcessableHeight` for now.

We need to always update the highest processable height unless it's paused, that's why I handle the Paused state first.
Also note, it's "highest" or "latest". The latest schedule could be a lower height than the highest. 

So there are a few test cases we need to add and pass:

1. if stop control has no stop height, then called BlockProcessable(20), and then set stop height 20, it should fail, because stop height 20 means the last block to execute is 19, but a 20 might have been processed. 
2. if stop control has no stop height, then called BlockProcessable(20), and then set stop height 21, it should succeed, because stop height 21 means the last block to execute is 20. 
3. if stop control has no stop height, then called BlockProcessable(20), and then called BlockProcessable(19), then set stop height 20 should fail, because we use highest processable height to guard, rather than latest.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+type StopControlState byte
+
+const (
+	// StopControlOff default state, envisioned to be used most of the time. Stopping module is simply off,
+	// blocks will be processed "as usual".
+	StopControlOff StopControlState = iota
+
+	// StopControlSet means stop height is set but not reached yet, and nothing related to stopping happened yet.
```
zhangchiqing, line 35: I found the word `reached` a bit unclear. 

Also please refer to my comments to add comments to the state variables, such as `height`, `crash`, etc

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+	// We can still go back to StopControlOff or progress to StopControlCommenced.
```
zhangchiqing, line 36: >We can still go back to StopControlOff

Is this implemented?

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+	StopControlCommenced
+
+	// StopControlPaused means EN has stopped processing blocks. It can happen by reaching set stop height, or
+	// if the node was started with a pause mode.
+	// It is a final state and cannot be changed
+	StopControlPaused
+)
+
+// NewStopControl creates new empty NewStopControl
+func NewStopControl(log zerolog.Logger, paused bool) *StopControl {
```
zhangchiqing, line 51: We need to initialize the highestProcessable with the HighestExecutedBlockHeight from badger database. So that we don't allow setting stop height lower or equal to that. 

Please refer to my comments before
```
// on startup, this is initialized as highest executed height
// after startup, this is updated by a Processable block with higher height 
// stop height can not be set below or equal to this height, otherwise we can't guarantee 
// no block above stopHeight has been executed. 
highestProcessableHeight uint64
```

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,221 @@
...
+	log.Debug().Msgf("created StopControl module with paused = %t", paused)
+	return &StopControl{
+		log:   log,
+		state: state,
+	}
+}
+
+// GetState returns current state of StopControl module
+func (s *StopControl) GetState() StopControlState {
+	return s.state
```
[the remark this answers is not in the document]
m4ksio: Yep, thanks!

engine/execution/ingestion/stop_control.go
m4ksio: Scheduled for execution is a bad approach in my mind.
Imagine EN which is far behind. Behind it has not executed blocks for the very same network it is a part of, for a significant number of blocks. 
It can schedule a large number of blocks for execution, sometime hours or days' worth of blocks. This would mean it's impossible to set a stop height to anything between current executed block and the actual latest block.

Additionally, I see no reason why we would prevent from stopping even one block from what's possible to properly stop, which is highest currently being executed+1.
I think practical solution would be to track highest block at the start of execution.

engine/execution/ingestion/stop_control.go
m4ksio: No, but that's topic for another PR which will add appropriate admin command, otherwise it will be moot operation

engine/execution/ingestion/stop_control.go
m4ksio: Let's keep discussion in one place https://github.com/onflow/flow-go/pull/3213#discussion_r993682312

engine/execution/ingestion/stop_control.go
m4ksio: Let's keep discussion in one place https://github.com/onflow/flow-go/pull/3213#discussion_r993682312

admin/errors.go
m4ksio: 👍

engine/verification/assigner/engine.go
m4ksio: 👍

engine/execution/ingestion/stop_control.go
m4ksio: Added the initialization.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+
+		if h.Height == s.height-1 {
+			s.stopExecution()
+		}
+	}
+}
+
+func (s *StopControl) stopExecution() {
+	if s.crash {
+		s.log.Fatal().Msgf("Crashing as finalization reached requested stop height %d", s.height)
```
zhangchiqing, line 217:
```suggestion
		s.log.Fatal().Msgf("Crashing as finalization reached requested stop height %d and the highest executed block is (%d - 1)", s.height, s.height)
```

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+	} else {
+		s.log.Debug().Int8("previous_state", int8(s.state)).Int8("new_state", int8(StopControlPaused)).Msg("StopControl state transition")
+		s.state = StopControlPaused
+		s.log.Warn().Msgf("Pausing execution as finalization reached requested stop height %d", s.height)
+	}
+}
+
+// executingBlockHeight should be called while execution of height starts, used for internal tracking of the minimum
+// possible value of height
+func (s *StopControl) executingBlockHeight(height uint64) {
```
zhangchiqing, line 227: Consider making this method public. 

If we allow caller to call the private method, it means they know the implementation detail, and nothing can prevent them from bypassing all the checks and calling `stopExecution` directly.

We could use interface to only expose certain methods to different callers.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+	if s.state == StopControlPaused {
+		return
+	}
+
+	if height > s.lastExecutingHeight {
```
zhangchiqing, line 232: Consider renaming to `highest`.
```suggestion
	// updating the highest executing height, which will be used to reject setting stop height that 
	// is too low. 
	if height > s.highestExecutingHeight {
```

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+
+		executed, err := state.IsBlockExecuted(ctx, execState, h.ParentID)
+		if err != nil {
+			// any error here would indicate unexpected storage error, so we crash the node
+			s.log.Fatal().Err(err).Str("block_id", h.ID().String()).Msg("failed to check if the block has been executed")
+			return
+		}
+
+		if executed {
+			s.stopExecution()
```
zhangchiqing, line 187: optional: wonder if would be useful to pass in the finalized block ID for it to be logged.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// StopControl is a specialized component used by ingestion.Engine to encapsulate
+// control of pausing/stopping blocks execution.
+// It is intended to work tightly with the Engine, not as a general mechanism or interface.
+// StopControl follows states described in StopState
+type StopControl struct {
+	sync.RWMutex
+	// desired stop height, the first value new version should be used, so this height WON'T
```
zhangchiqing, line 20: Reading this highlighted "WON'T", I wonder if it makes sense to add some sanity checks in `BlockExecuted` and `executingBlockHeight` and `blockExecuted` event to if stop height is set then we never execute blocks above or equal to that height?

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+		return
+	}
+
+	if s.stopAfterExecuting == h.ID() {
+		// double check. Even if requested stop height has been changed multiple times,
+		// as long as it matches this block we are safe to terminate
+
+		if h.Height == s.height-1 {
+			s.stopExecution()
+		}
```
zhangchiqing, line 211: I think the implementation is different from what is described here. 

Is the `else` case expected or unexpected? If is unexpected, better to crash. 

Either way, it's better to log whether the condition is true or not.

```suggestion
		} else {
		     s.Fatal()
		}
```

engine/execution/ingestion/engine.go
```diff
@@ -581,6 +586,8 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
 
 	startedAt := time.Now()
 
+	e.stopControl.executingBlockHeight(executableBlock.Block.Header.Height)
```
zhangchiqing, line 589: I got your idea, good point. We only need to track executing height, rather than processable height.

zhangchiqing (approved):
Great works! Thanks for making the state transition clear and adding tests 👍

engine/execution/ingestion/stop_control.go
m4ksio: This is intentional. `StopControl` is intended as internal part of ingestion engine, so we expect responsible behaviour from it. I have also made other methods private, and only keep the `SetHeight` and read-only methods as public, so we expose clean API outside.

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+	// +1 because we track last executing height, so +1 is the lowest possible block to stop
+	if height <= s.lastExecutingHeight+1 {
+		return oldHeight, oldCrash, fmt.Errorf("cannot update stop height, given height %d at or below last executed %d", height, s.lastExecutingHeight)
+	}
+
+	s.log.Info().
+		Uint64("height", height).Bool("crash", crash).
+		Uint64("old_height", oldHeight).Bool("old_crash", oldCrash).Msg("new stop height set")
+
+	s.log.Debug().Int8("previous_state", int8(s.state)).Int8("new_state", int8(StopControlSet)).Msg("StopControl state transition")
```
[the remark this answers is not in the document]
m4ksio: Yep

engine/execution/ingestion/stop_control.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	if s.state == StopControlPaused {
+		return false
+	}
+
+	// skips blocks at or above requested stop height
+	if b.Height >= s.height {
+		s.log.Warn().Msgf("Skipping execution of %s at height %d because stop has been requested at height %d", b.ID(), b.Height, s.height)
+		s.log.Debug().Int8("previous_state", int8(s.state)).Int8("new_state", int8(StopControlCommenced)).Msg("StopControl state transition")
```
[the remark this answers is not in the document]
m4ksio: I was initially skeptical because it would print for every block, but that will only happen for a few blocks until crash or pause

engine/execution/ingestion/stop_control.go
m4ksio: Yep, won't hurt

engine/execution/ingestion/stop_control.go
m4ksio: I think it hold true - what I meant was that if you, say, changed stop height from 100 to 200 and back to 100, it should still give the same finalized block. 
I wonder if the second check (height) is necessary. 
I will leave it and add logs in else, but if you think we should simply remove it, I'm find with that

engine/execution/ingestion/stop_control.go
m4ksio: I'm not sure what we would gain from this.
Its too late to halt execution in those methods, and the one which allow skipping blocks already has this check.

PR #3218, opened 2022-09-13
Bootstrap to single address

janezpodhostnik:
This changes bootstrapping to bootstrap all core contracts to one address (the service account address)

The benefit is that all contracts are on one account which makes it easier for all transient chain users to manage stuff. Transient chains being: Benchnet, Emulator, (Playground), ...

the drawback is that the bootstrapped account structure is not the same as on mainnet or testnet or canary or Sandboxnet. But that was already the case for the epoch part of bootstrapping.


janezpodhostnik:
bors try

Kay-Zee:
what's the status here @janezpodhostnik

janezpodhostnik:
This is probably something we still want to do, but this PR is old, and its easier to just do it again.

PR #3219, opened 2022-09-13
[FVM] Remove old quick fixes

janezpodhostnik:
Both of these were already properly fixed either in the FVM or in cadence. 

The reason I am removing them now is because the fuzz tests test exactly what these quick fixes were for: a scenario where the node crashes but shouldn't, because of reaching the limits at a bad location in the code.

I ran the fuzzer for an entire day after removing this.


pattyshack (approved):

Tonix517 (approved):

janezpodhostnik:
bors merge

Issue #3102, opened 2022-08-26
[FVM] Pre-fetch FVM settings from the state once per block


janezpodhostnik:
## Problem Definition

Currently FVM read some settings from the state on every transaction:
- execution effort weights
- memory weights
- memory limit

Even though the registers themselves are cached and publicly accessible, this still takes a big chunk of time. These reads are also a blocker for concurrent transaction processing.

## Proposed Solution

Only read these settings once per block.

Thinks to keep in mind:
- verification nodes only get a subset of transactions to verify, and they only get the date that was touched during running the transactions on the execution nodes. The registers where the settings are kept should also be sent to the verification nodes
- Changing the settings should only take affect at the next block

## Definition of Done

- FVM reads settings from the state only once per block

PR #3216, opened 2022-09-12
Simplify Process, Processor, Environment initialization

pattyshack:
Drop vm argument + skip creating programs in tests.

PR #3231, opened 2022-09-14
[FVM] Removing unused VM param from NewScript\TransationEnvironment()

Tonix517:
As a prep work for #3102. 

At present the `vm` param is not used in either `NewScriptEnvironment()` or `NewTransactionEnvironment()` method, also given that VirtualMachine is not part of `commonEnv`, it should be safe to remove it from the two functions. 

Didn't continue this removal to `Process()` of `ScriptProcessor` and `TransactionProcessor` because according to my understanding these two functions might need to use VirtualMachine.

### DIFF
--- a/cmd/util/ledger/reporters/account_reporter.go
+++ b/cmd/util/ledger/reporters/account_reporter.go
@@ -150,7 +150,7 @@ func NewBalanceReporter(chain flow.Chain, view state.View) *balanceProcessor {
 	)
 	accounts := environment.NewAccounts(stTxn)
 
-	env := fvm.NewScriptEnvironment(context.Background(), ctx, vm, stTxn, progs)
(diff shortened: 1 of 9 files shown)


pattyshack:
This PR won't work because it will break integration testing (I did the same thing last week =P).  Integration testing using emulator, which is pinned against an older version of flow.  We need to preserve the original api until the emulator can be updated.  I ended up doing this #3216  instead

pattyshack (commented):
(move this PR off my review queue)

Tonix517:
OK, i was wondering.. thanks for letting me know. Will close it.

> This PR won't work because it will break integration testing (I did the same thing last week =P). Integration testing uses the emulator, which is pinned against an older version of flow. We need to preserve the original api until the emulator can be updated. I ended up doing this #3216 instead

Issue #3103, opened 2022-08-26
[FVM] Move meter-able things into meter 


janezpodhostnik:
## Problem Definition

The FVM Meter is currently responsible for metering computation (a.k.a.: execution effort) and memory.

However the FVM is also metering/limiting other things:
- storage interaction
- events emitted

These should also be part of the meter, so there is a single way how to meter and limit stuff.

## Proposed Solution

Move metering and limiting of  storage interaction and events emitted into the FVM meter

## Definition of Done

- the FVM has a single way of metering and limiting meter-able quantities.

PR #3236, opened 2022-09-15
[FVM] Adding storage metering to Meter

Tonix517:
#3103 - we want to move existing storage metering logic in `state.go` into `Meter` interface.

As the first step, now adding storage read\write byte count metering interface to `Meter` interface. Next PR will replace related logic in `state.go` with the new Meter interface.

### DIFF
--- a/fvm/meter/meter.go
+++ b/fvm/meter/meter.go
@@ -1,6 +1,10 @@
 package meter
 
-import "github.com/onflow/cadence/runtime/common"
+import (
+	"github.com/onflow/cadence/runtime/common"
+
+	"github.com/onflow/flow-go/model/flow"
+)
 
 type MetringOperationType uint
 
@@ -41,6 +45,7 @@ const (
 
 type MeteredComputationIntensities map[common.ComputationKind]uint
 type MeteredMemoryIntensities map[common.MemoryKind]uint
+type MeteredStorageInteractionMap map[StorageInteractionKey]uint64
 
 type Meter interface {
 	// merge child funcionality
@@ -59,10 +64,17 @@ type Meter interface {
 	TotalMemoryEstimate() uint64
 	TotalMemoryLimit() uint64
 
-	// TODO move storage metering to here
-	// MeterStorageRead(byteSize uint) error
-	// MeterStorageWrite(byteSize uint) error
-	// TotalBytesReadFromStorage() int
-	// TotalBytesWroteToStorage() int
-	// TotalBytesOfStorageInteractions() int
+	// storage metering
+	MeterStorageRead(storageKey StorageInteractionKey, value flow.RegisterValue, enforceLimit bool) error
+	MeterStorageWrite(storageKey StorageInteractionKey, value flow.RegisterValue, enforceLimit bool) error
+	StorageUpdateSizeMap() MeteredStorageInteractionMap
+	TotalBytesReadFromStorage() uint64
+	TotalBytesWrittenToStorage() uint64
+	TotalStorageReadCount() uint64
+	TotalStorageWriteCount() uint64
+	TotalBytesOfStorageInteractions() uint64
+}
+
+type StorageInteractionKey struct {
+	Owner, Key string
 }
--- a/fvm/meter/meter.go
+++ b/fvm/meter/meter.go
@@ -70,8 +70,6 @@ type Meter interface {
 	StorageUpdateSizeMap() MeteredStorageInteractionMap
 	TotalBytesReadFromStorage() uint64
 	TotalBytesWrittenToStorage() uint64
-	TotalStorageReadCount() uint64
-	TotalStorageWriteCount() uint64
 	TotalBytesOfStorageInteractions() uint64
 }
 
--- a/fvm/meter/weighted_meter.go
+++ b/fvm/meter/weighted_meter.go
@@ -449,7 +449,8 @@ func (m *WeightedMeter) TotalMemoryEstimate() uint64 {
 
 // MeterStorageRead captures storage read bytes count and returns an error
 // if it goes beyond the total interaction limit and limit is enforced
-func (m *WeightedMeter) MeterStorageRead(storageKey StorageInteractionKey,
+func (m *WeightedMeter) MeterStorageRead(
+	storageKey StorageInteractionKey,
 	value flow.RegisterValue,
 	enforceLimit bool) error {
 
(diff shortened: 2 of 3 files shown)


pattyshack (changes requested):
"storage" (aka interaction) metering logic already exist in https://github.com/onflow/flow-go/blob/master/fvm/state/state.go .  I think the first step is refactoring that into meter (Your PR currently doesn't do that).

(how interaction should be measured in the future is up for debate, but that's a separate discussion.  For additional context, see #2922)

Tonix517:
I was not clear in the description - that is the part we are trying to move into `Meter` interface. The next step will be replacing existing metering logic in state.go with the new meter.

> "storage" (aka interaction) metering logic already exist in https://github.com/onflow/flow-go/blob/master/fvm/state/state.go . I think the first step is refactoring that into meter (Your PR currently doesn't do that).
> 
> (how interaction should be measured in the future is up for debate, but that's a separate discussion. For additional context, see #2922)

Tonix517:
Also I'm seeing two copies of `meter.go`, one in fvm/meter and one in fvm/environment - which one should be used? and should we remove the old one?

> I was not clear in the description - that is the part we are trying to move into `Meter` interface. The next step will be replacing existing metering logic in state.go with the new meter.
> 
> > "storage" (aka interaction) metering logic already exist in https://github.com/onflow/flow-go/blob/master/fvm/state/state.go . I think the first step is refactoring that into meter (Your PR currently doesn't do that).
> > (how interaction should be measured in the future is up for debate, but that's a separate discussion. For additional context, see #2922)

pattyshack:
I don't think we need to preserve the commented out APIs in the Meter interface.  They are outdated and most of them don't make any sense (I've already replace a few of them).

For now, just add new APIs that works with the current implementation in state.go.  For example, you've added WithAddress, presumably to work around the interface.  Instead of bending backwards to work around the interface, we can just pass the address directly into MeterStorageRead (or something along that line).

Also, take a look at updateSize in state.go.  The current interaction limiting uses that to avoid double counting.

In this PR, I think we should preserve the existing interaction metering behavior.  We can iterate on metering behavior in future PRs (I'm in favor of dumbing down interaction metering, but I don't have much stake in this).

janezpodhostnik:
Like Patrick mentioned, both meter storage read and write need to also accept a location (address, key) so that read can ignore double reads, and write can make sure if a write was already counted, that it first deducts it and then counts the new value.

> In this PR, I think we should preserve the existing interaction metering behavior. 
I totally agree. Lets preserve the existing logic in the first step.

pattyshack:
fvm/environment's version is a wrapper on top of fvm/meter.  EnforceLimits checks can probably be eventually removed based on the conversation in #2922, and the cancallable stuff is basically a hack to work around cadence (with some work, this can be removed).  What remains are some environments specific meter constants, which I think should think within the env package

Tonix517:
Thanks for the clarification. I was in tunnel vision for this part. Now existing storage metering logic in `state.go` has been copied over to Meter interface. Next PR will replace state.go to use this new interface.

fvm/meter/weighted_meter.go
```diff
@@ -324,6 +336,12 @@ type WeightedMeter struct {
 
 	computationIntensities MeteredComputationIntensities
 	memoryIntensities      MeteredMemoryIntensities
+
+	storageUpdateSizeMap     map[StorageInteractionKey]uint64
+	storageReadCounter       uint64
```
pattyshack, line 341: I think we can drop the read / write counter. if I remember correctly, we don't use it anywhere.

fvm/meter/weighted_meter.go
```diff
@@ -417,3 +446,95 @@ func (m *WeightedMeter) MemoryIntensities() MeteredMemoryIntensities {
...
+	if _, ok := m.storageUpdateSizeMap[storageKey]; !ok {
+		m.storageReadCounter++
+		readByteSize := getStorageKeyValueSize(storageKey, value)
+		m.totalStorageBytesRead += readByteSize
+		m.storageUpdateSizeMap[storageKey] = readByteSize
+	}
+
+	if enforceLimit &&
+		m.TotalBytesOfStorageInteractions() > m.storageInteractionLimit {
+		return errors.NewStorageCapacityExceededError(
```
pattyshack, line 466: the error return in state is different? errors.NewLedgerInteractionLimitExceededError

also maybe refactor the enforce limit check into method

Tonix517: addressed.

pattyshack (approved):

fvm/meter/weighted_meter.go
```diff
@@ -417,3 +446,88 @@ func (m *WeightedMeter) MemoryIntensities() MeteredMemoryIntensities {
...
+func (m *WeightedMeter) StorageUpdateSizeMap() MeteredStorageInteractionMap {
+	return m.storageUpdateSizeMap
+}
+
+func getStorageKeyValueSize(storageKey StorageInteractionKey,
+	value flow.RegisterValue) uint64 {
+	return uint64(len(storageKey.Owner) + len(storageKey.Key) + len(value))
+}
+
+func GetStorageKeyValueSizeForTesting(storageKey StorageInteractionKey,
```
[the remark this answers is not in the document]
Tonix517: addressed

fvm/meter/weighted_meter.go
Tonix517: discussed offline with Janez. Yes we can remove these counters from code. They don't seem to be useful at least in the foreseeable future.

fvm/meter/weighted_meter.go
```diff
@@ -368,6 +386,13 @@ func (m *WeightedMeter) MergeMeter(child Meter) {
 	for key, intensity := range child.MemoryIntensities() {
 		m.memoryIntensities[key] += intensity
 	}
+
+	// merge storage meters
+	for key, value := range child.StorageUpdateSizeMap() {
+		m.storageUpdateSizeMap[key] += value
```
janezpodhostnik, line 392: They should not be summed. Only take the child value.

fvm/meter/meter.go
```diff
@@ -59,10 +64,15 @@ type Meter interface {
...
-	// TODO move storage metering to here
-	// MeterStorageRead(byteSize uint) error
-	// MeterStorageWrite(byteSize uint) error
-	// TotalBytesReadFromStorage() int
-	// TotalBytesWroteToStorage() int
-	// TotalBytesOfStorageInteractions() int
+	// storage metering
+	MeterStorageRead(storageKey StorageInteractionKey, value flow.RegisterValue, enforceLimit bool) error
+	MeterStorageWrite(storageKey StorageInteractionKey, value flow.RegisterValue, enforceLimit bool) error
+	StorageUpdateSizeMap() MeteredStorageInteractionMap
```
janezpodhostnik, line 70: This should not be part of the interface.

janezpodhostnik (changes requested):
Overall looks good, just 2 issues.

fvm/meter/weighted_meter.go
Tonix517: addressed.

fvm/meter/meter.go
Tonix517: our current MergeMeter method takes `Meter` instead of `WeightedMeter`, thus to access WeightedMeter internals for merge purposes, we will need this interface to retrieve child data to merge. Similar for `ComputationIntensities()` and `MemoryIntensities()` above. Unless we do downcast from `*Meter` to `*WeightedMeter` in `MergeMeter()`, we can get rid of these interfaces. WDYT

Tonix517: sounds good, I can do this deletion in a separate PR.

janezpodhostnik (approved):

Tonix517:
bors merge

PR #3238, opened 2022-09-15
Clean up environment api and structs

pattyshack:
1. renamed commonEnv to facadeEnvironment
2. removed TransactionEnv and ScriptEnv structs
3. simplify env initialization a bit
4. updated Environment interface
5. renamed Commit to FlushPendingUpdates
6. removed ModifiedSets type alias
7. exposed a few more TransactionInfo methods

PR #3242, opened 2022-09-16
[Testing] de-flake TestGenerateTestMatrix

SaveTheRbtz:
Noticed in #3238

@gomisha, this is a good example of a case where CODEOWNERS-based annotation could help: in theory we would want to notify test owners of their flakes.

This is also an example of a test that should be auto-quarantined.


gomisha (approved):

gomisha:
> 

Thanks for fixing this! Yes, will update CODEOWNERS for automation, CI related folders. Auto quarantine is on the automation roadmap.

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #3245, opened 2022-09-16
[FVM] Adding IsMemoryLimitSet() to MeterParameters

Tonix517:
For `MeterParameters`, memoryLimit is [overridden](https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L86) only when state returns valid value ([code](https://github.com/onflow/flow-go/blob/master/fvm/executionParameters.go#L106)).

As a prep work for #3102, we will be moving out state reading for MeterParameters out of `fvm.RunV2()` and will do the read before each collection execution, then passing in the read meter parameters into `fvm.RunV2()`, so if we still want to treat memoryLimit from state as higher priority, we would need to know that with a given MeterParameters obj, if memoryLimit value has been set, as the condition to override existing value from [Procedures](https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L84).

Adding a new interface to serve this purpose.

### DIFF
--- a/fvm/meter/weighted_meter.go
+++ b/fvm/meter/weighted_meter.go
@@ -252,8 +252,9 @@ type MeterParameters struct {
 	computationLimit   uint64
 	computationWeights ExecutionEffortWeights
 
-	memoryLimit   uint64
-	memoryWeights ExecutionMemoryWeights
+	isMemoryLimitSet bool
+	memoryLimit      uint64
+	memoryWeights    ExecutionMemoryWeights
 }
 
 func DefaultParameters() MeterParameters {
@@ -284,6 +285,7 @@ func (params MeterParameters) WithComputationWeights(
 func (params MeterParameters) WithMemoryLimit(limit uint64) MeterParameters {
 	newParams := params
 	newParams.memoryLimit = limit
+	newParams.isMemoryLimitSet = true
 	return newParams
 }
 
@@ -306,6 +308,13 @@ func (params MeterParameters) TotalComputationLimit() uint {
 
 func (params MeterParameters) MemoryWeights() ExecutionMemoryWeights {
 	return params.memoryWeights
+}
+
+// IsMemoryLimitSet returns if memory limit has been set by WithMemoryLimit()
+// It is useful when metering parameters are loaded from state. We need to
+// know if state has this value available to decide whether to override existing value.
+func (params MeterParameters) IsMemoryLimitSet() bool {
+	return params.isMemoryLimitSet
 }
 
 // TotalMemoryLimit returns the total memory limit
--- a/fvm/meter/weighted_meter_test.go
+++ b/fvm/meter/weighted_meter_test.go
@@ -376,3 +376,18 @@ func TestMemoryWeights(t *testing.T) {
 		)
 	}
 }
+
+func TestMeterParameters(t *testing.T) {
+	t.Run("IsMemoryLimitSet return false - without WithMemoryLimit() called", func(t *testing.T) {
+		meterParams := meter.DefaultParameters()
+		require.False(t, meterParams.IsMemoryLimitSet())
+		require.Equal(t, uint64(math.MaxUint64), meterParams.TotalMemoryLimit())
+	})
+
+	t.Run("IsMemoryLimitSet return true - with WithMemoryLimit() called", func(t *testing.T) {
+		testMemoryLimit := uint64(123)
+		meterParams := meter.DefaultParameters().WithMemoryLimit(testMemoryLimit)
+		require.True(t, meterParams.IsMemoryLimitSet())
+		require.Equal(t, testMemoryLimit, meterParams.TotalMemoryLimit())
+	})
+}


pattyshack (commented):
we can't actually move the meterParameter setup out of RunV2 cuz there are other binaries (verifier, emulator, etc.), which should still do the setup within RunV2.  We can however specify an optional parameter in fvm.Context (similar to WithBlockPrograms), and have RunV2 use the optional parameters when it's specified.

for example, you can call getEnvironmentMeterParameters (https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L86) from within the computer, and pass that to RunV2 via the context (note that you will also need to pass the state, not just the meter parameter)

Tonix517:
Yes that's actually what my draft PR https://github.com/onflow/flow-go/pull/3240 does, and this PR is mainly for this specific line: https://github.com/onflow/flow-go/pull/3240/files#diff-a3887556c2c019d31cbf16a7fcf06822da0c5618482925349511c040d8901cabR82

> we can't actually move the meterParameter setup out of RunV2 cuz there are other binaries (verifier, emulator, etc.), which should still do the setup within RunV2. We can however specify an optional parameter in fvm.Context (similar to WithBlockPrograms), and have RunV2 use the optional parameters when it's specified.
> 
> for example, you can call getEnvironmentMeterParameters (https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L86) from within the computer, and pass that to RunV2 via the context (note that you will also need to pass the state, not just the meter parameter)

Tonix517:
Let me put more details in the description.

> we can't actually move the meterParameter setup out of RunV2 cuz there are other binaries (verifier, emulator, etc.), which should still do the setup within RunV2. We can however specify an optional parameter in fvm.Context (similar to WithBlockPrograms), and have RunV2 use the optional parameters when it's specified.
> 
> for example, you can call getEnvironmentMeterParameters (https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L86) from within the computer, and pass that to RunV2 via the context (note that you will also need to pass the state, not just the meter parameter)

fvm/meter/weighted_meter.go
```diff
@@ -308,6 +310,13 @@ func (params MeterParameters) MemoryWeights() ExecutionMemoryWeights {
...
 
+// IsMemoryLimitSet returns if memory limit has been set by WithMemoryLimit()
+// It is useful when metering parameters are loaded from state. We need to
+// know if state has this value available to decide whether to override existing value.
+func (params MeterParameters) IsMemoryLimitSet() bool {
+	return params.isMemoryLimitSet
+}
+
 // TotalMemoryLimit returns the total memory limit
 func (params MeterParameters) TotalMemoryLimit() uint64 {
```
zhangchiqing, line 321: This is not a safe pattern, because we assumes the caller will call `IsMemoryLimitSet` before `TotalMemoryLimit`. 
 It would cause a bug if the caller forget to call `IsMemoryLimitSet` before `TotalMemoryLimit`.

A safer change would be letting `TotalMemoryLimit` to return two values:
```
// TotalMemoryLimit returns the total memory limit
// it returns (0, false) if memory limit was not set
// it returns (limit, true) if memory limit was set by WithMemoryLimit()
func (params MeterParameters) TotalMemoryLimit() (uint64, bool) {
```

Tonix517: Good point Leo. I thought about the use case of these interfaces, and considered it is fine to call `TotalMemoryLimit()` with or without `IsMemoryLimitSet()` since it will always returns valid value, even it is the default `math.MaxUint64`. For now the only scenario that `IsMemoryLimitSet()` is necessary is [here](https://github.com/onflow/flow-go/blob/master/fvm/executionParameters.go#L103), where we do need to know if state contains this setting to decide whether to override existing value. @janezpodhostnik

zhangchiqing: In that case we don't need the `isMemoryLimitSet`, we could just do:

```
func (params MeterParameters) IsMemoryLimitSet() bool {
	return params.TotalMemoryLimit() == math.MaxUint64
}
```

Tonix517: According to https://github.com/onflow/flow-go/blob/master/fvm/executionParameters.go#L103, we override existing meter setting only when state does contain this value, so if state does have `math.MaxUint64`, we still want to read it out and override the setting.

zhangchiqing: OK. It's not clear to me where is `IsMemoryLimitSet` being used? 

> // We need to know if state has this value available to decide whether to override existing value.

How does this work? It seems we only set `isMemoryLimitSet`, never read from it, no?

Tonix517: I have a draft PR (#3240) in which I plan to move out the state reading for meter params from fvm `RunV2()` which is called per tx, and do state reading at per collection level. the read state values will be passed into `RunV2()` in which this `IsMemoryLimitSet()` will be used to override existing setting values from `*Procedure`.

janezpodhostnik: I gave this a long think.

The problem is that the limit is coming from three sources:
- FVM hardcoded default
- context (set at node boot)
- state (controlled by governance)

the limit in the state can bu null at which point we should take the context one.

We currently do this  by doing:
```
meterParams := meter.DefaultParameters().
		WithComputationLimit(uint(proc.ComputationLimit(ctx))).
		WithMemoryLimit(proc.MemoryLimit(ctx))

	meterParams, err := getEnvironmentMeterParameters( 
		ctx,
		v,
		blockPrograms,
		meterParams, // <--- we only modify the limit if its actually set in the state. getEnvironmentMeterParameters already knows how to do this
	)
``` 

In PR https://github.com/onflow/flow-go/pull/3240/ you are moving `getEnvironmentMeterParameters` to the collection level, so the state parameters are read before the context ones and this causes that you need to keep track if the limit was set or not. 

I think the solution might be that the context read is also moved to the collection level. See my comment here: https://github.com/onflow/flow-go/pull/3240/files#r978947929

Tonix517: that makes sense. so basically we should be move all meter setting logic out of `RunV2` and then we can get rid of this flag.

Tonix517:
Given https://github.com/onflow/flow-go/pull/3240/files#r978947929, we don't need this change anymore.

PR #3246, opened 2022-09-16
[Benchmark] Ability to dynamically adjust TPS

SaveTheRbtz:
Added `SetTPS` API that can be used to adjust the load on the fly.  Switched `manual` loader to using this instead of re-creating benchmark client on every run.

[It's better to review this PR w/o whitespace](https://github.com/onflow/flow-go/pull/3246/files?diff=split&w=1).

### DIFF
--- a/integration/benchmark/contLoadGenerator.go
+++ b/integration/benchmark/contLoadGenerator.go
@@ -168,8 +168,10 @@ func New(
 		lg.workFunc = lg.sendAddKeyTx
 	case ConstExecCostLoadType:
 		lg.workFunc = lg.sendConstExecCostTx
-	default:
+	case CompHeavyLoadType, EventHeavyLoadType, LedgerHeavyLoadType:
 		lg.workFunc = lg.sendFavContractTx
+	default:
+		return nil, fmt.Errorf("unknown load type: %s", loadParams.LoadType)
 	}
 
 	return lg, nil
--- a/integration/benchmark/contLoadGenerator.go
+++ b/integration/benchmark/contLoadGenerator.go
@@ -262,7 +262,6 @@ func (lg *ContLoadGenerator) startWorkers(numWorkers, diff int) {
 		worker.Start()
 		lg.workers = append(lg.workers, worker)
 	}
-	lg.workerStatsTracker.AddWorkers(diff)
 }
 
 func (lg *ContLoadGenerator) stopWorkers(numWorkers, diff int) {
@@ -280,7 +279,6 @@ func (lg *ContLoadGenerator) stopWorkers(numWorkers, diff int) {
 	wg.Wait()
 
 	lg.workers = lg.workers[:start]
-	lg.workerStatsTracker.AddWorkers(-diff)
 }
 
 func (lg *ContLoadGenerator) SetTPS(num uint) error {
@@ -296,6 +294,7 @@ func (lg *ContLoadGenerator) SetTPS(num uint) error {
 	case diff < 0:
 		lg.stopWorkers(numWorkers, -diff)
 	}
+	lg.workerStatsTracker.AddWorkers(diff)
 	return nil
 }
 
--- a/integration/benchmark/contLoadGenerator.go
+++ b/integration/benchmark/contLoadGenerator.go
@@ -281,18 +281,22 @@ func (lg *ContLoadGenerator) stopWorkers(numWorkers, diff int) {
 	lg.workers = lg.workers[:start]
 }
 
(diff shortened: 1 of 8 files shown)


integration/localnet/Makefile
```diff
@@ -76,7 +76,7 @@ stop:
 
 .PHONY: load
 load:
-	go run --tags relic ../benchmark/cmd/manual -log-level info -tps 1,0,10,0,100 -tps-durations 30s,10s,30s,10s
+	go run --tags relic ../benchmark/cmd/manual -log-level info -tps 1,10,100 -tps-durations 30s,30s
```
SaveTheRbtz, line 79: 0's are not needed.  This was previously used because we did not wait for benchmark to stop and these added 10s sleeps between runs.

integration/benchmark/contLoadGenerator.go
```diff
@@ -239,50 +254,55 @@ func (lg *ContLoadGenerator) setupFavContract() error {
...
 			w.Stop()
-		}()
+		}(lg.workers[i])
 	}
 	wg.Wait()
+
+	lg.workers = lg.workers[:start]
+}
+
+func (lg *ContLoadGenerator) SetTPS(num uint) error {
```
zhangchiqing, line 282:
```suggestion
// SetTPS compares the given tps to the existing tps to determine whether to increase
// or decrease the load.
// it increases/decreases the load by adjusting the number of workers, since each worker
// is responsible for sending the load at 1 tps.
func (lg *ContLoadGenerator) SetTPS(num uint) error {
```

integration/benchmark/contLoadGenerator.go
```diff
@@ -157,7 +158,19 @@ func New(
...
+	// TODO(rbtz): hide load implementation behind an interface
+	switch loadParams.LoadType {
+	case TokenTransferLoadType:
+		lg.workFunc = lg.sendTokenTransferTx
+	case TokenAddKeysLoadType:
+		lg.workFunc = lg.sendAddKeyTx
+	case ConstExecCostLoadType:
+		lg.workFunc = lg.sendConstExecCostTx
+	default:
+		lg.workFunc = lg.sendFavContractTx
```
zhangchiqing, line 170: better to be more strict:
```suggestion
	case FavContractLoadType: // Leo: I'm guessing the type here, please use the actual type
		lg.workFunc = lg.sendFavContractTx
	default:
		return nil, fmt.Errorf("unknown load type: %v", loadParams.LoadType)	
```

zhangchiqing (approved):

integration/benchmark/contLoadGenerator.go
```diff
@@ -239,50 +254,55 @@ func (lg *ContLoadGenerator) setupFavContract() error {
...
+	numWorkers := len(lg.workers)
+	diff := int(num) - numWorkers
+
+	switch {
+	case diff > 0:
+		lg.startWorkers(numWorkers, diff)
+	case diff < 0:
+		lg.stopWorkers(numWorkers, -diff)
+	}
+	lg.workerStatsTracker.AddWorkers(diff)
```
pattyshack, line 301: increment / decrement the count inside the startWorkers/stopWorkers, in case we accidentally call those directly in the future

integration/benchmark/contLoadGenerator.go
```diff
@@ -239,50 +256,59 @@ func (lg *ContLoadGenerator) setupFavContract() error {
...
-	lg.stopped = true
+func (lg *ContLoadGenerator) stopWorkers(numWorkers, diff int) {
 	wg := sync.WaitGroup{}
-	wg.Add(len(lg.workers))
-	for _, w := range lg.workers {
-		w := w
+	wg.Add(diff)
 
-		go func() {
+	start := numWorkers - diff
```
pattyshack, line 271: be paranoid and check for bound

integration/benchmark/contLoadGenerator.go
```diff
@@ -58,6 +58,8 @@ type ContLoadGenerator struct {
 	accounts            []*flowAccount
 	availableAccounts   chan *flowAccount // queue with accounts available for workers
 	workerStatsTracker  *WorkerStatsTracker
+	mu                  sync.Mutex
```
pattyshack, line 61: it's impossible to tell what's guarded by this mutex (if not looking at this specific PR).  add comment next to fields that are guarded by this mutex.  some newlines to divide the fields into sections would be useful as well.

pattyshack (approved):

AndrewM-SDET (approved):

SaveTheRbtz:
bors merge

PR #3247, opened 2022-09-17
suggested amendments to PR #2861

AlexHentschel:
## _Optional_ suggested amendments to PR https://github.com/onflow/flow-go/pull/2861

* adopted go1.19 formatting guidelines https://go.dev/doc/comment#syntax
* renamed `aggregationEmptyListError` -> to `blsEmptyListError` to reflect the more general usage patterns
* compactified some code by removing white spaces
* in `bls_multisig.go`: refactored hasher checks into one function
* in `bls_multisig.go`: copied documentation from error detecting function to the error struct 
* typos


tarakby:
Thanks @alex for the nice cleanup! Apologies for leaving many typos behind. 

I merged the PR. I will keep all the suggestions but I will modify the naming and doc for `blsEmptyListError`. The reason is that the error is not related to a software/language constraint where an empty list causes a problem. It is a specific case to aggregation where we don't want to return the "zero" element to avoid equivocation issues in signatures.

PR #3248, opened 2022-09-19
Collection`compliance`, `epochmgr` Engines Implement `component.Component`

jordanschalm:
⚠️ Depends on https://github.com/onflow/flow-go/pull/3121

This PR updates `collection/compliance.Engine` and `collection/epochmgr.Engine` to implement `component.Component` (https://github.com/onflow/flow-go/pull/3121, but for collection nodes).


engine/collection/compliance/engine.go
```diff
@@ -505,23 +502,20 @@ func (e *Engine) BroadcastProposalWithDelay(header *flow.Header, delay time.Dura
...
 		if err != nil {
-			log.Error().Err(err).Msg("could not broadcast proposal")
-			return
+			if errors.Is(err, network.EmptyTargetList) {
+				return
+			}
+			log.Err(err).Msg("could not send proposal message")
 		}
 
-		log.Info().Msg("cluster proposal was broadcast")
```
zhangchiqing, line 516: I think this log is still useful

engine/collection/epochmgr/epoch_components.go
```diff
@@ -0,0 +1,107 @@
...
+		// wait until all components start
+		<-util.AllReady(components.prop, components.sync, components.voteAggregator, components.timeoutAggregator)
+
+		// signal that startup has finished, and we are ready to go
+		ready()
+
+		// wait for shutdown to be commenced
+		<-ctx.Done()
+		// wait for compliance engine and event loop to shut down
+		<-util.AllDone(components.prop, components.sync, components.voteAggregator, components.timeoutAggregator)
```
zhangchiqing, line 60: can itself shut down before other components to shut down?

would it miss a few events if it's shut down earlier?

jordanschalm: > can itself shut down before other components to shut down?

Do you mean, "can the `epochmgr` engine shut down before the epoch components shut down?" (What is `itself` referring to?)

> would it miss a few events if it's shut down earlier?

Yes, although I think we can miss events regardless of the order in which we are shutting down. Shutting down means, we are either turning off an old cluster consensus or shutting down the whole node. In neither case are missed events problematic from my perspective.

zhangchiqing: Oh, I see. the ctx is actually the shut down signal itself. 

I thought it was an engine's Done channel, actually we are moving away from that pattern.

zhangchiqing (approved):

engine/collection/compliance/engine.go
jordanschalm: 👍 agree, this was a mistake - I re-added the log

jordanschalm:
CI Test Failure:
```
--- FAIL: TestClusterSwitchover_MultiCollectorCluster (11.04s)
    cluster_switchover_test.go:262: expecting transaction 08f3dc749f467a58c4e7da5a4d5607df6defa9403dd7c276517d143f962d70a7 in epoch 1 for cluster 0
    cluster_switchover_test.go:403: got guarantee from 909b9654e4c1256195ff459ab9f56a3135d403f886f8c53e5bc5355085ced986
    cluster_switchover_test.go:403: got guarantee from 4e4cdb4c29b887885c54c628be0daab61d8602df492034f66a15d8dfa87e6b96
    cluster_switchover_test.go:262: expecting transaction ec7bf89c3d4b59ce6977d34d02a26b8792c9cade28b998137172317aca7513da in epoch 2 for cluster 0
    cluster_switchover_test.go:403: got guarantee from 909b9654e4c1256195ff459ab9f56a3135d403f886f8c53e5bc5355085ced986
    cluster_switchover_test.go:403: got guarantee from 4e4cdb4c29b887885c54c628be0daab61d8602df492034f66a15d8dfa87e6b96
    irrecoverable.go:17: 
        	Error Trace:	/home/runner/work/flow-go/flow-go/engine/collection/test/irrecoverable.go:17
        	            				/home/runner/work/flow-go/flow-go/engine/collection/test/asm_amd64.s:1594
        	Error:      	Received unexpected error:
        	            	could not process proposal feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226: could not add block (feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226) to vote aggregator: could not process block: feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226, internal error updating VoteProcessor's status from VoteCollectorStatusCaching to VoteCollectorStatusVerifying for block feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226: failed to create VerifyingVoteProcessor for block feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226: instantiating vote processor for block feee1c1dae702e7645966b68249d6d7d230b1d573b2001476fa10840fa556226 failed: error retrieving consensus participants: could not retrieve resource: key not found
        	Test:       	TestClusterSwitchover_MultiCollectorCluster
        	Messages:   	observed unexpected irrecoverable error
FAIL
coverage: [no statements]
FAIL	github.com/onflow/flow-go/engine/collection/test	32.680s
```
[CI Run](https://github.com/onflow/flow-go/actions/runs/3114437099/jobs/5050281221)

engine/collection/synchronization/engine.go
```diff
@@ -307,7 +307,12 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.Cluster
 			OriginID: originID,
 			Block:    block,
 		}
-		e.comp.SubmitLocal(synced)
+		// forward the block to the compliance engine for validation and processing
+		// we use the network.MessageProcessor interface here because the block is un-validated
+		err := e.comp.Process(channels.SyncCluster(block.Header.ChainID), originID, synced)
```
durkmurder, line 312: I think we need to pass me.NodeID() here instead of originID, original sender is already part of `SyncedClusterBlock`

durkmurder (commented):
First batch of comments

network/errors.go
```diff
@@ -2,8 +2,39 @@ package network
...
 )
+
+// TransientError represents an error returned from a network layer function call
+// which may be interpreted as non-critical. In general, we desire that all expected
+// error return values are enumerated in a function's documentation - any undocumented
+// errors are considered fatal. However, 3rd party libraries don't always conform to
+// this standard, including the networking libraries we use. This error type can be
+// used to wrap these 3rd party errors on the boundary into flow-go, to explicitly
+// mark them as non-critical.
+type TransientError struct {
```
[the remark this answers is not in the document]
jordanschalm: It's definitely not a good practice to do this when critical vs non-critical errors can be programmatically differentiated. But not every dependency we use returns errors that can be parsed this way. The reason for introducing this error type is to make interactions with these dependencies compatible with our error handling approach as much as possible. 

We already treat [eg. networking errors](https://github.com/onflow/flow-go/blob/ccbd21bee9f1f9fab8b1fdb72db3cb66a8229650/engine/collection/compliance/engine.go#L367-L371) like this regularly, by interpreting generic errors as non-critical, like:
```go
err := conduit.Unicast(...)
if err != nil {
    log.Err(err).Msg("failed to send message)
    return nil
}
```

If we wrap these errors with a documented type (`TransientError`), we can at least:
* handle unexpected errors further up the stack
* retry on networking failures
For example:
```go
// Error returns:
// - TransientError if dependencyWithoutDefinedErrors returns any error
// - generic error in case of unexpected 
func DoSomethingIncludingANetworkCall() error {
    err := dependencyWithoutDefinedErrors.SomethingThatMightTransientlyFail(...)
    if err != nil {
        return NewTransientError(err)
    }

    err := internalComponent.DoSomething()
    if err != nil {
        return fmt.Errorf("unexpected error in internalComponent: %w", err)
    }

    return nil
}
```

If we don't wrap the first error above, then we can't differentiate between the generic error from `dependencyWithoutDefinedErrors` (in the networking layer we generally want to treat these as non-critical) and the generic error form `internalComponent` (we must crash the component or node if we observe this).

engine/collection/epochmgr/epoch_components.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/util"
+	"github.com/onflow/flow-go/state/cluster"
+)
+
+// EpochComponents represents all dependencies for running an epoch.
+type EpochComponents struct {
+	*component.ComponentManager
+	state             cluster.State
+	prop              component.Component
```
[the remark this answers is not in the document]
jordanschalm: Nice catch - it used to be called `proposal` engine. We renamed it to compliance a while ago, but missed it here.

engine/collection/epochmgr/epoch_components.go
```diff
@@ -0,0 +1,77 @@
...
+		<-ctx.Done()
+		// wait for compliance engine and event loop to shut down
+		<-util.AllDone(components.prop, components.sync, components.voteAggregator, components.timeoutAggregator)
+	})
+	components.ComponentManager = builder.Build()
+
+	return components
+}
+
+type StartableEpochComponents struct {
```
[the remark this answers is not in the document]
jordanschalm: 👍 added, and also renamed this for more clarity

engine/collection/epochmgr/engine.go
```diff
@@ -1,233 +1,189 @@
...
-
 	components, err := e.createEpochComponents(epoch)
-	// don't set up consensus components if we aren't authorized in current epoch
-	if errors.Is(err, ErrNotAuthorizedForEpoch) {
-		return e, nil
+	if err != nil {
+		if errors.Is(err, ErrNotAuthorizedForEpoch) {
+			// don't set up consensus components if we aren't authorized in current epoch
+			e.log.Info().Msg("node is not authorized for current epoch - skipping initializing cluster consensus")
+			return
```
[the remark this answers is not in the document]
jordanschalm: Yes we do, because at the next epoch transition we will need to handle the protocol event and start up consensus for the new epoch. We also need to respond to events in order to submit our root block vote, even if we aren't authorized for the current epoch.

network/errors.go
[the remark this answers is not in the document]
durkmurder: I agree that this is useful, and we probably want to use it, but I would like to see some checks for _some_ errors. I mean I am not very comfortable when we swallow _all_ errors without checking what has exactly happened.

module/epochs/qc_client.go
```diff
@@ -145,19 +151,25 @@ func (c *QCContractClient) SubmitVote(ctx context.Context, vote *model.Vote) err
...
+		return false, network.NewTransientErrorf("could not execute voted script: %w", err)
+	}
+
+	if _, ok := ret.(cadence.Bool); !ok {
+		return false, fmt.Errorf("unexpected cadence type (%T) returned from Voted script", ret)
 	}
 
 	// check if node has voted
-	if !hasVoted.(cadence.Bool) {
+	if !ret.(cadence.Bool) {
```
durkmurder, line 171: you can use previous value from cast to avoid double casting

engine/collection/compliance/engine.go
```diff
@@ -223,68 +233,37 @@ func (e *Engine) WithSync(sync module.BlockRequester) *Engine {
...
-		e.log.Info().Msg("shutting down hotstuff eventloop")
-		e.stopHotstuff()
-		<-e.core.hotstuff.Done()
-		e.log.Info().Msg("all components have been shut down")
-		<-e.unit.Done()
-	})
-	return e.lm.Stopped()
-}
+	e.log.Info().Msg("starting hotstuff")
+	e.core.hotstuff.Start(ctx)
```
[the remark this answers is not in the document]
jordanschalm: Totally agree - might be better to do that along with `MessageHub` though. I feel like changing it now would largely get clobbered by `MessageHub` rewiring

engine/collection/epochmgr/engine.go
```diff
@@ -1,233 +1,189 @@
...
+	log            zerolog.Logger
+	me             module.Local
+	state          protocol.State
+	pools          *epochpool.TransactionPools // epoch-scoped transaction pools
+	factory        EpochComponentsFactory      // consolidates creating epoch for an epoch
+	voter          module.ClusterRootQCVoter   // manages process of voting for next epoch's QC
+	heightEvents   events.Heights              // allows subscribing to particular heights
+	startupTimeout time.Duration               // how long we wait for epoch components to start up
+
+	mu     sync.Mutex                           // protects epochs map
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3248/commits/f386f51c0853fab041070c67b3f71d0edb5f5dff 👍

engine/collection/synchronization/engine.go
jordanschalm: Yeah, I feel like we're mixing practices a bit here. 
* We decided to use `Process` API, because the messages have not been validated (should be treated as originating from another node)
* But passing `me.NodeID()` as `originID` indicates the message originates from this node (trusted)

The MessageProcessor interface says
https://github.com/onflow/flow-go/blob/6a39c8c64184231a05dd5c756db91307f0361e78/network/engine.go#L43-L47

Overall I think the API doesn't fit well with this usage pattern, where an untrusted message is passing through an internal component, without being validated first. Though setting `originID` to the local node ID for an untrusted message seems like the worse of two not-great options here.

engine/collection/epochmgr/engine.go
```diff
@@ -1,233 +1,189 @@
...
+		heightEvents:                 heightEvents,
+		epochs:                       make(map[uint64]*StartableEpochComponents),
+		startupTimeout:               DefaultStartupTimeout,
+		epochTransitionEvents:        make(chan *flow.Header, 1),
+		epochSetupPhaseStartedEvents: make(chan *flow.Header, 1),
+		epochStopEvents:              make(chan uint64, 1),
 	}
 
+	e.cm = component.NewComponentManagerBuilder().
+		AddWorker(e.handleEpochSetupPhaseStartedLoop).
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3248/commits/eccca74b92646130c74b073a77af20b65855e5bc

network/errors.go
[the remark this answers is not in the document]
jordanschalm: After our discussion, I added comments to each point where we wrap the error saying that we're treating it as non-critical.

engine/collection/synchronization/engine.go
durkmurder: I am not sure here. We need to define clear rules how should we treat such cases. 
For instance sync engine doesn't do any checks so technically it's an untrusted input on other side it didn't come from network layer but from the sync engine. 
I would treat this case as general one as if message comes from unknown origin and only use trusted callbacks when we actually need different processing logic.

durkmurder (approved):
Looks good to me, thanks, very solid. The only thing is we need to decide on message passing pattern between different components, otherwise all good.

engine/collection/synchronization/engine.go
jordanschalm: OK, for now I will do that and pass `me.NodeID` with a comment.

jordanschalm:
bors merge

PR #3249, opened 2022-09-19
[FVM] Replacing existing metering in state.go with new interface

Tonix517:
It replaces existing storage metering functions in state.go with the new interface from #3236

### DIFF
--- a/fvm/state/state_test.go
+++ b/fvm/state/state_test.go
@@ -92,36 +92,6 @@ func TestState_ChildMergeFunctionality(t *testing.T) {
 
 }
 
-func TestState_InteractionMeasuring(t *testing.T) {
-	view := utils.NewSimpleView()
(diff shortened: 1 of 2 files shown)


fvm/state/state_test.go
```diff
@@ -92,36 +92,6 @@ func TestState_ChildMergeFunctionality(t *testing.T) {
 
 }
 
-func TestState_InteractionMeasuring(t *testing.T) {
```
Tonix517, line 95: the coverage of this removed test case has already been taken care of in https://github.com/onflow/flow-go/blob/master/fvm/meter/weighted_meter_test.go#L380

zhangchiqing (approved):

pattyshack (approved):

Tonix517:
bors retry

PR #3261, opened 2022-09-20
Add CodedError / CodedFailure to the errors package

pattyshack:
We can replace most (if not all) of our error / failure structs with CodedError and CodedFailure.

I've converted one error struct to CodedError, and one failure struct to CodeFailure.  I'll convert the rest if we're happy with this approach.

### DIFF
--- a/fvm/errors/codes.go
+++ b/fvm/errors/codes.go
@@ -4,26 +4,27 @@ import "fmt"
 
 type ErrorCode uint16
 
+func (ec ErrorCode) IsFailure() bool {
+	return ec >= FailureCodeUnknownFailure
+}
+
 func (ec ErrorCode) String() string {
+	if ec.IsFailure() {
+		return fmt.Sprintf("[Failure Code: %d]", ec)
+	}
 	return fmt.Sprintf("[Error Code: %d]", ec)
 }
 
-type FailureCode uint16
-
-func (fc FailureCode) String() string {
-	return fmt.Sprintf("[Failure Code: %d]", fc)
-}
-
 const (
-	FailureCodeUnknownFailure     FailureCode = 2000
-	FailureCodeEncodingFailure    FailureCode = 2001
-	FailureCodeLedgerFailure      FailureCode = 2002
-	FailureCodeStateMergeFailure  FailureCode = 2003
-	FailureCodeBlockFinderFailure FailureCode = 2004
+	FailureCodeUnknownFailure     ErrorCode = 2000
+	FailureCodeEncodingFailure    ErrorCode = 2001
+	FailureCodeLedgerFailure      ErrorCode = 2002
+	FailureCodeStateMergeFailure  ErrorCode = 2003
+	FailureCodeBlockFinderFailure ErrorCode = 2004
 	// Deprecated: No longer used.
-	FailureCodeHasherFailure FailureCode = 2005
+	FailureCodeHasherFailure ErrorCode = 2005
 	// Deprecated: No longer used.
-	FailureCodeMetaTransactionFailure FailureCode = 2100
+	FailureCodeMetaTransactionFailure ErrorCode = 2100
 )
 
 const (
--- a/fvm/errors/errors.go
+++ b/fvm/errors/errors.go
@@ -2,6 +2,7 @@ package errors
 
 import (
 	stdErrors "errors"
+	"fmt"
 
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/errors"
@@ -23,7 +24,7 @@ type Error interface {
 // execution.
 type Failure interface {
 	// FailureCode returns the failure code for this error
-	FailureCode() FailureCode
+	FailureCode() ErrorCode
 	// and anything else that is needed to be an error
 	error
 }
@@ -96,4 +97,93 @@ func HandleRuntimeError(err error) error {
 
 	// All other errors are non-fatal Cadence errors.
 	return NewCadenceRuntimeError(runErr)
+}
+
+// TODO(patrick): combine ErrorCodeChecker, Error and Failure interface.
+// We can also combine CodedError and CodedFailure once the interface is
+// unified.
+type ErrorCodeChecker interface {
+	// TODO(patrick): add Code() ErrorCode
+	// TODO(patrick): add IsFailure() bool
+
+	HasErrorCode(code ErrorCode) bool
+
+	error
+}
+
+func HasErrorCode(err error, code ErrorCode) bool {
+	var checker ErrorCodeChecker
+	return As(err, &checker) && checker.HasErrorCode(code)
+}
+
+type codedError struct {
+	code ErrorCode
+
+	errorWrapper
+}
+
+func wrapError(
+	code ErrorCode,
+	rootCause error,
+) codedError {
+	// TODO(patrick): check if rootCause is a failure, and if so, invert the
+	// error nesting (convert the new code into an additional error).
+	return codedError{
+		code:         code,
+		errorWrapper: errorWrapper{rootCause},
+	}
+}
+
+func newError(
+	code ErrorCode,
+	format string,
+	formatArguments ...interface{},
+) codedError {
+	return wrapError(code, fmt.Errorf(format, formatArguments...))
+}
+
+func (err codedError) Error() string {
+	return fmt.Sprintf("%v %v", err.code, err.err)
+}
+
+// This returns true if the codedError's code is set to the specified code, or
+// if any one of the codeError's errors has the error code.
+func (err codedError) HasErrorCode(code ErrorCode) bool {
+	if err.code == code {
+		return true
+	}
+
+	return HasErrorCode(err.err, code)
+}
+
+type CodedError struct {
+	codedError
+}
+
+func NewCodedError(
+	code ErrorCode,
+	format string,
+	arguments ...interface{},
+) *CodedError {
+	return &CodedError{
+		codedError: newError(code, format, arguments...),
+	}
+}
+
+func (err CodedError) Code() ErrorCode {
+	return err.code
+}
+
+type CodedFailure struct {
+	codedError
+}
+
+func WrapCodedFailure(code ErrorCode, err error) *CodedFailure {
+	return &CodedFailure{
+		codedError: wrapError(code, err),
+	}
+}
+
+func (err CodedFailure) FailureCode() ErrorCode {
+	return err.code
 }
--- a/fvm/errors/accounts.go
+++ b/fvm/errors/accounts.go
@@ -7,35 +7,16 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 )
 
-// AccountNotFoundError is returned when account doesn't exist for the given address
-type AccountNotFoundError struct {
-	address flow.Address
-}
-
-// NewAccountNotFoundError constructs a new AccountNotFoundError
-func NewAccountNotFoundError(address flow.Address) AccountNotFoundError {
-	return AccountNotFoundError{
-		address: address,
-	}
-}
-
-func (e AccountNotFoundError) Error() string {
-	return fmt.Sprintf(
-		"%s account not found for address %s",
-		e.Code().String(),
-		e.address.String(),
-	)
-}
-
-// Code returns the error code for this error type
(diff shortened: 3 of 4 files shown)


fvm/errors/errors.go
```diff
@@ -97,3 +99,162 @@ func HandleRuntimeError(err error) error {
...
+func HasErrorCode(err error, code ErrorCode) bool {
+	var checker ErrorCodeChecker
+	return As(err, &checker) && checker.HasErrorCode(code)
+}
+
+type codedError struct {
+	code ErrorCode
+
+	// The first error in the list is the root cause.
+	errors []error
```
zhangchiqing, line 137: Thanks for the idea. 

@janezpodhostnik  and I actually discussed a few improvement ideas for error handling and they are summarized here : https://github.com/dapperlabs/flow-go/issues/6364

For the error code, we do have a solution that is to provide a separate method `FirstCode` which will go through the chain of wrapped errors, and see if they are an instance of Error, and return the code of the first instance of Error, which is most likely the root cause.

fvm/errors/codes.go
```diff
@@ -4,26 +4,27 @@ import "fmt"
 
 type ErrorCode uint16
 
-func (ec ErrorCode) String() string {
-	return fmt.Sprintf("[Error Code: %d]", ec)
+func (ec ErrorCode) IsFailure() bool {
+	return ec >= FailureCodeUnknownFailure
```
zhangchiqing, line 8: It's cool idea that we combine both error type and reuse code.

I had a different solution: since failures are fatal errors, instead of making a Failure type, we could just use fmt.Errorf to create them. 

For instance: 
```
// NewEncodingFailuref formats and returns a new EncodingFailure
func NewEncodingFailuref(msg string, err error) error {
        // Note we don't want to use %w to wrap the error, because that might cause the errors.As to return true 
        // when checking if the error is non-fatal sentinel error.
	return fmt.Errorf("%v, error: %v", msg, err) 
}
```

Based on [our error handling guideline](https://github.com/onflow/flow-go/blob/master/CodingConventions.md#best-practice-guidelines), any non-sentinal errors are fatal errors, so we don't have to define types for them. We can still keep the error code, they can be included in the error message, and get printed out into logs.

zhangchiqing (commented):
Thanks for the idea, please see my comments.

fvm/errors/errors.go
pattyshack: Rather than FirstCode, we should probably return the highest priority error code (sorted by failure vs error, then use first code as tie breaker)

fvm/errors/codes.go
pattyshack: I agree that we can ignore the error code altogether if the error is not user facing, and if we never special case error handling (which we clearly do with the SplitError stuff).

right now, both fvm and flow transactions users are stuck with checking the error string, rather than checking the error code if we (or the user) want to handle specific types of error.  Dropping the error code (/ wrapping fmt.Errorf) doesn't help in that regard.

fvm/errors/errors.go
zhangchiqing: I'm not sure the use case for sorting errors by code or priority. Is it to distinguish failure and error?

If we hit a failure, then it's a fatal error, we might just turn all wrapped errors into an error string, and include it in a fmt.Errorf error. The fatal error will be used to crash the component. In other words, when we hit a failure, the failure error won't be wrapped into a sentinel error. 

This way, the caller will either get a failure or a chain of wrapped non-fatal errors (sentinel errors). 

With a chain of wrapped non-fatal errors, a caller still have full access of all the errors, although they can't be accessed by index (like `errors[len(errors) - 1]`), any error is still accessible by unwrapping.

fvm/errors/codes.go
zhangchiqing: We have plan to get rid of the `SplitErrorType`. It will go away when getting rid of the `Failure` type. 

>right now, both fvm and flow transactions users are stuck with checking the error string, rather than checking the error code if we (or the user) want to handle specific types of error. Dropping the error code (/ wrapping fmt.Errorf) doesn't help in that regard.

I don't understand the difference. 

flow transaction users will only see non-fatal error, probably won't see fatal errors, we will include the first code as the code in the error message, so that for instance, if a user hits fail to check storage capacity due to interaction limit hit error, the error message will include error code for interaction limit hit error. And whether the error code is included in the error message or in a separate field of the response is a separate question, right? 

for fvm, when they see fatal errors, they probably don't care about the error code, because no matter what code, it will crash the component anyway. And when they see non-fatal errors, since all errors are wrapped, they just need to unwrap each layer in order to handle a specific error and get that error code. In other words, they can check the error code instead of checking the error string. right?

pattyshack:
Sounds like your main concern is wrt failure.  That's actually the piece I'm least interested in.  I think it's fine with don't classify failures by code, but i'm interesting in propagating the error code (all the way back to the users)

pattyshack:
I had a chat with Leo.   To summarize the conversation:
1. the primary intention of this PR is to reduce the repetitive error/failure struct definitions (i.e., reuse CodedError for all the errors and failures).
2. we agree that we don't need to handle different types of failure differently (i.e., we can drop the failure code if we want)
3. we agree that we need a way to extract the first user error code (this PR helps in that regard cuz we can extract the error code recursively, and don't need to redefine FirstCode in 30 different structs)

fvm/errors/errors.go
```diff
@@ -97,3 +99,162 @@ func HandleRuntimeError(err error) error {
...
+}
+
+func (err *codedError) prependMsgToRootCause(
+	format string,
+	arguments ...interface{},
+) {
+	err.errors[0] = prependMsgError{
+		msg: fmt.Sprintf(format, arguments...),
+		err: err.errors[0],
+	}
```
zhangchiqing, line 197: What about
```suggestion
	msg := fmt.Sprintf(format, arguments...)
	err.errors[0] = fmt.Errorf("%v: %w", msg, err.errors[0])
	}
```

This way we don't need the `prependMsgError` type.

fvm/errors/errors.go
```diff
@@ -97,3 +99,162 @@ func HandleRuntimeError(err error) error {
...
+	return &CodedError{
+		codedError: newError(code, format, arguments...),
+	}
+}
+
+func (err CodedError) Code() ErrorCode {
+	return err.code
+}
+
+func (err *CodedError) PrependMsgToRootCause(
```
zhangchiqing, line 222: Maybe just call it `Errorf`, as the behavior is very similar to `fmt.Errorf`.
```suggestion
func (err *CodedError) Errorf(
```

zhangchiqing (approved):
Looks good. Nice idea. Just had some minor suggestion

fvm/errors/errors.go
pattyshack: Errorf returns a new error whereas this modify and return the original error.  How about Prependf instead of Errorf?

fvm/errors/errors.go
pattyshack: sgtm

fvm/errors/errors.go
pattyshack: hmm.  I think i'll drop this all together.  most of time, we immediate wrap the newly created error with fmt.Errorf.  I think pass the msg prefix directly into New<blah>Error

fvm/errors/errors.go
```diff
@@ -97,3 +99,147 @@ func HandleRuntimeError(err error) error {
...
+		return fmt.Sprintf("%v %v", err.code, err.errors[0])
+	}
+	return fmt.Sprintf(
+		"%v: %s",
+		err.code,
+		multierror.ListFormatFunc(err.errors))
+}
+
+func (err codedError) Unwrap() error {
+	return err.errors[0]
```
zhangchiqing, line 158: If we use `Caused` to append another error, then that error object seems won't be accessible, because `Unwrap` only returns the first error "errors[0]", is it useful to allow callers to iterate through other errors?

zhangchiqing (commented):
Have two more questions

fvm/errors/errors.go
```diff
@@ -97,3 +99,147 @@ func HandleRuntimeError(err error) error {
...
+
+func (err *CodedError) Prependf(
+	format string,
+	arguments ...interface{},
+) *CodedError {
+	err.prependf(format, arguments...)
+	return err
+}
+
+func (err *CodedError) Caused(additionalError error) *CodedError {
```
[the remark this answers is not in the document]
pattyshack: yeah, I agree that we would need to use/return *CodedError whenever possible within fvm, with some special handling when we call into cadence.  The good news is we can do this piecemeal

fvm/errors/errors.go
```diff
@@ -97,3 +99,123 @@ func HandleRuntimeError(err error) error {
...
+
+// This returns true if the codedError's code is set to the specified code, or
+// if any one of the codeError's errors has the error code.
+func (err codedError) HasErrorCode(code ErrorCode) bool {
+	if err.code == code {
+		return true
+	}
+
+	for _, e := range err.errors {
+		if HasErrorCode(e, code) {
```
janezpodhostnik, line 169: I'm not sure I understand this pattern.

1. I create a NewCodedError which in reality creates 2 errors. The CE itself, and the root cause which is in the `CE.errors[0]`.
2. This gets passed down, and it happens that this causes other problems so I call `CE.Caused`, but I might want to call that with a coded error. But If I do that I have to create another CE2 that contains another root cause...

Would it not be easier to just have a linked list of errors?

fvm/errors/errors.go
janezpodhostnik: If we get rid of failure type, I think its better that the root cause is just the first coded error (in the error.Unwrap stack). I don't think any sorting is needed. But we definitely need to still present the entire stack so its clear which way to read it i.e:

- [Code 123] this happened
- then this happened: [Code 23] asfasf
- then some non-coded error
- which caused the transaction to fail.

fvm/errors/codes.go
janezpodhostnik: I think lets (eventually) get rid of the failure codes (they will never be seen by the user) and failures as a concept, and just have everything that is not a coded error (or does not unwrap to one) to be a failure. 

Since we currently have location where we indiscriminately wrap errors into coded errors and then later try to unwrap them as a failure first, this change is not something we can easily do.

fvm/errors/errors.go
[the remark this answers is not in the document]
pattyshack: i guess i haven't really answered your question.  one solution is to not cast at not.  for example, we could do something like

```
func CastOrWrap(err error) *CodeError {
  coded, ok := err.(*CodedError)
  if ok {
    return coded
  }

  var codeChecker ErrorCodeChecker
  if errors.As(err, &codeChecker) {
    // wrap the error with a new *CodedError with the extract code
  }
  return NewUnknownFailure(err)
}
```

janezpodhostnik (commented):
I like the CodedError change. 
I'm not sure of the internal structure of the coded error. Right now it is necessarily a wrapper for  another error, which might be weird for stuff where there is no other error causing it, its just a failed condition.

I think sorting error codes only makes sense if both failures and normal errors have a code, and I don't think failures need a code, since the user is not supposed to ever see them. We should jut present them in the the order the errors happened and have the actual code be the code of the first coded error in the wrapped errors stack.

Naming suggestions: UserError, FVMError, ExecutionError, but honestly Coded Error sounds fine to me.

pattyshack:
On a high level, the internal structure is a bit weird since I'm trying to preserve most of the existing error vs failure behavior.  My intention is to first swap out the existing error / failure structs with CodedError and CodedFailure, then combine CodedError and CodedFailure (and codedError).  At the end, there will only be a single CodedError struct.

Wrt the []errors inside codedError, the list is only really useful for TransactionInvoker, where we can have multiple different / unwrappable errors but (right now) only return of one them.   Everything else should wrap the errors normally (the error wrapping is essentially a link list).  Asides from the fact that we should return all the errors, I don't feel strongly about how this should be structured.  Let me drop this list for now since it's clearly controversial.  We can figure out the rest later.

pattyshack:
i've dumb down the codedError's internal.  take another look.

janezpodhostnik (approved):

pattyshack:
bors merge

Issue #3104, opened 2022-08-26
[FVM] Refactor Metering


janezpodhostnik:
## Problem Definition

This continues on from https://github.com/onflow/flow-go/issues/3103

After all meter-able quantities are part of the meter, its worth splitting the meter into sub-part, each responsible for metering/limiting its own quantity.

## Definition of Done

- refactor meter into a more modular form where each sub-part of the meter is responsible for a single quantity

PR #3268, opened 2022-09-21
[FVM] Moving event emit size metering logic to Meter interface

Tonix517:
The existing event emit size metering logic in `fvm/environment/event_emitter.go` can also be moved into `Meter` interface.

Changes below are created in individual commits
- Adding new event metering interface in `fvm/meter/meter.go`
- Adding new event metering interface in `State`, `StateTransaction` and `fvm/environment/meter.go` 
   we should use environment Meter interface in environment code to avoid potential dep issue with `meter/meter.go`
- Replacing existing event metering logic in event_emitter.go with the new interface

Test

- [X] existing and new UTs
- [X] run localnet to see debug logs to make sure valid new config values passed in

### DIFF
--- a/fvm/environment/event_emitter.go
+++ b/fvm/environment/event_emitter.go
@@ -103,7 +103,9 @@ func NewEventEmitter(
 }
 
 func (emitter *eventEmitter) Reset() {
-	emitter.eventCollection = NewEventCollection()
+	// TODO: for now we are not resetting meter here because we don't check meter
+	//		 metrics after the first metering failure and when limit is disabled.
+	emitter.eventCollection = NewEventCollection(emitter.meter)
 }
 
 func (emitter *eventEmitter) EventCollection() *EventCollection {
@@ -128,23 +130,16 @@ func (emitter *eventEmitter) EmitEvent(event cadence.Event) error {
 
 	payloadSize := uint64(len(payload))
 
-	// skip limit if payer is service account
-	if emitter.payer != emitter.chain.ServiceAddress() {
-		totalSize := emitter.eventCollection.TotalByteSize() + payloadSize
-		if totalSize > emitter.EventCollectionByteSizeLimit {
-			return errors.NewEventLimitExceededError(
-				totalSize,
-				emitter.EventCollectionByteSizeLimit)
-		}
-	}
-
 	flowEvent := flow.Event{
 		Type:             flow.EventType(event.EventType.ID()),
 		TransactionID:    emitter.txID,
 		TransactionIndex: emitter.txIndex,
-		EventIndex:       emitter.eventCollection.eventCounter,
+		EventIndex:       emitter.eventCollection.TotalEventCounter(),
 		Payload:          payload,
 	}
+
+	// TODO: to set limit to maximum when it is service account and get rid of this flag
+	isServiceAccount := emitter.payer == emitter.chain.ServiceAddress()
 
 	if emitter.ServiceEventCollectionEnabled {
 		ok, err := IsServiceEvent(event, emitter.chain.ChainID())
@@ -152,13 +147,22 @@ func (emitter *eventEmitter) EmitEvent(event cadence.Event) error {
 			return fmt.Errorf("unable to check service event: %w", err)
 		}
 		if ok {
-			emitter.eventCollection.AppendServiceEvent(flowEvent, payloadSize)
+			eventEmitError := emitter.eventCollection.AppendServiceEvent(flowEvent, payloadSize)
+			// skip limit if payer is service account
+			if !isServiceAccount && eventEmitError != nil {
+				return eventEmitError
+			}
 		}
 		// We don't return and append the service event into event collection
 		// as well.
 	}
 
-	emitter.eventCollection.AppendEvent(flowEvent, payloadSize)
+	eventEmitError := emitter.eventCollection.AppendEvent(flowEvent, payloadSize)
+	// skip limit if payer is service account
+	if !isServiceAccount {
+		return eventEmitError
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


fvm/meter/meter.go
```diff
@@ -557,3 +573,23 @@ func GetStorageKeyValueSizeForTesting(
 func (m *Meter) GetStorageUpdateSizeMapForTesting() MeteredStorageInteractionMap {
 	return m.storageUpdateSizeMap
 }
+
+func (m *Meter) MeterEmittedEvent(byteSize uint64) error {
+	m.totalEventEmitBytes += uint64(byteSize)
+	m.eventCounter++
+
+	if m.totalEventEmitBytes > m.eventEmitByteLimit {
```
Tonix517, line 581: in current use case the `enforceLimit` is handled at `fvm/environment/meter.go`, so I'm not adding an extra flag here.

fvm/environment/event_emitter.go
```diff
@@ -115,37 +112,38 @@ func (emitter *eventEmitter) EmitEvent(event cadence.Event) error {
...
-	// skip limit if payer is service account
-	if emitter.payer != emitter.chain.ServiceAddress() {
-		totalSize := emitter.eventCollection.TotalByteSize() + payloadSize
-		if totalSize > emitter.eventCollectionByteSizeLimit {
-			return errors.NewEventLimitExceededError(
-				totalSize,
-				emitter.eventCollectionByteSizeLimit)
-		}
-	}
-
```
Tonix517, line 127: A little behavior change here: original behavior actually did a pre-metering. It checked sizes _before_ emit an event. The new behavior utilizes the error returned from new interface, and returns error immediately. Please confirm if this will expose some risks

fvm/environment/event_emitter.go
```diff
@@ -82,15 +80,14 @@ func NewEventEmitter(
...
-		eventCollectionByteSizeLimit:  eventCollectionByteSizeLimit,
 	}
 
 	emitter.Reset()
 	return emitter
 }
 
 func (emitter *eventEmitter) Reset() {
-	emitter.eventCollection = NewEventCollection()
+	emitter.eventCollection = NewEventCollection(emitter.meter)
```
pattyshack, line 90: This change seems wrong.  In particular, this no longer reset the totalByteSize / eventCounter to zero.  This (potentially) prevent the environment from being reusable since emitting an event after Reset will automatically fail.

pattyshack (changes requested):

fvm/environment/event_emitter.go
Tonix517: Good catch - to make sure that EventEmitter still uses the Meter instance from Context, a `Reset()` interface seems to be necessary in `Meter`, in which we may need to distinguish which part of metrics to reset, so that when we do reset in event_emitter, other metrics for storage\computation etc. can remain the same. I'd rather not creating new Meter instance here, so in case MeterParameter changes from Context can also be reflected automatically. WDYT

pattyshack: it's kind of weird to have a Reset in meter that reset some, but not all, limits.  We should either reset consistently, or just keep the event limits as a separate thing.

Tonix517: given #3104 is also on our list, maybe having a separate meter for event emitter is viable solution for now, which can make things easier when we get to splitting existing meter into sub-parts afterwards. @janezpodhostnik

pattyshack: o, you know what, i totally missed the enabledLimits check in environment/meter.go .   I think what you're doing is actually ok, as long as we're happy with not checking the emitted event size when limits are disabled.  I don't have any objections to your changes, but we should double check with @janezpodhostnik cuz i don't really know why we have this limit in the first place =P

fvm/environment/event_emitter.go
```diff
@@ -115,37 +112,38 @@ func (emitter *eventEmitter) EmitEvent(event cadence.Event) error {
...
+			}
 		}
 		// We don't return and append the service event into event collection
 		// as well.
 	}
 
-	emitter.eventCollection.AppendEvent(flowEvent, payloadSize)
+	eventEmitError := emitter.eventCollection.AppendEvent(flowEvent, payloadSize)
+	// skip limit if payer is service account
+	if !isServiceAccount && eventEmitError != nil {
```
janezpodhostnik, line 143:
```suggestion
	if !isServiceAccount {
```

fvm/meter/meter.go
```diff
@@ -557,3 +573,23 @@ func GetStorageKeyValueSizeForTesting(
...
+		return errors.NewEventLimitExceededError(
+			m.totalEventEmitBytes,
+			m.eventEmitByteLimit)
+	}
+	return nil
+}
+
+func (m *Meter) TotalEventEmitBytes() uint64 {
```
janezpodhostnik, line 589:
```suggestion
func (m *Meter) TotalEmittedEventBytes() uint64 {
```

fvm/environment/event_emitter.go
janezpodhostnik: This change is fine, since we throw away all events in case of event limit hit, so it doesn't really matter if we add the last one to the list or not.

fvm/environment/event_emitter.go
janezpodhostnik: Yeah since we disable the metering after first failure, for fee deduction, this should be fine. We don't reset the meter on other stuff that we reset either.

The limit is here because emitting to many events causes problems with getting the block event list through the API. In the past blocks with large event lists were unfetchable. Also I think because of storage.

fvm/environment/event_emitter.go
janezpodhostnik, line 143: In a future PR we should consider getting rid of this `isServiceAccount` check here, and instead set the event limit to maxuint if this is the service account.

janezpodhostnik (approved):

fvm/environment/event_emitter.go
pattyshack: +1.  we can extend ShouldDisableMemoryAndInteractionLimits to ShouldDisableNonComputeLimit or something

pattyshack (approved):

fvm/environment/event_emitter.go
Tonix517: addressed

Tonix517:
bors merge

Tonix517:
bors retry

PR #3269, opened 2022-09-22
Perform stricter unexpected nested transaction checks.

pattyshack:
We can now perform stricter checks since nested transaction restart / commit no longer fail due to metering.

1. Halt execution if successfully executed transaction has unexpected nested transaction.
2. Halt execution if failed transaction cannot restart


fvm/transactionInvoker.go
```diff
@@ -49,31 +49,41 @@ func (i TransactionInvoker) Process(
...
+				proc.Logs = make([]string, 0)
+				proc.Events = make([]flow.Event, 0)
+				proc.ServiceEvents = make([]flow.Event, 0)
 			}
-
-			msg := "transaction has unexpected nested transactions"
-			ctx.Logger.Error().
-				Msg(msg)
-
-			proc.Err = errors.NewFVMInternalErrorf(msg)
```
pattyshack, line 70: I'm going to remove this error/error code in a different PR.  This is currently classified as an error rather than a failure.  This should not be the case.

fvm/transactionInvoker.go
```diff
@@ -49,31 +49,41 @@ func (i TransactionInvoker) Process(
 
 	var modifiedSets programsCache.ModifiedSetsInvalidator
 	defer func() {
-		// based on the contract and frozen account updates we decide how to
-		// clean up the programs for failed transactions we also do the same as
-		// transaction without any deployed contracts
-		programs.Cleanup(modifiedSets)
-
 		if sth.NumNestedTransactions() > 1 {
```
Tonix517, line 52: for my understanding: when there are still nested tx here, does it mean some nested tx was not popped bcz of error?

pattyshack: in case of processErr == nil, it means that we forgot to call commit somewhere.  in case of processErr != nil, it potentially mean that we trampoline the control flow due to the error

pattyshack: added some more comment to the processErr == nil branch

janezpodhostnik (approved):

fvm/transactionInvoker.go
```diff
@@ -49,31 +49,42 @@ func (i TransactionInvoker) Process(
...
+					processErr = fmt.Errorf(
+						"cannot restart nested transaction on error: %w "+
+							"(original txError: %v)",
+						err,
+						processErr)
+					return
+				}
+
+				ctx.Logger.Warn().
+					Msg("transaction has unexpected nested transactions")
```
zhangchiqing, line 74:
```suggestion
					Msg("transaction had unexpected nested transactions, which have been restarted")
```

zhangchiqing (approved):

fvm/transactionInvoker.go
```diff
@@ -49,31 +49,42 @@ func (i TransactionInvoker) Process(
...
-			if err != nil {
+			if processErr == nil {
+				// This is a fvm internal programming error.  We forgot to
+				// call Commit somewhere in the control flow.  We should halt.
 				processErr = fmt.Errorf(
-					"cannot restart nested transaction: %w",
-					err,
-				)
+					"successfully executed transaction has unexpected " +
+						"nested transactions.")
```
zhangchiqing, line 58: wonder if useful to log more context just for debugging purpose. For instance, `nextedTxnId`.

pattyshack: the nestedTxnId is opaque (it holds a pointer) doesn't provide much debugging value.

Looks like we don't log the TxID in general.  I'll look into logging that consistent (in a separate PR)

pattyshack:
bors merge

PR #3270, opened 2022-09-22
Khalil/1789 Update debug logs and engine metrics

kc1116:
This PR makes the following updates to debug logs and engine metrics
- Add originID to the labels for messages received/handled 
- Add debug logs that log some request and response info in the entity provider/requester engines


engine/collection/compliance/engine.go
```diff
@@ -109,7 +109,7 @@ func NewEngine(
 			Match: func(msg *engine.Message) bool {
 				_, ok := msg.Payload.(*messages.ClusterBlockProposal)
 				if ok {
-					core.metrics.MessageReceived(metrics.EngineClusterCompliance, metrics.MessageClusterBlockProposal)
+					core.metrics.MessageReceived(metrics.EngineClusterCompliance, metrics.MessageClusterBlockProposal, msg.OriginID)
```
peterargue, line 112: after digging a bit, adding the nodeID as a label will significantly increase the number of time series that prometheus would need to track. for now, let's stick with just including the nodeIDs in the logs and extract metrics from there.

engine/common/provider/engine.go
```diff
@@ -163,16 +168,33 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
 	// try to retrieve each entity and skip missing ones
 	entities := make([]flow.Entity, 0, len(req.EntityIDs))
 	entityIDs := make([]flow.Identifier, 0, len(req.EntityIDs))
+	seen := make(map[flow.Identifier]struct{})
 	for _, entityID := range req.EntityIDs {
+		// skip requesting duplicate entity IDs
+		if _, ok := seen[entityID]; ok {
+			e.log.Debug().
```
peterargue, line 175: any idea if this happens with the node software as it is today? I think we should return an error in this case and flag this as suspected malicious behavior

engine/common/requester/engine.go
```diff
@@ -382,15 +383,20 @@ func (e *Engine) dispatchRequest() (bool, error) {
 	}()
 
 	e.metrics.MessageSent(e.channel.String(), metrics.MessageEntityRequest)
+	e.log.Debug().
+		Str("sender_node_id", e.me.NodeID().String()).
```
peterargue, line 387: we probably don't need to print our own node_id. it's included in messages by default

engine/common/provider/engine.go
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/f22aa498026edae0ddca4be1c05f28eb4024a6c2

engine/collection/compliance/engine.go
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/fad2b8f9bf81df6a7ca3fa9f33b277a6782a7182

engine/common/requester/engine.go
```diff
@@ -404,6 +410,11 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
 }
 
 func (e *Engine) onEntityResponse(originID flow.Identifier, res *messages.EntityResponse) error {
```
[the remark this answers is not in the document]
kc1116: There is no efficient way to check for unrequested entities, we are skipping them anyway [here](https://github.com/onflow/flow-go/blob/5d2273c199421733efcc5b2a477dfbb6ae085901/engine/common/requester/engine.go#L457) . We just assume that the entity was returned in another request rather than it "had not been requested at all"

engine/common/requester/engine.go
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/1b7cdf8fad7f2603c66d5e96f613834dcad24e78

engine/common/provider/engine.go
peterargue: why? wouldn't we want to ignore/fail any suspected malicious behavior instead of continuing on a best effort basis? It seems to me this would either indicate there's a bug in the implementation, or the peer is attempting something malicious.

kc1116: Yes your right since this is only possible during an attempt to do something malicious

peterargue (approved):

engine/common/requester/engine.go
```diff
@@ -404,6 +409,9 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
...
+	log := e.log.With().Str("origin_id", originID.String()).Uint64("nonce", res.Nonce).Logger()
```
yhassanzadeh13, line 412:
```suggestion
	lg := e.log.With().Str("origin_id", originID.String()).Uint64("nonce", res.Nonce).Logger()
```
to avoid any variable and package name collision.

engine/common/provider/engine_test.go
```diff
@@ -165,6 +165,76 @@ func TestOnEntityRequestPartial(t *testing.T) {
...
+
+	retrieve := func(entityID flow.Identifier) (flow.Entity, error) {
+		entity, ok := entities[entityID]
+		if !ok {
+			return nil, storage.ErrNotFound
+		}
+		return entity, nil
+	}
+
+	final := &protocol.Snapshot{}
```
yhassanzadeh13, line 192:
```suggestion
	final := protocol.NewSnapshot(t)
```
Let's always use the default provided `testify` mock constructor that checks for expectations automatically.

engine/common/provider/engine.go
```diff
@@ -144,6 +144,11 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
 }
 
 func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityRequest) error {
+	log := e.log.With().Uint64("nonce", req.Nonce).Str("origin_id", originID.String()).Logger()
```
yhassanzadeh13, line 147:
```suggestion
	lg := e.log.With().Uint64("nonce", req.Nonce).Str("origin_id", originID.String()).Logger()
```

yhassanzadeh13 (approved):

engine/common/requester/engine.go
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/3eb42cb65e1f6f08f51edcfdd93c4e1d466c8608

engine/common/provider/engine_test.go
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/226016109bb39372228bb31964a928ecafac5f1e

engine/common/provider/engine.go
kc1116: https://github.com/onflow/flow-go/pull/3270/commits/c9cb5f2f483e2e2a72c1794186a3a339b80d6e27

kc1116:
bors merge

peterargue:
bors merge

PR #3273, opened 2022-09-23
Checkpointing V6 - support concurrent checkpoint encoding/decoding

### LINKED ISSUE #3075: [EN Performance] Determine if checkpoint file should be split up and processed in parallel to resolve several issues
## Problem

Checkpoint file is around 160GB and growing.  Also, state extraction during mainnet spork should be faster.  Lastly, we currently evict entire 160GB checkpoint file from Linux page cache when the file is closed.

Updates #1744 

## The Proposed Solution

Determine if we should split up the checkpoint file and process portions in parallel to:
- Speedup checkpointing.
- Speedup state extraction during mainnet spork.
- Eliminate the need to evict segments of 160GB checkpoint file from Linux page cache during use.

PR #3050 (merged on Aug 23 and not deployed to August Spork) set the foundation for parallel processing of subtries by splitting up the mtrie into 16 subtries.   It will reduce peak RAM use (max RSS) by dozens of GB once deployed.   Parallel processing in future PR will use up of some or all of that memory reduction.

## Caveats

Considerations include third parties who currently read and parse the checkpoint file with their own software.

zhangchiqing:
Closes #3075 

This PR implements the checkpointing V6. 

Checkpointing Version 6 splits the single checkpoint file into 18 files in total. The main benefits are:
- The benefit of splitting the checkpoint file is to support concurrent writes to multiple sub files which speeds up checkpoint generation, and concurrent reads which speeds up reading checkpoint.
- V6 is benefited from V5, where it builds the sub trees first to be encoded, which built the ground for allowing concurrent processing.

See complete design in this doc:
https://www.notion.so/dapperlabs/Checkpoint-V6-8c7b97937da54c5b9e6c18b5b4598f2e

Comparison between V5 and V6 using latest mainnet19 data snapshot:
- checkpoint writing is reduced from `16mins` to `3mins`, 5.3 times faster
- checkpoint reading is reduced from `12mins` to `2mins`, 6 times faster

This is a feature branch. There are more TODO items to be done in separate PRs. Once this PR is approved, I will close this PR until all TODO items are done, then I will re-open this PR, and merge to master.


ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+
+	logger.Info().Msgf("checkpoint file has been successfully stored at %v/%v", outputDir, outputFile)
+
+	return nil
+}
+
+// 		1. version
+//		2. checksum of each part file (17 in total)
+// 		3. checksum of the main file itself
+func storeCheckpointSummary(
```
SaveTheRbtz, line 90: Does this file need to have a custom binary format?  Maybe we can just store a json here?

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+	defer func() {
+		closeErr := closable.Close()
+		// Return close error if there isn't any prior error to return.
+		if err == nil {
+			err = closeErr
+		}
+	}()
+
+	writer := NewCRC32Writer(closable)
```
SaveTheRbtz, line 167: Not sure how much crc32 is useful here: it is usually used for very small quantities of data.  Is there a downside of using a strong sha2 checksum here?

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+
+type resultStoringSubTrie struct {
+	Index     int
+	Roots     map[*node.Node]uint64 // node index for root nodes
+	NodeCount uint64
+	Checksum  uint32
+	Err       error
+}
+
+// StoreCheckpointV6 stores checkpoint file into a main file and 17 file parts.
```
SaveTheRbtz, line 32: Since it is concurrent now, you can consider adding tracing here so you can visualize store/load in grafana.  Thisi may simplify finding potential bottlenecks.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+				Err:       err,
+			}
+			close(resultCh)
+		}(i, subTrieRoot)
+		resultChs = append(resultChs, resultCh)
+	}
+
+	logger.Info().Msgf("subtrie roots have been stored")
+
+	results := make(map[*node.Node]uint64, 1<<(subtrieLevel+1))
```
SaveTheRbtz, line 259: Did you mean `subtrieCount+1`?

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	nodeCounter := uint64(1)
+
+	logging := logProgress(fmt.Sprintf("storing %v-th sub trie roots", i), estimatedSubtrieNodeCount, logger)
+	scratch := make([]byte, 1024*4)
+	for _, root := range roots {
+		// Note: nodeCounter is to assign an global index to each node in the order of it being seralized
+		// into the checkpoint file. Therefore, it has to be reused when iterating each subtrie.
+		// storeUniqueNodes will add the unique visited node into traversedSubtrieNodes with key as the node
+		// itself, and value as n-th node being seralized in the checkpoint file.
+		nodeCounter, err = storeUniqueNodes(root, traversedSubtrieNodes, nodeCounter, scratch, writer, logging)
```
SaveTheRbtz, line 374: How compressible is this data?  Can we try lz4 or zstd here to see if there is an upside in compressing this?

ledger/complete/ledger.go
```diff
@@ -434,14 +434,15 @@ func (l *Ledger) ExportCheckpointAt(
...
-	if err != nil {
-		return ledger.State(hash.DummyHash), fmt.Errorf("failed to create a checkpoint writer: %w", err)
-	}
 
 	l.logger.Info().Msg("storing the checkpoint to the file")
 
-	err = realWAL.StoreCheckpoint(writer, newTrie)
+	err = os.MkdirAll(outputDir, os.ModePerm)
+	if err != nil {
+		return ledger.State(hash.DummyHash), fmt.Errorf("could not create output dir %v: %w", outputDir, err)
```
SaveTheRbtz, line 442: A general nit: you are using `%v` a lot which may prevent static analyzer from detecting a class of bugs when a programmer accidentally passes a different variable than they actually wanted to log.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+func storeSubTrieConcurrently(
+	subtrieRoots [subtrieCount][]*node.Node,
+	estimatedSubtrieNodeCount int,
+	outputDir string,
+	outputFile string,
+	logger *zerolog.Logger,
+) (
+	map[*node.Node]uint64, // node indices
```
SaveTheRbtz, line 232: nit: you may consider a struct here, given that you need to comment them anyway.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+		return nil, err
+	}
+	if utilsio.FileExists(fullPath) {
+		return nil, fmt.Errorf("checkpoint file for %v-th sub trie %s already exists", index, fullPath)
+	}
+
+	return createClosableWriter(dir, logger, subTriesFileName, fullPath)
+}
+
+func createClosableWriter(dir string, logger *zerolog.Logger, fileName string, fullPath string) (io.WriteCloser, error) {
```
SaveTheRbtz, line 303: nit: you likely don't need to thread dir through all function calls here: you can just use `filepath.Dir()` here.  File moves are guaranteed to be atomic only within a directory anyway.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+// ReadCheckpointV6 reads checkpoint file from a main file and 17 file parts.
+// the main file stores:
+// 		1. version
+//		2. checksum of each part file (17 in total)
+// 		3. checksum of the main file itself
+// 	the first 16 files parts contain the trie nodes below the subtrieLevel
+//	the last part file contains the top level trie nodes above the subtrieLevel and all the trie root nodes.
+func ReadCheckpointV6(dir string, fileName string) ([]*trie.MTrie, error) {
```
SaveTheRbtz, line 22: Since it is concurrent now, you can consider adding tracing to the whole state load/store codepath so you can visualize in grafana (for example right now it would likely show that trie construction is the bottleneck).

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	"github.com/onflow/flow-go/ledger/complete/mtrie/trie"
+	utilsio "github.com/onflow/flow-go/utils/io"
+)
+
+const subtrieLevel = 4
+const subtrieCount = 1 << subtrieLevel
+
+type NodeEncoder func(node *trie.MTrie, index uint64, scratch []byte) []byte
+
+type resultStoringSubTrie struct {
```
ramtinms, line 24: we might find a better naming for this one like `SubTrieStorable`

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/ledger/complete/mtrie/flattener"
+	"github.com/onflow/flow-go/ledger/complete/mtrie/node"
+	"github.com/onflow/flow-go/ledger/complete/mtrie/trie"
+	utilsio "github.com/onflow/flow-go/utils/io"
+)
+
+const subtrieLevel = 4
+const subtrieCount = 1 << subtrieLevel
```
ramtinms, line 20: Do we need to handle the very rare case of trie not having 4 levels and only have very small number of registers? maybe just returning error in that case is good enough

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	if err != nil {
+		return fmt.Errorf("could not store checkpoint summary: %w", err)
+	}
+
+	logger.Info().Msgf("checkpoint file has been successfully stored at %v/%v", outputDir, outputFile)
+
+	return nil
+}
+
+// 		1. version
```
ramtinms, line 87: godoc seems malformed

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+
+	// write version
+	version := encodeVersion(MagicBytes, VersionV6)
+	writer := NewCRC32Writer(closable)
+	_, err = writer.Write(version)
+	if err != nil {
+		return fmt.Errorf("cannot write checksum summary header: %w", err)
+	}
+
+	// TODO: write checksums
```
ramtinms, line 117: I think you can remove this TODO

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+
+// 17th part file contains:
+// 1. checkpoint version TODO
+// 2. checkpoint file part index TODO
+// 3. subtrieNodeCount TODO
+// 4. top level nodes
+// 5. trie roots
+// 6. node count
+// 7. trie count
+// 6. checksum
```
ramtinms, line 145: you might need to update this part and remove TODOs

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("cannot write CRC32 checksum to top level part file: %w", err)
+	}
+	return checksum, nil
+}
+
+func createSubTrieRoots(tries []*trie.MTrie) [subtrieCount][]*node.Node {
+	var subtrieRoots [subtrieCount][]*node.Node
+	for i := 0; i < len(subtrieRoots); i++ {
+		subtrieRoots[i] = make([]*node.Node, len(tries))
```
ramtinms, line 203: hmm, not sure why we use `len(tries)` here ? can't we just use it as a capacity instead of actuall allocation considering most of the sub-tries would be shared by a few tries ?

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	}
+	return subtrieRoots
+}
+
+func estimateSubtrieNodeCount(tries []*trie.MTrie) int {
+	if len(tries) == 0 {
+		return 0
+	}
+	estimatedTrieNodeCount := 2*int(tries[0].AllocatedRegCount()) - 1
```
ramtinms, line 221: I would suggest we take the last trie and use the allocatedRegCount from there considering its most likely have more registers than the trie with 0 index.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: In V5, there is only a single file named "root.checkpoint", for instance, and this summary file in V6 is essentially the same file as that. The name summary is not the best name, I changed it into header in another PR.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: We have been using crc32, I'm not sure how much difference they are, changing it would be out of scope of this PR.

But thanks for pointing this out. Worth to explore the difference.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: maybe 50%, it's a trade-off to compress the data, as we have to de-compress it. 

During spork, we used to compress them, but it ended up taking longer than not compressing

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: TODOs are done and cleaned in a different PR

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: TODOs are done and cleaned in a different PR

ledger/complete/wal/checkpoint_v6_writer.go
SaveTheRbtz: Interesting!  Lz4 is quite capable at decompression (with C decoder reaching speeds of ~5Gbyte/s).  Did we use [ConcurrencyOption](https://github.com/pierrec/lz4/blob/v4.1.17/options.go#L114)?

ledger/complete/wal/checkpoint_v6_writer.go
SaveTheRbtz: My question is mostly why do we need to deal with all that custom binary encoding/decoding instead of just creating a struct that specifies a "summary" of the checkpoint and just `Marshal`/`Unmarshal` it in any format that we want.  Is custom marshaling/unmarshaling code worth it?

ledger/complete/wal/checkpoint_v6_writer.go
SaveTheRbtz: nevermind the checksumming point-- i do see the benefit of a single end-to-end checksum instead of many per-frame ones.

ledger/complete/wal/checkpoint_v6_reader.go
zhangchiqing: Yeah, I will be adding logs in this function in future PR. I think logs would be more useful than tracing. 

Since we will only read the checkpoint once on every restart, it's OK to print info logs with each key steps during the read.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: I didn't change this part, it's [the same as before ](https://github.com/onflow/flow-go/blob/master/ledger/complete/wal/checkpointer.go#L355). Maybe might know why it's this value.

This also doesn't matter too much, we are allocating memory for the root nodes at subtrie level, which usually only have at most hundreds of nodes.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: It's actually the result of storing sub trie, not the subtrie itself to be stored. 

The result could be succeeded or failed, as you can see the result also contains `Err`.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: This is already handled. In this case, there would be no subtrie nodes, we would still produce part files, but they have no subtrie nodes. The nodes are just top level nodes.

ledger/complete/wal/checkpointer.go
```diff
@@ -589,6 +655,8 @@ func readCheckpoint(f *os.File) ([]*trie.MTrie, error) {
 		return readCheckpointV4(f)
 	case VersionV5:
```
pattyshack, line 656: why do we need to keep v1-v4? since spork forces all binaries to upgrade, these are just deadcode / bugs when to happen? would be great if we can delete these (and their associate functions) first since I have to understand what the system currently does in order to provide useful review feedback.

ledger/complete/wal/concurrent_checkpointer_test.go
```diff
@@ -0,0 +1 @@
+package wal
```
pattyshack, line 1: this does nothing?

cmd/util/cmd/checkpoint-list-tries/cmd.go
```diff
@@ -29,7 +30,8 @@ func init() {
 func run(*cobra.Command, []string) {
 
 	log.Info().Msgf("loading checkpoint %v", flagCheckpoint)
-	tries, err := wal.LoadCheckpoint(flagCheckpoint, &log.Logger)
+	dir, file := filepath.Split(flagCheckpoint)
+	tries, err := wal.LoadCheckpoint(dir, file, &log.Logger)
```
pattyshack, line 34: could you break out the LoadCheckpoint changes into a different PR (we can probably get that merge now, but the reviewing the rest of the PR will take longer).

pattyshack (commented):
it's going to take me a while to review this since i don't have any context.  it would be great if you can clean up the code base a bit (v1 - v4 are deadcode?) since I'm not sure what's actually relevant.

ledger/complete/wal/checkpointer.go
zhangchiqing: yeah, we can probably delete v1-v4. we can do in separate PR. just to double check with @fxamacker @ramtinms

ledger/complete/wal/checkpointer.go
```diff
@@ -314,34 +320,57 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
 		return fmt.Errorf("cannot write checkpoint header: %w", err)
 	}
 
-	// Nodes are serialized in two steps to save memory.
```
pattyshack, line 317: looks like most of the changes to this file is actually v5 specific? maybe break this into a different PR

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,559 @@
...
+		_, err := writer.Write(encTrie)
+		if err != nil {
+			return fmt.Errorf("cannot serialize trie: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func storeFooter(topLevelNodesCount uint64, rootTrieCount uint16, writer io.Writer) error {
```
pattyshack, line 491: encoding the header using proto or json (or whatever std we use here) instead of hand packing the bytes.

also, can you structure them as objects instead of global functions?

pattyshack (commented):
why do we need to hardcode this to 16 partitions?  we should number of partitions via a flag.  you only need to keep track this number as part of the manifest.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,559 @@
...
+	outputDir string,
+	outputFile string,
+	logger *zerolog.Logger,
+) error {
+	closable, err := createWriterForCheckpointSummary(outputDir, outputFile, logger)
+	if err != nil {
+		return fmt.Errorf("could not store checkpoint summary: %w", err)
+	}
+	defer func() {
+		closeErr := closable.Close()
```
pattyshack, line 107: instead of closing closable here, can you close the crc writer instead?

```
        checksum := writer.Crc32()
	_, err = writer.Write(encodeCRC32Sum(checksum))
	if err != nil {
		return fmt.Errorf("cannot write CRC32 checksum to checkpoint summary: %w", err)
	}
```
Should be part of crc writer's Close method

zhangchiqing: `Crc32Writer` wraps a io.Writer interface, which doesn't have `Close` method, so itself also don't have `Close` method. In other words, `Crc32Writer` doesn't care whether the writer is closable or not.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: >encoding the header using proto or json (or whatever std we use here) instead of hand packing the bytes.

The header has always been hand packed bytes, so that we can read the same format and recognize the version number encoded in it. And using the same format allows me to do round trip testing to confirm converting to v6 then converting back to v5 would produce the same result as converting to v5 directly. The round trip testing helps verifying the v6 algorithm 

>also, can you structure them as objects instead of global functions?
with functions, it's clear to see what it depends on, easy to write tests. Not sure what's the benefit of using objects.

ledger/complete/wal/checkpointer.go
zhangchiqing: That's right. 

I forgot to mention, this PR is [on top of this PR actually](https://github.com/onflow/flow-go/pull/3139/files), which includes these v5 specific changes. Better to review that first.

fxamacker (changes requested):
Performance looks promising!  PR text should add the line "Closes #3075".

There are some remaining items that need to be implemented.  Some are commented in the code by you, and some are mentioned in my comments.  I put them here so we can track better. 

- [x] encode versions in all checkpoint files

- [x] encode subtrie checksums in summary file

- [x] remove all created subtrie files on error during checkpoint serialization

- [ ] remove all subtrie files when we clean up checkpoint files by Compactor

- [ ] modify Compactor to create checkpoint v6

- [ ] make number of goroutines configurable for checkpoint serialization and deserialization

- [ ] add benchmarks for checkpoint serialization especially memory usage and allocations 

- [x] perform some basic sanity checks before starting goroutines to process subtrie files, such as: 
  * check number of subtrie files matches number of checksums in summary file
  * check subtrie files have no gap (missing files)
  * check version in subtrie file  
  * check checksum in subtrie file against checksum in summary file

Some general suggestions:

- It might be useful to make number of subtries (level) configurable for flexibility.

- We will eventually replace CRC-32 (issue #3302).   Adding concurrency to checkpointing, compression, and replacing CRC-32 are postponed TODOs left over from PR #1944.

- We can combine checkpoint of top level nodes with summary file.  Subtrie files are independent of each other and can be processed in parallel.  However, top level nodes file depends on subtrie files and it needs to be processed differently from subtrie files.  I think it is confusing to have subtrie files at the same level as top level nodes file.

ledger/complete/wal/checkpoint_v6_writer.go
pattyshack: we own Crc32Writer, so why can't we change it to take a io.WriteCloser instead of just a io.Writer?

ledger/complete/wal/checkpoint_v6_writer.go
pattyshack: i'm not against hand packing the version (and the header content size), but the header content does not need to be hand packed.

ledger/complete/wal/checkpointer.go
```diff
@@ -314,34 +320,57 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
...
+	// Multiple tries might have shared nodes at higher level, However, we don't want to
+	// seralize duplicated nodes in the checkpoint file. In order to deduplicate, we build
+	// a map from unique nodes while iterating and seralizing the nodes to the checkpoint file.
+	//
+	// The map for deduplication contains all the trie nodes, which uses a lot of memory.
+	// In fact, we don't have to build a map for all nodes, since there are certain nodes whose children
+	// will never have shared nodes. Parent nodes could only share child nodes if and only if parent nodes
+	// are on the same path. In other words, parent nodes on different path won't share nodes.
```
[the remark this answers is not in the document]
zhangchiqing: Thanks for these suggestions. 

I forgot to mention, this PR is on top of [this PR](https://github.com/onflow/flow-go/pull/3139) . 

I will update your comments there.

ledger/complete/wal/checkpointer.go
```diff
@@ -314,34 +320,57 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
...
+	// With this idea in mind, the seralization can be done in two steps:
 	// 1. serialize nodes in subtries (tries with root at subtrieLevel).
 	// 2. serialize remaining nodes (from trie root to subtrie root).
-	// Using subtries saves memory because we only need to track unique nodes
-	// of subtries instead of unique nodes of the entire trie.
-	// NOTE: nodes are serialized in order of Descendents-First-Relationship.
-
+	// For instance, if there are 3 top tries, and subtrieLevel is 4, then there will be
+	// 	(2 ^ 4) * 3 = 48 subtrie root nodes at level 4.
+	// Then step 1 will seralize the 48 subtrie root nodes into the checkpoint file, and
```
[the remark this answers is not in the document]
zhangchiqing: I was trying to highlight that it's the root nodes of the subtries. Rather than all the child nodes in subtries.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	f, err := os.Open(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("could not open file %v: %w", filepath, err)
+	}
+	defer func(f *os.File) {
+		f.Close()
+	}(f)
+
+	nodesCount, err := readSubTriesNodeCount(f)
```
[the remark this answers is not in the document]
zhangchiqing: https://github.com/onflow/flow-go/pull/3319/commits/30dc86c63f1fb4fdd667646992443ce65066d3b6

ledger/complete/wal/checkpointer.go
fxamacker: I understand but I think we may need to handle decoding of all previous checkpoint versions.  For example, DPS or some third-party software might need to load old checkpoint file formats.

pattyshack: third party software can use older version of flow-go to read old checkpoints?  also, the code is still in git history ...

ledger/complete/wal/checkpoint_v6_writer.go
fxamacker: Footer is hand packed using fix sized so it can always be read at same offset from EOF.  Using a codec can make the offset from EOF a nondeterministic distance (depending on the format and/or codec settings used).

Footer contains data that need to be read before processing main file content.

I think subtrie file header is hand packed because it just contains magic bytes + version.  I have some thoughts about summary file but that's a separate topic from footers and headers.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+	// TODO: read the main file and check the version
+	// TODO: read the checksum of 17 file parts
+
+	headerPath := filePathHeader(dir, fileName)
+	subtrieChecksums, topTrieChecksum, err := readHeader(headerPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not read header: %w", err)
+	}
+
+	subtrieNodes, err := readSubTriesConcurrently(dir, fileName, subtrieChecksums)
```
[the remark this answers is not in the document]
zhangchiqing: [Implemented here](https://github.com/onflow/flow-go/pull/3321)

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: [Will be tracked and done by this issue](https://github.com/onflow/flow-go/issues/3302)

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: @fxamacker why isn't `subtrieCount * len(subtrieRoots[0])`?

If there is only one trie, since there are at most subtrieCount number of nodes at level `subtrieLevel`, then for `subtrieRoots[0]` number of tries, there will be `subtrieCount * subtrieRoots[0]` number of nodes at most.

The number of top level nodes is nearly the same as total number of nodes at subtrie level. right?

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+		return nil, err
+	}
+	f, err := os.Open(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("could not open file %v: %w", filepath, err)
+	}
+	defer func(f *os.File) {
+		f.Close()
+	}(f)
+
```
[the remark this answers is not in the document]
fxamacker: >  since the part files don't have magic code and version, then you won't read the part file as checkpoint header file of any version by mistake.

I think different file types should contain their own unique magic id.  This helps prevent accidental collisions (a part file's data being equal to checkpoint header file's magic id by coincidence).

Also, including a version number with the magic id is useful for detecting out-of-sync versions during development/testing.

We can use a tiny sanity-check header in part files to:
- know it's a part file (magic ID for checkpoint part file)
- know version of part file format (without relying on other files in case they become out of sync)

When handling a collection of related files, we want info that can help us debug more quickly unless that is cost prohibitive.  Adding a tiny header with unique magic id + version to identify file type is cheap.

E.g. during development/testing, it would take longer to debug file checksum mismatch than a file header showing file format versions are out of sync.

ledger/complete/wal/checkpoint_v6_writer.go
fxamacker: @zhangchiqing Although what you say is right about this specific block of code, returned `results` map is used by other function which adds more elements.

In this function `storeSubTrieConcurrently()`, `results` only contains subtrie root nodes for all tries, which is `subtrieCount * subtrieRoots[0]` number of nodes.

However, its caller `StoreCheckpointV6()` [passes returned `results` map to `storeTopLevelNodesAndTrieRoots()` as function parameter](https://github.com/onflow/flow-go/blob/db755dea37553d61ba1ca6432e1cc2c3142b5f51/ledger/complete/wal/checkpoint_v6_writer.go#L58-L75).

`storeTopLevelNodesAndTrieRoots()` continues to add top level nodes (from level 0 to including level 3) to this map.  So this map contains top level nodes (from 0 to including level 4) for all tries at the end.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+func filePathSubTries(dir string, fileName string, index int) (string, string, error) {
+	if index < 0 || index > 15 {
+		return "", "", fmt.Errorf("index must be between 1 to 16, but got %v", index)
+	}
+	subTrieFileName := fmt.Sprintf("%s.%03d", fileName, index)
+	return path.Join(dir, subTrieFileName), subTrieFileName, nil
+}
+
+func filePathTopTries(dir string, fileName string) (string, string) {
```
zhangchiqing, line 58: >We can combine checkpoint of top level nodes with summary file. Subtrie files are independent of each other and can be processed in parallel. However, top level nodes file depends on subtrie files and it needs to be processed differently from subtrie files. I think it is confusing to have subtrie files at the same level as top level nodes file.

@fxamacker Let's discuss here about your comments regarding merging top level nodes to summary file.

One benefit of keeping them separate is that, you can [verify the checksums of subtrie files in the summary file](https://github.com/onflow/flow-go/pull/3319/files#diff-2ff6507ee260a1e37156b4cae4cc259e7688bbd0007993b20fa193824073105aR116) before using them to [verify the checksum of each sub trie part file](https://github.com/onflow/flow-go/pull/3319/files#diff-2ff6507ee260a1e37156b4cae4cc259e7688bbd0007993b20fa193824073105aR203). 

Otherwise, I would be using unverified data (subtrie checksums in the summary file) to verify other data (subtrile part file's checksum), which might cause the error message to be confusing. 

Of course, I could verify the summary file checksum first before using those subtrie checksums , but then I will have to read all top level trie nodes, however, since I won't use those top level trie nodes until after processing all subtrie nodes, I would either have to buffer those top level tries data or read them again. 

That's why I chose to have the top level tries nodes stored in a separate file.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func filePathHeader(dir string, fileName string) string {
+	return path.Join(dir, fileName)
+}
+
+func filePathSubTries(dir string, fileName string, index int) (string, string, error) {
+	if index < 0 || index > 15 {
+		return "", "", fmt.Errorf("index must be between 1 to 16, but got %v", index)
+	}
```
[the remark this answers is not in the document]
zhangchiqing: I think it's good to keep this check inside this function in case the caller forgot to call it, as this function is being used in multiple places, such as [the check here ](https://github.com/onflow/flow-go/pull/3321/files#diff-2ff6507ee260a1e37156b4cae4cc259e7688bbd0007993b20fa193824073105aR102)and[ the test case here](https://github.com/onflow/flow-go/pull/3321/files#diff-7c9a29fd60f7021a8fd387c9e44b5c42dbeac85f35e4466de4b413208ba0444dR169).

I will also change this function to take additional `subtrieLevel` parameter, so that we can replace the hardcoded `15` here to be more flexible.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+// 	}
+//
+// 	return decodeFooter(footer)
+// }
+
+func readSubTriesConcurrently(dir string, fileName string, subtrieChecksums []uint32) ([][]*node.Node, error) {
+	// TODO: replace 16 with const
+	if len(subtrieChecksums) != 16 {
+		return nil, fmt.Errorf("expect subtrieChecksums to be %v, but got %v", 16, len(subtrieChecksums))
+	}
```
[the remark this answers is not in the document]
zhangchiqing: I have added [the check you mentioned](https://github.com/onflow/flow-go/pull/3321/files#diff-2ff6507ee260a1e37156b4cae4cc259e7688bbd0007993b20fa193824073105aR37). 

However, I still think it's good to keep this check here as a sanity check. Since this function returns an error anyway, removing this check wouldn't save us too much.

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: Track this idea here for now https://github.com/dapperlabs/flow-go/issues/6403

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: Thanks for the idea, [tracked here for now](https://github.com/onflow/flow-go/issues/3323)

ledger/complete/wal/checkpoint_v6_writer.go
zhangchiqing: Good point

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	first, last := tries[0], tries[len(tries)-1]
+	logger.Info().
+		Str("first_hash", first.RootHash().String()).
+		Uint64("first_reg_count", first.AllocatedRegCount()).
+		Str("last", last.RootHash().String()).
+		Uint64("last_reg_count", last.AllocatedRegCount()).
+		Msgf("storing checkpoint for %v tries to %v", len(tries), outputDir)
+
```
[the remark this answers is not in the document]
zhangchiqing: Added in this PR

https://github.com/onflow/flow-go/pull/3321/files#diff-d20ae1cfa1445683365b31fa2b1ce057d2b3fec4e4930094dd9469b68536c5bdR56

ledger/complete/wal/checkpoint_v6_reader.go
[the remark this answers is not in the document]
zhangchiqing: makes sense to me. [Issued up here](https://github.com/onflow/flow-go/issues/3325), will address that in separate PR

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,340 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("cannot read node %d: %w", i, err)
+		}
+		nodes[i] = node
+	}
+
+	// TODO: validate checksum again
+
+	return nodes, nil
```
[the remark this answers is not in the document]
zhangchiqing: That's a good idea. Will make a separate PR to refactor this

ledger/complete/wal/checkpointer.go
zhangchiqing: might need further discussion, [tracked in this issue](https://github.com/dapperlabs/flow-go/issues/6404)

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,529 @@
...
+	outputDir string,
+	outputFile string,
+	logger *zerolog.Logger,
+) error {
+	closable, err := createWriterForCheckpointSummary(outputDir, outputFile, logger)
+	if err != nil {
+		return fmt.Errorf("could not store checkpoint summary: %w", err)
+	}
+	defer func() {
+		closeErr := closable.Close()
```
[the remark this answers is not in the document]
zhangchiqing: [will apply a similar change](https://github.com/onflow/flow-go/pull/3356/files#diff-4d387a87aa55ecd3590e6cfeaa400f14d6aa9823fcbf4cded30e23c0117b767fR296) to here , once that PR is approved

fxamacker (changes requested):
My review mainly focused on non-test code.

To reduce bloat, we can remove CLI program that compares checkpoint v5 in this PR.  This program isn't likely to be needed after this PR is merged.  

This is in more than one comment so I'm highlighting it here.  I think we should remove flagVersion from execution state extraction program because we should only support creating checkpoint file in the latest version.  Adding this flag gives the wrong impression that encoding to any version is supported. It also may become a potential bug when this flag is not explicitly set after new checkpoint version is available.

Making number of goroutines configurable for reading subtries can help us test the code on machines that don't have as much RAM as EN by using fewer goroutines. 

Some of my other comments require changes, such as the need to use named return args for error used by `defer`.

ledger/complete/compactor.go
```diff
@@ -331,19 +331,8 @@ func createCheckpoint(checkpointer *realWAL.Checkpointer, logger zerolog.Logger,
...
-		closeErr := writer.Close()
-		// Return close error if there isn't any prior error to return.
-		if err == nil {
-			err = closeErr
-		}
-	}()
-
-	err = realWAL.StoreCheckpoint(writer, tries...)
+	fileName := realWAL.NumberToFilename(checkpointNum)
+	err := realWAL.StoreCheckpointV6Concurrent(tries, checkpointer.Dir(), fileName, &logger)
```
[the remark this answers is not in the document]
fxamacker: This comment is resolved by new commit.

cmd/util/cmd/execution-state-extract/cmd.go
```diff
@@ -26,6 +26,7 @@ var (
 	flagChain             string
 	flagNoMigration       bool
 	flagNoReport          bool
+	flagVersion           int
```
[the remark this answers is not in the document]
zhangchiqing: You are right, we will eventually remove the version, and only keep the latest.

However, for the coming spork, we would like to generate both V5 and V6 in order to give community time to migrate. 

Then for the spork after that, we will remove the version. 

That being said, we need to keep this for the coming spork

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,722 @@
...
+	return parts, nil
+}
+
+var errCheckpointFileExist = errors.New("checkpoint file exists already")
+
+// noneCheckpointFileExist check if none of the checkpoint header file or the part files exist
+// it returns nil if none exists
+// it returns errCheckpointFileExist if a checkpoint file exists already
+// it returns err if running into any other exception
+func noneCheckpointFileExist(dir string, fileName string, totalSubtrieFiles int) error {
```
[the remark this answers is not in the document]
zhangchiqing: Checkpointing will fail if some part file exists, that's why we need to ensure none Checkpoint part file exist.

Not only the header file, but also other part files must not exist.

Ideally, if some part file does exist, it's better that the error message contains the part file we found instead of just a boolean value. So that we could include that part file number into the error message for debugging purpose. 

In other words, errCheckpointFileExist will contain the part file number that we found. 

Since it's possible to hit an exception when checking file existence, I had to return two type of errors, and let caller to distinguish.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,680 @@
...
+// 		1. version
+//		2. checksum of each part file (17 in total)
+// 		3. checksum of the main file itself
+func storeCheckpointHeader(
+	subTrieChecksums []uint32,
+	topTrieChecksum uint32,
+	outputDir string,
+	outputFile string,
+	logger *zerolog.Logger,
+) error {
```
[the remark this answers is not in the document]
zhangchiqing: Agreed. 

[I will merge this PR, which have addressed this](https://github.com/onflow/flow-go/pull/3368)

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,680 @@
...
+	nWorker uint,
+) (
+	map[*node.Node]uint64, // node indices
+	uint64, // node count
+	[]uint32, //checksums
+	error, // any exception
+) {
+	logger.Info().Msgf("storing %v subtrie groups with average node count %v for each subtrie", subtrieCount, estimatedSubtrieNodeCount)
+
+	resultChs := make(chan chan *resultStoringSubTrie, nWorker)
```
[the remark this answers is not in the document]
zhangchiqing: Good catch

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+
+	return subtrieChecksums, topTrieChecksum, nil
+}
+
+// allPartFileExist check if all the part files of the checkpoint file exist
+// it returns nil if all files exist
+// it returns os.ErrNotExist if some file is missing, use (os.IsNotExist to check)
+// it returns err if running into any exception
+func allPartFileExist(dir string, fileName string, totalSubtrieFiles int) error {
```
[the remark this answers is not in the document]
zhangchiqing: It's a bit confusing to use `totalParts` as I consider the header file as one of the part file. However, the header file has different name pattern than others, since other files has `.xx` in the name, whereas the header file doesn't have it. 

That's why I think it's easier to let this function decide the total parts

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,722 @@
...
+	toplevelPath, _ := filePathTopTries(dir, fileName)
+
+	_, ok = lookup[toplevelPath]
+	if ok {
+		parts = append(parts, toplevelPath)
+		delete(lookup, toplevelPath)
+	}
+
+	return parts, nil
+}
```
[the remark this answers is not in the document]
zhangchiqing: I'm not clear what's the benefit of sorting the files. 

I build a lookup map first, and then go through each part file and check if they exists. 

IMO, having gap or not is an implementation detail, this function is to find all files, it doesn't need to assume there is gap or not, even though there shouldn't be.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,722 @@
...
+	matched, err := findCheckpointPartFiles(dir, fileName)
+	if err != nil {
+		return fmt.Errorf("could not check all checkpoint part file exist: %w", err)
+	}
+
+	// header + subtrie files + top level file
+	if len(matched) != 1+totalSubtrieFiles+1 {
+		return fmt.Errorf("some checkpoint part file is missing. found part files %v. err :%w",
+			matched, os.ErrNotExist)
+	}
```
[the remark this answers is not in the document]
zhangchiqing: `findCheckpointPartFiles` uses a lookup map to find all part files, there shouldn't have any duplication, so checking length should be enough, no?

Having gap or not is an implementation detail to me.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,740 @@
...
+
+	// ensure all checkpoint part file exists, might return os.ErrNotExist error
+	// if a file is missing
+	err = allPartFileExist(dir, fileName, len(subtrieChecksums))
+	if err != nil {
+		return nil, fmt.Errorf("fail to check all checkpoint part file exist: %w", err)
+	}
+
+	// TODO making number of goroutine configable for reading subtries, which can help us
```
zhangchiqing, line 52: added as a TODO here @fxamacker , I don't think it's a blocker for now

ledger/complete/wal/wal.go
```diff
@@ -204,23 +215,36 @@ func (w *DiskWAL) replay(
...
+				checkpointLoaded = true
 			}
 		}
 	}
 
+	// if should load checkpoint, but not loaded, then return error
+	if useCheckpoints && !checkpointLoaded {
+		return fmt.Errorf("no checkpoint file was found, tried all checkpoint files, but none of them is valid: %v",
+			allCheckpoints)
+	}
```
zhangchiqing, line 246: this is to fix confusing error message like this:
```
trie with the given rootHash d21b0c1f3653cf074a35f365e70c1840ce5a017287715fca60e1edd5758c20e8 not found
```

into this:
```
could not replay segments [0:11]: no checkpoint file was found, try all checkpoint files, but they are invalid: []
```

zhangchiqing: It seems we allow starting EN without checkpoint? So the checks here might be wrong. I'm going to revert this change

ledger/complete/wal/checkpoint_v6_reader.go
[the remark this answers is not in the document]
fxamacker: It makes sense to log existing part files when none is expected.

I think it's cleaner to remove `noneCheckpointFileExist` since it is only called in `storeCheckpointV6`.  

Instead of wrapping `findCheckpointPartFiles` in `noneCheckpointFileExist()`, `storeCheckpointV6` can directly call `findCheckpointPartFiles` and return error if any part files are found (with enough information to log).

Thoughts?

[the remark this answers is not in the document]
zhangchiqing: makes sense, updated

ledger/complete/wal/checkpoint_v6_reader.go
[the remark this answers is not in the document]
fxamacker: The main benefit of suggested change is to return a list of part numbers on disk.  The list isn't filtered by what is expected, such as part file or top level file.

In general, sorting can simplify testing, fuzzing, and troubleshooting (by allowing deterministic results) so the deciding factor is usually whether the small performance hit outweighs those benefits.

Not a blocker, we have bigger fish to fry.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,721 @@
...
+	[]uint32, //checksums
+	error, // any exception
+) {
+	logger.Info().Msgf("storing %v subtrie groups with average node count %v for each subtrie", subtrieCount, estimatedSubtrieNodeCount)
+
+	if nWorker == 0 || nWorker > subtrieCount {
+		return nil, 0, nil, fmt.Errorf("invalid nWorker %v, the valid range is [1,%v]", nWorker, subtrieCount)
+	}
+
+	resultChs := make(chan chan *resultStoringSubTrie, nWorker-1)
```
fxamacker, line 330: Although `nWorker-1` makes sense for single goroutine checkpointing, it isn't a good fit for multi-goroutine checkpointing.

For example, `nWorker` is 16  in `StoreCheckpointV6Concurrent()`, and it doesn't make sense to change it to 17.

To simplify, I think we can think of `nWorker` as extra workers.  It is 0 (default value) for single goroutine, and `16` for parallel.  So this block of code doesn't need to modify `nWorker`.

https://github.com/onflow/flow-go/blob/44e630a0064d68c793598ef1fd48a056681ce524/ledger/complete/wal/checkpoint_v6_writer.go#L46-L48

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,721 @@
...
+//		}()
+//
+// Better to use as below:
+// func XXX() (
+//
+//	errToReturn error,
+//	) {
+//		def func() {
+//			// good, because the error to returned is only updated here, and guaranteed to be returned
+//			errToReturn = closeAndMergeError(closable, err)
```
fxamacker, line 707: Need to use `errToReturn` as function argument. 

```suggestion
//			errToReturn = closeAndMergeError(closable, errToReturn)
```

cmd/execution_config.go
```diff
@@ -63,6 +64,7 @@ func (exeConf *ExecutionConfig) SetupFlags(flags *pflag.FlagSet) {
 	flags.Uint32Var(&exeConf.mTrieCacheSize, "mtrie-cache-size", 500, "cache size for MTrie")
 	flags.UintVar(&exeConf.checkpointDistance, "checkpoint-distance", 20, "number of WAL segments between checkpoints")
 	flags.UintVar(&exeConf.checkpointsToKeep, "checkpoints-to-keep", 5, "number of recent checkpoints to keep (0 to keep all)")
+	flags.BoolVar(&exeConf.outputCheckpointV5, "outputCheckpointV5", false, "output checkpoint file in v5")
```
zhangchiqing, line 67: @fxamacker added a bool flag for using v5

fxamacker:
@zhangchiqing 

Instead of using bool flag to indicate v5, what do you think about using int flag for checkpoint version and set default to `MaxVersion` like this:  
https://github.com/onflow/flow-go/blob/124f1c581388b9da2783a391a0aa1d46ef505c16/ledger/complete/wal/checkpointer.go#L61

fxamacker:
@zhangchiqing 

We also need to modify `RemoveCheckpoint` to remove all checkpoint files for v6, because v6 generates multiple files.

https://github.com/onflow/flow-go/blob/124f1c581388b9da2783a391a0aa1d46ef505c16/ledger/complete/wal/checkpointer.go#L632-L634

cmd/execution_config.go
zhangchiqing: >Instead of using bool flag to indicate v5, what do you think about using int flag for checkpoint version and set default to MaxVersion like this:

I've set the default to use MaxVersion. But I think it still makes sense to use bool flag here. 

We don't want to support old versions, but only v5. 

Using bool can reduce human mistake, and we will anyway remove this flag after next spork.

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -327,37 +333,48 @@ func storeSubTrieConcurrently(
...
+
+	// push all jobs into the channel
+	for i, roots := range subtrieRoots {
+		resultCh := make(chan *resultStoringSubTrie)
+		resultChs[i] = resultCh
+		jobs <- job{
+			Index:  i,
+			Roots:  roots,
+			Result: resultCh,
+		}
```
fxamacker, line 347: Need to close input channel so goroutines exit properly.
```suggestion
		}
		close(jobs)
```

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -327,37 +333,48 @@ func storeSubTrieConcurrently(
...
+	}
+
+	// start nWorker number of goroutine to take the job from the jobs channel concurrently
+	// and work on them, after finish, continue until the jobs channel is drained
+	for i := 0; i < int(nWorker); i++ {
+		go func(i int) {
```
fxamacker, line 353: We can remove worker index parameter because we only use job index in the goroutine.

```suggestion
		go func() {
```

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -327,37 +333,48 @@ func storeSubTrieConcurrently(
...
+			for job := range jobs {
 				roots, nodeCount, checksum, err := storeCheckpointSubTrie(
-					i, subTrieRoot, estimatedSubtrieNodeCount, outputDir, outputFile, logger)
-				resultCh <- &resultStoringSubTrie{
+					job.Index, job.Roots, estimatedSubtrieNodeCount, outputDir, outputFile, logger)
+
+				job.Result <- &resultStoringSubTrie{
 					Index:     i,
```
fxamacker, line 359: Need job index here instead of worker index.

```suggestion
		                        Index:     job.Index
```

fxamacker (changes requested):

fxamacker (commented):
Thanks for resolving comments.  Left some minor comments.  BTW, looks like unit test is broken by today's commit.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -0,0 +1,748 @@
...
+	for _, match := range matched {
+		lookup[match] = struct{}{}
+	}
+
+	parts := make([]string, 0)
+	// check header exists
+	_, ok := lookup[headerPath]
+	if ok {
+		parts = append(parts, headerPath)
+		delete(lookup, headerPath)
```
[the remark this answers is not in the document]
zhangchiqing: It is to prevent duplicated file paths

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -0,0 +1,740 @@
...
+) (
+	map[*node.Node]uint64, // node indices
+	uint64, // node count
+	[]uint32, //checksums
+	error, // any exception
+) {
+	logger.Info().Msgf("storing %v subtrie groups with average node count %v for each subtrie", subtrieCount, estimatedSubtrieNodeCount)
+
+	if nWorker == 0 || nWorker > subtrieCount {
+		return nil, 0, nil, fmt.Errorf("invalid nWorker %v, the valid range is [1,%v]", nWorker, subtrieCount)
```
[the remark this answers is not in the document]
zhangchiqing: I disagree here. We should be as strict as possible. If it self-correct here, developer or operator might not be aware of such self-correction. 

We should fail and highlight the problem, [rather than trying "best effort"](https://github.com/onflow/flow-go/blob/feature-checkpoint-v6/CodingConventions.md#error-handling)

ledger/complete/wal/checkpointer.go
```diff
@@ -442,6 +478,10 @@ func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
 	// included.
 	for _, t := range tries {
 		rootNode := t.RootNode()
+		if !t.IsEmpty() && rootNode.Height() != ledger.NodeMaxHeight {
+			return fmt.Errorf("height of root node must be %d, but is %d",
+				ledger.NodeMaxHeight, rootNode.Height())
+		}
```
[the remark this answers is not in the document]
zhangchiqing: A similar check is in V5, I think this would be consistent
https://github.com/onflow/flow-go/blob/feature-checkpoint-v6/ledger/complete/wal/checkpointer.go#L492

fxamacker (approved):
Thanks for the updates and fixing failed unit test on Friday night.  I'll respond to some comments separately later so I don't delay this PR.

I think it is probably best to wait for Ramtin's review before merging.

Maybe update PR text with latest benchmark results you shared on Friday.  E.g. this PR:
- slows down regular checkpointing by 1.2 minutes (was 10.6 mins, will be 11.8 mins).
- speeds up spork state extraction (serialize) by 7.8 mins (was 10.6 mins, will be 2.8 mins)
- speeds up spork state extraction (deserialize) by ___ mins (need benchmark results)
- speeds up EN startup (deserialize) by ... Exact same speedup as prior line (deserialize) but different scenario not limited to sporks (tradeoff is one-time cost of using more memory during startup which is OK).

ramtinms (approved):
Looks good to me, sorry that was a very large PR, took me a while to go over the changes.

zhangchiqing:
bors merge

PR #3275, opened 2022-09-26
Replace error/failure structs with CodedError/CodedFailure

pattyshack:
NOTE:
- I've deprecated fvm internal error since it's no longer used and is wrong ( fvm internal error should be treated as failure)
- I've added an bound check to the valueStr in StateValueSizeLimitError (the refactor discovered a previously hidden out-of-bound panic)
- NewTransactionFeeDeductionFailedError now takes the TxFees as argument as well (We forgot to initalize the TxFees value)
- I've left FrozenAccountError and InvalidProposalSeqNumberError alone since these requires additional error context (I'll handles these in a followup PR)

### DIFF
--- a/fvm/errors/execution.go
+++ b/fvm/errors/execution.go
@@ -1,8 +1,6 @@
 package errors
 
 import (
-	"errors"
-	"fmt"
 	"strings"
 
 	"github.com/onflow/cadence/runtime"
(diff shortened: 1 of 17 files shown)


fvm/errors/execution.go
```diff
@@ -1,374 +1,220 @@
...
+
+// NewStateValueSizeLimitError constructs a CodedError which indicates that the
+// provided value has exceeded the size limit allowed by the storage.
+func NewStateValueSizeLimitError(
+	value flow.RegisterValue,
+	size uint64,
+	limit uint64,
+) *CodedError {
+	valueStr := ""
+	if len(value) > 23 {
```
pattyshack, line 136: This check is new.  The out-of-bound error was previously hidden

fvm/errors/execution.go
```diff
@@ -1,374 +1,220 @@
...
+func IsCadenceRuntimeError(err error) bool {
+	return HasErrorCode(err, ErrCodeCadenceRunTimeError)
+}
+
+// NewTransactionFeeDeductionFailedError constructs a new CodedError which
+// indicates that a there was an error deducting transaction fees from the
+// transaction Payer
+func NewTransactionFeeDeductionFailedError(
+	payer flow.Address,
+	txFees uint64,
```
pattyshack, line 40: We forgot to pass this in previously

pattyshack:
hmm.  this involves more changes than I wanted.  I'll split this into multiple PRs.

PR #3279, opened 2022-09-27
Add IsValueError to errors/

pattyshack:


zhangchiqing (approved):

Tonix517 (approved):

fvm/errors/base.go
```diff
@@ -135,6 +135,11 @@ func (e ValueError) Code() ErrorCode {
 	return ErrCodeValueError
 }
 
+func IsValueError(err error) bool {
+	var valErr ValueError
+	return As(err, &valErr)
```
SaveTheRbtz, line 140: it looks a bit weird that we need to wrap `errors.As` nowadays.

PR #3281, opened 2022-09-27
Deprecate/delete FVMInternalError

pattyshack:
it's no longer used and is wrong (fvm internal error should be treated as failure)

### DIFF
--- a/fvm/errors/codes.go
+++ b/fvm/errors/codes.go
@@ -47,6 +47,7 @@ const (
 	ErrCodeInvalidEnvelopeSignatureError ErrorCode = 1009
 
 	// base errors 1050 - 1100
+	// Deprecated: No longer used.
 	ErrCodeFVMInternalError            ErrorCode = 1050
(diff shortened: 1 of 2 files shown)


Tonix517 (approved):

fvm/errors/codes.go
```diff
@@ -47,6 +47,7 @@ const (
 	ErrCodeInvalidEnvelopeSignatureError ErrorCode = 1009
 
 	// base errors 1050 - 1100
+	// Deprecated: No longer used.
 	ErrCodeFVMInternalError            ErrorCode = 1050
```
zhangchiqing, line 51: why still keeping this error?

pattyshack: i'm just following janez's convention

pattyshack: my guess is we want to keep these around just to be sure we don't accidentally reuse the same number

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3289, opened 2022-09-27
Replace error/failure structs with CodedError/CodedFailure (Part 1)

pattyshack:
I'm going to splay the replacement across multiple PRs.  The changes are unreviewable on github if I try to replace them all in one go since that causes too many newline changes for diff-ing.

Note that I've change NewEncodingFailuref a bit.  Previously, the original encoding error was sometimes dropped from the returned error string.

Note: I've added a todo_rm_emulator_hack.go to this diff to deal with emulator circular dependency hell.  Looks like the error type classification in the emulator are incorrect even prior to my changes, so there's no point in trying to keep backwards compatibility.

### DIFF
--- a/fvm/environment/accounts.go
+++ b/fvm/environment/accounts.go
@@ -367,8 +367,10 @@ func (a *StatefulAccounts) setContractNames(contractNames contractNames, address
 	cborEncoder := cbor.NewEncoder(&buf)
 	err = cborEncoder.Encode(contractNames)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pattyshack:
i've added todo_rm_emulator_hack.go to this PR to deal with emulator / integration test circular hell.  The emulator error classification is broken even before this PR.

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors merge

PR #3290, opened 2022-09-27
[Execution] Return error when closing gcp connection fails

peterargue:
Execution nodes have an option to upload block computation results to a GCS bucket. The upload has a retry feature, but does not check the error returned from flushing/closing the connection when deciding if the upload succeeded. Occasionally, the close operation fails with a 50x error from the GCS API, resulting in miss block uploads.

This PR also checks the response from closing the stream and returns an error encountered.

### DIFF
--- a/engine/execution/computation/computer/uploader/gcp_uploader.go
+++ b/engine/execution/computation/computer/uploader/gcp_uploader.go
@@ -45,14 +45,21 @@ func (u *GCPBucketUploader) Upload(computationResult *execution.ComputationResul
 	object := u.bucket.Object(objectName)
 
 	writer := object.NewWriter(u.ctx)
-	defer func() {
-		err := writer.Close()
-		if err != nil {
-			u.log.Warn().Err(err).Str("object_name", objectName).Msg("error while closing GCP object")
-		}
-	}()
 
-	return WriteComputationResultsTo(computationResult, writer)
+	// serialize and write computation result to upload stream
+	err := WriteComputationResultsTo(computationResult, writer)
+	if err != nil {
+		return fmt.Errorf("failed to write computation result to GCP stream: %w", err)
+	}
+
+	// flush and close the stream
+	// this occasionally fails with HTTP 50x errors due to flakiness on the network/GCP API
+	err = writer.Close()
+	if err != nil {
+		return fmt.Errorf("error while closing GCP object: %w", err)
+	}
+
+	return nil
 }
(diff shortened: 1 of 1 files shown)


koko1123 (approved):

engine/execution/computation/computer/uploader/gcp_uploader.go
```diff
@@ -45,14 +45,21 @@ func (u *GCPBucketUploader) Upload(computationResult *execution.ComputationResul
...
-			u.log.Warn().Err(err).Str("object_name", objectName).Msg("error while closing GCP object")
-		}
-	}()
-
-	return WriteComputationResultsTo(computationResult, writer)
+
+	// serialize and write computation result to upload stream
+	err := WriteComputationResultsTo(computationResult, writer)
+	if err != nil {
+		return fmt.Errorf("failed to write computation result to GCP stream: %w", err)
```
Tonix517, line 52: if this error block is reached, is `Close()` still called? imho the `Close()` should still be called in a defer, so that in any case the writer will be closed. in that case we can set error from `Close()` to a named return value(error) of this function, or we use multi-error which @janezpodhostnik has been working on.

engine/execution/computation/computer/uploader/gcp_uploader.go
```diff
@@ -39,20 +40,33 @@ func NewGCPBucketUploader(ctx context.Context, bucketName string, log zerolog.Lo
...
+
+		err = errs.ErrorOrNil()
 	}()
 
-	return WriteComputationResultsTo(computationResult, writer)
+	// serialize and write computation result to upload stream
+	writeErr := WriteComputationResultsTo(computationResult, writer)
+	if writeErr != nil {
+		errs = multierror.Append(errs, fmt.Errorf("error while writing computation result to GCP object: %w", writeErr))
+		return
```
Tonix517, line 66: nit: this return can be removed.

Tonix517 (approved):
LGTM

engine/execution/computation/computer/uploader/gcp_uploader.go
peterargue: I added it for completeness. that way if this code is ever extended to do something after that conditional, there's less chance the implementor forgets to re-add the return.

koko1123 (approved):

peterargue:
bors merge

PR #3294, opened 2022-09-28
Front-Load HotStuff Verification Pt. 1: Update Compliance Engines

jordanschalm:
This PR modifies compliance engines to front-load HotStuff verification:
* Once a block connects to the finalized state, we validate the proposal using HotStuff first, and only then attempt to extend the state (compliance checks)
* This means, all blocks in storage have been fully validated (there is no need for a separate `MarkValid` step)

⚠️ Depends on https://github.com/onflow/flow-go/pull/3248
👉  Removing `MarkValid` will be done in a separate PR: https://github.com/onflow/flow-go/pull/3303


engine/collection/compliance/core.go
```diff
@@ -200,21 +204,33 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
...
-		return nil
-	}
 	if err != nil {
+		if engine.IsOutdatedInputError(err) {
+			// child is outdated by the time we started processing it
+			// => node was probably behind and is catching up. Log as warning
+			log.Info().Msg("dropped processing of abandoned fork; this might be an indicator that the node is slightly behind")
+			return nil
+		}
+		if engine.IsInvalidInputError(err) {
```
durkmurder, line 222: :warning: At this point we need to notify vote aggregator that there is an invalid block to potentially slash byzantine voters. We could directly notify `VoteAggregator` or use a pub-sub mechanism(requires extending `Consumer` I suppose)

engine/collection/compliance/core.go
```diff
@@ -261,32 +282,57 @@ func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) err
...
-	// if the block proposes an invalid extension of the protocol state, then the block is invalid
-	if state.IsInvalidExtensionError(err) {
-		return engine.NewInvalidInputErrorf("invalid extension of protocol state (block: %x, height: %d): %w",
-			header.ID(), header.Height, err)
-	}
-	// protocol state aborted processing of block as it is on an abandoned fork: block is outdated
-	if state.IsOutdatedExtensionError(err) {
-		return engine.NewOutdatedInputErrorf("outdated extension of protocol state: %w", err)
-	}
+	err = c.state.Extend(block)
```
durkmurder, line 320: Can we extend godoc of `Extend` with returned errors? Will be easier to check if all sentinels are covered. Thanks

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +259,148 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
 	cs.pending.AssertNotCalled(cs.T(), "Add", originID, mock.Anything)
 }
 
-func (cs *CoreSuite) TestOnBlockProposalValidAncestor() {
+// TestOnBlockProposal_InvalidProposal tests that a proposal which fails HotStuff validation.
+//   - should not go through compliance checks
+//   - should not be added to the state
+//   - we should not attempt to process its children
+func (cs *CoreSuite) TestOnBlockProposal_InvalidProposal() {
```
durkmurder, line 266: also it needs to report `VoteAggregator.InvalidBlock` by any means

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +259,148 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
+		// the block fails HotStuff validation
+		*cs.validator = *hotstuff.NewValidator(cs.T())
+		cs.validator.On("ValidateProposal", model.ProposalFromFlow(block.Header, parent.Header.View)).Return(model.ErrViewForUnknownEpoch)
+
+		// the expected error should be handled within the Core
+		err := cs.core.OnBlockProposal(originID, proposal)
+		require.NoError(cs.T(), err, "proposal with invalid extension should fail")
+
+		// we should not extend the state with the header
+		cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything, block)
```
durkmurder, line 304:
```suggestion
		cs.state.AssertNotCalled(cs.T(), "Extend", block)
```

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +259,148 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
+		unexpectedErr := errors.New("generic unexpected error")
+		*cs.validator = *hotstuff.NewValidator(cs.T())
+		cs.validator.On("ValidateProposal", model.ProposalFromFlow(block.Header, parent.Header.View)).Return(unexpectedErr)
+
+		// the error should be propagated
+		err := cs.core.OnBlockProposal(originID, proposal)
+		require.ErrorIs(cs.T(), err, unexpectedErr)
+
+		// we should not extend the state with the header
+		cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything, block)
```
durkmurder, line 320:
```suggestion
		cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything)
```

engine/consensus/compliance/core.go
```diff
@@ -230,22 +234,33 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
...
 
+	log := c.log.With().
+		Str("block_id", blockID.String()).
+		Uint64("block_height", proposal.Header.Height).
+		Uint64("block_view", proposal.Header.View).
+		Logger()
+
 	// process block itself
 	err := c.processBlockProposal(proposal)
 	if err != nil {
```
durkmurder, line 245: same as with collection compliance engine, need to notify vote aggregator

engine/collection/compliance/core.go
jordanschalm: 👍 I will notify directly for now. We also call `AddVote` directly and they seem to be similar cost.

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +259,148 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
-	err := cs.core.OnBlockProposal(originID, proposal)
-	require.NoError(cs.T(), err, "valid block proposal should pass")
+		// the expected error should be handled within the Core
+		err := cs.core.OnBlockProposal(originID, proposal)
+		require.NoError(cs.T(), err, "proposal with invalid extension should fail")
 
-	// we should extend the state with the header
-	cs.state.AssertCalled(cs.T(), "Extend", &block)
+		// we should not extend the state with the header
+		cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything, block)
```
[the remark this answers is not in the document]
jordanschalm: 👌 Nice catch - thank you

engine/collection/compliance/core.go
```diff
@@ -261,32 +282,57 @@ func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) err
...
 	if err != nil {
-		return fmt.Errorf("could not extend protocol state (block: %x, height: %d): %w", header.ID(), header.Height, err)
-	}
-
-	// retrieve the parent
-	parent, err := c.headers.ByBlockID(header.ParentID)
-	if err != nil {
-		return fmt.Errorf("could not retrieve proposal parent: %w", err)
+		if state.IsInvalidExtensionError(err) {
```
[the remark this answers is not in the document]
jordanschalm: Yes, good point - I'll add a note like we have in consensus core.

state/cluster/badger/mutator.go
```diff
@@ -36,7 +36,12 @@ func NewMutableState(state *State, tracer module.Tracer, headers storage.Headers
 	return mutableState, nil
 }
 
+// Extend validates that the given cluster block passes compliance rules, then inserts
+// it to the cluster state.
 // TODO (Ramtin) pass context here
+// Expected errors during normal operations:
+//   - state.OutdatedExtensionError if the candidate block is outdated (e.g. orphaned)
+//   - state.InvalidExtensionError if the candidate block is invalid
 func (m *MutableState) Extend(block *cluster.Block) error {
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3294/commits/08a04b3b9a70d2791af3dc9148afed7af71b5aa9

engine/collection/compliance/core.go
jordanschalm: https://github.com/onflow/flow-go/pull/3294/commits/08a04b3b9a70d2791af3dc9148afed7af71b5aa9

engine/collection/compliance/core.go
```diff
@@ -200,21 +205,50 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
...
+			if err != nil {
+				if mempool.IsBelowPrunedThresholdError(err) {
+					log.Warn().Msg("received invalid block, but is below pruned threshold")
+					return nil
+				} else {
+					return fmt.Errorf("unexpected error notifying vote aggregator about invalid block: %w", err)
+				}
+			}
+			return nil
+		} else if engine.IsUnverifiableInputError(err) {
```
durkmurder, line 246: I believe this is also slashable, malicious replicas could have voted for this proposal, need to notify VA

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +253,152 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
+		// we should not attempt to process the children
+		cs.pending.AssertNotCalled(cs.T(), "ByParentID", mock.Anything)
+	})
 }
 
-func (cs *CoreSuite) TestOnBlockProposalInvalidExtension() {
+// TestOnBlockProposal_InvalidExtension tests processing a proposal which passes HotStuff validation,
+// but fails compliance checks.
+//   - should not be added to the state
+//   - we should not attempt to process its children
```
durkmurder, line 327: VA needs to be notified

engine/collection/compliance/core_test.go
```diff
@@ -230,69 +253,152 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
+		err := cs.core.OnBlockProposal(originID, proposal)
+		require.NoError(cs.T(), err, "proposal with invalid extension should fail")
+
+		// we should extend the state with the header
+		cs.state.AssertCalled(cs.T(), "Extend", &block)
+		// we should not pass the block to hotstuff
+		cs.hotstuff.AssertNotCalled(cs.T(), "SubmitProposal", mock.Anything, mock.Anything)
+		// we should not attempt to process the children
+		cs.pending.AssertNotCalled(cs.T(), "ByParentID", mock.Anything)
+		cs.voteAggregator.AssertExpectations(cs.T())
```
durkmurder, line 363: a suggestion to use `mocks.NewVoteAggregator` and get rid of `AssertExpectations` in other places.

engine/consensus/compliance/core.go
```diff
@@ -230,22 +235,52 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
...
+				if mempool.IsBelowPrunedThresholdError(err) {
+					log.Warn().Msg("received invalid block, but is below pruned threshold")
+					return nil
+				} else {
+					return fmt.Errorf("unexpected error notifying vote aggregator about invalid block: %w", err)
+				}
+			}
+			return nil
+		}
+		if engine.IsUnverifiableInputError(err) {
```
durkmurder, line 278: Same as collector cluster, I believe we should notify VA as well

engine/consensus/compliance/core_test.go
```diff
@@ -332,40 +332,156 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
 	cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything)
+	cs.validator.AssertNotCalled(cs.T(), "ValidateProposal", mock.Anything)
 	cs.pending.AssertNotCalled(cs.T(), "Add", originID, mock.Anything)
 }
 
-func (cs *CoreSuite) TestOnBlockProposalInvalidExtension() {
+// TestOnBlockProposal_InvalidProposal tests that a proposal which fails HotStuff validation.
+//   - should not go through compliance checks
+//   - should not be added to the state
+//   - we should not attempt to process its children
```
durkmurder, line 342: should notify VA

engine/consensus/compliance/core_test.go
```diff
@@ -332,40 +332,156 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
...
+		cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything, block)
+		// we should not attempt to process the children
+		cs.pending.AssertNotCalled(cs.T(), "ByParentID", mock.Anything)
+	})
+}
+
+// TestOnBlockProposal_InvalidExtension tests processing a proposal which passes HotStuff validation,
+// but fails compliance checks.
+//   - should not be added to the state
+//   - we should not attempt to process its children
```
durkmurder, line 410: should notify VA

engine/consensus/compliance/core.go
```diff
@@ -230,22 +235,52 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
 func (c *Core) processBlockAndDescendants(proposal *messages.BlockProposal) error {
 	blockID := proposal.Header.ID()
 
+	// retrieve the parent block
+	//  - once we reach this point, the parent block must have been validated
+	//    and inserted to the protocol state
+	parent, err := c.headers.ByBlockID(proposal.Header.ParentID)
```
zhangchiqing, line 241: We are querying this parent second time here. Consider using the parent queried [the first time here](https://github.com/onflow/flow-go/pull/3294/files#diff-91e83ea2456f635c82eea90128d3f2820974b95947215b9a75da64c0f513b5e5R194)

engine/consensus/compliance/core.go
```diff
@@ -280,26 +315,49 @@ func (c *Core) processBlockAndDescendants(proposal *messages.BlockProposal) erro
...
+	span, ctx, isSampled := c.tracer.StartBlockSpan(context.Background(), blockID, trace.ConCompProcessBlockProposal)
 	if isSampled {
 		span.SetAttributes(
 			attribute.String("proposer", header.ProposerID.String()),
 		)
 	}
 	defer span.End()
 
+	hotstuffProposal := model.ProposalFromFlow(header, parent.View)
+	err := c.validator.ValidateProposal(hotstuffProposal)
```
zhangchiqing, line 335: The processBlockProposal will not be called concurrently right? because the engine's `processAvailableMessages` method delivers one message at a time, right?

Since `processBlockProposal` is called in a single thread manner, then this `ValidateProposal` is also called in a single thread manner right?

Before the validation is done in hotstuff in a different thread, then the payload validation (state.Extend) and header validation (validator.ValidateProposal) are done concurrently. 

Note what's concurrently validated is not for the same block, but different block, meaning it's concurrently validating a payload for one block, and validating a header for a different block. 

Now in this PR, it seems both the payload and header validation are done by compliance sequentially, as there is no concurrent validation. Wouldn't it slow down the process of a proposal?

jordanschalm: Yes, excellent point. There is a subtle concurrency we're losing here. The reason we could extract that concurrency previously is that we stored potentially invalid blocks, and validated payloads before signatures, which we would like to avoid doing. 

Earlier in this feature branch, we added the ability to [query identities by view, for known epochs](https://github.com/onflow/flow-go/blob/6a39c8c64184231a05dd5c756db91307f0361e78/consensus/hotstuff/committee.go#L81). 

With the change to front-load QC verification, conceptually, we can perform different concurrent processing:
1. We can validate block payloads (compliance checks) in different forks concurrently, so long as we validate parents before children. 
  * This requires ensuring `MutableState` is concurrency safe
  * This only helps in the unhappy path, when we have competing forks
2. We can validate block headers (HotStuff checks) all concurrently, so long as they refer to a view for a known epoch (true the vast majority of the time). 
  * To do this, we need to know the parent view of the block, which we currently only discover by directly observing and processing the parent block itself (ie. in practice we need to process the parent block first, anyway, so we lose most of the concurrency possibilities)

engine/consensus/compliance/core.go
jordanschalm: Nice catch, thanks https://github.com/onflow/flow-go/pull/3294/commits/64ca513cd99b5fc0216fe9a7747d266fd1d1baec

engine/collection/compliance/core.go
jordanschalm: Right, to actually submit a slashing challenge we would need slightly more information from elsewhere, in particular a confirmation of a finalized block before the epoch commitment deadline, but it is knowable. Agree we should notify VA for this case.

engine/collection/compliance/core_test.go
jordanschalm: 👍 Was already using `NewVoteAggregator`, so `AssertExpectations` is indeed unnecessary - good point.

engine/consensus/compliance/core.go
durkmurder: In my mind we would to take a gradual approach for this problem. First we update compliance engine to correctly process blocks including header validation, payload validation, state extension and passing to hotstuff. We do this as Jordan did this. In single threaded fashion. Once we verify our implementation is correct and actually works we take advantage of stateless header validation and build a pipelines where we validate headers concurrently and then pass them to single-threaded pipeline where we validate payload and apply state changes in strict parent-child order. 
This approach doesn't get a speedup when producing blocks in steady pace but gives speedup when blocks are delivered faster than we can process them(in case of sync for instance). 
After this we can think about concurrent payload validation but it gets tricky at this point and requires careful organization due to parent-child requirement. Might not worth the effort.

To summarize, I am happy how Jordan approached this, we should incrementally update our implementation. In next iterations we will be adding concurrency support at least for header validation.

durkmurder (approved):
Really nice, great job. Excited to see everything running in the wild.

zhangchiqing (approved):

jordanschalm:
bors merge

jordanschalm:
bors retry

engine/consensus/compliance/core.go
AlexHentschel: Mentioning the following just for completeness. Conceptually, there is an even more efficient approach possible:
* Concurrency mainly pays off, if a node is catching up and has a longer backlog of blocks pending validation
* In many cases, we will receive such blocks from the sync engine in batches, such as `A <- B <- C <- D <- E`
* Technically, it is sufficient to validate
   - validate the hashes of the blocks and payload for all blocks `A, .. , E`
   - validate the QC from block `E` (pointing to block `D`)
   - validate payload of block `E`
   
   This is sufficient because consensus is an inductive protocol. Since block `D` received a QC, we  can be certain that all ancestor blocks up to and including `D` must be correct, including their payload. Hence, we could just write blocks `A, .. , D` into storage without _any_ validation (except for recomputing the hashes to confirm integrity of the blocks)

Note:
* In this approach, we firmly assume that byzantine nodes will never reach a supermajority of stake (_more_ than 2/3). If this assumption is violated, an honest node that is catching up might not notice and erroneously vote for an extension of an invalid fork. 
* Nevertheless, safety of finality already breaks at byzantine nodes reaching 1/3 of stake, at which double-spend attacks become theoretically possible. 

However, _exceeding_ 2/3 of Byzantine stake is a much larger hurdle for an attacker. Hence, I feel the assumption of having at most 2/3 byzantine stake is fine because other important consensus guarantees break much earlier.

AlexHentschel: I think **practically most impactful** would be concurrent validation of the _same block_. Reasoning:
* Concurrent header and payload validation of _different_ blocks is mainly useful for nodes that are catching up. I consider it unlikely that this parallelism will result in noteworthy performance improvements on the happy path. 
* In comparison, validating different aspects of the _same_ proposal concurrently has the potential to dramatically speed up the consensus hotpath and thereby improve block rate and reduce platform latency overall. A block has many different aspects that could be checked in parall:
  * header, incl. proposer signature, height, view etc
  * QC for parent
  * seals (validity of seals only depends on ancestor blocks but not on the current block's payload)
  * execution results & receipts 
  * collection guarantees

  At the moment, we check these 5 aspects sequentially. I assume some will be much more expensive to check, but if we check all 5 aspects in parallel, we could potentially still cut verification time down to half.

  With similar improvements for block production, we could reach 1s block time (with a multitude of further improvements also being possible).

engine/collection/compliance/core.go
```diff
@@ -261,32 +295,48 @@ func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) err
...
+		if model.IsInvalidBlockError(err) {
+			return engine.NewInvalidInputErrorf("invalid block proposal: %w", err)
+		} else if errors.Is(err, model.ErrViewForUnknownEpoch) {
+			// We have received a proposal, but we don't know the epoch its view is within.
+			// We know:
+			//  - the parent of this block is valid and inserted (ie. we knew the epoch for it)
+			//  - if we then see this for the child, one of two things must have happened:
+			//    1. the proposer malicious created the block for a view very far in the future (it's invalid)
+			//      -> in this case we can disregard the block
+			//    2. no blocks have been finalized the epoch commitment deadline, and the epoch end
```
AlexHentschel, line 310:
```suggestion
			//    2. no blocks have been finalized within the epoch commitment deadline, and the epoch ended
```

AlexHentschel (commented):
first batch of comments. Didn't quite get through the PR.

AlexHentschel:
Btw, also created issue https://github.com/onflow/flow-go/issues/3374 where I noted suggestions for potential future revisions. Feel free to address points, if you find that they fit nicely within the scope of the PR (we can also always follow up with subsequent PRs)

AlexHentschel:
Generally, I find the terminology "HotStuff validation" and "compliance checks" ambiguous. 
https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/collection/compliance/core_test.go#L324-L325

This is probably a legacy from our old design, where the compliance engine would check the payload and hotstuff would check the header. However, now the compliance engine checks everything, which makes the terminology confusing in my opinion. 

#### suggestion
Use terminology: 
* "header validation" 
* and "payload validation"

I would appreciate if we could update the naming and documentation of the [tests](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/collection/compliance/core_test.go):
* rename [`TestOnBlockProposal_InvalidProposal()`](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/collection/compliance/core_test.go#L256-L261) to `TestOnBlockProposal_InvalidHeader()` and update the godoc 
   * "fails HotStuff validation" -> "fails Header validation" 
   * "compliance checks" -> "payload checks"
* analogous for [`TestOnBlockProposal_InvalidExtension()`](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/collection/compliance/core_test.go#L324-L329):

engine/common/follower/engine.go
```diff
@@ -381,6 +386,36 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
...
+			// We know:
+			//  - the parent of this block is valid and inserted (ie. we knew the epoch for it)
+			//  - if we then see this for the child, one of two things must have happened:
+			//    1. the proposer malicious created the block for a view very far in the future (it's invalid)
+			//      -> in this case we can disregard the block
+			//    2. no blocks have been finalized the epoch commitment deadline, and the epoch end
+			//       (breaking a critical assumption - see EpochCommitSafetyThreshold in protocol.Params for details)
+			//      -> in this case, the network has encountered a critical failure
+			//  - we assume in general that Case 2 will not happen, therefore we can discard this proposal
+			log.Err(err).Msg("unable to validate proposal with view from unknown epoch")
```
AlexHentschel, line 414: > therefore we can discard this proposal

To disregard the proposal, wouldn't we need to return here? Otherwise, we trigger an exception in [line 416](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/common/follower/engine.go#L416)

```suggestion
			log.Err(err).Msg("unable to validate proposal with view from unknown epoch")
			return nil
```

engine/common/follower/engine.go
```diff
@@ -390,7 +425,7 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
 	// check whether the block is a valid extension of the chain.
 	// it only checks the block header, since checking block body is expensive.
 	// The full block check is done by the consensus participants.
-	err := e.state.Extend(ctx, block)
+	err = e.state.Extend(ctx, block)
```
AlexHentschel, line 428:
```suggestion
	// TODO: CAUTION we write a block to disk, without validating its payload yet. This is vulnerable to malicious primaries. 
	err = e.state.Extend(ctx, block)
```

engine/consensus/compliance/core.go
```diff
@@ -227,25 +232,48 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
...
 			// => node was probably behind and is catching up. Log as warning
-			c.log.Info().Msg("dropped processing of abandoned fork; this might be an indicator that the node is slightly behind")
+			log.Info().Msg("dropped processing of abandoned fork; this might be an indicator that the node is slightly behind")
 			return nil
-		}
-		if engine.IsInvalidInputError(err) {
-			// the block is invalid; log as error as we desire honest participation
-			// ToDo: potential slashing
-			c.log.Warn().Err(err).Msg("received invalid block from other node (potential slashing evidence?)")
+		} else if engine.IsInvalidInputError(err) || engine.IsUnverifiableInputError(err) {
```
AlexHentschel, line 253: similarly to [my comment before](https://github.com/onflow/flow-go/pull/3294#discussion_r994034443), it might be nice to not duplicate checks here and encapsulate some of the logic in auxiliary functions.

engine/consensus/compliance/core.go
```diff
@@ -280,26 +308,49 @@ func (c *Core) processBlockAndDescendants(proposal *messages.BlockProposal) erro
...
+	hotstuffProposal := model.ProposalFromFlow(header, parent.View)
+	err := c.validator.ValidateProposal(hotstuffProposal)
+	if err != nil {
+		if model.IsInvalidBlockError(err) {
+			return engine.NewInvalidInputErrorf("invalid block proposal: %w", err)
+		}
+		if errors.Is(err, model.ErrViewForUnknownEpoch) {
+			// We have received a proposal, but we don't know the epoch its view is within.
+			// We know:
+			//  - the parent of this block is valid and inserted (ie. we knew the epoch for it)
```
AlexHentschel, line 336:
```suggestion
			//  - the parent of this block is valid and was appended to the state (ie. we knew the epoch for it)
```

engine/consensus/compliance/core.go
```diff
@@ -280,26 +308,49 @@ func (c *Core) processBlockAndDescendants(proposal *messages.BlockProposal) erro
...
+			//  - if we then see this for the child, one of two things must have happened:
+			//    1. the proposer malicious created the block for a view very far in the future (it's invalid)
+			//      -> in this case we can disregard the block
+			//    2. no blocks have been finalized the epoch commitment deadline, and the epoch end
```
AlexHentschel, line 340:
```suggestion
			//    2. no blocks have been finalized within  the epoch commitment deadline, and the epoch ended
```

engine/consensus/compliance/core_test.go
```diff
@@ -332,40 +332,156 @@ func (cs *CoreSuite) TestOnBlockProposalSkipProposalThreshold() {
 
 	// block should be dropped - not added to state or cache
 	cs.state.AssertNotCalled(cs.T(), "Extend", mock.Anything)
+	cs.validator.AssertNotCalled(cs.T(), "ValidateProposal", mock.Anything)
 	cs.pending.AssertNotCalled(cs.T(), "Add", originID, mock.Anything)
 }
 
-func (cs *CoreSuite) TestOnBlockProposalInvalidExtension() {
+// TestOnBlockProposal_InvalidProposal tests that a proposal which fails HotStuff validation.
```
AlexHentschel, line 339: similar naming suggestion as for collector compliance tests (see [my comment here](https://github.com/onflow/flow-go/pull/3294#issuecomment-1278325803))

AlexHentschel (approved):

AlexHentschel:
Thank you. Very nice work.

I have created [PR #3408](https://github.com/onflow/flow-go/pull/3408) with some stylistic revisions targeting your branch.

engine/collection/compliance/core.go
```diff
@@ -85,12 +89,13 @@ func NewCore(
 // OnBlockProposal handles incoming block proposals.
 func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.ClusterBlockProposal) error {
```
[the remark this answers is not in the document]
jordanschalm: added in https://github.com/onflow/flow-go/pull/3408

engine/common/follower/engine.go
jordanschalm: Hmm, we do return already 🤔

jordanschalm: fixed in https://github.com/onflow/flow-go/pull/3408

engine/common/follower/engine.go
jordanschalm: added in https://github.com/onflow/flow-go/pull/3408

engine/collection/compliance/core.go
```diff
@@ -197,24 +202,48 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
...
+					return nil
+				} else {
+					return fmt.Errorf("unexpected error notifying vote aggregator about invalid block: %w", err)
+				}
+			}
+
+			return nil
+		}
 		// unexpected error: potentially corrupted internal state => abort processing and escalate error
 		return fmt.Errorf("failed to process block %x: %w", blockID, err)
```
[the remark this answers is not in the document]
jordanschalm: Nice, good idea

jordanschalm:
> suggestion
Use terminology:
"header validation"
and "payload validation"

This suggestion is not much more accurate, IMO. It's true that "HotStuff validation" is _limited_ to the header, but it does not validate all parts of the header, and `state.Extend` validates parts of the header (height, parent ID, chain ID) in addition to the payload.

The separation is reflective of different layers of the logic, than parts of the data structure, to me. HotStuff validation, as in the portion of validation relevant to the HotStuff layer, seems right to me. We used to call the layer above HotStuff, which mainly enforces validity of the payloads, the "compliance layer". Maybe referring to this as "protocol state validation" is more accurate.

engine/consensus/compliance/core.go
jordanschalm: https://github.com/onflow/flow-go/pull/3294/commits/a2c8220c13e22450ac7bdbd1df224366bc917212

engine/collection/compliance/core.go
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3294/commits/a2c8220c13e22450ac7bdbd1df224366bc917212

engine/consensus/compliance/core_test.go
jordanschalm: See https://github.com/onflow/flow-go/pull/3294#issuecomment-1282354892

jordanschalm:
I think the failing consensus tests here would be fixed by the addition of a block message queue to `VoteAggregator` in https://github.com/onflow/flow-go/pull/3277

durkmurder:
> I think the failing consensus tests here would be fixed by the addition of a block message queue to `VoteAggregator` in #3277

Yup, both https://github.com/onflow/flow-go/pull/3277 and https://github.com/onflow/flow-go/pull/3357 fix flaky behavior in different tests.

AlexHentschel:
Thanks @jordanschalm  for [explaining the motivation behind the naming convention/ terminology](https://github.com/onflow/flow-go/pull/3294#issuecomment-1282354892). Makes sense. 

Very nice PR altogether!

jordanschalm:
bors merge

Issue #3157, opened 2022-09-01
Loader - New GithubRunner Integration


AndrewM-SDET:
## Problem Definition

With the new self-hosted github runner the Loader will now be able to run as part of CI on a set schedule. There is some work in adding appropriate yml files to dapperlabs/flow-go and adding scripts so that the tests can be run once on the new runner. 

## Proposed Solution

Add new yml to the private repo and write any additional scripts required in order to run tests on a predefined schedule (starting at 1 hour). 

## Definition of Done

There is a scheduled job that will run the performance tests every hour pointed at the head of master. 
Note: We are okay with running the same test even if there have been no new PRs merged into master as we already see fluctuations in TPS when using the same code and the additional data could be very revealing of any possible intermittent issues.

AndrewM-SDET:
On Friday @sjonpaulbrown set up a github runner for us, this requires our YML file to live in the private repo dapperlabs/flow-go. Today we are setting up that file and working to get it to pull its code from onflow/flow-go for initial testing.

PR #3299, opened 2022-09-28
#3157 Updates to integrate loader with self-hosted github runner

AndrewM-SDET:


integration/benchmark/cmd/ci/main.go
```diff
@@ -281,3 +282,18 @@ func sendDataToBigQuery(
 	}
 	return nil
 }
+
+func outputLocalData(slices []dataSlice) {
+	jsonText, err := json.MarshalIndent(slices, "", "    ")
+	if err != nil {
+		println("Error converting slice data to json")
```
gomisha, line 289: this is going to silently ignore the json marshaling error - can we use `log.Fatal()` instead?

integration/localnet/bootstrap.go
```diff
@@ -310,7 +310,7 @@ func prepareServiceDirs(role string, nodeId string) (string, string) {
 	if nodeId != "" {
 		profilerDir = "./" + filepath.Join(ProfilerDir, role, nodeId)
 	}
-	err = os.MkdirAll(profilerDir, 0755)
+	err = os.MkdirAll(profilerDir, 0777)
```
gomisha, line 313: can you add a comment on why we're giving full access to all users?

integration/localnet/bootstrap.go
```diff
@@ -401,7 +401,7 @@ func prepareExecutionService(container testnet.ContainerConfig, i int, n int) Se
 
 	// create the execution state dir for the node
 	trieDir := "./" + filepath.Join(TrieDir, container.Role.String(), container.NodeID.String())
-	err := os.MkdirAll(trieDir, 0755)
+	err := os.MkdirAll(trieDir, 0777)
```
gomisha, line 404: can you add a comment on why we're giving full access to all users?

integration/benchmark/cmd/ci/main.go
```diff
@@ -281,3 +282,18 @@ func sendDataToBigQuery(
...
+	}
+
+	// output human-readable json blob
+	timestamp := time.Now()
+	fileName := fmt.Sprintf("tps-results-%v.json", timestamp.Format("2006-02-01 15:04"))
+	err = os.WriteFile(fileName, jsonText, 0666)
+	if err != nil {
+		fmt.Println(err)
```
gomisha, line 297: this is going to silently ignore the json marshaling error - can we use log.Fatal() instead?

integration/localnet/Makefile
```diff
@@ -70,6 +70,11 @@ start:
 	DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 docker-compose -f docker-compose.nodes.yml build
 	DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 docker-compose -f docker-compose.nodes.yml up -d
 
+.PHONY: start-thin
+start-thin:
```
gomisha, line 74: can you add a comment on what a "thin start" means and why we need it?

cmd/Dockerfile
```diff
@@ -70,3 +70,5 @@ FROM gcr.io/distroless/base-debian11 as production-transit-nocgo
 COPY --from=build-transit-production-nocgo /app/app /bin/app
 
 ENTRYPOINT ["/bin/app"]
+
+FROM build-setup as environment-clean
```
AndrewM-SDET, line 74: This thin docker image exists because I was unable to run a docker build on build-setup without major extra work, this thin image does all the same work but is available directly in the Makefile

integration/benchmark/cmd/ci/main.go
AndrewM-SDET: Updated

integration/localnet/bootstrap.go
AndrewM-SDET: This was an attempt at resolving this issue, reverted now that we have a more valid solution.

integration/benchmark/cmd/ci/main.go
AndrewM-SDET: Done

integration/localnet/Makefile
AndrewM-SDET: Added, this resolves the port collision we were seeing on the runner.

AndrewM-SDET:
bors retry

AndrewM-SDET:
bors retry

integration/benchmark/cmd/ci/main.go
```diff
@@ -103,9 +103,6 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	sp := benchmark.NewStatsPusher(ctx, log, pushgateway, "loader", prometheus.DefaultGatherer)
```
pattyshack, line 106: why was this deleted?  do you still need it for non-localDev mode?

pattyshack (approved):

integration/benchmark/cmd/ci/main.go
AndrewM-SDET: When running in CI we are not using prometheus and because of how the self-hosted runner works actually don't run the prometheus node at all (see `start-thin` in the Makefile). If we leave it in then the output is full of warnings instead of our data. This still exists in the manual version of the loader for when local testing is needed and prometheus is actually accessible.

SaveTheRbtz: Do we have a Prometheus setup for this environment?  Can we just send the data there?

SaveTheRbtz: (If there is no Prometheus setup there, you can consider hiding it behind the same `localDev` flag.)

integration/localnet/Makefile
```diff
@@ -71,6 +73,13 @@ start:
 	DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 docker-compose -f docker-compose.nodes.yml build
```
SaveTheRbtz, line 73: you can consider using `$(MAKE) start-thin` here

SaveTheRbtz: (or possibly even renaming it so there is `start-with-metrics` and `start`.)

integration/benchmark/cmd/ci/main.go
AndrewM-SDET: There is no Prometheus on the CI environment, will hide this behind the localDev flag.

integration/localnet/Makefile
AndrewM-SDET: I think we want to keep `start` as the main entrypoint but I can use `$(MAKE) start-thin` in start for reuse/clarity. Change incoming.

AndrewM-SDET:
bors merge

PR #3300, opened 2022-09-28
Replace error/failure structs with CodedError/CodedFailure (Part 2)

pattyshack:
Note that I've left CadenceRuntimeError, FrozenAccountError and InvalidProposalSeqNumberError alone since these require special handling. (I'll deal with these in a followup PR)

### DIFF
--- a/fvm/errors/base.go
+++ b/fvm/errors/base.go
@@ -1,8 +1,6 @@
 package errors
 
 import (
-	"fmt"
-
 	"github.com/onflow/cadence/runtime"
 
 	"github.com/onflow/flow-go/model/flow"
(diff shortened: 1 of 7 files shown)


fvm/errors/base.go
```diff
@@ -66,38 +51,21 @@ func NewInvalidLocationErrorf(
...
+// valid value.
+func NewValueErrorf(
+	valueStr string,
+	msg string,
+	args ...interface{},
+) *CodedError {
+	return NewCodedError(
+		ErrCodeValueError,
+		"invalid value (%s): "+msg,
+		append([]interface{}{valueStr}, args...)...)
```
zhangchiqing, line 64: Would this work?
```suggestion
		valueStr, args...)
```

fvm/errors/todo_rm_emulator_hack.go
```diff
@@ -23,10 +23,10 @@ type InvalidEnvelopeSignatureError struct{ CodedError }
 //type InvalidProposalSeqNumberError struct{ CodedError }
 
 // DO NOT USE.
-//type InvalidPayloadSignatureError struct{ CodedError }
+type InvalidPayloadSignatureError struct{ CodedError }
```
zhangchiqing, line 26: what is this file for?

And it has a comment says "DO NOT USE"

pattyshack: (See the comment at the top of the file)

We need to keep the original struct around because we're in circular dependency hell.  The flow-go integration test depends on the emulator, which in turn depends on flow-go.  We can't delete fvm structs / apis until we've upgraded the emulator, and we can't do that until after a spork.  fvm is littered with 10+ similar TODOs =...(

fvm/errors/base.go
pattyshack: sadly no.  i wish it would

fvm/errors/todo_rm_emulator_hack.go
pattyshack: btw, for additional context, the emulator uses these struct for error classification.  But the checks are done incorrectly and the switch cases always ends up in the default branch.  I'm not breaking anything that wasn't already broken

zhangchiqing (approved):

pattyshack:
bors merge

pattyshack:
bors retry

PR #3303, opened 2022-09-29
Front-Load HotStuff Verification Pt. 2: Remove `MarkValid` API

jordanschalm:
This PR removes the `MarkValid` and `ValidDescendants` API methods:
* In https://github.com/onflow/flow-go/pull/3294, we move HotStuff validation before compliance checks and persistence - this means that all stored blocks have passed both HotStuff and compliance validation
* Therefore, there is no need for an additional validity flag - all blocks are valid by virtue of being persisted

⚠️ Depends on https://github.com/onflow/flow-go/pull/3294


durkmurder (approved):

zhangchiqing (approved):

jordanschalm:
bors merge

state/protocol/events.go
```diff
@@ -31,6 +31,8 @@ type Consumer interface {
 	// chain and its source of randomness is available).
 	// Formally, this callback is informationally idempotent. I.e. the consumer
 	// of this callback must handle repeated calls for the same block.
+	// TODO trigger this on block insertion (used to be on MarkValid)
+	//  - don't trigger for root block or below
 	BlockProcessable(block *flow.Header)
```
AlexHentschel, line 36: ## ⚠️ 
will this be fixed in a downstream PR? IIUC, other node roles depend on this callback for their business logic (like Verification nodes checking blocks). I think we are currently not calling this at all anymore.

jordanschalm: No, this should be fixed here!

I've moved this event to block insertion and updated the tests to check for this event more clearly: https://github.com/onflow/flow-go/pull/3303/commits/2a82b71aa8d86cdd32f8d8df90c59d6d659735b5. This was caught by unit tests, but broader-scope tests where it wasn't easy to see what was going wrong.

state/protocol/badger/mutator.go
```diff
@@ -425,8 +425,21 @@ func (m *FollowerState) insert(ctx context.Context, candidate *flow.Block, last
...
+	if err != nil {
+		return fmt.Errorf("could not retrieve block header for %x: %w", parentID, err)
+	}
+	// root blocks and blocks below the root block are considered as "processed",
+	// so we don't want to trigger `BlockProcessable` event for them.
+	var rootHeight uint64
+	err = m.db.View(operation.RetrieveRootHeight(&rootHeight))
+	if err != nil {
+		return fmt.Errorf("could not retrieve root block's height: %w", err)
+	}
```
AlexHentschel, line 441: I think `rootHeight` is a constant, that should not change while the node is up, right? Therefore, I think we should run this query in the `FollowerState` constructor and cache the result. We can then remove the query here and also in the [`guaranteeExtend` method](https://github.com/onflow/flow-go/blob/2401a43f30735ffe1527befd46681895c4d70282/state/protocol/badger/mutator.go#L253)

AlexHentschel (approved):
looks good.

state/protocol/events.go
AlexHentschel: Thanks for the fix. I noticed that  we are querying the root height for each inserted block twice, despite this being a constant. Please see [my comment below](https://github.com/onflow/flow-go/pull/3303#discussion_r1004980707)

state/protocol/badger/mutator.go
jordanschalm: Yes, good point. Added a cache in https://github.com/onflow/flow-go/pull/3303/commits/7cfdba093264fd6b1e1016333cf0490147a1d4f6 https://github.com/onflow/flow-go/pull/3303/commits/fd59222414bfd27e0c48eddbb684713447eef3f4

jordanschalm:
bors merge

PR #3305, opened 2022-09-29
Automating generations of values.yml

LLsq2:
BN2 creation requires a dynamically created values.yml file for helm to bring up an environment.
This PR takes bootstrapped information and dynamically generates a values file with the correct NodeIDs, also allows for the env to be dynamically sized.


SaveTheRbtz:
Any reason to not use the https://pkg.go.dev/gopkg.in/yaml.v3#Marshal ?

integration/benchnet2/automate/bootstrap_automation_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"log"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateTestTemplates(t *testing.T) {
+	fmt.Printf("Starting tests")
+	expectedValues := textReader("templates/test_templates/expected_values.yml")
```
gomisha, line 14: Isn't this using `textReader` to calculate the expected values? But aren't we testing `textReader`? How can we trust that the results are correct from `textReader`?

integration/benchnet2/automate/bootstrap_automation_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateTestTemplates(t *testing.T) {
+	fmt.Printf("Starting tests")
```
gomisha, line 13: no need for logging messages in tests

sjonpaulbrown:
In general, I think we should take a different approach to the creation and generation of the `values.yml` file. Currently, we are creating the file in an "unstructured" manner by appending to the file itself. This is very difficult to maintain and could result in issues when modifying the data structure and inputs that need to be templated.

I would recommend that we take one of two possible approaches:
1. Model the data using structs and write the data as we do in the[ localnet bootstrap process](https://github.com/onflow/flow-go/blob/master/integration/localnet/bootstrap.go#L551-L570)
2. Standardize a given set of variables for templates and use [go templating](https://pkg.go.dev/text/template) to inject them into a template that will be written

The advantage of using https://pkg.go.dev/text/template would be that we could leverage the templating engine to structure the data while injecting "standard" inputs that may or may not be used. This makes it very easy to create new templates that require no code changes, and subsequently, it allows for flexible definition of specific templates for different usecases.

As an example, if we define a "standard" set of inputs that is a combination of the bootstrap info and variables injected from the CLI, we can create different templates that have statically defined configuration rather than dynamically generated values. This sort of functionality would allow us to create different templates for different usecases such as BFT testing, performance testing, etc. Furthermore, it makes it incredibly flexible as we could "choose" the template by specifying a template name on the CLI.

Overall, I think either of the above approaches will result in a more flexible and maintainable solution. While some of the templating may eventually be able to go away, we may inevitably need to support the dynamic generation of different `values.yml` files, and it is important that the approach is flexible and maintainable.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+
+	executionTemplatePath := templateFolder + EXECUTION_TEMPLATE
+	executionNodeMap := nodeStruct("execution", executionTemplatePath, DEFAULT_EXECUTION_IMAGE, nodesData, nodeConfig)
+	var executionNodes = NodesDefs{Defaults: *nodeTypeResources, Nodes: executionNodeMap}
+
+	verificationTemplatePath := templateFolder + VERIFICATION_TEMPLATE
+	verificationNodeMap := nodeStruct("verification", verificationTemplatePath, DEFAULT_VERIFICATION_IMAGE, nodesData, nodeConfig)
+	var verificationNodes = NodesDefs{Defaults: *nodeTypeResources, Nodes: verificationNodeMap}
+
+	return &Values{Branch: "fake-branch", Commit: "123456", Defaults: EmptyStruct{}, Access: accessNodes, Collection: collectionNodes,
```
sjonpaulbrown, line 117: The `Branch` and `Commit` do not need to be hard coded in the return. Ideally, these should be passed through via CLI args.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return target
+}
+
+func marshalToYaml(source interface{}, outputFilePath string) {
+	yamlData, err := yaml.Marshal(source)
+	if err != nil {
+		log.Fatal(err)
+	}
```
sjonpaulbrown, line 194: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	return doc.String()
+}
+
+func createTemplate(templatePath string) *template.Template {
+	template, err := template.New("todos").Parse(string(textReader(templatePath)))
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 175: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	file.Close()
+}
+
+func replaceTemplateData(template *template.Template, data ReplacementData) string {
+	var doc bytes.Buffer
+	err := template.Execute(&doc, data)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 167: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return file
+}
+
+func writeYamlBytesToFile(filepath string, yamlData []byte) {
+	file, err := os.OpenFile(filepath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 152: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	_, err = file.Write(yamlData)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 157: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+		nodeMap[name] = *nodeStruct
+	}
+
+	return nodeMap
+}
+
+func textReader(path string) []byte {
+	file, err := os.ReadFile(path)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 143: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+	nodeConfig["consensus"] = 0
+	nodeConfig["execution"] = 0
+	nodeConfig["verification"] = 0
+
+	nodeMap := map[string]Node{}
+	nodeNameRe := regexp.MustCompile(`\w{6,}\d{1,3}`)
+	nodeTypeRe := regexp.MustCompile(`\D{6,}`)
+	for _, node := range nodes {
+		name := nodeNameRe.FindStringSubmatch(node.Address)
+		nodeType := nodeTypeRe.FindStringSubmatch(node.Address)
```
sjonpaulbrown, line 85: I don't think we need a regex for this. This is available at `node.Role`.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+		log.Fatal(err)
+	}
+
+	return template
+}
+
+func unmarshalToStruct(source string, target interface{}) interface{} {
+	err := yaml.Unmarshal([]byte(source), target)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 184: We need to return the error here after logging it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// Loads node info json file and returns unmarshaled struct and node counts
+func loadNodeJsonData(nodeInfoPath string) (map[string]Node, map[string]int) {
+	jsonFileBytes := textReader(nodeInfoPath)
+
+	var nodes []Node
+	err := json.Unmarshal(jsonFileBytes, &nodes)
+	if err != nil {
+		log.Fatal(err)
```
sjonpaulbrown, line 70: We need to return this error and handle it.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,196 @@
...
+func unmarshalToStruct(source string, target interface{}) interface{} {
+	err := yaml.Unmarshal([]byte(source), target)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	return target
+}
+
+func marshalToYaml(source interface{}, outputFilePath string) {
```
sjonpaulbrown, line 190: Could we log the `outpFilePath` that this file will be written to so that the user understands where the file is being written to? This would be valuable for troubleshooting and local testing.

sjonpaulbrown (changes requested):
Do we have plans to ensure that the templating of the nodes can be flexible so that we can dynamically use different templates based on input from a CLI?

My primary concern with the current approach to the statically defined templates is that they will be difficult to alter. If we were to collapse all of templates into a single template, we could push the templating complexity to the individual template. This would allow us to dynamically "choose" which template we want to use while injecting in the same variables. 

In general, I think this would be a relatively simple change as we have all of the data loaded that would need to be templated. We would just need to move the "looping" logic to the template for each node type.

integration/benchnet2/automate/bootstrap_automation.go
```diff
@@ -0,0 +1,230 @@
...
+		nodeType := nodeTypeRe.FindStringSubmatch(node.Address)
+		nodeMap[name[0]] = node
+		nodeConfig[nodeType[0]]++
+	}
+
+	return nodeMap, nodeConfig, err
+}
+
+// Builds and returns values yaml struct
+func buildValuesStruct(templateFolder string, nodesData map[string]Node, nodeConfig map[string]int, branch string, commit string,
```
gomisha, line 109: does this assume there will a single node of each type? What if I wanted a different number of access and verification nodes?

gomisha:
Replaced by https://github.com/onflow/flow-go/pull/3488

PR #3307, opened 2022-09-29
[CI] Increase lint timeout

peterargue:
The lint step in CI occasionally [fails with a timeout ](https://github.com/onflow/flow-go/actions/runs/3152898274/jobs/5128702326). This PR increases the timeout to 3 minutes from the default of 1 minute.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       with:
         # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
         version: v1.49
-        args: -v --build-tags relic
+        args: -v --build-tags relic --timeout 180
         working-directory: ${{ matrix.dir }}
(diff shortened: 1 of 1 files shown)


SaveTheRbtz:
Interesting, I though we have it already set to 10m: https://github.com/onflow/flow-go/blob/master/.golangci.yml#L2

Are we ignoring this config for some reason?

peterargue:
I think the issue is that the `golangci-lint` tool itself is timing out. It has a default 60 second timeout

gomisha (approved):

SaveTheRbtz:
@peterargue but `.golangci.yml` is the config file for this tool, so it should pick up the `10m` timeout already.  Or it doesn't do that for some reason?

peterargue:
@SaveTheRbtz is right. this change doesn't modify the timeout in the way I expected. going to close this out for now

PR #3308, opened 2022-09-29
fix Windows build

SaveTheRbtz:
`syscall.Stat_t` is linux-only and would fail on Windows and Plan9.  Separate codepaths into unix/windows.

Verified with:
```
GOOS=windows GOARCH=amd64 go build ./utils/io/...
```

PS. @gomisha given that `flow-cli` [depends on windows builds](https://github.com/onflow/flow-cli/actions/runs/3136555718/jobs/5093616496#step:4:227) maybe we can add a windows build on CI?

### DIFF
--- a/utils/io/file.go
+++ b/utils/io/file.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"syscall"
 )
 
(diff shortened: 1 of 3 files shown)


turbolent (approved):
Nice!

jordanschalm (approved):

SaveTheRbtz:
bors merge

PR #3314, opened 2022-09-29
[Networking] Override timeout for unicast Connect

SaveTheRbtz:
Connect would eventually call dial that can block.  At the same time `CreateStream` can use a pretty aggressive timeout of up to 16 minutes: https://github.com/onflow/flow-go/blob/8efd7b31b112e8e54f2ae7e1148084afd51ff5ef/network/p2p/middleware/middleware.go#L377

Connection timeout is currently set to 1s which is up for debate, but should be enough for ~5-10 roundtrips around the world.


SaveTheRbtz:
Not needed based on a brief debug-log analysis

PR #3315, opened 2022-09-30
[Execution] Adding more logging and a fix to block upload retry

Tonix517:
Some minor tunings to block upload retry logic:

- adding more logging to upload behavior for easier tracking for DPS debugging purposes
- fixing the improper badger obj init order (thanks @peterargue)

### DIFF
--- a/cmd/execution_builder.go
+++ b/cmd/execution_builder.go
@@ -277,8 +277,8 @@ func (exeNode *ExecutionNode) LoadGCPBlockDataUploader(
 		exeNode.events = storage.NewEvents(node.Metrics.Cache, node.DB)
 		exeNode.commits = storage.NewCommits(node.Metrics.Cache, node.DB)
 		exeNode.computationResultUploadStatus = storage.NewComputationResultUploadStatus(node.DB)
-		exeNode.collections = storage.NewCollections(node.DB, exeNode.transactions)
 		exeNode.transactions = storage.NewTransactions(node.Metrics.Cache, node.DB)
+		exeNode.collections = storage.NewCollections(node.DB, exeNode.transactions)
 		exeNode.txResults = storage.NewTransactionResults(node.Metrics.Cache, node.DB, exeNode.exeConf.transactionResultsCacheSize)
 
 		// Setting up RetryableUploader for GCP uploader
--- a/engine/execution/computation/computer/uploader/retryable_uploader_wrapper.go
+++ b/engine/execution/computation/computer/uploader/retryable_uploader_wrapper.go
@@ -140,6 +140,8 @@ func (b *BadgerRetryableUploaderWrapper) RetryUpload() error {
 		go func(blockID flow.Identifier) {
 			defer wg.Done()
 
+			log.Debug().Msgf("retrying upload for computation result of block %s", blockID.String())
+
 			var cr_err error
 			retComputationResult, err := b.reconstructComputationResult(blockID)
 			if err != nil {
@@ -152,8 +154,10 @@ func (b *BadgerRetryableUploaderWrapper) RetryUpload() error {
 			// Do Upload
 			if cr_err = b.uploader.Upload(retComputationResult); cr_err != nil {
 				log.Error().Err(cr_err).Msgf(
-					"Failed to update ComputationResult with BlockID %s", blockID)
+					"Failed to re-upload ComputationResult with BlockID %s", blockID)
 				retErr = cr_err
+			} else {
+				log.Debug().Msgf("computation result of block %s was successfully re-uploaded", blockID.String())
 			}
 
 			b.metrics.ExecutionComputationResultUploadRetried()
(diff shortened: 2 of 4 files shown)


cmd/execution_builder.go
```diff
@@ -277,8 +277,8 @@ func (exeNode *ExecutionNode) LoadGCPBlockDataUploader(
 		exeNode.events = storage.NewEvents(node.Metrics.Cache, node.DB)
 		exeNode.commits = storage.NewCommits(node.Metrics.Cache, node.DB)
 		exeNode.computationResultUploadStatus = storage.NewComputationResultUploadStatus(node.DB)
-		exeNode.collections = storage.NewCollections(node.DB, exeNode.transactions)
 		exeNode.transactions = storage.NewTransactions(node.Metrics.Cache, node.DB)
+		exeNode.collections = storage.NewCollections(node.DB, exeNode.transactions)
```
Tonix517, line 281: somehow my local testings (localnet) didn't catch this failure..

peterargue: some of these should already be available via `exeNode.Storage.*` 
https://github.com/onflow/flow-go/blob/master/cmd/scaffold.go#L723

Tonix517: confirmed with @m4ksio , in this context we are using the storage instances from `ExecutionNode` itself instead from scaffold. Also in `execution_builder.go` the FlowNodeBuilder.Build() is not called as in other types of node builders (access\observer\follower_builder.go) in which the storage instances in scaffold are initialized.

peterargue: `FlowNodeBuilder.Build()` is called [here](https://github.com/onflow/flow-go/blob/master/cmd/execution/main.go#L18-L18)

Are we using new versions intentionally here? I think using existing instances is best when possible as it uses less resources and avoids gotcha's in instantiation like we hit here.

Tonix517: yes that's what i meant. `ExecutionNodeBuilder` has all module\components initialization task in `LoadComponentsAndModules()` instead of a `Build()` as other node builders where they call `FlowNodeBuilder.Build()`. 

However to get all `exeNode.Storage.*` ready when needed like at this line here, `FlowNodeBuilder.Build()` needs to be called before node-specific initialization(in our case `LoadComponentsAndModules()`). Sth can be refactored more in this part but I'd say we can do it in a separate PR.

Tonix517: synced offline with @peterargue. so `exeNode.Storage.*` should have been ready in `LoadComponentsAndModules()` since those `Module()` and `Component()` are actually callback. But we do need to revisit why `ExecutionNodeBuilder` does not reuse what's already available in `FlowNodeBuilder`. Some clean up can be done here but probably in a separate PR.

m4ksio (approved):

engine/execution/computation/computer/uploader/retryable_uploader_wrapper.go
```diff
@@ -140,6 +140,8 @@ func (b *BadgerRetryableUploaderWrapper) RetryUpload() error {
 		go func(blockID flow.Identifier) {
 			defer wg.Done()
 
+			log.Debug().Msgf("retrying upload for computation result of block %s", blockID.String())
```
pattyshack, line 143: you need to check for blockID == nil here (and elsewhere) as well?

pattyshack (approved):

engine/execution/messages.go
```diff
@@ -50,3 +50,11 @@ func (cr *ComputationResult) AddComputationUsed(inp uint64) {
 func (cr *ComputationResult) AddStateSnapshot(inp *delta.SpockSnapshot) {
 	cr.StateSnapshots = append(cr.StateSnapshots, inp)
 }
+
+func (cr *ComputationResult) GetBlockID() *flow.Identifier {
+	if cr.ExecutableBlock != nil && cr.ExecutableBlock.Block != nil {
+		availBlockID := cr.ExecutableBlock.Block.ID()
```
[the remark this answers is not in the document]
Tonix517: the purpose of this function is simply to handle the case when `cr.ExecutableBlock` or `cr.ExecutableBlock.Block` is nil, because both of them are pointers. If they will never be `nil`, I agree we will never need it, however they expose risks that in future changes they might be initialized\set to nil.

engine/execution/computation/computer/uploader/uploader.go
```diff
@@ -69,6 +69,11 @@ func (a *AsyncUploader) Upload(computationResult *execution.ComputationResult) e
 		a.metrics.ExecutionBlockDataUploadStarted()
 		start := time.Now()
 
+		blockID := computationResult.GetBlockID()
+		if blockID != nil {
+			a.log.Debug().Msgf("computation result of block %s is being uploaded", blockID.String())
```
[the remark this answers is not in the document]
Tonix517: updated. I will take the existing assumption that `computationResult.ExecutableBlock` will never be `nil`, thus the code now looks less confusing.

engine/execution/messages.go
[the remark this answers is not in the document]
Tonix517: discussed offline - i will simply take the existing assumption that `computationResult.ExecutableBlock` will never be `nil`. Now this PR is much simpler.

Tonix517:
bors merge

Tonix517:
bors retry

PR #3318, opened 2022-10-02
SignatureAlgorithm's indexes to match its Cadence definition

SiriusDely:
SignatureAlgorithm's indexes to match its Cadence definition at https://developers.flow.com/cadence/language/crypto#signing-algorithms

Currently causing issue on https://github.com/onflow/flow-go-sdk/blob/7c53dc1/crypto/crypto.go#L45 `crypto.StringToSignatureAlgorithm("ECDSA_P256")` resulting into 2 instead of 1. And then using the result for https://github.com/onflow/cadence `SignatureAlgorithm(rawValue: 2)` giving us back `SignatureAlgorithm.ECDSA_secp256k1` instead of `SignatureAlgorithm.ECDSA_P256`.

### DIFF
--- a/crypto/types.go
+++ b/crypto/types.go
@@ -17,17 +17,17 @@ type SigningAlgorithm int
 const (
 	// Supported signing algorithms
 	UnknownSigningAlgorithm SigningAlgorithm = iota
-	// BLSBLS12381 is BLS on BLS 12-381 curve
-	BLSBLS12381
 	// ECDSAP256 is ECDSA on NIST P-256 curve
 	ECDSAP256
 	// ECDSASecp256k1 is ECDSA on secp256k1 curve
 	ECDSASecp256k1
+	// BLSBLS12381 is BLS on BLS 12-381 curve
+	BLSBLS12381
 )
 
 // String returns the string representation of this signing algorithm.
 func (f SigningAlgorithm) String() string {
-	return [...]string{"UNKNOWN", "BLS_BLS12381", "ECDSA_P256", "ECDSA_secp256k1"}[f]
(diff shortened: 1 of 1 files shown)


tarakby:
The signature algorithms defined in Cadence and the ones defined in `flow-go/crypto`  are not necessarily correlated. `flow-go/crypto` may be extended to support more signature algorithms in the future while Cadence may not. This because the `flow-go/crypto`'s main purpose is to provide the tools necessary for Flow node implementation, which requires more cryptography tools than what Cadence offers. 
Even if we update `flow-go/crypto` with this PR, there will be no guarantee that the enum will not change in the future. Therefore it's preferable that Cadence definitions remain abstracted from `flow-go` evolutions. 

I agree though that this can cause confusions, now that the signature algorithms have different definitions in each tool (Cadence, transaction format, flow-go). Note that the transaction format is also different from Cadence: https://developers.flow.com/learn/concepts/accounts-and-keys#supported-signature--hash-algorithms

SiriusDely:
Thanks for the explanation, it really helps. So these SignatureAlgorithm & HashAlgorithm enums, although it have similar names on those repos, however they cannot pass their own repo's boundaries to be used in relation to each other.

In this case, I think we need similar `cadence/runtime/sema.StringToSignatureAlgorithm("ECDSA_P256")` helper function to be defined on https://github.com/onflow/cadence.

tarakby:
cc @turbolent for visibility on this

turbolent:
@SiriusDely 

The definitions inside of Cadence, like `sema` should be considered implementation details, not publicly supported ABI. Instead, use the definitions in Cadence's `runtime` package. 

Once you use the types in `runtime`, you can convert them to/from flow-go `crypto` types using the conversion functions in https://github.com/onflow/flow-go/blob/master/fvm/crypto/crypto.go. 

HTH. Let me know if you still have problems.

PR #3326, opened 2022-10-04
[Checkpoint V6] Checkpoint v6 support v5

zhangchiqing:
This PR allows ledger to support both V5 and V6 checkpointing


cmd/util/cmd/compare-checkpoint/cmd.go
```diff
@@ -0,0 +1,57 @@
...
+
+	file2, err := os.Open(flagCheckpoint2)
+	if err != nil {
+		panic(err)
+	}
+	defer func() {
+		_ = file2.Close()
+	}()
+
+	err = wal.CompareV5(file1, file2)
```
zhangchiqing, line 53: The compare-checkpoint file is for comparing two v5 checkpoint file. Useful for round-trip testing

ledger/complete/mtrie/flattener/decode.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/onflow/flow-go/ledger"
+	"github.com/onflow/flow-go/ledger/common/hash"
+)
+
+// DecodeNodeData reads the node data from reader, and uses scratch as reusable buffer for holding data read from disk.
+// it returns raw data so that they are allocated on stack instead of heap, useful when reading tons of nodes from a
+// checkpoint file.
+// any error returned means the file is malformated.
+func DecodeNodeData(reader io.Reader, scratch []byte) (
```
zhangchiqing, line 16: This function is extracted from The encoding/ReadNode function, so that it can be reused by compare-checkpoint cmd

ledger/complete/mtrie/flattener/decode.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Decode right child index (8 bytes)
+	rchildIndex = binary.BigEndian.Uint64(scratch[pos:])
+
+	return true, height, lchildIndex, rchildIndex, ledger.DummyPath, nil, nodeHash, nil
+}
+
+// DecodeTrieData reads the trie data and decode it.
+// any error returned means the file is malformated
+func DecodeTrieData(reader io.Reader, scratch []byte) (
```
zhangchiqing, line 109: This is extracted from encoding/ReadTrie function, so that it can be reused by compare-checkpoint cmd

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -386,6 +393,12 @@ func storeCheckpointSubTrie(
 	// be used to calculate CRC32 checksum
 	writer := NewCRC32Writer(closable)
 
+	// write file part index
```
zhangchiqing, line 396: This PR also adds the file part index to the subtrie part file

ledger/complete/wal/checkpoint_v6_writer.go
```diff
@@ -195,6 +196,12 @@ func storeTopLevelNodesAndTrieRoots(
 
 	writer := NewCRC32Writer(closable)
 
+	// write subTriesNodeCount
```
zhangchiqing, line 199: this PR also encodes sub trie node count into the top level part file.

fxamacker (commented):
Some questions/suggestions:

- Why does part index need to be encoded in the file?  Isn't index already in the filename?

- Maybe move `ledger/complete/wal/compare_v5.go` to `cmd/util/cmd/compare-checkpoint` directory because it's only used for debugging.

- I think `CompareV5` should have its own `DecodeNodeData` and `DecodeTrieData`, instead of using those functions in ledger package.  Decoding nodes and tries can change over time in ledger package, while `CompareV5` expects consistent decoding behavior for v5 format.

zhangchiqing:
>Maybe move ledger/complete/wal/compare_v5.go to cmd/util/cmd/compare-checkpoint directory because it's only used for debugging.

The round-trip test will also need this compare function as well.


>I think CompareV5 should have its own DecodeNodeData and DecodeTrieData, instead of using those functions in ledger package. Decoding nodes and tries can change over time in ledger package, while CompareV5 expects consistent decoding behavior for v5 format.

Not sure if we want to duplicate the code now just to prepare for future change. 

But for now, I've made a copy of the DecodeNodeData, as you see I didn't change the ReadNode function to use DecodeNodeData. 

Why?

For some reason, when I use DecodeNodeData in ReadNode, there is a test case always fails. I had to revert back

PR #3327, opened 2022-10-04
[Checkpoint V6] Add EOF check

zhangchiqing:
It's important to check EOF, because we read the node count from the end of the file and then re-read from the start of the file, preventing from manipulating the node count.


ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -520,3 +536,14 @@ func getNodeByIndex(subtrieNodes [][]*node.Node, topLevelNodes []*node.Node, ind
 	// +1 because first item is always nil
 	return topLevelNodes[offset+1], nil
 }
+
+// reachedEOF checks if the reader has reached end of file
+// it returns nil if reached EOF
+// any error returned are exception
+func reachedEOF(reader io.Reader) error {
+	_, err := reader.Read(make([]byte, 1))
```
Tonix517, line 544: i'd rather using `File.Stat()` to use file size check EOF. so if in future we need to call `read` after `reachedEOF`, we won't miss the byte read here.

pattyshack: you can't use File here since it's a io.Reader, which may or may not be a file.

Changing the argument to take an io.ReadSeeker instead of io.Reader would enable to seek position to determine EOF

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -520,3 +536,14 @@ func getNodeByIndex(subtrieNodes [][]*node.Node, topLevelNodes []*node.Node, ind
...
+	if errors.Is(err, io.EOF) {
+		return nil
+	}
+	return fmt.Errorf("expect to reach EOF, but didn't: %w", err)
```
fxamacker, line 548: `err` could be `nil` here.  Maybe just return a sentinel error if EOF isn't reached.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -491,3 +517,15 @@ func getNodeByIndex(subtrieNodes [][]*node.Node, topLevelNodes []*node.Node, ind
 	// +1 because first item is always nil
 	return topLevelNodes[offset+1], nil
 }
+
+// reachedEOF checks if the reader has reached end of file
+// it returns nil if reached EOF
+// any error returned are exception
+func reachedEOF(reader io.Reader) error {
```
m4ksio, line 524: maybe name like `ensureEOF` would be better name?

m4ksio (approved):

PR #3328, opened 2022-10-04
[Checkpoint v6] add more tests

zhangchiqing:
- Add more tests that tests storing and reading more complex tries.
- Add tests for round-trip testing (converting to v6 then back to v5 and compare the result against v5)


ledger/complete/wal/checkpoint_v6_test.go
```diff
@@ -82,10 +89,54 @@ func createSimpleTrie(t *testing.T) []*trie.MTrie {
...
+}
+
+func createMultipleRandomTries(t *testing.T) []*trie.MTrie {
+	tries := make([]*trie.MTrie, 0)
+	activeTrie := trie.NewEmptyMTrie()
+
+	// add tries with no shared paths
+	for i := 0; i < 100; i++ {
+		paths, payloads := randNPathPayloads(100)
+		activeTrie, _, err := trie.NewTrieWithUpdatedRegisters(activeTrie, paths, payloads, false)
```
fxamacker, line 117: Need to reuse `activeTrie` inside loop if we want new trie to be created from previous trie.
```suggestion
                var err error
		activeTrie, _, err = trie.NewTrieWithUpdatedRegisters(activeTrie, paths, payloads, false)
```

ledger/complete/wal/checkpoint_v6_test.go
```diff
@@ -82,10 +89,54 @@ func createSimpleTrie(t *testing.T) []*trie.MTrie {
...
+		require.NoError(t, err, "update registers")
+		tries = append(tries, activeTrie)
+	}
+
+	// add trie with some shared path
+	sharedPaths, payloads1 := randNPathPayloads(100)
+	activeTrie, _, err := trie.NewTrieWithUpdatedRegisters(activeTrie, sharedPaths, payloads1, false)
```
fxamacker, line 124:
```suggestion
	activeTrie, _, err = trie.NewTrieWithUpdatedRegisters(activeTrie, sharedPaths, payloads1, false)
```

ledger/complete/wal/checkpoint_v6_test.go
```diff
@@ -167,3 +218,130 @@ func TestWriteAndReadCheckpointV6(t *testing.T) {
...
+
+func TestWriteAndReadCheckpointV6MultipleTries(t *testing.T) {
+	unittest.RunWithTempDir(t, func(dir string) {
+		tries := createMultipleRandomTries(t)
+		fileName := "checkpoint-multi-file"
+		logger := unittest.Logger()
+		require.NoErrorf(t, StoreCheckpointV6Concurrent(tries, dir, fileName, &logger), "fail to store checkpoint")
+		decoded, err := ReadCheckpointV6(dir, fileName, &logger)
+		require.NoErrorf(t, err, "fail to read checkpoint %v/%v", dir, fileName)
+		requireTriesEqual(t, tries, decoded)
```
fxamacker, line 230: Maybe we can test checkpoint v6 roundtrip here:
- after `ReadCheckpointV6()` loads tries, call `StoreCheckpointV6Concurrent()` again to serialize tries to checkpoint files
- compare generated checkpoint files

ledger/complete/wal/checkpoint_v6_test.go
```diff
@@ -167,3 +218,130 @@ func TestWriteAndReadCheckpointV6(t *testing.T) {
...
+
+		require.NoErrorf(t, storeCheckpointV5(tries, dir, fileName, &logger), "fail to store checkpoint")
+		decoded, err := LoadCheckpoint(dir, fileName, &logger)
+		require.NoErrorf(t, err, "fail to load checkpoint")
+		requireTriesEqual(t, tries, decoded)
+	})
+}
+
+// test that converting a v6 back to v5 would produce the same v5 checkpoint as
+// producing directly to v5
```
fxamacker, line 326: Good idea!

fxamacker (approved):

ledger/complete/wal/checkpoint_v6_test.go
zhangchiqing: This has sort of tested by `TestCheckpointV6IsDeterminstic`. 

Since `ReadCheckpointV6` produces the same decoded tries as the original tries, then storing the tries again should be equivalent to storing the original tries again, which essentially to test if generating v6 is deterministic.

ledger/complete/wal/checkpoint_v6_test.go
```diff
@@ -167,3 +219,130 @@ func TestWriteAndReadCheckpointV6(t *testing.T) {
...
+		requireTriesEqual(t, tries, decoded)
+	})
+}
+
+// test running checkpointing twice will produce the same checkpoint file
+func TestCheckpointV6IsDeterminstic(t *testing.T) {
+	unittest.RunWithTempDir(t, func(dir string) {
+		tries := createMultipleRandomTries(t)
+		logger := unittest.Logger()
+		require.NoErrorf(t, StoreCheckpointV6Concurrent(tries, dir, "checkpoint1", &logger), "fail to store checkpoint")
```
m4ksio, line 240: Maybe we should run both function in two separate goroutines to rule out possible state sharing?

m4ksio (approved):

PR #3329, opened 2022-10-04
[Checkpoint v6] Control concurrency

zhangchiqing:


fxamacker (commented):
Some suggestions:

- Can we make number of workers configurable for checkpoint serialization?  

- Maybe remove `StoreCheckpointV6SingleThread()` (with hard-coded 1 worker) and `StoreCheckpointV6Concurrent()` (with hard-coded 16 workers), and directly use `StoreCheckpointV6` which receives number of workers as function parameter.

PR #3332, opened 2022-10-04
Modify SplitError's behavior.

pattyshack:
1. Instead of returning the unwrapped coded error, which loses a lot of error message information, always return the original error wrapped with the important error code.
2. For non-failure, use the root cause error code instead of the shallowest error code.


fvm/errors/errors.go
```diff
@@ -53,19 +50,17 @@ func findImportantCodedError(err error) (CodedError, bool) {
...
+			return coded, false
 		}
 
-		if coded.Code().IsFailure() {
+		var nextCoded CodedError
+		if !As(coded.Unwrap(), &nextCoded) {
 			return coded, false
 		}
+
+		coded = nextCoded
```
zhangchiqing, line 63: how does the value of nextCoded get updated? shouldn't this be `coded = coded.Unwrap()`?

pattyshack: nextCoded is updated by `As(coded.Unwrap(), &nextCoded)`.  The error is already unwrapped to the next nested coded error

janezpodhostnik (approved):

pattyshack:
bors merge

pattyshack:
bors merge

PR #3333, opened 2022-10-04
[Network] Log all inbound connections

peterargue:
Closes: https://github.com/dapperlabs/flow-internal/issues/1801

To ensure operators have enough information to deal with malicious peers, log the remote peer's IP:port for all connection attempts


network/p2p/connection/connGater.go
```diff
@@ -92,24 +92,27 @@ func (c *ConnGater) InterceptAccept(cm network.ConnMultiaddrs) bool {
...
+			Str("peer_id", p.Pretty()).
+			Str("remote_address", addr.RemoteMultiaddr().String()).
+			Logger()
+
 		if len(c.onInterceptSecuredFilters) == 0 {
-			c.log.Debug().
-				Str("peer_id", p.Pretty()).
-				Msg("allowing inbound connection intercept secured has no peer filters set")
+			logger.Info().Msg("inbound connection established")
+			logger.Debug().Msg("intercept secured has no peer filters set")
```
yhassanzadeh13, line 102: Suggestion: 

```go
		lg := c.log.With().
			Str("peer_id", p.String()).
			Str("remote_address", addr.RemoteMultiaddr().String()).
			Int("intercept_secure_filters_set", len(c.onInterceptSecuredFilters)).
			Logger()

		if len(c.onInterceptSecuredFilters) == 0 {
			lg.Info().Msg("inbound connection established")
			return true
		}
```
1. Let's rename local variable `logger` to `lg` hence not confusing it with the package `logger`. 
2. `peer.Pretty()` has been deprecated and we should use `peer.String()` instead.
3. It makes it cheaper to add the number of intercept secure filters as a key-value on the log entry than a separate log message.

peterargue: 2c7cd21

re: 3, I opted to just drop the filter length as it would add 30+ characters to every event and I don't it's necessary to include in an info level message

yhassanzadeh13 (approved):

peterargue:
bors merge

PR #3334, opened 2022-10-04
Ensure cadence don't access unexpected operations while parsing programs (part 1)

pattyshack:
I'll add more checks in followup PRs.  If all goes well, these checks should do nothing (If these checks triggered, then we have program cache invalidation bugs).


fvm/environment/account_creator.go
```diff
@@ -36,6 +36,35 @@ type BootstrapAccountCreator interface {
 	)
 }
 
+type ParseRestrictedAccountCreator struct {
+	txnState *state.TransactionState
+	impl     AccountCreator
```
Tonix517, line 41: will `ParseRestricted` check be applied to other operations too? if so it can be a generic decorator instead of just for AccountCreator. But it really depends on how it will be used.

Tonix517 (approved):

fvm/environment/account_creator.go
pattyshack: The ParsRestricted checks will be applied to everything, except Programs and ContractReader (and maybe Meter).

The different method signatures (in particular the arguments and return types) are interfering with using a generic decorator in general.  maybe missed something?

pattyshack: o, i guess i can use closures for some of these, if the return type is only error

Tonix517: yea i was thinking about closures. just curious to how code would look like with closures, if it is elegant enough, it may worth a try.

fvm/environment/account_creator.go
```diff
@@ -36,6 +36,35 @@ type BootstrapAccountCreator interface {
...
+}
+
```
zhangchiqing, line 43:
```suggestion
// Ensure ParseRestrictedAccountCreator is a AccountCreator
var _ AccountCreator = (*ParseRestrictedAccountCreator)(nil)
```

zhangchiqing (approved):

pattyshack:
bors merge

pattyshack:
bors retry

Issue #3325, opened 2022-10-04
[Checkpoint V6] Add magic id and version to sub trie file


zhangchiqing:
Addressing [this comments](https://github.com/onflow/flow-go/pull/3273#discussion_r986270264) from @fxamacker 

>I think different file types should contain their own unique magic id. This helps prevent accidental collisions (a part file's data being equal to checkpoint header file's magic id by coincidence).

PR #3335, opened 2022-10-04
[Checkpoint V6] Add version to part files

zhangchiqing:
Close https://github.com/onflow/flow-go/issues/3325

This PR adds magic code and version to checkpoint part files


Tonix517 (approved):
one question: how do we handle backward compatibility when we change file format? or we don't have to bcz we always create checkfile files first before reading new formats?

zhangchiqing:
>how do we handle backward compatibility when we change file format? or we don't have to bcz we always create checkfile files first before reading new formats?

This is the logic that checks version and use the corresponding function to read the format.
https://github.com/onflow/flow-go/blob/master/ledger/complete/wal/checkpointer.go#L623-L632

When we change format, we will bump up the format version.

fxamacker (approved):

PR #3338, opened 2022-10-05
[Benchmark] Add ability to auto-detect maximum TPS

SaveTheRbtz:
This PR adds an ability of finding the maximum TPS that the network can handle to the `ci` benchmark
It is currently using a simple AIMD (additive increase/multiplicative decrease) algorithm.

The algorithm starts with `-initial-tps` as a target.  Each time it is able to reach the target TPS, it increases the target by `additiveIncrease`. Each time it fails to reach the target TPS, it decreases the target by `multiplicativeDecrease` factor.  

There are two optimizations on top:
* To avoid oscillation and speedup conversion we skip the adjustment stage if TPS grew compared to the last round.  
* We do not shrink target TPS if it would bring them below the current TPS.

(We disable both optimizations if inflight transactions are above `maxInflight`)

Target TPS is always bounded by [`-initial-tps`, `-max-tps`].

Result wise, it is quite aggressive but reliably backs off once TPS "in" gets much higher than TPS "out":
Screen Shot 2022-10-05 at 11 10 31 PM 1

Depends on: #3292

### DIFF
--- a/integration/benchmark/cmd/ci/main.go
+++ b/integration/benchmark/cmd/ci/main.go
@@ -330,21 +330,6 @@ func adjustTPS(
 			inflight := currentSentTxs - int(currentTxs)
 			currentTPS := float64(currentTxs-lastTxs) / now.Sub(lastTs).Seconds()
 
-			// Do not touch target TPS if TPS rate incresed since last check.
-			if (currentTPS > float64(lastTPS)) && (inflight < int(maxInflight)) {
-				log.Info().
-					Float64("lastTPS", lastTPS).
-					Float64("currentTPS", currentTPS).
-					Int("inflight", inflight).
-					Msg("skipped adjusting TPS")
-
-				lastTxs = currentTxs
-				lastTPS = currentTPS
-				lastTs = now
-
-				continue
-			}
-
 			unboundedTPS := uint(math.Ceil(currentTPS))
 
 			// To avoid setting target TPS below current TPS,
@@ -356,6 +341,21 @@ func adjustTPS(
 
 				unboundedTPS = targetTPS + additiveIncrease
 			} else {
+				// Do not reduce the target if TPS incresed since the last round.
+				if (currentTPS > float64(lastTPS)) && (inflight < int(maxInflight)) {
+					log.Info().
+						Float64("lastTPS", lastTPS).
+						Float64("currentTPS", currentTPS).
+						Int("inflight", inflight).
+						Msg("skipped adjusting TPS")
+
+					lastTxs = currentTxs
+					lastTPS = currentTPS
+					lastTs = now
+
+					continue
+				}
+
 				unboundedTPS = uint(float64(targetTPS) * multiplicativeDecrease)
 			}
 			boundedTPS := boundTPS(unboundedTPS, minTPS, maxTPS)
--- a/integration/benchmark/cmd/ci/main.go
+++ b/integration/benchmark/cmd/ci/main.go
@@ -324,44 +324,40 @@ func adjustTPS(
 		select {
 		// NOTE: not using a ticker here since adjusting worker count in SetTPS
 		// can take a while and lead to uneven feedback intervals.
-		case now := <-time.After(interval):
+		case nowTs := <-time.After(interval):
 			currentSentTxs := lg.GetTxSent()
 			currentTxs := uint(lg.GetTxExecuted())
+
 			inflight := currentSentTxs - int(currentTxs)
-			currentTPS := float64(currentTxs-lastTxs) / now.Sub(lastTs).Seconds()
 			inflightPerWorker := inflight / int(targetTPS)
 
-			unboundedTPS := uint(math.Ceil(currentTPS))
+			skip, currentTPS, unboundedTPS := computeTPS(
+				lastTxs,
+				currentTxs,
+				lastTs,
+				nowTs,
+				lastTPS,
+				targetTPS,
+				inflight,
+				maxInflight,
+			)
 
-			// To avoid setting target TPS below current TPS,
-			// we decrease the former one by the multiplicativeDecrease factor.
-			//
-			// This shortcut is only applicable when current inflight is less than maxInflight.
-			if ((float64(unboundedTPS) >= float64(targetTPS)*multiplicativeDecrease) && (inflight < int(maxInflight))) ||
-				(unboundedTPS >= targetTPS) {
+			if skip {
+				log.Info().
+					Float64("lastTPS", lastTPS).
+					Float64("currentTPS", currentTPS).
+					Int("inflight", inflight).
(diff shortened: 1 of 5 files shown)


AndrewM-SDET (approved):

integration/localnet/Makefile
```diff
@@ -55,9 +55,10 @@ else
 endif
 
 # Creates a light version of the localnet with just 1 instance for each node type
+# (Except for consensus and execution nodes, which are set to 2)
 .PHONY: init-light
 init-light:
-	$(MAKE) -e COLLECTION=1 CONSENSUS=2 EXECUTION=1 VERIFICATION=1 ACCESS=1 NCLUSTERS=1 init
+	$(MAKE) -e COLLECTION=1 CONSENSUS=2 EXECUTION=2 VERIFICATION=1 ACCESS=1 NCLUSTERS=1 init
```
pattyshack, line 61: why are these numbers hardcoded?  should they be the same as the constants defined at the top of the file?

integration/benchmark/cmd/ci/main.go
```diff
@@ -52,9 +54,14 @@ const (
...
 	accessNodeAddress           = "127.0.0.1:3569"
 	pushgateway                 = "127.0.0.1:9091"
-	accountMultiplier           = 100
+	accountMultiplier           = 50
 	feedbackEnabled             = true
 	serviceAccountPrivateKeyHex = unittest.ServiceAccountPrivateKeyHex
+
+	// Auto TPS scaling constants
+	additiveIncrease       = 50
+	multiplicativeDecrease = 0.7
```
pattyshack, line 63: you may want to use an even larger number, say 0.8 or 0.9, if you to narrow the bound

integration/benchmark/cmd/ci/main.go
```diff
@@ -271,3 +298,96 @@ func sendDataToBigQuery(
...
+//
+// To avoid oscillation and speedup conversion we skip the adjustment stage if TPS grew
+// compared to the last round.
+//
+// Target TPS is always bounded by [minTPS, maxTPS].
+func adjustTPS(
+	lg *benchmark.ContLoadGenerator,
+	log zerolog.Logger,
+	interval time.Duration,
+	minTPS, maxTPS, maxInflight uint,
```
pattyshack, line 315: nit:  even through golang supports this style of arg declaration, could you specify the types individually (and on different lines)?  It makes skimming the method signature easier

integration/benchmark/cmd/ci/main.go
```diff
@@ -271,3 +298,96 @@ func sendDataToBigQuery(
...
+) error {
+	targetTPS := minTPS
+	lastTs := time.Now()
+	lastTPS := float64(minTPS)
+	lastTxs := uint(lg.GetTxExecuted())
+	for {
+		select {
+		case <-time.After(interval):
```
pattyshack, line 323: use a ticker?

integration/benchmark/cmd/ci/main.go
```diff
@@ -271,3 +298,96 @@ func sendDataToBigQuery(
...
+			currentSentTxs := lg.GetTxSent()
+			currentTxs := uint(lg.GetTxExecuted())
+			inflight := currentSentTxs - int(currentTxs)
+			currentTPS := float64(currentTxs-lastTxs) / math.Floor(time.Since(lastTs).Seconds())
```
pattyshack, line 327: the math seems wrong here.  drop the math.Floor?

integration/benchmark/cmd/ci/main.go
```diff
@@ -271,3 +298,96 @@ func sendDataToBigQuery(
...
+			if (currentTPS > float64(lastTPS)) && (inflight < int(maxInflight)) {
+				log.Info().
+					Float64("lastTPS", lastTPS).
+					Float64("currentTPS", currentTPS).
+					Int("inflight", inflight).
+					Msg("skipped adjusting TPS")
+
+				lastTxs = currentTxs
+				lastTPS = currentTPS
+				lastTs = time.Now()
```
pattyshack, line 339: you're using different "now" time within the loop (here, time.Since, etc).  maybe just use the one returned by `<-time.After(interval)`

integration/localnet/Makefile
SaveTheRbtz: these are overrides for the pre-defined ones to bootstrap the smalelst-viable-cluster.  We could reuse the default values from the top in cases where they match, but we do not want them to accidentally change in the future.

I can factor out the "2" constant, but it looks like it would make code less readable.

integration/benchmark/cmd/ci/main.go
```diff
@@ -271,3 +298,96 @@ func sendDataToBigQuery(
...
+	for {
+		select {
+		case <-time.After(interval):
+			currentSentTxs := lg.GetTxSent()
+			currentTxs := uint(lg.GetTxExecuted())
+			inflight := currentSentTxs - int(currentTxs)
+			currentTPS := float64(currentTxs-lastTxs) / math.Floor(time.Since(lastTs).Seconds())
+
+			// Do not touch target TPS if TPS rate incresed since last check.
+			if (currentTPS > float64(lastTPS)) && (inflight < int(maxInflight)) {
```
pattyshack, line 330: you probably want to have a fudge factor here.  we should probably increase tps if currentTPS is only slightly larger than lastTPS

pattyshack (commented):
any chance you can add a test for this?

pattyshack:
wrt testing, just the math part btw, i don't care about the loop

integration/benchmark/cmd/ci/main.go
SaveTheRbtz: ticker is tricky since SetTPS is a blocking call which means I would need to reset the ticker on every run, so I went with a manual `time.After`.
(Without it I would sometimes get "doubleticks" if TPS adjustment took too long.)

integration/benchmark/cmd/ci/main.go
SaveTheRbtz: fair enough-- I was using a pretty conservative constant from early TCP implementations, let's bump it up a bit.

integration/benchmark/cmd/ci/main.go
SaveTheRbtz: Though about that but it works quite well without it so i went for a simpler model.

integration/benchmark/cmd/ci/main.go
pattyshack: fair enough.  that's useful context.  maybe add a comment

SaveTheRbtz:
Added tests.

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #3310, opened 2022-09-29
Refactor load checkpoint

zhangchiqing:
Extracting changes to LoadCheckpoint for checkpointing v4

[Addressing comments](https://github.com/onflow/flow-go/pull/3273/files#r983940431)

PR #3342, opened 2022-10-05
[Checpoint V6] Revert "Refactor load checkpoint"

zhangchiqing:
Reverts onflow/flow-go#3310

Reverting this change, because:

1. We could [split the directory and filename from a file path](https://github.com/onflow/flow-go/pull/3336/files#diff-2ff6507ee260a1e37156b4cae4cc259e7688bbd0007993b20fa193824073105aR31)
2. [DPS depends on a single full path of the checkpoint file](https://github.com/onflow/flow-dps/blob/d119e8c713e9f058dc82d57048a186e7533b529f/service/loader/checkpoint.go#L47), they can keep passing a single path

### DIFF
--- a/cmd/util/cmd/checkpoint-list-tries/cmd.go
+++ b/cmd/util/cmd/checkpoint-list-tries/cmd.go
@@ -2,7 +2,6 @@ package checkpoint_list_tries
 
 import (
 	"fmt"
-	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


pattyshack (approved):

Tonix517 (approved):

zhangchiqing:
bors merge

Issue #3324, opened 2022-10-04
[Tech debt] error documentation for `p2p.IDTranslator`s


AlexHentschel:
Looking at the [`IDTranslator` interface](https://github.com/onflow/flow-go/blob/3539e21e900a85ac3a2786b16fed0753d9ddea3e/network/p2p/id_translator.go):
* the interface does not document any error returns
* implementations return generic errors leaving higher-level code completely in the dark whether the error is critical or non-critical  

The current implementation does not live up to our quality guidelines and best practises (see [CodingConventions](https://github.com/onflow/flow-go/blob/c07fdad6304260fc971cf9b74a3262d71720c00c/CodingConventions.md)), i.e. we have technical debt. Furthermore, the current implementation can actually loos safety-critical errors in the following code: https://github.com/onflow/flow-go/blob/c07fdad6304260fc971cf9b74a3262d71720c00c/network/p2p/translator/hierarchical_translator.go#L27-L34
* One of the `translators` could return a safety-critical error. But since there is no error documentation or error convention, the application logic at this level cannot differentiate critical from benign errors. It just puts all the errors into a `multierror`.
* Then, if one of the later translators finds a mapping from FlowID to PeerID, all previously accumulated errors will be dropped, including the safety-critical one. 

### suggestions
* LibP2P provides primitives (-> [example](https://github.com/onflow/flow-go/blob/c07fdad6304260fc971cf9b74a3262d71720c00c/network/p2p/translator/identity_provider_translator.go#L21)) that we could build on top
*  We need a relatively broad set of sentinel errors that cover all existing implementations. Something like the following could be used as a starting point:
    * `UnknownID`: id is unknown
    * `InvalidID` the format of the ID is syntactically invalid 
  
   These error returns should be specified in [`IDTranslator` interface](https://github.com/onflow/flow-go/blob/3539e21e900a85ac3a2786b16fed0753d9ddea3e/network/p2p/id_translator.go) as benign errors
* _All_ interface implementations should comply with the interface specification. This should be covered by tests where possible, 

### Design comment:

It seems we generate `flow.Identifier`s for non-staked nodes. I consider this a sub-optimal design, because only nodes that are part of the identity table have a `flow.Identifier`. It is not a big problem, but slightly unclean. Not sure if we can get around this problem by splitting up the interface into two: 
1. `flow.Identifier -> peer.ID` converter
2. `peer.ID -> flow.Identifier` converter

PR #3344, opened 2022-10-05
Alex/6401 blocking nodes on network level

AlexHentschel:
technical foundation for [issue #6401](https://github.com/dapperlabs/flow-go/issues/6401)

### High-level design
* Every time another node tries to establish a libP2P connection, libP2P consults the [`ConnectionGater`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/connmgr#ConnectionGater) during the handshake whether communication with the respective node is allowed
* [`ProtocolStateIDCache`](https://github.com/onflow/flow-go/blob/master/network/p2p/cache/protocol_state_provider.go) is already used by the [`DefaultLibP2PNodeFactory`](https://github.com/onflow/flow-go/blob/a7a69d24bb28a4c909376d3b94f0d5e56dbffe25/network/p2p/p2pbuilder/libp2pNodeBuilder.go#L61-L66) to implement a [`ConnectionGater`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/connmgr#ConnectionGater). Connections with nodes whose `Ejected` flag is true are denied.
* Hence, in this PR, I implemented a wrapper around [`ProtocolStateIDCache`](https://github.com/onflow/flow-go/blob/master/network/p2p/cache/protocol_state_provider.go), which overrides the `Ejected` flag of nodes on a blacklist to `true`

### Overview

**central technical contributions of this PR**:
* wrapper [`NodeBlocklistWrapper`](https://github.com/onflow/flow-go/blob/alex/6401_blocking_nodes_on_network_level/network/p2p/cache/node_blocklist_wrapper.go). This is a wrapper around the [`id.IdentityProvider` interface](https://github.com/onflow/flow-go/blob/b2b181b20d4b3bf4bc2ed8a2d443d7e4591a5175/module/id/id_provider.go#L18)
   * For nodes contained in a operator-specified blacklist, the wrapper sets the `Identity.Ejected` flag to `true`
   * I added the wrapper to all [`ProtocolStateIDCache`](https://github.com/onflow/flow-go/blob/alex/6401_blocking_nodes_on_network_level/network/p2p/cache/protocol_state_provider.go) instances for production nodes
* the [`NodeBlocklistWrapper`](https://github.com/onflow/flow-go/blob/alex/6401_blocking_nodes_on_network_level/network/p2p/cache/node_blocklist_wrapper.go) provides the method `Update` for concurrently updating the blacklist during normal operations without restart. Though, existing connections to newly blacklisted nodes might still be maintained. 
* The wrapper intrinsically persists its current blacklist in the data base and loads it at initialization. 
* Also, the method `ClearBlacklist()` for purging the blacklist concurrently during operations is implemented. 

The PR also cleans up various **misc technical debt** (probably the majority of line changes):
   * replace the deprecated method `peer.ID.Pretty()` by `peer.ID.String()`
   * consolidated imports (removed line breaks and legacy import of `github.com/onflow/flow-go/utils/binstat`)
   * made it explicit in code that structs implement interfaces. Example: `ProtocolStateIDCache` implements these interfaces https://github.com/onflow/flow-go/blob/b2b181b20d4b3bf4bc2ed8a2d443d7e4591a5175/network/p2p/cache/protocol_state_provider.go#L40-L42 which is now codified as part of the structs implementation
   * In the godoc for structs, added package name to references to other structs or interfaces. Example: 
https://github.com/onflow/flow-go/blob/b2b181b20d4b3bf4bc2ed8a2d443d7e4591a5175/network/p2p/translator/unstaked_translator.go#L14
may IDEs are able to directly resolve references in this format, so you can click on it, which aids ease of development  
   * added lots of goDoc, updated goDoc
   * added errors to the API specification of the [`IDTranslator`](https://github.com/onflow/flow-go/blob/alex/6401_blocking_nodes_on_network_level/network/p2p/id_translator.go), added several TODO's where implementations don't document any errors. This is a first step towards implementing https://github.com/onflow/flow-go/issues/3324 (subsequent work)

### Remaining work and open technical questions:

1. For [issue #6401](https://github.com/dapperlabs/flow-go/issues/6401), we still need the admin command to be implemented (@zhangchiqing was hoping you could help here with a follow-up PR)
2. While the [`DefaultLibP2PNodeFactory`](https://github.com/onflow/flow-go/blob/a7a69d24bb28a4c909376d3b94f0d5e56dbffe25/network/p2p/p2pbuilder/libp2pNodeBuilder.go#L61-L66) already instantiates a [`ConnectionGater`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/connmgr#ConnectionGater) including the desired filter for ejected nodes, there are potentially other places in the code, where we need to also add a `ConnectionGater` (?)
   * The [`FlowAccessNodeBuilder`](https://github.com/onflow/flow-go/blob/7cbc2f218dabd19533a65b31334da2994fa11a74/cmd/access/node_builder/access_node_builder.go#L982-L1001) does not have a `ConnectionGater`. 
   * neither does the [`ObserverServiceBuilder`](https://github.com/onflow/flow-go/blob/7cbc2f218dabd19533a65b31334da2994fa11a74/cmd/observer/node_builder/observer_builder.go#L843-L858)
   * or the [`FollowerServiceBuilder`](https://github.com/onflow/flow-go/blob/7cbc2f218dabd19533a65b31334da2994fa11a74/follower/follower_builder.go#L582-L597)

   While the network-external services (Observer and Follower) are certainly much less safety-critical from the perspective of the network, they might still benefit from the ability to block malicious access nodes? Not sure. But I feel strongly that at least the access node should allow an honest operator to block other malicious nodes. 

   Would appreciate guidance from @yhassanzadeh13 and @vishalchangrani here. Thanks


network/p2p/id_translator.go
```diff
@@ -1,17 +1,45 @@
...
+
+	// ErrUnknownId indicates that the given ID (either `peer.ID` or `flow.Identifier`) is unknown.
+	ErrUnknownId = errors.New("unknown ID")
+)
+
 // IDTranslator provides an interface for converting from Flow ID's to LibP2P peer ID's
 // and vice versa.
 type IDTranslator interface {
-	// GetPeerID returns the peer ID for the given Flow ID
+	// GetPeerID returns the peer ID for the given Flow ID.
```
yhassanzadeh13, line 22:
```suggestion
	// GetPeerID returns the peer ID for the given Flow identifier.
```

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -0,0 +1,169 @@
...
+		peerID := (peer.ID)("some_peer_ID")
+		s.provider.On("ByPeerID", peerID).Return(identity, true)
+
+		i, found := s.wrapper.ByPeerID(peerID)
+		require.True(s.T(), found)
+		require.Equal(s.T(), i, identity)
+	})
+	s.Run("Identities", func() {
+		identities := unittest.IdentityListFixture(11)
+		filter := filter.In(identities[3:4])
```
yhassanzadeh13, line 65:
```suggestion
		f := filter.In(identities[3:4])
```
`filter` collides with the package name.

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -0,0 +1,169 @@
+package cache_test
```
yhassanzadeh13, line 1: General comment: I think by the term block list in most places, you mean a _black_ list, right?

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -0,0 +1,169 @@
...
+//     While returning (non-nil identity, false) is not a defined return value,
+//     we expect the wrapper to nevertheless handle this case to increase its
+//     generality.
+func (s *NodeBlocklistWrapperTestSuite) TestBlacklistedNode() {
+	blocklist := unittest.IdentityListFixture(11)
+	err := s.wrapper.Update(blocklist.NodeIDs())
+	require.NoError(s.T(), err)
+
+	index := atomic.NewInt32(0)
+	for _, b := range []bool{true, false} {
```
yhassanzadeh13, line 96: Based on the documentation of the `IdentityProvider` interface, the boolean return value of the `ByPeerID` and `ByNodeID` is _"true if and only if Identity has been found, i.e. `Identity` is not nil."_. Accordingly, there will be no happy path case where we have the combination of a non-nil identity and `false` as the returned value. Along the same line, it seems we don't need to iterate over `[]bool{true, false}` rather, all _boolean_ returned values for these functions in this test must be set to `true` (and `false` for the subsequent `TestUnknownNode`).

yhassanzadeh13 (approved):
LGTM, just added a few cosmetic comments.

network/p2p/cache/node_blocklist_wrapper_test.go
AlexHentschel: I used the term "blacklist" before but @peterargue requested a terminology change 
> can you switch the terminology from `blacklist` to `blocklist`

My assumption is that "blocklist" is the common terminology for this particular scenario. Maybe @peterargue can add more insights.

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Iterate over all returned identities and set ejected flag to true. We
+	// copy both the return slice and identities of blocked nodes to avoid
+	// any possibility of accidentally modifying the wrapped IdentityProvider
+	idtx := make(flow.IdentityList, 0, len(identities))
+	w.m.RLock()
+	for _, identity := range identities {
+		if w.blocklist.Includes(identity.NodeID) {
+			var i flow.Identity = *identity // shallow copy is sufficient, because `Ejected` flag is in top-level struct
+			i.Ejected = true
```
zhangchiqing, line 115: Just to confirm, this is NOT an in-place mutation, right? So next time, when `Identities` is called, this Eject field is still false, before setting to true again, right?

Are there test cases that verify:
- if a Node is not originally Ejected, after it's blacklisted and then un-blacklisted, then `Ejected` should be false.
- if a Node is not originally Ejected (i.e. due to slashing), after it's blacklisted and then un-blacklisted, then `Ejected` should still be true.

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -0,0 +1,191 @@
...
+func (w *NodeBlocklistWrapper) Update(blocklist flow.IdentifierList) error {
+	b := blocklist.Lookup() // convert slice to
+	err := persistBlocklist(b, w.db)
+	if err != nil {
+		return fmt.Errorf("failed to persist set of blocked nodes to the data base: %w", err)
+	}
+
+	w.m.Lock()
+	w.blocklist = b
+	w.m.Unlock()
```
zhangchiqing, line 75: Why not lock before updating database? 

I'm not sure if this is concurrent safe. Consider the following events:

1. process A calls `Update(N1, N2)`
2. process B concurrently calls `Update(N3, N4)`
3. process A persist database with [N1, N2]
4. process B persist database with [N3, N4]
5. process B updates w.blocklist with [N3, N4]
6. process A updates w.blocklist with [N1, N2]

Now database has `[N3, N4]`, but w.blocklist has `[N1, N2]`, they are inconsistent.

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// ByPeerID returns the full identity for the node with the given peer ID,
+// where ID is the way the libP2P refers to the node. The function
+// has the same semantics as a map lookup, where the boolean return value is
+// true if and only if Identity has been found, i.e. `Identity` is not nil.
+// Caution: function returns include ejected nodes. Please check the `Ejected`
+// flag in the identity.
+func (w *NodeBlocklistWrapper) ByPeerID(p peer.ID) (*flow.Identity, bool) {
+	identity, b := w.identityProvider.ByPeerID(p)
```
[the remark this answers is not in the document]
AlexHentschel: Both methods [`ByNodeID`](https://github.com/onflow/flow-go/blob/f1bdd42a29a882a0674d36cbd286411c6f0f01c0/network/p2p/cache/node_blocklist_wrapper.go#L131-L134) as well as [`ByPeerID`](https://github.com/onflow/flow-go/blob/f1bdd42a29a882a0674d36cbd286411c6f0f01c0/network/p2p/cache/node_blocklist_wrapper.go#L166-L169) must implement this check. 
In my implementation, both methods call into `applyBlacklist`. If I was to move this check out, I would need to unnecessarily duplicate these checks in [`ByNodeID`](https://github.com/onflow/flow-go/blob/f1bdd42a29a882a0674d36cbd286411c6f0f01c0/network/p2p/cache/node_blocklist_wrapper.go#L131-L134) and [`ByPeerID`](https://github.com/onflow/flow-go/blob/f1bdd42a29a882a0674d36cbd286411c6f0f01c0/network/p2p/cache/node_blocklist_wrapper.go#L166-L169). My preference is to avoid unnecessary code duplication. 

Another way to read your comment could maybe be that the method name `applyBlacklist` is unclear. I have renamed the method to `setEjectedIfBlocked`.

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -0,0 +1,191 @@
...
+// IdentifierSet represents a set of node IDs (operator-defined) whose communication should be blocked.
+type IdentifierSet map[flow.Identifier]struct{}
+
+// Contains returns true iff id ∈ s
+func (s IdentifierSet) Contains(id flow.Identifier) bool {
+	_, found := s[id]
+	return found
+}
+
+// NodeBlocklistWrapper is a wrapper for an `id.IdentityProvider` instance, where the
```
zhangchiqing, line 26: What about calling it `NodeBlacklist` or `NodeBlockedList`? I found Blocklist is a bit confusing

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -0,0 +1,191 @@
...
+	identifiers := make(flow.IdentifierList, 0, len(w.blocklist))
+	for i := range w.blocklist {
+		identifiers = append(identifiers, i)
+	}
+	return identifiers
+}
+
+// Identities returns the full identities of _all_ nodes currently known to the
+// protocol that pass the provided filter. Caution, this includes ejected nodes.
+// Please check the `Ejected` flag in the returned identities (or provide a
```
zhangchiqing, line 99: Should we make it explicit by calling this method `IdentitiesWithEjected` and add another method `IdentitiesWithoutEjected`?

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: sorry. For me the term "blocklist" is also unfamiliar. Please see [my comment here](https://github.com/onflow/flow-go/pull/3344#discussion_r990353927) regarding the naming. Hope Peter will help us understand the terminology.

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: The sole purpose of the `NodeBlocklistWrapper` is to implement the [`IdentityProvider` interface](https://github.com/onflow/flow-go/blob/7baa3f8b09ff8161349fba2f32d7fe61c6944d94/module/id/id_provider.go#L18-L40), which only has the method https://github.com/onflow/flow-go/blob/7baa3f8b09ff8161349fba2f32d7fe61c6944d94/module/id/id_provider.go#L23. 

Happy to add an additional method (like `UnmodifiedIdentities`) which would be outside of the [`IdentityProvider` interface](https://github.com/onflow/flow-go/blob/7baa3f8b09ff8161349fba2f32d7fe61c6944d94/module/id/id_provider.go#L18-L40), but not sure if there is use for it. We can always add it later. Do you have any use case in mind in relation to the admin command or so?

zhangchiqing: I see. Looks good. Let's keep as is then

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: Good point. I didn't pay too much attention to this aspect because we are intending to update the wrapper state manually (through an admin command). 
But you are right, better safe than inconsistent. Since calls to `Update` are going to be _very_ infrequent there are no practical performance concerns with doing the data base write within the lock.

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: revision: https://github.com/onflow/flow-go/pull/3344/commits/bd14684c0656b18aa4e22365d200bb94a7229fb7

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: revision: https://github.com/onflow/flow-go/pull/3344/commits/bd14684c0656b18aa4e22365d200bb94a7229fb7

network/p2p/cache/node_blocklist_wrapper_test.go
peterargue: I suggested it as a less controversial term given this is global open source project. There's a general trend away from `blacklist` and `whitelist` towards `blocklist`/`denylist` and `allowlist`.

network/p2p/cache/node_blocklist_wrapper.go
AlexHentschel: >  this is NOT an in-place mutation, right?

Exactly. Its a little buried in the [`NodeBlocklistWrapper` documentation](https://github.com/onflow/flow-go/blob/bd14684c0656b18aa4e22365d200bb94a7229fb7/network/p2p/cache/node_blocklist_wrapper.go#L26-L35): https://github.com/onflow/flow-go/blob/bd14684c0656b18aa4e22365d200bb94a7229fb7/network/p2p/cache/node_blocklist_wrapper.go#L28-L30

I have some checks (see [here](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L112-L113), [here](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L128-L129), and [here](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L151-L155)) in the existing test [`TestBlacklistedNode`](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L83-L84) that confirm that the wrapper does not write back into the memory returned by the underlying `IdentityProvider`. 

I have refactored those a bit to make it more apparent that we check this in the tests

AlexHentschel: > * if a Node is not originally Ejected, after it's blacklisted and then un-blacklisted, then Ejected should be false.
> * if a Node is originally Ejected (i.e. due to slashing), after it's blacklisted and then un-blacklisted, then Ejected should still be true.

Also very good point. Added the test [`TestBlocklistAddRemove`](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L184-L236) that checks your scenarios.

AlexHentschel: also added the tests
* [`TestUpdate`](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L243) for testing updating, clearing and retrieving the blocklist
* and [`TestDataBasePersist`](https://github.com/onflow/flow-go/blob/32c9d741db63f80c09f2dded083380b9e72f3fc3/network/p2p/cache/node_blocklist_wrapper_test.go#L266) which verifies that blocklist is persisted in database

network/p2p/cache/node_blocklist_wrapper_test.go
AlexHentschel: Completely agree that from the perspective of `IdentityProvider` the return value `(<non-nil identity>, false)` is not specified. 
However, I beliefe there i

network/p2p/cache/node_blocklist_wrapper.go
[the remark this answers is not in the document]
AlexHentschel: Please also see [my comment below](https://github.com/onflow/flow-go/pull/3344#discussion_r990570878), why I prefer to only look at the returned identity, irrespective of the boolean value (defensive programming)

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -0,0 +1,294 @@
...
+		})
+	}
+}
+
+// TestBlocklistAddRemove checks that adding and subsequently removing a node from the blocklist
+// it in combination a no-op. We test two scenarious
+//   - Node whose original `Identity` has `Ejected = false`:
+//     After adding the node to the blocklist and then removing it again, the `Ejected` should be false.
+//   - Node whose original `Identity` has `Ejected = true`:
+//     After adding the node to the blocklist and then removing it again, the `Ejected` should be still be false.
```
zhangchiqing, line 189:
```suggestion
//     After adding the node to the blocklist and then removing it again, the `Ejected` should be still be true.
```

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -0,0 +1,294 @@
...
+// TestDataBasePersist verifies that blocklist is persisted in database
+func (s *NodeBlocklistWrapperTestSuite) TestDataBasePersist() {
+	blocklist := unittest.IdentifierListFixture(8)
+
+	// update blocklist and check DB that the new value is there
+	err := s.wrapper.Update(blocklist)
+	require.NoError(s.T(), err)
+
+	var b1 map[flow.Identifier]struct{}
+	err = s.DB.View(operation.RetrieveBlocklist(&b1))
```
zhangchiqing, line 274: This would imply that we know the implementation detail of database. And therefore coupled by the database design, if we change design, this test would have to change as well. 

The database implementation is better to be tested by their own unit test. 

So here, a better strategy is to create a new block list wrapper that wraps the database. Since a new wrapper is created, it should have forgotten what is possibly being cached in memory and if it returns the correct data, then it must be from the database.

zhangchiqing (approved):
Looks good. Just a minor suggestion to the test case.

network/p2p/cache/node_blocklist_wrapper_test.go
AlexHentschel: Thanks for the great suggestion on how to decouple testing of `NodeBlocklistWrapper` from the lower-level data base design!
* Added separate [unity tests](https://github.com/onflow/flow-go/blob/426c830174ecdb34b277fea56d9112a67cca4e9c/storage/badger/operation/bft_test.go#L18) for the lower-level data-base abstraction
* Refactored the unit tests for `NodeBlocklistWrapper` to implement you suggestion
   > create a new block list wrapper that wraps the database. Since a new wrapper is created, it should have forgotten what is possibly being cached in memory and if it returns the correct data, then it must be from the database.

   updated test: [`TestDataBasePersist`](https://github.com/onflow/flow-go/blob/4b252e1e4377db871dc6b3e312e73fa9c5e92402/network/p2p/cache/node_blocklist_wrapper_test.go#L265-L276)

AlexHentschel:
bors merge

turbolent:
@AlexHentschel please avoid using oppressive terminology and opt for neutral terms instead. See https://github.com/mbmccormick/oppressivelanguage#blacklist--whitelist

AlexHentschel:
thanks @turbolent for the link, glad I changed blacklist to **block**list 😅

PR #3345, opened 2022-10-05
Ensure cadence don't access unexpected operations while parsing (part 2)

pattyshack:
Added utility generic functions for checking parse restriction.

Sorry for the poor `parseRestrict<x>Arg<y>Ret` naming, but I can't think of anything better.

The generic functions are of the form
    `parseRestrict<x>Arg<y>Ret(txnState, opCode, callback, arg1, ..., argX)`
where the callback expects `<x>` number of arguments, and `<y>` number of return values (not counting error). If the callback expects no argument, `<x>Arg` is omitted, and similarly for return value.

### DIFF
--- a/fvm/environment/parse_restricted_checker.go
+++ b/fvm/environment/parse_restricted_checker.go
@@ -1,0 +1,135 @@
+package environment
+
+import (
+	"github.com/onflow/flow-go/fvm/errors"
+	"github.com/onflow/flow-go/fvm/state"
+)
+
+func parseRestricted(txnState *state.TransactionState, opCode string) error {
+	if txnState.IsParseRestricted() {
+		return errors.NewParseRestrictedModeInvalidAccessFailure(opCode)
+	}
+
+	return nil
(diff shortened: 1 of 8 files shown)


fvm/environment/parse_restricted_checker.go
```diff
@@ -0,0 +1,37 @@
...
+func (checker parseRestrictedChecker) parseRestricted(op string) error {
+	if checker.txnState.IsParseRestricted() {
+		return errors.NewParseRestrictedModeInvalidAccessFailure(op)
+	}
+
+	return nil
+}
+
+func (checker parseRestrictedChecker) getAddressValue(
+	op string,
```
Tonix517, line 25: nit: maybe in future we can have opCode instead of string. but no strong preference here.

Tonix517 (approved):

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #3346, opened 2022-10-05
Add searchable labels to bft related log events

peterargue:


kc1116 (approved):

jordanschalm (approved):
Nice, good idea. 

Only suggestion would be to make this a constant, like `logging.KeySuspicious`, then we can provide some documentation about when it's appropriate to attach this key alongside the constant declaration, avoid typos.

peterargue:
Good call. added in 4a12c21

peterargue:
bors merge

PR #3347, opened 2022-10-05
[Checkpoint V6] refactor getNodeByIndex

zhangchiqing:


ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -197,9 +197,9 @@ func readCheckpointSubTrie(dir string, fileName string, index int, checksum uint
 
 	// TODO: validate checksum again
 
-	// TODO: simplify getNodeByIndex() logic if we reslice nodes here to remove the nil node at index 0.
-	// return nodes[1:], nil
-	return nodes, nil
+	// since nodes[0] is always `nil`, returning a slice without nodes[0] could simplify the
+	// implementation of getNodeByIndex
+	return nodes[1:], nil
```
zhangchiqing, line 202: Addressing comment from @fxamacker 

>We can simplify getNodeByIndex() logic if we reslice nodes here to remove the nil node at index 0.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -326,37 +326,43 @@ func readNodeCount(f *os.File) (uint64, uint16, error) {
...
-			return nil, fmt.Errorf("subtrie[0] %v isn't nil", subtries[0])
+		if int(offset) < len(subtries) {
+			return subtries[offset], nil
 		}
 
-		if offset < uint64(len(subtries)-1) {
-			// +1 because first item is always nil
-			return subtries[offset+1], nil
-		}
-		offset -= uint64(len(subtries) - 1)
```
zhangchiqing, line 356: Addressing comment from @fxamacker 

>This block of code can be simplified if subtries doesn't contain nil node as the first element. See comment in readCheckpointSubTrie to reslice returned nodes.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -365,10 +371,12 @@ func getNodeByIndex(subtrieNodes [][]*node.Node, topLevelNodes []*node.Node, ind
...
-		return nil, fmt.Errorf("can not find node by index: %v in subtrieNodes %v, topLevelNodes %v", index, subtrieNodes, topLevelNodes)
+	nodePos := index - totalSubTrieNodeCount
+
+	if nodePos >= uint64(len(topLevelNodes)) {
+		return nil, fmt.Errorf("can not find node by index %v, nodePos >= len(topLevelNodes) => (%v > %v)",
+			index, nodePos, len(topLevelNodes))
 	}
 
-	// +1 because first item is always nil
-	return topLevelNodes[offset+1], nil
```
zhangchiqing, line 373: Addressing comment from @fxamacker , I extracted a `getTopNodeByIndex` function

>Maybe this part of the function can be moved to the caller.

>This function can focus on getSubTrieNodeByIndex and the caller can get node like this:

>if node index == 0, node is nil
if node index <= totalSubTrieNodeCount, call this function to look up node in the subtrie groups
else, look up node in topLevelNodes after adjusting index

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -326,37 +326,43 @@ func readNodeCount(f *os.File) (uint64, uint16, error) {
...
-		}
-		offset -= uint64(len(subtries) - 1)
+		offset -= uint64(len(subtries))
 	}
 
-	// TODO: move this check outside
+	return nil, fmt.Errorf("could not find node by index %v, totalSubTrieNodeCount %v", index, totalSubTrieNodeCount)
+}
+
+func getTopNodeByIndex(totalSubTrieNodeCount uint64, topLevelNodes []*node.Node, index uint64) (*node.Node, error) {
```
fxamacker, line 364: `getTopNodeByIndex` can be removed because we can lookup top level nodes directly in `getNodeByIndex`.

I think this function isn't necessary because:
- checking if `topLevelNodes[0]` is nil isn't necessary because this function is internal and only used by `readTopLevelTries`.  In `readTopLevelTries`, `topLevelNodes[0]` is nil and documented.
-  `if nodePos >= uint64(len(topLevelNodes))` isn't necessary because caller [filters out invalid node index](https://github.com/onflow/flow-go/blob/821bffa3d75c87bdfe954abd888f451c9824a6d1/ledger/complete/wal/checkpoint_v6_reader.go#L281-L283)  
- remaining functionality of looking up node in `topLevelNodes` with adjusted index can be moved to the caller.  See comment in `getNodeByIndex` for modified code.

fxamacker (commented):
Thanks for the refactor!  I added a few comments to simplify it further.

ledger/complete/wal/checkpoint_v6_reader.go
zhangchiqing: I do want to keep these checks, a checkpoint file might get manipulated in a certain way to bypass the early checks. 

For instance, a malicious actor might change the top level node to include an index for non-existing node, they could still produce a correct checksum, node count, etc, but then it could cause an out of index panic.

fxamacker: @zhangchiqing I agree with you that we need to check for malicious data.  That is why we use multiple checks when processing checkpoint file.

However, the checks here don't detect malicious data because:
- a check is done to internal data structure not affected by potentially malicious data, or
- more strict check is already done by the caller.

In the first case, `topLevelNodes` is a slice we create and guarantee that the first element is nil.  No malicious data would change that guarantee.  

Your concern applies to the second case.  But a stricter check is already done in the caller.  https://github.com/onflow/flow-go/blob/821bffa3d75c87bdfe954abd888f451c9824a6d1/ledger/complete/wal/checkpoint_v6_reader.go#L281-L283

Given this, I think these specific checks here won't actually detect any malicious data in practice.  

Thoughts?

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -365,10 +371,12 @@ func getNodeByIndex(subtrieNodes [][]*node.Node, topLevelNodes []*node.Node, ind
 		return nil, fmt.Errorf("top trie [0] %v isn't nil", topLevelNodes[0])
 	}
 
-	if offset >= uint64(len(topLevelNodes)-1) {
-		return nil, fmt.Errorf("can not find node by index: %v in subtrieNodes %v, topLevelNodes %v", index, subtrieNodes, topLevelNodes)
+	nodePos := index - totalSubTrieNodeCount
+
+	if nodePos >= uint64(len(topLevelNodes)) {
```
zhangchiqing, line 376: Yeah, that makes sense. I can remove the sanity check above.

But I think I should still keep the check here, because it's essentially to replace [the existing check here](https://github.com/onflow/flow-go/blob/master/ledger/complete/wal/checkpointer.go#L906-L908)

PR #3349, opened 2022-10-06
Update rate limited blobstore to drop requests

smnzhu:
Drop requests if rate limit doesn't allow, instead of waiting for timeout


peterargue (commented):
these changes look good. are there any tests that cover this? if not, can you add some?

peterargue (approved):
There are a few linter errors, but otherwise looks good.

koko1123 (approved):

smnzhu:
bors merge

PR #3350, opened 2022-10-06
[Checkpoint V6] Encode subtrie node count

zhangchiqing:


Tonix517:
looks like in foreseeable future, to maintain our in-house checkpoint file format, extra dev\testing work will be required - like this PR itself, we have to write\test\maintain code to simply add a new field in the file. 

How about using some existing solutions? to me https://google.github.io/flatbuffers/ is a good candidate. It is fast and efficient. Also changes like this PR can be as simple as one-liner change to the scheme file. Some food for thought.

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -258,16 +257,31 @@ func readTopLevelTries(dir string, fileName string, subtrieNodes [][]*node.Node,
...
 
 	reader := NewCRC32Reader(bufio.NewReaderSize(file, defaultBufioReadSize))
 
+	// read subtrie count
+	buf := make([]byte, encNodeCountSize)
+	_, err = io.ReadFull(reader, buf)
+	if err != nil {
+		return nil, fmt.Errorf("could not read subtrie node count: %w", err)
+	}
+	readSubtrieNodeCount, err := decodeSubtrieFooter(buf)
```
fxamacker, line 273: Calling `decodeSubtrieFooter()` here can be confusing because it isn't a footer.  Maybe rename the function or directly decode uint64 here.
```suggestion
	readSubtrieNodeCount := binary.BigEndian.Uint64(buf)
```

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -258,16 +257,31 @@ func readTopLevelTries(dir string, fileName string, subtrieNodes [][]*node.Node,
...
+	if err != nil {
+		return nil, fmt.Errorf("could not decode node count: %w", err)
+	}
```
fxamacker, line 276: This error checking isn't needed if we decode subtrie node count directly above.

fxamacker (approved):
Thanks for adding subtrie node count.

PR #3354, opened 2022-10-06
[FVM] Lazy seed unsafe random

janezpodhostnik:
Just noticed that seeding the unsafe random is not negligible. Its 90% of creating a new environment (but still very tiny ~1.5% of transaction execution).

Since very few transactions/scripts actually use the unsafe random, It might be better to just lazily seed it.


fvm/environment/unsafe_random_generator.go
```diff
@@ -11,34 +12,49 @@ import (
...
-	return gen
+	})
 }
 
 // UnsafeRandom returns a random uint64, where the process of random number
 // derivation is not cryptographically secure.
 func (gen *UnsafeRandomGenerator) UnsafeRandom() (uint64, error) {
 	defer gen.tracer.StartExtensiveTracingSpanFromRoot(trace.FVMEnvUnsafeRandom).End()
 
+	gen.seed()
```
zhangchiqing, line 56: It seems we want `seed` to ensure that `gen.rng` is ready after `seed` is called. 

Does sync.Once guarantee that?

I think sync.Once only guarantee if two calls happen concurrently, only one of them will run certain code. Is it possible the following event happens:

1. process A calls UnsafeRandom, triggers gen.seed to generate seed
2. process B concurrently calls UnsafeRandom, since process A is already generating seed, it skip immediately, and continue, then it hits `gen.rng == nil` and got a OperationNotSupportedError

I think we might need to use a mutex lock to block the call and ensure rng has been set before unblocking the call.

Please add a test to verify concurrent safe here.

SaveTheRbtz: https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/sync/once.go;l=55-61

> Do guarantees that when it returns, f has finished.

janezpodhostnik: > process B concurrently calls UnsafeRandom

unsafe random cannot be concurrently called, as `gen.rng.Read(buf)` is already not concurrency safe. Not sure if concurrency is needed her, since this is within one transaction.Another argument against concurrency is that random needs to be deterministic. if process A and B call unsafe random at the same time, the random number A and B get should not depend on which is first. I think if we wanted concurrency here, each process would have their own instance of a deterministically pre-seeded random.

pattyshack (approved):

zhangchiqing (approved):
Looks good. 

Better adding comments for the blocking behavior of the `Do` method, and the concurrency safetiness of the UnsafeRandom method, in case these assumptions is changed in the future.

janezpodhostnik:
bors merge

janezpodhostnik:
bors merge

PR #3356, opened 2022-10-07
Refactor StoreCheckpoint inputs

zhangchiqing:
In order to prepare for using checkpoint V6, this PR changes the StoreCheckpoint method's parameter to take dir and fileName.


ledger/complete/wal/checkpointer_test.go
```diff
@@ -443,57 +434,57 @@ func Test_Checkpointing(t *testing.T) {
...
+//
+// 		checkpointer, err := wal2.NewCheckpointer()
+// 		require.NoError(t, err)
+//
+// 		t.Run("write error", func(t *testing.T) {
+// 			errWrite := errors.New("unexpected write error")
+//
+// 			err = checkpointer.Checkpoint(1, func() (io.WriteCloser, error) {
+// 				return newWriteCloserWithErrors(errWrite, nil), nil
+// 			})
```
zhangchiqing, line 474: writer is now created internally, not sure how to re-create this case here, any idea?

ledger/complete/wal/checkpointer.go
```diff
@@ -293,7 +286,18 @@ func CreateCheckpointWriterForFile(dir, filename string, logger *zerolog.Logger)
 // as for each node, the children have been previously encountered.
 // TODO: evaluate alternatives to CRC32 since checkpoint file is many GB in size.
 // TODO: add concurrency if the performance gains are enough to offset complexity.
-func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
+func StoreCheckpoint(dir string, fileName string, logger *zerolog.Logger, tries ...*trie.MTrie) error {
```
zhangchiqing, line 289: checkpoint V6 will start using dir and fileName, changing it to make the transition easier

ledger/complete/wal/checkpointer.go
```diff
@@ -293,7 +286,18 @@ func CreateCheckpointWriterForFile(dir, filename string, logger *zerolog.Logger)
...
+	writer, err := CreateCheckpointWriterForFile(dir, fileName, logger)
+	if err != nil {
+		return fmt.Errorf("could not create writer: %w", err)
+	}
+	defer func() {
+		closeError := writer.Close()
+		// Return close error if there isn't any prior error to return.
```
Tonix517, line 296: what if `err` is not nil, do we need to use `multierror` to return both errors?

ledger/complete/wal/checkpointer_test.go
pattyshack: if you convert StoreCheckpoint into a method of checkpointer, you can store an extra field in checkpointer which normally points to CreateCheckpointWriterForFile, but can be override during test.

Tonix517 (approved):

ledger/complete/wal/checkpointer.go
```diff
@@ -293,7 +287,22 @@ func CreateCheckpointWriterForFile(dir, filename string, logger *zerolog.Logger)
 // as for each node, the children have been previously encountered.
 // TODO: evaluate alternatives to CRC32 since checkpoint file is many GB in size.
 // TODO: add concurrency if the performance gains are enough to offset complexity.
-func StoreCheckpoint(writer io.Writer, tries ...*trie.MTrie) error {
+func StoreCheckpoint(dir string, fileName string, logger *zerolog.Logger, tries ...*trie.MTrie) error {
```
fxamacker, line 290: We need to use named return argument in order to capture and set returned error in `defer`.

```suggestion
func StoreCheckpoint(dir string, fileName string, logger *zerolog.Logger, tries ...*trie.MTrie) (err error) {
```

fxamacker (changes requested):

ledger/complete/wal/checkpointer_test.go
zhangchiqing: Yeah, but I feel that's a trade off. 

If StoreCheckpoint is a function, then the behavior of storing a checkpoint is only determined by the StoreCheckpoint's inputs.
But if StoreCheckpoint is a method, then the behavior is not only determined by the inputs but also could possibly be altered by the object's state. If there is bug, the surface would be extended a lot, we have to check whether StoreCheckpoint internally depend on any state that could change, which makes the reasoning harder.

network/p2p/test/sporking_test.go
```diff
@@ -37,6 +37,7 @@ import (
 // TestCrosstalkPreventionOnNetworkKeyChange tests that a node from the old chain cannot talk to a node in the new chain
 // if it's network key is updated while the libp2p protocol ID remains the same
 func TestCrosstalkPreventionOnNetworkKeyChange(t *testing.T) {
+	unittest.SkipUnless(t, unittest.TEST_FLAKY, "flakey test")
```
zhangchiqing, line 40: cc @peterargue

zhangchiqing:
Close as it's been merged to the feature branch
https://github.com/onflow/flow-go/pull/3273

PR #3359, opened 2022-10-07
Ensure cadence don't access unexpected operations while parsing (part 3)

pattyshack:


fvm/environment/parse_restricted_checker.go
```diff
@@ -133,3 +167,54 @@ func parseRestrict4Arg1Ret[
...
+) {
+	err := parseRestricted(txnState, opCode)
+	if err != nil {
+		var value1 Ret1T
+		var value2 Ret2T
+		return value1, value2, err
+	}
+
+	return callback(arg)
+}
```
Tonix517, line 220: iirc golang does not have plan to support generic variadic functions. i'm thinking combining https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source and "text/template" may save us some typing here. but just for entertainment purposes, it is too hacky for production.

Tonix517 (approved):

fvm/environment/parse_restricted_checker.go
pattyshack: not supporting adhoc (i.e., overloading / variadic) polymorphism is the correct thing to do from the language designer's perspective.  unlike parametric polymorphism, there's no solid theoretical foundation for adhoc polymorphism.

Tonix517: this is why i hate C++ ;)

SaveTheRbtz (approved):

pattyshack:
bors merge

janezpodhostnik (approved):
Hmmm... It might be time to unify `opCode`, `SpanName` and `common.ComputationKind` and use a operation wrapper like the `parseRestricted` to not only check for `IsParseRestricted` but also do:
```
defer info.tracer.StartSpanFromRoot(opCode).End()
err := info.meter.MeterComputation(opCode, intensity())
```

pattyshack:
yeah, that's reasonable.

Tonix517:
+1
> Hmmm... It might be time to unify `opCode`, `SpanName` and `common.ComputationKind` and use a operation wrapper like the `parseRestricted` to not only check for `IsParseRestricted` but also do:
> 
> ```
> defer info.tracer.StartSpanFromRoot(opCode).End()
> err := info.meter.MeterComputation(opCode, intensity())
> ```

PR #3367, opened 2022-10-11
Fix TransactionPrograms Commit error handling.

pattyshack:
It is not safe to ignore txnPrograms' commit error for transactions that trigger programs invalidation.

Note: commit never returns error under sequential execution (except with bad test setup).

### DIFF
--- a/fvm/fvm.go
+++ b/fvm/fvm.go
@@ -93,22 +93,6 @@ func (vm *VirtualMachine) Run(
 		return fmt.Errorf("error creating transaction programs: %w", err)
 	}
 
-	// Note: it is safe to skip committing the parsed programs for non-normal
-	// transactions (i.e., bootstrap and script) since this is only an
-	// optimization.
-	if proc.Type() == TransactionProcedureType {
-		defer func() {
-			commitErr := txnPrograms.Commit()
-			if commitErr != nil {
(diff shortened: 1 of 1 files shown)


fvm/fvm.go
```diff
@@ -141,7 +125,21 @@ func (vm *VirtualMachine) Run(
...
+		return err
+	}
+
+	// Note: it is safe to skip committing the parsed programs for non-normal
+	// transactions (i.e., bootstrap and script) since these do not invalidate
+	// programs.
+	if proc.Type() == TransactionProcedureType {
+		// NOTE: It is not safe to ignore txnPrograms' commit error for
+		// transactions that trigger programs invalidation.
+		return txnPrograms.Commit()
```
Tonix517, line 139: for understanding purpose, how should we handle the error from `Commit()` here, is the current error handling enough for this new error source?

Tonix517 (approved):

fvm/fvm.go
pattyshack: this will halt the execution node.

janezpodhostnik (approved):

pattyshack:
bors merge

pattyshack:
bors retry

PR #3368, opened 2022-10-11
[Checkpoint v6] Refactor file close error

zhangchiqing:


ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -78,15 +78,18 @@ func ReadCheckpointV6(headerFile *os.File, logger *zerolog.Logger) ([]*trie.MTri
 }
 
 // OpenAndReadCheckpointV6 open the checkpoint file and read it with ReadCheckpointV6
-func OpenAndReadCheckpointV6(dir string, fileName string, logger *zerolog.Logger) ([]*trie.MTrie, error) {
+func OpenAndReadCheckpointV6(dir string, fileName string, logger *zerolog.Logger) (
+	tries []*trie.MTrie,
+	errToReturn error,
```
zhangchiqing, line 83: if you are thinking why not just `err`, then please read the comments on `closeAndMergeError`

ledger/complete/wal/checkpoint_v6_reader.go
```diff
@@ -126,7 +129,11 @@ func filePaths(dir string, fileName string, subtrieLevel uint16) []string {
 
 // readCheckpointHeader takes a file path and returns subtrieChecksums and topTrieChecksum
 // any error returned are exceptions
-func readCheckpointHeader(filepath string, logger *zerolog.Logger) ([]uint32, uint32, error) {
+func readCheckpointHeader(filepath string, logger *zerolog.Logger) (
+	checksumsOfSubtries []uint32,
```
zhangchiqing, line 133: I intended to have this verbose name, it is to ensure it's not being used and only for definition.

PR #3369, opened 2022-10-12
Admin command for dynamic config updates

jordanschalm:


integration/localnet/bootstrap.go
```diff
@@ -503,6 +503,7 @@ func defaultService(role, dataDir, profilerDir string, i int) Service {
 			fmt.Sprintf("--tracer-enabled=%t", tracing),
 			"--profiler-dir=/profiler",
 			"--profiler-interval=2m",
+			fmt.Sprintf("--admin-addr=0.0.0.0:%d", AdminToolPort),
```
zhangchiqing, line 506: 👍

module/updatable_configs/manager.go
```diff
@@ -0,0 +1,217 @@
...
+// If you need to add a new configurable config field with a type not below:
+//  1. Add a new setter and getter type below
+//  2. Add a Register*Config method to the Registerer interface and Manager implementation below
+//  3. Add a TestManager_Register*Config test to the manager_test.go file.
+
+type (
+	// Set*ConfigFunc is a setter function for a single updatable config field.
+	// Returns ValidationError if the new config value is invalid.
+
+	SetUintConfigFunc     func(uint) error
```
zhangchiqing, line 51:
```suggestion
	SetUintConfigFunc     func(uint) (uint, error)
```
It's better to return a uint as old value, would be useful for logging. So that we log when it's changed from a certain old value to a new value. 

Does generic support type restriction? Meaning restricting both type T here to be the same type.
```
	SetAnyConfigFunc func(T) (T, error)
```

zhangchiqing (commented):
Thanks for the refactor. I had a few comments and questions

consensus/hotstuff/pacemaker/timeout/config.go
```diff
@@ -104,11 +108,41 @@ func NewConfig(
...
-		BlockRateDelayMS:               float64(blockRateDelay.Milliseconds()),
+		BlockRateDelayMS:               atomic.NewFloat64(float64(blockRateDelay.Milliseconds())),
 	}
 	return tc, nil
 }
 
+// validBlockRateDelay validates a block rate delay config.
+// Returns model.ConfigurationError for invalid config inputs.
+func validBlockRateDelay(blockRateDelay time.Duration) error {
+	if blockRateDelay < 0 {
```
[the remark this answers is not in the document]
jordanschalm: We've actually intentionally set this value greater than the default timeout in the past (ie. the node configured this way would always send their block proposal _after_ everyone else has timed out). It behaves similarly to setting it below the default timeout, in that it lowers the average block rate, just more. So I'm not sure where the "sanity check" cutoff should really be here. 

Another perspective: One operator accidentally setting a 10m delay isn't a big deal - it's similar to having one node offline, the impact is very small. Many operators accidentally setting a 10m delay is definitely bad. But it's also quite bad if many operators accidentally set a delay in the range that we desire to be available for block rate adjustments.

admin/README.md
```diff
@@ -38,18 +38,31 @@ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"
...
-### To set required approvals for sealing
+#### To get a list of all updatable configs
+```
+curl localhost:3705/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "list-configs"}'
+```
+
+### To get a config value
 ```
-curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-required-approvals-for-sealing", "data": 1}'
+curl localhost:3705/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-config", "data": "consensus-required-approvals-for-sealing"}'
```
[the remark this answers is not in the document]
jordanschalm: Nice catch - I'll update these

module/profiler/profiler.go
```diff
@@ -67,15 +54,28 @@ func New(log zerolog.Logger, uploader Uploader, dir string, interval time.Durati
...
+func (p *AutoProfiler) Enabled() bool {
+	return p.enabled.Load()
+}
+
 func (p *AutoProfiler) Ready() <-chan struct{} {
 	delay := time.Duration(float64(p.interval) * rand.Float64())
 	p.unit.LaunchPeriodically(p.start, p.interval, delay)
 
-	if profilerEnabled.Load() {
+	if p.enabled.Load() {
```
SaveTheRbtz, line 78:
```suggestion
	if !p.Enabled() {
```

consensus/hotstuff/pacemaker/timeout/config.go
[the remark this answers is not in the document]
AlexHentschel: we could also print a warning when the admin command is executed: 
* For example, we could warn if `MinReplicaTimeout` is smaller or equal to the new delay. In this scenario, the node would be too late if the protocol was on the happy path.

zhangchiqing (approved):
Great work. Nice refactor 👍

consensus/config.go
```diff
@@ -74,3 +91,9 @@ func WithBlockRateDelay(delay time.Duration) Option {
 		cfg.BlockRateDelay = delay
 	}
 }
+
+func WithConfigRegisterer(reg updatable_configs.Registerer) Option {
```
kc1116, line 95: nit: rename references using Registerer -> Registrar

kc1116 (approved):
🚀

consensus/hotstuff/pacemaker/timeout/config.go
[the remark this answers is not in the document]
jordanschalm: 👍 Added a warn log for this case https://github.com/onflow/flow-go/pull/3369/commits/54b0bf4b708211e67907e553673c0bd60326f789

consensus/config.go
jordanschalm: https://github.com/onflow/flow-go/pull/3369/commits/ea13d09c415ae103b425215221efaffdc3434acb

jordanschalm:
bors merge

PR #3371, opened 2022-10-12
[FVM] Keeping event counter inside EventEmitter

Tonix517:
In PR #3268 the event counter was also moved into meter interface. It caused a problem in event index ([slack thread](https://dapperlabs.slack.com/archives/C01ANSMH5FF/p1665462723443849)) because meter itself has `enforceLimit` condition which event index should not be applied. Event index should increase unconditionally.
Fixed by keeping the event counter inside EventEmitter as usual.

Test
- [X] existing and patched UTs

### DIFF
--- a/fvm/fvm_fuzz_test.go
+++ b/fvm/fvm_fuzz_test.go
@@ -198,6 +198,12 @@ var fuzzTransactionTypes = []transactionType{
 			fees, deducted := getDeductedFees(t, tctx, results)
 			require.True(t, deducted, "Fees should be deducted.")
 			require.GreaterOrEqual(t, fees.ToGoValue().(uint64), fuzzTestsInclusionFees)
+			// event indices have to be an increasing uint sequence (0, 1, 2 ...)
+			expectedEventIndex := uint32(0)
+			for _, event := range results.tx.Events {
+				require.Equal(t, expectedEventIndex, event.EventIndex)
+				expectedEventIndex++
+			}
 		},
 	},
 }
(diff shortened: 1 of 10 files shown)


pattyshack (approved):
since the event counter is not essential to meter limiting, should we just remove the eventCounter from meter?

Tonix517:
addressed. good idea to keep our code lean.
> since the event counter is not essential to meter limiting, should we just remove the eventCounter from meter?

pattyshack (commented):
lgtm

fvm/fvm_fuzz_test.go
```diff
@@ -198,6 +198,12 @@ var fuzzTransactionTypes = []transactionType{
 			fees, deducted := getDeductedFees(t, tctx, results)
 			require.True(t, deducted, "Fees should be deducted.")
 			require.GreaterOrEqual(t, fees.ToGoValue().(uint64), fuzzTestsInclusionFees)
+			// event indices have to be an increasing uint sequence (0, 1, 2 ...)
+			expectedEventIndex := uint32(0)
+			for _, event := range results.tx.Events {
+				require.Equal(t, expectedEventIndex, event.EventIndex)
+				expectedEventIndex++
+			}
```
janezpodhostnik, line 206: can you package this into a function and add it to all the test cases. (You can do that in a separate PR so this can be merged faster)

janezpodhostnik (approved):
Can you (in a separate PR or in this one) also add two more tests:
One testing that not only the fee deduction events, but also events emitted during transaction invocation are numbered correctly, and one testing that service events also respect the increasing index.

Tonix517:
bors merge

Tonix517:
bors retry

PR #3373, opened 2022-10-12
[BFT] Authorize execution data requests over bitswap

peterargue:
ENs and ANs use bitswap to share ExecutionData within the staked network. To support permissionless nodes, all messages must be authorized to the specific node/roles that are allowed to participate.

This PR adds an authorization check to all ExecutionData requests over bitswap. It also adds configuration to Execution nodes that allows them to specify an allow list of nodes that can request ExecutionData directly. This enable additional options for handling malicious peers, as well as scaling the execution sync protocol.

### DIFF
--- a/cmd/execution_builder.go
+++ b/cmd/execution_builder.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ipfs/go-bitswap"
 	"github.com/ipfs/go-cid"
 	badger "github.com/ipfs/go-ds-badger2"
-	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/onflow/flow-core-contracts/lib/go/templates"
 	"github.com/rs/zerolog"
 	"github.com/shirou/gopsutil/v3/cpu"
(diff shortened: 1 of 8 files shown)


kc1116 (approved):

cmd/execution_builder.go
```diff
@@ -375,7 +374,42 @@ func (exeNode *ExecutionNode) LoadProviderEngine(
...
+	}
+
+	opts := []network.BlobServiceOption{
+		blob.WithBitswapOptions(
+			// Only allow block requests from staked ENs and ANs on the allowedANs list (if set)
+			bitswap.WithPeerBlockRequestFilter(
+				blob.AuthorizedRequester(allowedANs, exeNode.builder.IdentityProvider, exeNode.builder.Logger),
+			),
+		),
+	}
```
smnzhu, line 412: Is it enough to only initialize this list once? What happens if AN's join / leave between epochs?

peterargue: the EN would have to update their list and reboot. 

The intention is for ENs to have the option to choose a group of ANs they trust as the first tier to support, then allow all other ANs to get the data from eachother. This would spread out the load across a larger number of nodes, and provide a tool to allow ENs to block spammy ANs.

smnzhu (approved):

peterargue:
borse merge

peterargue:
bors merge

PR #3380, opened 2022-10-13
[Execution] Adding more metrics to the execution node

ramtinms:
This PR 
- updates the structure of the ComputationResult to enable better capturing of meta data, ComputationResult still needs a major refactoring in my view but that was outside the scope of this PR. 
- adds some metrics on chunk data pack generation (time spent on generation, proofsize, number of transactions) 
- adds an approximate light way of measuring the event sizes
- adds an extra metric to transaction to capture size of event 
- adds extra metrics to collection execution step ( event counts and sizes, number of registers touched, number of bytes written to registers, total computation and memory usage) 
- adds extra metrics to block execution step (aggregated memory usage, aggregated event count and size) 
- removed StateReads metric as it was not that much meaning considering total number of register reads per block is not a very helpful metric.


engine/execution/messages.go
```diff
@@ -17,18 +17,17 @@ type ComputationOrder struct {
...
 type ComputationResult struct {
-	ExecutableBlock    *entity.ExecutableBlock
-	StateSnapshots     []*delta.SpockSnapshot
-	StateCommitments   []flow.StateCommitment
-	Proofs             [][]byte
-	Events             []flow.EventsList
-	EventsHashes       []flow.Identifier
-	ServiceEvents      flow.EventsList
-	TransactionResults []flow.TransactionResult
-	ComputationUsed    uint64
```
ramtinms, line 28: this aggregated value is not computed on demand.

engine/execution/messages.go
```diff
@@ -17,18 +17,17 @@ type ComputationOrder struct {
...
-	StateReads         uint64
```
ramtinms, line 29: this was a meaning less metric, we care about state reads per collection level, so removed this and added other metrics to capture better signals.

model/flow/event.go
```diff
@@ -79,6 +79,16 @@ func wrapEventID(e Event) eventIDWrapper {
...
 
+// byteSize returns the number of bytes needed to store the wrapped version of the event.
+// returned int is an approximate measure, ignoring the number of bytes needed as headers.
+func (e Event) byteSize() int {
+	return IdentifierLen + // txID
+		4 + // Index
+		len(e.Type) + // Type
+		4 + // TransactionIndex
+		len(e.Payload) // Payload
+}
```
ramtinms, line 90: this is an approximate size, I was trying to avoid encoding to capture the size, ideally we could do something smarter based on how we encode and store events.

engine/execution/computation/computer/computer.go
```diff
@@ -481,9 +491,10 @@ func (e *blockComputer) executeTransaction(
 	e.metrics.ExecutionTransactionExecuted(
 		time.Since(startedAt),
 		tx.ComputationUsed,
-		memAllocAfter-memAllocBefore,
 		tx.MemoryEstimate,
+		memAllocAfter-memAllocBefore,
```
pattyshack, line 495: maybe we should just get rid of per txn memory usage estimate base on mem alloc.

it doesn't seem very meaningful since memory usage is dominated by background task (snapshotting, lib p2p, etc.) and semi-permanent allocation (trie / storage).  The number will become even less meaningful when we start parallelizing some aspects of transaction (pre)processing.

ramtinms: @dsainati1 thoughts on this? do we still need it?

janezpodhostnik (commented):
This will have merge conflicts with https://github.com/onflow/flow-go/pull/3403 @Tonix517

model/flow/event.go
ramtinms: we actually do collect the size when emitting the events, maybe we could consolidate and capture the size here as a field.

ramtinms:
@janezpodhostnik I just updated the branch and resolved conflicts

model/flow/event.go
pattyshack: +1 wrt janez's comment

pattyshack (approved):
i still don't think we need to measure mem alloc on a per transaction basis, but I don't feel strongly about it.

engine/execution/computation/computer/computer.go
dsainati1: Sorry I think I missed this comment originally, but Cadence needs some way to sanity check our memory metering estimates to make sure changes we make or features we add (especially in light of the upcoming Stable Cadence changes) don't underestimate memory usage. So if we get rid of this ideally we'd replace it with a better estimate.

Tonix517 (approved):

engine/execution/computation/computer/computer.go
ramtinms: got it lets leave it the way it is in this PR and we could address it later.

ramtinms:
bors merge

PR #3386, opened 2022-10-13
Adding Event Index Sequence checking to UTs

Tonix517:
Adding new checking in UT cases to make sure numbers in event index sequence are unique and increasing. As a follow-up to https://github.com/onflow/flow-go/pull/3371#pullrequestreview-1139592903

changes to FVM and execution computer UT are split into separate commits.

### DIFF
--- a/engine/execution/computation/computer/computer_test.go
+++ b/engine/execution/computation/computer/computer_test.go
@@ -445,6 +445,11 @@ func TestBlockExecutor_ExecuteBlock(t *testing.T) {
 		result, err := exe.ExecuteBlock(context.Background(), block, view, programs.NewEmptyBlockPrograms())
 		require.NoError(t, err)
 
+		// make sure event index sequence are valid
+		for _, eventsList := range result.Events {
+			ensureEventsIndexSeq(t, eventsList, execCtx.Chain.ChainID())
+		}
+
 		// all events should have been collected
 		require.Len(t, result.ServiceEvents, 2)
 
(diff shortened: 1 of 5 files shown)


engine/execution/computation/computer/computer_test.go
```diff
@@ -947,3 +952,15 @@ func generateEvents(eventCount int, txIndex uint32) []flow.Event {
 	}
 	return events
 }
+
+func ensureEventsIndexSeq(t *testing.T, events []flow.Event, chainID flow.ChainID) {
+	expectedEventIndex := uint32(0)
+	for _, event := range events {
+		require.Equal(t, expectedEventIndex, event.EventIndex)
+		if testutil.IsServiceEvent(event, chainID) {
+			expectedEventIndex += 2
```
Tonix517, line 961: In our code at https://github.com/onflow/flow-go/blob/master/fvm/environment/event_emitter.go#L154-L155, for service events we do double counting (original behavior), so the event after one service event will have index as `+2`. 
However if `AppendServiceEvent()` fails and `AppendEvent()` is skipped, then if `EmitEvent()` is called again, we can continue to have unique index values. If this use case can happen, we may want to keep the existing counting logic.

janezpodhostnik: This looks like a bug. Can you open an issue for it?

janezpodhostnik (approved):
Nice!

fvm/fvm_fuzz_test.go
```diff
@@ -307,3 +310,25 @@ func bootstrapFuzzStateAndTxContext(tb testing.TB) (bootstrappedVmTest, transact
 			chain:        bootstrappedVMTest.chain,
 		}
 }
+
+func ensureEventsIndexSeq(t *testing.T, events []flow.Event, chainID flow.ChainID) {
```
pattyshack, line 314: you're duplicating these two functions.  maybe add it to a shared testutil

pattyshack (approved):
see my duplicated code comment, but otherwise i don't have issues with this

engine/execution/computation/computer/computer_test.go
Tonix517: filed #3393

Tonix517:
bors merge

Tonix517:
will merge it first and treat https://github.com/onflow/flow-go/issues/3393 as a separate fix in future PR

PR #3396, opened 2022-10-14
"Charge" users for contract updates.

pattyshack:
In reality, this is a no-op change since flushing pending contracts does not modify computation/memory usage (it only changes interaction count).

This moves all metered execution into a single continuous section in the normal execution path, and simplifies fees deduction a bit since we no longer need to worry that metered usage may change afterward.


Tonix517 (approved):
nice

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3398, opened 2022-10-14
code gen parseRestrict wrappers

pattyshack:
Prep work for adding trace/meter logic to the wrapper.

(I don't really like python3 and haven't done any serious development using it, but ppl will probably hate me if I use python2 instead)


Tonix517:
exciting initiative. 

- question: why use Python instead of Golang's own `text/template`? if we add new language in flow-go, we may want to update the `install-tools` target in the main Makefile, though i'd still prefer not to introduce new language dep into the repo, unless there's strong reasons for that.
- you may also want to add this new code generation step into https://github.com/onflow/flow-go/blob/master/Makefile#L122
- accidental commit of that binary file ".codes.go.swap"?

pattyshack:
1. python cuz i was lazy.  but it's a fair point that not everyone will have python install.  i'll convert to golang instead (i hate python3, but I hate golang text/template even more)

2. i don't think we need to regenerate these regularly

3. yeah, .codes.go.swap was an accidental commit

Tonix517:
did a quick search on github and found this: https://github.com/dave/jennifer. not sure if it's battle-tested though.

pattyshack:
(my hobby template generator looks better https://github.com/pattyshack/abc/tree/master/src/template =P)

i'll probably go with (gogo)proto style generation instead of text/template.  it works pretty well in practice.

pattyshack:
switched to golang based code gen + added to makefile

fvm/environment/generate-wrappers/main.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"bytes"
+	"fmt"
+	"go/format"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const header = `// AUTO-GENERATED BY %s.  DO NOT MODIFIED.
```
Tonix517, line 12: nit: typo "MODIFY"

Tonix517 (approved):
LGTM - please remove `.codes.go.swp` before merging.

pattyshack:
> LGTM - please remove `.codes.go.swp` before merging.

.codes.go.swp is already in the main branch (I merged it accidently in a previous pr).  this removes it

janezpodhostnik (approved):
Automation is always nice.

Can you just double check that `go fmt` doesn't have anything to complain about in the generated file.

fvm/environment/generate-wrappers/main.go
```diff
@@ -0,0 +1,186 @@
...
+
+	generateWrapper(1, 2, content)
+
+	buffer := &bytes.Buffer{}
+	_, err := content.WriteTo(buffer)
+	if err != nil {
+		panic(err) // This should never happen
+	}
+
+	source, formatErr := format.Source(buffer.Bytes())
```
pattyshack, line 169: @janezpodhostnik the code generator runs go fmt here

pattyshack:
bors merge

pattyshack:
bors retry

PR #3400, opened 2022-10-17
Remove read last trie

zhangchiqing:
The `ReadLastTrieRootHashFromCheckpoint` is incorrect for checkpoint V6. 

Since state extraction will always specify block hash, we could simply remove this function.

### DIFF
--- a/cmd/util/cmd/execution-state-extract/cmd.go
+++ b/cmd/util/cmd/execution-state-extract/cmd.go
@@ -3,14 +3,12 @@ package extract
 import (
 	"encoding/hex"
 	"fmt"
-	"os"
 	"path"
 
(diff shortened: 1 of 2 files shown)


ramtinms:
A common use case for this when Cadence team wants to run their tooling over latest checkpoints collected from one of the running execution nodes in the active spork. We could not stop the nodes to take an snapshot and we don't need to transfer the whole protocol state for this goal.

SaveTheRbtz (approved):

zhangchiqing:
I created an issue for tracking adding the functionality back.

I will merge this for now, and the current tool is broken.

zhangchiqing:
bors merge

PR #3403, opened 2022-10-17
[FVM][Execution] Adding metrics for computation effort vector

Tonix517:
Now emitting the raw(unweighted) computation intensities at per-block level, for sampled transactions only. So even if there are many items in the computation effort vector, we don't expect there'll be obvious extra overhead for network traffic.

Also, we don't apply `stringer` on the constants of `ComputationKind` [defined](https://github.com/onflow/flow-go/blob/master/fvm/environment/meter.go#L14) in flow-go:
- [stringer](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) does not support generating string() for the cross-module constant values - in our case the type `ComputationKind` is defined in Cadence, so unless we define a new type in fvm, stringer won't generate String() for FVM's values.
- Instead we utilize Grafana override feature to map certain value like `ComputationKind(2020)` to the actual command string manually.

**Test**
- [X] show dashboard for localnet 
unweighted_tx_transx


engine/execution/messages.go
```diff
@@ -17,18 +18,19 @@ type ComputationOrder struct {
...
+	ExecutableBlock        *entity.ExecutableBlock
+	StateSnapshots         []*delta.SpockSnapshot
+	StateCommitments       []flow.StateCommitment
+	Proofs                 [][]byte
+	Events                 []flow.EventsList
+	EventsHashes           []flow.Identifier
+	ServiceEvents          flow.EventsList
+	TransactionResults     []flow.TransactionResult
+	ComputationUsed        uint64
+	ComputationIntensities meter.MeteredComputationIntensities
```
janezpodhostnik, line 30: Can you check if these (+memory) could be packaged into their own struct called `MeteringResults`?

fvm/transactionInvoker.go
```diff
@@ -171,6 +171,9 @@ func (i TransactionInvoker) Process(
 	proc.Logs = append(proc.Logs, env.Logs()...)
 	proc.ComputationUsed = proc.ComputationUsed + env.ComputationUsed()
 	proc.MemoryEstimate = proc.MemoryEstimate + env.MemoryEstimate()
+	if proc.IsSampled() {
+		proc.ComputationIntensities = env.ComputationIntensities()
+	}
```
janezpodhostnik, line 176: maybe wrap these in a function.

engine/execution/computation/computer/computer.go
```diff
@@ -456,6 +458,9 @@ func (e *blockComputer) executeTransaction(
 	res.AddServiceEvents(tx.ServiceEvents)
 	res.AddTransactionResult(&txResult)
 	res.AddComputationUsed(tx.ComputationUsed)
+	if tx.IsSampled() {
```
janezpodhostnik, line 461: This condition probably isn't necessary. Just make sure `MergeComputationEffortVector` can handle nil values.

janezpodhostnik (approved):
Some suggestions.

engine/execution/messages.go
```diff
@@ -47,6 +49,13 @@ func (cr *ComputationResult) AddComputationUsed(inp uint64) {
 	cr.ComputationUsed += inp
 }
 
+func (cr *ComputationResult) MergeComputationEffortVector(
+	computationIntensities meter.MeteredComputationIntensities) {
```
pattyshack, line 53: add newline to clarify end of method signature.

Tonix517:
for clarity i'll address your comments in a separate PR specifically for refactoring.

Tonix517:
bors merge

Tonix517:
after [discussion](https://docs.google.com/document/d/1b0HtraCXs57o44JexkbidXj4F-iAlhPuVFj7NsCrNPc/edit?usp=sharing) we decided to use state-replay cache to do cached state reads while keeping the original reg touch records.

Tonix517:
mistook it with #3240

Tonix517:
bors merge

PR #3404, opened 2022-10-18
[Common Provider Engine] Adds inbound queue

yhassanzadeh13:
Addresses https://github.com/dapperlabs/flow-go/issues/6174


engine/common/provider/engine.go
```diff
@@ -3,22 +3,31 @@ package provider
 import (
 	"errors"
 	"fmt"
+	"math/rand"
```
peterargue, line 6: please use `crypto/rand`

engine/common/provider/engine.go
```diff
@@ -3,22 +3,31 @@ package provider
...
 	"github.com/onflow/flow-go/module/metrics"
 	"github.com/onflow/flow-go/network"
 	"github.com/onflow/flow-go/network/channels"
 	"github.com/onflow/flow-go/state/protocol"
 	"github.com/onflow/flow-go/storage"
+	"github.com/onflow/flow-go/utils/logging"
 )
 
+// DefaultRequestProviderWorkers is the default number of workers used to process entity requests.
+const DefaultRequestProviderWorkers = uint(100)
```
peterargue, line 29: 100 seems like a lot. any idea what volume of requests we see today?

engine/common/provider/engine.go
```diff
@@ -30,40 +39,81 @@ type RetrieveFunc func(flow.Identifier) (flow.Entity, error)
...
+		engine.Pattern{
+			// Match is called on every new message coming to this engine.
+			// Provider engine only expects EntityRequest.
+			// Other message types are discarded by Match.
+			Match: func(message *engine.Message) bool {
+				request, ok := message.Payload.(*messages.EntityRequest)
+				if ok {
+					log.Info().
+						Str("entity_ids", fmt.Sprintf("%v", request.EntityIDs)).
+						Hex("requester_id", logging.ID(message.OriginID)).
```
peterargue, line 93: let's keep the field name consistent across messages for easier searching
```suggestion
						Hex("origin_id", logging.ID(message.OriginID)).
```

engine/common/provider/engine.go
```diff
@@ -73,81 +123,57 @@ func New(log zerolog.Logger, metrics module.EngineMetrics, net network.Network,
...
-	default:
-		return engine.NewInvalidInputErrorf("invalid message type (%T)", message)
+	err := e.requestHandler.Process(originID, event)
+	if err != nil {
+		if engine.IsIncompatibleInputTypeError(err) {
+			e.log.Warn().
+				Hex("origin_id", logging.ID(originID)).
+				Str("channel", channel.String()).
+				Str("event", fmt.Sprintf("%+v", event)).
+				Msgf("received unsupported message type")
```
peterargue, line 159: This would mean a node sent a different message type on this channel, and none of the authorization checks detected it.  please add `Bool(logging.KeySuspicious, true).` here to make this easier to identify in the logs.

engine/common/provider/engine.go
```diff
@@ -210,7 +236,7 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
 
 	// send back the response
 	res := &messages.EntityResponse{
-		Nonce:     req.Nonce,
+		Nonce:     rand.Uint64(),
```
peterargue, line 239: why reset it here? I thought the purpose of the nonce was to correlate request/responses on the requester side

engine/common/provider/engine.go
```diff
@@ -224,3 +250,66 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
...
+func (e *Engine) processQueuedRequestsShovellerWorker(ctx irrecoverable.SignalerContext, ready component.ReadyFunc) {
+	ready()
+
+	e.log.Debug().Msg("process entity request shoveller worker started")
+
+	for {
+		select {
+		case <-e.requestHandler.GetNotifier():
+			// there is at least a single request in the queue, so we try to process it.
+			e.shovelEntityRequests()
```
peterargue, line 263: I find this terminology confusing. Elsewhere we use `processAvailableMessages`. Can you switch to that?

module/mempool/entity_requests.go
```diff
@@ -0,0 +1,14 @@
+package mempool
+
+import "github.com/onflow/flow-go/engine"
+
+const DefaultEntityRequestCacheSize = 100_000
+
+// EntityRequestStore is a FIFO (first-in-first-out) size-bound queue for maintaining EntityRequests.
+// It is designed to be utilized at the common ProviderEngine to maintain and respond entity requests.
```
peterargue, line 8:
```suggestion
// It is designed to be utilized at the common ProviderEngine to maintain and respond to entity requests.
```

module/mempool/queue/entityRequests.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/onflow/flow-go/model/messages"
+	"github.com/onflow/flow-go/module"
+	"github.com/onflow/flow-go/module/mempool"
+	herocache "github.com/onflow/flow-go/module/mempool/herocache/backdata"
+	"github.com/onflow/flow-go/module/mempool/herocache/backdata/heropool"
+	"github.com/onflow/flow-go/module/mempool/queue/internal"
+)
+
+// EntityRequestStore is a FIFO (first-in-first-out) size-bound queue for maintaining EntityRequests.
+// It is designed to be utilized at the common ProviderEngine to maintain and respond entity requests.
```
peterargue, line 19:
```suggestion
// It is designed to be utilized at the common ProviderEngine to maintain and respond to entity requests.
```

engine/common/provider/engine.go
```diff
@@ -73,81 +123,57 @@ func New(log zerolog.Logger, metrics module.EngineMetrics, net network.Network,
...
+		return fmt.Errorf("unexpected error while processing engine event: %w", err)
 	}
+
+	return nil
 }
 
-func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityRequest) error {
-	lg := e.log.With().Uint64("nonce", req.Nonce).Str("origin_id", originID.String()).Logger()
+// onEntityRequest processes an entity request message from a remote node.
+// All errors returned by this function are benign and should not cause the engine to crash.
```
jordanschalm, line 169: This kind of error handling is OK, only as an exception from the general rule, for example for side-effect-free functions. Here, and as a general rule, it is dangerous, because it can hide unexpected failure conditions. 

For example:
* any error when requesting identities for the latest finalized block, is not expected
* an error besides NotFound is not expected for the `retrieve` function

It's challenging to update our existing code when dependencies don't have error returns documented. Here's how I suggest we approach it:

**Optimal:** When updating / creating a new function, update the error documentation of the sub-functions called by that function, then implement error handling for documented errors. Then we can correctly document errors for the new/updated function, and don't leave technical debt for later.

If it's too onerous to update error documentation for every sub-function, then for sub-functions we don't update:
* Add a TODO to the sub-function godoc noting that the function's errors must be documented
* If possible, implement error handling at the call-site of the sub-function, so that the new/updated function can have exhaustive error documentation, despite the sub-function, for example by:
  * wrapping errors from the sub-function
  * document reasoning for why certain errors are expected or not

**Backup:** If we aren't going to fix error handling for an updated function, we should add a TODO to the updated function noting the tech debt that errors must be documented. 

```suggestion
// Error returns:
// * InvalidInputError if the list of requested entities is invalid (empty)
// * generic error in case of unexpected failure or implementation bug
```

engine/common/provider/engine.go
```diff
@@ -224,3 +250,66 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
...
+		request, ok := <-e.requestChannel
+		if !ok {
+			e.log.Trace().Msg("processing entity request worker terminated")
+			return
+		}
+		lg := e.log.With().
+			Hex("origin_id", logging.ID(request.OriginId)).
+			Str("requested_entity_ids", fmt.Sprintf("%v", request.EntityIds)).Logger()
+		lg.Trace().Msg("worker picked up entity request for processing")
+		err := e.onEntityRequest(request.OriginId, request.EntityIds)
```
jordanschalm, line 309: Please handle expected errors here, and `ctx.Throw` unexpected errors.

module/mempool/queue/entityRequests.go
```diff
@@ -0,0 +1,103 @@
+package queue
```
jordanschalm, line 1: This is a lot of largely duplicated code and tests (~300 lines), to add a new `HeroCache` for the `EntityRequest` type. We would need to duplicate and maintain this code each time we use `HeroCache` as the queue backend for a different type. To me, this is not worth the trade-off here. 

Why don't we default to using `FifoQueue`, which supports generic entities and is used extensively for this role elsewhere, instead? Alternatively, we could consider modify `HeroCache` to support generic entities.

module/mempool/entity_requests.go
```diff
@@ -0,0 +1,14 @@
+package mempool
+
+import "github.com/onflow/flow-go/engine"
+
+const DefaultEntityRequestCacheSize = 100_000
```
jordanschalm, line 5: This seems much too big to me. I think in the range of 1000-10000 would be better, if not even smaller.

Dropping entity requests generally is low-cost - they will be retried.

module/mempool/queue/entityRequests.go
yhassanzadeh13: Generalized `HeroStore` and `HeroQueue` to support `engine.Message` and `flow.Entity`, respectively in https://github.com/onflow/flow-go/pull/3404/commits/73825e4dc2a8997ba93c5ac4876c4454084b6e8a, and replaced the `ChunkDataPackRequestStore` and `EntityRequestStore` with `HeroStore`.

module/mempool/entity_requests.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/cc2ed9962e43cca9e8994c1395587cf2c037e9b9

engine/common/provider/engine.go
```diff
@@ -30,40 +39,81 @@ type RetrieveFunc func(flow.Identifier) (flow.Entity, error)
...
+						Hex("requester_id", logging.ID(message.OriginID)).
+						Msg("entity request received")
+				}
+				return ok
+			},
+			// Map is called on messages that are Match(ed) successfully, i.e.,
+			// EntityRequest.
+			Map: func(message *engine.Message) (*engine.Message, bool) {
+				return message, true
+			},
```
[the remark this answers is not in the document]
yhassanzadeh13: Based on changes in https://github.com/onflow/flow-go/commit/73825e4dc2a8997ba93c5ac4876c4454084b6e8a, `Map` dereferences the `Message.Payload` for storing in `HeroCache`.

engine/common/provider/engine.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/e54fc258a0f093fe6a15308dec7ae30d7c252910

module/mempool/queue/entityRequests.go
yhassanzadeh13: `EntityRequests` has totally removed and generalized as `HeroStore`.

engine/common/provider/engine.go
yhassanzadeh13: Looking at `mainnet` data, we expect 3-5 requests on average. I think 5 is a safe number of workers hence: https://github.com/onflow/flow-go/pull/3404/commits/323fa27ab4a9d380ddc23b12b9ddfe25fc031316

engine/common/provider/engine.go
yhassanzadeh13: Added flag in here: https://github.com/onflow/flow-go/pull/3404/commits/dea665dc291bb96148bc2af179fe00bb221169b5
However, an unsupported message type will be flagged and detected by the slashing consumer earlier down the pipeline.

engine/common/provider/engine.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/1793639ab5041665d21f8ad3b8c1a9ff538bfce3

engine/common/provider/engine.go
yhassanzadeh13: https://github.com/onflow/flow-go/commit/1793639ab5041665d21f8ad3b8c1a9ff538bfce3

module/mempool/entity_requests.go
yhassanzadeh13: This implementation has been removed and consolidated with `HeroQueue` and `HeroStore`.

engine/common/provider/engine.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/ea22a1c4bf9cb025e44bd71773528ad4ce7c407c

module/mempool/queue/heroQueue.go
```diff
@@ -0,0 +1,71 @@
...
+			heropool.NoEjection,
+			logger.With().Str("mempool", "hero-queue").Logger(),
+			collector),
+		sizeLimit: uint(sizeLimit),
+	}
+}
+
+// Push stores the entity into the queue.
+// Boolean returned variable determines whether push was successful, i.e.,
+// push may be dropped if queue is full or already exists.
```
zhangchiqing, line 35: Now we reuse the same heroQueue module to back the providers in collection and execution nodes. And they also have the same config. 
```
	// DefaultRequestProviderWorkers is the default number of workers used to process entity requests.
	DefaultRequestProviderWorkers = uint(5)

	DefaultEntityRequestCacheSize = 1000
```

However, the number of requests they are serving might be very different. I'm OK to use the same config, but it's better we know whether these config is fine turned. 

For that, we probably need some metrics to know:
- the speed of adding jobs to the queue, meaning the message request is received and waiting to be processed.
- the speed of removing jobs to the queue, meaning the messages have been processed
- the number of jobs remaining in the queue. 
- whether there is message dropped from the queue when it's full

Do we have visibility over the above metrics?

module/mempool/queue/heroQueue.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/module"
+	herocache "github.com/onflow/flow-go/module/mempool/herocache/backdata"
+	"github.com/onflow/flow-go/module/mempool/herocache/backdata/heropool"
+)
+
+// HeroQueue implements a HeroCache-based in-memory queue.
```
zhangchiqing, line 14: It's better to also mention briefly the feature of HeroCache, such as "a key-value cache with zero heap allocation"

engine/errors.go
```diff
@@ -46,6 +46,38 @@ func IsInvalidInputError(err error) bool {
...
+func IsNetworkTransmissionError(err error) bool {
+	var errNetworkTransmissionError NetworkTransmissionError
+	return errors.As(err, &errNetworkTransmissionError)
+}
+
+// NetworkTransmissionError captures the general sentinel errors upon network transmission. It is used to distinguish
+// network transmission errors from other errors.
+type NetworkTransmissionError struct {
+	err error
+}
```
jordanschalm, line 59: Thank you for adding this. This is really useful, given that the network layer doesn't have error returns defined. 

If we don't already have one, it would be great to create a tech debt issue to eventually replace this with a similar error (or errors) which are returned directly directly by `network.Conduit`. That way we can:
* detect unexpected failure cases from calls to `Conduit.*` functions
* define errors in `Conduit` implementation once, rather than wrapping at all `Conduit.*` call-sites

module/mempool/queue/heroQueue.go
```diff
@@ -0,0 +1,71 @@
...
+		// although HeroCache is on no-ejection mode and discards pushes beyond limit,
+		// we save an id computation by just checking the size here.
+		return false
+	}
+
+	return c.cache.Add(entity.ID(), entity)
+}
+
+// Pop removes and returns the head of queue, and updates the head to the next element.
+// Boolean return value determines whether pop is successful, i.e., poping an empty queue returns false.
```
jordanschalm, line 51:
```suggestion
// Boolean return value determines whether pop is successful, i.e., popping an empty queue returns false.
```

module/mempool/queue/heroStore.go
```diff
@@ -0,0 +1,41 @@
...
+// Put enqueues the message into the message store.
+//
+// Boolean returned variable determines whether enqueuing was successful, i.e.,
+// put may be dropped if queue is full or already exists.
+func (c *HeroStore) Put(message *engine.Message) bool {
+	return c.q.Push(internal.NewMessageEntity(message))
+}
+
+// Get pops the queue, i.e., it returns the head of queue, and updates the head to the next element.
+// Boolean return value determines whether pop is successful, i.e., poping an empty queue returns false.
```
jordanschalm, line 32:
```suggestion
// Boolean return value determines whether pop is successful, i.e., popping an empty queue returns false.
```

module/mempool/queue/internal/requestEntity.go
```diff
@@ -0,0 +1,37 @@
+package internal
+
+import "github.com/onflow/flow-go/model/flow"
+
+var _ flow.Entity = (*RequestEntity)(nil)
+
+// RequestEntity is an internal data structure for storing entity requests.
+type RequestEntity struct {
```
jordanschalm, line 8: I think this is unused now, can we remove?

jordanschalm (approved):
Thank you, nice work.

engine/common/provider/engine.go
peterargue: yea, agreed this should never happen. It would mean that somehow a valid message was passed to the wrong engine, or there was a bug somewhere

engine/common/provider/engine.go
```diff
@@ -31,40 +41,93 @@ type RetrieveFunc func(flow.Identifier) (flow.Entity, error)
...
+			// Map is called on messages that are Match(ed) successfully, i.e.,
+			// EntityRequest.
+			Map: func(message *engine.Message) (*engine.Message, bool) {
+				request, ok := message.Payload.(*messages.EntityRequest)
+				if !ok {
+					// should never happen, unless there is a bug.
+					log.Warn().
+						Str("entity_ids", fmt.Sprintf("%v", request.EntityIDs)).
+						Hex("requester_id", logging.ID(message.OriginID)).
```
peterargue, line 108:
```suggestion
						Hex("origin_id", logging.ID(message.OriginID)).
```

engine/common/provider/engine.go
```diff
@@ -212,17 +255,93 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
...
+			close(e.requestChannel)
+			e.log.Trace().Msg("processing entity request worker terminated")
+			return
+		}
+	}
+}
+
+func (e *Engine) processAvailableMessages() {
+	for {
+		msg, ok := e.requestQueue.Get()
```
peterargue, line 294: I think we should check if the context was cancelled first and bail early. otherwise this could process forever if there is a consistent stream of new requests

module/mempool/queue/heroQueue.go
yhassanzadeh13: `HeroCache` provides the following metrics: 

> the speed of adding jobs to the queue, meaning the message request is received and waiting to be processed.

[KeyPutAttempt](https://github.com/onflow/flow-go/blob/ba75711b2133ad556fdf571b92bda4e28f33535a/module/metrics/herocache.go#L113)

> the speed of removing jobs to the queue, meaning the messages have been processed

[KeyRemoved](https://github.com/onflow/flow-go/blob/ba75711b2133ad556fdf571b92bda4e28f33535a/module/metrics/herocache.go#L141)

> the number of jobs remaining in the queue.

[Size](https://github.com/onflow/flow-go/blob/ba75711b2133ad556fdf571b92bda4e28f33535a/module/metrics/herocache.go#L92)

> whether there is message dropped from the queue when it's full.

[KeyPutDrop](https://github.com/onflow/flow-go/blob/ba75711b2133ad556fdf571b92bda4e28f33535a/module/metrics/herocache.go#L120)

engine/common/provider/engine.go
```diff
@@ -212,17 +255,93 @@ func (e *Engine) onEntityRequest(originID flow.Identifier, req *messages.EntityR
...
+		if !ok {
+			// no more requests, return
+			return
+		}
+
+		requestEvent, ok := msg.Payload.(messages.EntityRequest)
+		if !ok {
+			// should never happen, as we only put EntityRequest in the queue,
+			// if it does happen, it means there is a bug in the queue implementation.
+			e.log.Fatal().Msg("invalid entity request type")
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/faeae648d92155f45b8d07f6332a89a07b9d6f8f

engine/errors.go
yhassanzadeh13: Added a place holder issue https://github.com/dapperlabs/flow-go/issues/6418 and will scope it shortly.

module/mempool/queue/internal/requestEntity.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/c067dca68997a8a1274f2dbb80fb8deaf8e9ba6c

engine/common/provider/engine.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3404/commits/ffeb6415857a892e7f1e105c677eae7948439482

peterargue (approved):

vishalchangrani:
bors merge

PR #3406, opened 2022-10-18
[Execution] Aggressive collection retries

SaveTheRbtz:
Based on the logs, more than 95% of collection requests are currently handled under a second. Hence there is no need to stall for minutes in case of a misbehaving collection node.

Given that we produce a block ~1.4 seconds geometric wait of 4, 8, 16, 32, 64, 128, etc seconds causes visible stalls even on a single failure. Hence switched to a simple linear backoff with base of 1 and max of 10.


engine/common/requester/engine.go
```diff
@@ -59,9 +59,9 @@ func New(log zerolog.Logger, metrics module.EngineMetrics, net network.Network,
 	cfg := Config{
 		BatchThreshold:  32,
 		BatchInterval:   time.Second,
-		RetryInitial:    4 * time.Second,
-		RetryFunction:   RetryGeometric(2),
-		RetryMaximum:    2 * time.Minute,
+		RetryInitial:    1 * time.Second,
+		RetryFunction:   RetryLinear(1 * time.Second),
+		RetryMaximum:    10 * time.Second,
```
jordanschalm, line 64: This is changing the default config for the `requester` engine in general, which is used for several different entity types. We only want to modify the config for the engine configured with the `request-collections` channel, which is set up here: https://github.com/onflow/flow-go/blob/master/cmd/execution_builder.go#L742-L751.

We should also be sure to change the `BatchInterval` (set to 1m in the link above). `BatchInterval` is the interval at which the engine will send off requests, without outside intervention. `Retry*` configs are used to decide whether to include a particular item in a batch request, not when to submit the request. So with a `BatchInterval` of 1m and a 1s retry, we may still be sending requests once per minute.

jordanschalm (changes requested):

engine/common/requester/engine.go
SaveTheRbtz: > So with a `BatchInterval` of 1m and a 1s retry, we may still be sending requests once per minute.

Good point!  `matchOrRequestCollections` does force the re-try almost constantly, so we should not retry on the batch interval too much. That said, lowered since 1 minute seems like an overkill.

jordanschalm (approved):

ramtinms (approved):
It looks good to me, the only thing we might be careful here is to set default values as constants and make sure its documented in the constant parts why we choose these values, changing the values for the linear retry in the future without proper knowledge by other team members might result in execution nodes DoS the collection nodes in case of collection nodes falling behind or other interruptions in the network.

cmd/execution_builder.go
```diff
@@ -711,6 +711,9 @@ func (exeNode *ExecutionNode) LoadIngestionEngine(
 		func() flow.Entity { return &flow.Collection{} },
 		// we are manually triggering batches in execution, but lets still send off a batch once a minute, as a safety net for the sake of retries
 		requester.WithBatchInterval(exeNode.exeConf.requestInterval),
+		requester.WithRetryFunction(requester.RetryLinear(exeNode.exeConf.requestRetryDelay)),
+		requester.WithRetryInitial(exeNode.exeConf.requestRetryInitial),
+		requester.WithRetryMaximum(exeNode.exeConf.requestRetryMaximum),
```
pattyshack, line 716: When i read requestRetryMaximum, i think of maximum number of retries rather than maximum delay

maybe rename the config fields to requestRetryAdditionalDelay, requestRetryInitialDelay and requestRetryMaximumDelay.

SaveTheRbtz:
bors merge

PR #3402, opened 2022-10-17
Update depenencies to use onflow/go-bitswap fork

peterargue:
Temporarily use [`onflow/go-bitswap` fork](https://github.com/onflow/go-bitswap) of `ipfs/go-bitswap` to pull in an unmerged fix.

PR #3376, opened 2022-10-13
[BFT] add suspicious flag to a few new log messages

peterargue:
This is a follow on from https://github.com/onflow/flow-go/pull/3346.

Add the `suspicious` flag to a few more log events.

PR #3410, opened 2022-10-18
Backport bitswap fix and bft logging

peterargue:
Backports
* https://github.com/onflow/flow-go/pull/3402
* https://github.com/onflow/flow-go/pull/3376

### DIFF
--- a/network/p2p/network.go
+++ b/network/p2p/network.go
@@ -26,6 +26,7 @@ import (
 	"github.com/onflow/flow-go/network/p2p/conduit"
 	"github.com/onflow/flow-go/network/queue"
 	_ "github.com/onflow/flow-go/utils/binstat"
+	"github.com/onflow/flow-go/utils/logging"
 )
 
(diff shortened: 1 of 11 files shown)


network/p2p/network.go
```diff
@@ -516,7 +517,10 @@ func (n *Network) queueSubmitFunc(message interface{}) {
 
 	eng, err := n.subscriptionManager.GetEngine(qm.Target)
 	if err != nil {
-		logger.Err(err).Msg("failed to submit message")
+		// This means the message was received on a channel that the node has not registered an engine for
```
vishalchangrani, line 520: but will the node ever receive a message on a channel it didn't subscribe to? wouldn't libp2p just discard that message without submitting it to the application?

peterargue: it's possible for unicast cases. There should be validation to prevent this, so it would be a bug/unexpected

PR #3420, opened 2022-10-19
[BFT Testing][CI][Testing] create `insecure` module for forked `libp2p-pubsub`, CI and flaky test updates

gomisha:
This PR
1) adds `libp2p-pubsub` support to BFT Testing Framework by making `insecure` into a separate module so we can use a forked version of `libp2p-pubsub`
2) updates to Flaky Test Monitor to make it more streamlined
3) CI updates to remove `crypto` building redundancies and streamline how we unit test standalone modules (`insecure`, `crypto`, `integration`)

### BFT Testing
In order to add support for network level attacks and testing, we need to create a corrupt version of libp2p's PubSub module.

We can do this by forking the repo into a personal Github account and then including it in our BFT Testing Framework as a dependency.

Public version of libp2p-pubsub: https://github.com/libp2p/go-libp2p-pubsub
Forked version of libp2p-pubsub: https://github.com/yhassanzadeh13/go-libp2p-pubsub

- created `insecure` standalone module in order to use forked `libp2p-pubsub` - this was necessary because the main `flow-go` code uses the public (non-forked) `libp2p-pubsub` and we need the BFT testing framework to use the forked version
- updated `Makefile` to build mocks and Docker images of `insecure` in a modified way

### CI Updates
- removed `crypto` building redundancy for each unit test job so `crypto` only built once instead of twice for each job
- added matrix driven CI job for unit testing stand alone modules (`crypto`, `insecure`, `integration`) since they're not available with the regular `Unit Tests` CI job that uses dynamic test matrix
- added race detection control from CI to `crypto/Makefile`, `insecure/Makefile`, `integration/Makefile`

### Flaky Test Updates
- removed `tools/test_monitor/test-setup.sh` from CI workflows; all test prep and test running logic handled by `tools/test_monitor/run-tests.sh`

### To Do
The following tasks need to be done when this PR is ready for merging (after all reviews are done).

- [x] before PR merge: final test for Flaky Test Monitor
- [ ] after PR merge: update CI required checks for new CI jobs 1) `unit-test-modules` 2) `tidy `


crypto/Makefile
```diff
@@ -13,26 +20,29 @@ setup:
...
 ############################################################################################
 # CAUTION: DO NOT MODIFY THIS TARGET! DOING SO WILL BREAK THE FLAKY TEST MONITOR
 
 # runs all tests of the crypto package 
 .PHONY: test
-test: relic_tests non_relic_tests
+test: setup test_main
+
+.PHONY: test_main
+test_main: relic_tests non_relic_tests
```
tarakby, line 45:
```suggestion
# sets up the crypto module and runs all tests
.PHONY: test
test: setup unittest

# runs the unit tests of the module (assumes the module was set up)
.PHONY: unittest
test_main: relic_tests non_relic_tests
```

Makefile
```diff
@@ -65,34 +60,33 @@ install-mock-generators:
...
+
+############################################################################################
 
 .PHONY: emulator-build
 emulator-build:
 	# test the fvm package compiles with Relic library disabled (required for the emulator build)
 	cd ./fvm && go test ./... -run=NoTestHasThisPrefix
 
-.PHONY: emulator-build
+.PHONY: fuzz-fvm
```
tarakby, line 83: This target isn't called by any job. Is that expected?

.github/workflows/ci.yml
```diff
@@ -119,6 +135,62 @@ jobs:
...
+          cache: true
+      - name: Run tests (${{ matrix.name }})
+        if: github.actor != 'bors[bot]'
+        env:
+          RACE_DETECTOR: ${{ matrix.race }}
+        # run `make1` target before running `make2` target inside each module's root
+        run: |
+          make ${{ matrix.make1 }}
+          cd ${{ matrix.name }}
+          make ${{ matrix.make2 }}
```
tarakby, line 176: These 2 lines could be replaced by:
```suggestion
make -C ${{ matrix.name }} ${{ matrix.make2 }}
```

.github/workflows/ci.yml
```diff
@@ -53,6 +53,22 @@ jobs:
         # https://github.com/golangci/golangci-lint-action/issues/244
         skip-cache: true
 
+  tidy:
+    name: Tidy Emulator
```
tarakby, line 57: Why is there a tidy job only for the emulator?
Btw `make emulator-build` is only there to make sure the emulator can build. It's a very small job that could be added to another job.

.github/workflows/ci.yml
```diff
@@ -119,6 +135,62 @@ jobs:
...
         name: codecov-umbrella
 
+  unit-test-modules:
+    name: Unit Tests (Modules)
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: crypto
+            make1: no-op
```
tarakby, line 145: I think it's clearer if we can see the setup phase on the CI level:
```suggestion
make1: -C crypto setup
```

That way we can also delete the new job added under `crypto/Makefile` (`test_main`).

tarakby: We could even call `make1` as `setup` , and `make2` as `test` for clarity.

.github/workflows/ci.yml
gomisha: I grouped both the Tidy and Emulator build steps into a single job because they're both so small. The Tidy job is for the entire code base, not just for the emulator.

gomisha: I created a new job for building the emulator bd55f35f955c9c3eef3dd0cb79db8bd2d9c39ea7

.github/workflows/ci.yml
gomisha: Yeah, good idea. I updated it in 639938f7074ad10068f847dcec8830060fcb546f

.github/workflows/ci.yml
gomisha: Good idea - updated in 406e9f80be1eb8b3d1880233b0488c72074cf0b2

Makefile
gomisha: Yes - that fuzz job doesn't end and keeps fuzzing until you manually stop it so I think it's meant for local running only. I thought it was an error that it was tagged with the same target name `emulator-build` as the actual emulator building target so I renamed it.

crypto/Makefile
gomisha: Updated in 6eec8310fde4775e5192951abd585420295d78e2

yhassanzadeh13 (approved):
LGTM, thanks Misha.

.github/workflows/ci.yml
gomisha: I updated the emulator job with a new name and put it back into the same job as the main `Tidy` job
https://github.com/onflow/flow-go/pull/3420/commits/107911a774c50c871dfe21d0239c07a2f2627d19

tarakby (approved):
Thanks for including all changes and sparing our CI hours of redundant work 🙏🏼

PR #3421, opened 2022-10-19
[Crypto] Identity BLS public keys

tarakby:
This PR updates the BLS-primitives to take into account edge case issues related to identity public keys. To implement these changes, a new field is added to the BLS `PublicKey` structure to cache whether the point is identity in G_2 (point at infinity). The field is written at the public key creation and is cached for faster access.

- BLS signatures: 
   -  signature verification: The [BLS IETF draft](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#section-5.2) avoids signature equivocation issues by verifying a public key is not identity during signature verification. This was enforced previously in the package by guaranteeing a decoded public key is never identity. However an aggregated public key could be equal to identity and be used afterwards in signature verifications. The package is now updated to enforce a public key check against identity in all signature verification functions. Moreover, decoding an identity public key has become possible, in case some applications need to encode/decode aggregated public keys. 
   -  identity signature: With the above update, applications that aggregate signatures may fall in the case where the resulting signature is equal to identity (and therefore is verifiable only against an identity public key). For uniformly distributed public keys, this case can only happen with a negligible probability. However, chosen private keys (in a trustless environment for instance) can easily lead to an identity signature. The package adds a new function to check whether a signature is an identity signature.

- BLS-DKG:
   - Feldman VSS, Feldman VSSQ and Joint-Feldman instances now fail locally for every participant that receives/computes a group public key or public key share equal to identity. This is considered a failure because any signature generated by the respective private key fails signature verification. This case can happen in Feldman VSS and Feldman VSSQ with a malicious dealer. It can also happen in Joint-Feldman if all qualified dealers are malicious (which contradict the safety pre-requisite of malicious dealers being less that the threshold). The failure check is added in Joint-Feldman although the probability of getting local identity keys is negligible (even with the known bias of public key distribution). 
   - Feldman VSSQ was not updated to disqualify the dealer when the verification vector shows a group public key or public key share equal to identity. This is because Feldman VSSQ is the building block of Joint-Feldman where all dealer's keys add up. Enforcing keys per dealer to not be identity does not add any guarantee on final joint-Feldman keys to not be identity.
 

- BLS-SPoCK:
In a SPoCK verification involving (pk1, p1, pk2, p2), the pairing check would succeed if pk1 and p1 are identities in respectively G_2 and G_1, for any given (pk2, p2). This would be a forgery issue as the identity couple is valid against any (key, proof) pair. 
To avoid this issue, SPoCK verification now includes a check for pk1 and pk2 against identities. Note that public keys are used in a SPoCK verifications only if their proofs of possession have been verified successfully. An identity public key should fail a PoP verification and should not be used in a SPoCK verifications. To avoid making assumptions about the PoP verification implementation, the identity check is added also to the SPoCK verification implementation. 
 
 Other updates:
  - more documentation around the identity key behaviour
  - new tests to cover most of the identity public key cases
  - more documentation related to signature verification pre-conditions related to the PoP scheme.


jordanschalm (commented):
Not through everything yet, leaving a first round of comments.

crypto/bls_multisig.go
```diff
@@ -162,7 +167,10 @@ func AggregateBLSPrivateKeys(keys []PrivateKey) (PrivateKey, error) {
 //
 // The order of the keys in the slice does not matter since the aggregation
 // is commutative. The slice should not be empty.
-// No check is performed on the input public keys.
+// No check is performed on the input public keys. The input keys are guaranteed to
+// be on the G2 subgroup.
```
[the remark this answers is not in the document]
tarakby: Yes exactly. I added clarifications to the doc.

crypto/bls_thresholdsign.go
```diff
@@ -572,9 +572,12 @@ func BLSThresholdKeyGen(size int, threshold int, seed []byte) ([]PrivateKey,
...
 	a := make([]scalar, threshold+1)
-	randZrStar(&a[0])
-	for i := 1; i < threshold+1; i++ {
-		randZr(&a[i])
+	randZrStar(&a[0]) // non-identity key
+	if threshold > 0 {
+		for i := 1; i < threshold+1; i++ {
+			randZrStar(&a[i])
+		}
+		randZrStar(&a[threshold]) // enforce the polynomial degree
```
jordanschalm, line 580: It seems like we are generating keys 0 and 1-n separately, and generating key n twice? I'm not sure why it's structured this way, rather than simply:
```suggestion
	for i := 0; i < threshold+1; i++ {
		randZrStar(&a[i])
	}
```

crypto/dkg_feldmanvss.go
```diff
@@ -256,17 +267,23 @@ func (s *feldmanVSSstate) generateShares(seed []byte) error {
 	s.a = make([]scalar, s.threshold+1)
```
jordanschalm, line 267: It would be great to update the [godoc for this function](https://github.com/onflow/flow-go/pull/3421/files#diff-5ae00e788402d6f6773388800a459c091ef6e6ca6872c186de3ad0df1ec30e13R259), it looks unfinished.

crypto/dkg_feldmanvss.go
```diff
@@ -256,17 +267,23 @@ func (s *feldmanVSSstate) generateShares(seed []byte) error {
...
+	randZrStar(&s.a[0])
 	generatorScalarMultG2(&s.vA[0], &s.a[0])
-	for i := 1; i < s.threshold+1; i++ {
-		C.bn_new_wrapper((*C.bn_st)(&s.a[i]))
-		randZr(&s.a[i])
-		generatorScalarMultG2(&s.vA[i], &s.a[i])
+	if s.threshold > 0 {
+		for i := 1; i < s.threshold; i++ {
+			C.bn_new_wrapper((*C.bn_st)(&s.a[i]))
+			randZr(&s.a[i])
```
jordanschalm, line 276: Why do we allow maybe-zero values for `a[1:threshold]`, but not `a[0]` and `a[threshold]`? I'm also wondering why this differs from the [similar code block in threshold signing](https://github.com/onflow/flow-go/pull/3421/files#diff-09d2f77aa5f9d78dc6b7d19a2c640a9fb7710671880625092cc4b1cfece6bd67R575-R580), where we use non-zero values for all entries in `a`

crypto/dkg_jointfeldman.go
```diff
@@ -211,6 +212,22 @@ func (s *JointFeldmanState) End() (PrivateKey, PublicKey, []PublicKey, error) {
...
+
+	// check if current public key share or group public key is identity.
+	// In that case all signatures generated by the current private key share or
+	// the group private key are invalid (as stated by the BLS IETF draft)
+	// to avoid equivocation issues.
+	//
+	// Assuming both private keys have entropy from at least one honest dealer, each private
+	// key is initially uniformly distributed over the 2^255 possible values. We can argue that
+	// the known uniformity-bias caused by malicious dealers in Joint-Feldman does not weaken
+	// the likelihood of generarting an identity key to practical probabilities.
```
jordanschalm, line 224:
```suggestion
	// the likelihood of generating an identity key to practical probabilities.
```

crypto/spock_test.go
```diff
@@ -67,6 +67,15 @@ func TestSPOCKProveVerifyAgainstData(t *testing.T) {
 		assert.True(t, IsNotBLSKeyError(err))
 		assert.False(t, result)
 	})
+
+	// test with an identity public key
```
jordanschalm, line 71:
```suggestion
	// verifying an identity signature against the identity public key should return false
```

crypto/spock_test.go
```diff
@@ -154,4 +163,17 @@ func TestSPOCKProveVerify(t *testing.T) {
 		assert.True(t, IsNotBLSKeyError(err))
 		assert.False(t, result)
 	})
+
+	// test with identity public key and proof
```
jordanschalm, line 167:
```suggestion
	// verifying any SPOCK where either (pk, proof) pair is the identity public key and identity proof, should return false 
```

crypto/spock_test.go
```diff
@@ -154,4 +163,17 @@ func TestSPOCKProveVerify(t *testing.T) {
...
+	t.Run("identity proof", func(t *testing.T) {
+		identityProof := make([]byte, signatureLengthBLSBLS12381)
+		identityProof[0] = identityBLSSignatureHeader
+		result, err := SPOCKVerify(IdentityBLSPublicKey(), identityProof, sk2.PublicKey(), pr2)
+		assert.NoError(t, err)
+		assert.False(t, result)
+
+		result, err = SPOCKVerify(sk1.PublicKey(), pr1, IdentityBLSPublicKey(), identityProof)
+		assert.NoError(t, err)
+		assert.False(t, result)
```
jordanschalm, line 177: Suggesting to test the case whether both pairs are identity key/proof
```suggestion
		assert.False(t, result)
		
		result, err = SPOCKVerify(IdentityBLSPublicKey(), identityProof, IdentityBLSPublicKey(), identityProof)
		assert.NoError(t, err)
		assert.False(t, result)
```

crypto/dkg_feldmanvss.go
tarakby: Good question. This is related to Shamir secret sharing, where we need to generate a polynomial `P` of degree `t` : P(x) = a_t.x^t + ... + a_1.x + a_0 
To make sure the polynomial is of degree `t` (and avoid possible security issues related to the threshold), we need to enforce `a_t` is non-zero (otherwise `P` is of degree at most `t-1` ).

Moreover, in Shamir secret sharing, the secret (here private key) is `a_0`. As part of this PR where we don't want to allow identity private keys and identity public keys, we want to enforce `a_0` is non-identity (non-zero). The rest of the values `a_i` do not matter.

crypto/bls_thresholdsign.go
tarakby: Thanks for pointing this. There are two issues/typos:
 - the `a_i` inside the loop don't have to be non-zero, so `randZr` is enough (that was my intention after separating `a_0` and `a_t`)
 - the loop should stop at `threshold` to not generate twice.

crypto/dkg_feldmanvss.go
tarakby: The other block in threshold signing is buggy https://github.com/onflow/flow-go/pull/3421#discussion_r1006000477

crypto/dkg_feldmanvss.go
tarakby: Can you please specify the function name, I couldn't find it with the link above

crypto/spock_test.go
tarakby: Yep, good suggestion!

crypto/dkg_feldmanvss.go
tarakby: Thanks! doc updated

crypto/bls.go
```diff
@@ -338,7 +391,11 @@ func (sk *prKeyBLSBLS12381) Size() int {
 func (sk *prKeyBLSBLS12381) computePublicKey() {
 	var newPk pubKeyBLSBLS12381
 	// compute public key pk = g2^sk
-	generatorScalarMultG2(&(newPk.point), &(sk.scalar))
+	generatorScalarMultG2(&newPk.point, &sk.scalar)
+
+	// cache the identity comparison
+	newPk.isIdentity = C.bn_is_zero((*C.bn_st)(&sk.scalar)) != valid
```
jordanschalm, line 397: It would be nice to wrap this in a utility function, since we use it several times:
* `bn_is_zero`'s definition [doesn't define its return value in terms of the `VALID` constant](https://github.com/relic-toolkit/relic/blob/15cb6d1dbe03a2780aae27fe6b31682d9c2d0942/include/relic_bn.h#L444-L450)
* `bn_is_zero`'s "true" return value interpreted in terms of the `VALID` constant maps to `INVALID`, which was the opposite of what I'd expect

```suggestion
	newPk.isIdentity = bnIsZero(sk.scalar)
```

crypto/bls.go
```diff
@@ -283,9 +330,15 @@ func (a *blsBLS12381Algo) decodePublicKey(publicKeyBytes []byte) (PublicKey, err
...
 	}
-	if !pk.point.checkValidPublicKeyPoint() {
+
+	// membership check in G2
+	if C.check_membership_G2((*C.ep2_st)(&pk.point)) != valid {
 		return nil, invalidInputsErrorf("input key is infinity or does not encode a BLS12-381 point in the valid group")
 	}
+
+	// check point is non-infinity and cache it
+	pk.isIdentity = C.ep2_is_infty((*C.ep2_st)(&pk.point)) != valid
```
jordanschalm, line 340: Same comment here, for making a wrapper function for this.

crypto/bls.go
```diff
@@ -378,16 +435,22 @@ func (sk *prKeyBLSBLS12381) String() string {
 type pubKeyBLSBLS12381 struct {
 	// public key data
 	point pointG2
+	// G2 identity check cache
```
jordanschalm, line 438:
```suggestion
	// G2 identity check cache - if a public key is identity, the package guarantees it to be cached here after construction
```

jordanschalm (approved):

crypto/bls.go
tarakby: > bn_is_zero's definition doesn't define its return value in terms of the VALID constant

Good point. `valid` constant is mapped to Relic's `RLC_OK` while `bn_is_zero` is simply using `0`.

crypto/bls.go
tarakby: https://github.com/onflow/flow-go/pull/3421/commits/4f3197b9e3843ed4b693744fff1bd4b90174cdbc

crypto/bls.go
tarakby: https://github.com/onflow/flow-go/pull/3421/commits/4f3197b9e3843ed4b693744fff1bd4b90174cdbc

crypto/bls.go
tarakby: https://github.com/onflow/flow-go/pull/3421/commits/5df46201149ad0f7309d0a6d297debe911c8511f

crypto/bls12381_utils.c
```diff
@@ -84,7 +85,10 @@ prec_st* init_precomputed_data_BLS12_381() {
 // Initializes Relic context with BLS12-381 parameters
 ctx_t* relic_init_BLS12_381() { 
     // check Relic was compiled with the right conf 
-    if (ALLOC != AUTO) return NULL;
+    assert(ALLOC == AUTO);
+
+    // sanity check of Relic constants the package is relying on
+    assert(RLC_OK == RLC_EQ);
```
ramtinms, line 91: 👍 do we need to do this for other places that we use Relic?

crypto/bls12381_utils.go
```diff
@@ -100,6 +100,17 @@ func (p *pointG2) equals(other *pointG2) bool {
...
+// Scalar must be already reduced modulo r
+func (x *scalar) isZero() bool {
+	return C.bn_is_zero((*C.bn_st)(x)) == 1
+}
+
+// Comparison to point at infinity in G2.
+func (p *pointG2) isInfinity() bool {
+	return C.ep2_is_infty((*C.ep2_st)(p)) == 1
+}
+
```
ramtinms, line 113: do we have some unit tests for these, it looks a bit complex logic to not have unit tests.

ramtinms (approved):
Looks good to me (though as far as I understand it).

crypto/bls12381_utils.c
tarakby: That function is called under the package main `init()` function, so it's guaranteed to be called whenever the crypto package is used.

crypto/bls12381_utils.go
tarakby: Usually, there aren't specific unit tests for non exported APIs in the crypto module. However, these internal tools are tested implicitly using high level scenarios and edge cases. If there is a problem with them, many exported API tests would fail. 

I've also added a test case to cover a code path that I noticed it may be non covered so far. https://github.com/onflow/flow-go/pull/3421/commits/d54c7ee077adcccc6265938a14a116ec3ff1ac8e

tarakby:
bors merge

PR #3422, opened 2022-10-19
Port rate limited blobstore changes

smnzhu:
port https://github.com/onflow/flow-go/pull/3349/


network/p2p/blob/blob_service.go
```diff
@@ -232,9 +237,9 @@ func (r *rateLimitedBlockStore) Get(ctx context.Context, c cid.Cid) (blocks.Bloc
 		return nil, err
 	}
 
-	err = r.limiter.WaitN(ctx, size)
-	if err != nil {
-		return nil, err
+	allowed := r.limiter.AllowN(time.Now(), size)
+	if !allowed {
+		return nil, rateLimitedError
```
pattyshack, line 242: just `return errors.New("rate limited")` here instead of defining a global constant?  you can use `assert.ErrorContains` instead of `assert.ErrorIs`in the test to check for the message

pattyshack (approved):
o, this is a port to v0.28 / the code is already in the main branch.   maybe address my comment in the main branch only.  approve to unblock

yhassanzadeh13 (approved):

Kay-Zee:
since this is a port to an old branch, going to close

PR #3423, opened 2022-10-19
Refactor transaction invoker.

pattyshack:
1. move errors collection out of the invoker + added test
2. don't commit transaction using defer.  The original defer commit function is also split into validateCommit and commit.  This is prep work for converting the Process() into continuation passing style executor.
3. move error execution path into a method

### DIFF
--- a/fvm/errors/errors_collector.go
+++ b/fvm/errors/errors_collector.go
@@ -1,0 +1,60 @@
+package errors
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+// Utility object for collecting Processor errors.
+type ErrorsCollector struct {
+	errors  error
+	failure error
+}
+
+func NewErrorsCollector() *ErrorsCollector {
+	return &ErrorsCollector{}
+}
+
+func (collector *ErrorsCollector) CollectedFailure() bool {
+	return collector.failure != nil
+}
+
+func (collector *ErrorsCollector) CollectedError() bool {
+	return collector.errors != nil || collector.failure != nil
+}
+
+func (collector *ErrorsCollector) ErrorOrNil() error {
+	if collector.failure == nil {
+		return collector.errors
+	}
+
+	if collector.errors == nil {
+		return collector.failure
+	}
+
+	return fmt.Errorf(
+		"%w (additional errors: %v)",
+		collector.failure,
+		collector.errors)
+}
+
+// This returns collector for chaining purposes.
+func (collector *ErrorsCollector) Collect(err error) *ErrorsCollector {
+	if err == nil {
+		return collector
+	}
+
+	if collector.failure != nil {
+		collector.errors = multierror.Append(collector.errors, err)
+		return collector
+	}
+
+	if IsFailure(err) {
+		collector.failure = err
+		return collector
+	}
+
+	collector.errors = multierror.Append(collector.errors, err)
+	return collector
+}
--- a/fvm/transactionInvoker.go
+++ b/fvm/transactionInvoker.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/hashicorp/go-multierror"
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/rs/zerolog"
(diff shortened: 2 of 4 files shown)


fvm/transactionInvoker.go
```diff
@@ -296,3 +201,106 @@ func (i TransactionInvoker) normalExecution(
...
+			Err(feesError).
+			Msg("transaction fee deduction executed with error")
+
+		if errs.Collect(feesError) {
+			return errs.ErrorOrNil()
+		}
+
+		// drop delta
+		restartErr = txnState.RestartNestedTransaction(nestedTxnId)
+		if errs.Collect(restartErr) {
```
pattyshack, line 244: note the minor change here.  we used to ignore this error.  now we collect and return it.  (in reality, this error should never occur)

fvm/transactionInvoker.go
```diff
@@ -296,3 +201,106 @@ func (i TransactionInvoker) normalExecution(
...
+	// log transaction as failed
+	env.Logger().Info().
+		Err(errs.ErrorOrNil()).
+		Msg("transaction executed with error")
+
+	env.Reset()
+
+	// drop delta since transaction failed
+	restartErr := txnState.RestartNestedTransaction(nestedTxnId)
+	if errs.Collect(restartErr) {
```
pattyshack, line 225: note the minor change here.  This used to be 

```
		if restartErr != nil {
			return fmt.Errorf(
				"cannot restart nested transaction: %w "+
					"(original processErr: %v)",
				restartErr,
				processErr)
		}
```

They are functionally identical

fvm/errors/errors_collector.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// This returns true if the processor should early exit (due to a fatal error).
+func (collector *ErrorsCollector) CollectShouldEarlyExit(err error) bool {
+	if err == nil {
+		return collector.failure != nil
+	}
+
+	if collector.failure != nil {
+		collector.errors = multierror.Append(collector.errors, err)
```
Tonix517, line 45: does it mean that we only keep the first failure, and put all errors afterwards into `errors` even when it is a failure?

Tonix517 (approved):

fvm/errors/errors_collector.go
pattyshack: we should early exit on failure in general so this case should never get triggered.  i added this just in case we forgot to do so (this enables us to track the secondary failure)

SaveTheRbtz (approved):

fvm/transactionInvoker.go
```diff
@@ -296,3 +212,113 @@ func (i TransactionInvoker) normalExecution(
...
+}
+
+func (i TransactionInvoker) validateCommit(
+	proc *TransactionProcedure,
+	txnState *state.TransactionState,
+	nestedTxnId state.NestedTransactionId,
+	env environment.Environment,
+	processErr error,
+) error {
+	if txnState.NumNestedTransactions() > 1 {
```
janezpodhostnik, line 270: invert to reduce nesting.

fvm/transactionInvoker.go
```diff
@@ -45,129 +44,29 @@ func (i TransactionInvoker) Process(
...
-	}
+	errs := errors.NewErrorsCollector()
 
 	env := NewTransactionEnvironment(ctx, txnState, programs, proc.Transaction, proc.TxIndex, span)
 
 	var txError error
 	modifiedSets, txError = i.normalExecution(proc, txnState, env)
-	if mergeErrorShouldEarlyExit(txError) {
-		return
+	if errs.CollectShouldEarlyExit(txError) {
```
janezpodhostnik, line 54: Just a minor suggestion:

Not the biggest fan of the name: `CollectShouldEarlyExit`. Since it is no longer part of the invoker, the `ShouldEarlyExit` becomes a bit un-intuitive what should be exited.

Perhaps this could be in this form: `if errs.Collect(txError).CollectedFailure() {`?

janezpodhostnik (approved):

fvm/transactionInvoker.go
pattyshack: I'll leave this as is for now.  I plan to simply this function (and move procedure result population into `commit()`) after this PR.

Since we now early exit on failures, we no longer need to check for `processErr == nil`.  it's always a programming error.

fvm/transactionInvoker.go
pattyshack: it's a reasonable suggestion.  i'll rename HasError to CollectedError as well

pattyshack:
bors merge

PR #3429, opened 2022-10-20
[Consensus] Own vote & timeout message routing

AlexHentschel:
To the best of my understanding:
*  the [Collector's `MessageHub`](https://github.com/onflow/flow-go/blob/bfb8881ce077251a6cb365cec076c15bd33812d4/engine/collection/message_hub/message_hub.go) does _not_ forward the node's own TOs to the timeout aggregator, which could break liveness in some edge cases. 
* Similarly for votes, the message hub always goes through the networking layer even if the node itself is the next primary and should receive the vote. 

I think it would be more consistent if the `MessageHub` also feeds the node's own Votes and TOs into the node's VoteAggregator and TimeoutAggregator. 

This PR attempts to add this logic. If is for **collector nodes only**. Similar revisions might be **required for consensus nodes as well, which are _not_ contained in this PR**.


durkmurder:
Created a separate issue to implement this: https://github.com/dapperlabs/flow-go/issues/6421

durkmurder:
Originally this PR was created a suggested amendments for https://github.com/onflow/flow-go/pull/3357. Due to quite large scope of that PR we have agreed to extract it into a PR and merge it separately. 
In following commits I have extended original scope of this PR with next changes:
- `VoteAggregator` and `TimeoutAggregator` are no longer injected into `EventHandler`.
- `MessageHub` and `compliance.Core` call `VoteAggregator.AddBlock` to perform vote aggregator
- Routing of votes and timeouts happens exclusively by `MessageHub`, including ones that we have generated.
- Updated how we construct `NoopConsumer` to allow partial mocking without workarounds.

engine/consensus/message_hub/message_hub.go
```diff
@@ -228,10 +228,14 @@ func (h *MessageHub) processQueuedMessages(ctx context.Context) error {
 	}
 }
 
-// processQueuedTimeout performs actual processing of model.TimeoutObject, as a result of successful invocation
-// broadcasts timeout object to consensus committee.
+// processQueuedTimeout propagates the timeout the collector cluster:
```
jordanschalm, line 231:
```suggestion
// processQueuedTimeout propagates the timeout to the consensus committee:
```

jordanschalm (approved):
Nice refactor 👍

engine/collection/message_hub/message_hub.go
```diff
@@ -241,10 +241,14 @@ func (h *MessageHub) processQueuedMessages(ctx context.Context) error {
...
 }
 
-// processQueuedTimeout performs actual processing of model.TimeoutObject, as a result of successful invocation
-// broadcasts timeout object to consensus committee.
+// processQueuedTimeout propagates the timeout to the collector cluster:
+//   - the node's internal `timeoutAggregator`
+//   - broadcast to all other cluster participants (excluding myself)
+//
 // No errors are expected during normal operations.
 func (h *MessageHub) processQueuedTimeout(timeout *model.TimeoutObject) error {
```
AlexHentschel, line 249: I find that there is a subtle design asymmetry for processing votes, proposals and timeouts:
* receiving such a message from a _different node_, we _directly_ forward the message to the `compliance.Engine`, `VoteAggregator` or `TimeoutAggregator` (-> [code](https://github.com/onflow/flow-go/blob/de7fbb380ffc10ded1db53626f7e329dfe789955/engine/collection/message_hub/message_hub.go#L433-L452))
* In comparison, for our own messages, we queue the messages first and then lump together the sending to _external nodes_ as well as forwarding the message to the _internal_ components.

To me, it would be more intuitive to have a symmetric implementation:
* as part of `OnOwnProposal`, `OnOwnTimeout` and `OnOwnVote`, we forward the message to the internal component _right away_. There is no need send the messages to the queue, because the internal components are non-blocking.
* the methods `processQueuedProposal`, `processQueuedTimeout`, `processQueuedVote` and `processQueuedMessages` are only for sending the messages to _other_ peers. We could rename them to `sendOwnProposal`,  `sendOwnTimeout`, `endOwnVote` and `sendOwnMessages`.

To me, this would make the logic more consistent.

engine/collection/message_hub/message_hub.go
```diff
@@ -333,8 +348,12 @@ func (h *MessageHub) processQueuedProposal(header *flow.Header) error {
 
 	log.Debug().Msg("processing cluster broadcast request from hotstuff")
 
+	hotstuffProposal := model.ProposalFromFlow(header)
+	// notify vote aggregator that new block proposal is available, in case we are next leader
+	h.voteAggregator.AddBlock(hotstuffProposal)
+
 	// TODO(active-pacemaker): replace with pub/sub?
-	h.hotstuff.SubmitProposal(model.ProposalFromFlow(header)) // non-blocking
+	h.hotstuff.SubmitProposal(hotstuffProposal) // non-blocking
```
AlexHentschel, line 356: I am not sure, but I am concerned about skipping the compliance engine here. One of the important tasks of the compliance engine is to persist blocks in the DB. 
I think it would help reduce complexity to treat our own blocks exactly the same as external blocks as much as we can. Thereby, we avoid extra code and special cases. In particular, I think it is totally fine for our own blocks to undergo verification. Other nodes have to do the same work before voting. I feel there is little use for the proposer to skip the verification, because the proposer is the only node that can do this, i.e. we don't speed up the consensus hot-path. I just don't think it is worth the extra complexity in the code.

AlexHentschel (commented):
Some preliminary comments ... still need to continue reviewing. 

I also created [issue #3545](https://github.com/onflow/flow-go/issues/3545) to document some tech debt. It might be best to cover this issue in a subsequent PR, but I am also happy if you want to add the revisions to this PR.

engine/collection/message_hub/message_hub.go
durkmurder: block will be added into DB by block builder but now I think it's a good idea to pass it through compliance engine and have a symmetrical approach for handling blocks, even though it does redundant job in terms of verification.

durkmurder: While implementing this I've noticed that compliance engine [skips]() already processed proposals, maybe we would like to move block extension from builder to compliance engine exclusively

durkmurder: Let's try to handle this as part of [this](https://github.com/onflow/flow-go/pull/3512) PR. It has strongly typed interface and it will be easier to replace it when experimenting with builder implementation.

AlexHentschel: I added an issue https://github.com/onflow/flow-go/issues/3552 for later

engine/consensus/message_hub/message_hub.go
```diff
@@ -302,10 +293,14 @@ func (h *MessageHub) processQueuedVote(packed *packedVote) error {
 	return nil
 }
 
-// processQueuedProposal performs actual processing of flow.Header, as a result of successful invocation
-// broadcasts block proposal to consensus committee and rest of Flow network.
+// sendOwnProposal propagates the block proposal to the consensus cluster and submits to non-consensus network:
```
AlexHentschel, line 296: ### ❗ 
In this file, we repeatedly use the terms "cluster consensus" and "consensus cluster", which I find creates unnecessary ambiguity with the collector clusters. 

For the main consensus' `MessageHub` I would suggest to avoid the term "cluster" entirely:

```suggestion
// sendOwnProposal propagates the block proposal to the consensus committee and submits to non-consensus network:
```

For the other two occurrences of "cluster consensus" ([here](https://github.com/onflow/flow-go/blob/5dfb05ce21739c27046e14a497a61e5c99b02fb7/engine/consensus/message_hub/message_hub.go#L48) and [here](https://github.com/onflow/flow-go/blob/5dfb05ce21739c27046e14a497a61e5c99b02fb7/engine/consensus/message_hub/message_hub.go#L62)), we could just remove the word 'cluster'

engine/consensus/message_hub/message_hub.go
```diff
@@ -302,10 +293,14 @@ func (h *MessageHub) processQueuedVote(packed *packedVote) error {
...
+//   - directly forwarded proposal to HotStuff core logic
+//     (skipping compliance engine as we assume our own proposals to be correct)
```
AlexHentschel, line 298: would need to update this comment, if we move the logic for forwarding the proposal to hotstuff and the vote aggregator.

AlexHentschel (commented):
Done with the review. Very concise and clean PR. 
**Please merge**

durkmurder (approved):

durkmurder:
bors merge

PR #3430, opened 2022-10-20
[Consensus] Timeout rebroadcast de-duplication

durkmurder:
This PR is based on next comment:
       
I am not sure that this will actually work in practise. Lets consider the following scenario
*  (simplified)  network topology  `A` <-> `B` <-> `C`
* assume that `C` crashed, and while it is re-booting, `A` and `B` time out. So explicitly `A` broadcasts its TO to its peers `B`. As `C` is offline, it misses the TOs. 
* Then `C` comes back online. As there is still no progress, `A` decides to re-broadcast its TO:
   * It sends exactly the same binary message to the networking layer.
* the networking layer is explicitly designed to filter out duplicates.  In detail: when `B` receives the broadcast-request from `A`, `B`'s networking layer will say: "oh, this is a duplicate. I have already broadcasted this message to my peers, there is nothing else for me to do" .  `B` drops the re-broadcast. And consequently, `A`'s TO never reaches `C` .

Automatic de-duplication is the desired default case. There are a few exceptions, including TOs, where the application layer has to explicitly take on the responsibility of making the different re-broadcast attempts distinguishable on the wire. 

Often, we achieve this by wrapping the protocol-level message (here `TimeoutObject`) and adding a random unit64 as nonce. 

_Outlook (beyond this PR):_
I feel it would be nice to include the tick counter, as this provides auxiliary information for receiving nodes as to how many ticks the sender claims to have  passed through. I think that this would fit very organically in our design. Instead of the current time, the ticker could send a TimerInfo over the `timeoutChannel`. The `TimerInfo` could include the tick number, which we increment on every tick.

_Originally posted by @AlexHentschel in https://github.com/onflow/flow-go/pull/3357#discussion_r1000015453_
      

We explicitly deal with de-duplication by introducing a sequence number for each timeout object which is tracked by `TimeoutController`


consensus/hotstuff/model/timeout.go
```diff
@@ -12,6 +12,7 @@ import (
 // The end of the time period is the timeout that will trigger pacemaker's view change.
 type TimerInfo struct {
 	View      uint64
+	Tick      uint64
 	StartTime time.Time
 	Duration  time.Duration
```
jordanschalm, line 17:
```suggestion
	// Duration is how long we waited before timing out this round. 
	// It does not include subsequent timeouts (ie. all timeout events emitted for the same
	// view will have the same Duration).
	Duration  time.Duration
```

consensus/hotstuff/model/timeout.go
```diff
@@ -12,6 +12,7 @@ import (
 // The end of the time period is the timeout that will trigger pacemaker's view change.
 type TimerInfo struct {
 	View      uint64
+	Tick      uint64
```
jordanschalm, line 15:
```suggestion
	// Tick is the number of times a timeout event has been emitted for this view (beginning with 0).
	// It is used to de-duplicate TimeoutObject re-broadcasts.
	Tick      uint64
```

model/messages/collection.go
```diff
@@ -41,8 +41,9 @@ type ClusterBlockVote struct {
 }
 
 // ClusterTimeoutObject is part of the collection cluster protocol and represents a collection node
-// timing out in given round.
+// timing out in given round. Contains a sequential number for deduplication purposes.
 type ClusterTimeoutObject struct {
```
jordanschalm, line 45: This and `ClusterBlockVote` are identical to their consensus equivalents. I'm not sure there's a benefit to defining them separately (or even having different types, really). 

We could alias them to avoid duplicate declarations, or just use the consensus message types in cluster consensus as well. 
```suggestion
type ClusterTimeoutObject TimeoutObject
```

engine/collection/message_hub/message_hub.go
```diff
@@ -249,7 +249,7 @@ func (h *MessageHub) processQueuedMessages(ctx context.Context) error {
 // processQueuedTimeout performs actual processing of model.TimeoutObject, as a result of successful invocation
 // broadcasts timeout object to consensus committee.
 // No errors are expected during normal operations.
-func (h *MessageHub) processQueuedTimeout(timeout *model.TimeoutObject) error {
+func (h *MessageHub) processQueuedTimeout(timeout *messages.ClusterTimeoutObject) error {
 	logContext := h.log.With().
```
jordanschalm, line 253: It would be great to include the tick number in the log, so we have visibility into when we're relying on re-broadcasts.

engine/consensus/message_hub/message_hub.go
```diff
@@ -241,7 +241,7 @@ func (h *MessageHub) processQueuedMessages(ctx context.Context) error {
 // processQueuedTimeout performs actual processing of model.TimeoutObject, as a result of successful invocation
 // broadcasts timeout object to consensus committee.
 // No errors are expected during normal operations.
-func (h *MessageHub) processQueuedTimeout(timeout *model.TimeoutObject) error {
+func (h *MessageHub) processQueuedTimeout(timeout *messages.TimeoutObject) error {
 	logContext := h.log.With().
```
jordanschalm, line 245: Same here - would be useful to include the tick number in the log

jordanschalm (commented):
Looks good - this fits in nicely with the re-broadcast mechanism.

jordanschalm (approved):

consensus/hotstuff/consumer.go
```diff
@@ -255,7 +255,7 @@ type CommunicatorConsumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	BroadcastTimeout(timeout *model.TimeoutObject)
+	BroadcastTimeout(timeout *model.TimeoutObject, timeoutTick uint64)
```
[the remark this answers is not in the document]
durkmurder: Do you mean for vote ? For vote we pass values because there is no structure which represents all arguments

model/messages/collection.go
durkmurder: I was always under the impression that we actually want to have different types to be able to distinguish between cluster and main consensus messages. I believe we have talked about this once and decided to have different messages, haven't we?

engine/collection/message_hub/message_hub.go
durkmurder: yup, added it.

consensus/hotstuff/consumer.go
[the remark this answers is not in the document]
jordanschalm: No, I'm comparing this line where we pass the timeout by reference:
```go
BroadcastTimeout(timeout *model.TimeoutObject, timeoutTick uint64)
```
to [this one](https://github.com/onflow/flow-go/pull/3430/files#diff-8a5523bf8e36aec4df64507165416390056719564e7b8a475b3d003bd7f3d672R117) where we pass by value:
```go
OnReachedTimeout(timeout model.TimerInfo)
```

AlexHentschel (commented):
The structure looks great. 

While reviewing, I started thinking about some [edge case around node restarts (see my comment below)](). I am sorry, I didn't think about this earlier. I don't think it changes the high-level approach, but where we place the implementing code.

consensus/hotstuff/eventhandler/event_handler.go
```diff
@@ -196,19 +196,19 @@ func (e *EventHandler) OnReceiveProposal(proposal *model.Proposal) error {
...
 // No errors are expected during normal operation.
-func (e *EventHandler) OnLocalTimeout() error {
+func (e *EventHandler) OnLocalTimeout(timerInfo model.TimerInfo) error {
 	curView := e.paceMaker.CurView()
 	e.log.Debug().Uint64("cur_view", curView).Msg("timeout received from event loop")
 	// TODO(active-pacemaker): update telemetry
 	defer e.notifier.OnEventProcessed()
 
-	err := e.broadcastTimeoutObjectIfAuthorized()
+	err := e.broadcastTimeoutObjectIfAuthorized(timerInfo.Tick)
```
AlexHentschel, line 211: if we followed [my previous comment](https://github.com/onflow/flow-go/pull/3430#discussion_r1016067102), I don't think we would need the `timerInfo.Tick` here anymore.

AlexHentschel:
Not sure if we wanted to keep the field [`TimerInfo.Tick`](https://github.com/onflow/flow-go/blob/e34c3409360f1c68ce32028e51c1ca905c2ae13a/consensus/hotstuff/model/timeout.go#L22-L24) ... I feel it is not really that helpful if we were to follow [my comment above](https://github.com/onflow/flow-go/pull/3430#discussion_r1016067102) 🤷 . Anyway, just in case we wanted to keep the field, it would be great to extend the documentation a bit (and maybe rename it to `TimeoutTick`):
```golang
// TimeoutTick is the number of times the `timeout.Controller` has (re-)emitted the
// timeout for this view. Initially, when the view just starts, `TimeoutTick` is zero
// because the view has not yet timed out. When the timer `Duration` expires, a `TimerInfo`
// with `TimeoutTick = 1` is emitted. Subsequently, `timeout.Controller` repeatedly
// emits `TimerInfo` based on some internal heuristic, where `TimeoutTick` is incremented
// each time.
// CAUTION: TimeoutTick is _not persisted across restarts_! Therefore, it is not suitable
// for reliably facilitating timeout re-broadcast. Instead, after a restart, too many peers
// might consider an intentional timeout rebroadcast as duplicate and drop it.
TimeoutTick uint64
```

consensus/hotstuff/pacemaker/timeout/controller.go
```diff
@@ -61,35 +60,30 @@ func NewController(timeoutConfig Config) *Controller {
...
+func (t *Controller) StartTimeout(ctx context.Context, view uint64) model.TimerInfo {
 	t.stopTicker() // stop old timeout
 
 	// setup new timer:
 	// when round duration is small react with faster timeout rebroadcasts
 	duration := t.ReplicaTimeout()
 	tickInterval := time.Duration(math.Min(float64(duration.Milliseconds()), t.cfg.MaxTimeoutObjectRebroadcastInterval))
-	t.timerInfo = &model.TimerInfo{View: view, StartTime: time.Now().UTC(), Duration: duration}
-	t.timeoutChannel = make(chan time.Time, 1)
+	timerInfo := model.TimerInfo{View: view, StartTime: time.Now().UTC(), Duration: duration}
```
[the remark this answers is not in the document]
durkmurder: @AlexHentschel I've applied suggested changes, please take another look

consensus/hotstuff/pacemaker/timeout/controller.go
```diff
@@ -74,22 +68,22 @@ func (t *Controller) Channel() <-chan time.Time {
...
-func (t *Controller) StartTimeout(ctx context.Context, view uint64) *model.TimerInfo {
+func (t *Controller) StartTimeout(ctx context.Context, view uint64) model.TimerInfo {
 	t.stopTicker() // stop old timeout
 
 	// setup new timer:
 	// when round duration is small react with faster timeout rebroadcasts
 	duration := t.ReplicaTimeout()
-	tickInterval := time.Duration(math.Min(float64(duration.Milliseconds()), t.cfg.MaxTimeoutObjectRebroadcastInterval))
-	t.timerInfo = &model.TimerInfo{View: view, StartTime: time.Now().UTC(), Duration: duration}
+	tickInterval := time.Duration(math.Min(float64(duration.Milliseconds()), t.cfg.MaxTimeoutObjectRebroadcastInterval) * float64(time.Millisecond))
```
AlexHentschel, line 77: To me, calculating the timeout values seems quite clunky due to the repeated conversations 
* from `time.Duration` -> `float64` for the purpose of numerical computations
* `float64` -> Duration for the purpose of _internal_ passing of values

The fact that we forgot the scaling factor `float64(time.Millisecond)` here is an indicator to me that the API is too error prone. 

### Suggestion:
[`ReplicaTimeout()`](https://github.com/onflow/flow-go/blob/e96e75f2aca1837728a2305104e35ed2064268d1/consensus/hotstuff/pacemaker/timeout/controller.go#L122) is only used internally within the `timeout.Controller`.  Therefore, I feel we should tune the return type to improve clarity. My suggestion would be:
*  rename `ReplicaTimeout()` to `replicaTimeout` and return the timeout as `float64` in (units of Milliseconds)
    ```golang
    // ReplicaTimeout returns the duration [in MILLISECONDS] of the current view before we time out in.
    func (t *Controller) replicaTimeout() float64 {
    	if t.r <= t.cfg.HappyPathMaxRoundFailures {
    		return t.cfg.MinReplicaTimeout
    	}
    	r := float64(t.r - t.cfg.HappyPathMaxRoundFailures)
    	if r >= t.maxExponent {
    		return t.cfg.MaxReplicaTimeout
    	}
    	// compute timeout duration [in milliseconds]:
    	return t.cfg.MinReplicaTimeout * math.Pow(t.cfg.TimeoutAdjustmentFactor, r)
    }
    ```
* only after all timings are computed as `float64` (by convention in milliseconds), I would convert to `time.Duration`:
    ```golang
    // StartTimeout starts the timeout of the specified type and returns the timer info
    func (t *Controller) StartTimeout(ctx context.Context, view uint64) model.TimerInfo {
    	t.stopTicker() // stop old timeout

    	// setup new timer
    	durationMs := t.replicaTimeout()                                                         // duration of current view in units of Milliseconds
    	rebroadcastIntervalMs := math.Min(durationMs, t.cfg.MaxTimeoutObjectRebroadcastInterval) // time between attempted re-broadcast of timeouts if there is no progress
    	t.timeoutChannel = make(chan time.Time, 1)                                               // channel for delivering timeouts

    	// start timeout logic for (re-)broadcasting timeout objects on regular basis as long as we are in the same round.
    	var childContext context.Context
    	childContext, t.stopTicker = context.WithCancel(ctx)
    	duration := time.Duration(durationMs) * time.Millisecond
    	rebroadcastInterval := time.Duration(rebroadcastIntervalMs) * time.Millisecond
    	go tickAfterTimeout(childContext, duration, rebroadcastInterval, t.timeoutChannel)

    	return model.TimerInfo{View: view, StartTime: time.Now().UTC(), Duration: duration}
    }
    ```

consensus/hotstuff/safetyrules/safety_rules.go
```diff
@@ -129,6 +129,12 @@ func (r *SafetyRules) ProduceVote(proposal *model.Proposal, curView uint64) (*mo
 func (r *SafetyRules) ProduceTimeout(curView uint64, newestQC *flow.QuorumCertificate, lastViewTC *flow.TimeoutCertificate) (*model.TimeoutObject, error) {
 	lastTimeout := r.safetyData.LastTimeout
 	if lastTimeout != nil && lastTimeout.View == curView {
+		// increment timeout tick, since we are using a pointer this change will be reflected in `safetyData` as well.
+		lastTimeout.TimeoutTick++
+		err := r.persist.PutSafetyData(r.safetyData)
+		if err != nil {
+			return nil, fmt.Errorf("could not persist safety data: %w", err)
+		}
 		return lastTimeout, nil
```
AlexHentschel, line 138: ## ❓❗❓ 
I am worried about the lack of concurrency control here. We are passing the `TimeoutObject` as a pointer through the consensus node. I don't think we should be modifying these structs for consistency reasons. I would suggest to treat `TimeoutObject` as immutable structs and make a copy here:

```suggestion
		// model.TimeoutObject are conceptually immutable, hence we create a shallow copy here, which allows us to increment TimeoutTick
		var updatedTimeout model.TimeoutObject = *lastTimeout
		updatedTimeout.TimeoutTick += 1

		// persist updated TimeoutObject in `safetyData` and return it
		r.safetyData.LastTimeout = &updatedTimeout
		err := r.persist.PutSafetyData(r.safetyData)
		if err != nil {
			return nil, fmt.Errorf("could not persist safety data: %w", err)
		}
		return &updatedTimeout, nil
```

engine/collection/message_hub/message_hub.go
```diff
@@ -244,27 +247,30 @@ func (h *MessageHub) processQueuedMessages(ctx context.Context) error {
...
-		NewestQC:   timeout.NewestQC,
-		LastViewTC: timeout.LastViewTC,
-		SigData:    timeout.SigData,
-	}
 
-	err = h.con.Publish(msg, recipients.NodeIDs()...)
+	err = h.con.Publish(timeout, recipients.NodeIDs()...)
+	if errors.Is(err, network.EmptyTargetList) {
+		return nil
+	}
```
AlexHentschel, line 273: I think this code duplicates behaviour two lines below: https://github.com/onflow/flow-go/blob/e96e75f2aca1837728a2305104e35ed2064268d1/engine/collection/message_hub/message_hub.go#L275-L278
wondering if I am missing something 🤔 

```suggestion
```

model/messages/collection.go
```diff
@@ -41,10 +41,11 @@ type ClusterBlockVote struct {
...
```
AlexHentschel, line 45: I like [Jordan's suggestion](https://github.com/onflow/flow-go/pull/3430/files#r1001453978) to define the cluster types through the consensus types:
```golang
// ...
type ClusterTimeoutObject TimeoutObject

// ...
type ClusterBlockVote BlockVote
```

AlexHentschel (approved):
very nice and clean work. Largely stylistic comments. Only comment with algorithmic relevance is [this one](url)

durkmurder:
bors merge

PR #3435, opened 2022-10-20
Add Pause/Resume functionality to TransactionState

pattyshack:
These will be used for implementing continuation passing style executors (The nested transaction will pause after preprocessing, and resume on actual execution).

Also made a bunch of improvements to support caching result (These should be useful for implementing meter weight prefetching):
 1. added guard to prevent modification on committed state
 2. Commit returns the committed state
 3. renamed AttachAndCommitParseRestricted to AttachAndCommit

### DIFF
--- a/fvm/state/transaction_state.go
+++ b/fvm/state/transaction_state.go
@@ -17,8 +17,6 @@ type nestedTransactionStackFrame struct {
 	// the parts of the runtime environment necessary for importing/parsing the
 	// program, specifically, environment.ContractReader and
 	// environment.Programs.
-	//
-	// TODO(patrick): restrict environment method access
 	parseRestriction *common.AddressLocation
 }
 
@@ -106,14 +104,7 @@ func (s *TransactionState) BeginNestedTransaction() (
 	}
 
 	child := s.currentState().NewChild()
-
-	s.nestedTransactions = append(
-		s.nestedTransactions,
-		nestedTransactionStackFrame{
-			state:            child,
-			parseRestriction: nil,
-		},
-	)
+	s.push(child, nil)
 
 	return NestedTransactionId{
 		state: child,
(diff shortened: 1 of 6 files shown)


fvm/state/transaction_state.go
```diff
@@ -198,28 +223,48 @@ func (s *TransactionState) CommitParseRestricted(
...
-		s.nestedTransactions,
-		nestedTransactionStackFrame{
-			state:            cachedNestedTransaction,
-			parseRestriction: nil,
-		},
-	)
+// Resume attaches the paused nested transaction (state) to the current
+// transaction.
+func (s *TransactionState) Resume(pausedState *State) {
+	s.push(pausedState, nil)
```
Tonix517, line 259: can `pausedState` be committed already? if so, we may want to manually mark `committed` as false so that it can be accessed.

pattyshack: right now, we can resume a committed state (but the only action you can perform in that case is commit).  i'm thinking of adding a wrapper around *State to prevent this case.

Tonix517 (approved):

fvm/state/transaction_state.go
pattyshack: just fyi.  I'll add the PausedNestedTransaction wrapper later today / re-request another review when it's ready.

pattyshack: as discussed on vc, i'll leaves this as is for now and will add the wrapper in a follow up pr.

fvm/state/transaction_state.go
```diff
@@ -198,28 +223,51 @@ func (s *TransactionState) CommitParseRestricted(
...
+}
+
+// Pause detaches the current nested transaction from the parent transaction,
+// and returns the paused nested transaction state.  The paused nested
+// transaction may be resume via Resume.
+//
+// WARNING: Pause and Resume are intended for implementing continuation passing
+// style behavior for the transaction invoker, with the assumption that no
+// other transaction processor modify the state (other than the spock and
+// metering). The paused nested transaction should not be reused across
```
janezpodhostnik, line 236: currently the proposer sequence number is modified during the verify tx processor.

fvm/state/transaction_state.go
```diff
@@ -129,47 +120,76 @@ func (s *TransactionState) BeginParseRestrictedNestedTransaction(
...
 		},
 	)
-
-	return NestedTransactionId{
-		state: child,
-	}, nil
 }
 
-func (s *TransactionState) mergeIntoParent() error {
+func (s *TransactionState) pop(op string) (*State, error) {
```
janezpodhostnik, line 143: Might be nice to introduce an operation enum.

janezpodhostnik (approved):

fvm/state/transaction_state.go
pattyshack: thanks for pointing this out.  i think we're ok in this case since program parsing shouldn't read the sequence number.   I'll update the comment and keep a look out for this.

pattyshack:
bors merge

PR #3437, opened 2022-10-21
Tighten unexpected nested txn check + delay procedure result population

pattyshack:
Since we now early exit fatal errors, we know for sure that unexpected nested transactions are due to fvm programming error (we forgot to commit somewhere along the execution path).


Tonix517 (approved):
nice - much cleaner

janezpodhostnik (approved):

pattyshack (commented):
bors merge

PR #3443, opened 2022-10-24
Switch parseRestricted to use span name instead of hardcoded strings

pattyshack:
This also fixes the span name / tracing in a few places (some were undefined, some were misnamed, and some were using the wrong span names).


SaveTheRbtz (approved):

SaveTheRbtz:
Thanks for fixing/improving tracing.

pattyshack:
bors merge

PR #3445, opened 2022-10-25
[Execution] Randomize collection fetch order

SaveTheRbtz:
Map iteration order is not random, but instead undefined ([as per go spec](https://go.dev/ref/spec#RangeClause)):

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

This [does not mean random and especially not uniformly random](https://go.dev/play/p/wfSzSXfVOJa). Therefore pre-randomize the order.

### DIFF
--- a/engine/common/requester/engine.go
+++ b/engine/common/requester/engine.go
@@ -398,7 +398,7 @@ func (e *Engine) dispatchRequest() (bool, error) {
 
 	// create a batch request, send it and store it for reference
 	req := &messages.EntityRequest{
-		Nonce:     rand.Uint64(),
+		Nonce:     e.rng.Uint64(),
 		EntityIDs: entityIDs,
 	}
 
--- a/engine/common/requester/engine.go
+++ b/engine/common/requester/engine.go
@@ -49,6 +49,7 @@ type Engine struct {
 	items                 map[flow.Identifier]*Item
 	requests              map[uint64]*messages.EntityRequest
 	forcedDispatchOngoing *atomic.Bool // to ensure only trigger dispatching logic once at any time
+	rng                   *rand.Rand
 }
 
 // New creates a new requester engine, operating on the provided network channel, and requesting entities from a node
@@ -114,6 +115,7 @@ func New(log zerolog.Logger, metrics module.EngineMetrics, net network.Network, 
 		items:                 make(map[flow.Identifier]*Item),          // holds all pending items
 		requests:              make(map[uint64]*messages.EntityRequest), // holds all sent requests
 		forcedDispatchOngoing: atomic.NewBool(false),
+		rng:                   rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 
 	// register the engine with the network layer and store the conduit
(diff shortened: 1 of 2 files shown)


bluesign:
Why do we need uniform random here?

SaveTheRbtz:
> Why do we need uniform random here?

The main reason for this randomness is to prevent the case scenario when we constantly request the same items over and over again because of the max batch size limit.

bluesign:
> > Why do we need uniform random here?
> 
> The main reason for this randomness is to prevent the case scenario when we constantly request the same items over and over again because of the max batch size limit.

I thought about that but doesn't https://github.com/onflow/flow-go/blob/bfa82951ae46240d12e03e83a60b070d8d94d1eb/engine/common/requester/engine.go#L306 prevent that ?

engine/common/requester/engine.go
```diff
@@ -286,11 +288,19 @@ func (e *Engine) dispatchRequest() (bool, error) {
 		return false, fmt.Errorf("could not get providers: %w", err)
 	}
 
+	// randomize order of items.
```
zhangchiqing, line 291:
```suggestion
	// randomize order of items, so that they can be requested in different order each time
```

zhangchiqing (approved):

engine/common/requester/engine.go
```diff
@@ -286,11 +288,19 @@ func (e *Engine) dispatchRequest() (bool, error) {
...
+	rndItems := make([]flow.Identifier, 0, len(e.items))
+	for k := range e.items {
+		rndItems = append(rndItems, e.items[k].EntityID)
+	}
+	e.rng.Shuffle(len(rndItems), func(i, j int) { rndItems[i], rndItems[j] = rndItems[j], rndItems[i] })
```
peterargue, line 296: do you have a sense of how large `len(rndItems)` gets in when a node is behind? I think in the happy case this should fine, but I'm worried about the unhappy path.

SaveTheRbtz:
> > > Why do we need uniform random here?
> > 
> > 
> > The main reason for this randomness is to prevent the case scenario when we constantly request the same items over and over again because of the max batch size limit.
> 
> I thought about that but doesn't
> 
> https://github.com/onflow/flow-go/blob/bfa82951ae46240d12e03e83a60b070d8d94d1eb/engine/common/requester/engine.go#L306
> 
> prevent that ?

it would help sometimes depending on the phase of moon (timeouts, error rates, frequency of calling this function, etc.). It is better to explicitly randomize the order though.

> Sorry maybe asking too much (trying to grasp flow-go lately) but also why aborting here?
> 
> https://github.com/onflow/flow-go/blob/bfa82951ae46240d12e03e83a60b070d8d94d1eb/engine/common/requester/engine.go#L339
> 
> Instead of moving to the next item ?

It's a good question. The answer to it is: it was simpler: the complexity of this function is already too high.  In reality we just need to refactor that piece of code to reduce complexity.

peterargue (approved):

SaveTheRbtz:
bors merge

PR #3454, opened 2022-10-26
[FVM] Adding state-replay cache implementation

Tonix517:
As the first step for #3102 , now adding a generic state-replay cache to be used for meter settings, protocol version number later.

Assumption: value retrieval can be wrapped up in a closure. Existing Cadence Get\SetProgram() call pattern is not wrap-able since code is in Cadence side, so for now `BlockProgram` is not able to utilize this cache, until Cadence switches to "GetOrRetrieve" call pattern.

**Test**
- [X] new UT cases

### DIFF
--- a/fvm/state/state_replay_cache.go
+++ b/fvm/state/state_replay_cache.go
@@ -1,0 +1,95 @@
+package state
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// StateReplayCache caches value read from state together with all register touches triggered
+// when reading that value from state. These register touches will be re/played on the target
+// view in every GetOrRetrieveValue value call.
+type StateReplayCache[T any] struct {
+	cachedValue            T
+	cachedState            *State
+	cachedUpdatedRegisters map[string]bool
+}
+
+func NewStateReplayCache[T any]() *StateReplayCache[T] {
+	return &StateReplayCache[T]{
+		cachedUpdatedRegisters: make(map[string]bool),
+	}
+}
+
+// GetOrRetrieveValue returns a state value. When value is not cached, `retriever` closure will be
+// called to retrieve the value and the state delta will be captured and cached too; when value is
+// cached the cached one will be returned, and the cached state delta will be replayed on `view`.
+func (stateReplayCache *StateReplayCache[T]) GetOrRetrieveValue(
+	view View,
+	retriever func(*TransactionState) (T, error),
+	stateParams StateParameters,
+) (T, error) {
+	if view == nil {
+		return stateReplayCache.cachedValue, fmt.Errorf("view parameter should not be nil")
+	}
+	if retriever == nil {
+		return stateReplayCache.cachedValue, fmt.Errorf("retriever closure should not be nil")
+	}
+
+	transactionState := NewTransactionState(view, stateParams)
+
+	if stateReplayCache.cachedState == nil {
+		nestedTxId, err := transactionState.BeginNestedTransaction()
+		if err != nil {
+			return stateReplayCache.cachedValue, fmt.Errorf("failed to start a nested tx: %w", err)
+		}
+
+		// retrieve state value by calling retriever closure
+		stateReplayCache.cachedValue, err = retriever(transactionState)
+		if err != nil {
+			return stateReplayCache.cachedValue, fmt.Errorf("failed to retrieve value: %w", err)
+		}
+
+		// apply reg touches on the target view and cache the delta state for replay purposes
+		stateReplayCache.cachedState, err = transactionState.Commit(nestedTxId)
+		if err != nil {
+			stateReplayCache.cachedState = nil
+			return stateReplayCache.cachedValue, fmt.Errorf("failed to commit state: %w", err)
+		}
+
+		stateReplayCache.refreshUpdatedRegisters()
+	} else {
+		// if value/state is cached already, simply replay it on the target view.
+		err := transactionState.AttachAndCommit(stateReplayCache.cachedState)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


fvm/state/state_replay_cache.go
```diff
@@ -0,0 +1,87 @@
...
+type StateReplayCache[T any] struct {
+	cachedValue T
+	cachedState *State
+}
+
+func NewStateReplayCache[T any]() *StateReplayCache[T] {
+	return &StateReplayCache[T]{}
+}
+
+// GetOrRetrieveValue returns a state value. When value is not cached, `retriver` closure will be
```
pattyshack, line 21: typo s/retriver/retriever

fvm/state/state_replay_cache.go
```diff
@@ -0,0 +1,87 @@
...
+// called to retrieve the value and the state delta will be captured and cached too; when value is
+// cached the cached one will be returned, and the cached state delta will be replayed on `view`.
+func (stateReplayCache *StateReplayCache[T]) GetOrRetrieveValue(
+	view View,
```
pattyshack, line 25: maybe pass in the transactionState rather than the view.  this enables us to compose nested transactions.

fvm/state/state_replay_cache.go
```diff
@@ -0,0 +1,87 @@
...
+	transactionState := NewTransactionState(view, stateParams)
+
+	if stateReplayCache.cachedState == nil {
+		nestedTxId, err := transactionState.BeginNestedTransaction()
+		if err != nil {
+			return stateReplayCache.cachedValue, fmt.Errorf("failed to start a nested tx: %w", err)
+		}
+
+		// retrieve state value by calling retriever closure
+		stateReplayCache.cachedValue, err = retriever(transactionState)
```
pattyshack, line 45: should we cache the error in stateReplayCache as well?

fvm/state/state_replay_cache.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return stateReplayCache.cachedValue, nil
+}
+
+// Invalidate clears the cached state so next GetOrRetrieveValue() call will do
+// actual state read again to retrieve latest value in state.
+func (stateReplayCache *StateReplayCache[T]) Invalidate(
+	updatedRegiestersInTx []flow.RegisterID,
+) bool {
+	capturedUpdatedRegisters := make(map[string]bool)
```
pattyshack, line 72: if we plan to invalidate frequently, than maybe pre-compute this map as part of retriever.

fvm/state/state_replay_cache.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, reg := range stateReplayCache.cachedState.UpdatedRegisters() {
+		capturedUpdatedRegisters[reg.String()] = true
+	}
+
+	// detect register collision
+	for _, reg := range updatedRegiestersInTx {
```
pattyshack, line 78: this could be expensive for large transactions.  the number of captured register is typically small in comparison to the number of registers accessed by a transaction.  it's probably better to iterate over captured registers and probe the updated registers rather than the other way around.

Tonix517: another option is to simply make `Invalidate()` parameter-less and simply clear the cache. we leave it to callers to decide when the cache should be invalidated. in that case caller should figure out an optimal way, like for meter settings, caller can simply maintain a flag and mark it as true when meter setting values are mutated.

fvm/state/state_replay_cache.go
Tonix517: i was thinking about that, but in the case of meter settings, in `fvm.Run` when a transactionState is created, meter settings is needed: https://github.com/onflow/flow-go/blob/master/fvm/fvm.go#L122, so if we pass in transactionState, we will need to create a separate transactionState and then manually Pause\Assume that state. And since state here is mostly for a View, passing a View can make that use cases slightly more elegant.

fvm/state/state_replay_cache.go
Tonix517: here i return any error to caller to let them handle it immediately, so that in this cache we can assume that the cached state is not corrupted. it is mostly the question when\how we should handle errors from this cache. if we need to address a previously occurred error we need to cache that error as well.

fvm/state/state_replay_cache.go
pattyshack: i'm fine with either approach.  but regardless of where this check is done (as part of this function, or as an external thing), we should iterate over the smaller set and probe the larger set in general.

fvm/state/state_replay_cache.go
Tonix517: addressed

fvm/state/state_replay_cache.go
Tonix517: addressed.

fvm/state/state_replay_cache.go
pattyshack: o, i take my comment back.  in general, we need to pass in the transaction state rather than the view because we need update the transaction state's meter / spock hash, which are not part of the view

pattyshack (commented):
btw, this implementation only works for sequential execution / doesn't work for speculative execution (e.g., speculative parsing/preprocessing).  either address this or add a todo for me to address this.

pattyshack (commented):
ooo, i just realize that there's another bug in the programs cache.  we should invalidate all of its program entries when meter weights are updated (we don't right now ...).

since we would need to invalidate those entries anyways, it may be better to keep the weights cached as part of the programs cache

Tonix517:
discussed offline - we decided to simply move meter settings into existing programCache to utilize its invalidation logic, instead of creating a separate cache entity.

PR #3461, opened 2022-10-27
Process service events over ordered seals

jordanschalm:
Based on https://github.com/onflow/flow-go/pull/2969#discussion_r1003733713
* Adds a helper function `OrderedSeals` which orders the seals in a block payload by block height
* Uses helper in service event processing to avoid edge case noted in the comment
* Uses helper in `lastSealed` function


model/flow/payload.go
```diff
@@ -6,11 +6,16 @@ import (
...
+	// only the system chunk is executed for this block.
 	Guarantees []*CollectionGuarantee
-	Seals      []*Seal
-	Receipts   ExecutionReceiptMetaList
-	Results    ExecutionResultList
+	// Seals contains the block seals, for ancestor blocks, included in this payload.
+	// The oldest seal must connect to the latest seal in the fork extended by this block.
+	// Seals must be internally connected, containing no seals with duplicate block IDs or heights.
+	// Seals may be empty.
+	Seals    []*Seal
```
durkmurder, line 16: Let's mention that seals are not ordered?

durkmurder (approved):
Nice and clean, good job!

model/flow/payload.go
jordanschalm: Of course, the entire reason I was updating the godoc 🤦. Added in https://github.com/onflow/flow-go/pull/3461/commits/61fbabc1c79d49ab7e530f656e005547a1779696

AlexHentschel (approved):
very nice.

jordanschalm:
bors merge

state/protocol/util.go
```diff
@@ -113,3 +115,36 @@ func FindGuarantors(state State, guarantee *flow.CollectionGuarantee) ([]flow.Id
...
+	// since seals in a valid payload must have consecutive heights we can populate
+	// the ordered output by shifting by minHeight
+	seals := make([]*flow.Seal, len(payload.Seals))
+	for i, seal := range payload.Seals {
+		height := heights[i]
+		seals[height-minHeight] = seal
+	}
+
+	return seals, nil
+}
```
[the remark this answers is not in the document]
jordanschalm: Added this suggestion in https://github.com/onflow/flow-go/pull/3461/commits/589ddbce92bb0fbe85d3cfa03a641632fdce7b9a

> We could even [add a nil-pointer check here](https://github.com/onflow/flow-go/blob/61fbabc1c79d49ab7e530f656e005547a1779696/state/protocol/util.go#L131), and as result could guarantee that this function never panics for any input (?)

We would still need to contend with nil values in the inputs `payload`, `payload.Seals[x]`, and `headers`. I feel this is crossing the threshold where defensive nil checks are overly burdensome and unrealistic to depend on.

PR #3468, opened 2022-10-28
[BFT] Add validation support to message handler

peterargue:
This PR adds a validation callback to the `engine.MessageHandler` that is called for each matched message passed into the engine. It also implements the validator for the common sync engine and request handler engines

### DIFF
--- a/engine/enqueue.go
+++ b/engine/enqueue.go
@@ -26,15 +26,17 @@ type Pattern struct {
 	Match MatchFunc
 	// Map is a function to apply to messages before storing them. If not provided, then the message is stored in its original form.
 	Map MapFunc
+	// Validate is a function to validate a message before storing it. If not provided, then the message is always stored.
+	Validate ValidateFunc
 	// Store is an abstract message store where we will store the message upon receipt.
 	Store MessageStore
 }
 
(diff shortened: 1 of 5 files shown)


engine/enqueue.go
```diff
@@ -26,16 +26,18 @@ type Pattern struct {
 	Match MatchFunc
 	// Map is a function to apply to messages before storing them. If not provided, then the message is stored in its original form.
 	Map MapFunc
+	// Validate is a function to validate a message before storing it. If not provided, then the message is always stored.
+	Validate ValidateFunc
```
jordanschalm, line 30: We're solving the same problem slightly different ways. Here's a PR on the Pacemaker branch addressing a similar problem: https://github.com/onflow/flow-go/pull/3470

I'll review the rest of this PR in isolation, but we should align on one of these approaches.

jordanschalm, line 30: I would suggest to document the kind of validation we expect to happen here, and the expected error returns. In particular:
* should only include cheap validation checks (else we're blocking networking layer threads again)
* if a message passes validation here, it may still be invalid (validation is not exhaustive)
* either define an expected sentinel error type, or "all generic errors are benign and indicate an invalid message"

PR #3469, opened 2022-10-28
[BFT] Add provider engine message validation

peterargue:
This PR adds a `MessageHandler` validator to the common provider engine


peterargue:
we're taking a different approach to message validation

Issue #3374, opened 2022-10-12
Suggested compliance revisions


AlexHentschel:
This is a list of suggested revisions and improvements that resulted from reviewing [PR #3294](https://github.com/onflow/flow-go/pull/3294).

* [x] https://github.com/onflow/flow-go/pull/3470 [**high prio**: safety critical]The compliance engines (consensus and collector) are still vulnerable to various malformed message attacks, for example:
    * [this code](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/consensus/compliance/engine.go#L135-L142) is vulnerable to a nil-pointer attack: `Block` could be nil; we rely on synchronization engine to check this, which it doesn't. Furthermore, `SyncedBlock` does not document to which extend it is trusted. Caution, this problem is amplified by current changes on `master`, where we [submit an entire _list_ of blocks to the compliance engine](https://github.com/onflow/flow-go/blob/3e6d9954752e86445cba854b67ccb6c2dd4ae73f/engine/common/synchronization/engine.go#L311)
   * [this code](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/consensus/compliance/core.go#L118-L119) is vulnerable to a nil-pointer attack: `Header` could be nil

   Ideally, the compliance engine should not rely on any other components to assert validity to increase modularity and robustness of the code base.
  My suggestion would be to include methods for `BlockProposal` that check _basic syntactic_ validity.  
* [x] https://github.com/onflow/flow-go/pull/3470 Similarly, I think there are a variety of nil-pointer checks missing in the [`badger.MutableState`](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/state/protocol/badger/mutator.go#L135-L171)
* [x] in the collector `compliance.Core`, method `OnBlockProposal`, we query the finalized block directly from the data base ([code](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/engine/collection/compliance/core.go#L130-L142))
   * this data base interaction is relatively expensive
[...]

PR #3470, opened 2022-10-28
Enforce structural validity for messages in decoder

jordanschalm:
Based on https://github.com/onflow/flow-go/issues/3374 (1, 2 - nil message field checking).

* Adds interface for basic message structural validity: `model.StructureValidator`
* Modifies `network.Codec` to enforce structural validity for messages implementing this interface during decoding. This way we can assume any message which passes the network layer (once it arrives at `MessageProcessor.Process`) passes basic structural/syntactic validation
* Implements `StructureValidator` for consensus-related messages 
* Adds structural validity checks to protocol state and cluster state

### DIFF
--- a/model/events/synchronization.go
+++ b/model/events/synchronization.go
@@ -19,13 +19,7 @@ var _ model.StructureValidator = (*SyncedBlock)(nil)
 
 // StructureValid checks basic structural validity of the message and ensures no required fields are nil.
 func (m *SyncedBlock) StructureValid() error {
-	if m.Block.Header == nil {
-		return model.NewStructureInvalidError("nil block header")
-	}
-	if m.Block.Payload == nil {
-		return model.NewStructureInvalidError("nil block payload")
-	}
-	return nil
+	return m.Block.StructureValid()
 }
 
 // SyncedClusterBlock is an internal message passed from the chainsync engine to the consensus message hub.
@@ -41,11 +35,5 @@ var _ model.StructureValidator = (*SyncedClusterBlock)(nil)
 
 // StructureValid checks basic structural validity of the message and ensures no required fields are nil.
 func (m *SyncedClusterBlock) StructureValid() error {
-	if m.Block.Header == nil {
-		return model.NewStructureInvalidError("nil block header")
-	}
-	if m.Block.Payload == nil {
-		return model.NewStructureInvalidError("nil block payload")
-	}
-	return nil
+	return m.Block.StructureValid()
 }
--- a/model/events/synchronization.go
+++ b/model/events/synchronization.go
@@ -15,6 +15,8 @@ type SyncedBlock struct {
 	Block    *flow.Block
 }
 
+var _ model.StructureValidator = (*SyncedBlock)(nil)
+
 // StructureValid checks basic structural validity of the message and ensures no required fields are nil.
 func (m *SyncedBlock) StructureValid() error {
 	if m.Block.Header == nil {
@@ -34,6 +36,8 @@ type SyncedClusterBlock struct {
 	OriginID flow.Identifier
 	Block    *cluster.Block
 }
+
+var _ model.StructureValidator = (*SyncedClusterBlock)(nil)
 
 // StructureValid checks basic structural validity of the message and ensures no required fields are nil.
 func (m *SyncedClusterBlock) StructureValid() error {
--- a/model/events/synchronization.go
+++ b/model/events/synchronization.go
@@ -1,16 +1,47 @@
 package events
 
 import (
+	"github.com/onflow/flow-go/model"
 	"github.com/onflow/flow-go/model/cluster"
 	"github.com/onflow/flow-go/model/flow"
 )
 
+// SyncedBlock is an internal message passed from the chainsync engine to the consensus message hub.
+// It contains a block obtained through the chainsync protocol.
+// These blocks have not been validated and should be considered untrusted inputs,
+// however they are guaranteed to be structurally valid: StructureValid()==nil
 type SyncedBlock struct {
 	OriginID flow.Identifier
 	Block    *flow.Block
 }
 
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *SyncedBlock) StructureValid() error {
+	if m.Block.Header == nil {
+		return model.NewStructureInvalidError("nil block header")
+	}
+	if m.Block.Payload == nil {
+		return model.NewStructureInvalidError("nil block payload")
+	}
+	return nil
+}
+
+// SyncedClusterBlock is an internal message passed from the chainsync engine to the consensus message hub.
+// It contains a block obtained through the chainsync protocol.
+// These blocks have not been validated and should be considered untrusted inputs,
+// however they are guaranteed to be structurally valid: StructureValid()==nil
 type SyncedClusterBlock struct {
 	OriginID flow.Identifier
 	Block    *cluster.Block
+}
+
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *SyncedClusterBlock) StructureValid() error {
+	if m.Block.Header == nil {
+		return model.NewStructureInvalidError("nil block header")
+	}
+	if m.Block.Payload == nil {
+		return model.NewStructureInvalidError("nil block payload")
+	}
+	return nil
 }
--- a/model/flow/header.go
+++ b/model/flow/header.go
@@ -7,6 +7,7 @@ import (
 	"github.com/fxamacker/cbor/v2"
 	"github.com/vmihailenco/msgpack/v4"
 
+	"github.com/onflow/flow-go/model"
 	cborcodec "github.com/onflow/flow-go/model/encoding/cbor"
 	"github.com/onflow/flow-go/model/fingerprint"
 )
@@ -74,6 +75,13 @@ func (h Header) Body() interface{} {
 		ProposerID:         h.ProposerID,
 		LastViewTCID:       h.LastViewTC.ID(),
 	}
+}
+
+func (h *Header) StructureValid() error {
+	if h == nil {
+		return model.NewStructureInvalidError("nil header")
+	}
+	return nil
 }
 
 func (h Header) Fingerprint() []byte {
--- a/model/flow/payload.go
+++ b/model/flow/payload.go
@@ -2,6 +2,8 @@ package flow
 
 import (
 	"encoding/json"
+
+	"github.com/onflow/flow-go/model"
 )
 
 // Payload is the actual content of each block.
@@ -38,6 +40,33 @@ func (p Payload) MarshalJSON() ([]byte, error) {
 	return json.Marshal(struct{ payloadAlias }{
 		payloadAlias: payloadAlias(p),
 	})
+}
+
+func (p *Payload) StructureValid() error {
+	if p == nil {
+		return model.NewStructureInvalidError("nil payload")
+	}
+	for _, guarantee := range p.Guarantees {
+		if guarantee == nil {
+			return model.NewStructureInvalidError("contains nil guarantee")
+		}
+	}
+	for _, seal := range p.Seals {
+		if seal == nil {
+			return model.NewStructureInvalidError("contains nil seal")
+		}
+	}
+	for _, receipt := range p.Receipts {
+		if receipt == nil {
+			return model.NewStructureInvalidError("contains nil receipt")
+		}
+	}
+	for _, result := range p.Results {
+		if result == nil {
+			return model.NewStructureInvalidError("contains nil result")
+		}
+	}
+	return nil
 }
 
 // Hash returns the root hash of the payload.
--- a/model/messages/consensus.go
+++ b/model/messages/consensus.go
@@ -1,14 +1,26 @@
 package messages
 
 import (
+	"github.com/onflow/flow-go/model"
 	"github.com/onflow/flow-go/model/flow"
 )
 
-// BlockProposal is part of the consensus protocol and represents the the leader
+// BlockProposal is part of the consensus protocol and represents the leader
 // of a consensus round pushing a new proposal to the network.
 type BlockProposal struct {
 	Header  *flow.Header
 	Payload *flow.Payload
+}
+
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *BlockProposal) StructureValid() error {
+	if m.Header == nil {
+		return model.NewStructureInvalidError("nil block header")
+	}
+	if m.Payload == nil {
+		return model.NewStructureInvalidError("nil block payload")
+	}
+	return nil
 }
 
 // BlockVote is part of the consensus protocol and represents a consensus node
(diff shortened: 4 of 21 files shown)


model/messages/synchronization.go
```diff
@@ -47,9 +48,44 @@ type BlockResponse struct {
 	Blocks []*flow.Block
 }
 
-// ClusterBlockResponse is the same thing as BlockResponse, but for cluster
-// consensus.
+var _ model.StructureValidator = (*BlockResponse)(nil)
+
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *BlockResponse) StructureValid() error {
+	for _, b := range m.Blocks {
```
durkmurder, line 55: Should we allow an empty range? I think we don't want to process empty range because it doesn't make sense

model/messages/consensus.go
```diff
@@ -1,16 +1,30 @@
...
 	Header  *flow.Header
 	Payload *flow.Payload
 }
 
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *BlockProposal) StructureValid() error {
+	if m.Header == nil {
+		return model.NewStructureInvalidError("nil block header")
+	}
+	if m.Payload == nil {
```
durkmurder, line 20: Payload has lists of pointers, it's possible that adversary constructs a list of nil pointers, probably that will get parsed correctly but most likely will crash the node with segmentation fault

model/messages/synchronization.go
durkmurder: Can we somehow compose those checks so we use `b.StructureValid()` ?
I feel like we would benefit a lot from such recursive approach

network/codec/cbor/codec.go
```diff
@@ -107,5 +107,14 @@ func (c *Codec) Decode(data []byte) (interface{}, error) {
 		return nil, codec.NewMsgUnmarshalErr(data[0], what, err)
 	}
 
+	if validatable, ok := msgInterface.(model.StructureValidator); ok {
```
durkmurder, line 110: I would say that eventually we would like every message to be validated

model/messages/consensus.go
jordanschalm: Added in https://github.com/onflow/flow-go/pull/3470/commits/8b47a64b24ac42d4f3df2b7c2a4ed1f6cacec715

model/messages/synchronization.go
jordanschalm: Good idea, added in https://github.com/onflow/flow-go/pull/3470/commits/8b47a64b24ac42d4f3df2b7c2a4ed1f6cacec715 and https://github.com/onflow/flow-go/pull/3470/commits/fc1605225f92c920ce630bcaa37eb4050d8fe957

model/events/synchronization.go
```diff
@@ -1,16 +1,39 @@
...
+// however they are guaranteed to be structurally valid: StructureValid()==nil
 type SyncedBlock struct {
 	OriginID flow.Identifier
 	Block    *flow.Block
 }
 
+var _ model.StructureValidator = (*SyncedBlock)(nil)
+
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *SyncedBlock) StructureValid() error {
```
AlexHentschel, line 21: when I read 
> checks basic structural validity of the message
I feel it is a reasonable assumption to think "I can assume that [`ProposerSigData`](https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/model/flow/header.go#L45) has the appropriate length, otherwise the block is clearly structurally invalid". I think documentation really needs to be really explicitly here about what is checked in detail.

model/events/synchronization.go
```diff
@@ -1,16 +1,39 @@
...
+// however they are guaranteed to be structurally valid: StructureValid()==nil
 type SyncedClusterBlock struct {
 	OriginID flow.Identifier
 	Block    *cluster.Block
 }
+
+var _ model.StructureValidator = (*SyncedClusterBlock)(nil)
+
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *SyncedClusterBlock) StructureValid() error {
```
peterargue, line 37: should this also check if `m` is nil?

model/events/synchronization.go
peterargue, line 21: should this also check if `m` is nil?

model/flow/header.go
```diff
@@ -76,6 +77,13 @@ func (h Header) Body() interface{} {
 	}
 }
 
+func (h *Header) StructureValid() error {
```
AlexHentschel, line 80: I think there are a lot of thinks that we could check here to cover structural validity of the header:
* `ParentVoterIndices`, `ParentVoterSigData`, `ProposerSigData` should all not be empty or nil
* `ParentID` and `ProposerID` should not be `flow.ZeroID`

We could check them here. Not sure if we want to from a design perspective. Also, there are exceptions for the genesis block 🤷. I think that would be worth-while documenting here.

model/flow/payload.go
```diff
@@ -40,6 +42,33 @@ func (p Payload) MarshalJSON() ([]byte, error) {
...
+		if seal == nil {
+			return model.NewStructureInvalidError("contains nil seal")
+		}
+	}
+	for _, receipt := range p.Receipts {
+		if receipt == nil {
+			return model.NewStructureInvalidError("contains nil receipt")
+		}
+	}
+	for _, result := range p.Results {
```
AlexHentschel, line 64: what is with the (`ChunkList`)[https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/model/flow/chunk.go#L59] in the (`ExecutionResult`)[https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/model/flow/execution_result.go#L15]. Elements (i.e. individual chunks) could be nil. We don't check this here. Should we?

model/validity.go
```diff
@@ -0,0 +1,45 @@
...
+// StructureValidator is a model type which exposes a method to validate basic structural
+// validity of a message or model. This DOES NOT enforce business logic rules,
+// it only enforces that required fields are non-nil and no fields are malformed.
+// If a message fails structure validation, the sender can be slashed.
+type StructureValidator interface {
+	// StructureValid validates the model's basic structural validity: all required fields
+	// are non-nil and well-formed. It does NOT enforce business logic rules.
+	// Implementations must be non-blocking and computationally inexpensive, because
+	// structural validity is enforced early in the message processing flow.
+	// Returns StructureInvalidError if the model is structurally invalid.
```
AlexHentschel, line 43:
```suggestion
	// Returns StructureInvalidError if the model is structurally invalid.
	// No other error returns are expected during normal operations. 
```

network/codec/cbor/codec.go
```diff
@@ -85,7 +85,7 @@ func (c *Codec) Encode(v interface{}) ([]byte, error) {
 // NOTE: 'envelope' contains 'code' & serialized / encoded 'v'.
 // i.e.  1st byte is 'code' and remaining bytes are CBOR encoded 'v'.
 // Expected error returns during normal operations:
-//   - codec.UnknownMsgCodeErr if message code byte does not match any of the configured message codes.
+//   - codec.ErrMsgUnmarshal if message code byte does not match any of the configured message codes.
 //   - codec.ErrMsgUnmarshal if the codec fails to unmarshal the data to the message type denoted by the message code.
 func (c *Codec) Decode(data []byte) (interface{}, error) {
```
AlexHentschel, line 90: I think the binstat framework is not used anymore. We started removing code like the following from the implementations: 
* https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/network/codec/cbor/codec.go#L91-L96
* https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/network/codec/cbor/codec.go#L103
* https://github.com/onflow/flow-go/blob/04b4bcaf206cac75e956a557a959668b100d9356/network/codec/cbor/codec.go#L105
would you mind taking it out here as well?

peterargue (commented):
Added a few comments. It would be great to get this onto master so we could include it with the other BFT updates.

network/codec/cbor/decoder.go
```diff
@@ -17,8 +18,10 @@ type Decoder struct {
 }
 
 // Decode will decode the next CBOR value from the stream.
+// If the CBOR value decodes to a Go type which implements model.StructureValidator
+// this function will validate the Go type's structure using this method.
 // Expected error returns during normal operations:
-//   - codec.UnknownMsgCodeErr if message code byte does not match any of the configured message codes.
+//   - codec.ErrUnknownMsgCode if message code byte does not match any of the configured message codes.
 //   - codec.ErrMsgUnmarshal if the codec fails to unmarshal the data to the message type denoted by the message code.
 func (d *Decoder) Decode() (interface{}, error) {
```
AlexHentschel, line 26: similar request to remove the binstat code

state/protocol/badger/mutator.go
```diff
@@ -104,8 +104,15 @@ func NewFullConsensusState(
 // Instead, the consensus follower relies on the consensus participants to
 // validate the full payload. Therefore, a follower a QC (i.e. a child block) as
 // proof that a block is valid.
+// Expected errors during normal operations:
+//   - state.OutdatedExtensionError if the candidate block is outdated (e.g. orphaned)
+//   - state.InvalidExtensionError if the candidate block is invalid
 func (m *FollowerState) Extend(ctx context.Context, candidate *flow.Block) error {
```
AlexHentschel, line 111:
```suggestion
```

state/protocol/badger/mutator.go
```diff
@@ -132,8 +139,15 @@ func (m *FollowerState) Extend(ctx context.Context, candidate *flow.Block) error
 
 // Extend extends the protocol state of a CONSENSUS PARTICIPANT. It checks
 // the validity of the _entire block_ (header and full payload).
+// Expected errors during normal operations:
+//   - state.OutdatedExtensionError if the candidate block is outdated (e.g. orphaned)
+//   - state.InvalidExtensionError if the candidate block is invalid
 func (m *MutableState) Extend(ctx context.Context, candidate *flow.Block) error {
```
AlexHentschel, line 146:
```suggestion
```

state/protocol/badger/mutator.go
```diff
@@ -104,8 +104,15 @@ func NewFullConsensusState(
 // Instead, the consensus follower relies on the consensus participants to
 // validate the full payload. Therefore, a follower a QC (i.e. a child block) as
 // proof that a block is valid.
+// Expected errors during normal operations:
+//   - state.OutdatedExtensionError if the candidate block is outdated (e.g. orphaned)
+//   - state.InvalidExtensionError if the candidate block is invalid
 func (m *FollowerState) Extend(ctx context.Context, candidate *flow.Block) error {
 
+	if err := candidate.StructureValid(); err != nil {
```
AlexHentschel, line 112: ## ❓ 

I am confused why we are calling `StructureValid` here again. Didn't the networking layer already check this. 
I would really prefer a design, where is very clear where certain checks happen. Otherwise, I think we amplify the risk of checking some things 10 times and still miss some other necessary checks.

AlexHentschel (commented):
I like your approach to put some basic checks into the networking layer. Though, I have some concerns:
* The design is not quite clear to me.  In particular, I don't think there is really a clear convention of what is checked where. On the one hand, moving to a less ambiguous design is probably beyond the scope of this PR. On the other had, I think we should be very cautious to add ambiguity, which I feel the code in its current state would:
  * I think a much more detailed documentation of what is checked for each implementation of `StructureValid` is highly desired (see [my comment here](url) for details)
  * `StructureValid` has an error return. Therefore, I would prefer if each implementation included a clear statement what errors are expected (just for completeness). 
  
    Something like the following would be great:
      ```golang
      // StructureValid enforces that the block adheres to the following syntactical conventions:
      //  * ... [exhaustive list of all checks]
      // Otherwise, a `model.StructureInvalidError` is returned. No other error returns are expected
      // during normal operations. 
      ```
   * I am worried about repeated checks, which for me is an indicator of too much ambiguity in the design. (see [comment here](url))

model/messages/consensus.go
```diff
@@ -1,16 +1,30 @@
...
-// BlockProposal is part of the consensus protocol and represents the the leader
+// BlockProposal is part of the consensus protocol and represents the leader
 // of a consensus round pushing a new proposal to the network.
 type BlockProposal struct {
 	Header  *flow.Header
 	Payload *flow.Payload
 }
 
+// StructureValid checks basic structural validity of the message and ensures no required fields are nil.
+func (m *BlockProposal) StructureValid() error {
```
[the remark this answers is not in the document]
AlexHentschel: +1

jordanschalm:
> It would be great to get this onto master so we could include it with the other BFT updates.

@peterargue We are planning to deploy Active Pacemaker as part of the next spork, in January. Would the other BFT updates you mention be deployed before that? If not, I feel this could wait to be ported to master until the Pacemaker branch as a whole is.

model/messages/consensus.go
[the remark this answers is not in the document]
jordanschalm: I didn't add nil checks for top-level messages because the codec will never decode to a nil value (tested [here](https://github.com/onflow/flow-go/pull/3470/files#diff-e829a255cdabbac01eaed3c7e5b334703ad1e1e9972875073c44f3aa7e8b877aR16-R26)).

model/validity.go
```diff
@@ -0,0 +1,45 @@
...
+func IsStructureInvalidError(err error) bool {
+	var structureInvalidErr StructureInvalidError
+	return errors.As(err, &structureInvalidErr)
+}
+
+// StructureValidator is a model type which exposes a method to validate basic structural
+// validity of a message or model. This DOES NOT enforce business logic rules,
+// it only enforces that required fields are non-nil and no fields are malformed.
+// If a message fails structure validation, the sender can be slashed.
+type StructureValidator interface {
```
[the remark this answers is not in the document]
jordanschalm: I left some detailed thoughts in the Notion doc you made: https://www.notion.so/dapperlabs/Message-validation-thoughts-and-proposal-ec9aee8ab6f949b8984cac5a42e235f1#f4fda8dade9444cb81b36919c4715773

Overall, I'm leaning toward more strictly limiting what `StructureValidate` should do to essentially: "make sure that downstream components can access required fields without panicking".

model/flow/header.go
jordanschalm: I omitted these exactly because of the root block exception that you mentioned.

model/flow/payload.go
jordanschalm: 👍 we should

jordanschalm:
After some discussions, going to replace the approach implemented here with external message types that do not have nil-able fields or items to begin with.
* We feel that, were this network decode validation layer to be kept, it would be better to limit it to safety-related validations -> then removing nil-able fields from external messages removes the need for this validation layer at all
* Creating a separate type for external messages creates a pathway toward using distinct types for Validated and Unvalidated messages, and caching computed values on Validated message types

jordanschalm:
closing in favour of https://github.com/onflow/flow-go/pull/3523

Issue #2767, opened 2022-07-08
[FVM] Short Circuit Transactions with insufficient balance


janezpodhostnik:
## Context

Fee deduction currently happens as the last step of the transaction execution. This means that the users transaction is fully executed before failing with insufficient balance on the payers account. This means it is possible to send computationally heavy transactions without having the funds to cover them.

This however also allows the use case where the payer does not have enough funds to pay for the transaction, but withdraws those funds from a different source (withdraw staking rewards) during the transaction, so by the end of the transaction its balance is OK.

## Solution

Main point: **Transactions where the payer has insufficient balance should fail earlier.**

### Define "earlier"

Ideally the payers balance would be checked even before the transaction gets included in a collection, so that it doesn't cost any resources on the execution/verification side. This is the next step and **not in the scope of this issue**.

Even if the transaction's payer's balance is checked before hitting the execution node, the execution node still has to re-check each transaction. 

The current transaction execution steps:
- check all signatures
- check sequence number
- regular path:
  - invoke users transactions
  - deduct transaction fees
  - check storage
- failure path (in case of error during the regular path):
  - deduct transaction fees

The transaction execution steps would Ideally look like this:

- Phase 1.
  - check payers signature
  - check sequence number
- Phase 2.
  - regular path
    - check payers balance
    - check other signatures
    - invoke users transactions
    - deduct transaction fees
    - check storage
  - failure path (in case of error during the regular path):
    - deduct transaction fees

A failure in the Phase 1. would mean we cannot charge the payer at all, because this isn't a validly signed transaction and someone might be impersonating the payer or re-sending old transactions in order to drain the payers balance.

A failure in the regular path of Phase 2. means fees will be deducted from the payer even if the payer is below the minimum account balance. This might lead to the payer eventually dropping to 0 balance if the payer keeps sending transactions.

For this issue just the first building block will be added which is the **check payers balance** step, and the transaction execution steps will look like this:

- check all signatures
- check sequence number
- regular path:
  - check payers balance
  - invoke users transactions
  - deduct transaction fees
  - check storage
- failure path (in case of error during the regular path):
  - deduct transaction fees

### Define "insufficient balance"

Insufficient balance means the payers balance is less then:
- Minimum account balance (MAB)
- PLUS transaction inclusion fees
- PLUS maximum transaction execution fees  (in respect to the max execution effort (gas) defined by the transaction itself)

Alternatives considered:
[...]

PR #3473, opened 2022-10-28
[FVM] Fail fast when payer cannot pay

janezpodhostnik:
This is the first part of onflow/flow-go#2767

Payer must have sufficient balance for the transaction execution to happen.
The required balance is defined by the smart contract.

related to: https://github.com/onflow/flow-emulator/pull/231 and https://github.com/onflow/flow-core-contracts/pull/318


bluesign:
I am late to the party, but why do we need `MaxTxFees` here? Why not MAB is not enough alone? 

Considering when I have <MAB, any transaction will fail, and I will end up with <MAB; isn't it enough check to filter transactions and reject? Why do we need to cast the net wider here?

janezpodhostnik:
@bluesign I went into more detail in the issue itself https://github.com/onflow/flow-go/issues/2767, but generally the Idea is that if the payer is at MAB and the transaction is run and the payers vault doesn't get a deposit during the transaction, the transaction will fail anyway because payers balance will be below MAB.

bluesign:
thanks @janezpodhostnik, but I fail to see the benefit here; probably I am missing something:

Now; in the case of MAB, limit is constant. But when we introduce limit as a function of gas_limit ( MAB + cost(msx_gas_limit) ) we only cover very limited cases ( which are not important ) but generate extra complication.

Considering my account has MAB + delta ( where delta is a small amount ) of balance, It is in a state of unknown, isn't it easier to cache without this delta and calculation?  If we have an account that has < MAB, we can cache the result, and even skip the MAB check ( until balance > MAB ), but when we have MAB + delta calculation, we are losing that possibility.

janezpodhostnik:
That is a good argument! I will bring that up with the team.

This change would go into the smart contract anyway, so it doesn't affect this PR in particular, but the feature in general.

fvm/environment/system_contracts.go
```diff
@@ -92,6 +92,34 @@ func ServiceAddress(chain flow.Chain) flow.Address {
...
+}
+
+// CheckPayerBalanceAndGetMaxTxFees executes the verifyPayersBalanceForTransactionExecution
+// on the FlowFees account.
+func (sys *SystemContracts) CheckPayerBalanceAndGetMaxTxFees(
+	payer flow.Address,
+	inclusionEffort uint64,
+	maxExecutionEffort uint64,
+) (cadence.Value, error) {
+	return sys.Invoke(
```
Tonix517, line 113: do you have profiling info about how much this extra overhead will do? wondering if it can be less than deductFees & checkStorage cadence calls.

Tonix517:
i'm thinking out loud: would it be helpful if Cadence can provide FVM more info like, if there's any withdraw in the contract, and pass it to FVM as a flag, so that FVM will have more info to make decisions like this.

bluesign:
> i'm thinking out loud: would it be helpful if Cadence can provide FVM more info like, if there's any withdraw in the contract, and pass it to FVM as a flag, so that FVM will have more info to make decisions like this.

It is very good idea, but this being a good idea makes implementing it a bad idea ( in a way ) If this is a need ( and I agree it is ) solution is not for cadence to provide FVM information but instead FVM should take responsibility for storage (atree and friends ) ( as we are resource-aware blockchain ) and it should provide storage functions to Cadence. ( I know it is easier said then done )

engine/execution/computation/execution_verification_test.go
```diff
@@ -190,7 +191,7 @@ func Test_ExecutionMatchesVerification(t *testing.T) {
 		// ensure fee deduction events are emitted even though tx fails
 		require.Len(t, cr.Events[1], 3)
 		// storage limit error
-		assert.Contains(t, cr.TransactionResults[1].ErrorMessage, "Error Code: 1103")
+		assert.Contains(t, cr.TransactionResults[1].ErrorMessage, fmt.Sprintf("Error Code: %d", errors.ErrCodeInsufficientPayerBalance))
```
pattyshack, line 194: errors.ErrCodeInsufficientPayerBalance.String() instead of sprintf.  same elsewhere

fvm/errors/execution.go
```diff
@@ -47,6 +48,33 @@ func NewTransactionFeeDeductionFailedError(
...
+// indicates that the payer has insufficient balance to attempt transaction execution.
+func NewInsufficientPayerBalanceError(
+	payer flow.Address,
+	requiredBalance cadence.UFix64,
+) CodedError {
+	return NewCodedError(
+		ErrCodeInsufficientPayerBalance,
+		"payer %s has insufficient balance to attempt transaction execution (required balance: %s)",
+		payer,
+		requiredBalance.String(),
```
pattyshack, line 61: don't need String()

fvm/transactionPayerBalanceChecker.go
```diff
@@ -0,0 +1,56 @@
...
+	if !env.TransactionFeesEnabled() {
+		// if the transaction fees are not enabled,
+		// the payer can pay for the transaction,
+		// and the max transaction fees are 0.
+		// This is also the condition that gets hit during bootstrapping.
+		return 0, nil
+	}
+
+	resultValue, err :=
+		env.CheckPayerBalanceAndGetMaxTxFees(
```
pattyshack, line 30: delete newline between := and this line

fvm/transactionInvoker.go
```diff
@@ -275,6 +276,21 @@ func (executor *transactionExecutor) normalExecution() (
 	modifiedSets programsCache.ModifiedSetsInvalidator,
 	err error,
 ) {
+	executor.txnState.RunWithAllLimitsDisabled(func() {
```
pattyshack, line 279: move RunWithAllLimitsDisabled into TransactionPayerBalanceChecker / CheckPayerBalanceAndReturnMaxFees.  The executor doesn't need to know about this detail.

fvm/fvm_blockcontext_test.go
```diff
@@ -1033,13 +1033,30 @@ func TestBlockContext_ExecuteTransaction_InteractionLimitReached(t *testing.T) {
...
+				ctx.MaxStateInteractionSize = 500_000
+
+				txBody = testutil.CreateContractDeploymentTransaction(
 					"Container",
 					script,
 					accounts[0],
 					chain)
 
-				txBody.SetProposalKey(chain.ServiceAddress(), 0, 0)
+				txBody.SetProposalKey(chain.ServiceAddress(), 0, 1)
```
pattyshack, line 1059: instead of using magic number, maybe create a seqNum variable to track the number

fvm/transactionInvoker.go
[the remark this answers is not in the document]
pattyshack: o ... nevermind.  i need to move getBodyMeterParameters into preprocessing, which would defeat this point.

fvm/transactionPayerBalanceChecker.go
```diff
@@ -0,0 +1,56 @@
...
+			proc.Transaction.Payer,
+			proc.Transaction.InclusionEffort(),
+			uint64(txnState.TotalComputationLimit()),
+		)
+	if err != nil {
+		return 0, errors.NewPayerBalanceCheckError(proc.Transaction.Payer, err)
```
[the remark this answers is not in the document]
janezpodhostnik: I'm not sure This should be a failure, but I don't have a good argument against it. 

Let me create a separate PR for making this a failure, so I can move this one forward.

pattyshack (approved):

janezpodhostnik:
bors merge

PR #3475, opened 2022-10-28
Collect fees on tx seq num / verify non-fatal error

pattyshack:
A single errors collector is shared by all transaction processor.  This enables the transaction invoker to detect tx seq num / verify errors and collect fees using the errorExecution path (the normalExecution path is skipped)

Note: I plan replace the processor loop in TransactionProcedure with a linear workflow.  It's hard to reason across processors.


pattyshack:
o, on second thought, this is a bad idea.  this could open us to replay attack to drain the payer's account.  i'll just refactor the processor workflow without changing the fees behavior ....

PR #3477, opened 2022-10-28
[Access] Add Block Status to BlockResponse

### LINKED ISSUE #3106: The Access API should return the block status for all GetBlockXXX calls
## Problem Definition

What problem is this solving?

After a spork, the old network can continue to report finalized blocks with the same block height as the starting blocks of the new network.

e.g.
```
mainnet18 : A (last sealed) <- B (35858811) <- C <- D (last finalized) <- E <- F <- G
mainnet19 :                 ^- H (root block) <- I <- J
```

This is a source of confusion for clients that are consuming blocks from the old network before switching over to the network as reported [here](https://forum.onflow.org/t/provide-on-chain-and-off-chain-information-for-spork-information/3523/2)

## Proposed Solution

What are the proposed solutions to this problem?

The solution is to report the Block status (`finalized`, `executed`, `sealed`) as part of the block response for all Block related Get queries that the access node responds to. This applies to both the gRPC API and the REST API.

## Definition of Done

What tests must pass for this issue to be considered solved?

All Block related responses (gRPC and REST) must carry the block status.

## Actions Needed Before Submitting

Update ticket status using the following (remove this section once ticket created)

- What *workstream* does this ticket deal with? Find the appropriate 'S-' label and add that label.
- Is it a specific *'type'* of ticket (ex: bug, documentation)? If yes, add that label as well.
- Is this ticket related to an *overarching theme* (ex: architecture, performance)? If yes, add that label as well.
- Add any/all *descriptive characteristic* labels needed (ex: Needs Estimation, Needs Test Cases).
- Now we should determine what *release* this ticket is associated with. If none, leave it blank. If it is associated with a specific release, please add it to the appropriate release.
- If this ticket is associated with a release, we want to assign it a *level of importance* within that release. These labels follow the standard [MoSCoW method](https://en.wikipedia.org/wiki/MoSCoW_method) rules. We want to look at releases and then the importance of tickets within those specific releases. So the MoSCoW label is ONLY valid when it is taken in conjunction with its release.
- Assign this ticket a *priority level* (High, Medium, Low) via the appropriate label. These labels control the importance of the ticket within the sprint. For example, all `P-High` tickets should be worked on first, then `P-Medium`, then `P-Low`. This gives us an easy way to identify the order of priority for tickets within a specific sprint.

Simiar [issue](https://github.com/onflow/flow-cli/issues/576) on Flow-cli

lolpuddle:
For: #3106 

Pending: https://github.com/onflow/flow/pull/1157


koko1123 (approved):

engine/access/rpc/backend/backend_block_details.go
```diff
@@ -58,3 +58,15 @@ func (b *backendBlockDetails) GetBlockByHeight(_ context.Context, height uint64)
 
 	return block, nil
 }
+
+func (b *backendBlockDetails) GetBlockStatus(_ context.Context, id flow.Identifier) flow.BlockStatus {
+	block, err := b.blocks.ByID(id)
+	if err != nil {
+		return flow.BlockStatusUnknown
+	}
+	_, err = b.blocks.ByHeight(block.Header.Height + 1)
```
peterargue, line 67: I don't think this is an accurate way to get the status. Instead, you can use `b.state.Sealed().Head()` to get the latest sealed header and compare heights.

engine/access/rpc/backend/backend_block_details.go
```diff
@@ -58,3 +58,15 @@ func (b *backendBlockDetails) GetBlockByHeight(_ context.Context, height uint64)
 
 	return block, nil
 }
+
+func (b *backendBlockDetails) GetBlockStatus(_ context.Context, id flow.Identifier) flow.BlockStatus {
+	block, err := b.blocks.ByID(id)
```
peterargue, line 63: since you're only checking if the block exists, it's best to use `b.headers.ByID(id)` since it requires fewer db lookups and uses a cache

peterargue (commented):
Thanks for adding this feature. Super useful!

I think it would also be great to include the status when getting headers since they are often used as a light weight way to get block metadata

lolpuddle:
Pending: https://github.com/onflow/flow/pull/1190 to update the block header response

engine/access/rpc/backend/backend_block_details.go
lolpuddle: I think it would make even more sense to skip this step completely by passing the block/blockheader into getBlockStatus because the lookups for get block / get header would have already been performed by the time we're checking block status.

peterargue (approved):

engine/access/rest/blocks.go
```diff
@@ -194,7 +194,7 @@ func NewBlockProvider(backend access.API, options ...blockProviderOption) *block
 
 func (blkProvider *blockProvider) getBlock(ctx context.Context) (*flow.Block, error) {
 	if blkProvider.id != nil {
-		blk, err := blkProvider.backend.GetBlockByID(ctx, *blkProvider.id)
+		blk, _, err := blkProvider.backend.GetBlockByID(ctx, *blkProvider.id)
```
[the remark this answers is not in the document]
peterargue: please make sure to open an issue to track this if there isn't one already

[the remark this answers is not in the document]
lolpuddle: https://github.com/onflow/flow-go/issues/3622

PR #3478, opened 2022-10-29
fix heap draining bug in synchronization engine

AlexHentschel:
The heap for pending requests has a bug, where the flushes the entire content upon the first ejection. Its my own bug 🤦 . Thanks Peter for noticing.

### DIFF
--- a/engine/common/synchronization/request_heap_test.go
+++ b/engine/common/synchronization/request_heap_test.go
@@ -81,17 +81,27 @@ func TestRequestQueue_PutAtMaxCapacity(t *testing.T) {
 		m, ok := q.Get()
 		require.True(t, ok)
 
+		// in the following code segment, we check that:
+		//  - only previously inserted elements are popped from the heap
+		//  - each element is only popped once
 		expectedMessage, found := messages[m.OriginID]
 		require.True(t, found)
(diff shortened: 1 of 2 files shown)


engine/common/synchronization/request_heap_test.go
```diff
@@ -61,33 +61,40 @@ func TestRequestQueue_Put(t *testing.T) {
...
+		require.True(t, found)
+		require.Equal(t, m, expectedMessage)
+		if m == lastMsg {
+			lastMessagePopped = true
 		}
+
+		delete(messages, m.OriginID)
 	}
-	require.Equal(t, 1, ejectedCount)
+
```
peterargue, line 93: can you also add a final check here to make sure the heap is empty?

durkmurder (approved):

engine/common/synchronization/request_heap_test.go
AlexHentschel: 👍 added in https://github.com/onflow/flow-go/pull/3478/commits/86377dbe419b4df2db4f52257e7ac1629cc1dbc9

peterargue (approved):

AlexHentschel:
bors merge

PR #3483, opened 2022-10-31
Prep work for converting transaction invoker into continuation passing style executor

pattyshack:
Added an executor object and converted variables that span the entire workflow (preprocessing, actual execution) into fields.  Variables that do not span the entire workflow (e.g., modifiedSets) are left unmodified.

Note that I've also made a few no-op stylistic changes.

### DIFF
--- a/fvm/transactionInvoker.go
+++ b/fvm/transactionInvoker.go
@@ -8,6 +8,7 @@ import (
 	"github.com/onflow/cadence/runtime/common"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/otel/attribute"
+	otelTrace "go.opentelemetry.io/otel/trace"
 
 	"github.com/onflow/flow-go/fvm/environment"
 	"github.com/onflow/flow-go/fvm/errors"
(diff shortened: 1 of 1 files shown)


fvm/transactionInvoker.go
```diff
@@ -252,23 +264,25 @@ func (i TransactionInvoker) commit(
 	}
 
 	// if tx failed this will only contain fee deduction logs
-	proc.Logs = append(proc.Logs, env.Logs()...)
-	proc.ComputationUsed = proc.ComputationUsed + env.ComputationUsed()
-	proc.MemoryEstimate = proc.MemoryEstimate + env.MemoryEstimate()
-	if proc.IsSampled() {
-		proc.ComputationIntensities = env.ComputationIntensities()
+	executor.proc.Logs = append(executor.proc.Logs, executor.env.Logs()...)
+	executor.proc.ComputationUsed += executor.env.ComputationUsed()
```
pattyshack, line 266: note: switched to `+=` here

fvm/transactionInvoker.go
```diff
@@ -181,69 +206,56 @@ func (i TransactionInvoker) normalExecution(
...
-func (i TransactionInvoker) errorExecution(
-	proc *TransactionProcedure,
-	txnState *state.TransactionState,
-	nestedTxnId state.NestedTransactionId,
-	env environment.Environment,
-	errs *errors.ErrorsCollector,
-) error {
-	txnState.DisableAllLimitEnforcements()
-	defer txnState.EnableAllLimitEnforcements()
+func (executor *invocationExecutor) errorExecution() {
```
pattyshack, line 217: note: this no longer returns the fatal error.  `errs.ErrorOrNil()` is called in `Process` instead (see line 103)

Tonix517 (approved):

Tonix517:
and what will we use CPS for?

janezpodhostnik (approved):

pattyshack:
the plan is to split `Process` into two phases: `Preprocess` and `Execute`.   Program parsing will be done as part of preprocessing while the rest will remain in execute.  This is the cps part

This enables us to initially parallelize preprocessing without having general OCC.  The general execution flow would be something like:
1. Preprocess the transaction
2. Wait for all earlier transactions to finish committing
3. Validate Preprocess (if invalidated, use the newest view and retry)
4. "Rebase" the transaction view to the latest view snapshot
5. Execute and commit transaction

Steps 1-2 are executed in parallel across transactions.  Steps 3-5 are executed sequentially across transactions.

Once we support general view OCC, the general execution flow would be something like:
1. Preprocess the transaction (in parallel with other transaction, using an older view snapshot)
2.  Valdiate Preprocess
3.  "Rebase" to latest committed view snapshot (without waiting)
4. Execute
5. Wait for all earlier transactions to finish committing
6. Validate Execute (if invalidated, use the newest view and retry)
7. Commit

Steps 1-5 are executed in parallel across transactions.  Steps 6-7 are executed sequentially across transactions.

Tonix517:
sounds promising overall. one question, for:

> 5. Execute and commit transaction

things like tx seq num, fee deductions etc., can all be handled by OCC then?

pattyshack:
Those are all just business logic at the end of the day.

OCC operates on the principle assumption that given a set of inputs (the read set), the business logic will deterministically generate a set of outputs (the write set).  With this assumption, as long as the input set remains unmodified, the generated output set should also be unmodified as well.

pattyshack:
bors merge

PR #3484, opened 2022-10-31
[FVM] Adding generic OCC implementation

Tonix517:
To facilitate parallel\speculative transaction execution, all derived data(programs, meter settings etc.) should be OCC-enabled. This PR separates out the OCC logic in existing BlockPrograms into its own individual generic type, in which:

- `block_derived_data.go`. 
Majority of code\comments are directly taken from block_programs.go, except that `nonAddressSet` is removed from the generic version, which is program specific.
- `block_derived_data_invalidator.go`
Majority of the code\comments are directly from invalidators.go, except some minor renaming, except that ContractUpdate\Key and ModifiedSetsInvalidator are not included. It will be included in new BlockPrograms specific implementations.

NOTE: test will be done in the next PR together with BlockProgram refactoring using this generic types. Existing UTs will be enough.

### DIFF
--- a/fvm/programs/block_derived_data.go
+++ b/fvm/programs/block_derived_data.go
@@ -1,0 +1,358 @@
+package programs
+
+import (
+	"fmt"
+	"sync"
+)
+
+type invalidatableEntry[TVal any] struct {
+	Entry TVal // Immutable after initialization.
+
+	isInvalid bool // Guarded by BlockDerivedData' lock.
+}
+
+// BlockDerivedData is a simple fork-aware OCC database for "caching" given types
+// of data for a particular block.
+//
+// Since data are derived from external source, the database need not be
+// durable and can be recreated on the fly.
+//
+// Furthermore, because data are derived, transaction validation looks
+// a bit unusual when compared with a textbook OCC implementation.  In
+// particular, the transaction's invalidator represents "real" writes to the
+// canonical source, whereas the transaction's readSet/writeSet entries
+// represent "real" reads from the canonical source.
+type BlockDerivedData[TKey comparable, TVal any] struct {
+	lock  sync.RWMutex
+	items map[TKey]*invalidatableEntry[TVal]
+
+	latestCommitExecutionTime LogicalTime
+
+	invalidators chainedDerivedDataInvalidators[TVal] // Guarded by lock.
+}
+
+type TransactionDerivedData[TKey comparable, TVal any] struct {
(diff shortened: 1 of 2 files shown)


fvm/programs/block_derived_data.go
```diff
@@ -0,0 +1,358 @@
...
+	// the valid write entries then return error.
+	err := block.unsafeValidate(item)
+	if err != nil {
+		return err
+	}
+
+	for key, entry := range item.writeSet {
+		_, ok := block.items[key]
+		if ok {
+			// A previous transaction already committed an equivalent program
```
pattyshack, line 216: s/program/derived data/g

fvm/programs/block_derived_data.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func (block *BlockDerivedData[TKey, TVal]) unsafeValidate(
+	item *BlockDerivedDataItem[TKey, TVal],
+) RetryableError {
+	if item.isSnapshotReadTransaction &&
+		item.invalidators.ShouldInvalidateItems() {
+
+		return newNotRetryableError(
+			"invalid TransactionProgram: snapshot read can't invalidate")
```
pattyshack, line 150: need to update the error strings

pattyshack (commented):
btw, the Get/Set api is only necessary to accommodate cadence runtime interface.  Your api `GetOrCompute` (or something like that) from the previous PR is probably better for general usage.  Maybe add that as a follow up PR

fvm/programs/block_derived_data.go
```diff
@@ -0,0 +1,358 @@
...
+	snapshotTime LogicalTime,
+	executionTime LogicalTime,
+	isSnapshotReadTransaction bool,
+) (
+	*BlockDerivedDataItem[TKey, TVal],
+	error,
+) {
+	if executionTime < 0 || executionTime > upperBoundExecutionTime {
+		return nil, fmt.Errorf(
+			"invalid TransactionPrograms: execution time out of bound: %v",
```
janezpodhostnik, line 264: TransactionPrograms => BlockDerivedData

janezpodhostnik (approved):
Looks good.

fvm/programs/block_derived_data.go
janezpodhostnik: oh, Patrick already mentioned this :).

Tonix517:
bors merge

PR #3487, opened 2022-11-01
convert seq num checker / txn verifier to CPS executor objects.

pattyshack:
also made invocationExecutor implement the same interface.


pattyshack:
i've changed my mind on this.  i'll get rid of the concept of transaction processor instead.

PR #3496, opened 2022-11-02
Convert ScriptProcedure.Run into executor object.

pattyshack:
This also removed the concept of ScriptProcessor since we never override the script execution behavior

### DIFF
--- a/fvm/context.go
+++ b/fvm/context.go
@@ -25,7 +25,6 @@ type Context struct {
 	MaxStateInteractionSize           uint64
 
 	TransactionProcessors []TransactionProcessor
-	ScriptProcessors      []ScriptProcessor
 
 	BlockPrograms *programs.BlockPrograms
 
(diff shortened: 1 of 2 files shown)


Tonix517 (approved):
could you also create a brief write-up on the overall design of current, and planned refactoring?

janezpodhostnik (approved):

pattyshack:
With respect to refactoring, we need to expose `Preprocess` / `Execute` at the Procedure level in order to use it in the computer.  (The end goal is to have "fvm transaction procedure's executor call out to fvm transaction processor's executor, which then call out to cadence's transaction executor).  We don't need to change script or bootstrap procedure, but they need to implement the same interface.

After the basic refactoring, we would need to expose validation methods at the Procedure executor level.  These validation methods will validate the txnPrograms (and txnState) changes

pattyshack:
bors merge

PR #3507, opened 2022-11-03
[Networking] Backporting PR 3499 to master

yhassanzadeh13:
https://github.com/onflow/flow-go/pull/3499/files


zhangchiqing (approved):

network/p2p/middleware/middleware.go
```diff
@@ -59,7 +59,7 @@ const (
 
 	// DefaultUnicastTimeout is the default maximum time to wait for a default unicast request to complete
 	// assuming at least a 1mb/sec connection
-	DefaultUnicastTimeout = 5 * time.Second
+	DefaultUnicastTimeout = 20 * time.Second
```
peterargue, line 62: 20 seconds seems like a long time for a default. is this what's running on mainnet?

peterargue (approved):

network/p2p/middleware/middleware.go
yhassanzadeh13: We experimentally run it on mainnet only for ENs, but as we now have more clues about the unicast failures in the first place, I reverted it back. Nevertheless, we can set it using the flags at runtime.

PR #3510, opened 2022-11-03
[Execution] Improve locking in requester and provider engines

peterargue:
This PR makes 2 changes:
* In the provider engine, remove the lock protecting `onEntityRequest`. This method doesn't have any shared state and shouldn't need locking
* In the requester engine, unlock before sending a request over unicast to avoid blocking the network layer if there are slow unicast requests.

### DIFF
--- a/engine/common/provider/engine.go
+++ b/engine/common/provider/engine.go
@@ -133,9 +133,6 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
 	e.metrics.MessageReceived(e.channel.String(), metrics.MessageEntityRequest)
 	defer e.metrics.MessageHandled(e.channel.String(), metrics.MessageEntityRequest)
 
-	e.unit.Lock()
-	defer e.unit.Unlock()
-
 	switch msg := message.(type) {
 	case *messages.EntityRequest:
 		return e.onEntityRequest(originID, msg)
--- a/engine/common/provider/engine.go
+++ b/engine/common/provider/engine.go
@@ -25,6 +25,7 @@ import (
 // related entities. It is important that the retrieve function return a
 // `storage.ErrNotFound` error if the entity does not exist locally; otherwise,
 // the logic will error and not send responses when failing to retrieve entities.
+// This method must be concurrency safe.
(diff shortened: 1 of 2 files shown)


engine/common/requester/engine.go
```diff
@@ -290,6 +285,9 @@ func (e *Engine) dispatchRequest() (bool, error) {
 	now := time.Now().UTC()
 	var providerID flow.Identifier
 	var entityIDs []flow.Identifier
+
+	e.unit.Lock()
```
zhangchiqing, line 289:
```suggestion
	// Lock here, and to unlock before calling Unicast
	e.unit.Lock()
```

zhangchiqing (approved):

engine/common/provider/engine.go
```diff
@@ -133,9 +133,6 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
 	e.metrics.MessageReceived(e.channel.String(), metrics.MessageEntityRequest)
 	defer e.metrics.MessageHandled(e.channel.String(), metrics.MessageEntityRequest)
 
-	e.unit.Lock()
```
zhangchiqing, line 136: this lock is not protecting anything, as the retrieve method is concurrent safe.

The provider engine is used in:
- collection node [providing collections](https://github.com/onflow/flow-go/blob/v0.28/cmd/collection/main.go#L381)
- execution node providing its own receipts](https://github.com/onflow/flow-go/blob/v0.28/cmd/execution_builder.go#L885)

Both retrieve functions are already concurrent safe, so we can safely remove this lock

engine/common/requester/engine.go
```diff
@@ -275,11 +275,6 @@ PollLoop:
 
 func (e *Engine) dispatchRequest() (bool, error) {
 
-	e.unit.Lock()
-	defer e.unit.Unlock()
```
zhangchiqing, line 279: This locks for too long, we'd like to lock more granular in order to exclude Unicast from the lock period

yhassanzadeh13 (approved):

PR #3513, opened 2022-11-04
[Consensus] Standalone startup of hotstuff participant

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6420

### Context

This PR implements standalone startup of hotstuff participant separating startup from compliance engine. 

### Important changes
- Hotstuff participant decoupled from compliance engine, hotstuff is created before compliance core and is injected using constructor. 
- Refactored epoch factories for collection nodes to split construction of sync core and sync engine to allow injection using construction for compliance engine.


jordanschalm (approved):
Looks great, nice clean up 🧹

engine/collection/compliance/engine.go
```diff
@@ -90,53 +89,6 @@ func NewEngine(
 	return eng, nil
 }
```
AlexHentschel, line 91: wondering if we still need the field `Engine.cm`: https://github.com/onflow/flow-go/blob/6a6b5eb8f5361813623ff4a26944590fb9b45716/engine/collection/compliance/engine.go#L43

I think the only place where we use it is in the constructor, where we assign its value to `eng.Component`: https://github.com/onflow/flow-go/blob/6a6b5eb8f5361813623ff4a26944590fb9b45716/engine/collection/compliance/engine.go#L82-L87

### suggestion:
could apply the same revision here, as you did for the [main consensus compliance engine](https://github.com/onflow/flow-go/blob/6a6b5eb8f5361813623ff4a26944590fb9b45716/engine/consensus/compliance/engine.go#L47)
Screen Shot 2022-11-21 at 8 32 56 PM

AlexHentschel (approved):
Great cleanup. Only stylistic comments. Thanks

durkmurder:
bors merge

PR #3515, opened 2022-11-04
[Requester] Fix blocking call in requester

zhangchiqing:
Fix a bug that causing EN to stuck on execution from time to time. 

See [comments ](https://github.com/onflow/flow-go/pull/3515/files#r1014327314) for reasoning and [solution](https://github.com/onflow/flow-go/pull/3515/files#r1014329994).

### DIFF
--- a/engine/common/requester/engine.go
+++ b/engine/common/requester/engine.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/rs/zerolog"
 	"github.com/vmihailenco/msgpack"
+	"go.uber.org/atomic"
 
 	"github.com/onflow/flow-go/engine"
 	"github.com/onflow/flow-go/model/flow"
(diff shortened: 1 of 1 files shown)


engine/common/requester/engine.go
```diff
@@ -234,18 +237,24 @@ func (e *Engine) addEntityRequest(entityID flow.Identifier, selector flow.Identi
 // Force will force the requester engine to dispatch all currently
 // valid batch requests.
 func (e *Engine) Force() {
-	count := uint(0)
-	for {
```
zhangchiqing, line 238: [Force is called](https://github.com/onflow/flow-go/blob/master/engine/execution/ingestion/engine.go#L988) by ingestion engine when requesting collections.

However, this Force method is blocking, which causes the call `matchOrRequestCollections` to be blocking, which created other side effects.

engine/common/requester/engine.go
```diff
@@ -234,19 +237,26 @@ func (e *Engine) addEntityRequest(entityID flow.Identifier, selector flow.Identi
...
-			e.log.Error().Err(err).Msg("could not dispatch requests")
-			return
-		}
-		if !dispatched {
-			e.log.Debug().Uint("requests", count).Msg("forced request dispatch")
-			return
+	// using Launch to ensure the caller won't be blocked
+	e.unit.Launch(func() {
+		// using atomic bool to ensure there is at most one caller would trigger dispatching requests
+		if e.isDispatching.CompareAndSwap(false, true) {
```
zhangchiqing, line 243: using Launch and atomic bool ensure:
1. the caller won't be blocked
2. only running it at most once at anytime

peterargue (approved):

m4ksio (approved):

PR #3516, opened 2022-11-04
Renamed programs to derived data

pattyshack:
We'll be caching non-programs derived data soon.  I'll rename the package in a followup PR (That involves creating aliases, needed for emulator/integration test backward compatibility)

gh: #3102


SaveTheRbtz (approved):
Please fix linter issues: a couple of `programs` mentions were not rewritten and produce the following error:
```use of package programs not in selector```

File: cmd/execution_config.go [hunk 1]
```diff
@@ -72,7 +72,7 @@ func (exeConf *ExecutionConfig) SetupFlags(flags *pflag.FlagSet) {
 	flags.UintVar(&exeConf.checkpointsToKeep, "checkpoints-to-keep", 5, "number of recent checkpoints to keep (0 to keep all)")
 	flags.BoolVar(&exeConf.outputCheckpointV5, "outputCheckpointV5", false, "output checkpoint file in v5")
 	flags.UintVar(&exeConf.stateDeltasLimit, "state-deltas-limit", 100, "maximum number of state deltas in the memory pool")
-	flags.UintVar(&exeConf.computationConfig.ProgramsCacheSize, "cadence-execution-cache", programs.DefaultProgramsCacheSize,
+	flags.UintVar(&exeConf.computationConfig.DerivedDataCacheSize, "cadence-execution-cache", programs.DefaultDerivedDataCacheSize,
```
zhangchiqing, line 75: Can you briefly explain the background of how "data derive" works?

- Who derive what data from whom?
- Is Derived data something shared between certain modules? What are those modules?

pattyshack: derived data is just any data computed (aka derived) from storage/ledger (aka primary index).

BlockPrograms currently is a specific secondary index which keeps track of parsed programs.  I'm generalizing this to support additional secondary indices, namely to keep track of parsed meter setting.

zhangchiqing: Deriving data is to speed up executing the next transaction, right?

Is Derived data the data derived from the execution environment of the previous transaction or previous block?

janezpodhostnik (approved):

cmd/execution_config.go
janezpodhostnik: Storing derived data in a cache allows us to not "re-derive" it later, as that requires either running smart contract functions or unpacking a cadence controlled register or parsing a smart contract.

Derived data should reflect the current relevant on chain state at the start of each transaction (for example execution effort weights for computation metering is derived from a storage location (storage map) from the service account). It is up to the FVM implementation to correctly trigger invalidation of derived data in case the relevant chain state is touched during the transaction.

pattyshack: the data is derived from previous transactions, regardless of which block the transaction originally came from.

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors retry

PR #3517, opened 2022-11-04
Backport: make requester's Force method non-blocking

zhangchiqing:
Backport https://github.com/onflow/flow-go/pull/3515 to master

### DIFF
--- a/engine/common/requester/engine.go
+++ b/engine/common/requester/engine.go
@@ -35,20 +35,20 @@ type CreateFunc func() flow.Entity
 // on the flow network. It is the `request` part of the request-reply
 // pattern provided by the pair of generic exchange engines.
 type Engine struct {
-	unit          *engine.Unit
-	log           zerolog.Logger
-	cfg           Config
-	metrics       module.EngineMetrics
-	me            module.Local
-	state         protocol.State
-	con           network.Conduit
-	channel       channels.Channel
-	selector      flow.IdentityFilter
-	create        CreateFunc
-	handle        HandleFunc
-	items         map[flow.Identifier]*Item
-	requests      map[uint64]*messages.EntityRequest
-	isDispatching *atomic.Bool // to ensure only trigger dispatching logic once at any time
+	unit                  *engine.Unit
+	log                   zerolog.Logger
+	cfg                   Config
+	metrics               module.EngineMetrics
+	me                    module.Local
+	state                 protocol.State
+	con                   network.Conduit
+	channel               channels.Channel
+	selector              flow.IdentityFilter
+	create                CreateFunc
+	handle                HandleFunc
+	items                 map[flow.Identifier]*Item
+	requests              map[uint64]*messages.EntityRequest
+	forcedDispatchOngoing *atomic.Bool // to ensure only trigger dispatching logic once at any time
 }
 
 // New creates a new requester engine, operating on the provided network channel, and requesting entities from a node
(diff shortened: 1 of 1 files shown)


engine/common/requester/engine.go
```diff
@@ -234,19 +237,26 @@ func (e *Engine) addEntityRequest(entityID flow.Identifier, selector flow.Identi
...
-		if !dispatched {
-			e.log.Debug().Uint("requests", count).Msg("forced request dispatch")
-			return
+	// using Launch to ensure the caller won't be blocked
+	e.unit.Launch(func() {
+		// using atomic bool to ensure there is at most one caller would trigger dispatching requests
+		if e.isDispatching.CompareAndSwap(false, true) {
+			count := uint(0)
+			for {
+				dispatched, err := e.dispatchRequest()
```
AlexHentschel, line 246: may I request a brief documentation of the return values of `dispatchRequest()`? Just a one-liner like:
```golang
// dispatchRequest dispatches a subset of requests (selection based on internal heuristic). 
// The boolean return value indicates whether a request was dispatched at all.
```

engine/common/requester/engine.go
```diff
@@ -234,19 +237,26 @@ func (e *Engine) addEntityRequest(entityID flow.Identifier, selector flow.Identi
...
-	for {
-		dispatched, err := e.dispatchRequest()
-		if err != nil {
-			e.log.Error().Err(err).Msg("could not dispatch requests")
-			return
-		}
-		if !dispatched {
-			e.log.Debug().Uint("requests", count).Msg("forced request dispatch")
-			return
+	// using Launch to ensure the caller won't be blocked
```
AlexHentschel, line 240: the performance problems seem to indicate that the `Force` might run for some time (?) If that is the case, we could include some logic for exiting early, because the `unit.Launch` has some overhead attached to it:

```suggestion
	// exit early in case a forced dispatch is currently ongoing
	if e.isDispatching.Load() { 
	   return
	}
	// using Launch to ensure the caller won't be blocked
```

engine/common/requester/engine.go
```diff
@@ -99,18 +101,19 @@ func New(log zerolog.Logger, metrics module.EngineMetrics, net network.Network,
...
+		metrics:       metrics,
+		me:            me,
+		state:         state,
+		channel:       channel,
+		selector:      selector,
+		create:        create,
+		handle:        nil,
+		items:         make(map[flow.Identifier]*Item),          // holds all pending items
+		requests:      make(map[uint64]*messages.EntityRequest), // holds all sent requests
+		isDispatching: atomic.NewBool(false),
```
AlexHentschel, line 116: I find the naming a bit counterintuitive, because it seems to indicate that `isDispatching` is true whenever _any_ dispatch is ongoing. But that is not the case, it is only true when a _forced_ dispatch is running. Would suggest to reflect it in the variable name. 
```suggestion
		forcedDispatchOngoing: atomic.NewBool(false),
```

AlexHentschel: Another thought:
I don't think there is much use running a simple dispatch, if a forced dispatch is already ongoing. We modify the logic for regularly-timed requests here
https://github.com/onflow/flow-go/blob/1f231ce7e3361320018aa65b5a319ef78b075ebe/engine/common/requester/engine.go#L271-L272 
and exit early 
```
case <-ticker.C:
  if e.isDispatching.Load() { 
    return
  }
  dispatched, err := e.dispatchRequest()
```

With this modification, a small time window remains, where a regular dispatch runs into `dispatchRequest()` and then the `unit.Lock`  is already taken by a forced dispatch. But assuming that Forced dispatches potentially take some time to run (which seems to be the case), this extra check might reduce unnecessary dispatches.

AlexHentschel (approved):
Awesome that you found the bug causing EN 1 to hang! 👏 
Couple stylistic comments.

engine/common/requester/engine.go
zhangchiqing: Good point 👍

jordanschalm (approved):

zhangchiqing:
bors merge

PR #3520, opened 2022-11-05
[Profile] Add ability to manually trigger profiler

SaveTheRbtz:
This PR adds an ability to manually trigger auto-profiler for ad-hoc troubleshooting, e.g.:
```
$ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' \
    -d '{"commandName": "set-config", "data": {"profiler-trigger": "1m"}}'
```

### DIFF
--- a/module/profiler/profiler.go
+++ b/module/profiler/profiler.go
@@ -37,6 +37,9 @@ type AutoProfiler struct {
 
 	uploader Uploader
 	enabled  *atomic.Bool
+
+	// used to trigger a profile run
+	trigger chan struct{}
 }
 
 // New creates a new AutoProfiler instance performing profiling every interval for duration.
@@ -55,7 +58,11 @@ func New(log zerolog.Logger, uploader Uploader, dir string, interval time.Durati
 		duration: duration,
 		uploader: uploader,
 		enabled:  atomic.NewBool(enabled),
+		trigger:  make(chan struct{}, 1),
 	}
+
+	go p.runForever()
+
 	return p, nil
 }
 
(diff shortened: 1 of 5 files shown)


module/profiler/profiler.go
```diff
@@ -71,28 +78,52 @@ func (p *AutoProfiler) Enabled() bool {
...
+	case p.trigger <- d:
+		return nil
+	default:
+		return errors.New("profiling is already in progress")
+	}
+}
+
+func (p *AutoProfiler) runForever() {
+	jitter := time.Duration(rand.Int63n(int64(p.interval)))
+	t := time.NewTicker(p.interval + jitter)
```
pattyshack, line 93: should we recompute the jitter on each iteration?

also, the expected interval is now 1.5 * p.interval.  should we normalized?

module/profiler/profiler.go
```diff
@@ -71,28 +78,52 @@ func (p *AutoProfiler) Enabled() bool {
...
 
-	if !p.Enabled() {
+	for {
+		select {
+		case <-t.C:
+			if p.Enabled() {
+				p.runOnce()
+			}
+		case d := <-p.trigger:
+			oldDuration := p.duration
```
pattyshack, line 103: nit: maybe change `type profileFunc func(io.Writer) error` to `type profileFunc func(io.Writer, duration) error`, and pass the duration into runOnce as an argument.

pattyshack (approved):

module/profiler/profiler.go
```diff
@@ -71,28 +78,52 @@ func (p *AutoProfiler) Enabled() bool {
...
+	select {
+	case p.trigger <- d:
+		return nil
+	default:
+		return errors.New("profiling is already in progress")
+	}
+}
+
+func (p *AutoProfiler) runForever() {
+	jitter := time.Duration(rand.Int63n(int64(p.interval)))
```
zhangchiqing, line 92: The randominess is used in initial delay, the interval doesn't have randominess. 
But your implementation also added randominess in intervals, why is that?

zhangchiqing (commented):
Looks good, just had one question

module/profiler/profiler.go
SaveTheRbtz: There is no random in initial delay, so I've added randomness to the interval itself.

module/profiler/profiler.go
```diff
@@ -71,33 +83,53 @@ func (p *AutoProfiler) Enabled() bool {
 	return p.enabled.Load()
 }
 
-func (p *AutoProfiler) Ready() <-chan struct{} {
-	delay := time.Duration(float64(p.interval) * rand.Float64())
```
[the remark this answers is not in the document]
SaveTheRbtz: i've removed the `LaunchPeriodically` because I wanted to not only have the the periodic trigger but also a manual one.  Since we now doing the the scheduling ourselves it is easier to randomize the interval itself instead of introducing the delay on the initial tick of the ticker. (I can't just wait on `time.After` since we want to be able to receive manual trigger during that time.)

zhangchiqing (approved):

SaveTheRbtz:
bors merge

PR #3523, opened 2022-11-07
Add consensus and sync-related message types without nil struct fields

jordanschalm:
This PR implements a new approach to structural validity, by adding new inbound message types without nil struct fields. Replaces https://github.com/onflow/flow-go/pull/3470.

The primary purpose is to ensure safety against nil struct fields or slice entries. The secondary purpose is to begin to separate the internal (trusted) and external (untrusted) representations of a message, to enable eg. caching the ID of a trusted entity model.

This is still a draft - I will add documentation for the new types if we're happy with the general approach.

### DIFF
--- a/engine/collection/compliance/core.go
+++ b/engine/collection/compliance/core.go
@@ -81,7 +81,9 @@ func NewCore(
 
 // OnBlockProposal handles incoming block proposals.
 func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.ClusterBlockProposal) error {
-	header := proposal.Header
+	block := proposal.Block.ToInternal()
+	header := block.Header
+
 	log := c.log.With().
 		Hex("origin_id", originID[:]).
 		Str("chain_id", header.ChainID.String()).
@@ -89,9 +91,9 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Uint64("block_view", header.View).
 		Hex("block_id", logging.Entity(header)).
 		Hex("parent_id", header.ParentID[:]).
-		Hex("ref_block_id", proposal.Payload.ReferenceBlockID[:]).
-		Hex("collection_id", logging.Entity(proposal.Payload.Collection)).
-		Int("tx_count", proposal.Payload.Collection.Len()).
+		Hex("ref_block_id", block.Payload.ReferenceBlockID[:]).
+		Hex("collection_id", logging.Entity(block.Payload.Collection)).
+		Int("tx_count", block.Payload.Collection.Len()).
 		Time("timestamp", header.Timestamp).
 		Hex("proposer", header.ProposerID[:]).
 		Hex("signers", header.ParentVoterIndices).
@@ -195,7 +197,8 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 // to a valid proposal are validly connected to the finalized state and can be
 // processed as well.
 func (c *Core) processBlockAndDescendants(proposal *messages.ClusterBlockProposal) error {
-	blockID := proposal.Header.ID()
+	block := proposal.Block.ToInternal()
+	blockID := block.ID()
 
 	// process block itself
 	err := c.processBlockProposal(proposal)
@@ -228,8 +231,10 @@ func (c *Core) processBlockAndDescendants(proposal *messages.ClusterBlockProposa
 	}
 	for _, child := range children {
 		childProposal := &messages.ClusterBlockProposal{
-			Header:  child.Header,
-			Payload: child.Payload,
+			Block: messages.UntrustedClusterBlockFromInternal(&cluster.Block{
+				Header:  child.Header,
+				Payload: child.Payload,
+			}),
 		}
 		cpr := c.processBlockAndDescendants(childProposal)
 		if cpr != nil {
@@ -247,7 +252,8 @@ func (c *Core) processBlockAndDescendants(proposal *messages.ClusterBlockProposa
 // processBlockProposal processes the given block proposal. The proposal must connect to
 // the finalized state.
 func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) error {
-	header := proposal.Header
+	block := proposal.Block.ToInternal()
+	header := block.Header
 	log := c.log.With().
 		Str("chain_id", header.ChainID.String()).
 		Uint64("block_height", header.Height).
@@ -262,10 +268,6 @@ func (c *Core) processBlockProposal(proposal *messages.ClusterBlockProposal) err
 	log.Info().Msg("processing block proposal")
 
 	// see if the block is a valid extension of the protocol state
-	block := &cluster.Block{
-		Header:  proposal.Header,
-		Payload: proposal.Payload,
-	}
 	err := c.state.Extend(block)
 	// if the block proposes an invalid extension of the protocol state, then the block is invalid
 	if state.IsInvalidExtensionError(err) {
--- a/engine/common/follower/engine.go
+++ b/engine/common/follower/engine.go
@@ -197,8 +197,7 @@ func (e *Engine) onSyncedBlock(originID flow.Identifier, synced *events.SyncedBl
 
 	// process as proposal
 	proposal := &messages.BlockProposal{
-		Header:  synced.Block.Header,
-		Payload: synced.Block.Payload,
+		Block: synced.Block,
 	}
 	return e.onBlockProposal(originID, proposal, false)
 }
@@ -206,8 +205,7 @@ func (e *Engine) onSyncedBlock(originID flow.Identifier, synced *events.SyncedBl
 func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockResponse) error {
 	for i, block := range res.Blocks {
 		proposal := &messages.BlockProposal{
-			Header:  block.Header,
-			Payload: block.Payload,
+			Block: block,
 		}
 
 		// process block proposal with a wait
@@ -221,11 +219,11 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockRe
 
 // onBlockProposal handles incoming block proposals. inRangeBlockResponse will determine whether or not we should wait in processBlockAndDescendants
 func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.BlockProposal, inRangeBlockResponse bool) error {
+	block := proposal.Block.ToInternal()
+	header := block.Header
 
-	span, ctx, _ := e.tracer.StartBlockSpan(context.Background(), proposal.Header.ID(), trace.FollowerOnBlockProposal)
+	span, ctx, _ := e.tracer.StartBlockSpan(context.Background(), header.ID(), trace.FollowerOnBlockProposal)
 	defer span.End()
-
-	header := proposal.Header
 
 	log := e.log.With().
 		Hex("origin_id", originID[:]).
@@ -355,11 +353,11 @@ func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.Bl
 // any children are therefore also connected to the finalized state and can be processed as well.
 // No errors are expected during normal operations.
 func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messages.BlockProposal, inRangeBlockResponse bool) error {
+	block := proposal.Block.ToInternal()
+	header := block.Header
 
 	span, ctx := e.tracer.StartSpanFromContext(ctx, trace.FollowerProcessBlockProposal)
 	defer span.End()
-
-	header := proposal.Header
 
 	log := e.log.With().
 		Str("chain_id", header.ChainID.String()).
@@ -373,12 +371,6 @@ func (e *Engine) processBlockAndDescendants(ctx context.Context, proposal *messa
 		Logger()
 
 	log.Info().Msg("processing block proposal")
-
-	// see if the block is a valid extension of the protocol state
-	block := &flow.Block{
-		Header:  proposal.Header,
-		Payload: proposal.Payload,
-	}
 
 	// check whether the block is a valid extension of the chain.
 	// it only checks the block header, since checking block body is expensive.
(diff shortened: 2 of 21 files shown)


model/messages/consensus.go
```diff
@@ -4,11 +4,103 @@ import (
...
-// BlockProposal is part of the consensus protocol and represents the the leader
+type UntrustedExecutionResult struct {
+	PreviousResultID flow.Identifier
+	BlockID          flow.Identifier
+	Chunks           []flow.Chunk
+	ServiceEvents    flow.ServiceEventList
+	ExecutionDataID  flow.Identifier
+}
+
+func (ur UntrustedExecutionResult) ToFlowResult() *flow.ExecutionResult {
```
jordanschalm, line 15:
```suggestion
func (ur UntrustedExecutionResult) ToInternal() *flow.ExecutionResult {
```

jordanschalm:
In this PR I modified inbound message types, to use a distinct "untrusted" type for blocks and some block field types, in particular `Payload` and `ExecutionResult`, because these contained values which could be nil when received over the network.

So we have:
* `flow.Block` - the internal representation of the block
* `messages.UntrustedBlock` - the external representation of the block
  * contains duplicated `UntrustedPayload`, `UntrustedExecutionResult` types

These types are both nominally and structurally distinct, because of the difference in pointer fields and slice elements. They are also not consistently structurally distinct, because only sub-entities which contained potentially unsafe pointer fields were duplicated as a distinct `UntrustedXYZ` type. Other sub-entities are represented as the same type. This results in:
* verbose and unnecessary type conversion code
* less ability to share utility code which operates on core models

What we want instead, is types like:
* `flow.Block` - shared model, no notion of trusted vs untrusted
  * utility code and helper functions use this model
*  `messages.UntrustedBlock` - the external untrusted representation of the block
```go
// same core model structure, nominally distinct
type UntrustedBlock struct { 
  flow.Block
}
```
* `flow.TrustedBlock` - trusted, validated, internal representation of the block
```go
// same core model structure, nominally distinct, can embed additional computed data
type TrustedBlock struct {
  flow.Block
  computedID flow.Identifier
  // ... 
}
```

However, implementing this would require removing pointer fields and slice elements from the core types, which is a much bigger time investment than only adding the additional types. The fields that need to be changed are heavily used (>100 usages each) and individual usages can require more than trivial changes to update.

Given that end goal, this PR will introduce some technical debt in the form of duplicated types and conversion code.

durkmurder (commented):
It really depends on what path we want to take with our data structures. I think we can all agree that currently we use pointers too much, and our structures are not optimal. 
I see a few ways that we can take in long-run:

1. Complete update of our core data structure to use values instead of pointers.
(+) better concurrency guarantees 
(+) less stress on GC
(+) save time on calculating ID, ability to extend trusted data with pre-computed values
(+) no need to add extra logic for sanity checking wired messages.
(-) lots of refactoring 
(-) extra copy when converting from wire format

This approach means that we have a separate wire format `messages.UntrustredEntity` and convert it to internal trusted format `flow.TrustedEntity` which is built from transforming untrusted data. 

2. Use value based structure for wire messages and leave our core data structures as pointer based but with extra data.
(+) save time on calculating ID, ability to extend trusted data with pre-computed values
(+) no need to add extra logic for sanity checking wired messages.
(+) little effort
(-) extra conversion logic with extra copy when converting from wire format

This approach means that we use a separate wire format same as 1) but convert it exactly the way it's implemented in this PR. We won't introduce breaking changing to our core data structures but it still leaves us room for adding pre-computed ID or other data. 

3. Use the same wire format as we have right now but add runtime sanity checks for data validity + update core data structures to have more data.
(+) little effort
(+) save time on calculating ID, ability to extend trusted data with pre-computed values
(-) constant overhead for runtime checks
(-) more stress on GC
(-) we do checks which could be done automatically by support of type system.

This approach means that we won't change wire messages, we will still use pointers all over the place and rely on runtime checks to assert data validity. 

TL; DR; There is no clear winner, all of proposed solutions have their cons and pros, I think we need to have more discussions to select a path which we will take forward and stop collecting technical dept.

engine/collection/compliance/core.go
```diff
@@ -81,17 +80,19 @@ func NewCore(
 
 // OnBlockProposal handles incoming block proposals.
 func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.ClusterBlockProposal) error {
-	header := proposal.Header
+	block := proposal.Block.ToInternal()
```
AlexHentschel, line 83: I am confused why this transformation `ToInternal()` is repeatedly happening while processing a block proposal: 
* https://github.com/onflow/flow-go/blob/4d1829104f468fdbe6a1399b457031006511a267/engine/collection/compliance/core.go#L199
* https://github.com/onflow/flow-go/blob/4d1829104f468fdbe6a1399b457031006511a267/engine/collection/compliance/core.go#L249

In my head, the conversion would be done once, which created certain guarantees in the resulting data structure. For example here, `ToInternal()` gives rise to the higher-level internal representation with certain guarantees, e.g. the relevant pointers are not nil. 

I would suggest to convert _once_ right here in the beginning of `OnBlockProposal`. (Probably beyond the scope of this PR, but that is where I would imagine the pre-computation of the ID would happen: also as part of the `ToInternal()` function.) If we want to work with `cluster.Block` from here on, we would need to update `PendingClusterBlockBuffer` to work with the internal representation instead, which I think would be worth it.

engine/common/follower/engine.go
```diff
@@ -221,12 +219,12 @@ func (e *Engine) onBlockResponse(originID flow.Identifier, res *messages.BlockRe
 
 // onBlockProposal handles incoming block proposals. inRangeBlockResponse will determine whether or not we should wait in processBlockAndDescendants
 func (e *Engine) onBlockProposal(originID flow.Identifier, proposal *messages.BlockProposal, inRangeBlockResponse bool) error {
+	block := proposal.Block.ToInternal()
```
AlexHentschel, line 222: Also here, it would be more consistent in my mind, to convert once and then work with `flow.Block` from here on

engine/consensus/compliance/core.go
```diff
@@ -93,21 +93,22 @@ func NewCore(
 
 // OnBlockProposal handles incoming block proposals.
 func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.BlockProposal, inBlockRangeResponse bool) error {
+	block := proposal.Block.ToInternal()
```
AlexHentschel, line 96: Would make the most sense to me, to only convert once here too (requiring changing the type of the pending block storage here as well)

model/messages/consensus.go
```diff
@@ -4,11 +4,132 @@ import (
...
+	}
+	return block
+}
+
+// BlockProposal is part of the consensus protocol and represents the leader
 // of a consensus round pushing a new proposal to the network.
 type BlockProposal struct {
-	Header  *flow.Header
-	Payload *flow.Payload
+	Block UntrustedBlock
```
AlexHentschel, line 126: Just for curiosity, would there be a disadvantage to writing
```golang
type BlockProposal struct {
	UntrustedBlock
}
```
?

AlexHentschel (approved):
Overall, I like the approach. It very nicely opens up the additional degree of freedom by decoupling the wire format from the internal representation. While I agree that it introduces some technical debt, I feel this PR is a good basis to allow further experimentation. I think blocks are a particularly good case for experimentation, because they (probably) have the most complex pipeline of verification.

engine/collection/compliance/core.go
jordanschalm: > In my head, the conversion would be done once, which created certain guarantees in the resulting data structure.

Good point. The multiple conversions (and usages) of the wire type is partly why I was unhappy with this approach - I think this will help to reduce the technical debt scope as well.

model/messages/consensus.go
jordanschalm: Not really, either seem OK to me.

jordanschalm:
* Wrote up our discussion from today [here](https://www.notion.so/dapperlabs/Message-validation-thoughts-and-proposal-ec9aee8ab6f949b8984cac5a42e235f1#ea70a7f6c4234ce083205b7db07f27b8)
* Updated this PR:
  * limited surface area of `messages.Block`-like wire types to one conversion per engine, removed from block buffer
  * removed `PendingBlock` types in favour of the `Slashable` interface Yurii introduced

module/buffer.go
```diff
@@ -3,19 +3,18 @@ package module
...
-	"github.com/onflow/flow-go/model/messages"
 )
 
 // PendingBlockBuffer defines an interface for a cache of pending blocks that
 // cannot yet be processed because they do not connect to the rest of the chain
 // state. They are indexed by parent ID to enable processing all of a parent's
 // children once the parent is received.
 type PendingBlockBuffer interface {
-	Add(originID flow.Identifier, proposal *messages.BlockProposal) bool
+	Add(originID flow.Identifier, block *flow.Block) bool
```
durkmurder, line 13: can we also accept `flow.Slashable` here?

module/buffer.go
```diff
@@ -3,19 +3,18 @@ package module
...
 	"github.com/onflow/flow-go/model/cluster"
 	"github.com/onflow/flow-go/model/flow"
-	"github.com/onflow/flow-go/model/messages"
 )
 
 // PendingBlockBuffer defines an interface for a cache of pending blocks that
 // cannot yet be processed because they do not connect to the rest of the chain
 // state. They are indexed by parent ID to enable processing all of a parent's
 // children once the parent is received.
 type PendingBlockBuffer interface {
```
durkmurder, line 12: We could also make this interface generic to work both with cluster blocks?

module/buffer.go
```diff
@@ -3,19 +3,18 @@ package module
...
-	Add(originID flow.Identifier, proposal *messages.BlockProposal) bool
+	Add(originID flow.Identifier, block *flow.Block) bool
 
-	ByID(blockID flow.Identifier) (*flow.PendingBlock, bool)
+	ByID(blockID flow.Identifier) (*flow.Slashable[flow.Block], bool)
 
-	ByParentID(parentID flow.Identifier) ([]*flow.PendingBlock, bool)
+	ByParentID(parentID flow.Identifier) ([]*flow.Slashable[flow.Block], bool)
```
durkmurder, line 17: I would strongly prefer to use `flow.Slashable` instead of `*flow.Slashable` here since it's a structure that just holds two pointers, no need to allocate it on heap

module/buffer.go
```diff
@@ -28,11 +27,11 @@ type PendingBlockBuffer interface {
 // PendingClusterBlockBuffer is the same thing as PendingBlockBuffer, but for
 // collection node cluster consensus.
 type PendingClusterBlockBuffer interface {
```
durkmurder, line 29: Same here, I have a feeling we can get rid of this interface by using generics.

utils/unittest/fixtures.go
```diff
@@ -198,28 +198,20 @@ func ProposalFixture() *messages.BlockProposal {
...
+	return messages.NewClusterBlockProposal(block)
 }
 
-func PendingFromBlock(block *flow.Block) *flow.PendingBlock {
-	pending := flow.PendingBlock{
-		OriginID: block.Header.ProposerID,
-		Header:   block.Header,
-		Payload:  block.Payload,
+// AsSlashable returns the input message T, wrapped as a flow.Slashable instance with a random origin ID.
+func AsSlashable[T any](msg *T) *flow.Slashable[T] {
```
durkmurder, line 209: Would strongly advocate for not using pointers for `flow.Slashable`

model/messages/collection.go
```diff
@@ -24,12 +24,70 @@ type CollectionResponse struct {
...
+	ReferenceBlockID flow.Identifier
+}
+
+// UntrustedClusterBlock is a duplicate of cluster.Block used within
+// untrusted messages. It exists only to provide a memory-safe structure for
+// decoding messages and should be replaced in the future by updating the core
+// cluster.Block type.
+// Deprecated: Please update cluster.Payload.Collection to use []flow.TransactionBody,
+// then replace instances of this type with cluster.Block
+type UntrustedClusterBlock struct {
```
durkmurder, line 44: Interested if we can use generics to deal with cluster blocks? Can we give it a try?

model/events/synchronization.go
```diff
@@ -1,16 +1,16 @@
...
 import (
-	"github.com/onflow/flow-go/model/cluster"
 	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/model/messages"
 )
 
 type SyncedBlock struct {
 	OriginID flow.Identifier
-	Block    *flow.Block
+	Block    messages.UntrustedBlock
```
durkmurder, line 10: Why are we using a value here? Have we considered using a pointer as before?

durkmurder (approved):
Looks great, I really like the changes. Left some comments for discussion but nothing critical

model/events/synchronization.go
jordanschalm: Because the structure is pretty small for copying, and we don't need reference semantics. I don't have a strong opinion though - do you think a pointer is better here?

model/messages/consensus.go
```diff
@@ -4,11 +4,132 @@ import (
...
+type UntrustedExecutionResult struct {
+	PreviousResultID flow.Identifier
+	BlockID          flow.Identifier
+	Chunks           []flow.Chunk
+	ServiceEvents    flow.ServiceEventList
+	ExecutionDataID  flow.Identifier
+}
+
+// ToInternal returns the internal representation of the type.
+func (ur UntrustedExecutionResult) ToInternal() *flow.ExecutionResult {
```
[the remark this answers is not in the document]
jordanschalm: https://github.com/onflow/flow-go/pull/3523/commits/0d4ecde53e46582fdbf081fe3b0e95ce8eeaab3a

module/buffer.go
jordanschalm: Agree - good point

model/messages/collection.go
jordanschalm: I did try this initially but didn't really find a simple way to apply generics here. It might be easier if `flow.Block` used generics too...

It isn't compiling but I'll throw it in `jordan/3374--generics-experimentation-mess` if you want to take a look at what I was trying. Maybe you have some different ideas for how to approach it.

module/buffer.go
jordanschalm: I have spent some time experimenting  with this - we can, but the exact approach has impacts beyond this module. I would prefer to spend some more time experimenting and choosing the right approach before merging anything. I've pushed all the WIP code to `jordan/3374--generics-experimentation-mess`.

jordanschalm:
bors merge

PR #3525, opened 2022-11-07
Prevent broadcasting unicast only messages

kc1116:
This PR adds a new field `AllowedPublish` to the message auth config and explicitly marks messages as authorized to be sent via publish. In the authorized sender validator, we reject any message that is intended to only be sent via unicast but is sent via publish.


network/message/authorization.go
```diff
@@ -15,6 +15,9 @@ type ChannelAuthConfig struct {
 
 	// AllowedUnicast indicates whether this message is allowed to be sent on the channel via unicast
 	AllowedUnicast bool
+
+	// AllowedPublish indicates whether this message is allowed to be sent on the channel via publish/multicast
+	AllowedPublish bool
```
peterargue, line 20: Looking over the message configs, only the test messages support both unicast and publish. Instead of 2 separate configs, how about we use a single one?

kc1116: Yes I noticed that as well, I intentionally added AllowedPublish so that it explicitly lets the developer know that some messages are allowed unicast/publish. If you think it's too verbose we can remove it. Without it, you cannot infer that a unicast message is not allowed to be published without going and looking at the validator code.

kc1116: https://github.com/onflow/flow-go/pull/3525/commits/116aaf544ad672ff1c22c1bea32248e47a8c6c61

network/message/authorization.go
```diff
@@ -64,10 +71,12 @@ func initializeMessageAuthConfigsMap() {
 			channels.ConsensusCommittee: {
 				AuthorizedRoles: flow.RoleList{flow.RoleConsensus},
 				AllowedUnicast:  false,
+				AllowedPublish:  true,
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/3525/commits/c756b342117fe434dc236cac63d08cae151162d5

network/message/init.go
```diff
@@ -1,9 +1,42 @@
...
+)
+
 // init is called first time this package is imported.
 // It creates and initializes AuthorizationConfigs for each message type.
 func init() {
 	initializeMessageAuthConfigsMap()
+	validateMessageAuthConfigsMap(validationExclusionList)
+}
+
+func validateMessageAuthConfigsMap(excludeList []string) {
```
peterargue, line 18: can you add a comment explaining why this check is here

kc1116: https://github.com/onflow/flow-go/pull/3525/commits/5c2ce17c2000894e614422ede5eaee8417b35918

network/message/authorization.go
```diff
@@ -7,16 +7,16 @@ import (
...
-var authorizationConfigs map[string]MsgAuthConfig
-
 type ChannelAuthConfig struct {
-	// AuthorizedRoles list of roles authorized to send this message on the channel
+	// AuthorizedRoles list of roles authorized to send this message on the channel.
 	AuthorizedRoles flow.RoleList
 
-	// AllowedUnicast indicates whether this message is allowed to be sent on the channel via unicast
-	AllowedUnicast bool
+	// AllowedProtocols list of protocols the message is allowed to be sent on.
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/3525/commits/5c2ce17c2000894e614422ede5eaee8417b35918

network/message/authorization.go
```diff
@@ -34,18 +35,18 @@ type MsgAuthConfig struct {
 // Expected error returns during normal operations:
 //   - ErrUnauthorizedMessageOnChannel: the channel is not included in the message's list of authorized channels
```
peterargue, line 36: please add the new broadcast error here too

peterargue (approved):
Looks good. Please update all of the godocs with the new error message as well

yhassanzadeh13 (approved):
Thanks for applying the comments!

network/message/authorization.go
kc1116: https://github.com/onflow/flow-go/pull/3525/commits/44f438d01acf94316e2f4b8cf9466a51a7301ead

PR #3526, opened 2022-11-07
[Networking] Add ability to change libp2p's logging level on the fly

SaveTheRbtz:
It's better to not couple golog and zerolog since they have different log levels and output formats.

### DIFF
--- a/admin/README.md
+++ b/admin/README.md
@@ -9,8 +9,16 @@ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"
 ```
 
 ### To change log level
+Flow, and other zerolog-based libraries:
+
 ```
 curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-log-level", "data": "debug"}'
+```
+
+libp2p, badger, and other golog-based libraries:
+
+```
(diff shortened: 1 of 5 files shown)


zhangchiqing (approved):

admin/README.md
```diff
@@ -9,10 +9,18 @@ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"
...
+Flow, and other zerolog-based libraries:
+
 ```
 curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-log-level", "data": "debug"}'
 ```
 
+libp2p, badger, and other golog-based libraries:
+
+```
+curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "set-golog-level", "data": "debug"}'
```
pattyshack, line 21: I have a really dumb question.  why are we doing this / why not just send a grpc request via a command line tool?  This seems like a poor man's rpc system ...

pattyshack (approved):

SaveTheRbtz:
bors merge

admin/README.md
SaveTheRbtz: @jordanschalm did we consider using grpc for the admin interface?

PR #3527, opened 2022-11-07
Move non address programs tracking to environment.

pattyshack:
Since we're generalizing BlockPrograms into a schema-ed database for storing all sorts of derived data, it no longer make sense to keep track of non address programs inside of the database.

gh: #3102

### DIFF
--- a/fvm/environment/programs.go
+++ b/fvm/environment/programs.go
@@ -15,10 +15,10 @@ import (
 )
 
 // TODO(patrick): remove and switch to *programs.TransactionPrograms once
-// emulator is updated.
+// https://github.com/onflow/flow-emulator/pull/229 is integrated.
 type TransactionPrograms interface {
-	Get(loc common.Location) (*interpreter.Program, *state.State, bool)
-	Set(loc common.Location, prog *interpreter.Program, state *state.State)
+	Get(loc common.AddressLocation) (*interpreter.Program, *state.State, bool)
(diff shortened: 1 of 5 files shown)


fvm/environment/programs.go
```diff
@@ -35,6 +35,10 @@ type Programs struct {
 	accounts Accounts
 
 	transactionPrograms TransactionPrograms
+
+	// NOTE: non-address programs are not reusable across transactions, hence
+	// they are kept out of the derived data database.
```
zhangchiqing, line 40: This comments makes me wonder what is the "derived data". Is that the `TransactionPrograms`?

Could you add comments to other fields to indicate what are derived, what are not.

zhangchiqing (approved):

fvm/environment/programs.go
pattyshack: TransactionPrograms is renamed to DerivedTransactionData in #3516.  The plan is to cache all pre-compute / derived data (e.g., programs, meter settings, protocol version, etc) in this.   Everything in this package will be derived from ledger data.

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors retry

pattyshack:
bors merge

PR #3529, opened 2022-11-08
[Execution] Do not depend on string conversion in register set tracking

SaveTheRbtz:
This PR:
* converts `map[string]flow.RegisterID` to `map[flow.RegisterID]struct{}` to represent sets.
* switches Delta from `map[string]flow.RegisterEntry` to `map[flow.RegisterID]flow.RegisterValue` to save a bit of space.

Currently these are responsible for ~15% of memory allocations:
Screen Shot 2022-11-07 at 9 50 17 PM

```
name                             old time/op    new time/op    delta
ComputeBlock/16/cols/128/txes-8     1.83s ± 3%     1.69s ± 2%   -7.59%  (p=0.000 n=10+10)

name                             old alloc/op   new alloc/op   delta
ComputeBlock/16/cols/128/txes-8    1.34GB ± 1%    1.21GB ± 2%   -9.38%  (p=0.000 n=9+10)

name                             old allocs/op  new allocs/op  delta
ComputeBlock/16/cols/128/txes-8     19.8M ± 0%     17.1M ± 0%  -13.48%  (p=0.000 n=8+8)
```

### DIFF
--- a/engine/execution/state/delta/delta.go
+++ b/engine/execution/state/delta/delta.go
@@ -10,19 +10,14 @@ import (
 
 // A Delta is a record of ledger mutations.
 type Delta struct {
-	Data map[string]flow.RegisterEntry
+	Data map[flow.RegisterID]flow.RegisterEntry
 }
 
 // NewDelta returns an empty ledger delta.
 func NewDelta() Delta {
 	return Delta{
-		Data: make(map[string]flow.RegisterEntry),
+		Data: make(map[flow.RegisterID]flow.RegisterEntry),
 	}
-}
-
-func toString(owner, key string) string {
-	register := flow.NewRegisterID(owner, key)
-	return register.String()
 }
 
 // Get reads a register value from this delta.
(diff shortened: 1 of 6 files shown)


model/flow/ledger.go
```diff
@@ -14,22 +14,11 @@ type RegisterID struct {
...
-	arr := make([]byte, requiredLen)
-
-	hex.Encode(arr, []byte(r.Owner))
-
-	arr[2*ownerLen] = byte('/')
-
-	hex.Encode(arr[(2*ownerLen)+1:], []byte(r.Key))
-
-	return string(arr)
+	return fmt.Sprintf("%x/%x", r.Owner, r.Key)
```
ramtinms, line 21: I'm not sure why we were doing it this way in the first place, I think it suppose to be some sort of optimized way of doing it

SaveTheRbtz: yeah. It is about 2x fast as the `Sprintf`.  The main problem here is that it was nod needed at all -- we were using this for map indexing, instead of just using the struct itself (which is infinitely faster and noes not require allocations.)

pattyshack (approved):

engine/execution/state/delta/delta.go
```diff
@@ -10,46 +10,37 @@ import (
...
 }
 
 // RegisterUpdates returns all registers that were updated by this delta.
 // ids are returned sorted, in ascending order
 func (d Delta) RegisterUpdates() ([]flow.RegisterID, []flow.RegisterValue) {
 	data := make(flow.RegisterEntries, 0, len(d.Data))
-
-	for _, v := range d.Data {
-		data = append(data, v)
+	for k, v := range d.Data {
```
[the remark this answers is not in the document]
ramtinms: I was going to suggest a similar thing, I think we don't need to sort based on both key and values since the `key` is used for the Data map is going to be anyway unique, sorting based on the key but appending the values while we are doing it through the sort method might be the more optimized approach here. I see with the current code you were trying to maintain the same logic but I think in this case is fine to change the sort to be based on the key, considering the `less` implementation of the KeyEntry is only limited to keys.

model/flow/ledger.go
ramtinms: got it, then all good.

ramtinms (approved):
Nice find, looks good to me, I think the original definition of Data struct was because of a legacy structure of Delta and it makes a lot of sense to optimize it this way. 

Just put a minor suggestion

SaveTheRbtz:
Ok. I'll land this as is to minimize the updates and followup with the PR to simplify `RegisterUpdates`.

PR #3530, opened 2022-11-08
Refactor fetch collections

zhangchiqing:
The `matchOrRequestCollection` hits a lock in requester, which might cause it to hold the lock in requester for too long. This PR moves the side effect (requesting collections) out of the function that holds the lock in order to reduce concurrent tension.

### DIFF
--- a/engine/common/requester/engine.go
+++ b/engine/common/requester/engine.go
@@ -35,19 +35,22 @@ type CreateFunc func() flow.Entity
 // on the flow network. It is the `request` part of the request-reply
 // pattern provided by the pair of generic exchange engines.
 type Engine struct {
-	unit          *engine.Unit
-	log           zerolog.Logger
-	cfg           Config
-	metrics       module.EngineMetrics
-	me            module.Local
-	state         protocol.State
-	con           network.Conduit
-	channel       channels.Channel
-	selector      flow.IdentityFilter
-	create        CreateFunc
-	handle        HandleFunc
-	items         map[flow.Identifier]*Item
-	requests      map[uint64]*messages.EntityRequest
+	unit     *engine.Unit
+	log      zerolog.Logger
+	cfg      Config
+	metrics  module.EngineMetrics
+	me       module.Local
+	state    protocol.State
+	con      network.Conduit
+	channel  channels.Channel
+	selector flow.IdentityFilter
+	create   CreateFunc
+	handle   HandleFunc
+
+	// changing the following state variables must be guarded by unit.Lock()
+	items    map[flow.Identifier]*Item
+	requests map[uint64]*messages.EntityRequest
+
 	isDispatching *atomic.Bool // to ensure only trigger dispatching logic once at any time
 }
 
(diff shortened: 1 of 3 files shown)


engine/common/requester/engine.go
```diff
@@ -435,9 +435,6 @@ func (e *Engine) process(originID flow.Identifier, message interface{}) error {
 	e.metrics.MessageReceived(e.channel.String(), metrics.MessageEntityResponse)
 	defer e.metrics.MessageHandled(e.channel.String(), metrics.MessageEntityResponse)
 
-	e.unit.Lock()
```
zhangchiqing, line 438: We don't have to lock it so early here

engine/common/requester/engine.go
```diff
@@ -474,6 +471,9 @@ func (e *Engine) onEntityResponse(originID flow.Identifier, res *messages.Entity
 			Msg("onEntityResponse entries received")
 	}
 
+	e.unit.Lock()
```
zhangchiqing, line 474: Move it down here, so that we don't lock the engine, when querying Identities with `e.state.Final().Identities`

engine/execution/ingestion/engine.go
```diff
@@ -404,12 +404,14 @@ func (e *Engine) reloadBlock(
...
 	}
 
-	err = e.enqueueBlockAndCheckExecutable(blockByCollection, executionQueues, block, false)
+	missingCollections, err := e.enqueueBlockAndCheckExecutable(blockByCollection, executionQueues, block, false)
 
 	if err != nil {
 		return fmt.Errorf("could not enqueue block %x on reloading: %w", blockID, err)
 	}
 
+	e.fetchCollections(block.Header, missingCollections)
```
zhangchiqing, line 413: requesting collections is now moved out of the mempool.Run lock

pattyshack (commented):
could you add comments to the struct/field definitions indicated the fields thread safety?  e.g., which fields are supposed to be guarded by unit's mutex.

pattyshack (approved):

storage/badger/commits.go
```diff
@@ -36,7 +36,7 @@ func NewCommits(collector module.CacheMetrics, db *badger.DB) *Commits {
 	c := &Commits{
 		db: db,
 		cache: newCache(collector, metrics.ResourceCommit,
-			withLimit(100),
+			withLimit(1000),
```
zhangchiqing, line 39: The origin limit 100 is a bit small

engine/execution/ingestion/engine.go
```diff
@@ -980,13 +990,14 @@ func (e *Engine) matchOrRequestCollections(
...
-	}
+	// // make sure that the requests are dispatched immediately by the requester
+	// if len(executableBlock.Block.Payload.Guarantees) > 0 {
+	// 	defer e.request.Force()
+	// 	defer e.metrics.ExecutionCollectionRequestSent()
+	// }
+	//
 
-	actualRequested := 0
+	missingCollections := make([]*flow.CollectionGuarantee, 0, len(executableBlock.Block.Payload.Guarantees))
```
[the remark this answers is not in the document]
zhangchiqing: I didn't change the overall behavior tho.

If the unittest is to check whether we will fetch those missing collections, the existing `TestBlocksArentExecutedMultipleTimes_collectionArrival` test case has covered that.

PR #3531, opened 2022-11-08
[FVM] Threadsafe event encoder

SaveTheRbtz:


pattyshack (approved):

zhangchiqing (approved):
Good catch 👍 

I approve because I understand your changes and reasoning. It makes sense to me. 

I also wonder why it's not initially implemented this simple way. Maybe it's worth double checking with original code owner, in case there is any special scenario.

SaveTheRbtz:
The original reasoning was to re-use the encoder and the buffer to improve performance.  There are two solutions for the concurrency problem: a lock or allocations.  I went with a simpler approach.

SaveTheRbtz:
bors merge

PR #3546, opened 2022-11-09
[Consensus] Switch to generic binary search

SaveTheRbtz:
Let's not support a custom `bsearch(3)` implementation.  As a side benefit new generic-based version is slightly faster:
```
$ benchstat old.txt new.txt                                                                                           
name               old time/op    new time/op    delta
LeaderSelection-8     520ms ± 0%     508ms ± 1%  -2.27%  (p=0.000 n=9+10)

name               old alloc/op   new alloc/op   delta
LeaderSelection-8    30.0MB ± 0%    30.0MB ± 0%    ~     (p=0.797 n=9+10)

name               old allocs/op  new allocs/op  delta
LeaderSelection-8      99.0 ± 0%      99.0 ± 0%    ~     (all equal)
```

### DIFF
--- a/consensus/hotstuff/committees/leader/leader_selection.go
+++ b/consensus/hotstuff/committees/leader/leader_selection.go
@@ -157,11 +157,11 @@ func weightedRandomSelection(
 
 	leaders := make([]uint16, 0, count)
 	for i := 0; i < count; i++ {
-		// pick a random number from 0 (inclusive) to cumsum (exclusive). Or [0, cumsum)
-		randomness := rng.UintN(cumsum)
+		// pick a random number [1, cumsum]
+		randomness := rng.UintN(cumsum) + 1
 
 		// binary search to find the leader index by the random number
-		leader, _ := slices.BinarySearch(weightSums, randomness+1)
+		leader, _ := slices.BinarySearch(weightSums, randomness)
 		leaders = append(leaders, uint16(leader))
 	}
 	return leaders, nil
--- a/consensus/hotstuff/committees/leader/leader_selection.go
+++ b/consensus/hotstuff/committees/leader/leader_selection.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/onflow/flow-go/crypto/random"
 	"github.com/onflow/flow-go/model/flow"
 )
@@ -159,35 +161,8 @@ func weightedRandomSelection(
 		randomness := rng.UintN(cumsum)
 
 		// binary search to find the leader index by the random number
-		leader := binarySearchStrictlyBigger(randomness, weightSums)
-
+		leader, _ := slices.BinarySearch(weightSums, randomness+1)
 		leaders = append(leaders, uint16(leader))
 	}
 	return leaders, nil
-}
-
-// binarySearchStriclyBigger finds the index of the first item in the given array that is
-// strictly bigger to the given value.
-// There are a few assumptions on inputs:
-// - `arr` must be non-empty
-// - items in `arr` must be in non-decreasing order
-// - `value` must be less than the last item in `arr`
-func binarySearchStrictlyBigger(value uint64, arr []uint64) int {
-	left := 0
-	arrayLen := len(arr)
-	right := arrayLen - 1
-	mid := arrayLen >> 1
-	for {
-		if arr[mid] <= value {
-			left = mid + 1
-		} else {
-			right = mid
-		}
-
-		if left >= right {
-			return left
-		}
-
-		mid = int(left+right) >> 1
-	}
 }
--- a/consensus/hotstuff/committees/leader/leader_selection_test.go
+++ b/consensus/hotstuff/committees/leader/leader_selection_test.go
@@ -3,7 +3,6 @@ package leader
 import (
 	"fmt"
 	"math/rand"
-	"sort"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


consensus/hotstuff/committees/leader/leader_selection.go
```diff
@@ -155,39 +157,12 @@ func weightedRandomSelection(
...
 	for i := 0; i < count; i++ {
-		// pick a random number from 0 (inclusive) to cumsum (exclusive). Or [0, cumsum)
-		randomness := rng.UintN(cumsum)
+		// pick a random number [1, cumsum]
+		randomness := rng.UintN(cumsum) + 1
 
 		// binary search to find the leader index by the random number
-		leader := binarySearchStrictlyBigger(randomness, weightSums)
-
+		leader, _ := slices.BinarySearch(weightSums, randomness)
```
AlexHentschel, line 164: For correctness, we depend here on a subtle property
* we must find the element with the _smallest_ index that is equal or larger than the given value `randomness` 

Note that this property is **not** guaranteed by [`slices.BinarySearch`](https://pkg.go.dev/golang.org/x/exp/slices#BinarySearch) in the case of equality, according to the documentation. For example, consider the slice `weightSums = [0, 0, 2, 3, 3, 3, 8, 16, 16]` and assume we have `randomness = 3`. An index satisfying the API contract would be `3` but also `4` or `5`. However,  in our context `4` or `5` would be invalid.

Nevertheless, the [_implementation_](https://cs.opensource.google/go/x/exp/+/9ce248df:slices/sort.go;l=65-67)  satisfies our stronger requirements:
```golang
// search returns the leftmost position where f returns true, or len(x) if f
// returns false for all x. This is the insertion position for target in x,
// and could point to an element that's either == target or not.
```

In other words, we are relying on non-documented, internal implementation details here, which is not great. 
I see two options:
1. add detailed documentation here, explaining the additional requirement _and_ add solid test coverage to ensure that we would detect any future changes in `slices.BinarySearch` that violate our property. 
2. Keep the existing code: it already has solid test coverage and good documentation. There is no speed disadvantage compared  to `slices.BinarySearch` because it is literally the same implementation.

AlexHentschel (changes requested):

consensus/hotstuff/committees/leader/leader_selection.go
AlexHentschel: This statement is not that convincing to me:
> As a side benefit new generic-based version is slightly faster

If you take a look at [the implementation](https://cs.opensource.google/go/x/exp/+/9ce248df:slices/sort.go;l=90-105), it is almost identical same. 2.2% speedup are marginal and my assumption is that you are seeing noise as opposed to a real speedup.

AlexHentschel:
How about the following modifications:
1. we include the following additional documentation [here](https://github.com/onflow/flow-go/blob/cf8364cec57bb56561b57de35949ec953323aed9/consensus/hotstuff/committees/leader/leader_selection.go#L164)
    ```golang
    // CAUTION: we must find the element with the **smallest index**, whose value is
    // equal or larger than `randomness`. For example, consider the slice 
    //   weightSums = [0, 0, 2, 5, 5, 8, 16, 16]
    // and assume we have randomness = 5. For correctness of the leader selection, we
    // require that index `3` is returned. However, at the time of writing this code, 
    // the _implementation_ of `slices.BinarySearch` satisfies our requirement to find
    // the smallest index. But the API documentation does not guarantee this property. 
    // In other words, we are relying on non-documented, internal implementation details
    // of `slices.BinarySearch`. We verify that this critical property is retained by
    // future versions of golang through the unit test Test_SmallestIndexReturned.
    ```
2. include the following fuzzy test:
    ```golang
    // Test_SmallestIndexReturned verifies that `slices.BinarySearch` satisfies the following property:
    //
    //	`slices.BinarySearch` returns the **smallest index**, whose value is equal or
    //	 larger than `target`
    //
    // This property is critical for the correctness of our leader selection. For example, consider the slice
    //
    //	weightSums = [0, 0, 2, 5, 5, 8, 16, 16]
    //
    // and assume we have randomness = 5. For correctness of the leader selection, we
    // require that index `3` is returned. However, at the time of writing this code,
    // the _implementation_ of `slices.BinarySearch` satisfies our requirement to find
    // the smallest index. But the API documentation does not guarantee this property.
    // In other words, we are relying on non-documented, internal implementation details
    // of `slices.BinarySearch`. We verify that this critical property is retained by
    // future versions of golang through this unit test.
    func Test_SmallestIndexReturned(t *testing.T) {
    	for sliceLength := 1; sliceLength <= 123; sliceLength++ {
    		for repetition := 10; repetition > 0; repetition-- {
    			firstIndex := rand.Intn(sliceLength)
    			weightSums := make([]uint64, sliceLength)
    			for k := 0; k < firstIndex; k++ {
    				weightSums[k] = 7
    			}
    			for k := firstIndex; k < sliceLength; k++ {
    				weightSums[k] = 13
    			}

    			i, found := slices.BinarySearch(weightSums, 13)
    			require.True(t, found)
    			require.Equal(t, firstIndex, i)
    		}
    	}
    }
    ```

consensus/hotstuff/committees/leader/leader_selection.go
SaveTheRbtz: The main goal of the diff to decrease the complexity.  I've verified that implementation does return the leftmost element, but given that we are considering that this property can change, then the only sustainable way is to maintain our own fork/reimplementation.

Issue #3548, opened 2022-11-09
Automated Performance Tests


SaveTheRbtz:
## Goal
We want to have a standardized automated set of performance benchmarks. 

## Milestones
- [x] Add ability to test different commits and change the BigQuery schema accordingly: https://github.com/onflow/flow-go/pull/3633
- [x] Recreate dashboards for TPS tests utilizing new schema and split into two:
   - [x] Raw data for one-off tests and overall system monitoring: https://dapperlabs.grafana.net/d/BLSHi8ZVz/flow-performance-benchmarks?orgId=1&from=now-90d&to=now&var-branch=All&var-env=All&var-type=All
   - [x] Aggregated data for performance analysis: https://dapperlabs.grafana.net/d/mA42tHK4z/flow-performance-benchmarks-raw?orgId=1&from=now-24h&to=now&var-branch=All&var-env=All&var-type=All
- [x] Improve error handling in benchmark code: https://github.com/onflow/flow-go/pull/3720
- [x] (optional) speedup test setup: https://github.com/onflow/flow-go/pull/3641 
- [ ] Add gRPC API to the `benchmark/ci` and make it a daemon.
- [ ] Add CLI interface to the daemon so we can control it manually.
- [ ] Move control code out of dataplane an use gRPC interface to drive tests.
- [ ] Add a CLI to control the scheduling (e.g. submit ad-hoc tests)

Full plan here: https://www.notion.so/dapperlabs/TPS-Benchmark-Plan-36e2109bfe7c4f13946c7671217721bc

## Previous art
* Vitess nightly: https://benchmark.vitess.io/macrobench
* Rust Compiler Nightly: https://perf.rust-lang.org/

PR #3551, opened 2022-11-09
[Benchmark] Rewrite worker stats

SaveTheRbtz:
This PR represents the long standing cleanup for the benchmark binary.  Mostly the goal was to remove the annoying `github.com/jedib0t/go-pretty/table` print with a structured log.

* Decouple worker stats collection and printing.
* Decouple worker stats collection and formatting (remove prettytable)
* Replace the leaking precise TPS calculation with a streaming EWMA.


integration/benchmark/worker_stats_tracker_test.go
```diff
@@ -1,25 +1,30 @@
...
 	st.AddWorkers(1)
 
-	startTime := time.Now()
-	endTime := startTime.Add(time.Second)
-	assert.Equal(t, 0., st.AvgTPSBetween(startTime, endTime))
-	st.AddTxSent()
-	st.AddTxSent()
-	assert.Equal(t, 2., st.AvgTPSBetween(startTime, endTime))
+	stats := st.getStats()
+	assert.Equal(t, 0, stats.txsSent)
```
gomisha, line 17: Shouldn't this test use `require.Equal()` instead of `assert.Equal()`? Otherwise the test won't fail if the values don't match.

integration/benchmark/worker_stats_tracker.go
```diff
@@ -1,48 +1,83 @@
...
 }
 
-// StopPrinting stops reporting of worker stats
-func (st *WorkerStatsTracker) StopPrinting() {
-	st.printer.Stop()
+func (st *WorkerStatsTracker) updateEWMAonce(lastStats, stats workerStats) {
+	st.mux.Lock()
+	defer st.mux.Unlock()
+
+	st.txsSentEWMA.Add(float64(stats.txsSent - lastStats.txsSent))
```
gomisha, line 72: can you add comments here explaining what's being tracked? Looks like we're keeping track of the transactions sent since the last time

integration/benchmark/worker_stats_tracker_test.go
SaveTheRbtz: It should fail at the end, but will continue to the end instead of failing fast.

integration/benchmark/worker_stats_tracker.go
```diff
@@ -1,48 +1,84 @@
...
 
-type WorkerStats struct {
-	workers     int
-	txsSent     int
-	txsExecuted int
+type workerStats struct {
+	workers         int
+	txsSent         int
+	txsExecuted     int
+	txsSentEWMA     float64
```
pattyshack, line 16: ewma is not meaningful acronym for most ppl.  use a more descriptive abbreviation or type out the full name

integration/benchmark/worker_stats_tracker.go
```diff
@@ -81,47 +114,26 @@ func (st *WorkerStatsTracker) GetTxSent() int {
...
-
-func (st *WorkerStatsTracker) Digest() string {
-	t := table.NewWriter()
-	t.AppendHeader(table.Row{
-		"workers",
-		"total TXs sent",
-		"total TXs finished",
-		"Avg TPS (last 10s)",
+func NewPeriodicStatsLogger(st *WorkerStatsTracker, log zerolog.Logger) *Worker {
+	w := NewWorker(0, 1*time.Second, func(workerID int) {
```
pattyshack, line 127: aside: use the main's ctx rather than Background to create the worker's internal ctx.

pattyshack (approved):

gomisha (approved):

SaveTheRbtz:
bors merge

PR #3553, opened 2022-11-10
Fix ingestion engine

zhangchiqing:
This PR fixes a bug that some collections were not requested by ingestion


SaveTheRbtz (approved):

m4ksio (approved):
Approving, but as I mention in a previous PR in this area, we need to have tests for this before merging to master

pattyshack (approved):

PR #3554, opened 2022-11-10
[CODEOWNERS] Update ownership

SaveTheRbtz:
In this diff:

* cherry-pick 7d1b1bf170f18d477d5f1c95bef77d4107095382 from the active-pacemaker branch,
* Add @gomisha as a reviewers for `/integration/benchmark`,
* Add "Dev Tools Stream",
* Remove @zhangchiqing from `localnet` to reduce spam,
* Remove @Kay-Zee to reduce spam,
* Remove ex-team members,
* Remove `/utils/binstat` since it is unsupported.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -36,10 +36,15 @@
 # Bootstrap and transit scripts
 /cmd/bootstrap/** @vishalchangrani
 
+# Dev Tools Stream
+.github/workflows/** @gomisha
+/insecure/** @gomisha @yhassanzadeh13
+/integration/benchnet2/** @gomisha
+/tools/test_monitor/** @gomisha
+
 # Performance Stream
-/integration/benchmark/** @SaveTheRbtz @Kay-Zee @AndrewM-SDET
-/integration/localnet/** @SaveTheRbtz @Kay-Zee @zhangchiqing @AndrewM-SDET
-/module/profiler/** @SaveTheRbtz @Kay-Zee @pattyshack
-/module/trace/** @SaveTheRbtz @Kay-Zee @pattyshack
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -36,10 +36,12 @@
 # Bootstrap and transit scripts
 /cmd/bootstrap/** @vishalchangrani
 
+# Automation Team
+/integration/** @gomisha
```
gomisha, line 40: can you please add the following directories under `Dev Tools Stream` and replace `Automation Team`:

`/insecure/**` @gomisha @yhassanzadeh13 
`/integration/benchnet2` @gomisha 
`/tools/test_monitor` @gomisha 
`.github/workflows` @gomisha 

Also, I think the `/integration/**` directory is too large to go all under me. Can we remove that and use sub-directories of `/integration/` to assign to different teams?

gomisha:
I also think we should add @durkmurder to `# Consensus Stream`

cc @durkmurder

durkmurder:
> I also think we should add @durkmurder to `# Consensus Stream`
> 
> cc @durkmurder

I am already added to consensus stream on [feature branch](https://github.com/onflow/flow-go/blob/feature/active-pacemaker/CODEOWNERS) but it will take some time to merge it so you could replicate changes from that branch. Thanks

SaveTheRbtz:
@jordanschalm @durkmurder i've cherry-picked 7d1b1bf170f18d477d5f1c95bef77d4107095382 so it would merge clearly in the future.  Thanks for fixing that!

zhangchiqing (approved):

SaveTheRbtz:
bors merge

Issue #3455, opened 2022-10-26
EN1 periodic execution stalls


SaveTheRbtz:
We have an issue with EN1 periodically stalling due to collection unavailability.  The root cause for the collection node connectivity problems is unknown but regardless we should be robust against collection node failures.

Solution with more aggressive retries (https://github.com/onflow/flow-go/pull/3406) fixed the stalling issue but overloaded network queues impacting liveness.

zhangchiqing:
Log shows sometimes EN1 received collection much later than other ENs, for instance:
For collection `19c7d5dacb4527ff0d42020703a40bfa7bce86f1665efd58ff30ae7c6b7b69f5`, EN1 received it about 8 seconds later than other ENs:
```
{
  '19c7d5dacb4527ff0d42020703a40bfa7bce86f1665efd58ff30ae7c6b7b69f5': {
    '9686399a8a5418a12e762cfaeff2ea348c2137f554560917760e0d47acf2cda4': '2022-11-15 10:07:00',
    '4ab025ab974e7ad7f344fbd16e5fbcb17fb8769fc8849b9d241ae518787695bd': '2022-11-15 10:06:52',
    '160241f88cbfaa0f361cf64adb0a1c9fc19dec1daf4b96550cd67b7a9fb26cd9': '2022-11-15 10:06:52',
    '0ca407c1da940952ebcc02283b60cd97c9a008e111a48ea6cf1ce8f36f1e0153': '2022-11-15 10:06:52'
  },
```
But checking the log from collection nodes, collection node actually received request to that collection twice from EN1:
```
2022-11-15 10:07:00	
{"level":"debug","node_role":"collection","node_id":"cae791d22cb582210b9c9a38b3e590603c50141099747d8c3cdbe4eeb085a84b","engine":"provider","nonce":8871037616407288599,"origin_id":"9686399a8a5418a12e762cfaeff2ea348c2137f554560917760e0d47acf2cda4","entity_ids":["19c7d5dacb4527ff0d42020703a40bfa7bce86f1665efd58ff30ae7c6b7b69f5"],"time":"2022-11-15T18:07:00.619190346Z","message":"entity request received"}
2022-11-15 10:06:52	
[...]

PR #3561, opened 2022-11-12
[Profiler] Add admin commands to control profiling rates

SaveTheRbtz:
This PR adds following admin commands:
* profiler-set-mem-profile-rate
* profiler-set-block-profile-rate
* profiler-set-mutex-profile-fraction (really want to rename that to "rate", but kept for consistency w/ golang naming)

While here, also remove the `-profiler-mem-profile-rate` command line argument and `/debug/pprof` interface from the metrics port.

Ref: https://github.com/onflow/flow-go/issues/3455

### DIFF
--- a/cmd/node_builder.go
+++ b/cmd/node_builder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 	"path/filepath"
-	"runtime"
 	"time"
 
 	"github.com/dgraph-io/badger/v2"
@@ -21,6 +20,7 @@ import (
 	"github.com/onflow/flow-go/module/chainsync"
 	"github.com/onflow/flow-go/module/compliance"
 	"github.com/onflow/flow-go/module/component"
+	"github.com/onflow/flow-go/module/profiler"
 	"github.com/onflow/flow-go/module/updatable_configs"
 	"github.com/onflow/flow-go/network"
 	"github.com/onflow/flow-go/network/codec/cbor"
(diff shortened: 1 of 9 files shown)


cmd/node_builder.go
```diff
@@ -289,18 +284,21 @@ func DefaultBaseConfig() *BaseConfig {
...
-		receiptsCacheSize:      bstorage.DefaultCacheSize,
-		guaranteesCacheSize:    bstorage.DefaultCacheSize,
+		metricsPort:         8080,
+		tracerEnabled:       false,
+		tracerSensitivity:   4,
+		MetricsEnabled:      true,
+		receiptsCacheSize:   bstorage.DefaultCacheSize,
+		guaranteesCacheSize: bstorage.DefaultCacheSize,
+
+		profilerConfig: profiler.ProfilerConfig{
```
zhangchiqing, line 294: Nice grouping them 👍

cmd/scaffold.go
```diff
@@ -688,36 +687,45 @@ func (fnb *FlowNodeBuilder) createProfileUploader() (profiler.Uploader, error) {
...
+		func(d time.Duration) error { return profiler.TriggerRun(d) })
 	fnb.MustNot(err).Msg("could not register profiler-trigger config")
 
+	fnb.MustNot(
+		fnb.ConfigManager.RegisterUintConfig(
+			"profiler-set-mem-profile-rate",
+			func() uint { return uint(runtime.MemProfileRate) },
+			func(r uint) error { runtime.MemProfileRate = int(r); return nil }),
+	).Msg("could not register profiler setting")
+	// There is no way to get the current block progile rate so we keep track of it ourselves.
```
zhangchiqing, line 714:
```suggestion
	// There is no way to get the current block profile rate so we keep track of it ourselves.
```

zhangchiqing (approved):

PR #3563, opened 2022-11-14
Add GetOrCompute to generic derived data implementation.

pattyshack:
The Get and Set api are too low-level for general usage.  We need to expose Get/Set to accommodate cadence's runtime interface, but we can use a higher level api for other derived data.

gh: #3102


SaveTheRbtz (approved):

fvm/programs/block_derived_data.go
```diff
@@ -350,6 +352,55 @@ func (txn *TransactionDerivedData[TKey, TVal]) Set(
...
+// then the pre-computed value is returned and the cached state is replayed on
+// txnState.  Otherwise, the value is computed using valFunc; both the value
+// and the states used to compute the value are captured.
+//
+// Note: valFunc must be an idempotent function and it must not modify
+// txnState's values.
+func (txn *TransactionDerivedData[TKey, TVal]) GetOrCompute(
+	txnState *state.TransactionState,
+	key TKey,
+	valFunc func() (TVal, error),
```
janezpodhostnik, line 365: Just wondering if it would be useful for valFunc to take a TKey parameter.

janezpodhostnik (approved):

PR #3566, opened 2022-11-15
[Benchmark] Notify on slow transactions

SaveTheRbtz:
Sometimes at high TPS we never receive a notification from a transaction being executed/sealed.  This leads to feedback loop stalling since it waits for ongoing transactions to finish.

While we are figuring out where and why this happens let's unblock the benchmarking by proceeding and logging such events.  Later we can cross-correlate logs for these transaction IDs.

### DIFF
--- a/integration/benchmark/contLoadGenerator.go
+++ b/integration/benchmark/contLoadGenerator.go
@@ -31,7 +31,7 @@ const (
 	ConstExecCostLoadType LoadType = "const-exec" // for an empty transactions with various tx arguments
 )
 
-const slowTransactionThreshold = 30 * time.Second
+const lostTransactionThreshold = 90 * time.Second
 
 var accountCreationBatchSize = 750 // a higher number would hit max gRPC message size
 const tokensPerTransfer = 0.01     // flow testnets only have 10e6 total supply, so we choose a small amount here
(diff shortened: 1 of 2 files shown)


gomisha (approved):

integration/benchmark/contLoadGenerator.go
```diff
@@ -719,26 +719,25 @@ func (lg *ContLoadGenerator) sendTokenTransferTx(workerID int) {
...
-			return
-		case <-time.After(slowTransactionThreshold):
-			// TODO(rbtz): add a counter for slow transactions
-			log.Trace().
-				Hex("txID", transferTx.ID().Bytes()).
-				Dur("duration", time.Since(startTime)).
-				Int("availableAccounts", len(lg.availableAccounts)).
-				Msg("is taking too long")
-		}
+	t := time.NewTimer(lostTransactionThreshold)
```
pattyshack, line 725: just use time.After?  you're not leaking anything

pattyshack (approved):

integration/benchmark/contLoadGenerator.go
SaveTheRbtz: Timeout for the After is quite big, so I've decided to release resources immediately if possible.

SaveTheRbtz:
bors merge

PR #3574, opened 2022-11-15
[Benchmark] Speedup benchmark convergence

SaveTheRbtz:
This speeds up the convergence to maximum sustainable TPS (`additiveIncrease` increase) and reduces variance of the results (`multiplicativeDecrease` increase.)

Also reduce timeout on `pollCollections` to 1s: a value that is closer to our block generation time.

### DIFF
--- a/integration/benchmark/cmd/ci/main.go
+++ b/integration/benchmark/cmd/ci/main.go
@@ -59,8 +59,8 @@ const (
 	serviceAccountPrivateKeyHex = unittest.ServiceAccountPrivateKeyHex
 
 	// Auto TPS scaling constants
-	additiveIncrease       = 50
-	multiplicativeDecrease = 0.8
+	additiveIncrease       = 100
+	multiplicativeDecrease = 0.9
 	adjustInterval         = 20 * time.Second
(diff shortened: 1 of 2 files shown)


integration/benchmark/cmd/ci/main.go
```diff
@@ -59,8 +59,8 @@ const (
 	serviceAccountPrivateKeyHex = unittest.ServiceAccountPrivateKeyHex
 
 	// Auto TPS scaling constants
-	additiveIncrease       = 50
-	multiplicativeDecrease = 0.8
+	additiveIncrease       = 100
+	multiplicativeDecrease = 0.9
```
SaveTheRbtz, line 63: @pattyshack you were right!  0.9 looks way more "gentle" on the graphs!

pattyshack: it's well known that tcp backoff factor is too aggressive.

pattyshack (approved):

gomisha (approved):

SaveTheRbtz:
bors merge

PR #3576, opened 2022-11-15
Enable nested transaction to use a different set of meter parameters

pattyshack:
This allows nested transaction to use a different set of meter parameters that are not inherited from the parent transaction, and allows us to meter different parts of the transaction using different parameters. (See discussion in #2922 for additional context)

gh: #3102


SaveTheRbtz (approved):

janezpodhostnik (approved):
nice

pattyshack:
bors merge

PR #3578, opened 2022-11-15
[Benchmark] Throttle on timeout

SaveTheRbtz:
Each time we encounter a timeout we want to throttle unconditionally.

### DIFF
--- a/integration/benchmark/cmd/ci/main.go
+++ b/integration/benchmark/cmd/ci/main.go
@@ -316,6 +316,7 @@ func adjustTPS(
 	lastTs := time.Now()
 	lastTPS := float64(minTPS)
 	lastTxs := uint(lg.GetTxExecuted())
+	lastTimedoutTxs := lg.GetTxTimedout()
 	for {
 		select {
 		// NOTE: not using a ticker here since adjusting worker count in SetTPS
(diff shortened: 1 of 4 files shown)


pattyshack (approved):
maybe allow downward (but not upward) adjustment on timeout

integration/benchmark/cmd/ci/main.go
```diff
@@ -400,6 +407,10 @@ func computeTPS(
 	currentTPS := float64(currentTxs-lastTxs) / timeDiff
 	unboundedTPS := uint(math.Ceil(currentTPS))
 
+	if timedout {
```
gomisha, line 410: can you add a comment here explaining what is happening when there's a timeout?

integration/benchmark/cmd/ci/main.go
```diff
@@ -316,13 +316,16 @@ func adjustTPS(
...
 	lastTxs := uint(lg.GetTxExecuted())
+	lastTimedoutTxs := lg.GetTxTimedout()
 	for {
 		select {
 		// NOTE: not using a ticker here since adjusting worker count in SetTPS
 		// can take a while and lead to uneven feedback intervals.
 		case nowTs := <-time.After(interval):
 			currentSentTxs := lg.GetTxSent()
 			currentTxs := uint(lg.GetTxExecuted())
+			currentTimedoutTxs := lg.GetTxTimedout()
```
gomisha, line 327: can you add descriptions here explaining what the vars represent? For example, what's the difference between `currentTxs` and `currentTimedoutTxs`?

gomisha (approved):

SaveTheRbtz:
> maybe allow downward (but not upward) adjustment on timeout

I've looked into it, but state machine becomes too complicated to read so I went with a simpler approach.

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

PR #3585, opened 2022-11-17
[Networking][BFT Testing] abstracts GossipSub-related primitives

yhassanzadeh13:
This PR lays the foundation for testing the Flow network against GossipSub-related attacks. In this PR, we abstract the `pubsub`-related dependencies into `interfaces` and adapters, which allows us to run tests with different implementations of GossipSub. 

For sake of BFT testing of GossipSub-related attacks, we are using forked versions of `pubsub`, and  this refactoring is needed because in order to use the forked `pubsub` module, we need to use the entire dependency tree of the forked module which is resolved to a different go module. Without this refactoring, we cannot use the original libp2p pubsub module in the same package.

It also improves the general abstraction of the codebase around libp2p `pubsub` by reducing the coupling to the specific `go-libp2p-pubsub` dependencies. 

In specific, the coupling side benefits introduced by this PR are: 
- **Degree Reduction**: connections between the Flow codebase and `pubsub` are made through narrow interfaces with few parameters. 
- **Ease**: connections among `pubsub` interfaces are made with the minimum possible understanding of implementing other modules. 
- **Flexibility**: `pubsub` modules are easily replaceable for something better in the future than being tightly coupled to `go-libp2p-pubsub`. 

The core contribution of this PR and the best place to start the review is: https://github.com/onflow/flow-go/blob/f41885295ebb5930a23996633ec8b1ebfe6aaaf7/network/p2p/pubsub.go

ref: https://github.com/dapperlabs/flow-go/issues/6423


network/internal/p2pfixtures/fixtures.go
```diff
@@ -10,14 +10,10 @@ import (
 	"time"
 
 	addrutil "github.com/libp2p/go-addr-util"
-	dht "github.com/libp2p/go-libp2p-kad-dht"
```
yhassanzadeh13, line 13: Moving some test fixtures to `p2p/test/fixtures.go` so that they become accessible from `insecure` package (for BFT testing).

network/internal/p2putils/utils.go
```diff
@@ -164,10 +163,3 @@ func IPPortFromMultiAddress(addrs ...multiaddr.Multiaddr) (string, string, error
 	}
 	return "", "", fmt.Errorf("ip address or hostname not found")
 }
-
-// AllowAllPeerFilter returns a peer filter that does not do any filtering.
-func AllowAllPeerFilter() p2p.PeerFilter {
```
yhassanzadeh13, line 169: moving this fixture to `/p2p/tests/fixtures.go`.

network/p2p/p2pbuilder/libp2pNodeBuilder.go
```diff
@@ -329,37 +344,25 @@ func defaultLibP2POptions(address string, key fcrypto.PrivateKey) ([]config.Opti
...
 		libp2p.ListenAddrs(sourceMultiAddr), // set the listen address
 		libp2p.Identity(libp2pKey),          // pass in the networking key
-		transport,                           // set the protocol
+		t,                                   // set the transport
 	}
 
 	return options, nil
 }
 
-func DefaultPubsubOptions(maxPubSubMsgSize int) []pubsub.Option {
```
yhassanzadeh13, line 351: moved to `p2pode/gossipsubAdapterConfig.go`

network/p2p/test/fixtures.go
```diff
@@ -0,0 +1,305 @@
+package p2ptest
```
yhassanzadeh13, line 1: No new logic in this file, mainly migrated some fixtures from `internal` packages.

insecure/corruptlibp2p/fixtures.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+		}
+		msg.Ihave = iHaves
+	}
+}
+
+// gossipSubMessageIdFixture returns a random gossipSub message ID.
+func gossipSubMessageIdFixture() string {
+	// TODO: messageID length should be a parameter.
+	return unittest.GenerateRandomStringWithLen(10)
```
kc1116, line 38: Move 10 to a const variable.

insecure/corruptlibp2p/pubsubAdapter.go
```diff
@@ -0,0 +1,120 @@
...
+		case p2p.ValidationIgnore:
+			return corrupt.ValidationIgnore
+		case p2p.ValidationReject:
+			return corrupt.ValidationReject
+		default:
+			// should never happen, indicates a bug in the topic validator
+			c.logger.Fatal().Msgf("invalid validation result: %v", result)
+		}
+		// should never happen, indicates a bug in the topic validator, but we need to return something
+		c.logger.Warn().
```
kc1116, line 62: Same comment as line 59 add some message context to the log.

insecure/corruptlibp2p/spammerGossipSub.go
```diff
@@ -0,0 +1,32 @@
...
+
+func NewSpammerGossipSubRouter(router *pubsub.GossipSubRouter) *SpammerGossipSub {
+	return &SpammerGossipSub{
+		router: router,
+	}
+}
+
+// SpamIHave spams the victim with junk iHave messages.
+// msgCount is the number of iHave messages to send.
+// msgSize is the number of messageIDs to include in each iHave message.
```
kc1116, line 24: Add note in godoc mentioning msgSize is currently static.

insecure/corruptlibp2p/fixtures.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3585/commits/dc80931f84d4cfa3fb7cf770d0fb2a481f008df8

insecure/corruptlibp2p/pubsubAdapter.go
```diff
@@ -0,0 +1,120 @@
...
+		switch result {
+		case p2p.ValidationAccept:
+			return corrupt.ValidationAccept
+		case p2p.ValidationIgnore:
+			return corrupt.ValidationIgnore
+		case p2p.ValidationReject:
+			return corrupt.ValidationReject
+		default:
+			// should never happen, indicates a bug in the topic validator
+			c.logger.Fatal().Msgf("invalid validation result: %v", result)
```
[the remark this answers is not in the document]
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3585/commits/2cbea2a4f0df1643b72ea1b2c1c8ed4d9ec030e7

insecure/corruptlibp2p/pubsubAdapter.go
yhassanzadeh13: https://github.com/onflow/flow-go/pull/3585/commits/2cbea2a4f0df1643b72ea1b2c1c8ed4d9ec030e7

insecure/corruptlibp2p/spammerGossipSub.go
yhassanzadeh13: they are not static, we use the parameters: https://github.com/onflow/flow-go/blob/dc80931f84d4cfa3fb7cf770d0fb2a481f008df8/insecure/corruptlibp2p/fixtures.go#L31-L37

kc1116: I was referring to the underlying method used. Doesn't look like we currently use the message size and create them all with a size of 10. 

https://github.com/onflow/flow-go/blob/2cbea2a4f0df1643b72ea1b2c1c8ed4d9ec030e7/insecure/corruptlibp2p/fixtures.go#L42-L46

kc1116 (approved):
Generally looks good. 
nit: I noticed some file names are using snakeCase while others are using camel_case , we should stick to a single format

insecure/corruptlibp2p/spammerGossipSub.go
```diff
@@ -0,0 +1,32 @@
...
+func (s *SpammerGossipSub) SpamIHave(victim peer.ID, msgCount int, msgSize int) {
```
kc1116, line 25: Do you think we need any unit test coverage for this file?

Makefile
```diff
@@ -52,7 +52,7 @@ cmd/util/util:
 .PHONY: unittest-main
 unittest-main:
 	# test all packages with Relic library enabled
-	go test -coverprofile=$(COVER_PROFILE) -covermode=atomic $(if $(RACE_DETECTOR),-race,) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic $(GO_TEST_PACKAGES)
+	go test -v -coverprofile=$(COVER_PROFILE) -covermode=atomic $(if $(RACE_DETECTOR),-race,) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) --tags relic $(GO_TEST_PACKAGES)
```
peterargue, line 55: did you mean to enable verbose logging here? I think that will get very noisy

insecure/cmd/build_helper1.sh
```diff
@@ -0,0 +1,12 @@
+# temporarily make insecure/ a non-module to allow Docker to use corrupt builders there
```
peterargue, line 1: can we name these to reflect what they're doing? maybe `override_mods.sh` and `restore_mods.sh` or something like that

network/p2p/p2pnode/gossipSubAdapter.go
```diff
@@ -0,0 +1,83 @@
...
+// that implements the PubSubAdapter interface for the Flow network.
+type GossipSubAdapter struct {
+	gossipSub *pubsub.PubSub
+	logger    zerolog.Logger
+}
+
+var _ p2p.PubSubAdapter = (*GossipSubAdapter)(nil)
+
+func NewGossipSubAdapter(ctx context.Context, logger zerolog.Logger, h host.Host, cfg p2p.PubSubAdapterConfig) (p2p.PubSubAdapter, error) {
+	gossipSubConfig, ok := cfg.(*GossipSubAdapterConfig)
```
peterargue, line 26: given this must be a `*GossipSubAdapterConfig`, why not use that as the parameter type and avoid this check?

integration/Makefile
```diff
@@ -22,48 +22,48 @@ ci-integration-test: access-tests ghost-tests mvp-tests epochs-tests consensus-t
 # Run unit tests for test utilities in this module
 .PHONY: test
 test:
-	go test $(if $(VERBOSE),-v,) -tags relic -coverprofile=$(COVER_PROFILE) $(RACE_FLAG) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) `go list ./... | grep -v -e integration/tests`
```
peterargue, line 25: why enable this by default? wont this get very noisy in the CI logs?

peterargue (approved):
looks good aside from a few minor comments

Makefile
gomisha: Yes, I enabled it because we have had issues in CI of test failures but without additional logging so it was hard to tell what was wrong. This is mostly for CI test failure diagnosis.

integration/Makefile
gomisha: The VERBOSE flag was never set in CI so it was never verbose. We have had CI test failures without verbose logging that made it difficult to determine what was wrong. This shouldn't add a lot of extra logging, just the name of each test as it's being executed and a more explicit failure message. See https://github.com/onflow/flow-go/pull/3585#discussion_r1037417458

insecure/cmd/build_helper1.sh
gomisha: https://github.com/onflow/flow-go/pull/3585/commits/d4937ff24a66108377e78b1512ad8a8eb26e2771

network/p2p/p2pnode/gossipSubAdapter.go
gomisha: `NewGossipSubAdapter()` is called by `defaultGossipSubFactory()` which expects a `GossipSubFactoryFuc` return type and that func expects the base type of `p2p.PubSubAdapterConfig`. Changing this to the sub-type would break creation of `GossibSubFactory`

https://github.com/onflow/flow-go/pull/3585/files#diff-4edd977ff4567bb1e4b7042a1d973e12b11732f4bb622bf934235fd30d444c59R126-R130

insecure/corruptlibp2p/spammerGossipSub.go
gomisha: Yes, this will have extensive unit test coverage in a separate PR. This is just the basic skeleton of the spammer that will be used by unit tests.

Makefile
peterargue: ok. another option is the add a `VERBOSE` flag like in the `make -e VERBOSE test` case to enable it specifically in CI.

but in either case we can just adjust it later if it becomes a problem

PR #3586, opened 2022-11-17
[Execution] Fix ingestion loading unexecuted blocks

zhangchiqing:
This PR fixes a bug that when ingestion engine can't load unexecuted blocks

### DIFF
--- a/engine/execution/ingestion/engine.go
+++ b/engine/execution/ingestion/engine.go
@@ -404,13 +404,29 @@ func (e *Engine) reloadBlock(
 		return fmt.Errorf("could not get block by ID: %v %w", blockID, err)
 	}
 
+	// enqueue the block and check if there is any missing collections
 	missingCollections, err := e.enqueueBlockAndCheckExecutable(blockByCollection, executionQueues, block, false)
 
 	if err != nil {
 		return fmt.Errorf("could not enqueue block %x on reloading: %w", blockID, err)
 	}
 
-	return e.addOrFetch(blockID, block.Header.Height, missingCollections)
+	// forward the missing collections to requester engine for requesting them from collection nodes,
+	// adding the missing collections to mempool in order to trigger the block execution as soon as
+	// all missing collections are received.
+	err = e.fetchAndHandleCollection(blockID, block.Header.Height, missingCollections, func(collection *flow.Collection) error {
+		err := e.addCollectionToMempool(collection, blockByCollection)
+
+		if err != nil {
+			return fmt.Errorf("could not add collection to mempool: %w", err)
+		}
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("could not fetch or handle collection %w", err)
+	}
+	return nil
 }
 
 // BlockProcessable handles the new verified blocks (blocks that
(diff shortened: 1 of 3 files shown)


engine/execution/ingestion/engine.go
```diff
@@ -410,7 +410,42 @@ func (e *Engine) reloadBlock(
 		return fmt.Errorf("could not enqueue block %x on reloading: %w", blockID, err)
 	}
 
-	return e.addOrFetch(blockID, block.Header.Height, missingCollections)
+	height := block.Header.Height
```
zhangchiqing, line 413: We can't reuse the `addOrFetch` function, because `addOrFetch` wants to acquire the lock of mempool, but reloadBlock has already acquired the same lock. Calling `addOrFetch` here would cause deadlock. 

So we had to duplicate the logic here in order to fetch collection and add directly the mempool data

engine/execution/ingestion/engine.go
```diff
@@ -876,62 +911,62 @@ func (e *Engine) handleCollection(originID flow.Identifier, collection *flow.Col
...
-	return e.addCollectionToMempool(collection)
+	return e.mempool.BlockByCollection.Run(
+		func(backdata *stdmap.BlockByCollectionBackdata) error {
+			return e.addCollectionToMempool(collection, backdata)
+		},
+	)
 }
 
-func (e *Engine) addCollectionToMempool(collection *flow.Collection) error {
+func (e *Engine) addCollectionToMempool(collection *flow.Collection, backdata *stdmap.BlockByCollectionBackdata) error {
```
zhangchiqing, line 921: Letting this function take the mempool data and moving the lock acquiring logic out of this function, so that it can also be used by `reloadBlock` function

engine/execution/ingestion/engine.go
```diff
@@ -1301,7 +1336,11 @@ func (e *Engine) addOrFetch(blockID flow.Identifier, height uint64, guarantees [
 		collection, err := e.collections.ByID(guarantee.CollectionID)
 
 		if err == nil {
-			err := e.addCollectionToMempool(collection)
+			err = e.mempool.BlockByCollection.Run(
```
zhangchiqing, line 1339: we now acquire lock outside of the `addCollectionToMempool` call.

engine/execution/ingestion/engine.go
pattyshack: can you refactor addOrFetch instead of copying/pasting the logic?  Maybe have addOrFetch acquire the lock and call unsafeAddOrFetch, and use unsafeAddOrFetch here.

pattyshack (changes requested):

engine/execution/ingestion/engine.go
zhangchiqing: Yeah, good point, but if we do that, then `addOrFetch` would hold the lock for too long. For instance, I will be holding the lock while calling the following methods:
- e.collection.ByID
- fetchCollection
- e.request.Force

I'd like to have `addOrFetch` holding the lock as minimum as possible, because holding for too long might cause other side effects.

Any other idea?

engine/execution/ingestion/engine.go
```diff
@@ -1294,16 +1310,39 @@ func GenerateExecutionReceipt(
...
+			return fmt.Errorf("could not add collection to mempool: %w", err)
+		}
+		return nil
+	})
+}
+
+// addOrFetch checks if there are stored collections for the given guarantees, if there is,
+// forward them to the handler to process the collection, otherwise fetch the collections.
+// any error returned are exception
+func (e *Engine) fetchAndHandleCollection(
```
zhangchiqing, line 1329: This function is extracted so that it can be reused by both `reloadBlock` and `addOrFetch`

engine/execution/ingestion/engine.go
pattyshack: I rather the code be correct than fast.  (There are a ton of thread-safe data races, aka byzantine failures, in the code base ...)

Which field(s)/method(s) does the mutex protect?  I'm not familiar with this code, but based on your list of methods, which I assume are data race free, the only thing you really need to guard is addCollectionToMempool. maybe restructure the code so that you only lock in addCollectionToMempool if that's the case.

zhangchiqing: yeah, I refactored it a bit, and extracted the fetchAndHandleCollection function to be reused, which takes a function for handling collections, so that addAndFetch can provide a lock which acquires a lock, and reloadBlock can add the collection directly to the mempool as it's already holding the lock.

pattyshack (approved):

m4ksio (approved):

PR #3588, opened 2022-11-17
Fix storage interaction limit setup

pattyshack:
Storage interaction limit is part of meter parameters and should be set along the other meter parameters.

gh: #3102

### DIFF
--- a/fvm/executionParameters.go
+++ b/fvm/executionParameters.go
@@ -24,6 +24,26 @@ type MeterParamOverrides struct {
 	MemoryLimit        *uint64                      // nil indicates no override
 }
 
+func getBasicMeterParameters(
+	ctx Context,
+	proc Procedure,
+) meter.MeterParameters {
+	params := meter.DefaultParameters().
+		WithComputationLimit(uint(proc.ComputationLimit(ctx))).
+		WithMemoryLimit(proc.MemoryLimit(ctx)).
+		WithEventEmitByteLimit(ctx.EventCollectionByteSizeLimit).
+		WithStorageInteractionLimit(ctx.MaxStateInteractionSize)
+
+	// NOTE: The memory limit (and interaction limit) may be overridden by the
+	// environment.  We need to ignore the override in that case.
+	if proc.ShouldDisableMemoryAndInteractionLimits(ctx) {
+		params = params.WithMemoryLimit(math.MaxUint64).
+			WithStorageInteractionLimit(math.MaxUint64)
+	}
+
+	return params
+}
+
 func getMeterParameters(
 	ctx Context,
 	proc Procedure,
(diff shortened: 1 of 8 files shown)


fvm/executionParameters.go
```diff
@@ -24,6 +24,26 @@ type MeterParamOverrides struct {
...
+	ctx Context,
+	proc Procedure,
+) meter.MeterParameters {
+	params := meter.DefaultParameters().
+		WithComputationLimit(uint(proc.ComputationLimit(ctx))).
+		WithMemoryLimit(proc.MemoryLimit(ctx)).
+		WithEventEmitByteLimit(ctx.EventCollectionByteSizeLimit).
+		WithStorageInteractionLimit(ctx.MaxStateInteractionSize)
+
+	// NOTE: The memory limit (and interaction limit) may be override by the
```
SaveTheRbtz, line 37: nit:
```suggestion
	// NOTE: The memory limit (and interaction limit) may be override by the
	// NOTE: The memory limit (and interaction limit) may be overridden by the
```

SaveTheRbtz (approved):

fvm/executionParameters.go
```diff
@@ -33,17 +53,14 @@ func getMeterParameters(
...
-		WithEventEmitByteLimit(ctx.EventCollectionByteSizeLimit)
+	procParams := getBasicMeterParameters(ctx, proc)
 
 	txnState := state.NewTransactionState(
 		view,
 		state.DefaultParameters().
 			WithMaxKeySizeAllowed(ctx.MaxStateKeySize).
 			WithMaxValueSizeAllowed(ctx.MaxStateValueSize).
-			WithMaxInteractionSizeAllowed(ctx.MaxStateInteractionSize))
+			WithMeterParameters(procParams))
```
zhangchiqing, line 63: I'm not familiar with the different between `procParams` and `overrides`. 

But we are passing `procParams` here, and later updates the procParams if overrides.ComputationWeights is not nil (L74). However txnState won't see those updates, is that what we wanted? If so, why?

fvm/executionParameters.go
```diff
@@ -24,6 +24,26 @@ type MeterParamOverrides struct {
...
+	proc Procedure,
+) meter.MeterParameters {
+	params := meter.DefaultParameters().
+		WithComputationLimit(uint(proc.ComputationLimit(ctx))).
+		WithMemoryLimit(proc.MemoryLimit(ctx)).
+		WithEventEmitByteLimit(ctx.EventCollectionByteSizeLimit).
+		WithStorageInteractionLimit(ctx.MaxStateInteractionSize)
+
+	// NOTE: The memory limit (and interaction limit) may be overridden by the
+	// environment.  We need to ignore the override in that case.
```
zhangchiqing, line 38: why do we need to ignore the override?

If we need to ignore the override, why do we override it in the first place?

fvm/executionParameters.go
pattyshack: The meter computation/memory weights (and memory limit) are dynamically configurable via these overrides.  (I'll give a more detailed answer about this in #3576, once I'm done responding to this PR)

Right now, we're using two different txnState in a single transaction:
1. The first txnState, without environment overrides, is used to fetch the environment overrides; as you pointed out, this txnState does not get updated.
2. The second txnState (created in fvm.go), with the environment overrides applied, is used to execute the actual transaction: `auth verification`, `seq num check`, and `body execution`.  Note that `get meter parameters` is not currently part of the transaction executor.  Also note that only `body execution` needs the updated meter parameters, the rest (`auth verification` and `seq num check`) don't need these overrides.

I'm planning to merge to the two txnState into one (and make `get meter parameters` part of the transaction executor).  `auth verification`, `seq num check`, and `get meter parameters` will use the basic meter parameters, and only the `body execution` will use the "enhanced" meter parameters.  See #2922 (and #3576) for additional discussion/detail.

fvm/executionParameters.go
pattyshack: This is prep work for what i've described above.  you can think of the overrides in terms of priorities:

1. ShouldDisableMemoryAndInteractionLimits has the highest priority
2. Environment meter overrides have the medium priority
3. Basic meter parameters has the lowest priority

zhangchiqing (approved):

pattyshack:
bors merge

pattyshack:
urgh ... I have to add back WithMaxInteractionSizeAllowed since it's used by the emulator / integration test.  I'll add a todo to clean it up instead ...

pattyshack:
bors merge

pattyshack:
bors merge

pattyshack:
bors retry

pattyshack:
bors retry

PR #3594, opened 2022-11-18
Meter param overrides cache skeleton.

pattyshack:
This adds a second (meter param overrides) table to the derived block data database.  The database is committed via Validate/Commit 2PC (For now, Validate is not called during actual execution since there are no conflicts in sequential execution)

Note that this PR triggers meter params invalidation when contracts are updated. This is incorrect, but it's no worst than before since updating meter params would also break programs invalidation.  I'll add correct meter param invalidation in a followup PR.

gh: #3102

### DIFF
--- a/fvm/executionParameters.go
+++ b/fvm/executionParameters.go
@@ -18,12 +18,6 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
-type MeterParamOverrides struct {
-	ComputationWeights meter.ExecutionEffortWeights // nil indicates no override
-	MemoryWeights      meter.ExecutionMemoryWeights // nil indicates no override
-	MemoryLimit        *uint64                      // nil indicates no override
-}
-
 func getBasicMeterParameters(
 	ctx Context,
 	proc Procedure,
@@ -63,10 +57,9 @@ func getMeterParameters(
 			WithMeterParameters(procParams))
 
 	// TODO(patrick): cache meter param overrides
-	overrides, err := getMeterParamOverrides(
-		ctx,
+	overrides, err := derivedTxnData.GetMeterParamOverrides(
 		txnState,
-		derivedTxnData)
+		meterParamOverridesComputer{ctx, derivedTxnData})
 	if err != nil {
 		return procParams, err
 	}
@@ -94,21 +87,22 @@ func getMeterParameters(
 	return procParams, nil
 }
 
-func getMeterParamOverrides(
-	ctx Context,
+type meterParamOverridesComputer struct {
+	ctx            Context
+	derivedTxnData *programs.DerivedTransactionData
+}
+
+func (computer meterParamOverridesComputer) Compute(
 	txnState *state.TransactionState,
-	derivedTxnData *programs.DerivedTransactionData,
+	_ struct{},
 ) (
-	MeterParamOverrides,
+	programs.MeterParamOverrides,
 	error,
 ) {
-	var overrides MeterParamOverrides
+	var overrides programs.MeterParamOverrides
 	var err error
 	txnState.RunWithAllLimitsDisabled(func() {
-		overrides, err = getMeterParamOverridesFromState(
-			ctx,
-			txnState,
-			derivedTxnData)
+		overrides, err = computer.getMeterParamOverrides(txnState)
 	})
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


fvm/programs/block_programs.go
```diff
@@ -20,6 +26,10 @@ type DerivedTransactionData struct {
 		common.AddressLocation,
 		*interpreter.Program,
 	]
+
+	// There's only a single entry in this table.  For simplicity, we'll use
+	// 0 as the entry's key.
+	meterParamOverrides *TransactionDerivedData[int, MeterParamOverrides]
```
SaveTheRbtz, line 32: maybe `struct{}{}`?

SaveTheRbtz (approved):

fvm/executionParameters.go
```diff
@@ -18,12 +18,6 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
-type MeterParamOverrides struct {
```
zhangchiqing, line 21: You mentioned a second override is being ordered. I have a few questions regarding this:

1. Why do we need to override meter parameter?
2. If we need, what is the first override?
3. why can't we keep just 1 override by merging the second override into the first one?
4. If we need second override, would any process see the meter parameter without the second override?
5. How do we ensure the 2 overrides are ordered correctly? What I'm concerned is that if the 2 overrides are updating the same data with different value, then a wrong order would cause the final overridden value to be wrong.

fvm/programs/block_programs.go
pattyshack: o, i didn't know that works as a key.  I'll switch to that.

fvm/programs/block_derived_data.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/onflow/flow-go/fvm/state"
 )
 
+type ValueComputer[TKey any, TVal any] interface {
+	Compute(txnState *state.TransactionState, key TKey) (TVal, error)
```
zhangchiqing, line 11: Can you add some comments to this method?

Is it to get value from the state under the given key? Why it's not called `GetValue`? What is the actual "Computation"?

fvm/programs/invalidator.go
```diff
@@ -18,13 +19,22 @@ type ContractUpdate struct {
...
+	MemoryWeights      meter.ExecutionMemoryWeights // nil indicates no override
+	MemoryLimit        *uint64                      // nil indicates no override
+}
+
 type ProgramInvalidator DerivedDataInvalidator[
 	common.AddressLocation,
 	*interpreter.Program,
 ]
 
+type MeterParamOverridesInvalidator DerivedDataInvalidator[int, MeterParamOverrides]
```
zhangchiqing, line 33: Is this the second table you are adding?

Is the `int` the type of `meterParamOverridesKey`?

fvm/programs/block_derived_data.go
zhangchiqing: Could you also explain why do we need to use those generic type `TKey` and `TVal`?

fvm/programs/block_programs.go
```diff
@@ -118,12 +154,47 @@ func (transaction *DerivedTransactionData) AddInvalidator(
...
 func (transaction *DerivedTransactionData) Commit() RetryableError {
-	return transaction.programs.Commit()
+	err := transaction.programs.Commit()
+	if err != nil {
+		return err
+	}
+
+	err = transaction.meterParamOverrides.Commit()
+	if err != nil {
+		return err
```
zhangchiqing, line 196: better to add error message to distinguish the error

fvm/programs/block_derived_data.go
pattyshack: (will add comment as suggested)

This is used by BlockDerivedData, which is a schema-ed ACI (not D) database table for stroing TKey/TVal pairs.  This is generic because BlockDerivedData is generic.

fvm/programs/block_programs.go
pattyshack: good point.  will do

fvm/programs/invalidator.go
pattyshack: yes, this is the second table.  as alexey pointed out, i should use `struct{}` instead of `int` as the key since the table is single valued.

fvm/programs/block_programs.go
```diff
@@ -20,6 +24,10 @@ type DerivedTransactionData struct {
 		common.AddressLocation,
 		*interpreter.Program,
 	]
+
+	// There's only a single entry in this table.  For simplicity, we'll use
+	// 0 as the entry's key.
```
zhangchiqing, line 29:
```suggestion
	// struct{} as the entry's key.
```

zhangchiqing (approved):

fvm/programs/block_programs.go
pattyshack: thx

pattyshack:
bors merge

pattyshack:
bors merge

PR #3598, opened 2022-11-20
Performance Improvements for Cluster Compliance Engine

jordanschalm:
Addresses some suggestions from https://github.com/onflow/flow-go/issues/3374:
* use notifications to cache finalized height/view in compliance core
* move filter for finalized height/view to top of `OnBlockProposal` - avoid DB query for any block which will be dropped based on its height or view
* log transaction IDs of received collections while in debug logging mode (extra)


engine/collection/compliance/core.go
```diff
@@ -107,9 +120,29 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
...
 		Hex("signers", header.ParentVoterIndices).
+		Uint64("final_height", finalHeight).
 		Logger()
+	if log.Debug().Enabled() {
+		log = log.With().Strs("tx_ids", flow.IdentifierList(proposal.Payload.Collection.Light().Transactions).Strings()).Logger()
+	}
 	log.Info().Msg("block proposal received")
 
+	// drop proposals below the finalized threshold
+	if header.Height < finalHeight || header.View < finalView {
```
durkmurder, line 131: I think we are interested only in final height, since height is property of finalized chain and there is no way that first check won't pass but second one will

durkmurder: This will also allow to drop `c.finalizedView` completely.

engine/collection/compliance/engine.go
```diff
@@ -216,14 +215,27 @@ func (e *Engine) OnSyncedClusterBlock(syncedBlock flow.Slashable[messages.Cluste
...
-	blockFinalizedSignal := e.finalizationEventsNotifier.Channel()
 	for {
 		select {
 		case <-doneSignal:
 			return
-		case <-blockFinalizedSignal:
-			e.core.ProcessFinalizedView(e.finalizedView.Value())
+		case finalizedBlock := <-e.finalizedBlocksNotifications:
+			// drop outdated notifications
+			if finalizedBlock.View < finalView {
```
durkmurder, line 228: 1. `finalView` is not updated.
2. Why do we need this? To prevent delivery out of order? This shouldn't happen at all, meaning we don't expect to receive height `n+1` before height `n`

jordanschalm: It's replicating what `if e.finalizedView.Set(block.View)` was doing previously. The only real purpose is to skip processing duplicate notifications, rather than out of order. I'd be happy to get rid of it though - dupe notifications are unlikely and their impact is low.

durkmurder: Yeah, same here, I would get rid of it, makes logic simpler and it's very unlikely that this will happen.

jordanschalm: Removed in https://github.com/onflow/flow-go/pull/3598/commits/36c3380942411252725382b95820f5f3375b6d1c

engine/collection/compliance/core.go
```diff
@@ -107,9 +120,29 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Clus
 		Time("timestamp", header.Timestamp).
 		Hex("proposer", header.ProposerID[:]).
 		Hex("signers", header.ParentVoterIndices).
+		Uint64("final_height", finalHeight).
```
zhangchiqing, line 123:
```suggestion
		Uint64("final_height", finalHeight).
		Uint64("final_view", finalView).
```

engine/collection/compliance/core.go
zhangchiqing, line 131: I think we could also drop a proposal if its height is equal to finalized height
```suggestion
	if header.Height <= finalHeight || header.View <= finalView {
```

zhangchiqing (approved):
Nice improvements. I wonder if we can do the same for consensus compliance engine?

jordanschalm:
> I wonder if we can do the same for consensus compliance engine?

Added in https://github.com/onflow/flow-go/pull/3598/commits/b72797ed881df6298c7bc3e1b0ccc471fd897629

engine/collection/compliance/engine.go
```diff
@@ -65,16 +62,16 @@ func NewEngine(
...
+		log:                          engineLog,
+		metrics:                      core.engineMetrics,
+		me:                           me,
+		headers:                      core.headers,
+		payloads:                     payloads,
+		state:                        state,
+		core:                         core,
+		pendingBlocks:                blocksQueue,
+		pendingBlocksNotifier:        engine.NewNotifier(),
+		finalizedBlocksNotifications: make(chan *model.Block, 10),
```
AlexHentschel, line 74: ## ⚠️ 
From a conceptual standpoint, I dislike the usage of a buffered channel here, especially of size 10. When there is a higher fraction of offline primaries, I believe it is not too unlikely that finalization could halt for more than 10 views. (see probability approximation here in [this notion doc](https://www.notion.so/dapperlabs/Probability-of-finalization-halt-for-n-views-72997a3bff2446ae96e1824e57eb04de))
After such a temporary finalization halt, there will be a quick burst of finalized blocks, once we have a sufficient sequence of 2 subsequent honest primaries. If we exceed the capacity of 10 finalized blocks, the compliance engine will start dropping the newer finalized blocks and therefore delay pruning. 


### Suggestion
I believe there is not really any reason to work with a channel here. In fact, we are only interested in the _latest_ finalized block (by view). We have high-performance, completely non-blocking implementations [here](https://github.com/onflow/flow-go/blob/b4a7dbb4a6b7cf09a481f7eaa49c5bd8057f2427/consensus/hotstuff/tracker/tracker.go) for tracking the latest QC and TC. It is really not a lot of code. 

I would suggest to add a similar Tracker for latest block by view. I believe this would be widely useful beyond this PR.

engine/consensus/compliance/engine.go
```diff
@@ -65,18 +61,18 @@ func NewEngine(
...
+		mempoolMetrics:               core.mempoolMetrics,
+		engineMetrics:                core.engineMetrics,
+		headers:                      core.headers,
+		payloads:                     core.payloads,
+		pendingBlocks:                blocksQueue,
+		state:                        core.state,
+		tracer:                       core.tracer,
+		core:                         core,
+		pendingBlocksNotifier:        engine.NewNotifier(),
+		finalizedBlocksNotifications: make(chan *model.Block, 10),
```
AlexHentschel, line 75: same comment as for the collectors' compliance engine

AlexHentschel (approved):
Looks great. I have a suggestion for a conceptual improvement, to more closely reflect the intension of the protocol.

engine/consensus/compliance/core.go
```diff
@@ -119,7 +133,11 @@ func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.Bloc
 		Time("timestamp", header.Timestamp).
 		Hex("proposer", header.ProposerID[:]).
 		Hex("parent_signer_indices", header.ParentVoterIndices).
+		Uint64("final_height", finalHeight).
+		Uint64("final_view", finalView).
```
jordanschalm, line 137:
```suggestion
		Uint64("finalized_height", finalHeight).
		Uint64("finalized_view", finalView).
```

engine/collection/compliance/engine.go
jordanschalm: Done

jordanschalm:
bors merge

PR #3602, opened 2022-11-21
[BFT] Limit inbound unicast streams

peterargue:
This PR introduces a configurable limit on the number of inbound unicast connections from a peer. The limit applies across all connections, and any streams created beyond the limit are reset.


module/metrics.go
```diff
@@ -36,6 +36,9 @@ type NetworkSecurityMetrics interface {
 
 	// OnRateLimitedUnicastMessage tracks the number of rate limited messages seen on the network.
 	OnRateLimitedUnicastMessage(role, msgType, topic, reason string)
+
+	// UnicastStreamDropped tracks the number of unicast streams that were dropped due to exceeding the limit.
+	UnicastStreamDropped()
```
kc1116, line 41: Should we rename this `UnicastStreamReset` ?

peterargue: we could, but it doesn't capture all resets, just streams that are reset due to exceeding the limit

network/p2p/middleware/middleware.go
```diff
@@ -482,6 +489,17 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
 		return
 	}
 
+	streamCount := p2putils.CountStreams(m.libP2PNode.Host(), remotePeer, s.Protocol(), libp2pnetwork.DirInbound)
```
yhassanzadeh13, line 492: Assume `m.unicastMaxStreamsPerPeer = x` and a remote peer concurrently making `y > x` streams to the current peer. Can it result in all `y` streams being deemed over the limit and closed? Ideally, we want `x` out of `y` inbound streams to pass through and `x - y` ones dropped and closed. If that is the case, should we consider a mutual exclusion on handing streams from the same remote peer? i.e., streams from the same remote peer are handled sequentially.

cmd/node_builder.go
```diff
@@ -182,20 +182,24 @@ type NetworkConfig struct {
...
 	// UnicastBandwidthRateLimit bandwidth size in bytes a peer is allowed to send via unicast streams per second.
 	UnicastBandwidthRateLimit int
 	// UnicastBandwidthBurstLimit bandwidth size in bytes a peer is allowed to send via unicast streams at once.
 	UnicastBandwidthBurstLimit int
-	PeerUpdateInterval         time.Duration
-	UnicastMessageTimeout      time.Duration
-	DNSCacheTTL                time.Duration
+
+	// UnicastMaxStreamsPerPeer is the maximum number of inbound unicast streams that can be opened with a peer.
+	UnicastMaxStreamsPerPeer uint
```
yhassanzadeh13, line 202:
```suggestion

	// UnicastMaxStreamsPerPeer is the maximum number of inbound unicast streams that can be opened from a remote peer to the current peer.
	UnicastMaxStreamsPerPeer uint
```

module/metrics.go
yhassanzadeh13: Seems we attempt on [closing](https://github.com/onflow/flow-go/blob/900b3188765f2a228284dd14d42749ba6dd25cd6/network/p2p/middleware/middleware.go#L472) (and not directly reset) upon meeting the limit. So `dropped` or `closed` may fit more than `reset`.

network/test/middleware_test.go
```diff
@@ -805,6 +811,80 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+
+	// configured mw.unicastMaxStreamsPerPeer (using default)
+	maxStreams := middleware.DefaultMaxUnicastStreamsPerPeer
+
+	receiveCount := 0
+	m.ov[receiver].On("Receive", senderNodeID, mockery.Anything, mockery.Anything).Return(nil).Run(func(_ mockery.Arguments) {
+		receiveCount++
+	})
+
+	var protocol protocol.ID
```
yhassanzadeh13, line 834:
```suggestion
	var streamProtocol protocol.ID
```
`protocol` collides with a package name.

network/test/middleware_test.go
```diff
@@ -805,6 +811,80 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
 	unittest.RequireNeverReturnBefore(m.T(), msgRcvdFun, 2*time.Second, "message received unexpectedly")
 }
 
+// TestUnicastStreamLimit tests that the middleware does not accept more than the maximum number of
+// unicast streams and streams created beyond the limit are reset
+func (m *MiddlewareTestSuite) TestUnicastStreamLimit() {
```
yhassanzadeh13, line 816: Following my previous comment, would you please also add a concurrent version of this test? i.e., creating streams concurrently and also closing them concurrently.

network/test/middleware_test.go
```diff
@@ -805,6 +811,80 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+
+	senderPeerID := m.nodes[sender].Host().ID()
+	receiverPeerID := m.nodes[receiver].Host().ID()
+
+	// configured mw.unicastMaxStreamsPerPeer (using default)
+	maxStreams := middleware.DefaultMaxUnicastStreamsPerPeer
+
+	receiveCount := 0
+	m.ov[receiver].On("Receive", senderNodeID, mockery.Anything, mockery.Anything).Return(nil).Run(func(_ mockery.Arguments) {
+		receiveCount++
```
yhassanzadeh13, line 831: Please also check the content of the received message against what was originally sent. Ideally, we want `maxStreams`-many unique messages received by the overlay. For this we need to send unique messages on the streams (e.g., `hello` + `i`). This way we ensure stream rate limiting does not interfere with the typical procedure of message delivery.

module/metrics.go
peterargue: it's actually [resetting](https://github.com/onflow/flow-go/blob/900b3188765f2a228284dd14d42749ba6dd25cd6/network/p2p/middleware/middleware.go#L477) since `success` is false. This is also confirmed by the unittests.

network/p2p/middleware/middleware.go
```diff
@@ -482,6 +489,17 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
...
+	if streamCount > m.unicastMaxStreamsPerPeer {
```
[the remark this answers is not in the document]
peterargue: what's the suggestion here? It looks like those are the same.

network/p2p/middleware/middleware.go
```diff
@@ -482,6 +489,17 @@ func (m *Middleware) handleIncomingStream(s libp2pnetwork.Stream) {
...
+		m.metrics.UnicastStreamDropped()
+		log.Debug().
+			Err(fmt.Errorf("too many inbound streams from peer")).
+			Int("stream_count", streamCount).
+			Bool(logging.KeySuspicious, true).
+			Msg("resetting new stream")
+		return
```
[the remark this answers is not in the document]
peterargue: it's actually [resetting](https://github.com/onflow/flow-go/blob/900b3188765f2a228284dd14d42749ba6dd25cd6/network/p2p/middleware/middleware.go#L477) since `success` is false. This is also confirmed by the unittests.

network/test/middleware_test.go
peterargue: b5c22ed

network/test/middleware_test.go
peterargue: b5c22ed

network/p2p/middleware/middleware.go
peterargue: I did some testing of this. When `y` is slightly larger than `x`, then usually about 90% of the expected streams are handled correctly. This is about the same as the serial case so the difference is most likely flakiness in the network setup.

When `y` is much larger than `x`, the % false positives grows. 
* at `y == x*4`, ~80% of the expected streams are handled correctly.
* at `y == x*16`, ~60% of the expected streams are handled correctly.

As you noted, this is because there's a race condition between when the stream is opened and when the check occurs in `handleIncomingStream`. Ideally, the add and check would happen atomically so peers would be able to open precisely `m.unicastMaxStreamsPerPeer` streams. 

However, I think the current behavior is OK for a few reasons:
* The number of streams allowed is always less than `m.unicastMaxStreamsPerPeer`, so a malicious peer cannot circumvent the limit.
* When the number of streams is close to the limit, the limit is enforced fairly precisely. This would mean an honest node that's opening slightly larger number of streams, would be treated fairly.
* As the number of concurrently opened streams far exceeds the limit, the percent of rejected streams increases, effectively penalizing bad behavior more.

network/p2p/middleware/middleware.go
```diff
@@ -175,6 +181,7 @@ func NewMiddleware(
 		codec:                      codec,
 		slashingViolationsConsumer: slashingViolationsConsumer,
 		unicastRateLimiters:        ratelimit.NoopRateLimiters(),
+		unicastMaxStreamsPerPeer:   int(unicastMaxStreamsPerPeer),
```
kc1116, line 184: Any reason why we are using 2 types uint & int instead of one?

network/internal/testutils/testUtil.go
```diff
@@ -49,7 +49,7 @@ import (
 	"github.com/onflow/flow-go/utils/unittest"
 )
 
-var sporkID = unittest.IdentifierFixture()
+var SporkID = unittest.IdentifierFixture()
```
kc1116, line 52:
```suggestion
func SporkIDFixture() flow.Identifier {
	return unittest.IdentifierFixture()
}
```
Add SporkIDFixture to the unittest package

network/test/middleware_test.go
```diff
@@ -805,6 +812,212 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+	// this test only works if streamCount > mw.unicastMaxStreamsPerPeer
+	require.Greater(m.T(), streamCount, maxStreams)
+
+	// create all streams concurrently
+	wg := sync.WaitGroup{}
+
+	streams := make(chan streamEvent, streamCount)
+	for i := 0; i < streamCount; i++ {
+		wg.Add(1)
+		i := i
```
kc1116, line 951:
```suggestion
		j := i
```

network/test/middleware_test.go
```diff
@@ -805,6 +812,212 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+			writer := ggio.NewDelimitedWriter(bufw)
+
+			err = writer.WriteMsg(msg)
+			require.NoError(m.T(), err)
+
+			err = bufw.Flush()
+			require.NoError(m.T(), err)
+		}()
+	}
+	wg.Wait()
```
kc1116, line 976:
```suggestion
	unittest.RequireReturnsBefore(m.T(), wg.Wait, 100*time.Millisecond, "could not create streams on time")
```

network/test/middleware_test.go
```diff
@@ -805,6 +812,212 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+	assert.GreaterOrEqual(m.T(), len(receivedMsgs), maxStreams*9/10)
+
+	// the rest should result in resets
+	assert.GreaterOrEqual(m.T(), resets, streamCount-maxStreams)
+}
+
+// TestUnicastStreamLimit_Concurrent tests that the middleware does not accept more than the maximum number of
+// unicast streams and streams created beyond the limit are reset. It does this by creating all of the streams
+// concurrently to additionally validate that the stream limit is enforced correctly.
+func (m *MiddlewareTestSuite) TestUnicastStreamLimit_Concurrent() {
```
kc1116, line 915: It looks like TestUnicastStreamLimit_Concurrent and TestUnicastStreamLimit have identical structures in the test setup only difference being the concurrency. Can you refactor the overlapping logic such that it is shared by both tests.

kc1116 (commented):
Left a few comments.

network/test/middleware_test.go
```diff
@@ -805,6 +812,212 @@ func (m *MiddlewareTestSuite) TestUnsubscribe() {
...
+				assert.False(m.T(), receivedMsgs[e.message], "message '%s' should not have been received", e.message)
+				resets.Inc()
+			} else {
+				assert.True(m.T(), receivedMsgs[e.message], "message '%s' should have been received", e.message)
+			}
+		}()
+		closedStreams++
+	}
+	assert.Equal(m.T(), streamCount, closedStreams)
+	wg.Wait()
```
kc1116, line 1007:
```suggestion
	unittest.RequireReturnsBefore(m.T(), wg.Wait, 100*time.Millisecond, "could not create streams on time")
```

peterargue:
closing. use the libp2p host `ResourceManager` instead

PR #3605, opened 2022-11-22
Index service events

### LINKED ISSUE #3535: Index service events by block height
For further development of rolling upgrade mechnism we need an ability to quickly find the nearest VersionUpdate service event.
As the mechanism is intended to work on all node types, we want to consensus follower to create an index allowing to iterate service events by finalized height.

m4ksio:
This is a first PR in a series of changes for on chain stop control.
For brevity, it adds a service event index in a general manner.
Closes #3535


storage/badger/operation/results.go
```diff
@@ -26,6 +26,11 @@ func IndexExecutionResult(blockID flow.Identifier, resultID flow.Identifier) fun
 	return insert(makePrefix(codeIndexExecutionResultByBlock, blockID), resultID)
 }
 
+// IndexByServiceEvent indexes result ID by a service event type and a sealed block height
```
zhangchiqing, line 29: Let's explain the reasoning of using such key
```suggestion
// IndexByServiceEvent indexes result ID by a service event type and a sealed block height
// this index allows us to answer questions such as what is the latest EpochSetup service event up to 
// block height 1000. 
//
// why indexing result ID rather than service events? because the service event is included in a result,
// the indexed result id could help join the result and find the service event
// 
// why using <service_event_type+sealed_height> as key rather than <sealed_height + service_event_type>?
// because having service_event_type key first allows us to scan through the index and 
// filter by service event type first, and then find the result for highest height, 
// see LookupLastExecutionResultForServiceEvents

File: storage/badger/operation/results_test.go [hunk 2]
```diff
@@ -27,3 +29,72 @@ func TestResults_InsertRetrieve(t *testing.T) {
...
+
+		result1 := unittest.ExecutionResultFixture()
+		result2 := unittest.ExecutionResultFixture()
+		result3 := unittest.ExecutionResultFixture()
+		height1 := uint64(21)
+		height2 := uint64(37)
+		height3 := uint64(55)
+		eventType := flow.ServiceEventCommit
+
+		t.Run("inserting", func(t *testing.T) {
```
zhangchiqing, line 44:
```suggestion
		t.Run("inserting 3 results at different height each has a ServiceEventCommit", func(t *testing.T) {
```

state/protocol/badger/mutator.go
```diff
@@ -556,6 +556,14 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 
 	// track service event driven metrics and protocol events that should be emitted
 	var events []func()
+
+	// track service events to index
+	var serviceEvents []struct {
```
zhangchiqing, line 561: This type is defined quite far away from where it's used. Better to move it close.

state/protocol/badger/mutator.go
```diff
@@ -556,6 +556,14 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 
 	// track service event driven metrics and protocol events that should be emitted
 	var events []func()
+
+	// track service events to index
```
zhangchiqing, line 560:
```suggestion
	// track service events to index results by event type and height
```

zhangchiqing (approved):

state/protocol/badger/mutator.go
m4ksio: https://github.com/onflow/flow-go/pull/3605#discussion_r1029935694 fixed this as well

state/protocol/badger/mutator.go
```diff
@@ -629,6 +637,34 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 		m.metrics.EpochEmergencyFallbackTriggered()
 	}
 
+	// FOUR AND A HALF: Index sealed service events by height
+	for _, seal := range block.Payload.Seals {
```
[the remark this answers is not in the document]
m4ksio: Not really - we iterate there only over `parent.Seals`, and we want to index all the blocks.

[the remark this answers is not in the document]
m4ksio: Hi Jordan! Thanks for very detailed answer, it clarified a lot, but I still have some questions.

> that service event comes into effect in the protocol state

Is "comes into effect" actually defined? It seems like it would depend on the even type. Or not?

> ake Leo's suggestion of moving this logic into Step 4

Do you both mean step 3? I don't see how this fit into step 4.

storage/badger/operation/events.go
```diff
@@ -3,11 +3,34 @@
...
 )
 
+// prefixes used to distinguish service events
+const (
+	ServiceEventPrefixSetup  = byte(1)
+	ServiceEventPrefixCommit = byte(2)
+)
+
+// serviceEventTypeToPrefix returns byte prefix for given service event type
+func serviceEventTypeToPrefix(eventType string) byte {
```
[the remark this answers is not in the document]
m4ksio: My thinking was that this is a storage layer which shouldn't really be concerned with validation. Should we maybe add extra validation step before?

state/protocol/badger/mutator.go
```diff
@@ -629,6 +630,41 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+			return fmt.Errorf("could not retrieve result (id=%x) for seal (id=%x) for service events indexing: %w", seal.ResultID, seal.ID(), err)
+		}
+		blockHeader, err := m.headers.ByBlockID(result.BlockID)
+		if err != nil {
+			return fmt.Errorf("could not retrieve header (id=%x) for result (id=%x) for service events indexing: %w", result.BlockID, result.ID(), err)
+		}
+
+		for _, serviceEvent := range result.ServiceEvents {
+			// it is possible, however very unlikely, that there are multiple service events of the same type in a single result.
+			// This will add multiple identical entries here, but the storage layer will only keep one index entry per combination of
```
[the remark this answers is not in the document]
m4ksio: Yep, good idea!

state/protocol/badger/mutator_test.go
```diff
@@ -143,7 +143,13 @@ func TestSealedIndex(t *testing.T) {
 		require.NoError(t, err)
 
 		// block 2(result B1)
+		result, _, err := rootSnapshot.SealedResult()
+		require.NoError(t, err)
+
+		serviceEvent := result.ServiceEvents[0]
+
 		b1Receipt := unittest.ReceiptForBlockFixture(b1)
+		b1Receipt.ExecutionResult.ServiceEvents = []flow.ServiceEvent{serviceEvent}
```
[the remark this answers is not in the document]
m4ksio: I would say that this event being invalid is even better for the purpose - since the index is supposed to simply index events, not consider their type and then relevance

storage/badger/operation/results_test.go
```diff
@@ -27,3 +29,72 @@ func TestResults_InsertRetrieve(t *testing.T) {
...
+			require.NoError(t, err)
+
+			err = db.Update(IndexExecutionResultByServiceEventTypeAndHeight(result2.ID(), eventType, height2))
+			require.NoError(t, err)
+
+			err = db.Update(IndexExecutionResultByServiceEventTypeAndHeight(result3.ID(), eventType, height3))
+			require.NoError(t, err)
+		})
+
+		t.Run("retrieve exact height match", func(t *testing.T) {
```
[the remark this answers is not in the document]
m4ksio: Yes, good point, easily fixed by making update part of outer test

storage/badger/operation/events.go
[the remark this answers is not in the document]
jordanschalm: Yeah, I see where you're coming from, I agree the storage layer shouldn't be responsible for this kind of validation in general. But I think it should also be reasonably safe against panics. In this case validation is partly coupled to safety against crashes. 

We should definitely add a validation step before the event reaches this point, otherwise this will be guaranteed to represent a vulnerability to crashes. My point is more about approaching this defensively in general, so that even if one code-path leading to this function fails to implement proper validation, we can handle the resulting failure more gracefully than a panic. 

I think a reasonable general rule of thumb is, if the input which might cause a panic (if unvalidated) is expected to come from an external source, then we should use errors, and if the input is expected to come from an internal source, then it's safe to use panics at a lower level like this example.

state/protocol/badger/mutator_test.go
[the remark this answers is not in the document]
jordanschalm: > index is supposed to simply index events, not consider their type and then relevance

That's a good point. I was thinking that we would not want to index invalid service events, but I can see why we would want to do that. Though it will mean that, the code querying this index might need to be able to handle the case where querying the latest event results in an event that was considered invalid when it was processed.

storage/badger/operation/events.go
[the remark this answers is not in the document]
m4ksio: Yes, you are right.

state/protocol/badger/mutator.go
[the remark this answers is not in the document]
m4ksio: Ok, I have applied suggestions

state/protocol/badger/mutator_test.go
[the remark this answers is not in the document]
m4ksio: Yes, I would assume the code using index must be aware of the possibility

storage/badger/operation/results.go
```diff
@@ -26,6 +26,21 @@ func IndexExecutionResult(blockID flow.Identifier, resultID flow.Identifier) fun
...
+// this index allows us to answer questions such as what is the latest EpochSetup service event up to
+// block height 1000.
+//
+// why indexing result ID rather than service events? because the service event is included in a result,
+// the indexed result id could help join the result and find the service event
+//
+// why using <service_event_type+sealed_height> as key rather than <sealed_height + service_event_type>?
+// because having service_event_type key first allows us to scan through the index and
+// filter by service event type first, and then find the result for highest height,
+// see LookupLastExecutionResultForServiceEventType
```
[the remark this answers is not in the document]
m4ksio: Not really, we simply override the value, assuming that there won't be two different seals for one height - which I think is quite valid assumption

state/protocol/badger/mutator.go
```diff
@@ -557,17 +557,38 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+			// height and event type, essentially disregarding duplicates. Since the situation is very unlikely and other layer will handle
+			// this situation gracefully we leave it like this instead of writing deduplication logic here.
+			serviceEvents = append(serviceEvents, struct {
+				height   uint64
+				resultID flow.Identifier
+				event    flow.ServiceEvent
+			}{
+				height:   parent.Header.Height,
+				resultID: result.ID(),
```
zhangchiqing, line 584: We don't want to compute the same ID multiple times.
```suggestion
				resultID: seal.ResultID,
```

state/protocol/badger/mutator.go
```diff
@@ -556,17 +556,39 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+			// it is possible, however very unlikely, that there are multiple service events of the same type in a single result.
+			// This will add multiple identical entries here, but the storage layer will only keep one index entry per combination of
+			// height and event type, essentially disregarding duplicates. Since the situation is very unlikely and other layer will handle
+			// this situation gracefully we leave it like this instead of writing deduplication logic here.
+			serviceEvents = append(serviceEvents, struct {
+				height   uint64
+				resultID flow.Identifier
+				event    flow.ServiceEvent
+			}{
+				height:   parent.Header.Height,
```
zhangchiqing, line 583: It's still not clear to me why the result is indexed by the height of the parent block here?

Shouldn't we index a result for a height where it comes to effect?

We'd better add some comments or examples to explain what height is being used, and why.

For instance, if a service event, which indicates VersionUpdate from v1 to v2 at height 2000, is included in a result included in block 100. Do we index the VersionUpdate event by height 100 or 2000?

m4ksio: That's to be consistent with how service event effectiveness is described in https://github.com/onflow/flow-go/blob/master/state/protocol/badger/mutator.go#L759 - so after a QC is there for finalized block.

This PR shouldn't contain any mentions of VersionUpdate events, but general purpose and reasoning has been explained https://github.com/onflow/flow-go/pull/3605/files#diff-c760f8136a91a3d5ffbcfa069cf895ce0e65f4a60b1df9ad8cda5dbd04211c55R29

jordanschalm: I think the index height here should be `header.Height` - the height of the block being inserted. 

If a service event is coming into effect at the block we are processing, then that block has the form of block D (definition linked below). The parent of D (C) is the block containing the relevant seal, so while processing D we iterate over C's seals. But the point at which the service event comes into effect is still block D (not C), so we should use `header.Height` (D's height) rather than `parent.Height` (C's height).

https://github.com/onflow/flow-go/blob/b8561103450733edaa270131932dc555e8513360/state/protocol/badger/mutator.go#L749-L757

m4ksio: Fixed

state/protocol/badger/state.go
```diff
@@ -164,6 +170,21 @@ func (state *State) bootstrapSealingSegment(segment *flow.SealingSegment, head *
...
+			// SealingSegment might not have all the blocks referenced by results.
+			// If the block is missing we skip it while indexing, and that's OK because
+			// we are bootstrapping and service event index before the segment shouldn't matter
+			block, has := blocksById[result.BlockID]
+			if !has {
+				continue
+			}
+
+			for _, serviceEvent := range result.ServiceEvents {
+				err := transaction.WithTx(operation.IndexExecutionResultByServiceEventTypeAndHeight(result.ID(), serviceEvent.Type, block.Header.Height))(tx)
```
jordanschalm, line 183: This is indexing the old way - now it isn't consistent with how service events are indexed when finalized.

storage/badger/operation/results_test.go
```diff
@@ -27,3 +29,78 @@ func TestResults_InsertRetrieve(t *testing.T) {
...
+		err = db.Update(IndexExecutionResultByServiceEventTypeAndHeight(result3.ID(), eventType, height3))
+		require.NoError(t, err)
+
+		// insert result 2 again to make sure we tolerate duplicates
+		// it is possible for two or more events of the same type to be from the same height
+		err = db.Update(IndexExecutionResultByServiceEventTypeAndHeight(result2.ID(), eventType, height2))
+		require.NoError(t, err)
+
+		t.Run("retrieve exact height match", func(t *testing.T) {
+			//t.Parallel()
```
jordanschalm, line 60: please uncomment or remove these before merging

storage/badger/operation/results.go
```diff
@@ -26,6 +26,28 @@ func IndexExecutionResult(blockID flow.Identifier, resultID flow.Identifier) fun
...
+func IndexExecutionResultByServiceEventTypeAndHeight(resultID flow.Identifier, eventType string, blockHeight uint64) func(*badger.Txn) error {
```
jordanschalm, line 41:
```suggestion
//
// Input eventType strings must be valid service event types, and be known to serviceEventTypeToPrefix.
// No errors are expected during normal operation.
func IndexExecutionResultByServiceEventTypeAndHeight(resultID flow.Identifier, eventType string, blockHeight uint64) func(*badger.Txn) error {
```

jordanschalm:
Thanks for the updates, once we update the [bootstrapping indexing](https://github.com/onflow/flow-go/pull/3605#discussion_r1038266718), this is good to merge from my perspective.

state/protocol/badger/state.go
jordanschalm: Adding more context:

Given a segment of blocks like:
```
A<-B(Result_A)<-C(Seal_A)<-D(QC_C)
```

In the bootstrapping here, we are indexing a service event emitted in block A, by A.height.

However, when we process blocks while the process is running ([here](https://github.com/onflow/flow-go/pull/3605/files#diff-c79d3fc5d32f153917b7cd923ab4ba64ef22f12268ddd13f353a06391f68a96dR574-R586)) we are indexing by D.height , like we discussed. 

So we are indexing inconsistently between bootstrapping and ongoing block processing. 

To index the same way in both places, we would need to look through the seals contained in the block paylods of the sealing segment used to bootstrap. In particular, if the sealing segment contains an execution result where a service event was emitted, but the result hasn't been sealed yet, we shouldn't index it during bootstrapping.

For example, if `Result_X` has a service event, and we're processing the sealing segment `W<-X<-Y<-Z` during bootstrapping, we should not index that service event yet, since it hasn't been sealed.
```
W<-X(Result_W)<-Y(Result_X)<-Z(Seal_W)
```

jordanschalm (approved):
Nice work. Main comments:
* I think we should keep these changes on a feature branch until we have a clearer solution for solving the service event initialization problem
* Because there are still some problems with the bootstrapping changes, and that ties into the main unknowns about how we'll initialize the service event index, suggesting that we remove that section to unblock this PR as a starting point.

m4ksio:
Superseded by https://github.com/onflow/flow-go/pull/3736

PR #3606, opened 2022-11-22
Use generics for `flow.GetIDs`, clean up `mempool` package

jordanschalm:
* Modifies `flow.GetIDs` to use generics rather than reflection
* Some mempool implementations had exposed a `Hash` method, which was intended to provide a fingerprint of the current state of the mempool's stored list of entities - this has been removed because:
  * it was unused
  * it was passing a map to `GetIDs`, which would have panicked and would not have produced a deterministic ordering of keys to be hashed anyway


kc1116 (approved):

peterargue (approved):

jordanschalm:
bors merge

Issue #3547, opened 2022-11-09
[Execution Node] Concurrent Execution


SaveTheRbtz:
- [x] Split block computation into preprocess() and process()
  - [x] Convert FVM into executor-style objects.
- [x] Optimistic concurent preprocessing.
  - [x] Block computer refactoring to execute blocks in pipeline
  - [x] Splitting process() function - inject basic concurrency management layer which handles scheduling the executors and rebasing the execution state when collisions are detected.
- [x] Add concurrency management layer to the execution.

Initial discussion: https://www.notion.so/dapperlabs/Concurrent-Execution-Sync-minutes-66bbe575748940ae906517856d7554ac

pattyshack:
(copy/paste from #4490description):

WARNING: it is currently unsafe to enable concurrent execution because it exposes a cadence type checker bug, which in turn causes non-deterministic execution.

Unsurprisingly, enabling concurrent execution today will slightly slow down block execution since transaction conflict rate is currently 100%. Transactions access pattern will need to be restructure in order to take advantage of concurrent execution, but it is reassuring from a correctness standpoint to see that conflicting transactions are correctly detected / retried. (For performance expectations, see "Concurrency Control Performance Modeling: Alternatives and Implications", 1987).

PR #3607, opened 2022-11-22
dedup executeCollection logic

pattyshack:
executeCollection and executeSystemCollection have identical logic, with only minor tracing and logging differences.

gh: #3547

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -31,6 +31,14 @@ const (
 	SystemChunkEventCollectionMaxSize = 256_000_000 // ~256MB
 )
 
+type collectionItem struct {
+	*entity.CompleteCollection
+
+	ctx fvm.Context
+
+	isSystemCollection bool
+}
+
 // VirtualMachine runs procedures
 type VirtualMachine interface {
 	Run(fvm.Context, fvm.Procedure, state.View) error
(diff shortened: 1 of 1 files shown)


zhangchiqing (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -276,96 +288,111 @@ func (e *blockComputer) executeSystemCollection(
...
-		err := e.executeTransaction(txBody, colSpan, collectionView, txCtx, collectionIndex, txIndex, res, false)
+		err := e.executeTransaction(txBody, colSpan, collectionView, collection.ctx, collectionIndex, txIndex, res, false)
 		txIndex++
 		if err != nil {
 			return txIndex, err
 		}
 	}
+
+	if collection.isSystemCollection {
+		systemChunkTxResult := res.TransactionResults[len(res.TransactionResults)-1]
```
SaveTheRbtz, line 369: do we need to be defensive here and verify that len is always == 1 here?

engine/execution/computation/computer/computer.go
```diff
@@ -276,96 +288,111 @@ func (e *blockComputer) executeSystemCollection(
...
+		if systemChunkTxResult.ErrorMessage != "" {
+			// This log is used as the data source for an alert on grafana.
+			// The system_chunk_error field must not be changed without adding
+			// the corresponding changes in grafana.
+			// https://github.com/dapperlabs/flow-internal/issues/1546
+			e.log.Error().
+				Str("error_message", systemChunkTxResult.ErrorMessage).
+				Hex("block_id", logging.Entity(collection.ctx.BlockHeader)).
+				Bool("system_chunk_error", true).
```
SaveTheRbtz, line 378: you can double-emit logs with both old and new names here so that we would be able to do the migration.

SaveTheRbtz (approved):

janezpodhostnik (approved):

engine/execution/computation/computer/computer.go
pattyshack: we can't check for len == 1 since TransactionResults holds results for the entire block

engine/execution/computation/computer/computer.go
pattyshack: add system_transaction double logging everywhere.  i'll look into renaming the variables later.

pattyshack:
bors merge

PR #3608, opened 2022-11-22
[Logging] Add log throttle for debug/trace level events

peterargue:
This PR uses `zerolog.BurstSampler`s to rate limit `Debug` and `Trace` level events. Any events that exceed the limit will be dropped. This helps protect nodes from significant performance issues when debug logging is enabled.

Note: This is being added to avoid liveness issues when a large volume of logs are generated. It may result in lost log events which would require further changes to logging to prevent.

This PR also sets the min log level to `Trace` which allows enabling trace events via the `--loglevel` flag or admin command.

### DIFF
--- a/cmd/node_builder.go
+++ b/cmd/node_builder.go
@@ -281,7 +281,7 @@ func DefaultBaseConfig() *BaseConfig {
 		secretsdir:       NotSet,
 		secretsDBEnabled: true,
 		level:            "info",
-		debugLogLimit:    1000,
+		debugLogLimit:    2000,
 
 		metricsPort:         8080,
 		tracerEnabled:       false,
--- a/cmd/node_builder.go
+++ b/cmd/node_builder.go
@@ -151,6 +151,7 @@ type BaseConfig struct {
 	secretsDBEnabled            bool
 	InsecureSecretsDB           bool
 	level                       string
+	debugLogLimit               uint
 	metricsPort                 uint
 	BootstrapDir                string
 	profilerConfig              profiler.ProfilerConfig
@@ -280,6 +281,7 @@ func DefaultBaseConfig() *BaseConfig {
 		secretsdir:       NotSet,
 		secretsDBEnabled: true,
 		level:            "info",
+		debugLogLimit:    1000,
 
 		metricsPort:         8080,
 		tracerEnabled:       false,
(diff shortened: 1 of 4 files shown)


cmd/node_builder.go
```diff
@@ -283,6 +284,7 @@ func DefaultBaseConfig() *BaseConfig {
 		secretsdir:       NotSet,
 		secretsDBEnabled: true,
 		level:            "info",
+		debugLogLimit:    1000,
```
peterargue, line 287: I'm not sure what an appropriate default is.

cmd/scaffold.go
```diff
@@ -573,11 +574,22 @@ func (fnb *FlowNodeBuilder) initLogger() {
 	// configure logger with standard level, node ID and UTC timestamp
 	zerolog.TimeFieldFormat = time.RFC3339Nano
 	zerolog.TimestampFunc = func() time.Time { return time.Now().UTC() }
+
+	// Drop all log events that exceed this rate limit
```
peterargue, line 578: another option is to sample N events/sec after exceeding the limit

cmd/node_builder.go
zhangchiqing: maybe this is a bit low. I just checked access-001  is producing 500 per second right now

peterargue: looking over the logs from the LN outage, it looks like nodes were peaking around 500 events/sec. It's hard to say. I updated the limit to 2000, but operators will likely need to tune this for their hardware

cmd/scaffold.go
```diff
@@ -573,11 +574,22 @@ func (fnb *FlowNodeBuilder) initLogger() {
...
+	throttledSampler := &zerolog.BurstSampler{
+		Burst:  fnb.BaseConfig.debugLogLimit,
+		Period: time.Second,
+	}
+
```
yhassanzadeh13, line 583: Please also add a similar `throttledSampler` dedicated for the [score option](https://github.com/onflow/flow-go/blob/57957a344efde12efffabbcbdb2d7b9e7ffd6edd/network/p2p/scoring/score_option.go#L96). It currently is triggering the logging per received message, but the content of log is mostly the same under the normal and honest operation. I think a burst of 100 within one second is sufficient for the score option. In this way we also safeguard the global logger against the score option spamming it. This is the only place of my knowledge that we experienced spamming at the networking layer.

yhassanzadeh13 (approved):
👍

cmd/scaffold.go
peterargue: b21e90d

zhangchiqing (approved):

peterargue:
bors merge

PR #3609, opened 2022-11-23
Reproduce `engine/collection/test` failure with error logs

jordanschalm:


jordanschalm:
Found the error by replacing fatal logs and using verbose test flag -> closing in favour of https://github.com/onflow/flow-go/pull/3610, which should address the test failure.

PR #3577, opened 2022-11-15
Small improvements to error types and messages in block extension logic

jordanschalm:
* Adds `UnverifiableExtensionError` to differentiate extensions which aren't certainly invalid, and cannot be processed (for cluster consensus `ReferenceBlockID`
* Updates some error message and documentation wording

PR #3610, opened 2022-11-23
Handle unverifiable input error in cluster compliance core

jordanschalm:
Add missing handling for error type added in https://github.com/onflow/flow-go/pull/3577


gomisha (approved):

yhassanzadeh13 (approved):

zhangchiqing:
Do we need to add the same logic in [consensus compliance core](https://github.com/onflow/flow-go/blob/master/engine/consensus/compliance/core.go#L248)?

PR #3611, opened 2022-11-23
Only override meter parameters for transaction/script body execution

pattyshack:
The transaction (/script) procedure originally uses two separate transaction state to process a single request.  This merges the two transaction states into one, and only override the meter params for body execution.  The rest of the transaction (i.e., auth check, seq num check, meter override parsing) uses basic/non-overridden meter setting.

gh: #3547 #3102

### DIFF
--- a/fvm/executionParameters.go
+++ b/fvm/executionParameters.go
@@ -18,6 +18,9 @@ import (
 	"github.com/onflow/flow-go/fvm/utils"
 )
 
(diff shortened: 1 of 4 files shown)


SaveTheRbtz (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3613, opened 2022-11-23
[Execution] Break blobservice out to own component in en builder

peterargue:
Backports https://github.com/onflow/flow-go/pull/3497 from v0.28 to master

This fixes the initialization order of the blobservice and uploaders in the execution node builder. Blobservice is needed by both the provider engine and GCP uploader, and the uploaders are also needed by the provider

### DIFF
--- a/cmd/execution_builder.go
+++ b/cmd/execution_builder.go
@@ -424,6 +424,10 @@ func (exeNode *ExecutionNode) LoadProviderEngine(
 	module.ReadyDoneAware,
 	error,
 ) {
+	if exeNode.blobService == nil {
+		return nil, errors.New("blob service is not initialized")
+	}
+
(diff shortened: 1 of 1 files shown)


cmd/execution_builder.go
```diff
@@ -422,7 +433,7 @@ func (exeNode *ExecutionNode) LoadProviderEngine(
 		node.Logger,
 		providerMetrics,
 		execution_data.DefaultSerializer,
-		bs,
+		exeNode.blobService,
```
zhangchiqing, line 436: Let's add a check to ensure this module has been set and is not `nil`. Type system can't ensure this, so we'd better add explicit check.

zhangchiqing (approved):

m4ksio (approved):

peterargue:
bors merge

PR #3555, opened 2022-11-10
[v0.28] Backport #3369 -  Admin command for dynamic config updates

jordanschalm:
Backport https://github.com/onflow/flow-go/pull/3369 to v0.28

This will enable block rate delay updates without rolling restarts going forward.

PR #3615, opened 2022-11-23
[Profiler] Start profiler asynchronously

peterargue:
This is a revision of https://github.com/onflow/flow-go/pull/3503 which updated `scaffold.go` to start the profiler as one of the first components. Due to changes needed for the admin command, this PR changes the implementation to use `DependableComponent` so the profiler will start asynchronously.

Before this PR, the profiler was started as a `Component` that was added after all other node specific components. This meant that the profiler would delay startup until after all other components were loaded. On nodes that perform a lot of startup logic like execution nodes, the profiler would not be able to capture any of that data.

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -711,7 +711,7 @@ func (fnb *FlowNodeBuilder) initProfiler() {
 			func() uint { return uint(runtime.MemProfileRate) },
 			func(r uint) error { runtime.MemProfileRate = int(r); return nil }),
 	).Msg("could not register profiler setting")
-	// There is no way to get the current block progile rate so we keep track of it ourselves.
+	// There is no way to get the current block profile rate so we keep track of it ourselves.
 	currentRate := new(uint)
 	fnb.MustNot(
 		fnb.ConfigManager.RegisterUintConfig(
@@ -726,9 +726,10 @@ func (fnb *FlowNodeBuilder) initProfiler() {
 			func(r uint) error { _ = runtime.SetMutexProfileFraction(int(r)); return nil }),
 	).Msg("could not register profiler setting")
 
-	fnb.Component("profiler", func(node *NodeConfig) (module.ReadyDoneAware, error) {
+	// registering as a DependableComponent with no dependencies so that it's started immediately on startup
+	fnb.DependableComponent("profiler", func(node *NodeConfig) (module.ReadyDoneAware, error) {
 		return profiler, nil
-	})
+	}, NewDependencyList())
 }
 
 func (fnb *FlowNodeBuilder) initDB() {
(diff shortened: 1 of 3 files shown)


cmd/scaffold.go
```diff
@@ -1113,7 +1114,7 @@ func (fnb *FlowNodeBuilder) handleComponents() error {
 	// Run all components
 	for _, f := range fnb.components {
 		// Components with explicit dependencies are not started serially
-		if f.dependencies != nil && len(f.dependencies.components) > 0 {
+		if f.dependencies != nil {
```
m4ksio, line 1117: wouldn't it be better to use `len(f.dependencies) > 0` which would handle both nil and empty slice

m4ksio (approved):

cmd/scaffold.go
peterargue: `dependencies` is a struct pointer so I can't directly check the length of `f.dependencies.components` if `f.dependencies` is nil

cmd/scaffold.go
```diff
@@ -726,9 +726,10 @@ func (fnb *FlowNodeBuilder) initProfiler() {
 			func(r uint) error { _ = runtime.SetMutexProfileFraction(int(r)); return nil }),
 	).Msg("could not register profiler setting")
 
-	fnb.Component("profiler", func(node *NodeConfig) (module.ReadyDoneAware, error) {
+	// registering as a DependableComponent with no dependencies so that it's started immediately on startup
```
zhangchiqing, line 729:
```suggestion
	// registering as a DependableComponent with no dependencies so that it's started immediately on startup
	// without being blocked by other component's Ready()
```

zhangchiqing (approved):
Nice 👍

SaveTheRbtz:
Please note that profile uploader needs access to `fnb.RootChainID` (which is set by the storage, IIRC) so it can't be started too early:

https://github.com/onflow/flow-go/blob/31b14721bf1aa3f18556600948a0b6692d9e2c88/cmd/scaffold.go#L660

 I think we broke it recently and profiles are marked as "unknown" instead of mainnet20 on recent builds:

Screen Shot 2022-11-24 at 7 45 13 PM

peterargue:
yea, looks like that happened in this [change](https://github.com/onflow/flow-go/pull/3555/files#diff-cae89c2f7daa439631988089cdd51c551e4c25db2452885527785d0774235041R609) because https://github.com/onflow/flow-go/pull/3503 moved `initProfiler` to the `Initialize()` step.

That should be fixed in this PR since it keeps the original initalization order, but just changes when the component actually starts (which is after `fnb.RootChainID` is set)

peterargue:
bors merge

PR #3616, opened 2022-11-23
[Execution] Disable block uploads by default

peterargue:
Backports github.com/onflow/flow-go/pull/3504 and https://github.com/onflow/flow-go/pull/3518 from `v0.28`

Currently, the uploader is enabled by default on all nodes, and only skipped because there are no configured uploaders. This causes nodes that don't have uploads enabled to check all blocks in their dbs for blocks to upload

### DIFF
--- a/engine/execution/computation/manager.go
+++ b/engine/execution/computation/manager.go
@@ -41,7 +41,7 @@ const (
 	ReusableCadenceRuntimePoolSize = 1000
 )
 
-var uploadEnabled = true
+var uploadEnabled = false
 
 func SetUploaderEnabled(enabled bool) {
 	uploadEnabled = enabled
--- a/cmd/execution_builder.go
+++ b/cmd/execution_builder.go
@@ -301,6 +301,8 @@ func (exeNode *ExecutionNode) LoadGCPBlockDataUploader(
 
 		exeNode.blockDataUploaders = append(exeNode.blockDataUploaders, retryableUploader)
 
+		computation.SetUploaderEnabled(true)
+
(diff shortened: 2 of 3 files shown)


m4ksio (approved):

engine/execution/computation/manager.go
```diff
@@ -41,7 +41,7 @@ const (
 	ReusableCadenceRuntimePoolSize = 1000
 )
 
-var uploadEnabled = true
+var uploadEnabled = false
```
zhangchiqing, line 44: I think we can get rid of this variable by checking whether `len(e.uploaders) > 0`. If uploaded is enabled, there must be at least one `e.uploaders` .

engine/execution/computation/manager.go
```diff
@@ -41,7 +41,7 @@ const (
...
 
 func SetUploaderEnabled(enabled bool) {
```
zhangchiqing, line 46: I wonder if `computation manager` needs to know about upload at all. That seems not the responsibility of computation manager to me. 

Should we let another module to handle and decide whether and how to upload?

cc @m4ksio

engine/execution/computation/manager.go
peterargue: That flag is there to support toggling uploads from the admin server

engine/execution/computation/manager.go
peterargue: I opened a different PR to implement this: https://github.com/onflow/flow-go/pull/3676

PR #3618, opened 2022-11-24
Simplify adding transaction result to ComputationResult

pattyshack:
All the necessary transaction result info are in TransactionProcedure. There's no need to set the ComputationResult's fields using individual calls.

gh: #3547

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -442,16 +442,6 @@ func (e *blockComputer) executeTransaction(
 			err)
 	}
 
-	txResult := flow.TransactionResult{
-		TransactionID:   tx.ID,
-		ComputationUsed: tx.ComputationUsed,
-		MemoryUsed:      tx.MemoryEstimate,
-	}
-
-	if tx.Err != nil {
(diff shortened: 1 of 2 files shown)


m4ksio (approved):
I don't see changes that might be problematic for DPS (such as new fields addition), so approving. A simple test would be great though

zhangchiqing (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -464,28 +454,23 @@ func (e *blockComputer) executeTransaction(
...
 		Uint64("memAlloc", memAllocAfter-memAllocBefore).
 		Int64("timeSpentInMS", time.Since(startedAt).Milliseconds()).
 		Logger()
 
 	if tx.Err != nil {
 		lg.Info().
-			Str("error_message", txResult.ErrorMessage).
+			Str("error_message", tx.Err.Error()).
 			Uint16("error_code", uint16(tx.Err.Code())).
 			Msg("transaction executed failed")
```
SaveTheRbtz, line 475: nit: "execution failed".

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3619, opened 2022-11-24
[Bootstrap] Improve error handling in node scaffold

peterargue:
This PR reduces the number of calls to `Fatal()` during node bootstrapping by returning errors to the caller so they can be handled appropriately. All of the errors returned are ultimately returned from the node builder's `Build()` method and result in a final call to `Fatal()` in the node's main method. The only exception is the case of the consensus follower library, which allows the implementor to handle the error directly.


zhangchiqing (approved):

tarakby (approved):

peterargue:
bors merge

peterargue:
bors retry

Issue #3614, opened 2022-11-23
[Execution] Making state migration concurrent by account


zhangchiqing:
## Problem Definition

The state migration happens during spork. It takes a list of payloads and [migrate](https://github.com/onflow/flow-go/blob/master/ledger/complete/ledger.go#L387) and [load into a new trie. ](https://github.com/onflow/flow-go/blob/master/ledger/complete/ledger.go#L422)

This process is currently single threaded. We could make it concurrent to speed up.

## Which part can be done concurrently
Among all the payloads to be migrated, there are two types of payloads:
- payloads belong to an account
- payloads belong a global state. (There are two instance of this type. The first is the global UUID generator, the second is the global account creator)

What can be migrated concurrently are the payloads belong to different account.

So the state migration will be break into the following steps:
- Iterate through each payload, group the payloads that belong to the same account
- Using workers to concurrently migrate different payload group
- All payloads within a payload group belong to the same account, so migrate them together using cadence runtime
- add the migrated payload to a newly constructed trie. The order of the payloads being added to the trie doesn't matter, it should produce the same trie root hash

-

zhangchiqing:
I did some baseline benchmark. 

Right now, without any concurrency optimization or any migration action, if we have a migration that go over all the payloads, the state extraction will take extra time on the following tasks:
1. 7 mins to group the payloads by account address
2. 2 mins to migrate without any migration action
3. 9 mins to preparing all the paths for all the migrated payloads
4. 16 mins to construct a new trie with the paths and the migrated payloads

```
1:26AM INF extracting state by block ID: 1ed665b91dc8f888824f43027f4bc85c5d61108db4b7141c2ec64197690a1fbc

1:26AM INF Extracting state from /var/flow/data/execution, exporting root checkpoint to /var/flow/v5-1/root.checkpoint, version: 6
1:26AM INF Block state commitment: e310cfebe945dfd997dc35987dc4b00a3bbbdbf8c3c4e304ec22eb31f6c697bf from /var/flow/data/execution, output dir: /var/flow/v5-1
1:26AM INF init WAL
1:26AM INF init ledger
1:26AM INF loading checkpoint with WAL from 0 to 5191 ledger_mod=diskwal
1:26AM INF loading checkpoint checkpoint=5111 ledger_mod=diskwal
1:26AM INF reading v5 checkpoint file ledger_mod=diskwal
1:27AM INF reading trie nodes completion percentage: 10 percent ledger_mod=diskwal
[...]

PR #3621, opened 2022-11-25
Optimize state migration

zhangchiqing:
For https://github.com/onflow/flow-go/issues/3614

This PR implemented a template for running state migration concurrently based on account.

It first group all payloads by account, and then migrate the payloads of different account concurrently.

The example migration reads the account storage usage and update the register in account status with new storage usage.

Since after migration, the storage usage is mostly likely changed, this example migration could be reused in the real migration.


cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return groups, nil
+}
+
+// AccountMigrator takes all the payloads that belong to the given account
+// and return the migrated payloads
+type AccountMigrator func(account string, payloads []ledger.Payload, pathFinderVersion uint8) ([]ledger.Payload, error)
+
+// MigrateByAccount teaks a migrator function and all the payloads, and return the migrated payloads
+func MigrateByAccount(migrator AccountMigrator, allPayloads []ledger.Payload, pathFinderVersion uint8) (
```
zhangchiqing, line 62: This function takes a migrator function, so that the caller can provide a different migrator function for different migration scenarios and reuse this function for running migration concurrently.

ledger/complete/ledger.go
```diff
@@ -405,15 +407,7 @@ func (l *Ledger) ExportCheckpointAt(
 			payloadSize = newPayloadSize
 		}
 
-		l.logger.Info().Msgf("creating paths for %v payloads", len(payloads))
-
-		// get paths
-		paths, err := pathfinder.PathsFromPayloads(payloads, targetPathFinderVersion)
```
zhangchiqing, line 411: Iterating through the paths again takes extra time. 

Since the `migrate` is going to return the payloads, we could let it return the path at the same time, so that computing the path could also be done concurrently.

cmd/util/ledger/migrations/account_migration.go
```diff
@@ -0,0 +1,94 @@
...
+	}, nil
+}
+
+func payloadSize(key ledger.Key, payload ledger.Payload) (uint64, error) {
+	ik, err := KeyToStorageInteractionKey(key)
+	if err != nil {
+		return 0, err
+	}
+
+	return meter.GetStorageKeyValueSizeForTesting(ik, payload.Value()), nil
```
zhangchiqing, line 33: In order to get the cached storage size for the payload, I need to call this function which is a testing function. Would it make sense to remove the `Testing` in the name, and make it public? @pattyshack .

Also , is this the right function to compute the storage size of a given payload? @janezpodhostnik

cmd/util/ledger/migrations/account_migration.go
```diff
@@ -0,0 +1,94 @@
...
+		return meter.StorageInteractionKey{}, err
+	}
+	return meter.StorageInteractionKey{
+		Owner: id.Owner,
+		Key:   id.Key,
+	}, nil
+}
+
+func payloadSize(key ledger.Key, payload ledger.Payload) (uint64, error) {
+	ik, err := KeyToStorageInteractionKey(key)
```
zhangchiqing, line 28: we can't directly get the cached storage size of a given payload. Instead, the method to get storage size of a payload requires a storage interaction key, that's why we need to first convert the key into a storage interaction key here.

cmd/util/ledger/migrations/account_migration.go
```diff
@@ -0,0 +1,94 @@
...
+				return nil, fmt.Errorf("could not parse account status: %w", err)
+			}
+
+		}
+
+		size, err := payloadSize(key, payload)
+		if err != nil {
+			return nil, err
+		}
+		totalSize += size
```
zhangchiqing, line 64: Simply sum up the size of each payload, which would be the storage used size.

cmd/util/ledger/migrations/account_migration.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	if status == nil {
+		return nil, fmt.Errorf("could not find account status for account %v", account)
+	}
+
+	// update storage used
+	status.SetStorageUsed(totalSize)
```
zhangchiqing, line 72: In this example, the totalSize is supposed to be equal to the cached value from `status.StorageUsed()`. 

But in real migration, some payload will be removed, and some payload might be added, in that case, the StorageUsed might be different.

cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -86,8 +87,10 @@ func extractExecutionState(
 	newState := ledger.State(targetHash)
 
 	if migrate {
-		migrations = []ledger.Migration{}
-
+		// add migration here
+		migrations = []ledger.Migration{
+			mig.MigrateAccountUsage,
```
zhangchiqing, line 92: To be removed

module/util/log.go
```diff
@@ -0,0 +1,20 @@
+package util
+
+import "github.com/rs/zerolog"
+
+func LogProgress(msg string, total int, logger *zerolog.Logger) func(current int) {
+	lookup := make(map[int]int)
+	// report every 10 percent
+	for i := 1; i < 10; i++ { // [1...9]
+		lookup[total/10*i] = i * 10
```
pattyshack, line 9: `lookup[total*i/10] = i * 10`

module/util/log.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+	if total > 0 {
+		lookup[total-1] = 100
+	}
+	return func(current int) {
+		percentage, ok := lookup[current]
```
pattyshack, line 15: if current is for whatever reason greater than total, than this stops logging.  instead of using a lookup map, maybe just do this

```
func LogProgress(msg string, total int, logger *zerolog.Logger) func(current int) {
	logThreshold := float64(0)
	return func(current int) {
		percentage := float64(100)
		if total > 0 {
			percentage = (float64(current+1) / float64(total)) * 100. // current+1 assuming zero based indexing 
		}

		if percentage >= logThreshold+10 {
			logger.Info().Msgf("%s completion percentage: %v percent", msg, percentage)
			logThreshold += 10
		}
	}
}
```

ledger/complete/ledger.go
pattyshack: this seems wasteful.  you're recomputing the paths on migration iteration, and throwing away everything except the last iteration.  just parallelize the path computing here instead.

cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,210 @@
...
+	k, err := p.Key()
+	if err != nil {
+		return "", false, fmt.Errorf("could not find key for payload: %w", err)
+	}
+	id, err := KeyToRegisterID(k)
+	if err != nil {
+		return "", false, fmt.Errorf("error converting key to register ID")
+	}
+	if len([]byte(id.Owner)) != flow.AddressLength {
+		return "", false, nil
```
pattyshack, line 28: is this check safe? if i remember correctly address "" is a reserved address.

cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,210 @@
...
+	if isAccount {
+		groups.Accounts[address] = append(groups.Accounts[address], payload)
+	} else {
+		groups.NonAccountPayloads = append(groups.NonAccountPayloads, payload)
+	}
+
+	return groups, nil
+}
+
+// AccountMigrator takes all the payloads that belong to the given account
```
pattyshack, line 57: create a proper interface please.  it's easier to compose / test

cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,210 @@
...
+	// add the non accounts which don't need to be migrated
+	migrated = append(migrated, groups.NonAccountPayloads...)
+	paths = append(paths, nonAccountPaths...)
+
+	log.Info().Msgf("finished migrating all account based payloads, total migrated payloads: %v", len(migrated))
+
+	return migrated, paths, nil
+}
+
+// MigrateGroupSequentially migrate the payloads in the given payloadsByAccount map which
```
pattyshack, line 108: dead code ...

cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,210 @@
...
+// MigrateGroupConcurrently migrate the payloads in the given payloadsByAccount map which
+// using the migrator
+// It's similar to MigrateGroupSequentially, except it will migrate different groups concurrently
+func MigrateGroupConcurrently(
+	migrator AccountMigrator,
+	payloadsByAccount map[string][]ledger.Payload,
+	pathFinderVersion uint8,
+) (
+	[]ledger.Payload, []ledger.Path, error) {
+	nWorker := 10
```
pattyshack, line 160: make this configurable

cmd/util/ledger/migrations/account_based_migration.go
zhangchiqing: Will be using this code to benchmark against the concurrent version when real migration is ready

module/util/log_test.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"testing"
+
+	"github.com/rs/zerolog/log"
+)
+
+func TestLogProgress(t *testing.T) {
+	logger := LogProgress("test", 41, &log.Logger)
+	for i := 0; i < 41; i++ {
+		logger(i)
```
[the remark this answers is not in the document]
zhangchiqing: Now sure how to mock the logger, but it's much easier to print and check the output

cmd/util/ledger/migrations/account_based_migration.go
janezpodhostnik: there are registers with "" as an address, but those are not for a special address they are for special registers that don't belong to any address. I think this check is ok, but might be nicer packaged in a function with a clear name.

cmd/util/ledger/migrations/account_based_migration.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return migrated, migratedPaths, nil
+}
+
+type jobMigrateAccountGroup struct {
+	Account  string
+	Payloads []ledger.Payload
+}
+
+type resultMigrating struct {
```
janezpodhostnik, line 148:
```suggestion
type migrationResult struct {
```

cmd/util/ledger/migrations/account_migration.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return uint64(registerSize(id, payload)), nil
+}
+
+func isAccountKey(key ledger.Key) bool {
+	return string(key.KeyParts[1].Value) == state.AccountStatusKey
+}
+
+type AccountUsageMigrator struct{}
```
janezpodhostnik, line 29: does this replace StorageUsedUpdateMigration?
should that one be deleted?

ledger/ledger.go
```diff
@@ -387,7 +387,8 @@ func (v *Value) UnmarshalJSON(b []byte) error {
 }
 
 // Migration defines how to convert the given slice of input payloads into an slice of output payloads
-type Migration func(payloads []Payload) ([]Payload, error)
+// and returns the paths of the payload
+type Migration func(payloads []Payload, pathFinderVersion uint8) ([]Payload, []Path, error)
```
janezpodhostnik, line 391: Its not clear to me why the migration should return paths. I feel like if anything it should accept paths...

cmd/util/ledger/migrations/account_based_migration.go
pattyshack: in that case, at least rm the commented out line of code at line 87.

if you refactor this into an interface, say GroupMigrator, you can switch out MigrateGroupConcurrently for MigrateGroupSequentially

janezpodhostnik (approved):

ledger/complete/ledger.go
```diff
@@ -378,6 +378,7 @@ func (l *Ledger) ExportCheckpointAt(
 		payloads = t.AllPayloads()
 		payloadSize := len(payloads)
 
+		var err error
```
pattyshack, line 381: not needed?

pattyshack (approved):

zhangchiqing:
bors merge

PR #3623, opened 2022-11-25
[Consensus] Implementation of missing consensus metrics

durkmurder:
https://github.com/dapperlabs/flow-go/issues/6319

### Context 

This PR fixes old metrics that were broken due to changes to the consensus. Moreover it enriches old metrics with new metrics by reporting more data. 

Main changes:
- Added ability to report dropped messages in `EngineMetrics`.
- Vote/Timeout aggregators report extra metrics including dropped/handled messages, processing duration, length of queue.
- `EventLoop` reports busy/wait/processing duration metrics on all events
- Updated compliance engine to report dropped messages, fixed old metrics
- `MessageHub` reports sent/received/dropped message metrics on all events


consensus/hotstuff/eventloop/event_loop.go
```diff
@@ -302,20 +299,22 @@ func (el *EventLoop) OnTcConstructedFromTimeouts(tc *flow.TimeoutCertificate) {
...
-
 	select {
 	case el.partialTcCreated <- event:
 	case <-el.ComponentManager.ShutdownSignal():
 		return
 	}
+
+	// the wait duration is measured as how long it takes from an event being
+	// received to event handler commencing the processing of the event
+	el.metrics.HotStuffWaitDuration(time.Since(received), metrics.HotstuffEventTypeOnPartialTc)
```
jordanschalm, line 317: [`e.partialTcCreated` has a channel size of 10](https://github.com/onflow/flow-go/blob/afdc0a830c6f61ff8ff40647d7a70b3c14982895/consensus/hotstuff/eventloop/event_loop.go#L51). So it's not measuring how long it takes for the Event Handler to start processing the event, it's measuring how long until the event is enqueued. 

This function is also blocking when the channel is full, which might block the `TimeoutProcessor` [here](https://github.com/onflow/flow-go/blob/94884ef4193fb6866060564a0c65d61e2ca2658f/consensus/hotstuff/timeoutcollector/timeout_processor.go#L142). 

Suppose we have 10 partial TC events queued, for views 10-20. Is it ever desirable to process the partial TC events for the views before 20? I don't think it is. So I'm wondering if we should use a pattern similar to how we process QCs and TCs here, where we store the latest partial TC by view in a `PartialTCTracker` and notify when it changes, rather than using a channel.

jordanschalm: The problem of measuring time-to-queueing rather than time-to-start-processing is similar for the proposals queue. The difference there could be large, since the queue has a max size of 1000 proposals. To measure how much time elapses between the EventLoop accepting the proposal and commencing processing, we could attach a timestamp to the proposal when enqueueing it in the channel, then record the difference between then and now when dequeueing the proposal in the event loop.

consensus/hotstuff/voteaggregator/vote_aggregator.go
```diff
@@ -157,7 +168,12 @@ func (va *VoteAggregator) processQueuedMessages(ctx context.Context) error {
 		msg, ok = va.queuedVotes.Pop()
 		if ok {
 			vote := msg.(*model.Vote)
+			startTime := time.Now()
 			err := va.processQueuedVote(vote)
+			// report duration of processing one vote
+			va.hotstuffMetrics.VoteProcessingDuration(time.Since(startTime))
+			va.engineMetrics.MessageHandled(metrics.EngineVoteAggregator, metrics.MessageBlockVote)
```
jordanschalm, line 175: This will use the same metric label for consensus and collection blocks.

engine/collection/message_hub/message_hub.go
```diff
@@ -382,6 +379,8 @@ func (h *MessageHub) OnOwnVote(blockID flow.Identifier, view uint64, sigData []b
 	}
 	if ok := h.ownOutboundVotes.Push(packed); ok {
 		h.ownOutboundMessageNotifier.Notify()
+	} else {
+		h.engineMetrics.MessageDropped(metrics.EngineCollectionMessageHub, metrics.MessageClusterBlockVote)
```
jordanschalm, line 383: This is changing convention slightly from previous usage of these metric labels, and other usages of `MessageDropped`. Usually it means, we received a message from another node, but dropped it before we processed it. Here it means, we are preparing to transmit a message we created, but dropped it before we could transmit it. I think it would be beneficial to use a label name that made this clear, and differentiated this case. 
```suggestion
		h.engineMetrics.MessageDropped(metrics.EngineCollectionMessageHub, metrics.MessageOutboundClusterBlockVote)
```

engine/collection/message_hub/message_hub.go
```diff
@@ -391,6 +390,8 @@ func (h *MessageHub) OnOwnTimeout(timeout *model.TimeoutObject) {
 	h.forwardToOwnTimeoutAggregator(timeout) // forward timeout to my own `timeoutAggregator`
 	if ok := h.ownOutboundTimeouts.Push(timeout); ok {
 		h.ownOutboundMessageNotifier.Notify()
+	} else {
+		h.engineMetrics.MessageDropped(metrics.EngineCollectionMessageHub, metrics.MessageClusterTimeoutObject)
```
jordanschalm, line 394:
```suggestion
		h.engineMetrics.MessageDropped(metrics.EngineOutboundCollectionMessageHub, metrics.MessageClusterTimeoutObject)
```

engine/collection/message_hub/message_hub.go
```diff
@@ -415,6 +416,8 @@ func (h *MessageHub) OnOwnProposal(proposal *flow.Header, targetPublicationTime
 
 		if ok := h.ownOutboundProposals.Push(proposal); ok {
 			h.ownOutboundMessageNotifier.Notify()
+		} else {
+			h.engineMetrics.MessageDropped(metrics.EngineCollectionMessageHub, metrics.MessageClusterBlockProposal)
```
jordanschalm, line 420:
```suggestion
			h.engineMetrics.MessageDropped(metrics.EngineOutboundCollectionMessageHub, metrics.MessageClusterBlockProposal)
```

engine/collection/message_hub/message_hub.go
```diff
@@ -468,6 +472,7 @@ func (h *MessageHub) forwardToOwnVoteAggregator(vote *messages.ClusterBlockVote,
 // forwardToOwnTimeoutAggregator logs timeout and forwards it to own `timeoutAggregator`.
 // Per API convention, timeoutAggregator` is non-blocking, hence, this call returns quickly.
 func (h *MessageHub) forwardToOwnTimeoutAggregator(t *model.TimeoutObject) {
+	h.engineMetrics.MessageReceived(metrics.EngineCollectionMessageHub, metrics.MessageClusterTimeoutObject)
```
jordanschalm, line 475:
```suggestion
```
Same comment here, to move this into `TimeoutAggregator`

engine/collection/message_hub/message_hub.go
```diff
@@ -450,6 +453,7 @@ func (h *MessageHub) Process(channel channels.Channel, originID flow.Identifier,
 // forwardToOwnVoteAggregator converts vote to generic `model.Vote`, logs vote and forwards it to own `voteAggregator`.
 // Per API convention, timeoutAggregator` is non-blocking, hence, this call returns quickly.
 func (h *MessageHub) forwardToOwnVoteAggregator(vote *messages.ClusterBlockVote, originID flow.Identifier) {
+	h.engineMetrics.MessageReceived(metrics.EngineCollectionMessageHub, metrics.MessageClusterBlockVote)
```
jordanschalm, line 456:
```suggestion
```
I have a feeling that spanning the received and processed/dropped metrics over two engines might making understanding them more difficult. In all existing cases I'm aware of, these metrics are used within one engine/component. Using the same engine label also lets you do queries that aggregate all the message types for a particular engine, like `processed - received by (engine)`. 

Suggestion: move the `MessageReceived` line into `VoteAggregator` and `TimeoutAggregator`.

engine/consensus/message_hub/message_hub.go
```diff
@@ -484,6 +487,7 @@ func (h *MessageHub) Process(channel channels.Channel, originID flow.Identifier,
 // forwardToOwnVoteAggregator converts vote to generic `model.Vote`, logs vote and forwards it to own `voteAggregator`.
 // Per API convention, timeoutAggregator` is non-blocking, hence, this call returns quickly.
 func (h *MessageHub) forwardToOwnVoteAggregator(vote *messages.BlockVote, originID flow.Identifier) {
+	h.engineMetrics.MessageReceived(metrics.EngineConsensusMessageHub, metrics.MessageBlockVote)
```
jordanschalm, line 490: same comment here for moving this into VA

engine/consensus/message_hub/message_hub.go
```diff
@@ -502,6 +506,7 @@ func (h *MessageHub) forwardToOwnVoteAggregator(vote *messages.BlockVote, origin
 // forwardToOwnTimeoutAggregator logs timeout and forwards it to own `timeoutAggregator`.
 // Per API convention, timeoutAggregator` is non-blocking, hence, this call returns quickly.
 func (h *MessageHub) forwardToOwnTimeoutAggregator(t *model.TimeoutObject) {
+	h.engineMetrics.MessageReceived(metrics.EngineConsensusMessageHub, metrics.MessageTimeoutObject)
```
jordanschalm, line 509: same comment here for moving this into TA

module/metrics.go
```diff
@@ -130,23 +131,34 @@ type HotstuffMetrics interface {
 	// HotStuffWaitDuration reports Metrics C6 HotStuff Idle Duration
```
jordanschalm, line 131:
```suggestion
	// HotStuffWaitDuration reports Metrics C6 HotStuff Idle Duration - the time between receiving and 
	// enqueueing a message to beginning to process that message.
```
Updating doc based on what I think this metric should measure. As mentioned in previous comments, we currently don't measure this correctly.

module/metrics.go
```diff
@@ -82,6 +82,7 @@ type EngineMetrics interface {
 	MessageSent(engine string, message string)
 	MessageReceived(engine string, message string)
 	MessageHandled(engine string, messages string)
+	MessageDropped(engine string, messages string)
```
jordanschalm, line 85:
```suggestion
	// MessageSent reports that the engine transmitted the message over the network.
	// Unicasts, broadcasts, and multicasts are all reported once. 
	MessageSent(engine string, message string)
	// MessageReceived reports that the engine received the message over the network.
	MessageReceived(engine string, message string)
	// MessageHandled reports that the engine has finished processing the message. 
	// Both invalid and valid messages should be reported. 
	// A message must be reported as either handled or dropped, not both.
	MessageHandled(engine string, messages string)
	// MessageDropped reports that the engine has dropped the message without processing it.
	// A message must be reported as either handled or dropped, not both.
	MessageDropped(engine string, messages string)
```

jordanschalm (commented):
Great work

engine/collection/message_hub/message_hub.go
durkmurder: I was thinking about it a bit differently. Cluster's `MessageHub` drops block vote so the respective metric message will be:
`metrics.EngineCollectionMessageHub` + `metrics.MessageClusterBlockVote` since it's the only vote that message hub is dropping it means it has to be an outbound block vote. I mean we can use the metrics more compactly this way instead of spawning a label for each use case which is harder to support in future not even talking about on-boarding someone into to the topic.

engine/collection/message_hub/message_hub.go
durkmurder: I see your argument here, when I was designing this I was thinking about tracking by _object_ not as much as by _engine_. This is the reason why messages are reported in different places, it gives a strong understanding where the message was received, handled and dropped and by which engine. Of course we can argue what approach is more suitable but I would like to have a generalized common pattern of reporting metrics.

engine/collection/message_hub/message_hub.go
jordanschalm: > since it's the only vote that message hub is dropping it means it has to be an outbound block vote

It's true that it has to be an outbound vote, but it isn't at all obvious without familiarity with the implementation details. At a high level, `MessageHub` both accepts and transmits votes - it would be completely reasonable for someone with high-level familiarity with the compliance engine design, and experience with message received/processed/dropped metrics reporting inbound messages, to assume this is referring to an inbound vote. Or at least for it to be unclear without further investigation. 

Another perspective is we have both message received and sent metrics so we can differentiate those cases - we could do the same for dropped messages: `InboundMessageDropped`, `OutboundMessageDropped`.

AlexHentschel: Agree with Jordan here, that the differentiation between inbound and outbound messages is too subtle. 

I can see two pragmatic ways how we can add sufficient differentiation:
1. Jordan's proposal
   > we have both message received and sent metrics so we can differentiate those cases - we could do the same for dropped messages: `InboundMessageDropped`, `OutboundMessageDropped`.

    i.e. we would differentiate by having _different metrics_
2. Alternatively, we could keep the same metric, but differentiate on the message type. Specifically, we could use different `message` strings here: https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/module/metrics.go#L91-L93

    For collector nodes, we would split these message types https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/module/metrics/labels.go#L120-L122
      and introduce one dedicated type for inbound and another for outbound, e.g

    ```golang
    MessageInboundClusterBlockProposal = "inbound_cluster_proposal"
    MessageOutboundClusterBlockProposal = "outbound_cluster_proposal"
    ```

module/metrics/hotstuff.go
```diff
@@ -166,24 +196,36 @@ func (hc *HotstuffCollector) HotStuffWaitDuration(duration time.Duration, event
...
 }
 
-// HotstuffCollector reports Metrics C8: Current View
+// CountSkipped counts the number of skips we did.
+func (hc *HotstuffCollector) CountSkipped() {
+	hc.skips.Inc()
+}
+
+// CountTimeout counts the number of timeouts we had.
+func (hc *HotstuffCollector) CountTimeout() {
```
AlexHentschel, line 205: I find we could improve the clarity of the documentation. From its documentation alone, is unclear, whether this is the number of times the consensus committee left a view with a timeout. It could also mean the number of times this node sent a timeout message. 

I would suggest to update
1. the godoc here 
    ```suggestion
    // CountTimeout tracks the number of views that this replica left due to observing a TC.
    func (hc *HotstuffCollector) CountTimeout() {
    ```
2. Update the metric documentation https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/module/metrics/hotstuff.go#L116 to 
     ```golang
    Help: "The number of views that this replica left due to observing a TC",
    ```

module/metrics/hotstuff.go
```diff
@@ -166,24 +196,36 @@ func (hc *HotstuffCollector) HotStuffWaitDuration(duration time.Duration, event
...
-// NewestKnownQC reports Metrics C9: View of Newest Known QC
+// SetQCView reports Metrics C9: View of Newest Known QC
 func (hc *HotstuffCollector) SetQCView(view uint64) {
 	hc.qcView.Set(float64(view))
 }
 
-// CountSkipped counts the number of skips we did.
-func (hc *HotstuffCollector) CountSkipped() {
-	hc.skips.Inc()
+func (hc *HotstuffCollector) SetTCView(view uint64) {
```
AlexHentschel, line 219:
```suggestion
// SetTCView reports the view of the newest known TC
func (hc *HotstuffCollector) SetTCView(view uint64) {
```

module/metrics/hotstuff.go
```diff
@@ -166,24 +196,36 @@ func (hc *HotstuffCollector) HotStuffWaitDuration(duration time.Duration, event
...
+	hc.tcView.Set(float64(view))
 }
 
-// CountTimeout counts the number of timeouts we had.
-func (hc *HotstuffCollector) CountTimeout() {
-	hc.timeouts.Inc()
+func (hc *HotstuffCollector) VoteProcessingDuration(duration time.Duration) {
```
AlexHentschel, line 223:
```suggestion
// VoteProcessingDuration reports the processing time for a single vote 
func (hc *HotstuffCollector) VoteProcessingDuration(duration time.Duration) {
```

module/metrics/hotstuff.go
```diff
@@ -166,24 +196,36 @@ func (hc *HotstuffCollector) HotStuffWaitDuration(duration time.Duration, event
...
+	hc.voteProcessingDuration.Observe(duration.Seconds())
+}
+
+func (hc *HotstuffCollector) TimeoutObjectProcessingDuration(duration time.Duration) {
```
AlexHentschel, line 227:
```suggestion
// TimeoutObjectProcessingDuration reports the processing time for a TimeoutObject
func (hc *HotstuffCollector) TimeoutObjectProcessingDuration(duration time.Duration) {
```

AlexHentschel (approved):
thanks for the solid, clean work, and especially for adding goDoc to the [`EngineMetrics` interface](https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/module/metrics.go#L81-L94) 🖖 

My comments are largely stylistic suggestions. Only detail where I would like to request revisions is for the `CountSkipped()` metric (see [my comment here](url)) and `CountTimeout()` ([comment]())

engine/collection/compliance/engine.go
```diff
@@ -120,6 +120,7 @@ func (e *Engine) processQueuedBlocks(doneSignal <-chan struct{}) error {
 		if ok {
 			inBlock := msg.(flow.Slashable[messages.ClusterBlockProposal])
 			err := e.core.OnBlockProposal(inBlock.OriginID, inBlock.Message)
+			e.core.engineMetrics.MessageHandled(metrics.EngineClusterCompliance, metrics.MessageClusterBlockProposal)
```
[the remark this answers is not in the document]
durkmurder: We have a metric for it https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/engine/consensus/compliance/core.go#L319, I will change it to Histogram though and use both for consensus and collection compliance engines.

module/metrics/hotstuff.go
```diff
@@ -166,24 +196,36 @@ func (hc *HotstuffCollector) HotStuffWaitDuration(duration time.Duration, event
 	hc.waitDuration.WithLabelValues(event).Observe(duration.Seconds()) // unit: seconds; with float64 precision
 }
 
-// HotstuffCollector reports Metrics C8: Current View
+// CountSkipped counts the number of skips we did.
+func (hc *HotstuffCollector) CountSkipped() {
```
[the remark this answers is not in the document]
durkmurder: yeah, we could do it like this, `CountSkipped` here tracks how many times we have used QC to change view instead of TC. i think it is still a useful metric to have. Maybe let's just rename `CountSkipped` to more obvious name?

engine/collection/message_hub/message_hub.go
durkmurder: I think I will stick to the approach 1) I would like to reuse as much as possible existing labels and avoid introducing more of them to reduce maintenance complexity. Additionally I would like to remove cluster labels(`MessageClusterConsensusObject`) and use instead `MessageConsensusObject` because the differentiation between consensus and collector engines comes from _engine_ label.

durkmurder: Separate metrics for outbound and inbound messages: https://github.com/onflow/flow-go/pull/3623/commits/a2902ae90b8a1e39058f3708b8052b8e4af2dfab
Cleanup of cluster specific messages: https://github.com/onflow/flow-go/blob/628b20cd267f34864960ab5834fdaeabddab7900/engine/consensus/compliance/core.go#L319

engine/collection/compliance/core.go
```diff
@@ -105,6 +109,9 @@ func NewCore(
 // OnBlockProposal handles incoming block proposals.
 // No errors are expected during normal operation.
 func (c *Core) OnBlockProposal(originID flow.Identifier, proposal *messages.ClusterBlockProposal) error {
+	startTime := time.Now()
+	defer c.hotstuffMetrics.BlockProcessingDuration(time.Since(startTime))
```
jordanschalm, line 113: ⚠️ Arguments for deferred functions are evaluated right away, so this will always report ~0s.

module/metrics/execution_data_sync.go
```diff
@@ -93,7 +93,7 @@ func NewExecutionDataRequesterV2Collector() *ExecutionDataRequesterV2Collector {
 			Name:      "responses_dropped",
 			Namespace: namespaceExecutionDataSync,
 			Subsystem: subsystemExeDataRequester,
-			Help:      "the number of dropped responses",
+			Help:      "the number of inboundDropped responses",
```
jordanschalm, line 96:
```suggestion
			Help:      "the number of dropped responses",
```

module/metrics/herocache.go
```diff
@@ -121,7 +121,7 @@ func NewHeroCacheCollector(nameSpace string, cacheName string, registrar prometh
 		Namespace: nameSpace,
 		Subsystem: subsystemHeroCache,
 		Name:      cacheName + "_" + "write_drop_count_total",
-		Help:      "total number of put queries dropped due to full capacity",
+		Help:      "total number of put queries inboundDropped due to full capacity",
```
jordanschalm, line 124:
```suggestion
		Help:      "total number of put queries dropped due to full capacity",
```

module/metrics/herocache.go
```diff
@@ -215,7 +215,7 @@ func (h *HeroCacheCollector) OnKeyPutSuccess(size uint32) {
 }
 
 // OnKeyPutDeduplicated is tracking the total number of unsuccessful writes caused by adding a duplicate key to the cache.
-// A duplicate key is dropped by the cache when it is written to the cache.
+// A duplicate key is inboundDropped by the cache when it is written to the cache.
```
jordanschalm, line 218:
```suggestion
// A duplicate key is dropped by the cache when it is written to the cache.
```

module/metrics/herocache.go
```diff
@@ -245,7 +245,7 @@ func (h *HeroCacheCollector) OnKeyPutAttempt(size uint32) {
 	h.size.Set(float64(size))
 }
 
-// OnKeyPutDrop is called whenever a new (key, entity) pair is dropped from the cache due to full cache.
+// OnKeyPutDrop is called whenever a new (key, entity) pair is inboundDropped from the cache due to full cache.
```
jordanschalm, line 248:
```suggestion
// OnKeyPutDrop is called whenever a new (key, entity) pair is dropped from the cache due to full cache.
```

module/metrics/network.go
```diff
@@ -82,7 +82,7 @@ func NewNetworkCollector(opts ...NetworkCollectorOpt) *NetworkCollector {
 			Namespace: namespaceNetwork,
 			Subsystem: subsystemGossip,
 			Name:      nc.prefix + "duplicate_messages_dropped",
-			Help:      "number of duplicate messages dropped",
+			Help:      "number of duplicate messages inboundDropped",
```
jordanschalm, line 85:
```suggestion
			Help:      "number of duplicate messages dropped",
```
There are several other instances of this bad auto-rename in this file.

durkmurder: oops, fixed those, thanks for spotting

engine/collection/compliance/core.go
durkmurder: https://github.com/onflow/flow-go/pull/3623/commits/7078c9cc5b14a6299357f809d50b3162c9091e3e

jordanschalm (approved):
🎸

consensus/hotstuff/eventloop/event_loop.go
durkmurder: https://github.com/onflow/flow-go/pull/3623/commits/a30e64688e838f0e4651a448e9d2d194085670a9

durkmurder:
bors merge

durkmurder:
bors merge

durkmurder:
bors-retry

durkmurder:
bors retry

PR #3625, opened 2022-11-25
Update Epoch Collection Node Join/Leave Integration Test

jordanschalm:
This PR updates and re-enables the `TestEpochStaticTransition` and `TestEpochJoinAndLeaveLN` integration tests.
* The static test was already passing and required no changes
* The LN test was failing due to an unhandled error when a node submits a vote in the epoch before it leaves the network
  * This is fixed by returning the appropriate error type in this case
  * Also updates the relevant unit tests to check for the error type
  * Created an issue for another scenario which arises in this test case, but doesn't impact the test result https://github.com/dapperlabs/flow-go/issues/6435


durkmurder (approved):
Nice, good thing having them back

kc1116 (approved):

PR #3627, opened 2022-11-28
Shorten bench and make it optional

janezpodhostnik:
Trade some benchstat accuracy for some time.

Also make benchstat optional since its only informative anyway.


Kay-Zee (approved):

janezpodhostnik:
bors merge

PR #3628, opened 2022-11-28
Only read and write checkpoint in v6

zhangchiqing:
From the next spork, the checkpoint V5 will be deprecated, and we will only support reading and generating checkpoint V6

### DIFF
--- a/cmd/util/cmd/execution-state-extract/cmd.go
+++ b/cmd/util/cmd/execution-state-extract/cmd.go
@@ -134,9 +134,10 @@ func run(*cobra.Command, []string) {
 		}
 	}
 
-	log.Info().Msgf("Extracting state from %s, exporting root checkpoint to %s, version: 6",
+	log.Info().Msgf("Extracting state from %s, exporting root checkpoint to %s, version: %v",
 		flagExecutionStateDir,
 		path.Join(flagOutputDir, bootstrap.FilenameWALRootCheckpoint),
+		6,
 	)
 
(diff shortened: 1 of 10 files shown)


ledger/complete/ledger_test.go
```diff
@@ -706,7 +706,7 @@ func TestLedgerFunctionality(t *testing.T) {
 }
 
 func Test_ExportCheckpointAt(t *testing.T) {
-	for _, version := range []int{5, 6} {
+	for _, version := range []int{6} {
```
fxamacker, line 709: nit: maybe remove `version` because checkpointer only produces v6 now.

cmd/util/cmd/execution-state-extract/cmd.go
```diff
@@ -137,10 +134,10 @@ func run(*cobra.Command, []string) {
 		}
 	}
 
-	log.Info().Msgf("Extracting state from %s, exporting root checkpoint to %s, version: %v",
+	log.Info().Msgf("Extracting state from %s, exporting root checkpoint to %s, version: 6",
```
fxamacker, line 137: nit: maybe use `wal.MaxVersion` instead of hard-coded version `6` assuming that we only extract checkpoint file in latest version.

fxamacker (approved):
LGTM!  I left a couple of nits.

ramtinms (approved):
Looks good to me

zhangchiqing:
bors merge

zhangchiqing:
bors merge

PR #3631, opened 2022-11-29
Re-enable AN test, increase test timeout, add logging, document outstanding issues

jordanschalm:
This PR fixes and re-enables the AN join/leave test, fixes several problems with the VN join/leave test, and documents outstanding issues which prevent tests from passing reliably. See https://github.com/dapperlabs/flow-go/issues/6443 for details.

* Modifies reference point for asserting sealed height to be a snapshot within epoch 2 (ensuring the joining VN is required to participate in sealing) rather than the snapshot used to bootstrap the joining VN
* Modifies test utilities to remove potential race conditions (use same source of truth for all queries)
* Increases test timeout to 20m from default 10m (takes around 13m currently to run the 3 enabled epoch tests)
* Adds periodic logger of status, to aid in debugging.
* Enables `require-approvals` flag for VN tests, increases DKG phase length accordingly.
* Removes warn-level logs from cluster sync engine, when clusters have only one member


gomisha (approved):

AlexHentschel (approved):
🎯

integration/Makefile
```diff
@@ -38,7 +38,8 @@ consensus-tests:
 
 .PHONY: epochs-tests
 epochs-tests:
-	go test $(if $(VERBOSE),-v,) $(RACE_FLAG) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) -tags relic ./tests/epochs/...
+	# Use a higher timeout of 20m for the suite of tests which span full epochs
+	go test $(if $(VERBOSE),-v,) $(RACE_FLAG) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) -tags relic -timeout 20m ./tests/epochs/...
```
jordanschalm, line 42: We have observed timeout failures in the past. The timeout is per test binary (package), not per test, and the epoch suite when passing is near enough to the default 10 minute timeout that the timeout failures appear to be real timeouts, rather than timeouts as a symptom of some other problem. For example, a recent passing test run (after increasing the test timeout) with 3 epoch tests enabled takes 13m for the `Run tests` step: https://github.com/onflow/flow-go/actions/runs/3596970055/jobs/6058931839

engine/collection/synchronization/engine.go
```diff
@@ -370,7 +371,7 @@ func (e *Engine) pollHeight() {
 		Height: head.Height,
 	}
 	err = e.con.Multicast(req, synccore.DefaultPollNodes, e.participants.NodeIDs()...)
-	if err != nil {
+	if err != nil && !errors.Is(err, network.EmptyTargetList) {
```
jordanschalm, line 374: Removes logging noise from test logs when there is a single-node collector cluster.

integration/tests/epochs/suite.go
```diff
@@ -52,6 +51,9 @@ type Suite struct {
 	DKGPhaseLen                uint64
 	EpochLen                   uint64
 	EpochCommitSafetyThreshold uint64
+	// Whether approvals are required for sealing (we only enable for VN tests because
+	// requiring approvals requires a longer DKG period to avoid flakiness)
+	RequiredSealApprovals uint // defaults to 0 (no approvals required)
```
jordanschalm, line 56: Adjust config so we can enable approvals for sealing only for the VN test case.

integration/tests/epochs/suite.go
```diff
@@ -673,13 +714,14 @@ func (s *Suite) submitSmokeTestTransaction(ctx context.Context) {
...
-// head of the rootSnapshot with the head of the snapshot we retrieved directly from the AN
-// 3. Check that we can execute a script on the AN
-func (s *Suite) assertNetworkHealthyAfterANChange(ctx context.Context, env templates.Environment, rootSnapshot *inmem.Snapshot, info *StakedNodeOperationInfo) {
+// assertNetworkHealthyAfterANChange performs a basic network health check after replacing an access node.
+//  1. Check that there is no problem connecting directly to the AN provided and retrieve a protocol snapshot
+//  2. Check that the chain moved at least 20 blocks from when the node was bootstrapped by comparing
+//     head of the rootSnapshot with the head of the snapshot we retrieved directly from the AN
+//  3. Check that we can execute a script on the AN
+//
+// TODO test sending and observing result of a transaction via the new AN (blocked by https://github.com/onflow/flow-go/issues/3642)
```
jordanschalm, line 723: Re-enable AN tests for script execution, but it would be best to test transaction API as well

integration/tests/epochs/epoch_join_and_leave_vn_test.go
```diff
@@ -10,14 +10,26 @@ import (
 )
 
 func TestEpochJoinAndLeaveVN(t *testing.T) {
-	unittest.SkipUnless(t, unittest.TEST_FLAKY, "epochs join/leave tests should be run on an machine with adequate resources")
+	// TODO this test is blocked by https://github.com/dapperlabs/flow-go/issues/6443
+	unittest.SkipUnless(t, unittest.TEST_FLAKY, "this test is flaky due to an unhandled case in service event processing following epoch transition https://github.com/dapperlabs/flow-go/issues/6443")
```
gomisha, line 14: So you're keeping this test in quarantine?

jordanschalm: Yes this test will be quarantined until https://github.com/dapperlabs/flow-go/issues/6443 is fixed.

integration/tests/epochs/suite.go
```diff
@@ -127,8 +129,10 @@ func (s *Suite) Ghost() *client.GhostClient {
 	return client
 }
 
-// TimedLogf logs the message using t.Log, but prefixes the current time.
+// TimedLogf logs the message using t.Log and the suite logger, but prefixes the current time.
+// This enables viewing logs inline with Docker logs as well as other test logs.
 func (s *Suite) TimedLogf(msg string, args ...interface{}) {
+	s.log.Info().Msgf(msg, args...)
```
durkmurder, line 135: do we need it?

integration/tests/epochs/suite.go
```diff
@@ -764,44 +804,50 @@ func (s *Suite) runTestEpochJoinAndLeave(role flow.Role, checkNetworkHealth node
...
+	rootSnapshot, err := s.client.GetLatestProtocolSnapshot(s.ctx)
 	require.NoError(s.T(), err)
-	testContainer.WriteRootSnapshot(snapshot)
-	testContainer.Container.Start(s.ctx)
 
-	header, err := snapshot.Head()
+	header, err := rootSnapshot.Head()
 	require.NoError(s.T(), err)
-	s.TimedLogf("retrieved header after entering EpochSetup phase: height=%d, view=%d", header.Height, header.View)
+	segment, err := rootSnapshot.SealingSegment()
```
durkmurder, line 816: do we use segment only for logging?

durkmurder (approved):
Nice, much needed test utils.

integration/tests/epochs/suite.go
jordanschalm: I found it useful because it puts the logs inline with Docker logs - makes it easier to see what's going on. You can always cross-reference by timestamp though, so not strictly necessary.

jordanschalm:
bors merge

PR #3634, opened 2022-11-30
[FVM] Check account storage before deducting fees

janezpodhostnik:
ref: https://github.com/onflow/flow-go/issues/2767

uses: https://github.com/onflow/flow-core-contracts/pull/323

Invert payer check  and fee deduction. This is possible by assuming that during the storage check the balance of the payer, is `maxFees` lower then it actually is.

The next step (PR) is metering the storage check and including it into the fees.


fvm/transactionStorageLimiter.go
```diff
@@ -60,3 +92,13 @@ func (_ TransactionStorageLimiter) CheckStorageLimits(
 
 	return nil
 }
+
+func (d TransactionStorageLimiter) getCommonAddressAndStorageUsed(env environment.Environment, address flow.Address) (common.Address, uint64, error) {
+	c := common.Address(address)
+	u, err := env.GetStorageUsed(c)
+	if err != nil {
+		err = fmt.Errorf("storage limit check failed: %w", err)
```
pattyshack, line 100: wrap the error around CheckStorageLimits instead of getCommonAddressAndStorageUsed to consistently report "storage limit check failed"

Also, maybe drop this method since it nothing after you've move the wrap

pattyshack (approved):

fvm/transactionStorageLimiter.go
```diff
@@ -14,29 +14,76 @@ import (
...
+	env environment.Environment,
+	addresses []flow.Address,
+	payer flow.Address,
+	maxTxFees uint64,
+) error {
 	commonAddresses := make([]common.Address, len(addresses))
 	usages := make([]uint64, len(commonAddresses))
+
+	commonAddressesContainPayer := false
+
```
SaveTheRbtz, line 53: nit: maybe first pre-process addresses so that we can remove the code duplication below?
(this would also avoid potential reallocation of `commonAddresses`/`usages` in cases where they are off by one)
```go
if maxTxFees > 0 {
	commonAddressesContainPayer := false
	for _, address := range addresses {
		if address == payer {
			commonAddressesContainPayer = true
			break
		}
	}

	if !commonAddressesContainPayer {
		addresses = append(addresses, payer)
	}
}
```

SaveTheRbtz (approved):

fvm/transactionStorageLimiter.go
```diff
@@ -14,33 +14,73 @@ import (
...
+	return nil
+}
+
+func (d TransactionStorageLimiter) checkStorageLimits(
+	env environment.Environment,
+	addresses []flow.Address,
+	payer flow.Address,
+	maxTxFees uint64,
+) error {
 	commonAddresses := make([]common.Address, len(addresses))
```
SaveTheRbtz, line 49: You probably want to initialize both of them to len+1 otherwise common addresses will have out of bounds access in case of `commonAddressesContainPayer`.

fvm/transactionStorageLimiter.go
```diff
@@ -14,29 +14,73 @@ import (
...
 type TransactionStorageLimiter struct{}
 
-func (_ TransactionStorageLimiter) CheckStorageLimits(
+// CheckStorageLimits checks each account that had its storage written to during a transaction, that its storage used
+// is less than its storage capacity.
+// Storage used is an FVM register and is easily accessible.
+// Storage capacity is calculated by the FlowStorageFees contract from the account's flow balance.
+//
+// The payers balance is considered to be maxTxFees lower that its actual balance, due to the fact that
+// the fee deduction step happens after the storage limit check.
```
zhangchiqing, line 23:
```suggestion
// CheckStorageLimits reads the storage usage of each account and the payer, 
// and reads their storage capacity from the given environment, and check if the storage
// usage of each account exceeds the storage capacity. 
// In additional, it also check if the given payer has enough balance to pay for the given maxTxFees
// It returns StorageCapacityExceededError if storage usage of a certain account exceeds its storage capacity
// It returns ErrCodeInsufficientPayerBalance if the payer has insufficient balance to cover maxTxFees.
// Storage used is an FVM register and is easily accessible without cadence.
// Storage capacity is calculated by the FlowStorageFees contract from the account's flow balance.
//
// The payers balance is considered to be maxTxFees lower that its actual balance, due to the fact that
// the fee deduction step happens after the storage limit check.
```

fvm/transactionStorageLimiter.go
```diff
@@ -14,29 +14,73 @@ import (
...
+
+		commonAddresses[i] = ca
 		usages[i] = u
 	}
 
-	result, invokeErr := env.AccountsStorageCapacity(commonAddresses)
+	result, invokeErr := env.AccountsStorageCapacity(
+		commonAddresses,
+		common.Address(payer),
+		maxTxFees,
```
zhangchiqing, line 82: Is `AccountsStorageCapacity` actually `CheckMaxTxFeeCanBeCoveredByPayerBalanceAndReadAccountsStorageCapacity`? Meaning it performs 1 check and 1 read:
1. check the payer has enough balance to cover maxTxFees
2. read each account's storage capacity

If so, why don't we separate into two calls instead?

SaveTheRbtz (approved):

fvm/transactionInvoker.go
```diff
@@ -298,27 +298,31 @@ func (executor *transactionExecutor) normalExecution() (
...
-	}
-
 	// Check if all account storage limits are ok
 	//
 	// disable the computation/memory limit checks on storage checks,
 	// so we don't error from computation/memory limits on this part.
-	// We cannot charge the user for this part, since fee deduction already happened.
+	// The storage limit check for the payer assumes that the payers
+	// balance is max_fees lower than the current payers balance,
+	// because the payer will have to pay for this tx.
```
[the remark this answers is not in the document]
janezpodhostnik: Let me try to describe this in a more concise way:

```suggestion
	// The storage limit check is performed for all accounts that were touched during the transaction.
	// The storage capacity of an account depends on its balance and should be higher than the accounts storage used.
	// The payer account is special cased in this check and its balance is considered max_fees lower than 
	// its actual balance, for the purpose of calculating storage capacity, because the payer will have to pay for this tx.
```

@zhangchiqing is this easier to understand whats going on?

fvm/transactionStorageLimiter.go
janezpodhostnik:
```
// In additional, it also check if the given payer has enough balance to pay for the given maxTxFees
...
// It returns ErrCodeInsufficientPayerBalance if the payer has insufficient balance to cover maxTxFees.
```
I different function does this part: `CheckPayerBalanceAndReturnMaxFees`

janezpodhostnik:
bors merge

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors retry

janezpodhostnik:
bors merge

PR #3636, opened 2022-11-30
[FVM] Meter storage check for fees

janezpodhostnik:
ref: @janezpodhostnik TODO - create issue

:warning: Do not merge yet.

Due to the changes in the storage check, we can now actually meter the storage check and include it in the fee calculation.

It should be included in the metering because it has a variable cost depending on the amount of accounts touched during the transaction. Which is something the user does have control over.

Changing this means transactions will cost slightly more computation (and consequentially FLOW) across the board. Transactions that touch a lot of accounts (rewards payout) will get a bigger bump in cost.

This might break any transaction that is currently very close to the limit and this additional cost will push it over the gas limit. So this has to be communicated clearly. 

**TODO:** 
- do not merge until we know estimates of how much more transactions will cost
- do not merge until we are ready to communicate the braking change out


pattyshack (approved):

janezpodhostnik:
This will need to be applied in the future, but it has conflicts, and I'm not up to constantly updating it. So I'll close it for now track this in this issue: https://github.com/onflow/flow-go/issues/3834.

PR #3629, opened 2022-11-29
Stop special casing system collection execution

pattyshack:
Asides from error string formating, there's no difference in how regular and system collections should be processed.

gh: #3547

PR #3637, opened 2022-11-30
Refactor computer result collection

pattyshack:
Refactor all execution result collection logic into ResultCollector.  In particular, BlockCommitter and EventHasher are foled into the collector.

Note that BlockCommiter's Commit and EventHasher's Hash are now part of CommitCollection and called inside executeCollection instead of executeBlock. The order of operation got reordered a bit to improve pipeline efficiency.

Also note that the system chunk error logging got moved into executeTransaction as is logged on a per transaction level.  This is a no-op in reality since the system chunk only has a singlge transaction.

gh: #3547

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -3,7 +3,6 @@ package computer
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -11,12 +10,10 @@ import (
 	otelTrace "go.opentelemetry.io/otel/trace"
 
 	"github.com/onflow/flow-go/engine/execution"
-	"github.com/onflow/flow-go/engine/execution/state/delta"
 	"github.com/onflow/flow-go/fvm"
 	"github.com/onflow/flow-go/fvm/blueprints"
 	"github.com/onflow/flow-go/fvm/programs"
 	"github.com/onflow/flow-go/fvm/state"
-	"github.com/onflow/flow-go/ledger"
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/module"
 	"github.com/onflow/flow-go/module/executiondatasync/execution_data"
@@ -32,6 +29,8 @@ const (
 )
 
 type collectionItem struct {
+	blockId string
+
 	*entity.CompleteCollection
 
 	ctx fvm.Context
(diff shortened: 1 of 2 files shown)


engine/execution/computation/computer/computer.go
```diff
@@ -163,53 +156,30 @@ func (e *blockComputer) executeBlock(
...
-		res:       res,
-	}
-	defer eh.Close()
+	collector := newResultCollector(
+		e.tracer,
+		blockSpan,
+		e.metrics,
+		e.committer,
+		block,
+		len(collections)+1)
```
janezpodhostnik, line 168: Maybe keep this comment `// + 1 system chunk` just to make it more obvious.

janezpodhostnik (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -220,23 +190,21 @@ func (e *blockComputer) executeBlock(
...
+	collector.Close()
 
 	err = e.mergeView(stateView, colView, blockSpan, trace.EXEMergeCollectionView)
 	if err != nil {
 		return nil, fmt.Errorf("cannot merge view: %w", err)
 	}
 
-	wg.Wait()
+	collector.Wait()
+	res := collector.result
```
SaveTheRbtz, line 207: does it make sense for Wait to return the result?

engine/execution/computation/computer/computer.go
```diff
@@ -163,53 +156,30 @@ func (e *blockComputer) executeBlock(
...
+		e.committer,
+		block,
+		len(collections)+1)
+	defer collector.Close()
 
-	go func() {
-		bc.Run()
-		wg.Done()
-	}()
+	collector.Start()
```
SaveTheRbtz, line 171: are there any downsides of auto-starting the collector?

SaveTheRbtz (approved):

engine/execution/computation/computer/computer.go
pattyshack: This comment stops making sense after #3629

engine/execution/computation/computer/computer.go
pattyshack: no, unless we test the collector's methods individually.  i'll change this to auto start for now

engine/execution/computation/computer/computer.go
pattyshack: that's reasonable.  I was thinking of combining Close/Wait/return result into a Finalize method anyways.

engine/execution/computation/computer/computer.go
```diff
@@ -354,71 +320,57 @@ func (e *blockComputer) executeCollection(
...
+			Str("block_id", collection.blockId).
 			Hex("collection_id", logging.Entity(collection.Guarantee)).
 			Msg("executing collection")
 	}
 
 	defer colSpan.End()
 
 	for _, txBody := range collection.Transactions {
-		err := e.executeTransaction(txBody, colSpan, collectionView, collection.ctx, collectionIndex, txIndex, res, false)
+		err := e.executeTransaction(collection.blockId, txBody, colSpan, collectionView, collection.ctx, collectionIndex, txIndex, collector, collection.isSystemCollection)
```
pattyshack, line 331: note that this was a bug (false should have been changed to collection.isSystemCollection) introduced in #3607.  This only impacted logging.

pattyshack:
bors merge

PR #3639, opened 2022-11-30
increase GRPC limit to 20MB

koko1123:


vishalchangrani (approved):

diaswrd (approved):

m4ksio:
Should we maybe make this a flag? Bumping default is fine, but we could have the other option

vishalchangrani:
> Should we maybe make this a flag? Bumping default is fine, but we could have the other option

Can you issue this up for later @diaswrd / @koko1123 ? For now I am going to merge this in so that we can start rolling it out.

PR #3641, opened 2022-12-01
[Benchmark] Support multiple keys

SaveTheRbtz:
This PR drastically speeds up account creation by using multiple keys on the service account.

Full list of changes from the @peterargue's original PR:
https://github.com/onflow/flow-go/compare/770759375d00e898798d38881153cdf0bf31fb03..rbtz/multiKey


integration/benchmark/scripts/addKeyToAccountTransaction.cdc
```diff
@@ -7,7 +7,7 @@ transaction(keys: [[UInt8]]) {
       )
       signer.keys.add(
         publicKey: publicKey,
-        hashAlgorithm: HashAlgorithm.SHA3_256,
+        hashAlgorithm: HashAlgorithm.SHA2_256,
```
gomisha, line 10: why the regression to an older hashing algo?

integration/benchmark/account/keys.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Done unlocks a key after use and puts it back into the pool.
+func (k *accountKey) Done() {
+	k.markUnused()
+	k.ks.availableKeys <- k
+}
+
+// Success is called when a key was successfully used to sign a transaction as the proposer
+// It increments the sequence number.
+func (k *accountKey) Success() {
```
peterargue, line 79: success is a little misleading since it also needs to be called if the tx failed. Maybe rename to `Increment()`?

integration/benchmark/scripts/addKeyToAccountTransaction.cdc
peterargue: this is the algorithm used by the service account key. It's easier to just use the same key everywhere so we don't need to track state between runs

integration/benchmark/contLoadGenerator.go
```diff
@@ -190,23 +198,104 @@ func (lg *ContLoadGenerator) stopped() bool {
...
+				err := lg.createAccounts(num)
+				if errors.Is(err, account.ErrNoKeysAvailable) {
+					lg.log.Warn().Err(err).Msg("error creating accounts, retrying...")
+					time.Sleep(1 * time.Second)
+					continue
+				}
+				return err
+			}
+		})
+		time.Sleep(1 * time.Second)
```
peterargue, line 294: you could probably reduce this time if you wanted

integration/benchmark/account/account.go
```diff
@@ -0,0 +1,87 @@
...
+	signer crypto.InMemorySigner
+}
+
+func New(i int, address *flowsdk.Address, accountKeys []*flowsdk.AccountKey, signer crypto.InMemorySigner) *FlowAccount {
+	keys := make([]*accountKey, 0, len(accountKeys))
+	for _, key := range accountKeys {
+		keys = append(keys, &accountKey{
+			AccountKey: key,
+			Address:    address,
+			Signer:     signer,
```
peterargue, line 28: given signer is not concurrency safe, is it safe to reuse the same object across keys?

integration/benchmark/account/keys.go
SaveTheRbtz: Made is a bit more verbose to explicitly saw what's incremented

peterargue (approved):

integration/benchmark/account/account.go
SaveTheRbtz: you were very right here-- it was useless to lock the key, given that we were reusing the signer across keys.

gomisha (approved):

SaveTheRbtz:
bors merge

PR #3644, opened 2022-12-01
Invalidate derived data when meter param overrides change.

pattyshack:
gh: #3102


fvm/environment/derived_data_invalidator.go
```diff
@@ -22,10 +22,42 @@ type ContractUpdate struct {
...
+
+	serviceAccount := string(env.chain.ServiceAddress().Bytes())
+	storageDomain := common.PathDomainStorage.Identifier()
+	for _, registerId := range updatedRegisterIds {
+		if registerId.Owner != serviceAccount {
+			continue
+		}
+
+		if registerId.Key == storageDomain ||
+			state.IsStabIndex(registerId.Key) {
```
SaveTheRbtz, line 53: Can you please add a comment why we need to invalidate meter params in these two cases?

SaveTheRbtz (approved):

fvm/state/state.go
```diff
@@ -353,10 +358,14 @@ func IsFVMStateKey(owner string, key string) bool {
 		key == AccountStatusKey
 }
 
+func IsSlabIndex(key string) bool {
+	return len(key) == 9 && key[0] == '$'
```
janezpodhostnik, line 362: Can we take this function from cadence somewhere. It feels like the FVM shouldn't be responsible for knowing this.

janezpodhostnik (approved):

fvm/state/state.go
pattyshack: i did a grep on the cadence code base, it seems like the only place that does a similar check is
`./runtime/cmd/decode-state-values/main.go:	return len(key) == slabKeyLength && key[0] == '$'`

can you point me to a function if there is one?

janezpodhostnik: Hmm... I cant find anything either. I'll open a cadence issue for this.

pattyshack:
bors merge

fvm/state/state.go
```diff
@@ -353,10 +358,14 @@ func IsFVMStateKey(owner string, key string) bool {
 		key == AccountStatusKey
 }
 
+func IsSlabIndex(key string) bool {
```
zhangchiqing, line 361: can we add some comment to this function?

fvm/environment/derived_data_invalidator.go
```diff
@@ -22,10 +22,47 @@ type ContractUpdate struct {
...
+
+func meterParamOverridesUpdated(env *facadeEnvironment) bool {
+	updatedRegisterIds, _ := env.txnState.RegisterUpdates()
+
+	serviceAccount := string(env.chain.ServiceAddress().Bytes())
+	storageDomain := common.PathDomainStorage.Identifier()
+	for _, registerId := range updatedRegisterIds {
+		if registerId.Owner != serviceAccount {
+			continue
+		}
```
zhangchiqing, line 50: Can we add some comments to explain why we skip the non service account? Just in case the assumption changes in the future.
```suggestion
		if registerId.Owner != serviceAccount {
			continue
		}
```

fvm/environment/derived_data_invalidator.go
```diff
@@ -22,10 +22,47 @@ type ContractUpdate struct {
...
+
+		// NOTE: This condition is empirically generated by running the
+		// MeterParamOverridesComputer to capture touched registers.
+		// The parameter settings are store in the service account's slabs.
```
zhangchiqing, line 54: This only explains why we check `state.IsSlabIndex(registerId.Key)` to me. 

Could we also explain why we also check `registerId.Key == storageDomain`?

fvm/environment/derived_data_invalidator.go
```diff
@@ -39,8 +76,8 @@ type ProgramInvalidator struct {
 }
 
 func (invalidator ProgramInvalidator) ShouldInvalidateEntries() bool {
-	// TODO(patrick): invalidate on meter param updates
-	return len(invalidator.ContractUpdateKeys) > 0 ||
+	return invalidator.MeterParamOverridesUpdated ||
+		len(invalidator.ContractUpdateKeys) > 0 ||
```
zhangchiqing, line 80: can you explain the case where `invalidator.MeterParamOverridesUpdated` is false, but `len(invalidator.ContractUpdateKeys) > 0 || len(invalidator.FrozenAccounts) > 0` is true?

Also why do we remove `len(invalidator.ContractUpdateKeys) > 0 || len(invalidator.FrozenAccounts) > 0` for `MeterParamOverridesInvalidator` ?

fvm/environment/derived_data_invalidator.go
pattyshack: (imo, slabs/storageDomain key, are a leaky abstraction that should not be exposed to cadence in the first place.  we should have provided register ids ordering guarantees instead and let the ledger handle the storage layout internally, but i digress ...)

Most, but not all, values/fields in an account are stored in an ordered map known only to cadence.  cadence encodes the map into bytes and split the bytes into slab chunks, then stores the slab chunks into the ledger.  When we check for stab id, we're basically checking for any changes to the ordered map.

I don't really know what's stored in `storageDomain` but that's one of the input for computing the meter param overrides and hence any changes to it will also invalidate the derived value.

I've tweak the comments a bit

fvm/environment/derived_data_invalidator.go
pattyshack: will do.

fvm/environment/derived_data_invalidator.go
pattyshack: any contract update to a regular account for example will trigger this condition.

i've remove this condition from MeterParamOverridesInvalidator because it's wrong (i only added that as a temporary placeholder).  in general, meter param overrides don't change when contract changes.

zhangchiqing (approved):
Looks good. Thanks for adding the comments

PR #3649, opened 2022-12-02
[Access] Add admin command to get execution data

peterargue:
This PR makes the max response size for the Admin server configurable, and adds the `read-execution-data` command to ANs running execution sync

### DIFF
--- a/admin/README.md
+++ b/admin/README.md
@@ -51,6 +51,11 @@ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"
 curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-transactions", "data": { "start-height": 340, "end-height": 343 }}'
 ```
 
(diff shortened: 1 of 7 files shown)


admin/README.md
```diff
@@ -51,6 +51,11 @@ curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"
 curl localhost:9002/admin/run_command -H 'Content-Type: application/json' -d '{"commandName": "get-transactions", "data": { "start-height": 340, "end-height": 343 }}'
 ```
 
+### To get execution data for a block by execution_data_id (only available on staked access and execution nodes)
```
zhangchiqing, line 54: Is this also added to execution node? There is no code change in execution_builder.go

peterargue: yea, ENs also have this:
https://github.com/onflow/flow-go/blob/master/cmd/execution_builder.go#L155-L157

koko1123 (approved):

diaswrd (approved):

PR #3651, opened 2022-12-02
Renamed fvm/programs to fvm/derived

pattyshack:
Also renamed some files at the same time.

Note that I've left fvm/programs/programs.go as is.  This entire directory can be removed once emulator/integration test is updated.

gh: #3102


SaveTheRbtz (approved):

pattyshack:
bors merge

PR #3652, opened 2022-12-02
Cleanup ComputationResult

pattyshack:
1. consistently use collection for naming, instead of using collection/chunk interchangeably
2. simplify adding collection to the result
3. return all collection/block stats as a struct using a single api call
4. clean up collection/block stats metric logging

gh: #3547

### DIFF
--- a/engine/execution/computation/computer/computer.go
+++ b/engine/execution/computation/computer/computer.go
@@ -351,14 +351,7 @@ func (e *blockComputer) executeCollection(
 		collection,
 		collectionView)
 
-	// TODO(patrick): refactor
-	e.metrics.ExecutionCollectionExecuted(time.Since(startedAt),
-		stats.ComputationUsed, stats.MemoryUsed,
-		stats.EventCounts, stats.EventSize,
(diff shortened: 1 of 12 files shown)


SaveTheRbtz (approved):

janezpodhostnik (commented):
The code changes look good.

I'm a bit concerned about the renaming. To my knowledge chunk and collection are not the same. 

- a collection is the incoming array of transactions 
- a chunk is an outgoing array of transactions

A chunk is a smallest verifiable unit by the verification nodes. These are currently the same. A collection becomes a chunk. 

In the future chunks could be arbitrarily constructed from the transactions in a block.

Maybe @m4ksio or @ramtinms  can check this out as well.

pattyshack:
we can switch to chunk consistently as well.  i don't have a strong preference.  the only reason i picked collection is because we seem refer to it more often, even within verification.

pattyshack:
bors merge

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3653, opened 2022-12-02
Improve result collector to return error instead of panic

pattyshack:
gh: #3547

### DIFF
--- a/engine/execution/computation/computer/result_collector.go
+++ b/engine/execution/computation/computer/result_collector.go
@@ -1,8 +1,10 @@
 package computer
 
 import (
+	"fmt"
 	"sync"
 
+	"github.com/hashicorp/go-multierror"
 	otelTrace "go.opentelemetry.io/otel/trace"
 
 	"github.com/onflow/flow-go/engine/execution"
@@ -37,13 +39,16 @@ type resultCollector struct {
 	metrics module.ExecutionMetrics
 
 	closeOnce sync.Once
-	wg        sync.WaitGroup
 
-	committer ViewCommitter
-	state     flow.StateCommitment
-	viewChan  chan state.View
+	committer          ViewCommitter
+	state              flow.StateCommitment
+	committerInputChan chan state.View
+	committerDoneChan  chan struct{}
+	committerError     error
 
-	eventsListChan chan flow.EventsList
+	hasherInputChan chan flow.EventsList
+	hasherDoneChan  chan struct{}
+	hasherError     error
 
 	result *execution.ComputationResult
 }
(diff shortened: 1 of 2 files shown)


engine/execution/computation/computer/result_collector.go
```diff
@@ -171,13 +190,32 @@ func (collector *resultCollector) CommitCollection(
...
+	<-collector.hasherDoneChan
+
+	var err error
+	if collector.committerError != nil {
+		err = multierror.Append(err, collector.committerError)
+	}
+
+	if collector.hasherError != nil {
+		err = multierror.Append(err, collector.hasherError)
+	}
```
SaveTheRbtz, line 214: nit: you can likely simplify this as:
```suggestion
	multierr.AppendInto(&err, collector.committerError)
	multierr.AppendInto(&err, collector.hasherError)
```

SaveTheRbtz (approved):

engine/execution/computation/computer/result_collector.go
pattyshack: there's no such function? https://pkg.go.dev/github.com/hashicorp/go-multierror#pkg-index

janezpodhostnik (approved):

engine/execution/computation/computer/result_collector.go
```diff
@@ -171,13 +190,32 @@ func (collector *resultCollector) CommitCollection(
...
+func (collector *resultCollector) Finalize() (
+	*execution.ComputationResult,
+	error,
+) {
 	collector.Stop()
-	collector.wg.Wait()
-	return collector.result
+
+	<-collector.committerDoneChan
+	<-collector.hasherDoneChan
```
zhangchiqing, line 205: what about moving the error to the channel?
```
err := <- collector.committerDoneChan

if err != nil {
  ...
}
```

engine/execution/computation/computer/result_collector.go
```diff
@@ -146,8 +154,19 @@ func (collector *resultCollector) CommitCollection(
...
+		// Do nothing
+	case <-collector.committerDoneChan:
+		// Committer exited (probably due to an error)
+	}
+
+	select {
+	case collector.hasherInputChan <- collector.result.Events[collectionIndex]:
+		// Do nothing
+	case <-collector.hasherDoneChan:
+		// Hasher exited (probably due to an error)
```
zhangchiqing, line 168: In this case, if there is error, should we return error instead of the ExecutionStats?

engine/execution/computation/computer/result_collector.go
```diff
@@ -54,30 +59,31 @@ func newResultCollector(
...
+	go collector.runCollectionCommitter()
+	go collector.runCollectionHasher()
 
 	return collector
 }
 
-func (collector *resultCollector) runChunkCommitter() {
-	defer collector.wg.Done()
+func (collector *resultCollector) runCollectionCommitter() {
+	defer close(collector.committerDoneChan)
```
zhangchiqing, line 84: 👍

engine/execution/computation/computer/result_collector.go
pattyshack: it's not safe to return the error in the channel since that could led to a loss event (only the first receiver will see the error, the rest will see nil)

pattyshack:
bors merge

pattyshack:
bors retry

zhangchiqing (approved):

pattyshack:
bors merge

PR #3659, opened 2022-12-03
[Execution] Remove useless condition from builder

SaveTheRbtz:
This check is always true on the `else if` branch


peterargue (approved):

zhangchiqing (approved):

SaveTheRbtz:
bors merge

PR #3671, opened 2022-12-06
[Testing] Add verbose envvar and set in CI

peterargue:
Running tests locally with `--verbose` set is very noisy and makes it hard to quickly find the specific failing tests. Update the makefile to use a `VERBOSE` environment variable to control whether or not to unable verbose logging. Also, update the CI workflows and scripts to use the flag.

the following tests use VERBOSE `go test` logging (`go test -v`):
- unit tests
- integration tests
- module tests


yhassanzadeh13 (approved):

SaveTheRbtz:
@gomisha any reason for disabling race detector?

gomisha (approved):
This PR showed that that there are further issues with flaky tests that need to be resolved, but the verbosity changes you did are outside the scope of that and don't break any existing functionality.

peterargue:
bors merge

PR #3480, opened 2022-10-31
[Network] Stop sending unused message.Message fields

peterargue:
These fields are deprecated, and will be removed in a future PR. For now, don't include them when sending a new message.

PR #3673, opened 2022-12-06
[Network] Collect network metrics after processing all fields

peterargue:
https://github.com/onflow/flow-go/pull/3480 removed some fields that were previously included in the network message. After that change, some of the metrics collectors are missing values for label fields.

This PR refactors the network metrics collectors so they run after all of the required fields are populated.

This also does some cleanup of the `readSubscription` to simplify its signalling and error handling.

### DIFF
--- a/network/p2p/network.go
+++ b/network/p2p/network.go
@@ -19,6 +19,7 @@ import (
 	"github.com/onflow/flow-go/module"
 	"github.com/onflow/flow-go/module/component"
 	"github.com/onflow/flow-go/module/irrecoverable"
+	"github.com/onflow/flow-go/module/metrics"
 	"github.com/onflow/flow-go/network"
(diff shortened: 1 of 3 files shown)


network/p2p/network.go
```diff
@@ -327,6 +328,8 @@ func (n *Network) Identity(pid peer.ID) (*flow.Identity, bool) {
 }
 
 func (n *Network) Receive(nodeID flow.Identifier, msg *message.Message, decodedMsgPayload interface{}) error {
+	n.metrics.NetworkMessageReceived(msg.Size(), msg.ChannelID, msg.Type)
```
peterargue, line 331: This needs to be updated to indicate which messages were received over unicast.

peterargue:
this PR was replaced by https://github.com/onflow/flow-go/pull/3675

PR #3676, opened 2022-12-07
[Execution] Refactor uploader into separate module

peterargue:
Refactor the execution block uploader into a separate module, to remove the code from the computation manager.


engine/execution/computation/manager.go
```diff
@@ -302,22 +287,8 @@ func (e *Manager) ComputeBlock(
...
-				defer span.End()
-
-				return uploader.Upload(result)
-			})
-		}
-
-		err = group.Wait()
-
+	if e.uploader.Enabled() {
+		err := e.uploader.Upload(ctx, result)
```
zhangchiqing, line 291: can we move the uploader.Upload to ingestion engine right after it's calling [manager.ComputeBlock()](https://github.com/onflow/flow-go/blob/master/engine/execution/ingestion/engine.go#L618)?

That way, manager doesn't need to depend on uploader. Ingestion engine depends on the uploader anyway, so we can just let ingestion engine take care of the upload.

If we think about it, the result upload has nothing to do with computation. It simply upload the result to cloud.

If we move the uploader to ingestion engine, then the uploader.Manager module could also move under ingestion package, which is right now in the manager package.

zhangchiqing (commented):
Thanks for refactoring the uploader. It's really helpful reducing the complexity. 

I just have one comment.

engine/execution/ingestion/engine.go
```diff
@@ -617,28 +618,32 @@ func (e *Engine) executeBlock(ctx context.Context, executableBlock *entity.Execu
...
-			Msg("error while computing block")
+		lg.Err(err).Msg("error while computing block")
 		return
 	}
 
+	if e.uploader.Enabled() {
+		err := e.uploader.Upload(ctx, computationResult)
+		if err != nil {
+			lg.Err(err).Msg("error while uploading block")
+			// continue processing. uploads should not block execution
```
peterargue, line 629: this is a change in behavior. Previously, if the uploader returned an error, `ComputeBlock` also [returned an error](https://github.com/onflow/flow-go/pull/3676/files#diff-d3ee6a44b1853d89f2bdaa968ac7a540797e841f88cd4bc1bb1edb3f8008a791L322).

Realistically, all implementations are currently `AsyncUploader`s which always [return `nil`](https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/uploader/uploader.go#L98-L98), but it seems incorrect for execution to fail if there's an issue with uploaders. I'm happy to change it back to returning here if the consensus is to fail.

engine/execution/computation/manager.go
peterargue: updated in 7b82842 and bd39327

engine/execution/ingestion/engine.go
zhangchiqing: I agree we should not block execution if uploading returns an error. 

However, I wonder how do we recover from this failure?

If block 3 failed to upload, and block 4 succeeded to upload, when and how to retry upload for block 3?

How does it work before?

cc @m4ksio

peterargue: Right now, all uploaders are `AsyncUploaders` which are hardcoded to return `nil` as the error. Internally, they have some [retry logic](https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/uploader/uploader.go#L75-L81) and make multiple attempts before eventually giving up. The `RetryableUploaderWrapper` that's used for GCP uploads will create a [db entry](https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/uploader/retryable_uploader_wrapper.go#L122-L125) that the block should be uploaded, but that's only check on startup.

So effectively, this error is never handled today (it's hardcoded `nil` and even complete failures aren't retried)

m4ksio (approved):

zhangchiqing (approved):
Looks good. Thanks for the refactor!

peterargue:
bors merge

peterargue:
bors retry

PR #3677, opened 2022-12-07
suggested amendments for PR 3441

AlexHentschel:
suggested amendments for PR https://github.com/onflow/flow-go/pull/3441

### DIFF
--- a/consensus/hotstuff/model/errors.go
+++ b/consensus/hotstuff/model/errors.go
@@ -246,7 +246,6 @@ func NewInvalidAggregatedSignatureErrorf(msg string, args ...interface{}) error 
 	return InvalidAggregatedSignatureError{fmt.Errorf(msg, args...)}
 }
 
-func (e InvalidAggregatedSignatureError) Error() string { return e.error.Error() }
 func (e InvalidAggregatedSignatureError) Unwrap() error { return e.error }
 
 // IsInvalidAggregatedSignatureError returns whether err is an InvalidAggregatedSignatureError
--- a/consensus/hotstuff/model/errors.go
+++ b/consensus/hotstuff/model/errors.go
@@ -234,10 +234,6 @@ func IsInvalidSignatureIncludedError(err error) bool {
 }
 
 // InvalidAggregatedSignatureError indicates that the aggregated signature is invalid.
-// (because it is equal to an identity signature).
-// This can happen because:
-//   - one or many signatures added via TrustedAdd are invalid to their respective public keys.
-//   - OR the signatures are valid but the public keys were forged to sum up to an identity public key.
 type InvalidAggregatedSignatureError struct {
 	error
 }
--- a/consensus/hotstuff/model/errors.go
+++ b/consensus/hotstuff/model/errors.go
@@ -238,12 +238,33 @@ func IsInvalidSignatureIncludedError(err error) bool {
 // This can happen because:
 //   - one or many signatures added via TrustedAdd are invalid to their respective public keys.
 //   - OR the signatures are valid but the public keys were forged to sum up to an identity public key.
-var InvalidAggregatedSignatureError = errors.New("aggregated signature is invalid (identity signature)")
+type InvalidAggregatedSignatureError struct {
+	error
+}
+
+func NewInvalidAggregatedSignatureError(err error) error {
+	return InvalidAggregatedSignatureError{err}
+}
+
+func NewInvalidAggregatedSignatureErrorf(msg string, args ...interface{}) error {
+	return InvalidAggregatedSignatureError{fmt.Errorf(msg, args...)}
+}
+
+func (e InvalidAggregatedSignatureError) Error() string { return e.error.Error() }
+func (e InvalidAggregatedSignatureError) Unwrap() error { return e.error }
 
 // IsInvalidAggregatedSignatureError returns whether err is an InvalidAggregatedSignatureError
 func IsInvalidAggregatedSignatureError(err error) bool {
-	return errors.Is(err, InvalidAggregatedSignatureError)
+	var e InvalidAggregatedSignatureError
+	return errors.As(err, &e)
 }
+
+//var InvalidAggregatedSignatureError = errors.New("aggregated signature is invalid (identity signature)")
+//
+//// IsInvalidAggregatedSignatureError returns whether err is an InvalidAggregatedSignatureError
+//func IsInvalidAggregatedSignatureError(err error) bool {
+//	return errors.Is(err, InvalidAggregatedSignatureError)
+//}
 
 // InsufficientSignaturesError indicates that not enough signatures have been stored to complete the operation.
 type InsufficientSignaturesError struct {
--- a/consensus/hotstuff/model/errors.go
+++ b/consensus/hotstuff/model/errors.go
@@ -259,13 +259,6 @@ func IsInvalidAggregatedSignatureError(err error) bool {
 	return errors.As(err, &e)
 }
 
-//var InvalidAggregatedSignatureError = errors.New("aggregated signature is invalid (identity signature)")
-//
-//// IsInvalidAggregatedSignatureError returns whether err is an InvalidAggregatedSignatureError
-//func IsInvalidAggregatedSignatureError(err error) bool {
-//	return errors.Is(err, InvalidAggregatedSignatureError)
-//}
-
 // InsufficientSignaturesError indicates that not enough signatures have been stored to complete the operation.
 type InsufficientSignaturesError struct {
 	err error
--- a/consensus/hotstuff/signature/packer_test.go
+++ b/consensus/hotstuff/signature/packer_test.go
@@ -84,6 +84,34 @@ func TestPackUnpack(t *testing.T) {
 	require.Equal(t, expectedSignerIDs, signers.NodeIDs())
 }
 
+// TestUnpack_EmptySignerList verifies that `Unpack` gracefully handles the edge case
+// of an empty signer list, as such could be an input from a byzantine node.
+func TestPackUnpack_EmptySigners(t *testing.T) {
+	// encode SignatureData with empty SigType vector (this could be an input from a byzantine node)
+	byzantineInput := model.SignatureData{
+		SigType:                      []byte{},
+		AggregatedStakingSig:         unittest.SignatureFixture(),
+		AggregatedRandomBeaconSig:    unittest.SignatureFixture(),
+		ReconstructedRandomBeaconSig: unittest.SignatureFixture(),
+	}
+	encoder := new(model.SigDataPacker)
+	sig, err := encoder.Encode(&byzantineInput)
+	require.NoError(t, err)
+
+	// create packer with a non-empty committee (honest node trying to decode the sig data)
+	committee := unittest.IdentityListFixture(6, unittest.WithRole(flow.RoleConsensus))
+	packer := newPacker(committee)
+	unpacked, err := packer.Unpack(make([]*flow.Identity, 0), sig)
+	require.NoError(t, err)
+
+	// check that the unpack data match with the original data
+	require.Empty(t, unpacked.StakingSigners)
+	require.Empty(t, unpacked.RandomBeaconSigners)
+	require.Equal(t, byzantineInput.AggregatedStakingSig, unpacked.AggregatedStakingSig)
+	require.Equal(t, byzantineInput.AggregatedRandomBeaconSig, unpacked.AggregatedRandomBeaconSig)
+	require.Equal(t, byzantineInput.ReconstructedRandomBeaconSig, unpacked.ReconstructedRandomBeaconSig)
+}
+
 // if signed by 60 staking nodes, and 50 random beacon nodes among a 200 nodes committee,
 // it's able to pack and unpack
 func TestPackUnpackManyNodes(t *testing.T) {
--- a/consensus/hotstuff/signature/weighted_signature_aggregator_test.go
+++ b/consensus/hotstuff/signature/weighted_signature_aggregator_test.go
@@ -252,7 +252,7 @@ func TestWeightedSignatureAggregator(t *testing.T) {
 		_, err = aggregator.TrustedAdd(ids[1].NodeID, sigs[1]) // stand-alone verification
 		require.NoError(t, err)
 
-		// Aggregation should validate its own aggregation result and error with sentinel ErrIdentitySignature
+		// Aggregation should validate its own aggregation result and error with sentinel model.InvalidAggregatedSignatureError
 		signers, agg, err := aggregator.Aggregate()
 		assert.Error(t, err)
 		assert.True(t, model.IsInvalidAggregatedSignatureError(err))
--- a/consensus/hotstuff/signature/weighted_signature_aggregator_test.go
+++ b/consensus/hotstuff/signature/weighted_signature_aggregator_test.go
@@ -252,7 +252,7 @@ func TestWeightedSignatureAggregator(t *testing.T) {
 		_, err = aggregator.TrustedAdd(ids[1].NodeID, sigs[1]) // stand-alone verification
 		require.NoError(t, err)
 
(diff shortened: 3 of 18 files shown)


crypto/bls.go
```diff
@@ -243,18 +247,7 @@ const identityBLSSignatureHeader = byte(0xC0)
...
-	if s[0] != identityBLSSignatureHeader {
-		return false
-	}
-	for _, b := range s[1:] {
-		if b != 0 {
-			return false
-		}
-	}
-	return true
+	return bytes.Compare(s, identityBLSSignature) == 0
```
AlexHentschel, line 250: I would expect it to be faster to use `bytes.Compare` as this potentially uses low-level optimized code (utilizing AVX instructions if available. I think Alexey did some detailed investigations into this)

module/signature/aggregation.go
```diff
@@ -224,15 +221,15 @@ func (s *SignatureAggregatorSameMessage) Aggregate() ([]int, crypto.Signature, e
 // keys corresponding to the input signers.
 // Aggregating the keys of the signers internally is optimized to only look at the keys delta
 // compared to the latest execution of the function. The function is therefore not thread-safe.
+// Per convention, this function rejects the identity signature as invalid.
 // Possible returns:
 //   - (true, nil): aggregate signature is valid
-//   - (false, nil): aggregate signature is cryptographically invalid
+//   - (false, nil): aggregate signature is cryptographically invalid (including identity signature)
```
AlexHentschel, line 227: > including identity signature 

is this correct?

tarakby: Yes. Identity signature fails the cryptographic signature verification, and is not rejected by this module by convention.

module/signature/aggregation.go
```diff
@@ -209,7 +206,7 @@ func (s *SignatureAggregatorSameMessage) Aggregate() ([]int, crypto.Signature, e
 	if !ok {
 		// check for identity signature (invalid aggregated signature)
 		if crypto.IsBLSSignatureIdentity(aggregatedSignature) {
-			return nil, nil, InvalidAggregatedSignatureError
+			return nil, nil, fmt.Errorf("invalid aggregated signature: %w", ErrIdentitySignature)
```
tarakby, line 209: nit and can be ignored: I think the outer and internal error describe exactly the same idea, maybe we could return `nil, nil, ErrIdentitySignature` in this case?

module/signature/aggregation.go
```diff
@@ -158,22 +158,19 @@ func (s *SignatureAggregatorSameMessage) HasSignature(signer int) (bool, error)
...
 // Returns:
 //   - InsufficientSignaturesError if no signatures have been added yet
-//   - InvalidAggregatedSignatureError if the aggregated signature is invalid. It's not clear whether included
-//     signatures via TrustedAdd are invalid. This case can happen even when all added signatures
-//     are individually valid.
+//   - ErrIdentitySignature if aggregation produced the identity signature, which is invalid
+//     by convention. This error can arise in two scenarios:
+//     1. Some signatures added via TrustedAdd were forged specifically with the goal to yield the
+//     identity signature. Here, these signatures would be invalid w.r.t to their respective public keys.
+//     2. The signatures are valid but the public keys were forged to sum up to an identity public key.
```
tarakby, line 173: nice 👌🏼

module/signature/aggregation.go
```diff
@@ -158,22 +158,19 @@ func (s *SignatureAggregatorSameMessage) HasSignature(signer int) (bool, error)
...
+// "post-check" of the aggregated signature is required for safety, as `TrustedAdd` allows
+// adding invalid signatures.
 // The function is not thread-safe.
 // Returns:
 //   - InsufficientSignaturesError if no signatures have been added yet
-//   - InvalidAggregatedSignatureError if the aggregated signature is invalid. It's not clear whether included
-//     signatures via TrustedAdd are invalid. This case can happen even when all added signatures
-//     are individually valid.
+//   - ErrIdentitySignature if aggregation produced the identity signature, which is invalid
+//     by convention. This error can arise in two scenarios:
```
tarakby, line 170: I would remove "by convention" so that we don't imply that only this module made the convention, while the invalidity of the identity signature comes from the low level crypto layer. 

```suggestion
//     This error can arise in two scenarios:
```

consensus/hotstuff/model/errors.go
```diff
@@ -238,13 +238,34 @@ func IsInvalidSignatureIncludedError(err error) bool {
...
+
+func NewInvalidAggregatedSignatureError(err error) error {
+	return InvalidAggregatedSignatureError{err}
+}
+
+func NewInvalidAggregatedSignatureErrorf(msg string, args ...interface{}) error {
+	return InvalidAggregatedSignatureError{fmt.Errorf(msg, args...)}
+}
+
+func (e InvalidAggregatedSignatureError) Error() string { return e.error.Error() }
```
tarakby, line 253: We can remove this definition because it's implicitly defined when `error` is embedded in `InvalidAggregatedSignatureError`.

```suggestion
```

consensus/hotstuff/model/errors.go
```diff
@@ -238,13 +238,34 @@ func IsInvalidSignatureIncludedError(err error) bool {
...
+	return errors.As(err, &e)
 }
 
+//var InvalidAggregatedSignatureError = errors.New("aggregated signature is invalid (identity signature)")
+//
+//// IsInvalidAggregatedSignatureError returns whether err is an InvalidAggregatedSignatureError
+//func IsInvalidAggregatedSignatureError(err error) bool {
+//	return errors.Is(err, InvalidAggregatedSignatureError)
+//}
+
```
tarakby, line 268:
```suggestion
```

consensus/hotstuff/model/errors.go
```diff
@@ -238,13 +238,34 @@ func IsInvalidSignatureIncludedError(err error) bool {
 // This can happen because:
 //   - one or many signatures added via TrustedAdd are invalid to their respective public keys.
 //   - OR the signatures are valid but the public keys were forged to sum up to an identity public key.
-var InvalidAggregatedSignatureError = errors.New("aggregated signature is invalid (identity signature)")
+type InvalidAggregatedSignatureError struct {
```
tarakby, line 241: Not sure if you initially meant to rename this error also to `IdentitySignatureError` like you did in `module/signature` ?

consensus/hotstuff/signature/weighted_signature_aggregator_test.go
```diff
@@ -252,7 +252,7 @@ func TestWeightedSignatureAggregator(t *testing.T) {
 		_, err = aggregator.TrustedAdd(ids[1].NodeID, sigs[1]) // stand-alone verification
 		require.NoError(t, err)
 
-		// Aggregation should validate its own aggregation result and error with sentinel InvalidAggregatedSignatureError
+		// Aggregation should validate its own aggregation result and error with sentinel ErrIdentitySignature
```
tarakby, line 255: I think this error is still called `model.InvalidAggregatedSignatureError`, it hasn't been renamed to `ErrIdentitySignature` (yet).

consensus/hotstuff/verification/combined_verifier_v2.go
```diff
@@ -118,9 +117,7 @@ func (c *CombinedVerifier) VerifyVote(signer *flow.Identity, sigData []byte, blo
 //   - model.ErrInvalidSignature if a signature is invalid
 //   - error if running into any unexpected exception (i.e. fatal error)
 func (c *CombinedVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
-	if len(signers) == 0 {
-		return model.NewInsufficientSignaturesErrorf("empty list of signers")
-	}
+	msg := MakeVoteMessage(block.View, block.BlockID)
```
tarakby, line 120: curious why `msg` was moved to be constructed early and not right before the verification?

consensus/hotstuff/verification/combined_verifier_v3.go
```diff
@@ -122,9 +122,7 @@ func (c *CombinedVerifierV3) VerifyVote(signer *flow.Identity, sigData []byte, b
 // This implementation already support the cases, where the DKG committee is a
 // _strict subset_ of the full consensus committee.
 func (c *CombinedVerifierV3) VerifyQC(signers flow.IdentityList, sigData []byte, block *model.Block) error {
-	if len(signers) == 0 {
-		return model.NewInsufficientSignaturesErrorf("empty list of signers")
-	}
+	msg := MakeVoteMessage(block.View, block.BlockID)
```
tarakby, line 125: Same question for moving `msg`

consensus/hotstuff/verification/staking_verifier.go
```diff
@@ -69,26 +70,26 @@ func (v *StakingVerifier) VerifyQC(signers flow.IdentityList, sigData []byte, bl
...
+		//  (i) In case no keys are provided, i.e. `len(signers) == 0`.
 		//      This scenario _is expected_ during normal operations, because a byzantine
 		//      proposer might construct an (invalid) QC with an empty list of signers.
 		// (ii) In case some provided public keys type is not BLS.
 		//      This scenario is _not expected_ during normal operations, because all keys are
 		//      guaranteed by the protocol to be BLS keys.
 		//
 		if crypto.IsBLSAggregateEmptyListError(err) {
-			return model.NewInvalidFormatErrorf("empty list of signers: %w", err)
+			return model.NewInsufficientSignaturesErrorf("empty list of signers: %w", err)
```
tarakby, line 84: Good catch

consensus/hotstuff/signature/packer_test.go
```diff
@@ -84,6 +84,34 @@ func TestPackUnpack(t *testing.T) {
 	require.Equal(t, expectedSignerIDs, signers.NodeIDs())
 }
 
+// TestUnpack_EmptySignerList verifies that `Unpack` gracefully handles the edge case
+// of an empty signer list, as such could be an input from a byzantine node.
+func TestPackUnpack_EmptySigners(t *testing.T) {
```
tarakby, line 89: Very nice!

tarakby:
Update: I ported all your suggestions I agree with to my branch (unfortunately I couldn't cherry-pick your code because the commits were not modular and I only needed some of the changes). 
I also merged my branch into yours so that we can clearly see the delta and what remains to discuss about.

consensus/hotstuff/model/errors.go
AlexHentschel: Are you sure? I think in this case, you might loose some information about the wrapped errors. I just typed up this example:
https://go.dev/play/p/8-RBskzm4oX
but I don't want to dismiss the possibility that I have a bug in my code ... its already late. See this example. I also added some explanation, why the implementation behaves that way.

module/signature/errors.go
```diff
@@ -19,6 +19,15 @@ var (
...
 	ErrInvalidChecksum = errors.New("index vector's checksum is invalid")
+
+	// ErrIdentitySignature indicates that a signature is an 'identity signature', which we
+	// reject as invalid by convention.
+	// Context: If nodes generate their public-private key pairs honestly, there is vanishing
+	// probability of generating identity signatures (individual as well as aggregated signatures).
+	// However, (colluding) byzantine nodes could force the generation of identity signatures, to
+	// undermine cryptographic safety. Therefore, we exclude the identity signature from the set
+	// of accepted signatures.
+	ErrIdentitySignature = errors.New("rejecting identity signature")
```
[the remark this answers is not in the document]
AlexHentschel: I have the soft preference to _not_ include statements about the logic leading to the error in the description of the error. If you use the error in a different location, the application-specific context usually doesn't carry over. 

For example, you write:
> failed the module safety final verification

But what if we wanted to use the error when checking an aggregated signature that somebody else sent us? For example, just hypothetically consider the scenario
*  `SignatureAggregatorSameMessage.VerifyAggregate` would respond only with error returns
* in this case, returning an `ErrIdentitySignature` would be reasonable from the perspective of `VerifyAggregate`
* but your error description specifically talks about "safety final verification", which `VerifyAggregate` does not do.

I am not sure I am able to explain this properly ... Just trying to say that in the context of the _current_ implementation, the error description makes sense ... but it imposes additional refactoring work (that often is overlooked) in case you want to use the same error in a different context. 

My recommendation is: 
* the error should describe the failure that you reached (identity signature) 
* and avoid describing the logic (safety final verification) that _leads_ to the failure case. As we document the error returns for each method, you can include the comments on the failure path in the method's documentation here: https://github.com/onflow/flow-go/blob/080d4ffbd04a927c7f2fa5178493ca8dcd9c886d/module/signature/aggregation.go#L169-L175

module/signature/aggregation.go
```diff
@@ -224,15 +221,15 @@ func (s *SignatureAggregatorSameMessage) Aggregate() ([]int, crypto.Signature, e
 // keys corresponding to the input signers.
 // Aggregating the keys of the signers internally is optimized to only look at the keys delta
 // compared to the latest execution of the function. The function is therefore not thread-safe.
+// Per convention, this function rejects the identity signature as invalid.
```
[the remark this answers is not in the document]
AlexHentschel: done

module/signature/aggregation.go
AlexHentschel: Well, that depends on the perspective. I assumed that `ErrIdentitySignature` might be used in other places too, and not only here. In context, your proposed error description ties the error exactly to this use case. See [my comment above](https://github.com/onflow/flow-go/pull/3677#discussion_r1051313145) for details

module/signature/aggregation.go
AlexHentschel: done

consensus/hotstuff/model/errors.go
AlexHentschel: applied

consensus/hotstuff/model/errors.go
AlexHentschel: I have the tendency to generalize errors as we move higher into the application stack. 
Similar to [my reasoning above](https://github.com/onflow/flow-go/pull/3677#discussion_r1051313145): the error is more general and the function returning the error carries the context that lead to the error: https://github.com/onflow/flow-go/blob/080d4ffbd04a927c7f2fa5178493ca8dcd9c886d/consensus/hotstuff/signature/weighted_signature_aggregator.go#L156-L158

But in this particular case, I am going against my own recommendation, because the error describes the application-specific context. Removed this context in

consensus/hotstuff/signature/weighted_signature_aggregator_test.go
AlexHentschel: changed godoc to `model.InvalidAggregatedSignatureError` in https://github.com/onflow/flow-go/pull/3677/commits/13d72333fe80c5ec9feada0c38d766a6dae1fee0

consensus/hotstuff/verification/combined_verifier_v2.go
AlexHentschel: 🤷 don't know. You are right, having the construction of `msg` right before verification is most intuitive. done in https://github.com/onflow/flow-go/pull/3677/commits/be9459382d0a82a8fdcf9a04358ad133f325519d

consensus/hotstuff/model/errors.go
tarakby: You are right that `Unwrap()` is required, otherwise `errors.As` wouldn't work. What I meant in the comment is that `Error()` is not needed, and it is inherited from the embedded field. 
I've updated your code to show that by keeping `Unwrap()` and removing all `Error()` definitions (by keeping the types you used). https://go.dev/play/p/Kk0bvxkeuhn

AlexHentschel: I was using unwrap to explain ... but you are right, the `Error()` method is subtly different, because the wrapped `error` actually contains the entire error string, and the wrapper does not add any additional text in this implementation. 💡

AlexHentschel:
all relevant suggestions have been included in PR https://github.com/onflow/flow-go/pull/3441

PR #3679, opened 2022-12-07
[FVM] Change payer balance check error to a failure

janezpodhostnik:
This should be a failure because it can only happen if the communication between the core contract and FVM is faulty, which means something isn't set up correctly.


zhangchiqing (approved):

pattyshack (approved):

janezpodhostnik:
bors merge

PR #3680, opened 2022-12-08
[Testing] Use localhost in unit tests to reduce firewall alerts

peterargue:
When running unit tests on a mac, there are many alert popups asking for permission to allow incoming connections. This is caused by using `0.0.0.0` as the listen address in the test setup. Update unit tests to instead use localhost.

### DIFF
--- a/utils/unittest/fixtures.go
+++ b/utils/unittest/fixtures.go
@@ -4,6 +4,7 @@ import (
 	crand "crypto/rand"
 	"fmt"
 	"math/rand"
+	"net"
 	"testing"
 	"time"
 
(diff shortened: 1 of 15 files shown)


utils/unittest/fixtures.go
```diff
@@ -42,8 +42,13 @@ import (
 
 const (
 	DefaultSeedFixtureLength = 64
+	DefaultAddress           = "localhost:0"
 )
 
+func IPPort(port string) string {
+	return fmt.Sprintf("localhost:%s", port)
```
SaveTheRbtz, line 49: nit: https://pkg.go.dev/net#JoinHostPort

SaveTheRbtz (approved):

gomisha (approved):

PR #3681, opened 2022-12-08
Upgrade Cadence

janezpodhostnik:
Cadence upgraded to https://github.com/onflow/cadence/commit/c1db9ec13ae369db0a0c1d7f90f63745b45187ed


fvm/meter/meter.go
```diff
@@ -48,7 +48,6 @@ var (
 
 		// Values
 
-		common.MemoryKindBoolValue:      8,
```
janezpodhostnik, line 51: @turbolent Some of these have been removed. Do some need to be added or adjusted?

PR #3685, opened 2022-12-08
add network collector from public Gater

koko1123:
Because the connection metrics have been moved, I have added the initialization for it in the observer, similar to the access node builder's metric init

### DIFF
--- a/cmd/access/node_builder/access_node_builder.go
+++ b/cmd/access/node_builder/access_node_builder.go
@@ -165,7 +165,7 @@ func DefaultAccessNodeConfig() *AccessNodeConfig {
 		apiBurstlimits:               nil,
 		PublicNetworkConfig: PublicNetworkConfig{
 			BindAddress: cmd.NotSet,
-			Metrics:     metrics.NewNoopCollector(),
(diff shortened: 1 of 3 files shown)


cmd/access/node_builder/access_node_builder.go
```diff
@@ -165,7 +165,7 @@ func DefaultAccessNodeConfig() *AccessNodeConfig {
 		apiBurstlimits:               nil,
 		PublicNetworkConfig: PublicNetworkConfig{
 			BindAddress: cmd.NotSet,
-			Metrics:     metrics.NewNoopCollector(),
+			Metrics:     metrics.NewNetworkCollector(zerolog.Logger{}),
```
peterargue, line 168: can you move this initialization to where the public network is set so it can use the node's actual logger

peterargue (approved):
looks good. can you double check the inbound/outbound connection count metrics are populated when running on localnet?

diaswrd (approved):

diaswrd:
Closing since this has been addressed at https://github.com/onflow/flow-go/pull/3716

PR #3687, opened 2022-12-09
[Spam Testing] Fixes race conditions 

yhassanzadeh13:
This PR fixes the race conditions (both internally as well as in interaction with GossipSub) and flakiness with the spam testing of `iHave` control messages.


network/internal/p2pfixtures/fixtures.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
```
yhassanzadeh13, line 7: Nothing changed in this file, just some functions re-ordered due to move and revert-move.

PR #3688, opened 2022-12-09
Change info level log for provider engine

zhangchiqing:
Changing the provider engine's log level into INFO for making debugging execution halt issue easier

### DIFF
--- a/engine/common/provider/engine.go
+++ b/engine/common/provider/engine.go
@@ -189,7 +189,7 @@ func (e *Engine) onEntityRequest(request *internal.EntityRequest) error {
 		Strs("entity_ids", flow.IdentifierList(request.EntityIds).Strings()).
 		Logger()
 
-	lg.Debug().
+	lg.Info().
 		Msg("entity request received")
 
 	// TODO: add reputation system to punish nodes for malicious behaviour (spam / repeated requests)
@@ -263,7 +263,7 @@ func (e *Engine) onEntityRequest(request *internal.EntityRequest) error {
 	}
 
 	e.metrics.MessageSent(e.channel.String(), metrics.MessageEntityResponse)
-	lg.Debug().Msg("entity response sent")
+	lg.Info().Msg("entity response sent")
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


engine/common/provider/engine.go
```diff
@@ -189,7 +189,7 @@ func (e *Engine) onEntityRequest(request *internal.EntityRequest) error {
 		Strs("entity_ids", flow.IdentifierList(request.EntityIds).Strings()).
 		Logger()
 
-	lg.Debug().
+	lg.Info().
```
zhangchiqing, line 192: The provider engine is only used by:
1. [collection engine returning collections to execution nodes](https://github.com/onflow/flow-go/blob/leo/info-log-provider-engine/cmd/collection/main.go#L395)
2. [execution nodes returning receipts to consensus nodes](https://github.com/onflow/flow-go/blob/master/cmd/execution_builder.go#L911)

The first log is very useful in debugging execution halt issue, where we need to distinguish between whether execution halt is due to execution node not sending collection requests or collection node not responding collection responses. If this log is debug level, then when execution is halted, we no longer have the log to see check whether the provider engine received the request.

The problem for changing this is it might increase the amount of logs, but this is ok, it's not too many logs, I have tested, it only add about 1-2 logs per second for collection node.
And for execution node, since consensus only request receipts when falling behind, this log should be rare.

engine/common/requester/engine.go
```diff
@@ -415,7 +415,7 @@ func (e *Engine) dispatchRequest() (bool, error) {
 
 	err = e.con.Unicast(req, providerID)
 	if err != nil {
-		return true, fmt.Errorf("could not send request: %w", err)
+		return true, fmt.Errorf("could not send request for entities %v: %w", logging.IDs(entityIDs), err)
```
zhangchiqing, line 418: This change would be useful for answering questions like why EN didn't receive the collection, did it fail to send the request for the collection?

Now since we log the entity ID, if we look for log for the collection ID on execution node, we could know whether it failed to request the collection

peterargue (approved):

SaveTheRbtz (approved):

PR #3690, opened 2022-12-09
[Build] Remove manual install of golangci-lint

SaveTheRbtz:
non-content addressable `curl | sh` does not look like a best security practice.  Let's see what breaks if we remove it.


pattyshack (approved):

gomisha (approved):

SaveTheRbtz:
bors merge

Issue #2976, opened 2022-08-15
[Access] Add metrics to the rest api


peterargue:
## Problem Definition

The REST api currently does not collect any metrics, so we lack visibility into key metrics.

## Proposed Solution

Add metrics collection to the Access node's REST api similar to what exists for the gRPC api.\

PR #3691, opened 2022-12-09
[Access] Add REST API metrics middleware

diaswrd:
Closes https://github.com/onflow/flow-go/issues/2976

This PR introduces the following metrics for the [Access REST API](https://developers.flow.com/http-api):
````
access_rest_api_http_requests_inflight
access_rest_api_http_requests_total
access_rest_api_http_request_duration_seconds_bucket
access_rest_api_http_request_duration_seconds_count
access_rest_api_http_request_duration_seconds_sum
access_rest_api_http_response_size_bytes_bucket
access_rest_api_http_response_size_bytes_count
access_rest_api_http_response_size_bytes_sum
````

Screen Shot 2023-02-08 at 3 49 41 PM

### CODE REVIEW

File: engine/access/rest/middleware/metrics.go [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	r := &recorder{
+		httpRequestDurHistogram: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: cfg.Prefix,
+			Subsystem: "http",
+			Name:      "request_duration_seconds",
+			Help:      "The latency of the HTTP requests.",
+			Buckets:   cfg.DurationBuckets,
+		}, []string{cfg.ServiceLabel, cfg.HandlerIDLabel, cfg.MethodLabel, cfg.StatusCodeLabel}),
```
peterargue, line 70: do handler and method map 1:1 or are there cases where they are different?

peterargue (commented):
This looks good. Can you move the metrics portion into https://github.com/onflow/flow-go/tree/master/module/metrics along with our other metrics? Also please integrate it with the `NoopCollector` so we can stub it out in tests

engine/access/rest/middleware/metrics.go
diaswrd: Method in this case is the type of HTTP method (like `GET` or `POST`) while Handler is the actual HTTP path of the request (like `/v1/blocks` for example). And both are always present in the HTTP request, if that answers your question, let me know 👀

diaswrd:
This is stalled for a while as we're dealing with more important priorities on the Archive node and Execution Data sync projects at the moment. I plan to rework some of this implementation to reuse an already defined prometheus registry instead of automatically adding a new one with promauto (which works but breaks some unit tests).

koko1123 (approved):
Looks good to me, thanks for the hands-on help with this!

lolpuddle (approved):

module/metrics/rest_api.go
```diff
@@ -0,0 +1,137 @@
...
+	metricsProm "github.com/slok/go-http-metrics/metrics/prometheus"
+	"github.com/slok/go-http-metrics/middleware"
+	"github.com/slok/go-http-metrics/middleware/std"
+
+	"github.com/gorilla/mux"
+)
+
+// Example recorder taken from:
+// https://github.com/slok/go-http-metrics/blob/master/metrics/prometheus/prometheus.go
+type CustomRecorder interface {
```
peterargue, line 20:
```suggestion
type RestCollector interface {
```

module/metrics/rest_api.go
```diff
@@ -0,0 +1,137 @@
...
+type recorder struct {
+	httpRequestDurHistogram   *prometheus.HistogramVec
+	httpResponseSizeHistogram *prometheus.HistogramVec
+	httpRequestsInflight      *prometheus.GaugeVec
+	httpRequestsTotal         *prometheus.GaugeVec
+}
+
+// NewRecorder returns a new metrics recorder that implements the recorder
+// using Prometheus as the backend.
+func NewRecorder(cfg metricsProm.Config) CustomRecorder {
```
peterargue, line 34: we typically use the term `Collector` for metrics objects
```suggestion
func NewRestCollector(cfg metricsProm.Config) RestCollector {
```

module/metrics/rest_api.go
```diff
@@ -0,0 +1,137 @@
...
+			r.AddTotalRequests(req.Context(), req.Method, req.URL.Path)
+
+			// Modify the writer
+			respWriter := &responseWriter{w, http.StatusOK}
+
+			// Record go-http-metrics/middleware metrics and continue to the next handler
+			std.Handler("", metricsMiddleware, next).ServeHTTP(respWriter, req)
+		})
+	}
+}
```
peterargue, line 119: from an organizational perspective, I think the `MetricsMiddleware` should be defined in `engine/access/rest` to keep it with the other HTTP server functionality

It could then accept a `RestCollector` implementation as an argument. This aligns more closely with our other uses of metrics collectors, and simplifies testing with the `NoopCollector`

module/metrics/noop.go
```diff
@@ -234,6 +234,7 @@ func (nc *NoopCollector) PrunedBlocks(totalByHeight, totalById, storedByHeight,
 func (nc *NoopCollector) RangeRequested(ran chainsync.Range)                                    {}
 func (nc *NoopCollector) BatchRequested(batch chainsync.Batch)                                  {}
 func (nc *NoopCollector) OnUnauthorizedMessage(role, msgType, topic, offense string)            {}
+func (nc *NoopCollector) MetricsMiddleware()                                                    {}
```
peterargue, line 237: These would be the metrics methods (e.g. `ObserveHTTPRequestDuration`). The idea is that in tests you can instantiate a `NoopCollector` and pass it into any method that expects a metrics object.

module/metrics/rest_api.go
```diff
@@ -0,0 +1,137 @@
+package metrics
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/slok/go-http-metrics/metrics"
```
peterargue, line 10: maybe give this an alias since this package is also called `metrics`. This should still work as-is, but renaming it would be less confusing.
```suggestion
	httpmetrics "github.com/slok/go-http-metrics/metrics"
```

peterargue (commented):
sorry for the many nits, but I'm hoping to align this with our other metrics implementations for a consistent interface

peterargue (approved):
looks great!

PR #3692, opened 2022-12-09
[Provider Engine] log actual entities in the response

zhangchiqing:


engine/common/provider/engine.go
```diff
@@ -263,7 +266,11 @@ func (e *Engine) onEntityRequest(request *internal.EntityRequest) error {
 	}
 
 	e.metrics.MessageSent(e.channel.String(), metrics.MessageEntityResponse)
-	lg.Info().Msg("entity response sent")
+	e.log.Info().
+		Str("origin_id", request.OriginId.String()).
+		Strs("entity_ids", flow.IdentifierList(entityIDs).Strings()).
```
zhangchiqing, line 271: The old log was logging the requested entity ids, which isn't useful because they are logged in the `entity request received` log already. 

Instead, we will log the actual entities sent in the response. 

I also added the nonce to the logs for both request and response. So that we can pair them up and check the entities ids in the request and entities ids in the response to know whether the provider had some specific entities that were requested.

jordanschalm (approved):

engine/common/provider/engine.go
```diff
@@ -189,7 +190,9 @@ func (e *Engine) onEntityRequest(request *internal.EntityRequest) error {
 		Strs("entity_ids", flow.IdentifierList(request.EntityIds).Strings()).
 		Logger()
 
+	nonce := rand.Uint64()
```
peterargue, line 193: we should use `request.Nonce`

m4ksio (approved):

peterargue (approved):

zhangchiqing:
bors merge

PR #3698, opened 2022-12-09
[Consensus] TC aggregation multi-message verification fix

durkmurder:
### Context

This PR fixes a bug which can occur under specific conditions and results in consensus halt effectively undermining liveness. 
Replicas perform a multi-message BLS aggregation to build a TC. To verify such signature we need to provide list of signers and messages used to construct the signature. We provide this data as part of `flow.TimeoutCertificate` structure as two separate lists `SignerIndices` and `NewestQCViews`. Strictly speaking TC verification depends on correct relation between signers and messages, should the order of signers or messages be altered and we cannot verify the signature anymore. 
And that's exactly what has happened when replicas contributed with different messages, signers got sorted in canonical order because of signer indices implementation but messages weren't which resulted in inability to verify the signature.

Proposed solution is to properly order both signers and messages when constructing a TC. As part of this PR I have modified our test to represent this scenario to make sure that problem is indeed solved.

### Motivation for implementation

There were two ways how to approach it:
1. Sort data in `TimeoutSignatureAggregator.Aggregate()`
2. Sort data in `TimeoutProcessor.buildTC()`.

While the 1. would be simpler to implement I have picked 2. because this resulted in more clear separation of concepts. Sig aggregator doesn't know anything what we are going to do aggregated data, it doesn't require ordered data, doesn't know anything about TC structure and signer indices. On other side `TimeoutProcessor` knows about all of those things and can make a decision to properly order data knowing that signer indices will introduce canonical ordering.

### DIFF
--- a/consensus/hotstuff/signature.go
+++ b/consensus/hotstuff/signature.go
@@ -131,8 +131,8 @@ type TimeoutSignatureAggregator interface {
 	Aggregate() (signersInfo []TimeoutSignerInfo, aggregatedSig crypto.Signature, exception error)
 }
 
-// TimeoutSignerInfo is a helper structure that stores what QC views each signer has contributed into TC's aggregated signature.
-// Used as result of TimeoutSignatureAggregator.Aggregate()
+// TimeoutSignerInfo is a helper structure that stores the QC views that each signer
+// contributed to a TC. Used as result of TimeoutSignatureAggregator.Aggregate()
 type TimeoutSignerInfo struct {
 	NewestQCView uint64
 	Signer       flow.Identifier
--- a/consensus/hotstuff/signature.go
+++ b/consensus/hotstuff/signature.go
@@ -128,7 +128,14 @@ type TimeoutSignatureAggregator interface {
 	// Caller can be sure that resulting signature is valid.
 	// Expected errors during normal operations:
 	//  - model.InsufficientSignaturesError if no signatures have been added yet
-	Aggregate() (signers []flow.Identifier, newestQCViews []uint64, aggregatedSig crypto.Signature, exception error)
+	Aggregate() (signersInfo []TimeoutSignerInfo, aggregatedSig crypto.Signature, exception error)
+}
+
+// TimeoutSignerInfo is a helper structure that stores what QC views each signer has contributed into TC's aggregated signature.
+// Used as result of TimeoutSignatureAggregator.Aggregate()
+type TimeoutSignerInfo struct {
+	NewestQCView uint64
+	Signer       flow.Identifier
 }
 
 // BlockSignatureData is an intermediate struct for Packer to pack the
--- a/consensus/hotstuff/timeoutcollector/timeout_processor.go
+++ b/consensus/hotstuff/timeoutcollector/timeout_processor.go
@@ -251,10 +251,11 @@ func (p *TimeoutProcessor) buildTC() (*flow.TimeoutCertificate, error) {
 	}
 
 	// IMPORTANT: To properly verify an aggregated signature included in TC we need to provide list of signers with corresponding
-	// messages(`TimeoutCertificate.NewestQCViews`) for each signer, this relation should be very strict.
+	// messages(`TimeoutCertificate.NewestQCViews`) for each signer. If the one-to-once correspondence of view and signer is not
+	// maintained, the TC's aggregated signature will be invalid. 
 	// Aggregate returns an unordered set of signers together with additional data.
-	// Due to implementation specifics of signer indices decoding step results in canonically ordered
-	// signer ids which means we need to canonically order messages, so we can properly map signer to message after decoding.
+	// Due to implementation specifics of signer indices, the decoding step results in canonically ordered signer ids, which means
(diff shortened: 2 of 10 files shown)


jordanschalm (approved):

consensus/hotstuff/signature.go
```diff
@@ -128,7 +128,14 @@ type TimeoutSignatureAggregator interface {
 	// Caller can be sure that resulting signature is valid.
 	// Expected errors during normal operations:
 	//  - model.InsufficientSignaturesError if no signatures have been added yet
-	Aggregate() (signers []flow.Identifier, newestQCViews []uint64, aggregatedSig crypto.Signature, exception error)
+	Aggregate() (signersInfo []TimeoutSignerInfo, aggregatedSig crypto.Signature, exception error)
+}
+
+// TimeoutSignerInfo is a helper structure that stores what QC views each signer has contributed into TC's aggregated signature.
+// Used as result of TimeoutSignatureAggregator.Aggregate()
```
AlexHentschel, line 135:
```suggestion
// TimeoutSignerInfo is a helper structure that stores the QC views that each signer
// contributed to a TC. Used as result of TimeoutSignatureAggregator.Aggregate()
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -243,17 +245,36 @@ func (p *TimeoutProcessor) validateTimeout(timeout *model.TimeoutObject) error {
...
 // Any error should be treated as exception.
 func (p *TimeoutProcessor) buildTC() (*flow.TimeoutCertificate, error) {
-	signers, highQCViews, aggregatedSig, err := p.sigAggregator.Aggregate()
+	signersData, aggregatedSig, err := p.sigAggregator.Aggregate()
 	if err != nil {
 		return nil, fmt.Errorf("could not aggregate multi message signature: %w", err)
 	}
 
+	// IMPORTANT: To properly verify an aggregated signature included in TC we need to provide list of signers with corresponding
+	// messages(`TimeoutCertificate.NewestQCViews`) for each signer, this relation should be very strict.
```
AlexHentschel, line 254:
```suggestion
	// messages(`TimeoutCertificate.NewestQCViews`) for each signer. If the one-to-once correspondence of view and signer is not
	// maintained, the TC's aggregated signature will be invalid. 
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -243,17 +245,36 @@ func (p *TimeoutProcessor) validateTimeout(timeout *model.TimeoutObject) error {
...
+	// Aggregate returns an unordered set of signers together with additional data.
+	// Due to implementation specifics of signer indices decoding step results in canonically ordered
+	// signer ids which means we need to canonically order messages, so we can properly map signer to message after decoding.
```
AlexHentschel, line 257:
```suggestion
	// Due to implementation specifics of signer indices, the decoding step results in canonically ordered signer ids, which means
	// we need to canonically order the respective `newestQCView`, so we can properly map signer to `newestQCView` after decoding.
```

AlexHentschel (approved):
Love it. Found the time for a proper review. Only some polishing of comments.

consensus/hotstuff/signature.go
```diff
@@ -128,7 +128,14 @@ type TimeoutSignatureAggregator interface {
 	// Caller can be sure that resulting signature is valid.
 	// Expected errors during normal operations:
 	//  - model.InsufficientSignaturesError if no signatures have been added yet
-	Aggregate() (signers []flow.Identifier, newestQCViews []uint64, aggregatedSig crypto.Signature, exception error)
+	Aggregate() (signersInfo []TimeoutSignerInfo, aggregatedSig crypto.Signature, exception error)
```
tarakby, line 131: 👌🏼

consensus/hotstuff/timeoutcollector/timeout_processor.go
tarakby: > the TC's aggregated signature will be invalid.

An aggregated signature is never invalid on its own. It all depends on how the verification is performed : It is valid if it is verified against the correct list of (pub_key, message) that initially generated the signature, and it is invalid if it is verified against a different list. 
I would suggest a slightly different comment : 
"If the one-to-once correspondence of view and signer is not maintained, it won't be possible to verify the aggregated signature"

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -243,17 +245,36 @@ func (p *TimeoutProcessor) validateTimeout(timeout *model.TimeoutObject) error {
...
+
+	// sort data in canonical order
+	slices.SortFunc(signersData, func(lhs, rhs hotstuff.TimeoutSignerInfo) bool {
```
tarakby, line 260: 👌🏼

tarakby (approved):
Nice I like this way of fixing the issue. As we mentioned during the call, option 1. isn't clean and implicitly relies on the aggregation module using the same signer ordering used by the signer indices packer.

durkmurder:
bors merge

durkmurder:
bors merge

PR #3700, opened 2022-12-09
[CD] Automate without-netgo image building

peterargue:
Currently, the docker images without `netgo` require an extra manual step to build. Add them into CD so they're built for every version

### DIFF
--- a/cmd/Dockerfile
+++ b/cmd/Dockerfile
@@ -25,12 +25,15 @@ RUN --mount=type=cache,sharing=locked,target=/go/pkg/mod \
 
 FROM build-env as build-production
 WORKDIR /app
+
 ARG GOARCH=amd64
+ARG TAGS="relic,netgo"
+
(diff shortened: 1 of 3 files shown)


sjonpaulbrown (approved):

cmd/Dockerfile
```diff
@@ -25,12 +25,15 @@ RUN --mount=type=cache,sharing=locked,target=/go/pkg/mod \
 
 FROM build-env as build-production
 WORKDIR /app
+
 ARG GOARCH=amd64
+ARG TAGS="relic,netgo"
```
gomisha, line 30: can you add a comment here explaining that `TAGS` is meant to be override if you don't want to build with `netgo` tag

gomisha (approved):
Minor comment, otherwise LGTM

peterargue:
bors merge

PR #3702, opened 2022-12-11
Fix flaky unicast rate limit messages test

kc1116:
This PR does the following
- Removes the usage of the testtime utility and instead sends 6 messages immediately. 
- Adds flaky test flag so that we can gather flaky test metrics before reenabling


peterargue (commented):
when I run this locally with `-count=10` it now fails with
```
--- FAIL: TestMiddlewareTestSuite (0.03s)
    --- FAIL: TestMiddlewareTestSuite/TestUnicastRateLimit_Messages (0.03s)
        middleware_test.go:287: 
                Error Trace:    /Users/pargue/go/src/github.com/onflow/flow-go/network/test/middleware_test.go:287
                Error:          Not equal: 
                                expected: 0x1
                                actual  : 0x0
                Test:           TestMiddlewareTestSuite/TestUnicastRateLimit_Messages
        network_metrics.go:247: PASS:   NetworkMessageReceived(string,string,string)
FAIL
FAIL    github.com/onflow/flow-go/network/test  0.724s
FAIL
```

Seems like the test isn't allowing all messages to process before ending.

peterargue (approved):

yhassanzadeh13 (approved):

PR #3703, opened 2022-12-12
[Execution] Remove runtime memory stats

SaveTheRbtz:
Runtime memory stats have non-zero overhead but quite useless since script execution is happening alongside with transactions and other scripts.  Since script execution is the main user of allocator, the data we get from the `GetHeapAllocsBytes` is useless.


peterargue:
I thought this was being used to get a sense of how far off our estimates were from the actual memory used (by taking the lowest value across all of the nodes). Do we have a better way to determining that now?

SaveTheRbtz:
> I thought this was being used to get a sense of how far off our estimates were from the actual memory used

Sadly, data there is very inaccurate since it relies on the fact that nothing is happening in the system in parallel.  The better way would just do that in unit tests where we can fully control the environment.

peterargue (approved):

Issue #3015, opened 2022-08-18
Improve FVM Program Cache Invalidation


j1010001:
Problem
Cadence programs (transactions, contracts, scripts) are submitted in source code form. This means we need to parse and type check the programs before we can execute them. As some contracts are used very often, e.g. the Flow token contract, or the Top Shot contract, we cache the results of parsing and type checking.

However, we also support contract updates, which means the cache needs to be invalidated.

Properly invalidating the cache requires dependency tracking, which complicates the cache even more (it is already complicated, as it needs to be fork-aware). As a short-term solution we went with the simplest solution: all contracts are invalidated.

This is reasonable for few contracts. However, with permissionless deployment contract updates will likely be frequent, or this cache invalidation could be attacked

We’ll want to find a solution for this, e.g. making it financially expensive, before enabling permissionless deployment, otherwise we’ll risk a performance degredation

Questions
What is the performance impact?
Solutions/Ideas
Make contract operations expensive (updates, maybe also initial deployment)
Currently fee calculation has low maximum fee (1 cent)
We just completed a vote on fees
We can add an extra cost to the fee contract (deductTransactionFee), easy
“Problem” is community, maybe want to announce, and maybe even allow vote (votes are difficult)
Given that impact
Implement proper cache invalidation which only invalidates contracts that need to be
Requires maintaining a dependency tree
This could also be implemented and deployed in the middle of a spork
Lazy invalidation? Avoids exploitation impact on unrelated contracts

@j1010001 commented on https://github.com/dapperlabs/cadence-private-issues/issues/49#issuecomment-1146241087

FVM issue

@turbolent commented on https://github.com/dapperlabs/cadence-private-issues/issues/49#issuecomment-1146245762

I opened https://github.com/onflow/cadence/pull/1634, which adds a dependency resolution package to Cadence.
It could be used to calculate which contracts need to be rechecked, instead of wiping the cache and re-checking all contracts

bluesign:
> “Problem” is community, maybe want to announce, and maybe even allow vote (votes are difficult) 

:))))))

bluesign:
> Cadence programs (transactions, contracts, scripts) are submitted in source code form. This means we need to parse and type check the programs before we can execute them. As some contracts are used very often, e.g. the Flow token contract, or the Top Shot contract, we cache the results of parsing and type checking.

Can't we do something like what `linker` is doing ? Cache programs without dependencies, link those at runtime. 

[...]

PR #3706, opened 2022-12-12
[FVM] Smart/partial programs cache invalidation

janezpodhostnik:
ref:  [Improve FVM Program Cache Invalidation](https://github.com/onflow/flow-go/issues/3015)

Smart contract invalidation.

This relies on Cadence properly calling Get/Set program (which should already be the case, because other parts of the code also depend on this). The proper use of Get/Set will be guaranteed once we replace Get/Set with a single `GetOrCreate(location, create_function)` function. This will be a future PR.

The principle is that we store program dependencies together with the program in the cache. These dependencies are filled when cadence calls Get/Set program.

**Example**
```mermaid
graph TD
  C-->D
  C-->B
  B-->A
```
With this dependency graph, if B is already in cache and we are loading C:

- Get C:
    - Get B. We have B with dependency A, so A is added to C as a dependency
    - Get D. 
        - D  is not loaded so load it here
    - Set D. and add it as a dependency of C
- Set C with dependencies { A, B, D, C[^1]} 

[^1]: C is listed as its own dependency to make some code simpler.

### DIFF
--- a/fvm/environment/derived_data_invalidator.go
+++ b/fvm/environment/derived_data_invalidator.go
@@ -5,11 +5,10 @@ import (
 
 	"github.com/onflow/flow-go/fvm/derived"
 	"github.com/onflow/flow-go/fvm/state"
-	"github.com/onflow/flow-go/model/flow"
 )
 
 type ContractUpdateKey struct {
-	Address flow.Address
+	Address common.Address
 	Name    string
 }
 
@@ -94,19 +93,23 @@ func (invalidator ProgramInvalidator) ShouldInvalidateEntry(
 	state *state.State,
 ) bool {
 	if invalidator.MeterParamOverridesUpdated {
+		// if meter parameters changed we need to invalidate all programs
 		return true
 	}
 
-	if len(invalidator.FrozenAccounts) > 0 {
-		// TODO: switch to fine grain invalidation.
-		return true
+	// if an account was (un)frozen we need to invalidate all
+	// programs that depend on any contract on that address.
+	for _, frozenAccount := range invalidator.FrozenAccounts {
+		_, ok := program.Dependencies[frozenAccount]
+		if ok {
+			return true
+		}
 	}
 
+	// invalidate all programs depending on any of the contracts that were updated
+	// A program has itself listed as a dependency, so that this simpler.
 	for _, key := range invalidator.ContractUpdateKeys {
-		_, ok := program.Dependencies[common.AddressLocation{
-			Address: common.MustBytesToAddress(key.Address.Bytes()),
-			Name:    key.Name,
-		}]
+		_, ok := program.Dependencies[key.Address]
 		if ok {
 			return true
 		}
--- a/fvm/environment/derived_data_invalidator.go
+++ b/fvm/environment/derived_data_invalidator.go
@@ -93,8 +93,25 @@ func (invalidator ProgramInvalidator) ShouldInvalidateEntry(
 	program *derived.Program,
 	state *state.State,
 ) bool {
-	// TODO(rbtz): switch to fine grain invalidation.
-	return invalidator.ShouldInvalidateEntries()
+	if invalidator.MeterParamOverridesUpdated {
+		return true
+	}
+
+	if len(invalidator.FrozenAccounts) > 0 {
+		// TODO: switch to fine grain invalidation.
+		return true
+	}
+
+	for _, key := range invalidator.ContractUpdateKeys {
+		_, ok := program.Dependencies[common.AddressLocation{
+			Address: common.MustBytesToAddress(key.Address.Bytes()),
+			Name:    key.Name,
+		}]
+		if ok {
+			return true
+		}
+	}
+	return false
 }
 
 type MeterParamOverridesInvalidator struct {
--- a/fvm/environment/derived_data_invalidator.go
+++ b/fvm/environment/derived_data_invalidator.go
@@ -2,7 +2,6 @@ package environment
 
 import (
 	"github.com/onflow/cadence/runtime/common"
-	"github.com/onflow/cadence/runtime/interpreter"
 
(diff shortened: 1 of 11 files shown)


fvm/environment/programs.go
```diff
@@ -127,6 +138,47 @@ func (programs *Programs) get(
...
 }
 
+// popDependencyTracker removes the dependencyTracker for the given location,
+// and returns the list of dependencies.
+func (programs *Programs) popDependencyTracker(
+	location common.AddressLocation,
+) derived.ProgramDependencies {
+	dependencies, ok := programs.dependencyTracker[location]
+	if !ok {
+		return make(derived.ProgramDependencies)
```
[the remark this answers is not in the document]
janezpodhostnik: I have concerns with cadence calling Set without calling Get first. There seems to be a couple of instances of that. I 'll verify they don't mess this up and change this to a stack.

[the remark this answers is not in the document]
pattyshack: txnState BeginParseRestrictedNestedTransaction / CommitParseRestricted already guard against this.  if commit and called before begin, this would error out

pattyshack:
o, btw, this may not be correct.  this handles explicit path import, but not *-style imports (aka import all).

e.g., if a unrelated program to the address, the next time the address is parsed, that program will also get imported, causing a spock hash change.

pattyshack:
you'll probably need some changes on the cadence side to figure out the star imports

pattyshack:
alternatively, we can just ignore the contract name, and invalid based only on address.  this will still be more fine grain than wholesale invalidation and is correct.  up to you

janezpodhostnik:
Good point on star imports, but I think it might be fine.

If contract `B` imports everything from address `a` this still gets resolved into concrete contracts `A1` and `A2` which will be listed as dependencies. If an `A3` is deployed on a this should have no effect on existing contract `B` because that contract doesn't resolve into `A3`. But that is just my understanding. @turbolent any thoughts about this? 

I will add a test.

pattyshack:
A3 won't impact the program's execution, but importing A3 (vs not importing) will change the spock hash.  you're end up with a fork

janezpodhostnik:
Switched to address level invalidation to cover the case with star imports. And changed tracking dependencies to a proper stack.

fvm/environment/derived_data_invalidator.go
```diff
@@ -91,11 +90,32 @@ func (invalidator ProgramInvalidator) ShouldInvalidateEntries() bool {
...
+		_, ok := program.Dependencies[frozenAccount]
+		if ok {
+			return true
+		}
+	}
+
+	// invalidate all programs depending on any of the contracts that were updated
+	// A program has itself listed as a dependency, so that this simpler.
+	for _, key := range invalidator.ContractUpdateKeys {
+		_, ok := program.Dependencies[common.MustBytesToAddress(key.Address.Bytes())]
```
pattyshack, line 113: we should precompute `common.MustBytesToAddress(key.Address.Bytes())` instead of using key directly / recomputing the same address for each program entry

fvm/environment/programs.go
```diff
@@ -196,3 +216,65 @@ func (programs *Programs) DecodeArgument(
...
+			nil,
+			fmt.Errorf("cannot pop the programs dependency stack, because it is empty")
+	}
+
+	// pop the last tracker
+	tracker := s.trackers[l-1]
+	s.trackers = s.trackers[:l-1]
+	if l > 1 {
+		// add the dependencies to the parent tracker
+		s.trackers[l-2].dependencies.Merge(tracker.dependencies)
```
pattyshack, line 277: off by one error?  you've already pop the tracker

fvm/environment/programs.go
```diff
@@ -82,7 +87,19 @@ func (programs *Programs) set(
 		return err
 	}
 
-	programs.derivedTxnData.SetProgram(address, program, state)
+	location, dependencies, err := programs.dependencyStack.pop()
+	if err != nil {
+		return err
+	}
+	if location != address {
```
pattyshack, line 94: this check and the pop stack empty check is probably not necessary.  i don't feel strongly about this / feel free to keep it around.

pattyshack (approved):

fvm/environment/programs.go
janezpodhostnik: Added comment.

fvm/environment/programs.go
```diff
@@ -105,16 +122,19 @@ func (programs *Programs) get(
...
 			panic(fmt.Sprintf(
 				"merge error while getting program, panic: %s",
 				err))
 		}
 
-		return program, true
+		return program.Program, true
 	}
 
+	programs.dependencyStack.push(address)
```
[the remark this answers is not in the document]
janezpodhostnik: added comments

janezpodhostnik:
@zhangchiqing I added a bunch of comments. Can you let me know if everything is clearer now?

fvm/environment/programs.go
```diff
@@ -82,7 +87,30 @@ func (programs *Programs) set(
...
 
-	programs.derivedTxnData.SetProgram(address, program, state)
+	// Get collected dependencies of the loaded program.
+	stackLocation, dependencies, err := programs.dependencyStack.pop()
+	if err != nil {
+		return err
+	}
+	if stackLocation != address {
+		// This should never happen, and indicates an implementation error.
+		// Get and set should be always called in pair.
```
zhangchiqing, line 97: Am I right here?
```suggestion
		// GetProgram and SetProgram should be always called in pair, this check depends on this assumption.
		// TODO: in future, the GetPrgram and SetProgram method will be replaced
		// with `GetOrCreateProgram`
```

zhangchiqing (approved):
Looks good, added a few suggestions on comments.

janezpodhostnik:
bors merge

PR #3707, opened 2022-12-12
Fix bug in cluster block validation. 

jordanschalm:
This check was duplicated to before we verified existence of the reference block. This caused an empty collection with an unknown reference block to pass validation. This breaks an assumption about the cluster state and would cause an irrecoverable error to be thrown later.


AlexHentschel (approved):

jordanschalm:
bors merge

PR #3708, opened 2022-12-12
fix race condition in PeerManager by starting libp2p node after start…

kc1116:
This PR fixes a race condition in the `TestUnicastAuthorizationTestSuite` where starting the libp2p nodes before starting the middlewares will cause a race condition with access to the peer manager.


gomisha (approved):

kc1116:
@gomisha I added another commit to fix the same race condition in the middleware test suite https://github.com/onflow/flow-go/pull/3708/commits/4037e1c9ff0e59e647b0006a65b26daa9e0cd28d

PR #3715, opened 2022-12-14
parallelize signature verification

pattyshack:
1. moved all inexpensive check ahead of crypto library calls
2. parallelize crypto library call

Note: there's no observed difference in speed in benchmarks since the benchmark transactions all involve a very small number of signatures (there's basically no parallelism to be had).  We can expect minor speed up for transactions with large number of signatures

### DIFF
--- a/fvm/transactionVerifier.go
+++ b/fvm/transactionVerifier.go
@@ -1,7 +1,9 @@
 package fvm
 
 import (
+	"context"
 	"fmt"
+	"sync"
 
 	"go.opentelemetry.io/otel/attribute"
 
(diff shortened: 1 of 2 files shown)


fvm/transactionVerifier.go
```diff
@@ -141,73 +237,121 @@ func (v *TransactionVerifier) verifyTransaction(
...
 
-	if valid {
+	for _, entry := range signatures {
+		toVerifyChan <- entry
+	}
+	close(toVerifyChan)
+
+	foundError := false
+	for i := 0; i < len(signatures); i++ {
+		entry := <-verifiedChan
```
SaveTheRbtz, line 326: this would stall if goroutine above exits earlier.

fvm/transactionVerifier.go
```diff
@@ -141,73 +237,121 @@ func (v *TransactionVerifier) verifyTransaction(
...
-// If the signature is valid, this function returns the associated account key.
-//
-// An error is returned if the account does not contain a public key that
-// correctly verifies the signature against the given message.
-func (v *TransactionVerifier) verifyAccountSignature(
-	accountKey flow.AccountPublicKey,
-	txSig flow.TransactionSignature,
-	message []byte,
-	errorBuilder func(flow.TransactionSignature, error) errors.CodedError,
+func (v *TransactionVerifier) verifyAccountSignatures(
```
SaveTheRbtz, line 276: you may want to add a test for this one-- it is quite compelx.

fvm/transactionVerifier.go
```diff
@@ -141,73 +237,121 @@ func (v *TransactionVerifier) verifyTransaction(
...
+				}
+			}
+		}()
 	}
 
-	if valid {
+	for _, entry := range signatures {
+		toVerifyChan <- entry
+	}
+	close(toVerifyChan)
```
SaveTheRbtz, line 322: nit: you may want to move this to the top of the function.

pattyshack: your suggestion serializes queuing ...

in the current form, the goroutines could start processing in the background while the main function is queuing up the entries

your proposal is to queue up all the entires first, then start the goroutines

SaveTheRbtz (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

fvm/transactionVerifier.go
```diff
@@ -14,6 +16,123 @@ import (
...
+		address flow.Address
+		index   uint64
+	}
+	duplicate := make(map[uniqueKey]struct{}, numSignatures)
+
+	for _, group := range list {
+		for _, signature := range group.signatures {
+			entry := &signatureEntry{
+				TransactionSignature: signature,
+				signatureType:        group.signatureType,
```
zhangchiqing, line 115: Why constructing an entry structure without `accountKey`?

Would this create a possibility of hitting nil panic when reading `entry.accountKey.PublicKey` in `verify()` method?

Would it be useful to create a `signatureEntryData` with these two fields only?

fvm/transactionVerifier.go
```diff
@@ -14,6 +16,123 @@ import (
...
+}
+
+type signatureEntry struct {
+	flow.TransactionSignature
+
+	signatureType
+
+	accountKey flow.AccountPublicKey
+
+	verifyErr errors.CodedError
```
zhangchiqing, line 34: I'm a bit concerned about this including the verifyErr in this data structure because there are two different cases we can't distinguish where verifyErr are both `nil`:

1. we have verified and the signature is valid
2. we haven't verified yet

In other words, if we forgot to call `verify()`, then the signature would seem to be verified and is valid.

It's better to add another field: `verified`, so that we can distinguish them more explicitly: for case 1, `verified` would be `true`; for case 2, `verified` would be `false`.

fvm/transactionVerifier.go
```diff
@@ -14,6 +16,123 @@ import (
...
+
+func (entry *signatureEntry) verify() errors.CodedError {
+	valid, err := crypto.VerifySignatureFromTransaction(
+		entry.Signature,
+		entry.message,
+		entry.accountKey.PublicKey,
+		entry.accountKey.HashAlgo,
+	)
+	if err != nil {
+		entry.verifyErr = entry.newError(err)
```
zhangchiqing, line 56: this would be an exception, better to add some context: 
```
entry.verifyErr = entry.newError(fmt.Errorf("failed to verify signature: %w", err))
```

fvm/transactionVerifier.go
pattyshack: This is intentional.  The struct is used to keep track of continuation states.   accountKey and verifyErr are set by `getAccountKeys` and `verifyAccountSignatures` respectively.  These method are not accessible outside of the `verifyTransaction()`, so there's no risk of uninitialized fields.

I can add some comment to the struct and an extra nil check for clarity

fvm/transactionVerifier.go
pattyshack: see my continuation state comment

PR #3716, opened 2022-12-14
[Access] Fix public network metrics

peterargue:
Currently, the public network's metrics collector is initialized when setting up the network objects. However, starting in `v0.28`, the public network's libp2p node is created before the network objects, so it is initialized with a noop metrics collector, resulting in all public network metrics being dropped.

This PR moves the metrics initialization to before all components are started.


yhassanzadeh13 (approved):

vishalchangrani (approved):

koko1123 (approved):

peterargue:
bors merge

PR #3718, opened 2022-12-14
[BN2][Testing] Generate public-private key pair for Benchnet2

gomisha:
This PR generates a public-private key pair that is used for BN2 (Benchnet2) network deployment as well as for TPS load tests.

ref: https://github.com/dapperlabs/dapper-flow-hosting/issues/957


SaveTheRbtz (approved):
you may consider:
* json format
* put it into benchnet2/cmd

gomisha:
> you may consider:
> 
> * json format
> * put it into benchnet2/cmd

Moved to benchnet2/cmd in ea98ede6963fa734f83ce5fbdd6e9d1b98c5c22d
As for json format - that will be done at a later time when the process is more fully automated so the key gen and updating the values that need to store those keys will be automatically done.

sjonpaulbrown (approved):

gomisha:
bors merge

PR #3722, opened 2022-12-15
[Benchmark] Non-blocking SetTPS

SaveTheRbtz:
Currently, `SetTPS` is a blocking call.  Since it only blocks in cases where we decrease TPS (waiting for current transaction to finish) -- it causes tests to be biassed towards lower numbers.

### DIFF
--- a/integration/benchmark/contLoadGenerator.go
+++ b/integration/benchmark/contLoadGenerator.go
@@ -279,13 +279,10 @@ func (lg *ContLoadGenerator) startWorkers(num int) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pattyshack (commented):
just delete the wg?  there's no call where you set blocking to true ....

SaveTheRbtz:
> just delete the wg? there's no call where you set blocking to true ....

fair enough!  let's make this code simpler!

gomisha (approved):

pattyshack (approved):

SaveTheRbtz:
bors merge

SaveTheRbtz:
bors merge

PR #3723, opened 2022-12-15
[Access] Add streaming API for BlockExecutionData

peterargue:
Implement streaming gRPC APIs for `BlockExecutionData` and events.

Protobuf: https://github.com/onflow/flow/pull/1275
FLIP: https://github.com/onflow/flips/pull/73


engine/access/state_stream/handler.go
```diff
@@ -14,30 +18,107 @@ type Handler struct {
...
 	blockID, err := convert.BlockID(request.GetBlockId())
 	if err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.InvalidArgument, "could not convert block ID: %v", err)
 	}
 
 	execData, err := h.api.GetExecutionDataByBlockID(ctx, blockID)
 	if err != nil {
-		return nil, err
+		return nil, rpc.ConvertError(err, "could no get execution data", codes.Internal)
```
SaveTheRbtz, line 37: it would be nice if we could get NotFound for not yet ready blocks

engine/access/state_stream/subscription.go
```diff
@@ -0,0 +1,115 @@
...
+	id string
+
+	// ch is the channel used to pass data to the receiver
+	ch chan interface{}
+
+	// err is the error that caused the subscription to fail
+	err error
+}
+
+func NewSubscription() *SubscriptionImpl {
```
SaveTheRbtz, line 40: nit: this seems like a good usecase for generics: `func NewSubscriptionFor[T any]() *SubscriptionImpl[T]`

engine/access/state_stream/handler.go
```diff
@@ -14,30 +18,107 @@ type Handler struct {
...
+		reqFilter := request.GetFilter()
+		filter = NewEventFilter(reqFilter.GetEventType(), reqFilter.GetAddress(), reqFilter.GetContract())
+	}
+
+	sub := h.api.SubscribeEvents(stream.Context(), startBlockID, request.GetStartBlockHeight(), filter)
+
+	for {
+		v, ok := <-sub.Channel()
+		if !ok {
+			return rpc.ConvertError(sub.Err(), "stream encountered an error", codes.Internal)
```
SaveTheRbtz, line 107: IIUC this would also be called on sub.Close()

engine/access/state_stream/streamer.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if err != nil {
+			s.sub.Fail(err)
+			return
+		}
+	}
+}
+
+// sendAllAvailable reads data from the streamable and sends it to the client until no more data is available.
+func (s *Streamer) sendAllAvailable(ctx context.Context) error {
```
SaveTheRbtz, line 78: nit: this function is basically Next+Send but there is quite a bit of code in between.  Maybe it is possible to simplify this a bit by moving code around (e.g. moving some for the error-processing code inside `if err`)

engine/access/state_stream/subscription.go
```diff
@@ -0,0 +1,115 @@
...
+	return &SubscriptionImpl{
+		id: uuid.New().String(),
+		ch: make(chan interface{}),
```
SaveTheRbtz, line 43: nit: non-buffered channels are prone to deadlocks.

engine/access/state_stream/backend_executiondata.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, fmt.Errorf("could not get block header for height %d: %w", height, err)
+	}
+
+	executionData, err := b.getExecutionData(ctx, header.ID())
+	if err != nil {
+		return nil, err
+	}
+
+	return &ExecutionDataResponse{
+		Height:        header.Height,
```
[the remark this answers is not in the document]
peterargue: yea, `BlockID` is part of the core data structure. I didn't add it to  `ExecutionDataResponse` since it would be redundant.

engine/access/state_stream/engine.go
```diff
@@ -79,23 +91,49 @@ func NewEng(
 
 	server := grpc.NewServer(grpcOpts...)
 
-	backend := New(headers, seals, results, execDataStore)
+	execDataCache := herocache.NewCache(
+		DefaultCacheSize,
+		herocache.DefaultOversizeFactor,
+		heropool.LRUEjection,
```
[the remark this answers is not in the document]
peterargue: That's a good point. Unfortunately, herocache only supports LRU and random. i'll open an issue to circle back and implement LFU or something similar for this usecase.

engine/access/state_stream/filter.go
```diff
@@ -0,0 +1,76 @@
...
+
+	parts := strings.Split(string(event.Type), ".")
+
+	if len(parts) < 2 {
+		// TODO: log the error
+		return false
+	}
+
+	// name := parts[len(parts)-1]
+	contract := parts[len(parts)-2]
```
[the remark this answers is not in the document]
peterargue: this needs a comment, but there are 2 formats:
* `flow.[SomeEvent]` for protocol level events (e.g. `flow.AccountKeyAdded`)
* `A.[Address].[Contract].[SomeEvent]`

so this is saying always take the second to last element (`flow` or `[Contract]`)

engine/access/state_stream/handler.go
peterargue: it should. `GetExecutionDataByBlockID` will return a `NotFound` status code if either the block or execution data doesn't exist
https://github.com/onflow/flow-go/blob/6397de16ba4323d474ea5a3c6d5d9996ed7a2694/engine/access/state_stream/backend_executiondata.go#L40

engine/access/state_stream/handler.go
peterargue: yea, if `sub.Err()` is nil, `rpc.ConvertError()` will return nil. I'll add an explicit check since that's not obvious

engine/access/state_stream/streamer.go
peterargue: it's mostly logging code. I'll clean it up

engine/access/state_stream/subscription.go
peterargue: I can add a buffer, but this is intentionally setup without one to apply backpressure. Send uses a timeout to fail and close the connection if the client is slow
https://github.com/onflow/flow-go/blob/6397de16ba4323d474ea5a3c6d5d9996ed7a2694/engine/access/state_stream/subscription.go#L78-L88

SaveTheRbtz: i think we can balance perf and backpressure by having a tiny buffer-- that way we'll pipeline fetching with sending.  (It should be already somewhat pipelined by async nature of Send, but we may still be blocked by tcp/h2 flow control)

engine/access/state_stream/event.go
```diff
@@ -0,0 +1,56 @@
...
+func ParseEvent(eventType flow.EventType) (*ParsedEvent, error) {
+	parts := strings.Split(string(eventType), ".")
+
+	switch parts[0] {
+	case "flow":
+		if len(parts) == 2 {
+			return &ParsedEvent{
+				Type:      ProtocolEventType,
+				EventType: eventType,
+				Contract:  parts[0],
```
peterargue, line 38: Someone could define a contract named `flow` with the same event names which would break anyone filtering on the contract `flow`.

engine/access/state_stream/event.go
```diff
@@ -0,0 +1,56 @@
...
+			}, nil
+		}
+
+	case "A":
+		if len(parts) == 4 {
+			return &ParsedEvent{
+				Type:      AccountEventType,
+				EventType: eventType,
+				Address:   parts[1],
+				Contract:  parts[2],
```
peterargue, line 49: This is incorrect. Contract should be the full contract. This implementation also needs a `ContractName` field to hold just `parts[2]`
```suggestion
				Contract:  fmt.Sprintf("%s.%s.%s", parts[0], parts[1], parts[2]),
```

engine/access/state_stream/filter.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, eventType := range eventTypes {
+		f.EventTypes[flow.EventType(eventType)] = struct{}{}
+	}
+	for _, address := range addresses {
+		// convert to flow.Address to ensure it's in the correct format, but use the string value
+		// for matching to avoid an address conversion for every event
+		f.Addresses[flow.HexToAddress(address).String()] = struct{}{}
+	}
+	for _, contract := range contracts {
+		f.Contracts[contract] = struct{}{}
```
peterargue, line 37: this should have validation to ensure contracts are `A.[Address].[ContractName]` or `flow`.

engine/access/state_stream/subscription.go
peterargue: that's fair. what would you say is an appropriate buffer size? 5-10?

module/state_synchronization/requester/distributer.go
```diff
@@ -0,0 +1,52 @@
...
+	lock      sync.Mutex
+}
+
+func NewExecutionDataDistributor() *ExecutionDataDistributor {
+	return &ExecutionDataDistributor{}
+}
+
+// LastBlockID returns the block ID of the most recent execution data received
+// Execution data is guaranteed to be received in height order
+func (p *ExecutionDataDistributor) LastBlockID() flow.Identifier {
```
peterargue, line 27: this isn't needed anymore

koko1123 (approved):
Aside from the nits, I think this looks good!

engine/access/state_stream/subscription.go
SaveTheRbtz: we can start with 1 and then go up

SaveTheRbtz (approved):

peterargue:
bors merge

PR #3724, opened 2022-12-15
[Networking] Implements configurable scaler for libp2p resource manager

yhassanzadeh13:
This PR implements a configurable scaler for the libp2p resource manager that allows the node operator to config the allowed memory and file descriptors limits of libp2p at runtime.


cmd/scaffold.go
```diff
@@ -167,6 +167,8 @@ func (fnb *FlowNodeBuilder) BaseFlags() {
 	fnb.flags.StringVar(&fnb.BaseConfig.AdminClientCAs, "admin-client-certs", defaultConfig.AdminClientCAs, "admin client certs (for mutual TLS)")
 	fnb.flags.UintVar(&fnb.BaseConfig.AdminMaxMsgSize, "admin-max-response-size", defaultConfig.AdminMaxMsgSize, "admin server max response size in bytes")
 
+	fnb.flags.Float64Var(&fnb.BaseConfig.LibP2PResourceManagerConfig.FileDescriptorsRatio, "libp2p-fd-ratio", defaultConfig.LibP2PResourceManagerConfig.FileDescriptorsRatio, "ratio of available file descriptors to be used by libp2p (in (0,1])")
+	fnb.flags.Float64Var(&fnb.BaseConfig.LibP2PResourceManagerConfig.MemoryLimitRatio, "libp2p-memory-limit", defaultConfig.LibP2PResourceManagerConfig.MemoryLimitRatio, "ratio of available memory to be used by libp2p (in (0,1])")
```
peterargue, line 171: maybe we can simplify this a bit an either use autoscale (default), or allow the operator to set explicit values. The number of file descriptors and total memory are static so I don't see much value for the operator to provide a ratio.

SaveTheRbtz (approved):

network/p2p/p2pbuilder/libp2pNodeBuilder.go
```diff
@@ -97,6 +104,21 @@ type NodeBuilder interface {
 	Build() (p2p.LibP2PNode, error)
 }
 
+// ResourceManagerConfig returns the resource manager configuration for the libp2p node.
+// The resource manager is used to limit the number of open connections and streams (as well as any other resources
+// used by libp2p) for each peer.
+type ResourceManagerConfig struct {
+	MemoryLimitRatio     float64 // maximum allowed fraction of memory to be allocated by the libp2p resources (in bytes)
```
jordanschalm, line 111:
```suggestion
	MemoryLimitRatio     float64 // maximum allowed fraction of memory to be allocated by the libp2p resources in (0,1]
```
ratio is unitless

network/p2p/p2pbuilder/libp2pNodeBuilder.go
```diff
@@ -97,6 +104,21 @@ type NodeBuilder interface {
...
+	FileDescriptorsRatio float64 // maximum allowed fraction of file descriptors to be allocated by the libp2p resources
```
jordanschalm, line 112:
```suggestion
	FileDescriptorsRatio float64 // maximum allowed fraction of file descriptors to be allocated by the libp2p resources in (0,1]
```

jordanschalm (approved):

cmd/scaffold.go
yhassanzadeh13: The ratio definition is more compliant with [the libp2p's one](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/limit_defaults.go#L308-L313), i.e., querying the kernel for the available resources and scaling it based on a ratio. Also, it seems safer to define the ratio than to supply arbitrary values that is system-dependent and less portable. 

Moreover, the [libp2p AutoScale](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/limit_defaults.go#L308-L313) seems non-performant for our application use-case. We need more than `1/8` of the total memory allocated for the libp2p resources, e.g., `1/5`. Accordingly, we define our [own default](https://github.com/onflow/flow-go/blob/8e15fa85e5eb636973f7ae9e395527c73e010cda/network/p2p/p2pbuilder/libp2pNodeBuilder.go#L115-L120).

peterargue (approved):

PR #3731, opened 2022-12-16
Make computer's tracing/logging more compact/consistent

pattyshack:


engine/execution/computation/computer/computer.go
```diff
@@ -449,39 +441,33 @@ func (e *blockComputer) executeTransaction(
...
-		Str("tx_id", txID.String()).
-		Str("block_id", blockIdStr).
-		Str("traceID", traceID).
+	logger = logger.With().
 		Uint64("computation_used", tx.ComputationUsed).
 		Uint64("memory_used", tx.MemoryEstimate).
-		Uint64("memAlloc", memAllocAfter-memAllocBefore).
-		Int64("timeSpentInMS", time.Since(startedAt).Milliseconds()).
+		Uint64("mem_alloc", memAllocAfter-memAllocBefore).
+		Int64("time_spent_in_ms", time.Since(startedAt).Milliseconds()).
```
SaveTheRbtz, line 448: you may consider renaming other two instead: `timeSpentInMS` is quite common in the repo

SaveTheRbtz (approved):

engine/execution/computation/computer/computer.go
```diff
@@ -449,39 +441,33 @@ func (e *blockComputer) executeTransaction(
...
+			Logger()
+		logger.Info().Msg("transaction execution failed")
 
 		if isSystemTransaction {
 			// This log is used as the data source for an alert on grafana.
 			// The system_chunk_error field must not be changed without adding
 			// the corresponding changes in grafana.
 			// https://github.com/dapperlabs/flow-internal/issues/1546
-			e.log.Error().
-				Str("error_message", errMsg).
```
janezpodhostnik, line 476: I wonder why we didnt just use `.Err(tx.Err)` here

janezpodhostnik (approved):

engine/execution/computation/computer/computer.go
pattyshack: i have no idea, but the comment looks scary so i'm leaving this as is

engine/execution/computation/computer/computer.go
pattyshack: all other attributes (block_id, tx_index, etc.) are logged using snake case.  it's better to be consistent within the same log line rather consistently inconsistent

pattyshack:
bors merge

PR #3732, opened 2022-12-16
Add canary missing keys

janezpodhostnik:
Quick migration to add keys to the core contract accounts that are missing one. 

Which is only a problem on canary.

The cause the canary accounts are missing keys was an old bug, that was fixed with https://github.com/onflow/flow-go/pull/2705

There is no need to merge this, we only need it to fix the state.

### DIFF
--- a/cmd/util/ledger/migrations/add_missing_keys_migration.go
+++ b/cmd/util/ledger/migrations/add_missing_keys_migration.go
@@ -1,0 +1,76 @@
+package migrations
+
+import (
+	"fmt"
+	"github.com/onflow/flow-go/fvm/environment"
+	"github.com/onflow/flow-go/fvm/state"
+
+	"github.com/onflow/flow-go/ledger"
+	"github.com/onflow/flow-go/model/flow"
+)
+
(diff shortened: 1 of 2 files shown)


cmd/util/ledger/migrations/add_missing_keys_migration.go
```diff
@@ -0,0 +1,76 @@
...
+	address := flow.HexToAddress(addressInHex)
+	ok, err := accounts.Exists(address)
+	if err != nil {
+		return err
+	}
+	if ok {
+		err = accounts.AppendPublicKey(address, accountKey)
+		if err != nil {
+			return err
+		}
```
ramtinms, line 70: you might need to adjust the storage used by this account as well

janezpodhostnik: Good point. I checked and AppendPublicKey goes through SetValue which updates storage used, so this should be taken care of.

SaveTheRbtz (approved):

janezpodhostnik:
This was applied. Closing

PR #3632, opened 2022-11-29
[Engine] Require explicit capacity for Fifo queue

peterargue:
Previously, the `FifoQueue`'s constructor took an optional argument that specified the queue's capacity, and defaulted to `maxint`. While all instantiations did specify a capacity, this pattern makes it easy to unintentionally omit the capacity override and configure an unlimited queue.

Update the `NewFifoQueue` constructor to take a required `maxCapacity` argument.

PR #3734, opened 2022-12-17
handle limited size of queue for holding incorporated block in sealing logic

AlexHentschel:
PR https://github.com/onflow/flow-go/pull/3632/files caused me to look into the queue limitation for the sealing logic. I concluded that the added capacity limit for the [incorporated result queue](https://github.com/onflow/flow-go/blob/b8d41adf7c138b49686c9e667dd7de8db59c9056/engine/consensus/sealing/engine.go#L46-L47) is fine, because we have [this logic](https://github.com/onflow/flow-go/blob/b8d41adf7c138b49686c9e667dd7de8db59c9056/engine/consensus/sealing/engine.go#L479-L484) that crashes the node to avoid dropping incorporated results. However, a similar logic for the queue for incorporated blocks was absent (-> [code](https://github.com/onflow/flow-go/blob/b8d41adf7c138b49686c9e667dd7de8db59c9056/engine/consensus/sealing/engine.go#L441)) 😨 . The latter could have caused a sealing halt in extreme edge cases. 

Going back to this logic took really long, because there was hardly any documentation in the code base and out [corresponding notion doc](https://www.notion.so/dapperlabs/Matching-Engine-and-Sealing-Engine-b20868df572a4653bb915800552b970a) was outdated and did not cover the recovery case. 

### This PR adds:

• logic to avoid accidentally dropping an incorporated block, which could undermine sealing liveness.
• Added lots of documentation, to avoid extensive time for reviewing similar aspects in the future.

### DIFF
--- a/engine/consensus/Sealing_Readme.md
+++ b/engine/consensus/Sealing_Readme.md
@@ -1,0 +1,89 @@
+# Matching Engine and Sealing Engine
+
+### Matching Engine
+
+- Matching engine ingests the Execution Receipts (ERs) from execution nodes and incorporated blocks, validates the ERs and stores then in its `ExecutionTree`.
+- Matching core is fully concurrent
+- if receipts are received before the block, we will drop the receipt
+- If receipts are received after the block, but the previous result is missing, we will cache the receipt in the mempool
+- If previous result exists, we can validate the receipt, and store it in the execution tree mempool and storage.
+- After processing a receipt, we try to find child receipts from the cache, and process them.
+- When constructing a new block, the block builder reads receipts and results from the execution tree and add them to the new block’s payload
+- processing finalized block is to prune pending receipts and execution tree mempool.
+- Matching core also contains logic for fetching missing receipts (method `requestPendingReceipts`)
+
+  Caution: this logic is not yet fully BFT compliant. The logic only requests receipts for blocks, where it has less than 2 consistent receipts. This is a temporary simplification and incompatible with the mature BFT protocol! There might be multiple consistent receipts that commit to a wrong result. To guarantee sealing liveness, we need to fetch receipts from *all*  ENs, whose receipts we don't have yet. There might be only a single honest EN and its result might occasionally get lost during transmission.
+
+- **Crash recovery:**
+    - crash recovery is implemented in the consensus block builder (`module/builder/consensus/builder.go` method `repopulateExecutionTree`)
+    - During normal operations, we query the tree for "all receipts, whose results are derived from the latest sealed and finalized result". This requires the execution tree to know what the latest sealed and finalized result is.
+
+      The builder adds the result for the latest block that is both sealed and finalized, without any Execution Receipts. This is sufficient to create a vertex in the tree. Thereby, we can traverse the tree, starting from the sealed and finalized result, to find derived results and their respective receipts.
+
+    - In a second step, the block builder repopulates the builder adds all known receipts for any finalized but unsealed block to the execution tree.
+    - Lastly, the builder adds receipts from any valid block that descends from the latest finalized block.
+    - Note:
+
+      It is *strictly required* to add the result for the latest block that is both sealed and finalized. This is necessary because the ID of the latest result is given as a query to find all descending results during block construction. Without the sealed result, the block builder could not proceed.
+
+      In contrast, conceptually it is *not strictly required* for the builder to also add the receipts from the known blocks (finalized unsealed blocks and all pending blocks descending from the latest finalized block). This is because this matching engine *would* re-request these (method `requestPendingReceipts`) by itself. Nevertheless, relying on re-feting already known receipts would add a large latency overhead after recovering from a crash.
+
+
+### Sealing Engine
+
+- The sealing engine ingests:
+    - Result approvals from the Verification nodes (message `ResultApproval`)
+    - It also re-requests missing result approvals and ingests the corresponding responses (`ApprovalResponse`)
+    - Execution Results from incorporated blocks. Sealing engine can trust those incorporated results are valid, because the protocol state has validated them.
+
+      Caution: to not compromise sealing liveness, the sealing engine cannot drop any incorporated blocks or any results contained in it. Otherwise, it cannot successfully create a seal for the incorporated results.
+
+- Sealing Engine will forward each receipts and approval to the sealing core. Receipts and approvals are concurrently processed by sealing core.
+- The purpose for the Sealing Core is to
+    1. compute verifier assignments for the incorporated results
+    2. track the approvals that match the assignments
+    3. generate a seal once enough approvals have been collected
+- For sealing core to concurrently process receipts, it maintains an `AssignmentCollectorTree`, which is a fork-aware structure holding the assignments + approvals for each incorporated result. Therefore, it can ingest results and approvals in arbitrary order.
+    - Each vertex (`AssignmentCollector`) in the tree is responsible for collecting approvals one particular incorporated result. Each collector is able to process approvals concurrently. But accessing the tree requires a lock.
+    - A result might be incorporated in multiple blocks on different block forks. Results incorporated on different forks will lead to different assignment for verification nodes. Although the approvals are the same for incorporated results on different fork, a valid approval for a certain incorporated result on a fork might be invalid for another incorporated result in a different fork, because the assignment is different and a verification node is not allowed to verify a result if the result is not assigned to it to verify.
+- The `AssignmentCollectorTree` tracks whether the result is derived from the latest result with a finalized seal. If interim results are unknown (due to concurrent ingestion of incorporated results), approvals are only cached (status `CachingApprovals`). Once all interim results are received, the `AssignmentCollectorTree` changes the status of the result and all its *connected* children to be processable (status `VerifyingApprovals`).
+
+  Furthermore, if results are orphaned because they conflict with a result that has a finalized seal, the processing of all approvals for the conflicting result and any of its derived results is stopped  (status `Orphaned`)
+
+- In order to limit the memory usage of the `AssignmentCollectorTree`, we need to prune it by sealed height. Therefore, we subscribe to finalized block event and use the sealed height to prune the tree.
+
+  When pruning the tree, we keep the node at the same height as the last sealed height, so that we know which fork actually connects to the sealed block
+
+- It means adding a new result to the tree requires a lock, and is a write lock. A write lock won't block reads, so that the worker to process approvals won't be blocked.
+- when a result has been added to a tree, it also means we've received the executed block, because otherwise the block, which includes the receipt, won't pass the validation
+- We create an approval collector for each incorporated result, when we receive an approval.  We verify the approval only once, if valid, we forward it to multiple approval collectors for each result incorporated in different fork. If any fork has collected enough approval, we will generate a seal, note a seal can only be used on one fork. If a consensus leader is building a new block on a fork, it will only add new seals for that fork to the new block.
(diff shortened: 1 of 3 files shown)


AlexHentschel:
Note: I also updated our notion doc https://www.notion.so/dapperlabs/Matching-Engine-and-Sealing-Engine-b20868df572a4653bb915800552b970a (and copied the updated content into [`engine/consensus/Sealing_Readme.md`](https://github.com/onflow/flow-go/blob/bb977286851a14e0586511d8a41f826eaa318d4d/engine/consensus/Sealing_Readme.md))

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,89 @@
...
+### Matching Engine
+
+- Matching engine ingests the Execution Receipts (ERs) from execution nodes and incorporated blocks, validates the ERs and stores then in its `ExecutionTree`.
+- Matching core is fully concurrent
+- if receipts are received before the block, we will drop the receipt
+- If receipts are received after the block, but the previous result is missing, we will cache the receipt in the mempool
+- If previous result exists, we can validate the receipt, and store it in the execution tree mempool and storage.
+- After processing a receipt, we try to find child receipts from the cache, and process them.
+- When constructing a new block, the block builder reads receipts and results from the execution tree and add them to the new block’s payload
+- processing finalized block is to prune pending receipts and execution tree mempool.
```
durkmurder, line 12:
```suggestion
- Processing finalized block is to prune pending receipts and execution tree mempool.
```

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,89 @@
...
+- Matching core also contains logic for fetching missing receipts (method `requestPendingReceipts`)
+
+  Caution: this logic is not yet fully BFT compliant. The logic only requests receipts for blocks, where it has less than 2 consistent receipts. This is a temporary simplification and incompatible with the mature BFT protocol! There might be multiple consistent receipts that commit to a wrong result. To guarantee sealing liveness, we need to fetch receipts from *all*  ENs, whose receipts we don't have yet. There might be only a single honest EN and its result might occasionally get lost during transmission.
+
+- **Crash recovery:**
+    - crash recovery is implemented in the consensus block builder (`module/builder/consensus/builder.go` method `repopulateExecutionTree`)
```
durkmurder, line 18:
```suggestion
    - Crash recovery is implemented in the consensus block builder (`module/builder/consensus/builder.go` method `repopulateExecutionTree`)
```

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,89 @@
...
+
+  Furthermore, if results are orphaned because they conflict with a result that has a finalized seal, the processing of all approvals for the conflicting result and any of its derived results is stopped  (status `Orphaned`)
+
+- In order to limit the memory usage of the `AssignmentCollectorTree`, we need to prune it by sealed height. Therefore, we subscribe to finalized block event and use the sealed height to prune the tree.
+
+  When pruning the tree, we keep the node at the same height as the last sealed height, so that we know which fork actually connects to the sealed block
+
+- It means adding a new result to the tree requires a lock, and is a write lock. A write lock won't block reads, so that the worker to process approvals won't be blocked.
+- when a result has been added to a tree, it also means we've received the executed block, because otherwise the block, which includes the receipt, won't pass the validation
```
durkmurder, line 58:
```suggestion
- When a result has been added to a tree, it also means we've received the executed block, because otherwise the block, which includes the receipt, won't pass the validation
```

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,89 @@
...
+
+  Furthermore, if results are orphaned because they conflict with a result that has a finalized seal, the processing of all approvals for the conflicting result and any of its derived results is stopped  (status `Orphaned`)
+
+- In order to limit the memory usage of the `AssignmentCollectorTree`, we need to prune it by sealed height. Therefore, we subscribe to finalized block event and use the sealed height to prune the tree.
+
+  When pruning the tree, we keep the node at the same height as the last sealed height, so that we know which fork actually connects to the sealed block
+
+- It means adding a new result to the tree requires a lock, and is a write lock. A write lock won't block reads, so that the worker to process approvals won't be blocked.
```
durkmurder, line 57: I think this is not accurate, when a goroutine obtains write lock it will block readers all readers until released. It's the opposite, read lock won't block other reads.

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,89 @@
...
+- when a result has been added to a tree, it also means we've received the executed block, because otherwise the block, which includes the receipt, won't pass the validation
+- We create an approval collector for each incorporated result, when we receive an approval.  We verify the approval only once, if valid, we forward it to multiple approval collectors for each result incorporated in different fork. If any fork has collected enough approval, we will generate a seal, note a seal can only be used on one fork. If a consensus leader is building a new block on a fork, it will only add new seals for that fork to the new block.
+- When adding a new result to the tree, we use a Write lock on the three to add the `AssignmentCollector`. However, when receiving an `AssignmentCollector` from the tree to add an approval, we use a read lock, so that processing approvals concurrently won’t be blocked by processing receipts.
```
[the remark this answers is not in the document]
AlexHentschel: fixed in https://github.com/onflow/flow-go/pull/3734/commits/ddb7402ed0699ea6abc6528ddfba09679f3089d5

AlexHentschel:
Thanks for noticing the wrong statements @durkmurder. I have taken another pass over the `Sealing_Readme.me`. Looking over it again today with a fresh mind, I also found some other slightly misleading statements and missing important details. 

I re-worked the `Sealing Engine` section. Would you mind reading that section again, please? Thanks

engine/consensus/sealing/engine.go
```diff
@@ -438,12 +446,18 @@ func (e *Engine) OnFinalizedBlock(*model.Block) {
 // CAUTION: the input to this callback is treated as trusted; precautions should be taken that messages
 // from external nodes cannot be considered as inputs to this function
 func (e *Engine) OnBlockIncorporated(incorporatedBlock *model.Block) {
-	e.pendingIncorporatedBlocks.Push(incorporatedBlock.BlockID)
+	added := e.pendingIncorporatedBlocks.Push(incorporatedBlock.BlockID)
```
durkmurder, line 449: One of the options to consider is using a buffered channel, this will allow us to avoid dropping but at cost of possibly blocking. I bet you have considered this option, would like to hear your thoughts on this.

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,109 @@
+# Matching Engine and Sealing Engine
+
+### Matching Engine
+
+- Matching engine ingests the Execution Receipts (ERs) from execution nodes and incorporated blocks, validates the ERs and stores then in its `ExecutionTree`.
+- Matching core is fully concurrent
+- if receipts are received before the block, we will drop the receipt
```
durkmurder, line 7:
```suggestion
- If receipts are received before the block, we will drop the receipt
```

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,109 @@
+# Matching Engine and Sealing Engine
+
+### Matching Engine
+
+- Matching engine ingests the Execution Receipts (ERs) from execution nodes and incorporated blocks, validates the ERs and stores then in its `ExecutionTree`.
+- Matching core is fully concurrent
```
durkmurder, line 6:
```suggestion
- Matching core is fully concurrent.
```

durkmurder: also some other dots are missing

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,109 @@
...
+
+    - In a second step, the block builder repopulates the builder adds all known receipts for any finalized but unsealed block to the execution tree.
+    - Lastly, the builder adds receipts from any valid block that descends from the latest finalized block.
+    - Note:
+
+      It is *strictly required* to add the result for the latest block that is both sealed and finalized. This is necessary because the ID of the latest result is given as a query to find all descending results during block construction. Without the sealed result, the block builder could not proceed.
+
+      In contrast, conceptually it is *not strictly required* for the builder to also add the receipts from the known blocks (finalized unsealed blocks and all pending blocks descending from the latest finalized block). This is because this matching engine *would* re-request these (method `requestPendingReceipts`) by itself. Nevertheless, relying on re-feting already known receipts would add a large latency overhead after recovering from a crash.
```
durkmurder, line 29:
```suggestion
      In contrast, conceptually it is *not strictly required* for the builder to also add the receipts from the known blocks (finalized unsealed blocks and all pending blocks descending from the latest finalized block). This is because this matching engine *would* re-request these (method `requestPendingReceipts`) by itself. Nevertheless, relying on re-fetching already known receipts would add a large latency overhead after recovering from a crash.
```

engine/consensus/Sealing_Readme.md
```diff
@@ -0,0 +1,109 @@
...
+- For sealing core to concurrently process receipts, it maintains an `AssignmentCollectorTree`, which is a fork-aware structure holding the assignments + approvals for each incorporated result. Therefore, it can ingest results and approvals in arbitrary order.
+    - Each vertex (`AssignmentCollector`) in the tree is responsible for managing approvals for one particular result.
+    - A result might be incorporated in multiple blocks on different block forks. 
+    - Results incorporated on different forks will lead to different assignment for verification nodes. For example, consider block `B` with two children `C` and `D`:
+      ```
+            ┌ C   
+         B <   
+            └ D
+      ```
+      `C` and `D` can each incorporate the same result `r` for block `B`, but the verifier assignment would be different. Hence, there is one `AssignmentCollector` for `r` stored in the `AssignmentCollectorTree`.
```
durkmurder, line 55: Can we show on diagram that `C` and `D` incorporates `ER` for `B`?

durkmurder (approved):
Very nice, thank you for adding a readme, it will be very useful for future maintenance and on-boarding new engineers.

engine/consensus/sealing/engine.go
AlexHentschel: Completely agree, a buffered channel is a totally fine approach too. My reasons are:
* [aesthetically] we have multiple queues in the sealing engine, and therefore using also a queue here is the most consistent
* The queue is fully compliant with the API convention to _never block_:  https://github.com/onflow/flow-go/blob/f2555090b6f672036d9368f7890b8e355ccb9736/consensus/hotstuff/consumer.go#L19-L31
   * if we exceed the queue capacity, the node will go down with an error, which we will see in the logs
   * In comparison,  if we exceed the channel capacity, we block the main consensus logic. This will look like the node just hangs, without anything in the logs. I don't even want to start thinking about how painful it would be to find the root cause. I am a big fan of "fail early and fail hard (loud)" as opposed to working around an edge case that we agree should not happen in practise. 
* Lastly, I think a buffered channel allocates the entire memory upfront. Therefore, we would always pay the price for the full queue. But in practise, this queue will be empty with overwhelming probability because the worker drains it immediately, maybe it has rarely one element in it, almost never two ... Don't really think it is worth-while paying the cost for 10k elements. But we need a large capacity to avoid crashing the node too early or blocking other components of the node. 

🤷

engine/consensus/Sealing_Readme.md
AlexHentschel: Thanks for this comment. I realized that there is quite a bit of information missing, to properly explain what is going on here. I am worried that adding additional context here to this figure, and using notation that I have to introduce might be too confusing. 

* I added two additional sections (exported from our notion documents) to explain more of the basics upfront ... also including a proper figure that illustrates this example a bit more in detail. 
* Instead of adding more complexity here, I could then simply just reference the example from the prior section.

jordanschalm (approved):
Thank you, the documentation improvements are great

AlexHentschel:
bors merge

PR #3633, opened 2022-11-30
[Benchmark] Change BigQuery schema

SaveTheRbtz:
This diff changes the BigQuery schema for the `benchmark/cmd/ci` binary so that we can have both long-term perf analysis and adhoc perf tests.

Sadly, this diff is quite enormous and lacks test coverage.  On the bright side it does move core logic out of `main.go` and compartmentalizes the codebase.

[Long-term perf analysis example](https://dapperlabs.grafana.net/d/BLSHi8ZVz/flow-performance-benchmarks?orgId=1&from=now-90d&to=now&var-branch=All&var-env=All&var-type=All):
long term

[Per-test in-depth analysis](https://dapperlabs.grafana.net/d/mA42tHK4z/flow-performance-benchmarks-raw?orgId=1&from=now-24h&to=now&var-branch=All&var-env=All&var-type=All):
short term

Filters for branch, environment, and test type are available:
Filters

Also all tests have annotations to point to a specific commit tested and its branch/tag if available:
Annotations

ref: #3548

PR #3735, opened 2022-12-17
[Benchmark] Switch to a PD controller

SaveTheRbtz:
This switches the code from a home-grown TCP's AIMD algorithm with a PD-part of a PID controller.  Controller tries to maintain TPS to the value that keeps number of inflight requests equal to the half of available accounts.

This PR does not use integral component of a PID controller to avoid all the problems with a windup, saturation, and clamping. (And since the inflight requests is already an integral of TPS -- we do not need another one.)
We also reduce the control period to 10s to shorten feedback loop's delay.

Old AIMD-based:
Screen Shot 2022-12-17 at 11 51 14 AM

New PD-based:
Screen Shot 2022-12-17 at 11 51 24 AM

Did a single round of a simple Cohen-Coon Tuning for PD controllers to make it converge and reduce oscillations:
Screen Shot 2022-12-17 at 3 11 09 PM

> This is a followup for the @pattyshack's comment in: https://github.com/onflow/flow-go/pull/3633#discussion_r1047528729.  Instead of increasing complexity of AIMD further, replaced it with a PID controller.


pattyshack (approved):

gomisha (approved):

SaveTheRbtz:
bors merge

Issue #3537, opened 2022-11-09
Support VersionUpdate service event


m4ksio:
As other service events, they must be hardcoded and explicitly supported by both FVM and flow-go in general.
This task is encompassing the work required for exactly that

PR #3736, opened 2022-12-18
#3537 Support version beacon service event

m4ksio:


storage/badger/operation/prefix.go
```diff
@@ -137,3 +139,15 @@ func b(v interface{}) []byte {
...
 	}
 }
+
+// d is a reverse of b
+func d[T uint64](bytes []byte) (ret T) {
+
+	// See https://stackoverflow.com/a/73382479
+	switch any(ret).(type) {
+	case uint64:
+		return T(binary.BigEndian.Uint64(bytes))
```
jordanschalm, line 149:
```suggestion
		return T(binary.BigEndian.Uint64(bytes))
	default:
		panic(...)
```
To avoid hiding failures from unimplemented types, which currently would return the empty value of type T.

storage/version_beacon.go
```diff
@@ -0,0 +1,11 @@
+package storage
+
+import "github.com/onflow/flow-go/model/flow"
+
+// VersionBeacons represents persistent storage for Version Beacons.
+type VersionBeacons interface {
+
+	// Highest finds the highest flow.VersionBeacon but no higher than maxHeight and the corresponding height it has been finalized.
+	// Returns storage.ErrNotFound if version beacon exists at or below the given height.
```
jordanschalm, line 9:
```suggestion
	// Returns storage.ErrNotFound if no version beacon exists at or below the given height.
```

utils/unittest/fixtures.go
```diff
@@ -1872,6 +1876,41 @@ func EpochCommitFixture(opts ...func(*flow.EpochCommit)) *flow.EpochCommit {
 	return commit
 }
 
+func VersionBeaconFixture(size uint, opts ...func(beacon flow.VersionBeacon)) *flow.VersionBeacon {
```
jordanschalm, line 1879:
```suggestion
func VersionBeaconFixture(size uint, opts ...func(beacon *flow.VersionBeacon)) *flow.VersionBeacon {
```
Need to operate on the reference, not a copy, otherwise options generally aren't being applied.

fvm/environment/event_emitter.go
```diff
@@ -274,5 +276,11 @@ func IsServiceEvent(event cadence.Event, chain flow.ChainID) (bool, error) {
 		}
 	}
 
+	fmt.Printf("SERVICE EVENt DEBUG\n")
+	spew.Dump(chain)
+	spew.Dump(event)
+	spew.Dump(events.All())
+
+	fmt.Printf("/SERVICE EVENT DEBUG\n")
```
jordanschalm, line 284:
```suggestion
```

state/protocol/badger/state.go
```diff
@@ -155,6 +163,12 @@ func Bootstrap(
 func (state *State) bootstrapSealingSegment(segment *flow.SealingSegment, head *flow.Block) func(tx *transaction.Tx) error {
 	return func(tx *transaction.Tx) error {
 
+		blocksById := make(map[flow.Identifier]*flow.Block, len(segment.Blocks))
```
jordanschalm, line 166: I think we can roll back all the changes to `bootstrapSealingSegment`. 

The additions to `bootstrapSealingSegment` here are to process any `VersionBeacon` service events which appear in the sealing segment. However, as the code mentions, it is quite unlikely that the most recent `VersionBeacon` service event actually happens to appear in the sealing segment. 

Therefore, we need to bootstrap the `VersionBeacon` information separately in the general case, which we do in `boostrapVersionBeacon`. 

Is there any case where `boostrapVersionBeacon` is insufficient on its own? I don't think so, here is my reasoning:

```
A <- B <- C <- D
```

We always initialize an initial `VersionBeacon` value, based on the value for the root block of the snapshot (the highest block: `D`). Therefore, we don't need to worry about separately processing service events which are applied as of `D`. 

Let's suppose there is a service event `S` which is not yet applied as of block `D`. That means that `S` will be sealed either by `D`, or a block later than `D` (after the sealing segment). 

If `S` is sealed by `D`, the `S` will be applied when finalizing `D`'s child. This will be handled by the `Mutator` and does not need a special case in bootstrapping. The same reasoning applies in case `S` is sealed by a block later than `D`.

jordanschalm (commented):
Nice work on this. I haven't completed reviewing all changes in detail, but have added comments to the main high-level changes. 

### Comments
* My main change request is removing the specific service event processing from bootstrapping - I think this is redundant with the addition of `boostrapVersionBeacon`
* Not strictly in scope here, but I'm not sure how the initial state of `VersionBeacon` looks like, for example after a spork. Is it valid to have no version specified initially?
* It looks like this PR is extending https://github.com/onflow/flow-go/pull/3605, which has been approved. As a reviewer, it is a lot easier to differentiate between the changes introduced in this PR, and the changes which have already been reviewed+approved, if the new PR (this one) is based on the previous branch, rather than master. Then I can review only the new changes, and the incremental diff is smaller and easier to work with.

storage/badger/operation/prefix.go
m4ksio: Hmm, this uses generics which is a fixed list of types, I don't think unsupported type will compile

model/flow/version.go
```diff
@@ -0,0 +1,55 @@
...
+	Height  uint64
+	Version string
+}
+
+// VersionBeacon represents a service event which specified required software versions for upcoming blocks.
+// It contains RequiredVersions field which is an ordered list of VersionControlRequirement.
+// RequiredVersions is an ordered table of VersionControlRequirement. It informs about upcoming required versions by
+// height. Table is ordered by height, ascending. While heights are strictly increasing, versions are equal or greater,
+// compared by semver semantics. First entry, if present, must indicate a current version - that is its height must be
+// at or below a block is has been emitted in.
```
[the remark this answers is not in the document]
m4ksio: Yes. The version is kept inside a contract, so especially for sporking requirement would mean that we need to update on-chain data, with no obvious benefits. Also many networks don't care about the version at all (localnet, emulator)

m4ksio:
> Nice work on this. I haven't completed reviewing all changes in detail, but have added comments to the main high-level changes.
> 
> ### Comments
> * My main change request is removing the specific service event processing from bootstrapping - I think this is redundant with the addition of `boostrapVersionBeacon`

Just to confirm - when we take a state snapshot by block - is this a a last or first block in subsequent sealing segment?

> * Not strictly in scope here, but I'm not sure how the initial state of `VersionBeacon` looks like, for example after a spork. Is it valid to have no version specified initially?

No version beacon is a valid state. That's because not all network needs version management, and version state is stored in contract which would make it hard to synchronize. We may revisit later, but for now we simply tolerate this

> * It looks like this PR is extending #3605, which has been approved. As a reviewer, it is a lot easier to differentiate between the changes introduced in this PR, and the changes which have already been reviewed+approved, if the new PR (this one) is based on the previous branch, rather than master. Then I can review only the new changes, and the incremental diff is smaller and easier to work with.

It looks similar but in fact its really rather different, and not many things are re-used apart from concepts.

storage/badger/operation/prefix.go
jordanschalm: Actually I think the problem is different than what I initially commented. We're type switching on the type of the return value, which is always `uint64` - not the input type, since the input is always `[]byte`. 

⚠️  So types for which decoding is not implemented would not pass through the switch statement and silently fail, they would be incorrectly interpreted as `uint64` and the function would return an incorrect value. 

For example, we can try passing in the unsupported type `Identifier` (https://github.com/onflow/flow-go/commit/6d33da2a1fb74f44bed7eaf35caf3748cdc3a538):
```go
id := unittest.IdentifierFixture()
fmt.Println(id)
enc := b(id)
dec := d(enc)
fmt.Println(dec)
```


**Suggestion:** Modify function `d` to either correctly handle all types supported by `b`, or explicitly fail for unsupported types. (It would also be great to add some documentation to these functions and/or give them more descriptive names.)

state/errors.go
```diff
@@ -142,3 +142,5 @@ func IsUnknownBlockError(err error) bool {
 	var e UnknownBlockError
 	return errors.As(err, &e)
 }
+
+var ErrNoVersionBeacon = errors.New("no version beacon exists")
```
janezpodhostnik, line 146: It might be nice to include the block height for which a version beacon was requested. In case there are version beacons, just not "low" enough.

storage/badger/version_beacon.go
```diff
@@ -0,0 +1,41 @@
...
+	res := &VersionBeacons{
+		db: db,
+	}
+
+	return res
+}
+
+// Highest finds the highest flow.VersionBeacon but no higher than maxHeight.
+// Returns storage.ErrNotFound if version beacon exists at or below the given height.
+func (r *VersionBeacons) Highest(maxHeight uint64) (*flow.VersionBeacon, uint64, error) {
```
janezpodhostnik, line 29: In most other places the wording is: "latest" version beacon, not "highest". Might be nice to also have that here.

storage/badger/operation/prefix.go
m4ksio: We do not pass unsupported type - this function always accepts `[]byte` - and its the raw encoding, without the type information. Type information is inferred from the return type parameter. In your case there is no other type possible than `uint64` - and thats how `d` interprets the bytes, and thats the type of `dec`.
 I have added another supported type `byte` to `d` - so now when you use it code won't compile without explicitly providing the type which I hope will make it more usable.

state/protocol/badger/state.go
m4ksio: https://github.com/onflow/flow-go/pull/3736#issuecomment-1360449088 - especially the first question. If that's correct I'm happy to remove changes in `bootstrapSealingSegment` - just want to be extra sure

model/flow/service_event.go
```diff
@@ -48,6 +49,16 @@ func (sel ServiceEventList) EqualTo(other ServiceEventList) (bool, error) {
 	return true, nil
 }
 
+func FilterServiceEvents[T *EpochSetup | *EpochCommit | *VersionBeacon](sel ServiceEventList) []T {
```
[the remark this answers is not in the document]
m4ksio: Yes - but we keep the list of service events fixed and rarely changing. Additionally, and this is the main reason, this method already returns properly typed slice, otherwise caller would have to rewrite it to get a proper type

state/errors.go
m4ksio: This error is intended for when snapshot doesn't have the version beacon available, so there is no given height per se.

storage/badger/version_beacon.go
m4ksio: Yes, but this is highest but not higher than. Latests doesn't have the same meaning in my opinion

integration/testnet/network.go
```diff
@@ -24,7 +26,6 @@ import (
 	"github.com/dapperlabs/testingdock"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/api/types/network"
```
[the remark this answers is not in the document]
m4ksio: Good point, I wonder why linter didn't flag this.

integration/testnet/client.go
```diff
@@ -218,7 +219,35 @@ func (c *Client) WaitForSealed(ctx context.Context, id sdk.Identifier) (*sdk.Tra
...
+	return result, err
+}
+
+func (c *Client) WaitForAtLeastStatus(ctx context.Context, txID sdk.Identifier, status sdk.TransactionStatus) (*sdk.TransactionResult, error) {
+
+	fmt.Printf("Waiting for transaction %s to be sealed...\n", txID)
+	errCount := 0
+	var result *sdk.TransactionResult
+	var err error
+	for result == nil || (result.Status < status) {
```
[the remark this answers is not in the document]
m4ksio: Good spot. But turns out we don't use this function :)

jordanschalm:
> Just to confirm - when we take a state snapshot by block - is this a a last or first block in subsequent sealing segment?

It is the highest block in the subsequent sealing segment (see [implementation](https://github.com/onflow/flow-go/blob/60cc312acee146f49ebb62bb9a9314908689be95/state/protocol/badger/snapshot.go#L269-L290) and [godoc](https://github.com/onflow/flow-go/blob/60cc312acee146f49ebb62bb9a9314908689be95/state/protocol/snapshot.go#L72-L75) for reference)

storage/badger/operation/prefix.go
jordanschalm: Adding more than one type parameter option works to fix this - sounds good 👍

state/protocol/badger/validity.go
```diff
@@ -338,6 +344,28 @@ func validateRootQC(snap protocol.Snapshot) error {
...
+		}
+	}
+
+	segment, err := snap.SealingSegment()
+	if err != nil {
+		return fmt.Errorf("could not get sealing segment: %w", err)
+	}
+	highest := segment.Highest()
+
+	if versionTableHeight > highest.Header.Height {
```
zhangchiqing, line 362:
```suggestion
	// version beacon must be included in a past block to be effective
	if versionTableHeight > highest.Header.Height {
```

fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc
```diff
@@ -1,9 +1,14 @@
 import FlowEpoch from 0xEPOCHADDRESS
+import NodeVersionBeacon from 0xNODEVERSIONBEACONADDRESS
 
 transaction {
   prepare(serviceAccount: AuthAccount) {
 	let heartbeat = serviceAccount.borrow<&FlowEpoch.Heartbeat>(from: FlowEpoch.heartbeatStoragePath)
       ?? panic("Could not borrow heartbeat from storage path")
     heartbeat.advanceBlock()
+
+    let versionBeacon = serviceAccount.borrow<&NodeVersionBeacon.NodeVersionAdmin>(from: NodeVersionBeacon.NodeVersionAdminStoragePath)
```
zhangchiqing, line 10: The code indentation looks different than the `heartbeat` above, can we format the code properly?

storage/badger/operation/version_beacon_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/onflow/flow-go/storage"
+	"github.com/onflow/flow-go/utils/unittest"
+)
+
+func TestResults_IndexByServiceEvents(t *testing.T) {
+	unittest.RunWithBadgerDB(t, func(db *badger.DB) {
+
+		vb1 := unittest.VersionBeaconFixture(2)
+		vb2 := unittest.VersionBeaconFixture(2)
+		vb3 := unittest.VersionBeaconFixture(2)
```
zhangchiqing, line 19: Let's make the three version beacon to be different, otherwise checking equal to one of them would always return true.
```suggestion
		vb1 := unittest.VersionBeaconFixture(1)
		vb2 := unittest.VersionBeaconFixture(2)
		vb3 := unittest.VersionBeaconFixture(3)
```

fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc
m4ksio: Not sure what you mean by "properly", we break line in the same place (before `??`) and indent in a same manner?

model/flow/version.go
```diff
@@ -0,0 +1,55 @@
...
+// Sequence is event sequence number, which can be used to verify that no event has been skipped by the follower
```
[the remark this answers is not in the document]
m4ksio: Not in this PR, we only need to validate this when in situations when consensus follower is far ahead to check if the we didn't skip a version

state/protocol/badger/mutator.go
```diff
@@ -659,6 +673,13 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
 			}
 		}
 
+		if versionBeacon != nil {
+			err := operation.IndexVersionBeaconByHeight(versionBeacon, header.Height)(tx)
```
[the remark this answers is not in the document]
m4ksio: Good point. This is part of the smart contract (it it's not supposed to emit event with version behind the current block) but it won't hurt to add the validation here.

storage/badger/operation/version_beacon_test.go
m4ksio: We use random versions for actual requirement table, this argument is just a size of the table

zhangchiqing: I see. The randomness is quite deep in the `VersionControlRequirementFixture`. 

I think it's still good to use `1,2,3`, as the real scenario would most likely have the increasing number of VersionControlRequirement.

fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc
zhangchiqing: I'm a bit nit here.
I meant the `	let heartbeat = ` used a tab as indentation, but the `    let versionBeacon = ` used 4 spaces as indentation. 
From github, it looks like the code is not formatted consistently:
image

state/protocol/badger/mutator.go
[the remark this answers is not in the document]
zhangchiqing: I think we should add this validation to the sealing segment to be BFT

state/protocol/badger/mutator.go
```diff
@@ -556,17 +556,21 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
 
 		result, err := m.results.ByID(seal.ResultID)
 		if err != nil {
 			return fmt.Errorf("could not retrieve result (id=%x) for seal (id=%x): %w", seal.ResultID, seal.ID(), err)
 		}
 		for _, event := range result.ServiceEvents {
+			// skip updating epoch-related metrics if EECC is triggered
+			if epochFallbackTriggered {
+				continue
+			}
```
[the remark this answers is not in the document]
jordanschalm: Also, a heads up: service event processing was refactored and cleaned up in the [Active Pacemaker branch](https://github.com/onflow/flow-go/blob/feature/active-pacemaker/state/protocol/badger/mutator.go), which will be merged fairly soon.

state/protocol/badger/validity.go
```diff
@@ -197,6 +199,57 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
...
 }
 
+func isValidVersionBeacon(vb *flow.VersionBeacon, header *flow.Header) error {
+	if len(vb.RequiredVersions) == 0 {
+		return fmt.Errorf("required versions empty")
+	}
+
+	// first entry in a table must be a current version, so the height must be below the current block
+	if vb.RequiredVersions[0].Height > header.Height {
+		return fmt.Errorf("lowest required version height %d below current block's height %d", vb.RequiredVersions[0].Height, header.Height)
```
jordanschalm, line 209:
```suggestion
		return fmt.Errorf("lowest required version height %d above current block's height %d", vb.RequiredVersions[0].Height, header.Height)
```
Error message doesn't match logic or comment.

I don't quite understand this condition though. When we require a new version, we would use a height a few thousand blocks in the future, to make sure all nodes have time to prepare. So, the first `VersionBeacon` event we process would be guaranteed to have some `Height > header.Height`, and therefore would always fail this check... 🤔

state/protocol/badger/validity_test.go
```diff
@@ -101,3 +115,121 @@ func TestBootstrapInvalidEpochCommit(t *testing.T) {
...
+
+		vbHeight := uint64(0)
+
+		snap.On("VersionBeacon").Return(vb, vbHeight, state.ErrNoVersionBeacon)
+
+		err := validateVersionBeacon(snap)
+		require.NoError(t, err)
+	})
+
+	t.Run("height must be below highest block", func(t *testing.T) {
```
jordanschalm, line 138:
```suggestion
	t.Run("height must be below snapshot head", func(t *testing.T) {
```

storage/badger/operation/version_beacon_test.go
```diff
@@ -0,0 +1,82 @@
...
+		height1 := uint64(21)
+		height2 := uint64(37)
+		height3 := uint64(55)
+
+		// inserting 3 results at different height each has a ServiceEventCommit
```
jordanschalm, line 24:
```suggestion
		// indexing 3 version beacons at different heights
```

storage/badger/operation/version_beacon_test.go
```diff
@@ -0,0 +1,82 @@
...
+		err := db.Update(IndexVersionBeaconByHeight(vb1, height1))
+		require.NoError(t, err)
+
+		err = db.Update(IndexVersionBeaconByHeight(vb2, height2))
+		require.NoError(t, err)
+
+		err = db.Update(IndexVersionBeaconByHeight(vb3, height3))
+		require.NoError(t, err)
+
+		// insert result 2 again to make sure we tolerate duplicates
```
jordanschalm, line 34:
```suggestion
		// index version beacon 2 again to make sure we tolerate duplicates
```

storage/badger/version_beacon.go
```diff
@@ -0,0 +1,41 @@
...
+
+	res := &VersionBeacons{
+		db: db,
+	}
+
+	return res
+}
+
+// Highest finds the highest flow.VersionBeacon but no higher than maxHeight.
+// Returns storage.ErrNotFound if version beacon exists at or below the given height.
```
jordanschalm, line 28:
```suggestion
// Returns storage.ErrNotFound if no version beacon exists at or below the given height.
```

storage/util/testing.go
```diff
@@ -8,31 +8,23 @@ import (
...
-	receipts := storage.NewExecutionReceipts(metrics, db, results, storage.DefaultCacheSize)
-	index := storage.NewIndex(metrics, db)
-	payloads := storage.NewPayloads(db, index, guarantees, seals, receipts, results)
-	blocks := storage.NewBlocks(db, headers, payloads)
-	setups := storage.NewEpochSetups(metrics, db)
-	commits := storage.NewEpochCommits(metrics, db)
-	statuses := storage.NewEpochStatuses(metrics, db)
-	return headers, guarantees, seals, index, payloads, blocks, setups, commits, statuses, results
+	all := bstorage.InitAll(metrics, db)
+	return all
```
jordanschalm, line 21: 👍 Thanks for cleaning this up

state/protocol/badger/mutator.go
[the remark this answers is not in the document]
m4ksio: https://github.com/onflow/flow-go/pull/3736/files#diff-c79d3fc5d32f153917b7cd923ab4ba64ef22f12268ddd13f353a06391f68a96dR893 I added it alongside other Service Events validations, which happens when we extend the state, not only while finalizing.

fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc
m4ksio: Interesting, it was showing the same in GoLand. Nevertheless, I have unified this with tabs everywhere and it seems consistent in github now.

storage/badger/operation/version_beacon_test.go
m4ksio: It certainly won't hurt - so changed as requested

fvm/systemcontracts/system_contracts.go
```diff
@@ -30,11 +30,13 @@ const (
 	ContractNameFlowFees     = "FlowFees"
 	ContractStorageFees      = "FlowStorageFees"
 	ContractDeploymentAudits = "FlowContractAudits"
```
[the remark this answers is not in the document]
m4ksio: Fixed

fvm/systemcontracts/system_contracts_test.go
```diff
@@ -94,10 +94,13 @@ func checkServiceEvents(t *testing.T, chainID flow.ChainID) {
...
 	epochContractAddr := addresses[ContractNameEpoch]
+	versionContractAddr := addresses[ContractNameVersion]
 	// entries may not be empty
 	assert.NotEqual(t, flow.EmptyAddress, epochContractAddr)
+	assert.NotEqual(t, flow.EmptyAddress, versionContractAddr)
 
 	// entries must match internal mapping
 	assert.Equal(t, epochContractAddr, events.EpochSetup.Address)
 	assert.Equal(t, epochContractAddr, events.EpochCommit.Address)
+	assert.Equal(t, epochContractAddr, events.VersionTable.Address)
```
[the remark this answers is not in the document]
m4ksio: Fixed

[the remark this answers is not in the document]
m4ksio: Added

fvm/systemcontracts/system_contracts_test.go
```diff
@@ -94,10 +94,13 @@ func checkServiceEvents(t *testing.T, chainID flow.ChainID) {
 	require.True(t, ok, "missing chain %w", chainID.String())
 
 	epochContractAddr := addresses[ContractNameEpoch]
+	versionContractAddr := addresses[ContractNameVersion]
```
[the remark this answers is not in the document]
m4ksio: Not really, we need contracts deployed and their addresses known. Skipping would make this PR passing CI while it should not.

integration/testnet/client.go
```diff
@@ -193,16 +193,17 @@ func (c *Client) Account() *sdk.Account {
...
+	fmt.Printf("Waiting for transaction %s to be sealed...\n", txID)
 	errCount := 0
 	var result *sdk.TransactionResult
 	var err error
 	for result == nil || (result.Status != sdk.TransactionStatusSealed) {
 		childCtx, cancel := context.WithTimeout(ctx, time.Second*5)
-		result, err = c.client.GetTransactionResult(childCtx, id)
+		result, err = c.client.GetTransactionResult(childCtx, txID)
 		cancel()
+		fmt.Printf("Results for %s = %d, err = %s, \n", txID, result.Status, err)
```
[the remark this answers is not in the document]
m4ksio: I think for debug its better to use raw value

integration/tests/upgrades/suite.go
```diff
@@ -0,0 +1,259 @@
...
+	// start tracking blocks
+	s.Track(s.T(), ctx, s.Ghost())
+
+}
+
+func (s *Suite) SetupTest() {
+	s.log = unittest.LoggerForTest(s.Suite.T(), zerolog.InfoLevel)
+	s.log.Info().Msg("================> SetupTest")
+
+	//s.ctx, s.cancel = context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
m4ksio: Fixed

integration/tests/upgrades/suite.go
```diff
@@ -0,0 +1,259 @@
...
+	s.net.Start(ctx)
+
+	// start tracking blocks
+	s.Track(s.T(), ctx, s.Ghost())
+
+}
+
+func (s *Suite) SetupTest() {
+	s.log = unittest.LoggerForTest(s.Suite.T(), zerolog.InfoLevel)
+	s.log.Info().Msg("================> SetupTest")
```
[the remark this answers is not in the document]
m4ksio: Fixed

integration/tests/upgrades/suite.go
```diff
@@ -0,0 +1,259 @@
...
+		confs,
+		// set long staking phase to avoid QC/DKG transactions during test run
+		testnet.WithViewsInStakingAuction(10_000),
+		testnet.WithViewsInEpoch(100_000),
+	)
+	// initialize the network
+	s.net = testnet.PrepareFlowNetwork(s.T(), netConfig, flow.Localnet)
+
+	// start the network
+
```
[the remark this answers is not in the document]
m4ksio: Fixed

integration/tests/upgrades/suite.go
```diff
@@ -0,0 +1,259 @@
...
+	// start the network
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	s.net.Start(ctx)
+
+	s.net.Start(ctx)
```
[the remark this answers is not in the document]
m4ksio: Fixed

integration/Makefile
```diff
@@ -52,6 +52,10 @@ mvp-tests:
 execution-tests:
 	go test $(if $(VERBOSE),-v,) $(RACE_FLAG) $(if $(JSON_OUTPUT),-json,) $(if $(NUM_RUNS),-count $(NUM_RUNS),) -tags relic ./tests/execution/...
 
+.PHONY: upgrades-tests
+upgrades-tests:
```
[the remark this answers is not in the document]
m4ksio: Good point, added

integration/tests/upgrades/version_beacon_service_event_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type TestServiceEventVersionControl struct {
+	Suite
+}
+
+func (s *TestServiceEventVersionControl) TestEmittingVersionBeaconServiceEvent() {
+
+	height := uint64(21)
```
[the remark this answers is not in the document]
m4ksio: Added

integration/tests/upgrades/version_beacon_service_event_test.go
```diff
@@ -0,0 +1,105 @@
...
+	s.Require().NoError(err)
+
+	txResult, err := s.AccessClient().WaitForSealed(ctx, tx.ID())
+	s.Require().NoError(err)
+
+	sealed := s.ReceiptState.WaitForReceiptFromAny(s.T(), flow.Identifier(txResult.BlockID))
+
+	s.Require().Len(sealed.ExecutionResult.ServiceEvents, 1)
+
+	s.Require().IsType(&flow.VersionBeacon{}, sealed.ExecutionResult.ServiceEvents[0].Event)
```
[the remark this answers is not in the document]
m4ksio: Changed spacing to be more mainstream

integration/tests/upgrades/version_beacon_service_event_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	versionTable := sealed.ExecutionResult.ServiceEvents[0].Event.(*flow.VersionBeacon)
+
+	s.Require().Equal(uint64(5), versionTable.Sequence)
+
+	s.Require().Len(versionTable.RequiredVersions, 1)
+
+	s.Require().Equal(height, versionTable.RequiredVersions[0].Height)
```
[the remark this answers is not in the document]
m4ksio: Fixed

model/convert/service_event_test.go
```diff
@@ -47,4 +47,20 @@ func TestEventConversion(t *testing.T) {
...
+	t.Run("version beacon", func(t *testing.T) {
+
+		fixture, expected := unittest.VersionBeaconFixtureByChainID(chainID)
+
+		// convert Cadence types to Go types
+		event, err := convert.ServiceEvent(chainID, fixture)
+		require.NoError(t, err)
+		require.NotNil(t, event)
+
+		// cast event type to epoch commit
```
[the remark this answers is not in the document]
m4ksio: Fixed

model/flow/service_event_test.go
```diff
@@ -72,11 +86,26 @@ func TestEncodeDecode(t *testing.T) {
...
+			require.NoError(t, err)
+
+			outer = new(flow.ServiceEvent)
+			t.Logf("- debug: outer=%+v <-- before .Unmarshal()\n", outer)
+			err = json.Unmarshal(b, outer)
+			t.Logf("- debug: outer=%+v <-- after .Unmarshal()\n", outer)
+			require.NoError(t, err)
+			gotVersionTable, ok := outer.Event.(*flow.VersionBeacon)
+			require.True(t, ok)
+			assert.DeepEqual(t, versionBeacon, gotVersionTable, comparePubKey)
```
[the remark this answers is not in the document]
m4ksio: Good point!

state/protocol/badger/mutator.go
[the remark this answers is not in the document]
m4ksio: I think I understand the problem - we want to process Version Beacons events even if the epoch fallback has been triggered. I have updated the code, please check if that's looking fine now

state/protocol/badger/mutator_test.go
```diff
@@ -131,7 +131,7 @@ func TestSealedIndex(t *testing.T) {
 		require.NoError(t, err)
 
 		// build a chain:
-		// G <- B1 <- B2 (resultB1) <- B3 <- B4 (resultB2, resultB3) <- B5 (sealB1) <- B6 (sealB2, sealB3) <- B7
+		// G <- B1 <- B2 (resultB1(vb1)) <- B3 <- B4 (resultB2(vb2), resultB3(vb3)) <- B5 (sealB1) <- B6 (sealB2, sealB3) <- B7
```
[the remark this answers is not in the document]
m4ksio: Its a matter of opinion of course, but I don't find this overly complicated. Test itself is rather easy, its just that we need to build a long chain of blocks to test all the aspects - so its input or test data issue.
I don't think separate test would be any simpler, as the gist of test is to check if its properly mutates the state, so if not by input the same date would have to be mocked or initial state somehow prepared.

state/errors.go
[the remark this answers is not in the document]
m4ksio: Nicely put, thanks!

state/protocol/badger/validity.go
m4ksio: First entry so first version, if present of course, should always be the current one - so it should be currently active and therefore the height must be at or below the current block (block at which we mark it as active)
It is to make sure we are on the proper version that interprets event as intended. 
Indeed, the error message was wrong, thanks for pointing that out.

state/protocol/badger/validity.go
```diff
@@ -197,6 +199,57 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
 	return nil
 }
 
+func isValidVersionBeacon(vb *flow.VersionBeacon, header *flow.Header) error {
```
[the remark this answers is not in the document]
m4ksio: Good idea with `InvalidServiceEventError` - changed

state/protocol/badger/validity_test.go
m4ksio: VB event will most of the time be send as a service event included in a block, so I don't think changing it to snapshot head is valid here.

jordanschalm: The "snapshot head" and "highest block (in sealing segment)" are the same block, so I'm not sure why the change would be invalid. The suggestion is aiming to make the meaning clearer rather than change it. We could also say `below highest block in sealing segment`.

state/protocol/badger/mutator_test.go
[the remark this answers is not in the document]
jordanschalm: I agree that adding a separate test case would need to generate similar state to test against, and would almost certainly be more code in total. My point is about keeping the scope of tests limited, so each one is more understandable and isolated on its own, even if we have more test code in aggregate.

`TestSealedIndex` is responsible for testing that the index of sealed blocks is updated correctly as we incorporate new blocks into the state. Many other things can happen when we seal a block, besides updating this index, one of which now is updating the `VersionTable`.

However, we don't include assertions for everything that might happen when sealing a block into `TestSealedIndex`, because that would make the test very complicated. The test would be more difficult to understand, because it is longer and dealing with more concepts. If the test fails, it is more difficult to determine what caused the failure - the sealing index? or something obliquely related to sealing which was bolted on to the test for convenience?

After the changes in this PR, `TestSealedIndex` is 146 lines long and mentions by my count 16 named entities in its test data. It's a complicated test - many tests in this package are. Some of the complexity is necessary, but some of it can be mitigated by isolating distinct functionality into distinct test cases.

fvm/systemcontracts/system_contracts_test.go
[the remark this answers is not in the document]
jordanschalm: Sounds good, I assumed we'd want to merge before those contracts were deployed, but happy to wait on them as well 👍

state/protocol/badger/validity.go
```diff
@@ -199,29 +200,44 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
 	return nil
 }
 
+// isValidVersionBeacon validates internal structure of flow.VersionBeacon and in a context of given block,
+// which is assumed to be a block flow.VersionBeacon has been put into effect.
+// InvalidServiceEventError with appropriate message is returned if any validation fails.
+// Please refer to documentation of flow.VersionBeacon for a description of what's expected of a valid VersionBeacon.
 func isValidVersionBeacon(vb *flow.VersionBeacon, header *flow.Header) error {
 	if len(vb.RequiredVersions) == 0 {
 		return fmt.Errorf("required versions empty")
```
jordanschalm, line 209:
```suggestion
		return protocol.NewInvalidServiceEventError("required versions empty")
```

state/protocol/badger/validity_test.go
m4ksio: Yes, however the validation happens not only in a context of a snapshot but also when event is emitted (or included) so that's why I find this notion to be more universal

model/flow/version.go
```diff
@@ -11,12 +11,35 @@ type VersionControlRequirement struct {
...
+//   - no previous versions were ever set - first entry must be at least VersionThreshold blocks from the current block's height
+//   - there were previous version set - first entry's height must be below the current block's height and it's version must be
+//     the current running versions. Next one, so first upcoming version must be at least VersionThreshold blocks in the future.
+//
+// In other words, current active version, if ever set,  must be the first entry in the table. Then, any upcoming change must be at least
+// VersionThreshold blocks in the future regardless.
+
+// testy maksio kurwa
+
 // Sequence is event sequence number, which can be used to verify that no event has been skipped by the follower
```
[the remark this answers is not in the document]
m4ksio: Yep, cleared that one!

state/protocol/badger/validity.go
```diff
@@ -199,29 +200,44 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
...
 
-	// first entry in a table must be a current version, so the height must be below the current block
-	if vb.RequiredVersions[0].Height > header.Height {
-		return fmt.Errorf("lowest required version height %d below current block's height %d", vb.RequiredVersions[0].Height, header.Height)
-	}
+	minimumThresholdHeight := header.Height + flow.VersionThreshold
 
 	// handle case when only one version is present
 	if len(vb.RequiredVersions) == 1 {
+		if vb.RequiredVersions[0].Height < minimumThresholdHeight {
```
jordanschalm, line 216: I'm wondering if we really want to enforce this constraint here. 

The `VersionThreshold` exists because we might occasionally not seal a block for an extended period of time. We assume that we will never have a sealing outage larger than 1000 blocks. So we make sure that the height distance between a new `VersionRequirement`'s addition to the table and its effective height is at least 1000. That way we guarantee that a `VersionRequirement` will be known to an EN before that EN executes the first height where the requirement is effective.

In the diagram below, we want to make sure that `C.height > B.height`, because the distance between `A` and `B` can vary due to sealing lags.
```
/- Height where VersionBeacon event is emitted
|           /- Height where VersionBeacon event is sealed, and known to EN
|           |           /- First height where VersionRequirement is effective
v           v           v 
A <- ... <- B <- ... <- C
```

However, when validating the service event, we are are comparing `B.height` to `C.height`. If a VersionBeacon is emitted with a "version requirement buffer" near the threshold, then this check might fail unpredictably, depending on the sealing lag.

For example, suppose that at height 987 (block A) we emit a VersionBeacon with `height=2000, version=v2`. We have included a buffer of 1,013 blocks so we satisfy the 1000-block constraint. But then there is a sealing lag lasting 50 blocks, so we don't seal the event until height 1037. Then, we will fail validation and throw an error here. **⚠️ With the current error handling, that error would halt the chain and cause an outage.**

state/protocol/badger/validity.go
```diff
@@ -199,29 +200,44 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
...
+		}
 		_, err := validateRequirement(vb.RequiredVersions[0])
 		return err
 	}
 
+	// if first requirement below current height, second one must obey threshold
+	if vb.RequiredVersions[0].Height <= header.Height {
+		if vb.RequiredVersions[1].Height < minimumThresholdHeight {
+			return protocol.NewInvalidServiceEventError("required version (index=%d) height %d below minimum possible height (current block %d + threshold %d = %d)", 1, vb.RequiredVersions[1].Height, header.Height, flow.VersionThreshold, minimumThresholdHeight)
+		}
```
jordanschalm, line 227: I'm having a hard time convincing myself that these conditions will always hold. I think they depend on safeguards implemented in the smart contract, to prevent the creation of subsequent version requirements within a certain number of blocks of one another. 

Here's an example where I think we could run into validation problems here:
- at block 100, add version requirement (height=2000, version=v2) via event `vb1`
- at block 200, add version requirement (height=2100, version=v3) via `vb2`
- at block 2050, add version requirement (height=4000, version=v4) via `vb3`

When we are processing `vb3`, it would contain this version table:
- Height=2000, Version=v2
- Height=2100, Version=v3
- Height=4000, Version=v4

Since the first requirement is below current height, we would trigger the condition on line 225, which would fail because the second version requirement's height is only 50 ahead of `head`. 

As mentioned in the previous comment, service event validation is a very sensitive codepath. Processing a service event which fails validation can cause the chain to halt, unless otherwise handled. Because of this, I think it would be worthwhile in general to:
- tend toward being more lenient on inputs here, so long as they are safe to process
- clearly document how the smart contract guarantees that all conditions for validation will always pass

state/protocol/badger/validity.go
m4ksio: Thanks for this comment! 
There are two aspects here.
1. This validation happens not when finalizing, but when we extends the chain - so using your diagram, we compare A and C (B is used to index an event and we don't enforce threshold there). 
2. Error handling - or rather invalid input data handling. I assumed that we don't really expect invalid service events, hence halting a chain seemed like a good idea. However, now I'm not certain, but can't really think of an alternative. Shall we simply skip an event, risking possibly running invalid version?

jordanschalm: > This validation happens not when finalizing, but when we extends the chain

Agreed - the validation happens when extending the chain by inserting a block. However, we are inserting the block where the service event is processed (block B), not the the block where the service event was emitted (block A). 

When we call `isValidVersionBeacon` here:
https://github.com/onflow/flow-go/blob/1c05cb7f754267e6f5622fcf4378c182bf53abd4/state/protocol/badger/mutator.go#L896
`block` is the child of the block sealing the service event (equivalent to block B in my example diagram above). So in the validation, we are using `B.height` in our comparison.

state/protocol/badger/validity.go
m4ksio: One problem with leniency is that if we miss the version or run a wrong one we risk permanent damage to a node (as in requiring manual intervention/maybe rollback).
In the situation you described update at `vb3` would be considered invalid, as we don't want modifications within 1000 blocks, which include adding versions further away. Or maybe its better described as freezing the table if a new version is expected to take place within 1000 blocks.
Otherwise, as you correctly noted, we would need to track when each requirement was added, and this would overly complicate things. 
Given projected usage - having a bit restricted window to make changes is a good tradeoff for simplified design.
I would however love to to hear your proposal what would be the best way to handle invalid events without crashing the chain.

jordanschalm: > One problem with leniency is that if we miss the version or run a wrong one we risk permanent damage to a node

Let's define what outcomes might happen here:
1. A version beacon may be discarded, resulting in failure to upgrade to the desired version
2. A version beacon may be incorrectly interpreted, resulting in an upgrade to an undesired version
3. A version beacon may be inconsistently interpreted by different nodes, resulting in different nodes executing the same chunks with different versions.

When you mention permanent damage, I think this is referring to Outcome 3. In that case a node may produce an Execution Result or Approval which is incorrect, creating evidence for them to be slashed (and in case of ENs polluting their execution state going forward). In Outcomes 1 and 2, all nodes agree on the version, even though the version might be different than specified in the VersionBeacon. Of course if there is a bug causing a VersionBeacon to be interpreted incorrectly (Outcome 2), we should fix this. By leniency, I mean handle more potential failures gracefully rather than throwing an error which will cause the chain to halt, not introducing bugs. From my perspective, Outcome 1 is relatively harmless and greatly preferred to halting the chain.

All nodes must process blocks in a consistent manner, and agree on the resulting protocol state after processing each block. We can say that each node executes a function `f(old_state,next_block)->new_state`. `f` must be deterministic, irrespective of the Version Beacon. Whether `f` is strict or lenient with respect to its inputs is unrelated to whether `f` is deterministic. So I don't think we need to worry about Outcome 3 here.

> I would however love to to hear your proposal what would be the best way to handle invalid events without crashing the chain.

I think a reasonable choice would be to discard all VersionBeacon events which fail validation (Outcome 1). Ideally, future VersionBeacon's could still be applied (the sequence number currently prevents this, I believe).

janezpodhostnik (approved):

state/protocol/badger/validity.go
jordanschalm: Resolving because we removed height-based validation

jordanschalm:
A couple more change requests, then this PR is good to go from my perspective:
* Separating the new VersionBeacon tests from the sealing index tests https://github.com/onflow/flow-go/pull/3736#discussion_r1070067071
* Error handling for VersionBeacon event validation https://github.com/onflow/flow-go/pull/3736#discussion_r1073946750

model/convert/service_event.go
```diff
@@ -173,6 +175,36 @@ func convertServiceEventEpochCommit(event flow.Event) (*flow.ServiceEvent, error
...
+
+	versionTable := new(flow.VersionBeacon)
+
+	payload, err := json.Decode(nil, event.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal event payload: %w", err)
+	}
+
+	for _, cadenceVal := range payload.(cadence.Event).Fields[0].(cadence.Array).Values {
+		height := cadenceVal.(cadence.Struct).Fields[0].(cadence.UInt64).ToGoValue().(uint64)
```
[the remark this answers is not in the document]
m4ksio: Good question, other conversions do not check this, but I guess extra validation won't hurt

state/protocol/badger/state.go
```diff
@@ -123,6 +125,12 @@ func Bootstrap(
 			return fmt.Errorf("could not bootstrap epoch values: %w", err)
 		}
 
+		// 4½) initialize version beacon
```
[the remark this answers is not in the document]
m4ksio: 5 is already there, so instead of renaming rest of the steps I just slipped in-between

state/protocol/badger/mutator.go
```diff
@@ -590,12 +593,14 @@ func (m *FollowerState) Finalize(ctx context.Context, blockID flow.Identifier) e
...
+				err := isValidVersionBeacon(ev)
 
-				// Service events are strictly controlled within a system, and we don't envision multiple Version Beacon
-				// events in a single result. However, should this happen we will keep only the last one, since it's newer
-				// and takes precedence anyway
-				versionBeacon = ev
+				if err == nil {
+					// Service events are strictly controlled within a system, and we don't envision multiple Version Beacon
+					// events in a single result. However, should this happen we will keep only the last one, since it's newer
+					// and takes precedence anyway
```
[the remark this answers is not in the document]
m4ksio: Good point, multiple events here would be bordering on a bug in the system - I have added extra comments to make sure we handle this unlikely case.

state/protocol/badger/validity.go
```diff
@@ -197,6 +200,58 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
...
+		if err != nil {
+			return err
+		}
+
+		nextVersion, err := validateRequirement(next)
+		if err != nil {
+			return err
+		}
+
+		if nextVersion.LessThan(*currentVersion) {
```
zhangchiqing, line 236: So we allow current and next to be equal?

state/protocol/badger/validity.go
```diff
@@ -197,6 +200,58 @@ func isValidEpochCommit(commit *flow.EpochCommit, setup *flow.EpochSetup) error
...
+			return protocol.NewInvalidServiceEventError("higher requirement (index=%d) semver(%s) lower than previous(%s) (index=%d)", i+1, nextVersion, currentVersion, i)
+		}
+	}
+
+	return nil
+}
+
+// validateRequirement validates an individual version requirement (one entry in the version table).
+// Returns the parsed semver version for a valid requirement input.
+// Returns protocol.InvalidServiceEventError for an invalid semver string
```
zhangchiqing, line 246:
```suggestion
// Returns protocol.InvalidServiceEventError for an invalid semver string
// for instance, "2.1.1" would be a valid semver version, and "2.1" would be invalid.
```

model/flow/version.go
```diff
@@ -0,0 +1,65 @@
...
+
+	for i, v := range v.RequiredVersions {
+		other := other.RequiredVersions[i]
+
+		if v.Height != other.Height {
+			return false
+		}
+		if semver.Compare(v.Version, other.Version) != 0 {
+			return false
+		}
```
zhangchiqing, line 61: maybe move to `VersionControlRequirement.Equal`?

state/protocol/badger/validity.go
m4ksio: Yes. It doesn't create any confusion (should basically be a no-op)

model/flow/version.go
m4ksio: I don't want to make `VersionControlRequirement` a first class citizen, just a data container inside other

zhangchiqing (approved):

utils/unittest/service_events_fixtures.go
```diff
@@ -109,15 +109,15 @@ func EpochSetupFixtureByChainID(chain flow.ChainID) (flow.Event, *flow.EpochSetu
 
 // EpochCommitFixture returns an EpochCommit service event as a Cadence event
 // representation and as a protocol model representation.
-func EpochCommitFixtureByChainID(chain flow.ChainID) (flow.Event, *flow.EpochCommit) {
+func /**/ EpochCommitFixtureByChainID(chain flow.ChainID) (flow.Event, *flow.EpochCommit) {
```
jordanschalm, line 112:
```suggestion
func EpochCommitFixtureByChainID(chain flow.ChainID) (flow.Event, *flow.EpochCommit) {
```

jordanschalm (approved):

jordanschalm:
Note: This PR indexes `VersionBeacon` events by the height at which the are processed in the protocol state. This height is changing (by 1) with this PR: https://github.com/onflow/flow-go/pull/4149 

@janezpodhostnik

janezpodhostnik:
All parts of this are now in separate PRs

PR #3740, opened 2022-12-19
Simplify setting transaction procedure result values

pattyshack:
Just set the result values unconditionally.

Also remove IsSampled check around ComputationIntensities.  The intensities are used for metric logging rather than tracing and hence should not be sampled.


SaveTheRbtz (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3748, opened 2022-12-20
Filter ejected peers in the middleware topologypeers provider

kc1116:
This PR fixes a bug in the node block list wrapper where ejected identities are returned from the Identities func when provided with the not ejected identity filter. It filters block listed nodes again after modifying the ejected field and adds a unit test for this edgecase.

### DIFF
--- a/network/p2p/cache/node_blocklist_wrapper_test.go
+++ b/network/p2p/cache/node_blocklist_wrapper_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/model/flow/filter"
 	mocks "github.com/onflow/flow-go/module/mock"
+	"github.com/onflow/flow-go/network/p2p"
 	"github.com/onflow/flow-go/network/p2p/cache"
 	"github.com/onflow/flow-go/utils/unittest"
 )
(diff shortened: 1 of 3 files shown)


network/p2p/middleware/middleware.go
```diff
@@ -237,7 +237,7 @@ func (m *Middleware) NewPingService(pingProtocol protocol.ID, provider network.P
 func (m *Middleware) peerIDs(flowIDs flow.IdentifierList) peer.IDSlice {
 	result := make([]peer.ID, len(flowIDs))
```
AlexHentschel, line 238: agree with the usage of `append` as the default pattern. However, than you need to instantiate the slice here also with length of zero and capacity len(flowIDs).
```suggestion
	result := make([]peer.ID, 0, len(flowIDs))
```

AlexHentschel (approved):
Thanks for adding the test and its great documentation 👏.
I think you might have forgotten to set the slice length (see comment here)

network/p2p/cache/node_blocklist_wrapper_test.go
```diff
@@ -151,7 +152,32 @@ func (s *NodeBlocklistWrapperTestSuite) TestBlacklistedNode() {
...
+		numIdentities := len(combinedIdentities)
+
+		s.provider.On("Identities", mock.Anything).Return(combinedIdentities)
+
+		identities := s.wrapper.Identities(p2p.NotEjectedFilter)
 
+		require.Equal(s.T(), len(honestIdentities), len(identities)) // expected only honest nodes to be returned
+		for _, i := range identities {
+			_, isBlocked := blocklistLookup[i.NodeID]
+			require.Equal(s.T(), isBlocked, i.Ejected)
```
peterargue, line 173: nit: make this check more explicit
```suggestion
			_, isBlocked := blocklistLookup[i.NodeID]
			require.False(s.T(), isBlocked)
			require.False(s.T(), i.Ejected)
```

peterargue (approved):
Looks good, added one testing comment.

Can you also test with localnet that blocking a peer using the admin command results in their connection getting pruned?

kc1116:
> Looks good, added one testing comment.
> 
> Can you also test with localnet that blocking a peer using the admin command results in their connection getting pruned?

Will make the updates, I've also already done localnet testing multiple times everything is working as expected.

network/p2p/middleware/middleware.go
kc1116: https://github.com/onflow/flow-go/pull/3748/commits/1c1af46f6e71c00de7db09eceae43f927cc08ebd

network/p2p/cache/node_blocklist_wrapper_test.go
kc1116: https://github.com/onflow/flow-go/pull/3748/commits/2ae1c8b903737e765c6c9423800824db5ad7506b

kc1116:
bors merge

PR #3750, opened 2022-12-21
[Benchmark] Make follower return results

SaveTheRbtz:
This is the first step to remove separate transaction tracker and move this to a follower.

This PR removes `WaitForTransactionResult`, in followup diffs I'll see how much of the `WorkerStatsTracker` can be removed.


integration/benchmark/contLoadGenerator.go
```diff
@@ -407,16 +407,18 @@ func (lg *ContLoadGenerator) createAccounts(num int) error {
...
 	}
 
-	result, err := WaitForTransactionResult(context.Background(), lg.flowClient, createAccountTx.ID())
-	if err != nil {
-		return fmt.Errorf("failed to get transactions result: %w", err)
+	var result flowsdk.TransactionResult
+	select {
+	case result = <-ch:
+		lg.workerStatsTracker.IncTxExecuted()
+	case <-time.After(60 * time.Second):
```
pattyshack, line 419: should this be some sort of cancellable context (pass into sendTx)?

pattyshack (approved):

integration/benchmark/follower.go
```diff
@@ -291,17 +290,19 @@ func NewNopTxFollower(ctx context.Context, client access.Client, opts ...followe
...
 		txFollowerImpl: impl,
-		closedCh:       closedCh,
 	}
 	return nop, nil
 }
 
-// Follow always returns a closed channel.
-func (nop *nopTxFollower) Follow(ID flowsdk.Identifier) <-chan struct{} {
-	return nop.closedCh
+// Follow immediately returns a result.
```
gomisha, line 299: Can you add explanation of why it immediately returns a result?

integration/benchmark/follower_test.go
```diff
@@ -52,9 +52,12 @@ func TestTxFollower(t *testing.T) {
 	require.Equal(t, blockID, f.BlockID())
 
 	// test that transactionID is eventually discovered
-	unittest.AssertClosesBefore(t, f.Follow(transactionID), 1*time.Second)
-	// test that random transaction is not found
-	unittest.AssertNotClosesBefore(t, f.Follow(flowsdk.Identifier{0x7}), 200*time.Millisecond)
+	unittest.AssertReturnsBefore(t, func() { <-f.Follow(transactionID) }, 1*time.Second)
+	select {
+	case <-f.Follow(flowsdk.Identifier{0x7}):
```
gomisha, line 57: can a comment be added here that explains why the random `0x7` value was put here? (assuming that it tests that a random transaction shouldn't be found but can be more explicit)

gomisha: Why can't we use `unittest.AssertNotClosesBefore()` for asserting random transaction not found, just like before?

integration/benchmark/follower_test.go
```diff
@@ -83,12 +86,12 @@ func TestNopTxFollower(t *testing.T) {
...
-	unittest.AssertClosesBefore(t, f.Follow(flowsdk.Identifier{}), 1*time.Second)
+	unittest.AssertReturnsBefore(t, func() { <-f.Follow(flowsdk.Identifier{}) }, 1*time.Second)
 
 	// test that multiple Stops are safe
 	f.Stop()
 	f.Stop()
 
 	// test that all further Follow calls return immediately
-	unittest.AssertClosesBefore(t, f.Follow(flowsdk.Identifier{}), 1*time.Second)
+	unittest.AssertReturnsBefore(t, func() { <-f.Follow(flowsdk.Identifier{}) }, 1*time.Second)
```
gomisha, line 96: shouldn't the timeout be much less than 1 second since it's expected to return right away?

gomisha (commented):
A few minor requests for clarification.

integration/benchmark/follower_test.go
SaveTheRbtz: here it does not matter since channel is closed immediately.  Sleep here is only needed for the race in `AssertReturnsBefore`.

integration/benchmark/follower_test.go
SaveTheRbtz: Sadly, `AssertNotClosesBefore` relies on channel being `struct{}` and we have a result there now.

SaveTheRbtz:
bors merge

gomisha (approved):

SaveTheRbtz:
bors merge

PR #3751, opened 2022-12-21
[Benchmark] Limit GOMAXPROCS on localnet

SaveTheRbtz:
In a large benchmark environments it does not make sense to give every binary all `NumCPU` threads.  Otherwise we are looking at the order of thousands of GC threads threshing caches across the cores.

Also while here, remove `BINSTAT` since we do not use it.


gomisha (approved):

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #3757, opened 2022-12-22
[Networking] Removes deprecated fields off the protobuf message

yhassanzadeh13:
This PR removes the deprecated fields off the protobuf message and revamps the protobuf coding of the fields.


peterargue (approved):

vishalchangrani (approved):

yhassanzadeh13:
bors merge

network/message/message.proto
```diff
@@ -8,9 +8,6 @@ package message;
 // gossip network
 message Message {
   string ChannelID = 1;
-  bytes EventID = 2 [deprecated = true];
-  bytes OriginID = 3 [deprecated = true];
-  repeated bytes TargetIDs = 4;
-  bytes Payload = 5;
-  string Type = 6 [deprecated = true];
+  repeated bytes TargetIDs = 2;
+  bytes Payload = 3;
```
SaveTheRbtz, line 12: you may consider keeping old IDs, since there is no downside in that.

You should also `reserve`  deprecated ids and names: https://developers.google.com/protocol-buffers/docs/proto3#reserved

network/message/message.proto
```diff
@@ -8,9 +8,6 @@ package message;
 // gossip network
```
SaveTheRbtz, line 8: nit: whitespace in:
```
 syntax = "proto3";
```

You should likely format the protofile with `buf` (buf.build)

SaveTheRbtz (changes requested):

SaveTheRbtz:
bors cancel

network/message/message.proto
SaveTheRbtz: From the [protobuf docs](https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers):
> As you can see, each field in the message definition has a unique number. These field numbers are used to identify your fields in the [message binary format](https://developers.google.com/protocol-buffers/docs/encoding), and should not be changed once your message type is in use.

This means this would break compatibility between hosts using different proto versions.  Just want to make sure that this was intended.

network/message/message.proto
yhassanzadeh13: done

network/message/message.proto
yhassanzadeh13: As far as we discussed with @peterargue, backward compatibility was not an issue (as we freshly spork). Nevertheless, I reserved the deprecated ids here: https://github.com/onflow/flow-go/pull/3757/commits/c481a66161e40ca8e3a0ed28ea76d71062fb240d

peterargue: I think in this case it's OK to break compatibility between proto versions at a spork boundry.
* We currently don't support running previous versions of node software on the new spork network since there are going to be backwards incompatible changes to the node software.
* We also prevent accidental communication between nodes running on different spork networks by using spork-specific libp2p protocol and topic names.
* This proto is specific to the libp2p message payloads, which are never saved or (intentionally) replayed.

I also think it's fine to reserve these fields, and may avoid strange errors on observers that don't upgrade their node software

SaveTheRbtz (approved):

SaveTheRbtz:
bors merge

PR #3760, opened 2022-12-23
Crypto: code improvement

tarakby:
- add no-relic build versions of SPoCK functions (required for the emulator build)
- fix new gcc-12 and gcc-11 warnings (not triggered in Clang) when compiling the module C code
- minor optimization to identity signature check
- consolidate CGO include cflags


crypto/bls.go
```diff
@@ -243,18 +247,7 @@ const identityBLSSignatureHeader = byte(0xC0)
 // suspected to be equal to identity, which avoids failing the aggregated
 // signature verification.
 func IsBLSSignatureIdentity(s Signature) bool {
```
pattyshack, line 249: maybe undo changes to this function.  the original version has no gc overhead.  the new one does

pattyshack (approved):
I would keep the original IsBLSSignatureIdentity implementation.  otherwise this looks fine.

crypto/bls.go
tarakby: Interesting! Is it because of the use of the global variable `identityBLSSignature` ? The global aspect of the variable doesn't make the gc overhead negligible?

SaveTheRbtz: IMHO, the new version simpler, so it should be fine, since the GC overhead is constant.

crypto/bls.go
```diff
@@ -243,18 +247,7 @@ const identityBLSSignatureHeader = byte(0xC0)
...
-	if s[0] != identityBLSSignatureHeader {
-		return false
-	}
-	for _, b := range s[1:] {
-		if b != 0 {
-			return false
-		}
-	}
-	return true
+	return bytes.Compare(s, identityBLSSignature) == 0
```
SaveTheRbtz, line 250:
```suggestion
	return bytes.Equal(s, identityBLSSignature)
```

tarakby: Yep that was also flagged by the linter, but I was waiting for @pattyshack's opinion in case I need to undo all the change.

pattyshack: i don't feel strongly about this.  feel free to follow alexey's suggestion

SaveTheRbtz (approved):

tarakby:
bors merge

PR #3762, opened 2022-12-25
Khalil/1855 Remove block listed peers 

kc1116:
This PR does the following: 
- Add node block list distributor. Following the distributor-consumer pattern, it is used to notify consumers of updates to the node block list
- Adds notifier to the node block list wrapper that invokes the distributor callback OnNodeBlockListUpdate for blocklist with len > 0
- Adds middleware option that adds a consumer to the node block list distributor, this consumer removes each peer in the updated block list.


network/p2p/cache/node_blocklist_distributor.go
```diff
@@ -0,0 +1,38 @@
+package cache
+
+import (
+	"sync"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-go/network/p2p"
+)
+
+type OnNodeBlockListConsumer = func(blockList flow.IdentifierList)
```
peterargue, line 10: please add a concurrency safe & non-blocking notice here as well

network/p2p/consumer.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/onflow/flow-go/model/flow"
+)
+
+// NodeBlockListConsumer consumes notifications from the cache.NodeBlocklistWrapper whenever the block list is updated.
+// Implementations must:
+//   - be concurrency safe
+//   - be non-blocking
+type NodeBlockListConsumer interface {
```
peterargue, line 11: This is pretty specific. Maybe this should be `IdentityProviderConsumer` so we can add other event types related to identity updates later

network/p2p/cache/node_blocklist_wrapper.go
```diff
@@ -73,6 +76,11 @@ func (w *NodeBlocklistWrapper) Update(blocklist flow.IdentifierList) error {
 		return fmt.Errorf("failed to persist set of blocked nodes to the data base: %w", err)
 	}
 	w.blocklist = b
+
+	if blocklist.Len() > 0 {
+		w.notifier.OnNodeBlockListUpdate(blocklist)
```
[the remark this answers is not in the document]
AlexHentschel: +1

AlexHentschel:
1. In your implementation, you define an abstract function https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/network/p2p/cache/node_blocklist_distributor.go#L10 and this definition is used to subscribe to events https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/network/p2p/cache/node_blocklist_distributor.go#L26

  This pattern is used by _some_ pub-sub implementations (like [`FinalizationDistributor`](https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/consensus/hotstuff/notifications/pubsub/finalization_distributor.go#L32-L42) and [`QCCreatedDistributor`](https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/consensus/hotstuff/notifications/pubsub/qc_created_distributor.go#L29-L33))
2. The alternative approach is to enforce that subscribers have to implement a specific interface (as opposed to just an abstract function):
https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/consensus/hotstuff/notifications/pubsub/distributor.go#L33-L38 Here, [`hotstuff.Consumer`](https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/consensus/hotstuff/consumer.go) is an interface
   
There are tradeoffs between both approaches:
* With 1. you can subscribe to individual events (as long as the distributor supports it). The downside is that we are conceptually duplicating the definition of the events. In your case, you have the definition in the distributor: https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/network/p2p/cache/node_blocklist_distributor.go#L10 which mirrors the interface definition https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/network/p2p/consumer.go#L12-L16
* With 2, a consumer always has to implement receivers for all notifications (which in your case doesn't matter, because you anyway only have only one event). However, implementation overhead is negligible, if we add an implementation of a NoOp consumer (example [`notifications.NoopConsumer`](https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/consensus/hotstuff/notifications/noop_consumer.go)).  The advantages of this pattern are
   * you avoid duplicating consumer specifications and having to keep to keep the documentation of the multiple specifications in sync (see [Peter's comment](https://github.com/onflow/flow-go/pull/3762#discussion_r1066486268))
   * It is easy to find all Consumer implementations in the code base, even if they are currently not subscribed to some event. This is because most IDEs support searching for interface implementations. 
   * It forces developers to use consistent function names across the code base. If you want to ingest the `OnNodeBlockListUpdate` notifications, your struct has to have the appropriately named method https://github.com/onflow/flow-go/blob/8a86ca8fb0eb068ab274736a703b8f45ecda3505/network/p2p/consumer.go#L16
   * In comparison, with approach 1. , I could add an arbitrarily named method to my struct, e.g.
      ```golang
      func (x *SomeStruct) Foo(blockList flow.IdentifierList) {
      	...
      }
      ```
      You would be surprised how common inconsistent method naming is unless explicitly enforced. For somebody reviewing the code, it is much harder to deduce that  `Foo` is the implementation to consume the notification compared to the case, where `SomeStruct` has a method called `OnNodeBlockListUpdate`.
   * Lastly, you can make it very clear in the consumer implementation, that you are consuming specific events by adding the following code to the final consumer
      ```golang
      var _ p2p.NodeBlockListConsumer = (*SomeStruct)(nil)
      ```

In the end, I feel we can enforce a lot of consistency through the compiler, if we go with option 2. With option 1 it is a lot easier (and frequently happens by accident) that the code is inconsistently named, thereby adding unnecessary complexity and ambiguity for reviewers and during later maintenance.

network/p2p/cache/node_blocklist_distributor.go
AlexHentschel: I'd prefer to avoid using abstract functions to specify consumers. See [my comment below](https://github.com/onflow/flow-go/pull/3762#issuecomment-1382490758) for additional details.

AlexHentschel (commented):
Thanks. Very nice PR. Only have some stylistic comments. For convenience, I implemented my suggested revisions in PR https://github.com/onflow/flow-go/pull/3814 targeting your branch (also including [this suggestion](https://github.com/onflow/flow-go/pull/3762#discussion_r1066421948) from Peter). Feel free to merge my PR, not merge it PR, or pick only the changes you agree with.

network/p2p/consumer.go
kc1116: Lets keep the name NodeBlockListConsumer because we are consuming events from the node block list wrapper the naming is consistent. When dynamic identity table is implemented we will remove this consumer and update the naming.

kc1116:
Merged suggested changes from @AlexHentschel https://github.com/onflow/flow-go/pull/3762/commits/ffa265335d8f6008c77222fbb4adadb935f963ba 

cc: @AlexHentschel @peterargue

AlexHentschel (approved):
🎉

peterargue (approved):

Issue #3728, opened 2022-12-16
[Active Pacemaker] Extended consensus telemetry


AlexHentschel:
## Context
Metrika's node agent can be installed by node operators to monitor their node. The node agent currently ingests notifications emitted by the node (currently via scraping the node's logs locally though, this might be subject to later change)

Metrica explained  ([slack conversation](https://dapperlabs.slack.com/archives/C01FNL8769W/p1666753678186439?thread_ts=1666738052.518069&cid=C01FNL8769W)), which events their node agents currently ingests:
> the list of telemetry events (as keywords) that we're currently scraping are as follows:
> • `OnFinalizedBlock`
> • `OnProposingBlock`
> • `OnReceiveProposal`
> • `OnVoting`
> • `block vote received, forwarding block vote to hotstuff vote aggregator`
> You can in fact see the keywords and the associated metadata for each event that we're gathering in our node agent github repo [here](https://github.com/Metrika-Inc/agent/blob/fd1fd8ec522e687cc136cd7acb0d95c4e2ef44ad/flow/event.go#L23-L39).

The events `OnFinalizedBlock`, `OnReceiveProposal` are also present in the the new Jolteon consensus implementation. The events `OnProposingBlock` and `OnVoting` still exist but have been renamed to [`OnOwnProposal`](https://github.com/onflow/flow-go/blob/a3da26c56607179bf681faedc9152c8a97cba531/consensus/hotstuff/notifications/telemetry.go#L165-L182) and [`OnOwnVote`](https://github.com/onflow/flow-go/blob/a3da26c56607179bf681faedc9152c8a97cba531/consensus/hotstuff/notifications/telemetry.go#L157-L163), respectively. There are some minor revisions necessary on Metrika's side, which I have outlined in [this issue](url) in Metrika's repo. 

## Goal

It is important to note that Metrika is currently ingesting a _Flow-internal_ log message [`block vote received, forwarding block vote to hotstuff vote aggregator`](https://github.com/onflow/flow-go/blob/a3da26c56607179bf681faedc9152c8a97cba531/engine/consensus/message_hub/message_hub.go#L496), because there is not corresponding telemetry event. This is highly discouraged, because internal log messages are subject to change without notice.

[...]

PR #3763, opened 2022-12-28
[Consensus] Vote/Timeout aggregation telemetry events

durkmurder:
https://github.com/onflow/flow-go/issues/3728

### Context

This PR enriches telemetry events for timeout and vote aggregation to report each processed vote and timeout. Additionally it adds new logs for vote and timeout aggregation for easier debugging of possible issues.

### What was done
- Extended `hotstuff.Consumer` to report new events
- Updated all objects implementing `hotstuff.Consumer`
- Implemented telemetry events without notion of current path when compared to `EventHandler` telemetry
- Added extra logs for vote/timeout aggregation

### DIFF
--- a/consensus/hotstuff/consumer.go
+++ b/consensus/hotstuff/consumer.go
@@ -133,6 +133,26 @@ type Consumer interface {
 	// and must handle repetition of the same events (with some processing overhead).
 	OnStartingTimeout(model.TimerInfo)
 
+	// OnVoteProcessed notifications are produced by Vote Aggregation logic.
+	// Such a notification indicates that we have successfully processed vote.
+	// Prerequisites:
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnVoteProcessed(vote *model.Vote)
+
+	// OnTimeoutProcessed notifications are produced by Timeout Aggregation logic.
+	// Such a notification indicates that we have successfully processed timeout.
+	// Prerequisites:
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnTimeoutProcessed(timeout *model.TimeoutObject)
+
+	// OnCurrentViewDetails notifications are produced by EventHandler when proposing logic is triggered.
+	// Prerequisites:
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnCurrentViewDetails(finalizedView uint64, currentLeader flow.Identifier)
+
 	// OnDoubleVotingDetected notifications are produced by the Vote Aggregation logic
 	// whenever a double voting (same voter voting for different blocks at the same view) was detected.
 	// Prerequisites:
--- a/consensus/hotstuff/consumer.go
+++ b/consensus/hotstuff/consumer.go
@@ -115,7 +115,7 @@ type Consumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	OnQcTriggeredViewChange(qc *flow.QuorumCertificate, newView uint64)
+	OnQcTriggeredViewChange(oldView uint64, newView uint64, qc *flow.QuorumCertificate)
 
 	// OnTcTriggeredViewChange notifications are produced by PaceMaker when it moves to a new view
 	// based on processing a TC. The arguments specify the tc (first argument), which triggered
@@ -123,7 +123,7 @@ type Consumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	OnTcTriggeredViewChange(tc *flow.TimeoutCertificate, newView uint64)
+	OnTcTriggeredViewChange(oldView uint64, newView uint64, tc *flow.TimeoutCertificate)
 
 	// OnStartingTimeout notifications are produced by PaceMaker. Such a notification indicates that the
 	// PaceMaker is now waiting for the system to (receive and) process blocks or votes.
@@ -161,7 +161,7 @@ type Consumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	OnCurrentViewDetails(finalizedView uint64, currentLeader flow.Identifier)
+	OnCurrentViewDetails(currentView, finalizedView uint64, currentLeader flow.Identifier)
 
 	// OnDoubleVotingDetected notifications are produced by the Vote Aggregation logic
 	// whenever a double voting (same voter voting for different blocks at the same view) was detected.
--- a/consensus/hotstuff/consumer.go
+++ b/consensus/hotstuff/consumer.go
@@ -175,7 +175,7 @@ type Consumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	OnInvalidVoteDetected(*model.Vote)
+	OnInvalidVoteDetected(err model.InvalidVoteError)
 
 	// OnVoteForInvalidBlockDetected notifications are produced by the Vote Aggregation logic
 	// whenever vote for invalid proposal was detected.
@@ -196,7 +196,7 @@ type Consumer interface {
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
-	OnInvalidTimeoutDetected(*model.TimeoutObject)
+	OnInvalidTimeoutDetected(err model.InvalidTimeoutError)
 }
 
 // QCCreatedConsumer consumes outbound notifications produced by HotStuff and its components.
--- a/consensus/hotstuff/consumer.go
+++ b/consensus/hotstuff/consumer.go
@@ -134,14 +134,14 @@ type Consumer interface {
 	OnStartingTimeout(model.TimerInfo)
 
 	// OnVoteProcessed notifications are produced by Vote Aggregation logic.
-	// Such a notification indicates that we have successfully processed vote.
+	// Such a notification indicates that we have processed a vote.
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
 	OnVoteProcessed(vote *model.Vote)
 
 	// OnTimeoutProcessed notifications are produced by Timeout Aggregation logic.
-	// Such a notification indicates that we have successfully processed timeout.
+	// Such a notification indicates that we have processed a timeout.
 	// Prerequisites:
 	// Implementation must be concurrency safe; Non-blocking;
 	// and must handle repetition of the same events (with some processing overhead).
--- a/consensus/hotstuff/consumer.go
+++ b/consensus/hotstuff/consumer.go
@@ -133,15 +133,15 @@ type Consumer interface {
 	// and must handle repetition of the same events (with some processing overhead).
 	OnStartingTimeout(model.TimerInfo)
 
-	// OnVoteProcessed notifications are produced by Vote Aggregation logic.
-	// Such a notification indicates that we have processed a vote.
+	// OnVoteProcessed notifications are produced by the Vote Aggregation logic, each time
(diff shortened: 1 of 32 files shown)


consensus/hotstuff/notifications/telemetry.go
```diff
@@ -215,10 +251,9 @@ type PathHandler struct {
 
 // NewPathHandler instantiate a new PathHandler.
 // The PathHandler has no currently open path
-func NewPathHandler(log zerolog.Logger, chain flow.ChainID) *PathHandler {
+func NewPathHandler(log zerolog.Logger) *PathHandler {
 	return &PathHandler{
-		chain:       chain,
-		log:         log.With().Str("hotstuff", "telemetry").Str("chain", chain.String()).Logger(),
+		log:         log.With().Str("hotstuff", "telemetry").Logger(),
```
AlexHentschel, line 256: I personally would prefer to keep `chain.String()` as part of the log message. Maybe the keyword "chain" is misleading. 

Context:
* We have multiple consensus instances running in parallel: main consensus instance as well as several collector clusters. 
* The telemetry's main purpose is to allow analyzing the collective "swarm behaviour" of a particular consensus instance. For collector nodes, it is not apparent which collectors nodes belong to each of the clusters. 

The purpose of `chain` was to provide an easy way to filter for the telemetry logs of a particular cluster. For example the following [query](https://dapperlabs.grafana.net/explore?left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnetwork%3D%5C%22mainnet20%5C%22%7D%20%7C%3D%20%5C%22telemetry%5C%22%20%7C%3D%20%5C%22cluster-58-bf8cea8d86557909a71c8113860e7d78e7c87c084140c35f3c6d6c07d29473ca%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-15m%22,%22to%22:%22now%22%7D%7D&orgId=1) collects the telemetry from one particular cluster: 
```
{network="mainnet20"} |= "telemetry" |= "cluster-58-bf8cea8d86557909a71c8113860e7d78e7c87c084140c35f3c6d6c07d29473ca"
```

engine/collection/epochmgr/factories/hotstuff.go
```diff
@@ -73,14 +73,16 @@ func (f *HotStuffFactory) CreateModules(
 	updater module.Finalizer,
 ) (*consensus.HotstuffModules, module.HotstuffMetrics, error) {
 
+	log := f.createLogger(cluster)
+
 	// setup metrics/logging with the new chain ID
```
AlexHentschel, line 78:
```suggestion
	// setup metrics/logging with the new chain ID
	log := f.createLogger(cluster)
```

AlexHentschel (commented):
looks good. Like the overall approach.

consensus/hotstuff/consumer.go
```diff
@@ -133,6 +133,26 @@ type Consumer interface {
...
+
+	// OnTimeoutProcessed notifications are produced by Timeout Aggregation logic.
+	// Such a notification indicates that we have processed a timeout.
+	// Prerequisites:
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnTimeoutProcessed(timeout *model.TimeoutObject)
+
+	// OnCurrentViewDetails notifications are produced by EventHandler when proposing logic is triggered.
+	// Prerequisites:
```
jordanschalm, line 151:
```suggestion
	// These notifications are generally not produced for all views (for example skipped views).
	// These notifications are guaranteed to be produced for all views we enter after fully processing a message.
	// Example 1: 
	//   - We are in view 8. We process a QC with view 10, causing us to enter view 11.
	//   - Then this notification will be produced for view 11.
	// Example 2: 
	//   - We are in view 8. We process a proposal with view 10, which contains a TC for view 9 and TC.NewestQC for view 8.
	//   - The QC would allow us to enter view 9 and the TC would allow us to enter view 10, 
	//     so after fully processing the message we are in view 10.
	//   - Then this notification will be produced for view 10, but not view 9
	// Prerequisites:
```

consensus/hotstuff/notifications/log_consumer.go
```diff
@@ -140,6 +140,26 @@ func (lc *LogConsumer) OnStartingTimeout(info model.TimerInfo) {
...
+}
+
+func (lc *LogConsumer) OnTimeoutProcessed(timeout *model.TimeoutObject) {
+	log := timeout.LogContext(lc.log).Logger()
+	log.Debug().Msg("processed timeout object")
+}
+
+func (lc *LogConsumer) OnCurrentViewDetails(finalizedView uint64, currentLeader flow.Identifier) {
+	lc.log.Info().
+		Uint64("finalized_view", finalizedView).
```
jordanschalm, line 158: can add current view field here, after adding to notification method

consensus/hotstuff/notifications/telemetry.go
```diff
@@ -25,20 +25,25 @@ import (
...
+// Additionally, the TelemetryConsumer reports events related to vote and timeout aggregation
+// but those events are not bound to a path, so they are reported differently.
 // Generally, the TelemetryConsumer could export the collected data to a variety of backends.
 // For now, we export the data to a logger.
 //
 // Telemetry does NOT capture slashing notifications
 type TelemetryConsumer struct {
 	NoopConsumer
-	pathHandler *PathHandler
+	pathHandler  *PathHandler
```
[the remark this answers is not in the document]
jordanschalm: +1 for option 2

consensus/hotstuff/timeoutcollector/timeout_collector.go
```diff
@@ -25,12 +28,17 @@ type TimeoutCollector struct {
...
-func NewTimeoutCollector(view uint64,
+func NewTimeoutCollector(log zerolog.Logger,
+	view uint64,
 	notifier hotstuff.Consumer,
 	collectorNotifier hotstuff.TimeoutCollectorConsumer,
 	processor hotstuff.TimeoutProcessor,
 ) *TimeoutCollector {
 	return &TimeoutCollector{
+		log: log.With().
+			Str("hotstuff", "timeout_collector").
```
jordanschalm, line 39:
```suggestion
			Str("component", "hotstuff.timeout_collector").
```

consensus/hotstuff/timeoutcollector/timeout_processor.go
```diff
@@ -73,6 +76,10 @@ func NewTimeoutProcessor(committee hotstuff.Replicas,
 		return nil, fmt.Errorf("could not retrieve timeout weight threshold for view %d: %w", view, err)
 	}
 	return &TimeoutProcessor{
+		log: log.With().
+			Str("hotstuff", "timeout_processor").
```
jordanschalm, line 80:
```suggestion
			Str("component", "hotstuff.timeout_processor").
```

consensus/hotstuff/notifications/telemetry.go
jordanschalm, line 256:
```suggestion
		log:         log.With().Str("component", "hotstuff.telemetry").Logger(),
```

consensus/hotstuff/notifications/telemetry.go
```diff
@@ -25,20 +25,25 @@ import (
...
+	noPathLogger zerolog.Logger
 }
 
 var _ hotstuff.Consumer = (*TelemetryConsumer)(nil)
 
-func NewTelemetryConsumer(log zerolog.Logger, chain flow.ChainID) *TelemetryConsumer {
+func NewTelemetryConsumer(log zerolog.Logger) *TelemetryConsumer {
```
jordanschalm, line 42: The telemetry consumer is of limited use without the logger being tagged with the `chain_id` we're reporting about. That used to be ensured here by passing it in explicitly - now we are relying on the caller to make sure the logger argument is tagged appropriately.

**Suggestion:** Add a note in the godoc here and in `NewPathHandler` that `log` must have a field called `chain` set to the chain ID of the consensus instance.

consensus/hotstuff/notifications/telemetry.go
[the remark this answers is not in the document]
AlexHentschel: I assume option 2 would entail a bunch of changes across many files. Maybe it would be best to commit this PR first, adding just a TODO comment and then follow up with a PR for the interface refactoring.

consensus/hotstuff/timeoutcollector/timeout_collector.go
```diff
@@ -77,6 +86,7 @@ func (c *TimeoutCollector) processTimeout(timeout *model.TimeoutObject) error {
 	err := c.processor.Process(timeout)
 	if err != nil {
 		if model.IsInvalidTimeoutError(err) {
+			c.log.Err(err).Msgf("invalid timeout detected")
```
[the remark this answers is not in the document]
durkmurder: Right, and we have that already, this will be logged as warning by the consumer of events but for know I think it's useful to have a message that specifically prints the error for us to get better observability if something goes wrong. During previous tests this type of logs proved to be useful.

consensus/hotstuff/consumer.go
```diff
@@ -133,6 +133,26 @@ type Consumer interface {
...
+	// Implementation must be concurrency safe; Non-blocking;
+	// and must handle repetition of the same events (with some processing overhead).
+	OnCurrentViewDetails(finalizedView uint64, currentLeader flow.Identifier)
```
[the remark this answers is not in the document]
durkmurder: This event is emitted as part of `EventHandler` logic, meaning the current view is included in "path opening" event, while this event is not a path opening event it just provides information.

consensus/hotstuff/votecollector/combined_vote_processor_v2.go
```diff
@@ -233,6 +233,9 @@ func (p *CombinedVoteProcessorV2) Process(vote *model.Vote) error {
 		}
 	}
 
+	totalWeight := p.stakingSigAggtor.TotalWeight()
+	p.log.Debug().Msgf("processed vote, total weight=(%d), required=(%d)", totalWeight, p.minRequiredWeight)
+
 	// checking of conditions for building QC are satisfied
 	if p.stakingSigAggtor.TotalWeight() < p.minRequiredWeight {
```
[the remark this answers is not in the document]
durkmurder: good catch

consensus/hotstuff/notifications/telemetry.go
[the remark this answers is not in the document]
durkmurder: While I agree that our consumer is rather big at the moment and some decoupling would be beneficial I don't particularly agree that we have an issue right now. Yes, we have some white-spaces in documentation where we don't specifically mention that we it's intended for single-threaded usage but we implicitly guarantee it based on how we use it. All calls to `pathHandler` happen exclusively in `EventHandler` or code which is called by `EventHandler`, this is bound to notion of _path_. Since `EventHandler` opens a path and closes it, and events that use `pathHandler` must be part of _some_ path. For other events(vote and timeout aggregation) we specifically use `noPathHandler` which is a plain logger atm. 

I also thing that option 2 is better and cleaner, created an issue: https://github.com/dapperlabs/flow-go/issues/6453

consensus/hotstuff/notifications/telemetry.go
durkmurder: Do we really need it for main consensus? 
I was under impression that `chain_id` really matters for cluster consensus for us to clearly see what log messages are for which epoch, but with main consensus it should be "mainnet" every-time? This is the main reason why I didn't add `chain_id` for main consensus events but applied it for cluster consensus.

jordanschalm: Eventually consensus nodes will need to run Follower instances for cluster consensus. Then it would be useful to be explicit about which chain we're referring to in logs, since consensus nodes would be worried about multiple chain IDs at the same time. I agree it's not strictly necessary now, but it doesn't hurt and then we don't need to change it later.

consensus/hotstuff/consumer.go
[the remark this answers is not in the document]
durkmurder: Decided to extend this and similar methods to report current view wherever it is possible.

jordanschalm (approved):

AlexHentschel (commented):
🎉

AlexHentschel (approved):

durkmurder:
bors merge

PR #3764, opened 2022-12-30
[Consensus] Extending sealing segment with extra information

durkmurder:
Implements https://github.com/dapperlabs/flow-go/issues/6443

### Context

This PR implements a few important changes for nodes bootstrapping and generating sealing segment for other nodes.
- Updated sealing segment to include extra blocks which are needed to extend history when bootstrapping from sealing segment
- Added `SporkRootBlockHeight` as global parameter for snapshot
- Updated collection guarantee validation logic to depend on `SporkRootBlockHeight`
- Updated scaffold logic to perform extra validation of root snapshot
- Added sanity checks for root snapshot for consensus nodes


state/protocol/badger/snapshot.go
```diff
@@ -284,11 +284,58 @@ func (s *Snapshot) SealingSegment() (*flow.SealingSegment, error) {
...
+	limitHeight := s.state.sporkRootBlockHeight
+	sealingSegmentMaxLength := s.state.maxSealingSegmentLength
+	if highestBlock.Height > sealingSegmentMaxLength {
+		lowestSealingSegmentHeight := highestBlock.Height - sealingSegmentMaxLength
+		if limitHeight < lowestSealingSegmentHeight {
+			limitHeight = lowestSealingSegmentHeight
+		}
+	}
+
+	lowestSealingSegmentBlock, err := s.state.headers.ByBlockID(lowestBlockID)
```
jordanschalm, line 311:
```suggestion
	// the highest sealed block is the first block above any extra blocks
	highestSealedBlock, err := s.state.headers.ByBlockID(lowestBlockID)
```
Similar to my comment about `SealingSegment.Lowest()`, I think a different name is more appropriate for this block now.

state/protocol/badger/state.go
```diff
@@ -33,6 +33,10 @@ type State struct {
 	}
 	// cache the root height because it cannot change over the lifecycle of a protocol state instance
 	rootHeight uint64
+	// cache the spork root block height because it cannot change over the lifecycle of a protocol state instance
+	sporkRootBlockHeight uint64
```
jordanschalm, line 37:
```suggestion
	rootSealingSegmentMinHeight uint64
```

storage/badger/operation/heights.go
```diff
@@ -14,6 +14,14 @@ func RetrieveRootHeight(height *uint64) func(*badger.Txn) error {
 	return retrieve(makePrefix(codeRootHeight), height)
 }
 
+func InsertSporkRootBlockHeight(height uint64) func(*badger.Txn) error {
```
jordanschalm, line 17:
```suggestion
func InsertRootSealingSegmentMinHeight(height uint64) func(*badger.Txn) error {
```

storage/badger/operation/heights.go
```diff
@@ -14,6 +14,14 @@ func RetrieveRootHeight(height *uint64) func(*badger.Txn) error {
...
+	return insert(makePrefix(codeSporkRootBlockHeight), height)
+}
+
+func RetrieveSporkRootBlockHeight(height *uint64) func(*badger.Txn) error {
```
jordanschalm, line 21:
```suggestion
func RetrieveRootSealingSegmentMinHeight(height *uint64) func(*badger.Txn) error {
```

storage/badger/operation/prefix.go
```diff
@@ -32,6 +32,7 @@ const (
 	codeExecutedBlock           = 23 // latest executed block with max height
 	codeRootHeight              = 24 // the height of the first loaded block
```
jordanschalm, line 33:
```suggestion
	codeRootHeight              = 24 // the height of the highest block contained in the root snapshot
```

storage/badger/operation/prefix.go
```diff
@@ -32,6 +32,7 @@ const (
...
 	codeLastCompleteBlockHeight = 25 // the height of the last block for which all collections were received
+	codeSporkRootBlockHeight    = 26 // the height of the lowest block in current spork that our node is aware of
```
jordanschalm, line 35:
```suggestion
	codeRootSealingSegmentMinHeight    = 26 // the height of the lowest block in current spork that our node is aware of
```

state/protocol/badger/state.go
AlexHentschel, line 37: The intention behind [my suggestion to introduce `sporkRootBlockHeight`](https://github.com/dapperlabs/flow-go/issues/6443#issuecomment-1364320554) was that this variable would hold the height of the spork's root block. 
* The local history in a node might not reach all the way back to the spork's root block
* Nevertheless, the node knows the height of the spork's root block

_Suggestion_:
* I would suggest to include a getter method `SporkRootBlockHeight() (uint, error)` into the https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/params.go#L43 interface, which is exposed by `Snapshot`: https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/snapshot.go#L123
* include `sporkRootBlockHeight` into the `EncodableSnapshot`, which we use for bootstrapping. I would place `sporkRootBlockHeight` into the `Params` https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/inmem/encodable.go#L19 section
   * we assign `root.Header.Height` to `sporkRootBlockHeight` here: https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/inmem/convert.go#L290-L295
* When bootstrapping the state, we would use the value returned by the root snapshot here: https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/badger/state.go#L261
* For consensus nodes, it would be great to include an additional sanity check that the sealing segment covers the history (see condition (i) here: https://github.com/dapperlabs/flow-go/issues/6443#issuecomment-1364320554
   ```golang
   limit = max(header.Height - m.cfg.transactionExpiry, SporkRootBlockHeight)
   ```
* replace the [`IsSporkRootSnapshot`](https://github.com/onflow/flow-go/blob/a848093dccc16618a51a9d16d0b6faece2a6234b/state/protocol/util.go#L67-L77) with a direct comparison of `root

state/protocol/badger/snapshot.go
```diff
@@ -284,11 +284,58 @@ func (s *Snapshot) SealingSegment() (*flow.SealingSegment, error) {
...
 
-	err = fork.TraverseForward(s.state.headers, s.blockID, scraper, fork.IncludingBlock(seal.BlockID))
+	lowestBlockID := seal.BlockID
+	err = fork.TraverseForward(s.state.headers, s.blockID, scraper, fork.IncludingBlock(lowestBlockID))
 	if err != nil {
 		return nil, fmt.Errorf("could not traverse sealing segment: %w", err)
 	}
 
+	// strictly speaking sealing segment length should be bounded by next formula:
+	// limitHeight = max(s.block.Height - maxSealingSegmentLength, SporkRootBlockHeight)
```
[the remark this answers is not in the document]
AlexHentschel: Regarding the [edge case described in Jordan's comment](https://github.com/onflow/flow-go/pull/3764#discussion_r1063812369): 
* I think this edge case could be covered by my suggestion https://github.com/onflow/flow-go/pull/3764#discussion_r1063864674 to include the `sporkRootHeight` as a global parameter in the protocol state. Therefore, the node ingesting the snapshot could decide whether the snapshot is long enough to satisfy its needs.

storage/badger/operation/heights.go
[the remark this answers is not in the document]
AlexHentschel: As [described below](https://github.com/onflow/flow-go/pull/3764#discussion_r1063864674), I think this method would be useful, provided that its implementation adheres to the name: namely that this methods inserts the height of the root block the spork started with. Of course, this height (additional) information needs to be contained in the snapshot / bootstrapping information.

model/flow/sealing_segment.go
```diff
@@ -65,6 +67,10 @@ type SealingSegment struct {
 	// Blocks contain the chain segment blocks in ascending height order.
 	Blocks []*Block
 
+	// ExtraBlocks contain the chain extra blocks in ascending height order. These blocks
+	// are connecting to the Blocks[0](the lowest block of sealing segment)
```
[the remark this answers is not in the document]
AlexHentschel: added in https://github.com/onflow/flow-go/commit/d95bc66b644c70fd61940ad6f218e2c4a4e51522

state/protocol/badger/validity.go
```diff
@@ -350,3 +350,38 @@ func validateClusterQC(cluster protocol.Cluster) error {
...
+	sporkRootBlockHeight, err := snapshot.Params().SporkRootBlockHeight()
+	if err != nil {
+		return fmt.Errorf("could not query spork root block height: %w", err)
+	}
+
+	cfg := DefaultConfig()
+	if head.Height+sporkRootBlockHeight < cfg.transactionExpiry {
+		return fmt.Errorf("invalid root snapshot length, expecting at least %d, got %d",
+			cfg.transactionExpiry, head.Height-sporkRootBlockHeight)
+	}
```
jordanschalm, line 384: This condition should be primarily concerned with the sealing segment length. I don't think the current check is sufficient.

The condition should be like:
```go
blocksInSpork := head.Height - sporkRootBlockHeight
if blocksInSpork < cfg. transactionExpiry {
  // the distance to spork root is less than transaction expiry, we need all blocks back to the spork root.
  if len(sealingSegment.AllBlocks()) != blocksInSpork {
    // err
  }
else {
  // the distance to spork root is more than transaction expiry, we need at least `transactionExpiry` many blocks
  if len(sealingSemgn.AllBlocks()) < cfg.transactionExpiry {
    // err
  }
}

// pass the check
```

durkmurder: :facepalm: thanks for spotting this, completely misimplemented it

state/protocol/badger/validity.go
```diff
@@ -377,11 +377,27 @@ func SanityCheckConsensusNodeRootSnapshotValidity(snapshot protocol.Snapshot) er
...
-	if head.Height+sporkRootBlockHeight < cfg.transactionExpiry {
-		return fmt.Errorf("invalid root snapshot length, expecting at least %d, got %d",
-			cfg.transactionExpiry, head.Height-sporkRootBlockHeight)
+	sealingSegment, err := snapshot.SealingSegment()
+	if err != nil {
+		return fmt.Errorf("could not query sealing segment: %w", err)
 	}
 
+	sealingSegmentLength := uint64(len(sealingSegment.AllBlocks()))
+	cfg := DefaultConfig()
```
jordanschalm, line 386: By using the default config rather than a config instance actually used by the protocol state instance, we're already assuming that the `transactionExpiry` value is constant. I think this would be clearer if we just use the constant value directly: `flow.DefaultTransactionExpiry`.

(The same argument holds for the `Config` type as well. As-is, it is not safe to configure any value for `config.transactionExpiry` besides the global constant, so it would be clearer to use the constant directly.)

jordanschalm (approved):

state/protocol/badger/validity.go
durkmurder: Yeah, agree, I have removed config at all since we use the default one all the time. Will bring it back when it's needed.

state/protocol/badger/state.go
```diff
@@ -461,6 +474,15 @@ func (state *State) bootstrapSporkInfo(root protocol.Snapshot) func(*badger.Txn)
...
 
+		sporkRootBlockHeight, err := params.SporkRootBlockHeight()
+		if err != nil {
+			return fmt.Errorf("could not get spork root block height: %w", err)
+		}
+		err = operation.InsertSporkRootBlockHeight(sporkRootBlockHeight)(tx)
+		if err != nil {
+			return fmt.Errorf("could not insert spork root block height: %w", err)
+		}
+
```
AlexHentschel, line 485: ## ⚠️ 
I think we are missing the `EpochCommitSafetyThreshold`

AlexHentschel (commented):
I am worried that the edge case Jordan is describing still persists in the current implementation. Added a PR, which I believe should fix it: [PR #3816](https://github.com/onflow/flow-go/pull/3816) targeting your branch updating the implementation to what I think should be the complete logic. I added detailed comments. _Caution_: in my PR, I  numbered the conditions differently than in [my comment in #6443](https://github.com/dapperlabs/flow-go/issues/6443) (Caution: did this late at night, and no tests. Please inspect thoroughly!)

state/protocol/badger/state.go
durkmurder: we insert it in [bootstrapSporkInfo](https://github.com/onflow/flow-go/blob/yurii/6443-sealing-segment-service-event-support/state/protocol/badger/state.go#L497)

state/protocol/badger/state.go
```diff
@@ -33,6 +33,10 @@ type State struct {
 	}
 	// cache the root height because it cannot change over the lifecycle of a protocol state instance
 	rootHeight uint64
+	// cache the spork root block height because it cannot change over the lifecycle of a protocol state instance
+	sporkRootBlockHeight uint64
+	// maximum length of sealing segment that replica will be building
+	maxSealingSegmentLength uint64
```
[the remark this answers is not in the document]
durkmurder: The big problem with this is that sealing segment building logic is coupled to the `Snapshot` I think it's doesn't look nice to include a parameter for `Snapshot.SealingSegment()`. That was the primary reason to parameterize it this way.

AlexHentschel (approved):

durkmurder:
bors retry

PR #3766, opened 2023-01-03
Set transaction procedure's trace span via context

pattyshack:
Transaction procedure's span was set using a snowflake api.  This change makes span setup consistent with the rest.

This also plumb trace span correctly through script environment.


SaveTheRbtz (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3776, opened 2023-01-05
Improve error code search.

pattyshack:
Don't rely on As() to unwrap errors when searching for a non-fatal error code since multierror.Error's As() will only inspects the first error in the list (other error codes in the errors list are ignored).

### DIFF
--- a/fvm/errors/errors.go
+++ b/fvm/errors/errors.go
@@ -4,15 +4,19 @@ import (
 	stdErrors "errors"
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/onflow/cadence/runtime"
 	"github.com/onflow/cadence/runtime/errors"
 )
 
+type Unwrappable interface {
+	Unwrap() error
+}
+
 type CodedError interface {
 	Code() ErrorCode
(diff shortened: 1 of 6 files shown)


janezpodhostnik (approved):

fvm/errors/errors.go
```diff
@@ -134,21 +138,36 @@ func HasErrorCode(err error, code ErrorCode) bool {
...
-	return Find(coded.Unwrap(), code)
+	// NOTE: we need to special case multierror.Error since As() will only
+	// inspect the first error within multierror.Error.
+	errors, ok := unwrappable.(*multierror.Error)
+	if !ok {
+		return Find(unwrappable.Unwrap(), code)
+	}
+
+	for _, internalErr := range errors.Errors {
+		coded = Find(internalErr, code)
```
zhangchiqing, line 164: internalErr sounds like a "bug"

```suggestion
	for _, innerError := range errors.Errors {
		coded = Find(innerError, code)
```

zhangchiqing (approved):

pattyshack:
bors merge

PR #3780, opened 2023-01-06
[Benchmark] Match log-level w/ production

SaveTheRbtz:
Right now we are logging too much data from ~30 daemons on the same box most of which debug data.  Since production runs with an info level by default, let's switch to it in benchmarks too.

PS. Keep default at DEBUG to preserve backwards compatibility with previous localnet setups (it is quite useful for local troubleshooting.)


gomisha (approved):

pattyshack (approved):

SaveTheRbtz:
bors merge

PR #3783, opened 2023-01-06
Use consistent default weight

jordanschalm:
Modifies default initial weight value in flow-go to be consistent with FlowIDTableStaking smart contract. In general it is allowed for these values to differ, but:
- it is confusing when they are different for no reason - conceptually they should be equal
- being different can cause an edge case in cluster consensus (see https://github.com/dapperlabs/flow-go/issues/6435)


zhangchiqing (approved):

durkmurder (approved):
Nice

jordanschalm:
bors merge

jordanschalm:
bors retry

PR #3786, opened 2023-01-06
Simplify Start Entity Span.

pattyshack:
Remove isSample from Start`<Block|Collection|Transaction>`Span's return value. The span attribute value overhead is minimal.

Note that we usually set the same attribute values into the logger at the same time.  If the attribute value formatting is expensive, we can just precompute the formatted value and reuse it for both the span and the logger.


SaveTheRbtz (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3787, opened 2023-01-07
[CI] Add github action to build boot-tools on demand

peterargue:
The PR adds a github action workflow that can be run on demand to build the `bootstrap`, `transit` and `util` tools, then upload them to GCS

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -1,0 +1,44 @@
+name: Build Tools
+
+on:
+  workflow_dispatch:
+    inputs: 
+      tag:
+        description: 'Tagged commit to build tools against'
+        required: true
+        type: string
+
+jobs:
+  docker-push:
+    name: Push to container registry
+    runs-on: ubuntu-latest
+    steps:
+    - id: auth
+      uses: google-github-actions/auth@v1
+      with:
+        credentials_json: ${{ secrets.GCR_SERVICE_KEY }} # TODO: we need a new key to allow uploads
+    - name: Setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.19'
+    - name: Set up Google Cloud SDK
+      uses: google-github-actions/setup-gcloud@v1
+      with:
+        project_id: flow # TODO: is this correct?
+    - name: Checkout repo
+      uses: actions/checkout@v2
+      with:
+        ref: ${{ inputs.tag }}
+    - name: Build relic
+      run: make crypto_setup_gopath
+    - name: Build Tools
+      run: |
+        make tool-bootstrap
+        make tool-util
+        make tool-transit
+    - name: Upload boot-tools
+      run: |
+        mkdir boot-tools
+        mv bootstrap util transit boot-tools/
+        tar -czf boot-tools.tar ./boot-tools/
+        gsutil cp boot-tools.tar gs://flow-genesis-bootstrap/boot-tools.tar
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -35,21 +35,18 @@ jobs:
         ref: ${{ inputs.tag }}
     - name: Build relic
       run: make crypto_setup_gopath
-    - name: Build Tools
+    - name: Build and upload boot-tools
       run: |
-        make tool-bootstrap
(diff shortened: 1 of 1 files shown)


.github/workflows/tools.yml
```diff
@@ -0,0 +1,44 @@
...
+  workflow_dispatch:
+    inputs: 
+      tag:
+        description: 'Tagged commit to build tools against'
+        required: true
+        type: string
+
+jobs:
+  docker-push:
+    name: Push to container registry
```
zhangchiqing, line 13: I don't think we are pushing containers to registry, maybe we can rename it to `build-boot-tools`?

.github/workflows/tools.yml
```diff
@@ -0,0 +1,44 @@
...
+      with:
+        credentials_json: ${{ secrets.GCR_SERVICE_KEY }} # TODO: we need a new key to allow uploads
+    - name: Setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.19'
+    - name: Set up Google Cloud SDK
+      uses: google-github-actions/setup-gcloud@v1
+      with:
+        project_id: flow # TODO: is this correct?
```
zhangchiqing, line 27: cc @sjonpaulbrown to confirm

zhangchiqing (approved):

.github/workflows/tools.yml
```diff
@@ -0,0 +1,44 @@
...
+      run: |
+        make tool-bootstrap
+        make tool-util
+        make tool-transit
+    - name: Upload boot-tools
+      run: |
+        mkdir boot-tools
+        mv bootstrap util transit boot-tools/
+        tar -czf boot-tools.tar ./boot-tools/
+        gsutil cp boot-tools.tar gs://flow-genesis-bootstrap/boot-tools.tar
```
sjonpaulbrown, line 44: Could we update this to conditionally support test builds that are uploaded with an artifact name that uses the ref rather than `boot-tools.tar`?

If we also address the comment above, we could essentially use the following to conditionally set the name. This would allow us to kick of on-demand builds that would allow us to build the boot tools for testing purposes.

```
ARTIFACT_NAME=$([ "$INPUT_PRODBUILD" = "1" ] && echo "boot-tools.tar" || echo "$(git rev-parse --short HEAD).tar")
gsutil cp boot-tools.tar gs://flow-genesis-bootstrap/$ARTIFACT_NAME
```

.github/workflows/tools.yml
```diff
@@ -0,0 +1,44 @@
+name: Build Tools
+
+on:
+  workflow_dispatch:
+    inputs: 
+      tag:
+        description: 'Tagged commit to build tools against'
+        required: true
+        type: string
```
sjonpaulbrown, line 9:
```suggestion
    inputs: 
      tag:
        description: 'Tagged commit to build tools against'
        required: true
        type: string
      prodbuild:
        description: 'Execute production artifact build. Defaults to test build that will create artifact with short commit'
        type: string
        default: 0
```

.github/workflows/tools.yml
sjonpaulbrown: This is the correct project id.

.github/workflows/tools.yml
peterargue: how about it uploads to a tag specific directory by default, and there's an option to promote it to the official image
e.g.
* upload to `tools/v0.x.x/boot-tools.tar
* if "promote" is set, also upload to `./boot-tools.tar`

.github/workflows/tools.yml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+  build-publish:
+    name: Build boot tools
+    runs-on: ubuntu-latest
+    steps:
+    - id: auth
+      uses: google-github-actions/auth@v1
+      with:
+        credentials_json: ${{ secrets.GCR_SERVICE_KEY }} # TODO: we need a new key to allow uploads
```
peterargue, line 23: @sjonpaulbrown @gomisha could one of your help create and add a new secret for this usecase?

gomisha (approved):

peterargue:
bors merge

PR #3788, opened 2023-01-08
[Crypto] KeyGen improvement

tarakby:
So far, key generation implementation assumes the input seed has uniformly distributed entropy. This is only valid if function callers use the output of a secure RNG as a seed. 
This PR relaxes the strong seed assumption and implements further mechanism to extract the input entropy and use it in the rest of the process. A seed with enough entropy is now sufficient to generate keys.

- update BLS key generation seed handling by implementing the[ IETF draft algorithm](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-keygen) in section 2.3. 
   - this improves compatibility to IETF draft recommendations.
   - improves handling "bad" seeds with non uniform entropy.
   - a test against BLST key gen is added for compatibility check.
- update ECDSA key generation by using HKDF to extract entropy from input seed and expand it into key bytes. 
  - improves handling "bad" seeds with non uniform entropy.
- minor improvement by overwriting sensitive data in memory after computation.
- consolidate minimum seed length as a module constant instead of an algorithm specific length (consequence of the algorithm update) 
  
Side change:
 - add SHA2-256 light computation function (outside of the existing interface).


crypto/common.go
```diff
@@ -25,6 +26,16 @@ const (
...
+func overwrite(data []byte) {
+	_, err := rand.Read(data) // checking err is enough
+	if err != nil {
+		// zero the buffer if randomizing failed
+		for i := 0; i < len(data); i++ {
+			data[i] = 0
+		}
+	}
+}
+
```
SaveTheRbtz, line 38: is there any benefit in overwriting with random data instead of just writing zeros in the first place?

```suggestion
func overwrite(data []byte) {
	for i := range data {
		data[i] = 0
	}
}

```

BoringSSL simply memsets the memory on free:
```
void OPENSSL_cleanse(void *ptr, size_t len) {
#if defined(OPENSSL_WINDOWS)
  SecureZeroMemory(ptr, len);
#else
  OPENSSL_memset(ptr, 0, len);

#if !defined(OPENSSL_NO_ASM)
  /* As best as we can tell, this is sufficient to break any optimisations that
     might try to eliminate "superfluous" memsets. If there's an easy way to
     detect memset_s, it would be better to use that. */
  __asm__ __volatile__("" : : "r"(ptr) : "memory");
#endif
#endif  // !OPENSSL_NO_ASM
}
```

(with an additional caveat of preventing compiler optimizations.)

tarakby: Good question. First, this mitigation (either randomizing or zeroing) can only be helpful against "static" memory dump attacks that happen after the computation has ended. However, a more powerful attacker could use "dynamic" memory dumps during the computation to read sensitive data. In that case, the final "static" dump can help identify the memory areas to look at. Zeroing the sensitive areas gives an obvious hint for such attacks. 
Of course, an experimented attacker may avoid this mitigation by comparing multiple dynamic memory dumps and seeing the buffers evolution (especially that Golang initializes new allocated buffers with zeroes 😞 ). Randomizing instead of zeroing makes the attack slightly harder but not impossible. Randomizing is also a common practice in "grey box" cryptography implementations. 

Another limitation of overwriting buffers after use (either by randomizing or zeroing) is that it assumes the sensitive buffer was not copied in other areas. Such copies could have been made to the stack or registers, and these need to be overwritten as well. Looking at the assembly level for a specific compiler/architecture can answer the question, but that's not possible in our case (lib is written without a specific target compiler/arch). 
This mitigation is nice to have but I don't rely much on it, the library is still vulnerable to memory dumps and should only be run in safe environments.

Now back to the compiler optimizations: I am still wondering how to force the Go complier to not omit the call to `overwrite()`. Any thoughts about that?

SaveTheRbtz: We may be overthinking it too much.  Our goal is to do a due diligence so that the memory we free won't have secret material in it.  As long as we can check that memset is not optimized out by compiler, we should be fine. (otherwise we down the rabbithole of software memory enclaves, mlock, PROT_NONE, and other "fun" system things)

For the long term solution, a `runtime/secret` is probably the right way of solving it: https://github.com/golang/go/issues/21865

PS. As far as physical access goes, I think this is an uphill battle: w/o enabled IOMMU one has full memory access via the PCIe DMA.

tarakby: > For the long term solution, a runtime/secret is probably the right way of solving it

Nice, yes I can wait for that to be available in Go. Thanks!

> As far as physical access goes, I think this is an uphill battle.

Once our model includes physical access, our library is totally insecure, even without memory access (because of side channel leakages). In my comment above, I was only considering remote memory access.  

My overall opinion:
- this mitigation makes the lib more secure only in a "weak" attack model. In other models it is not very relevant IMHO. 
- It doesn't cost much to add it, so I decided to add a slightly stronger mitigation, that is commonly used in grey box crypto implementations (overthinking there is required and common).
- It is noticeable in security audits so it's good to add it for that purpose.

crypto/bls12381_utils.c
```diff
@@ -189,36 +189,51 @@ void bn_randZr_star(bn_t x) {
...
     bn_new_size(x, BITS_TO_DIGITS(Fr_BITS + SEC_BITS));
     bn_rand(x, RLC_POS, Fr_BITS + SEC_BITS);
-    bn_mod(x, x, r);
-    bn_free(r);
+    bn_mod(x, x, &core_get()->ep_r);
 }
 
 // reads a scalar from an array and maps it to Zr
+// the resulting scalar is in the range 0 <= a < r
+// len must be less than BITS_TO_BYTES(RLC_BN_BITS).
```
AlexHentschel, line 205: > the resulting scalar is in the range 0 <= a < r

is `a` the scalar? Hard to differentiate between the symbol and the english word "a". 


```suggestion
// reads a scalar from an array and maps it to Zr;
// the resulting scalar `a` is in the range 0 <= a < r;
// len must be less than BITS_TO_BYTES(RLC_BN_BITS).
```

crypto/bls12381_utils.c
```diff
@@ -189,36 +189,51 @@ void bn_randZr_star(bn_t x) {
...
+    bn_free(tmp);
+    if (bn_cmp_dig(a, 0) == RLC_EQ) {
+        return VALID;
+    }
+    return INVALID;
+}
+
+// reads a scalar from an array and maps it to Zr*
 // the resulting scalar is in the range 0 < a < r
 // len must be less than BITS_TO_BYTES(RLC_BN_BITS)
```
AlexHentschel, line 223:
```suggestion
// reads a scalar from an array and maps it to Zr*;
// the resulting scalar `a` is in the range 0 < a < r;
// len must be less than BITS_TO_BYTES(RLC_BN_BITS)
```

crypto/bls12381_utils.go
```diff
@@ -122,18 +122,14 @@ func randZrStar(x *scalar) {
...
-func mapToZrStar(x *scalar, src []byte) error {
-	if len(src) > maxScalarSize {
-		return invalidInputsErrorf(
-			"input slice length must be less than %d",
-			maxScalarSize)
-	}
-	C.bn_map_to_Zr_star((*C.bn_st)(x),
+// mapToZr reads a scalar from a slice of bytes and maps it to Zr
+// the resulting scalar is in the range 0 <= k < r.
+// It returns true if scalar is zero and false otherwise.
```
AlexHentschel, line 127:
```suggestion
// mapToZr reads a scalar from a slice of bytes and maps it to Zr.
// The resulting scalar `k` is in the range 0 <= k < r.
// It returns true if scalar is zero and false otherwise.
```

crypto/ecdsa.go
```diff
@@ -199,16 +195,34 @@ func goecdsaGenerateKey(c elliptic.Curve, seed []byte) *ecdsa.PrivateKey {
...
+
+	// use SHA2-256 as the building block H in HKDF
+	hashFunction := sha256.New
+	salt := []byte("") // HKDF salt
+	info := []byte("") // HKDF info
+	// use extra 128 bits to reduce the modular reduction bias
+	Nlen := bitsToBytes((a.curve.Params().N).BitLen())
+	okmLength := Nlen + (securityBits / 8)
+
+	// instanciate HKDF and extract okm
```
AlexHentschel, line 215:
```suggestion
	// instantiate HKDF and extract okm
```

AlexHentschel:
misc additional question regarding https://github.com/onflow/flow-go/blob/03ca88932ab719370034ae5887493d88d9f56150/crypto/hash/sha2.go#L9-L12
I didn't find a struct `commonHasher` in the code base. Is this outdated? Or is it just a typo, and the comment should read "... embeds common hasher" ?

crypto/hash/sha2.go
```diff
@@ -66,3 +66,14 @@ func (s *sha2_384Algo) ComputeHash(data []byte) Hash {
 func (s *sha2_384Algo) SumHash() Hash {
 	return s.Sum(nil)
 }
+
+// ComputeSHA2_256 computes the SHA2-256 (commonly known as SHA256)
+// digest of data and copies the result to the result buffer.
+//
+// The function is not part of the Hasher API. It is a light API
+// that allows a simple computation of a hash and minimizes
```
AlexHentschel, line 74:
```suggestion
// The function is not part of the Hasher API. It is pure function
// for simple hash computation with minimal heap allocation.
```

crypto/sign.go
```diff
@@ -104,13 +104,13 @@ func SignatureFormatCheck(algo SigningAlgorithm, s Signature) (bool, error) {
...
 //
+// The seed minimum length is 32 bytes and it should have enough entropy.
 // It is recommended to use a secure crypto RNG to generate the seed.
-// The seed must have enough entropy and should be sampled uniformly at random.
 //
 // The function returns:
 //   - (false, invalidInputsErrors) if the signing algorithm is not supported or
 //
-// if the seed length is not valid (too short or too long)
+// if the seed length is not valid (less than 32 bytes or larger than 256 bytes)
```
AlexHentschel, line 113: I feel like the formatting is a bit messed up here
```suggestion
//   - (false, invalidInputsErrors) if the signing algorithm is not supported or
//     the seed's length is invalid (less than 32 bytes or larger than 256 bytes)

```

crypto/sign.go
```diff
@@ -104,13 +104,13 @@ func SignatureFormatCheck(algo SigningAlgorithm, s Signature) (bool, error) {
...
 //   - (false, error) if an unexpected error occurs
 //   - (sk, nil) otherwise
```
AlexHentschel, line 115:
```suggestion
//   - (sk, nil) if key generation was successful
```

crypto/sign_test_utils.go
```diff
@@ -106,22 +106,49 @@ func testGenSignVerify(t *testing.T, salg SigningAlgorithm, halg hash.Hasher) {
 }
 
 func testKeyGenSeed(t *testing.T, salg SigningAlgorithm, minLen int, maxLen int) {
```
AlexHentschel, line 108: may I request just a very brief goDoc describing the objective of this test 🙏 🙇

AlexHentschel (approved):
Looks good to me (but don't have the background to assess the correctness of the low-level crypto implementation)

crypto/bls.go
```diff
@@ -250,30 +249,67 @@ func IsBLSSignatureIdentity(s Signature) bool {
...
+	// salt = H(UTF-8("BLS-SIG-KEYGEN-SALT-")) as per draft-irtf-cfrg-bls-signature-05 section 2.3.
+	saltString := "BLS-SIG-KEYGEN-SALT-"
+	hasher := hashFunction()
+	hasher.Write([]byte(saltString))
+	salt := make([]byte, hasher.Size())
+	hasher.Sum(salt[:0])
+
+	// L is the OKM length
+	// L = ceil((3 * ceil(log2(r))) / 16) which makes L (security_bits/8)-larger than r size
+	okmLength := (3 * PrKeyLenBLSBLS12381) / 2
```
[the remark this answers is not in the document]
tarakby: In theory we can move it yes. However, in the `const` section, I am only trying to keep the high level BLS design choices, rather than the smaller implementation details. I feel like it makes less sense for that specific constant to live far from where it's used.

SaveTheRbtz (approved):

tarakby:
bors merge

PR #3789, opened 2023-01-08
[Tools] Add bootstrap command to generate grpc TLS keys

peterargue:
This new command allows generating Access API TLS keys

### DIFF
--- a/cmd/bootstrap/cmd/access_keygen.go
+++ b/cmd/bootstrap/cmd/access_keygen.go
@@ -9,7 +9,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"math/big"
-	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -23,6 +23,7 @@ import (
 const certValidityPeriod = 100 * 365 * 24 * time.Hour // ~100 years
 
 var (
+	flagSANs           string
 	flagCommonName     string
 	flagNodeInfoFile   string
 	flagOutputKeyFile  string
@@ -44,6 +45,7 @@ func init() {
 	accessKeyCmd.Flags().StringVar(&flagOutputKeyFile, "key", "./access-tls.key", "path to output private key file")
 	accessKeyCmd.Flags().StringVar(&flagOutputCertFile, "cert", "./access-tls.crt", "path to output certificate file")
 	accessKeyCmd.Flags().StringVar(&flagCommonName, "cn", "", "common name to include in the certificate")
+	accessKeyCmd.Flags().StringVar(&flagSANs, "sans", "", "subject alternative names to include in the certificate, comma separated")
 }
 
 // accessKeyCmdRun generate an Access node TLS key and certificate
@@ -60,8 +62,12 @@ func accessKeyCmdRun(_ *cobra.Command, _ []string) {
 
 	if flagCommonName != "" {
 		log.Info().Msgf("using cn: %s", flagCommonName)
-		// needed to pass envoy's certificate validity check
 		certTmpl.Subject.CommonName = flagCommonName
+	}
+
+	if flagSANs != "" {
+		log.Info().Msgf("using SANs: %s", flagSANs)
+		certTmpl.DNSNames = strings.Split(flagSANs, ",")
 	}
 
 	cert, err := grpcutils.X509Certificate(networkKey, grpcutils.WithCertTemplate(certTmpl))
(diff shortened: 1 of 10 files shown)


cmd/bootstrap/cmd/access_keygen.go
```diff
@@ -0,0 +1,145 @@
...
+
+func init() {
+	rootCmd.AddCommand(accessKeyCmd)
+
+	accessKeyCmd.Flags().StringVar(&flagNodeInfoFile, "node-info", "", "path to node's node-info.priv.json file")
+	_ = accessKeyCmd.MarkFlagRequired("node-info")
+
+	accessKeyCmd.Flags().StringVar(&flagOutputKeyFile, "key", "./access-tls.key", "path to output private key file")
+	accessKeyCmd.Flags().StringVar(&flagOutputCertFile, "cert", "./access-tls.crt", "path to output certificate file")
+	accessKeyCmd.Flags().StringVar(&flagCommonName, "cn", "", "common name to include in the certificate")
```
SaveTheRbtz, line 46: fwiw: Common Name was deprecated by [rfc2818](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1) (May 2000).

cmd/bootstrap/cmd/access_keygen.go
```diff
@@ -0,0 +1,145 @@
...
+	var info bootstrap.NodeInfoPriv
+	err = json.Unmarshal(data, &info)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse private node info (path=%s): %w", nodeInfoPath, err)
+	}
+
+	return info.NetworkPrivKey.PrivateKey, nil
+}
+
+func writeFile(path string, data []byte) error {
```
SaveTheRbtz, line 110: nit: there is also a `io.WriteFile` or `os.WriteFile`

SaveTheRbtz (approved):

cmd/bootstrap/cmd/access_keygen.go
peterargue: 465383f

cmd/bootstrap/cmd/access_keygen.go
peterargue: added support for SANs c6614d6

cmd/bootstrap/cmd/access_keygen.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func loadNetworkKey(nodeInfoPath string) (crypto.PrivateKey, error) {
+	data, err := os.ReadFile(nodeInfoPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not read private node info (path=%s): %w", nodeInfoPath, err)
+	}
+
+	var info bootstrap.NodeInfoPriv
+	err = json.Unmarshal(data, &info)
```
tarakby, line 108: I'm a bit confused about the networking key curves in the protocol: 
- the function in this PR assumes there are on P-256 because I think it uses this [unmarshaling](https://github.com/onflow/flow-go/blob/master/model/encodable/keys.go#L92-L103). Are all access nodes keys on P-256? I think I remember a difference between staked and unstaked nodes but I may be wrong.
- Observer nodes [seem to use](https://github.com/onflow/flow-go/blob/master/cmd/observer/node_builder/observer_builder.go#L1069) secp256k1 (btw the function code and purpose seem very similar to what is in the PR).

cmd/bootstrap/cmd/access_keygen.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse private node info (path=%s): %w", nodeInfoPath, err)
+	}
+
+	return info.NetworkPrivKey.PrivateKey, nil
+}
+
+func defaultCertTemplate() (*x509.Certificate, error) {
+	bigNum := big.NewInt(1 << 62)
```
tarakby, line 117: what is the reason of keeping `sn` and `subjectSN` specifically on 62 bits?

peterargue: This was taken from the libp2p cert template: https://github.com/libp2p/go-libp2p/blob/master/p2p/security/tls/crypto.go#L250

I'm not totally sure but I suspect it doesn't matter.

cmd/bootstrap/cmd/access_keygen.go
peterargue: This uses the same logic ANs use to load the networking key:
https://github.com/onflow/flow-go/blob/master/cmd/scaffold.go#L1832

This is only intended for staked Access nodes since observers are untrusted and don't necessarily have a static network identity.

tarakby: Yes,  unstaked access nodes use a secp256k1 networking key: https://github.com/onflow/flow-go/blob/master/cmd/bootstrap/utils/key_generation.go#L52-L53 (unlike staked nodes)

tarakby (approved):

peterargue:
bors merge

PR #3790, opened 2023-01-09
[Access] Do not copy protobuf messages

SaveTheRbtz:
This is a cleanup for the switch to Protobuf v2 API.

* It is now forbidden to copy ptotobuf messages by value since they have [DoNotCopy](https://godoc.org/google.golang.org/protobuf/internal/pragma#DoNotCopy) embedded in them.  So switch to passing pointers.
* It is incorrect to compare protobuf messages by value w/ [DeepEqual](https://developers.google.com/protocol-buffers/docs/reference/go/faq#deepequal), so switched tests to use `go-cmp` and [protocmp.Transform()](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp#Transform)

ref: https://github.com/onflow/flow/pull/1215

### DIFF
--- a/engine/access/rpc/backend/backend_accounts.go
+++ b/engine/access/rpc/backend/backend_accounts.go
@@ -78,7 +78,7 @@ func (b *backendAccounts) getAccountAtBlockID(
 	blockID flow.Identifier,
 ) (*flow.Account, error) {
 
-	exeReq := execproto.GetAccountAtBlockIDRequest{
+	exeReq := &execproto.GetAccountAtBlockIDRequest{
 		Address: address.Bytes(),
 		BlockId: blockID[:],
 	}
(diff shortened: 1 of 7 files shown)


engine/access/rpc/backend/backend_accounts.go
```diff
@@ -111,7 +111,7 @@ func getAccountError(err error) error {
 	return status.Errorf(codes.Internal, "failed to get account from the execution node: %v", err)
 }
 
-func (b *backendAccounts) getAccountFromAnyExeNode(ctx context.Context, execNodes flow.IdentityList, req execproto.GetAccountAtBlockIDRequest) (*execproto.GetAccountAtBlockIDResponse, error) {
+func (b *backendAccounts) getAccountFromAnyExeNode(ctx context.Context, execNodes flow.IdentityList, req *execproto.GetAccountAtBlockIDRequest) (*execproto.GetAccountAtBlockIDResponse, error) {
```
pattyshack, line 114: nit: some newlines would make the method signature more readable

pattyshack (approved):

peterargue (approved):
thanks for cleaning this up!

SaveTheRbtz:
bors merge

PR #3792, opened 2023-01-10
Fix data race in DKG integration test

jordanschalm:
The DKG Client wrapper used in the dkg emulator test made use of an `enabled` flag which could be concurrently written by the test thread and read by the DKG controller thread, causing a data race.

### DIFF
--- a/integration/dkg/dkg_client_wrapper.go
+++ b/integration/dkg/dkg_client_wrapper.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"go.uber.org/atomic"
+
 	sdk "github.com/onflow/flow-go-sdk"
 
 	"github.com/onflow/flow-go/crypto"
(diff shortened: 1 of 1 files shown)


durkmurder:
Is there a data race or the value is not propagated to other thread?
From description I thought it's a data race, but atomic variable won't help in such case.

jordanschalm:
> Is there a data race or the value is not propagated to other thread?

The Go Race Detector detects it as a data race. The value might be written and read at the same time, because the value is not synchronized. Why would an atomic variable not address the data race here?

durkmurder:
> > Is there a data race or the value is not propagated to other thread?
> 
> The Go Race Detector detects it as a data race. The value might be written and read at the same time, because the value is not synchronized. Why would an atomic variable not address the data race here?

I see, in reality it's not exactly accurate. Atomic variable doesn't give any guarantees except values will be not read/written at the same time, so the update will be sequentially consistent. But this doesn't protect of ordering issues, meaning it will just order read/writes so they don't happen at the same time, but you don't get benefit that read will actually _wait_ for writer to finish. Meaning you don't get critical section on this code compared to mutexes. This might not be an issue here, just making sure we get the desired behavior.

durkmurder (approved):
Nice!

jordanschalm:
> But this doesn't protect of ordering issues, meaning it will just order read/writes so they don't happen at the same time, but you don't get benefit that read will actually wait for writer to finish. Meaning you don't get critical section on this code compared to mutexes.

Yeah, that is all we need here functionally. For concurrent reads and writes, it doesn't matter which happens first.

jordanschalm:
bors merge

PR #3793, opened 2023-01-10
Simplify block computer tracing

pattyshack:
1. Unconditionally trace block execution.  Block span data volume is low.  We don't need to sample at block level as long as we control the amount of span data at transaction level.
2. Skip creating spans for collection execution (we'll flatten the collection execution loop soon)
3. Create transaction execution trace spans under block root span instead of a separate transaction root span.  The transaction span data volume is controlled via sampling.


pattyshack:
hmm.  benchmark isn't happy with this change.  let me tweak it a bit

engine/execution/computation/computer/computer.go
```diff
@@ -241,27 +244,52 @@ func (e *blockComputer) getCollections(
 			isSystemCollection: true,
 		})
 
-	return collections, nil
+	shouldSample := e.tracer.ShouldSample(block.ID())
```
zhangchiqing, line 247:
```suggestion
	// sample the block. if the block is not sampled, sample the transaction
	shouldSample := e.tracer.ShouldSample(block.ID())
```

module/trace/noop.go
```diff
@@ -73,19 +76,21 @@ func (t *NoopTracer) StartTransactionSpan(
...
-	context.Context,
-) {
-	return NoopSpan, ctx
+) trace.Span {
+	return NoopSpan
 }
 
-func (t *NoopTracer) StartSpanFromParent(
+func (t *NoopTracer) ShouldSample(entityID flow.Identifier) bool {
+	return true
```
zhangchiqing, line 88: should it return `false` instead, since we don't want to trace any.

module/trace/trace.go
```diff
@@ -236,14 +225,28 @@ func (t *Tracer) StartSpanFromParent(
...
+	return entityID.IsSampled(t.sensitivity)
+}
+
+func (t *Tracer) StartSampledSpanFromParent(
+	parentSpan trace.Span,
+	entityID flow.Identifier,
+	operationName SpanName,
+	opts ...trace.SpanStartOption,
+) trace.Span {
+	if !entityID.IsSampled(t.sensitivity) {
```
zhangchiqing, line 243:
```suggestion
	if !ShouldSample(entityID) {
```

module/trace/log_tracer.go
```diff
@@ -62,6 +62,11 @@ func (t *LogTracer) Start(
 	return ctx, sp
 }
```
SaveTheRbtz, line 64: maybe we should just remove log tracer-- i haven't ever seen anyone use it and it's just slowly rotting away.

pattyshack: sure.  makes my life easier

module/trace/noop.go
pattyshack: it doesn't matter since the child span would be a no-op span either way

zhangchiqing (approved):

pattyshack:
as alexey pointed out, the perf regression is due to extra txn ID computation.  #3804 should address the regression.  I'm going to revert the block trace span back to not sampled.

the probability of sampling the block is 1 - ((1 -  1 / (1 << sensitivity)) ^ (# txns + 1)).  The probability approaches 1 pretty quickly.

at sensitivity = 4 (prod setting), with 20 txns (current load), block sampling prob is 72%

janezpodhostnik (approved):

pattyshack:
it's strange, but it looks like the regression is due to extra txn id calls within fvm rather at the computer ...

pattyshack:
will check again after merging #3819 ...

module/trace/trace.go
```diff
@@ -236,14 +225,28 @@ func (t *Tracer) StartSpanFromParent(
 	operationName SpanName,
 	opts ...trace.SpanStartOption,
 ) trace.Span {
-	if !IsSampled(parentSpan) {
```
pattyshack, line 239: I finally figured out the real regression.  Removing this caused the tracer to create a real span, but it doesn't show up on profile.

pattyshack: on the plus side, execution just got 1-2% faster cuz of the txn id pre-computations

pattyshack:
bors merge

PR #3796, opened 2023-01-11
Define vm interface in fvm instead of computer

pattyshack:
Note that we can't simply redefine fvm.VirtualMachine from a struct into an interface since that breaks the emulator.

### DIFF
--- a/fvm/fvm.go
+++ b/fvm/fvm.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/onflow/cadence/runtime"
-
 	"github.com/onflow/flow-go/fvm/derived"
 	"github.com/onflow/flow-go/fvm/environment"
 	errors "github.com/onflow/flow-go/fvm/errors"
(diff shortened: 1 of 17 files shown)


fvm/fvm.go
```diff
@@ -67,10 +65,14 @@ type Procedure interface {
...
 
-func NewInterpreterRuntime(config runtime.Config) runtime.Runtime {
-	return runtime.NewInterpreterRuntime(config)
+// VM runs procedures
+type VM interface {
+	Run(Context, Procedure, state.View) error
+	GetAccount(Context, flow.Address, state.View) (*flow.Account, error)
 }
 
+var _ VM = &VirtualMachine{}
```
zhangchiqing, line 74:
```suggestion
var _ VM = (*VirtualMachine)(nil)
```

zhangchiqing (approved):

janezpodhostnik (approved):

pattyshack:
bors merge

PR #3797, opened 2023-01-11
Change RegisterUpdates's return type

pattyshack:
return RegisterEntries instead of ([]RegisterID, []RegisterValue)

it's safer, and we can potentially skip sorting in the future.  It probably does less work as well.

### DIFF
--- a/engine/execution/state/state.go
+++ b/engine/execution/state/state.go
@@ -170,6 +170,21 @@ func RegisterValuesToValues(values []flow.RegisterValue) []ledger.Value {
 	return vals
 }
 
+func RegisterEntriesToKeysValues(
+	entries flow.RegisterEntries,
+) (
+	[]ledger.Key,
+	[]ledger.Value,
+) {
+	keys := make([]ledger.Key, len(entries))
+	values := make([]ledger.Value, len(entries))
+	for i, entry := range entries {
+		keys[i] = RegisterIDToKey(entry.Key)
+		values[i] = entry.Value
+	}
+	return keys, values
+}
+
 func LedgerGetRegister(ldg ledger.Ledger, commitment flow.StateCommitment) delta.GetRegisterFunc {
 
 	readCache := make(map[flow.RegisterID]flow.RegisterEntry)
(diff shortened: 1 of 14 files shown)


janezpodhostnik (approved):

engine/execution/state/state.go
```diff
@@ -213,17 +228,13 @@ func (s *state) NewView(commitment flow.StateCommitment) *delta.View {
 }
 
 type RegisterUpdatesHolder interface {
-	RegisterUpdates() ([]flow.RegisterID, []flow.RegisterValue)
+	RegisterUpdates() flow.RegisterEntries
 }
 
 func CommitDelta(ldg ledger.Ledger, ruh RegisterUpdatesHolder, baseState flow.StateCommitment) (flow.StateCommitment, *ledger.TrieUpdate, error) {
-	ids, values := ruh.RegisterUpdates()
+	keys, values := RegisterEntriesToKeysValues(ruh.RegisterUpdates())
```
zhangchiqing, line 235: This is the trade-off for the safety. We have to do additional loop to create the keys and values to create the ledger update. 

But I guess we could consider changing ledger update to be a list of key-value pairs as well, instead of a list of keys and a list of values?

zhangchiqing (approved):
Looks good, just a comment to discuss

engine/execution/state/state.go
pattyshack: This PR reduces the number of loops by 2 when compare to the original (one less loop in RegisterUpdates and one less loop in converting to ledger's keys/values)

I agree that we should convert ledger to accept key-value pair list as input as well, preferably RegisterEntries directly.  I decided against doing it in this PR to keep the PR simple.

pattyshack:
bors merge

pattyshack:
this breaks the emulator.  I have to add a new method to replace the old one =...(

SaveTheRbtz:
@j1010001 Who owns the emulator tests in flow-go codebase? maybe we can switch it to use a pinned container/binary instead of a source code.  Otherwise we'll always would need to deal with this circular dependency.

j1010001:
@SaveTheRbtz yes that sounds like a good idea - @sideninja, @janezpodhostnik what do you think ?

pattyshack:
i've renamed `RegisterUpdates` to `UpdatedRegisters`, and added an adaptor `RegisterUpdates` function to `Delta` to maintain emulator compatibility.

pattyshack:
bors merge

PR #3798, opened 2023-01-11
[CI] Remove --ssh flag from docker build for util

peterargue:
The `--ssh default` flag is only needed for building when the repo contains a dependency within a private repo. This is generally not needed so removing.

Running with this flag causes CI builds to fail: https://github.com/onflow/flow-go/actions/runs/3896458835/jobs/6653020995


zhangchiqing (approved):

sjonpaulbrown (approved):

peterargue:
bors merge

Issue #3597, opened 2022-11-18
Exposing spork ID as a Prometheus Metric


vishalchangrani:
## Problem Definition

What problem is this solving?

The Metrika node agent needs to identify which spork the node software is currently running on. If the node software is not on the current spork, then Metrika analytics can raise an alert and also stop reporting the node metrics.
However, the metrika node agent does not have access to the genesis data used by the node software. It can only query the Prometheus metrics and the node logs.

To allow for the Metrika node agent to detect the spork on which the node is running, a new metric needs to be introduced that reports the spork ID.

## Proposed Solution

What are the proposed solutions to this problem?

Introduce a new info metric to report the Spork ID. Here is an [example](https://github.com/prometheus/client_golang/blob/449b46435075e6e069e05af920fe028b941033cf/prometheus/example_metricvec_test.go#L56).

## Definition of Done

What tests must pass for this issue to be considered solved?
The new metrics gets reported when the node software is running.

This should be ideally included in the Jan 2023 spork so that by the time Metrika dashboard is made available to operators, this metric is available.

## Nice to have
If it will be great to also have the metric report other static information such as the git commit/tag of the node software.

AlexHentschel:
To the best of my knowledge, Prometheus only supports numerical metrics. [This stackoverflow thread](https://stackoverflow.com/questions/65850083/prometheus-java-client-export-string-based-metrics) is suggesting to labels to export string values
```
version_info{version="1.23", builtOn="Windows", built_by="myUserName" gitTag="version_1.0"} = 1
```

PR #3800, opened 2023-01-12
[General] Reporting node info metrics

durkmurder:
https://github.com/onflow/flow-go/issues/3597

### Context 

This PR implements reporting of static metrics that are common for each node type. 
Currently we report _sporkID_, _version_ and _commit_. 
Prometheus doesn't support reporting static metrics, the recommended way is to use `GaugeVec` with dummy data, the metrics itself are reported as gauge labels. 
Metrics are reported as part of startup process in scaffold code, this allows to utilize same logic for all nodes.

### DIFF
--- a/module/metrics/node_info.go
+++ b/module/metrics/node_info.go
@@ -1,9 +1,10 @@
 package metrics
 
 import (
+	"strconv"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"strconv"
 )
 
 // NodeInfoCollector implements metrics to report static information about node.
--- a/module/metrics/node_info.go
+++ b/module/metrics/node_info.go
@@ -1,0 +1,37 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// NodeInfoCollector implements metrics to report static information about node.
+// Such information can include: version, commit, sporkID.
+type NodeInfoCollector struct {
+	nodeInfo *prometheus.GaugeVec
+}
+
+const (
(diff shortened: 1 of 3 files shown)


module/metrics/node_info_test.go
```diff
@@ -0,0 +1,40 @@
...
+				}
+			}
+		}
+		assert.Failf(t, "metric not found", "except to find value %s", value)
+	}
+
+	for _, value := range []string{version, commit, sporkID} {
+		assertReported(value)
+	}
+	//t.Logf(proto.MarshalTextString(metricsFamilies[0]))
```
jordanschalm, line 39:
```suggestion
```

module/metrics/node_info.go
```diff
@@ -0,0 +1,36 @@
...
+// NodeInfoCollector implements metrics to report static information about node.
+// Such information can include: version, commit, sporkID.
+type NodeInfoCollector struct {
+	nodeInfo *prometheus.GaugeVec
+}
+
+const (
+	sporkIDLabel = "spork_id"
+	versionLabel = "version"
+	commitLabel  = "commit"
```
jordanschalm, line 17: We also have [`ProtocolVersion`](https://github.com/onflow/flow-go/blob/e250dcbb895ce604acc2a535816c647af879d9d2/state/protocol/params.go#L45-L47), which is an integer version number meant to map to the major version for the spork. 

We set this value when bootstrapping sporks now, but don't really use it otherwise. I think this would be a good place to expose it. Then, Metrika could potentially directly compare `version.MajorVersion()` to `ProtocolVersion`, rather than maintaining a separate mapping of `SporkID -> SporkVersionRange`.

jordanschalm (approved):

vishalchangrani (approved):

durkmurder:
bors merge

PR #3820, opened 2023-01-16
Update view's register ids apis.

pattyshack:
1. rename AllRegisters to AllRegisterIDs
2. added un-sorted UpdatedRegisterIDs
3. switch meter param override invalidation over to UpdatedRegisterIDs


engine/execution/state/delta/delta.go
```diff
@@ -36,6 +36,15 @@ func (d Delta) Set(owner, key string, value flow.RegisterValue) {
 	d.Data[k] = value
 }
 
+// UpdatedRegisterIDs returns all register ids that were updated by this delta.
+func (d Delta) UpdatedRegisterIDs() []flow.RegisterID {
```
janezpodhostnik, line 40: For certain use cases it is very important that we know that these are not deterministically sorted.
Could you add to the comment that the order of these is not deterministic.

janezpodhostnik (approved):

engine/execution/state/delta/delta.go
pattyshack: good point.  i'm thinking of making fvm/state/View interface the primary view interface.  i'll add the comment there

pattyshack: shouldn't the sorting be done within the ledger / NewUpdate in that case?

fvm/state/view.go
```diff
@@ -8,8 +8,19 @@ type View interface {
...
+	// UpdatedRegisters returns all registers that were updated by this view.
+	// The returned entries are sorted by ids.
 	UpdatedRegisters() flow.RegisterEntries
-	AllRegisters() []flow.RegisterID
+
+	// UpdatedRegisterIDs returns all register ids that were updated by this
+	// view.  The returned ids are unsorted.
+	UpdatedRegisterIDs() []flow.RegisterID
+
+	// UpdatedRegisterIDs returns all register ids that were touched by this
```
SaveTheRbtz, line 20: nit: comment

fvm/state/state.go
```diff
@@ -146,7 +146,12 @@ func (s *State) InteractionUsed() uint64 {
 	return s.meter.TotalBytesOfStorageInteractions()
 }
 
-// UpdatedRegisters returns the lists of register id / value that were updated.
+// UpdatedRegisters returns the lists of register ids that were updated.
```
SaveTheRbtz, line 149: nit: comment

engine/execution/state/delta/delta.go
zhangchiqing: We could. We need to make the returned slice to be deterministic, so that there is no other side effect. And I just wanted to point out your change can't ensure the order is deterministic.

SaveTheRbtz (approved):

pattyshack:
bors merge

PR #3828, opened 2023-01-18
Add sandboxnet missing keys and upgrade contract

janezpodhostnik:
The FlowFees contract is from core contracts version c1b31d5a4722aba3bc4ccaae9a7ec6e660887591

This adds missing keys to 3 core accounts and upgrades the FlowFees core contract.

### DIFF
--- a/cmd/util/ledger/migrations/add_missing_keys_migration.go
+++ b/cmd/util/ledger/migrations/add_missing_keys_migration.go
@@ -1,0 +1,140 @@
+package migrations
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-core-contracts/lib/go/templates"
+	"github.com/onflow/flow-go/fvm"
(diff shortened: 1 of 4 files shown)


pattyshack (commented):
This should go into v0.29 instead?

Maybe have someone more familiar with the system contracts look at this as well.

cmd/util/ledger/migrations/add_missing_keys_migration.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+}
+
+func addMissingKeysAndUpgradeContractsMigration(
+	chain flow.Chain,
+	payloads []ledger.Payload,
+	logger zerolog.Logger,
+) ([]ledger.Payload, error) {
+
+	// only run this migration on sandboxnet and benchnet
```
pattyshack, line 39: this will error if you try to run migrate on other chain.  maybe change the migrate list in execution_state_extract.go instead.

janezpodhostnik:
yes, this should be on v0.29, moving to new PR

PR #3830, opened 2023-01-18
Add sandboxnet missing keys and upgrade contract

janezpodhostnik:
The FlowFees contract is from core contracts version c1b31d5a4722aba3bc4ccaae9a7ec6e660887591

This adds missing keys to 3 core accounts and upgrades the FlowFees core contract.

This should not be merged, and should fail anywhere but Sandboxnet (or benchnet, since that is where its being tested)

### DIFF
--- a/cmd/util/cmd/execution-state-extract/execution_state_extract.go
+++ b/cmd/util/cmd/execution-state-extract/execution_state_extract.go
@@ -7,6 +7,7 @@ import (
 	"github.com/rs/zerolog"
 	"go.uber.org/atomic"
 
+	mig "github.com/onflow/flow-go/cmd/util/ledger/migrations"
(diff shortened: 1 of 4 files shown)


cmd/util/cmd/execution-state-extract/execution_state_extract.go
```diff
@@ -90,6 +91,7 @@ func extractExecutionState(
 		migrations = []ledger.Migration{
```
pattyshack, line 91: as pointed out previously, creating the migration list unconditionally meant that running migration for non sandbox / benchnet chain will fail.  maybe create different migration list based on the chain rather returning error in addMissingKeysAndUpgradeContractsMigration, line 42.

janezpodhostnik:
this was applied

PR #3831, opened 2023-01-19
Khalil/1844 topic validator skip decoding

kc1116:
This PR optimizes message payload decoding for the happy path by removing message payload decoding from the TopicValidator and decoding the payload in the middleware instead. The authorized sender validator will assume the message code byte encoded in the message payload is correct, messages with invalid message codes will be dropped when decoding fails in the middleware before the IncomingMessageScope is passed to the Overlay.


network/validator/validator.go
```diff
@@ -3,14 +3,15 @@ package validator
...
 
+	"github.com/onflow/flow-go/network/message"
 	"github.com/onflow/flow-go/network/p2p"
 )
 
 // MessageValidator validates the given message with original sender `from` and returns an error if validation fails
 // else upon successful validation it should return the decoded message type string.
 // Note: contrarily to pubsub.ValidatorEx, the peerID parameter does not represent the bearer of the message, but its source.
-type MessageValidator func(from peer.ID, msg interface{}) (string, error)
+type MessageValidator func(from peer.ID, msg *message.Message) (string, error)
```
yhassanzadeh13, line 13: I've found no usage for the `MessageValidator` type. If we are not planning to use it, we may remove it.

network/validator/authorized_sender_validator.go
```diff
@@ -105,16 +105,24 @@ func (av *AuthorizedSenderValidator) Validate(from peer.ID, msg interface{}, cha
 //
 // Expected error returns during normal operations:
 //   - ErrSenderEjected: if identity of sender is ejected from the network
-//   - message.ErrUnknownMsgType if message auth config us not found for the msg
+//   - codec.ErrUnknownMsgCode: if interface for the encoded message code byte is unknown
+//   - message.UnknownMsgTypeErr if message auth config us not found for the msg
 //   - message.ErrUnauthorizedMessageOnChannel if msg is not authorized to be sent on channel
 //   - message.ErrUnauthorizedRole if sender role is not authorized to send msg
-func (av *AuthorizedSenderValidator) isAuthorizedSender(identity *flow.Identity, channel channels.Channel, msg interface{}, protocol message.Protocol) (string, error) {
+func (av *AuthorizedSenderValidator) isAuthorizedSender(identity *flow.Identity, channel channels.Channel, msgCode uint8, protocol message.Protocol) (string, error) {
```
yhassanzadeh13, line 112: Instead of using `uint8` variables to represent message codes throughout the codebase, let's define the `MessageCode` type. This will make it clear to anyone reading the code that a variable of this type represents a message code, and will also prevent any potential bugs caused by accidentally using a non-message code value. In this way, we take advantage of the type system in Go to improve the readability and maintainability of the code.

```go
type MessageCode uint8
```

network/validator/authorized_sender_validator.go
```diff
@@ -105,16 +105,24 @@ func (av *AuthorizedSenderValidator) Validate(from peer.ID, msg interface{}, cha
...
 	if identity.Ejected {
 		return "", ErrSenderEjected
 	}
 
+	// attempt to get the message interface from the message code encoded into the first byte of the message payload
+	// this will be used to get the message auth configuration.
+	iFace, _, err := codec.InterfaceFromMessageCode(msgCode)
```
yhassanzadeh13, line 119:
```suggestion
	msgInterface, _, err := codec.InterfaceFromMessageCode(msgCode)
```

network/validator/authorized_sender_validator.go
```diff
@@ -105,16 +105,24 @@ func (av *AuthorizedSenderValidator) Validate(from peer.ID, msg interface{}, cha
...
+	if err != nil {
+		return "", err
```
yhassanzadeh13, line 121:
```suggestion
		return "", fmt.Errorf("could not extract interface from message code %v: %w", msgCode, err)
```

network/validator/authorized_sender_validator.go
```diff
@@ -105,16 +105,24 @@ func (av *AuthorizedSenderValidator) Validate(from peer.ID, msg interface{}, cha
...
+	}
+
 	// get message auth config
-	conf, err := message.GetMessageAuthConfig(msg)
+	conf, err := message.GetMessageAuthConfig(iFace)
 	if err != nil {
 		return "", err
```
yhassanzadeh13, line 127:
```suggestion
		return "", fmt.Errorf("could not get authorization config for interface %T: %w", iFace, err)
```

yhassanzadeh13 (commented):
Overall the PR looks good, thanks for consolidating the message validation logic. Added a few comments.

network/p2p/middleware/middleware.go
```diff
@@ -693,7 +667,7 @@ func (m *Middleware) processUnicastStreamMessage(remotePeer peer.ID, msg *messag
 
 	// if message channel is not public perform authorized sender validation
 	if !channels.IsPublicChannel(channel) {
-		_, err := m.authorizedSenderValidator.Validate(remotePeer, decodedMsgPayload, channel, message.ProtocolUnicast)
+		messageType, err := m.authorizedSenderValidator.Validate(remotePeer, msg.Payload[0], channel, message.ProtocolUnicast)
```
peterargue, line 670: same comment about not accessing the `Payload` slice directly

network/p2p/middleware/middleware.go
```diff
@@ -726,6 +700,25 @@ func (m *Middleware) processAuthenticatedMessage(msg *message.Message, decodedMs
...
+	}
+	if codec.IsErrMsgUnmarshal(err) || codec.IsErrInvalidEncoding(err) {
+		// slash if peer sent a message that could not be marshalled into the message type denoted by the message code byte
+		violation := &slashing.Violation{
+			PeerID: peerID.String(), Channel: channel, Protocol: message.ProtocolUnicast, Err: err,
+		}
+		m.slashingViolationsConsumer.OnInvalidMsgError(violation)
+		return
+	}
+
```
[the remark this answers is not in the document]
peterargue: agreed. this was missed during the refactor. the original version on line 666 had the check

network/validator/authorized_sender_validator.go
```diff
@@ -39,8 +40,8 @@ func NewAuthorizedSenderValidator(log zerolog.Logger, slashingViolationsConsumer
 
 // PubSubMessageValidator wraps Validate and returns PubSubMessageValidator callback that returns pubsub.ValidationReject if validation fails and pubsub.ValidationAccept if validation passes.
 func (av *AuthorizedSenderValidator) PubSubMessageValidator(channel channels.Channel) PubSubMessageValidator {
-	return func(from peer.ID, msg interface{}) p2p.ValidationResult {
-		_, err := av.Validate(from, msg, channel, message.ProtocolPublish)
+	return func(from peer.ID, msg *message.Message) p2p.ValidationResult {
+		_, err := av.Validate(from, msg.Payload[0], channel, message.ProtocolPublish)
```
peterargue, line 44: same comment about not indexing into `Payload`

peterargue (changes requested):
overall this looks pretty good. There are a few places where we're accessing the message code directly without size checks. This introduces a vulnerability where a nil payload can cause the node to crash.

We should instead create a helper method on the codec that extracts and returns the code with proper checking.

network/codec/codes.go
```diff
@@ -68,75 +68,75 @@ func MessageCodeFromInterface(v interface{}) (uint8, string, error) {
 	switch v.(type) {
 	// consensus
 	case *messages.BlockProposal:
-		return CodeBlockProposal, "CodeBlockProposal", nil
+		return CodeBlockProposal, "message.BlockProposal", nil
```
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/520ed98b68df5c130adeb4db5f52831724961fda

network/codec/codes.go
```diff
@@ -151,75 +151,75 @@ func InterfaceFromMessageCode(code uint8) (interface{}, string, error) {
 	switch code {
 	// consensus
 	case CodeBlockProposal:
-		return &messages.BlockProposal{}, "BlockProposal", nil
+		return &messages.BlockProposal{}, "messages.BlockProposal", nil
```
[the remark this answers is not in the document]
kc1116: Fixed in the same commit as comment above https://github.com/onflow/flow-go/pull/3831/commits/520ed98b68df5c130adeb4db5f52831724961fda

network/p2p/middleware/middleware.go
```diff
@@ -726,6 +700,25 @@ func (m *Middleware) processAuthenticatedMessage(msg *message.Message, decodedMs
 		return
 	}
 
+	channel := channels.Channel(msg.ChannelID)
+	decodedMsgPayload, err := m.codec.Decode(msg.Payload)
+	if codec.IsErrUnknownMsgCode(err) {
+		// slash peer if message contains unknown message code byte
+		violation := &slashing.Violation{
+			PeerID: peerID.String(), Channel: channel, Protocol: message.ProtocolUnicast, Err: err,
+		}
```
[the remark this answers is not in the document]
kc1116: 1. https://github.com/onflow/flow-go/pull/3831/commits/36c277b5a7a59f1429f5b1e52fe0afce94bc0346
2. https://github.com/onflow/flow-go/pull/3831/commits/d51b48106798badc918c07afc243b1f264abc050

network/p2p/middleware/middleware.go
```diff
@@ -706,14 +680,14 @@ func (m *Middleware) processUnicastStreamMessage(remotePeer peer.ID, msg *messag
...
 
 // processAuthenticatedMessage processes a message and a source (indicated by its peer ID) and eventually passes it to the overlay
 // In particular, it populates the `OriginID` field of the message with a Flow ID translated from this source.
 // The assumption is that the message has been authenticated at the network level (libp2p) to originate from the peer with ID `peerID`
 // this requirement is fulfilled by e.g. the output of readConnection and readSubscription
-func (m *Middleware) processAuthenticatedMessage(msg *message.Message, decodedMsgPayload interface{}, peerID peer.ID, protocol network.ProtocolType) {
+func (m *Middleware) processAuthenticatedMessage(msg *message.Message, peerID peer.ID, protocol network.ProtocolType) {
 	originId, err := m.idTranslator.GetFlowID(peerID)
 	if err != nil {
 		// this error should never happen. by the time the message gets here, the peer should be
```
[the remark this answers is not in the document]
kc1116: @peterargue The inaccuracy lies in the fact that originID is not always the origin of the message but could be the source (relayed from another peer) https://github.com/onflow/flow-go/pull/3831/commits/d51b48106798badc918c07afc243b1f264abc050

network/p2p/middleware/middleware.go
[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/fdbd9f754567fdca53e9ef3252c05062ee37c5fa

network/validator/validator.go
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/299d371c6b78472e077c28d82d4b90a385ec5954

network/validator/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/938612350daa63faf74941be04e42211fc6c9a7e

network/validator/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/2a28965e3c2566b9174c1f01959ff829226a76b1

network/validator/authorized_sender_validator.go
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/4616da0d6e5b39e55b5b1b51efa8c50a7ff36f39

network/p2p/middleware/middleware.go
kc1116: https://github.com/onflow/flow-go/pull/3831/commits/4616da0d6e5b39e55b5b1b51efa8c50a7ff36f39

yhassanzadeh13 (approved):
Looks great 👏 thanks for applying the comments.

network/codec/codes.go
```diff
@@ -147,81 +154,96 @@ func MessageCodeFromInterface(v interface{}) (uint8, string, error) {
 // of the message code represents.
 // Expected error returns during normal operations:
 //   - ErrUnknownMsgCode if message code does not match any of the configured message codes above.
-func InterfaceFromMessageCode(code uint8) (interface{}, string, error) {
+func InterfaceFromMessageCode(code MessageCode) (interface{}, string, error) {
 	switch code {
 	// consensus
 	case CodeBlockProposal:
-		return &messages.BlockProposal{}, "BlockProposal", nil
+		return &messages.BlockProposal{}, what(messages.BlockProposal{}), nil
```
peterargue, line 161: does it matter that these will return `messages.BlockProposal` instead of the actual type `*messages.BlockProposal`?

network/p2p/middleware/middleware.go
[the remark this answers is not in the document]
peterargue: My understanding is that the peer reported by the [topic validator](https://github.com/onflow/flow-go/blob/master/network/validator/pubsub/topic_validator.go#L87) was the original sender of the message. This is confirmed with checks against the signing public key [here](https://github.com/onflow/flow-go/blob/master/network/validator/pubsub/topic_validator.go#L38-L57). So by the time we get to `processAuthenticatedMessage`, `peerID` is guaranteed to be the sender.

Am I missing something?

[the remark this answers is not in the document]
yhassanzadeh13: @peterargue is correct in stating that the peer.ID in a signed message refers to the node that has signed the message, which is the original sender, and not the relayer. There appears to be some confusion regarding the role of nodes in relaying messages _as pieces of evidence_ (e.g., slashing), which is outside the scope of the networking layer and requires the relaying node to create a new message and sign it, which in turn the relaying node becomes the original sender. However, relaying a message (as a piece of evidence) is an application layer protocol and is built on top of networking primitives without interfering with them. It is important to note that even after a peer is ejected from the network, its messages may still be routed to Flow nodes through malicious routers, and in such cases, the original comment above is no longer applicable: `// this error should never happen. by the time the message gets here, the peer should be...`

peterargue (approved):
looks good!

network/p2p/middleware/middleware.go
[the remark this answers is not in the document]
peterargue: we discussed offline to make sure authentication is working as expected. for posterity:
* at this point in processing, `peerID` is verified to be the original sender (node that signed the original message).
* the original sender has been checked against the identity list and verified to be part of the network, not ejected and authorized to send the received message type.
* there's a caveat for the public (unstaked) network where all participants are allowed to communicate

network/codec/codes.go
kc1116: No because this value is only used for logs

PR #3832, opened 2023-01-19
Log invalid vote in bootstrap finalize

zhangchiqing:
See comments.

If we fail to construct QC, we will log the error and also log invalid votes if there is any.
If we succeeded to construct a QC, we will also log invalid votes if there is any.

### DIFF
--- a/cmd/bootstrap/cmd/qc.go
+++ b/cmd/bootstrap/cmd/qc.go
@@ -20,7 +20,13 @@ func constructRootQC(block *flow.Block, votes []*model.Vote, allNodes, internalN
 		log.Fatal().Err(err).Msg("failed to generate QC participant data")
 	}
 
-	qc, err := run.GenerateRootQC(block, votes, participantData, identities)
+	qc, invalidVotesErr, err := run.GenerateRootQC(block, votes, participantData, identities)
+	if len(invalidVotesErr) > 0 {
+		for _, err := range invalidVotesErr {
+			log.Warn().Err(err).Msg("invalid vote")
+		}
+	}
+
 	if err != nil {
 		log.Fatal().Err(err).Msg("generating root QC failed")
 	}
--- a/cmd/bootstrap/run/qc.go
+++ b/cmd/bootstrap/run/qc.go
@@ -44,11 +44,18 @@ func (pd *ParticipantData) Identities() flow.IdentityList {
 // which assumes that only consensus participants construct QCs, which also have produce votes.
 //
 // TODO: modularize QC construction code (and code to verify QC) to be instantiated without needing private keys.
-func GenerateRootQC(block *flow.Block, votes []*model.Vote, participantData *ParticipantData, identities flow.IdentityList) (*flow.QuorumCertificate, error) {
+// It returns (qc, nil, nil) if a QC can be constructed with enough votes, and there is no invalid votes
+// It returns (qc, invalidVotes, nil) if there are some invalid votes, but a QC can still be constructed
+// It returns (nil, invalidVotes, err) if no qc can be constructed with not enough votes or running any any exception
+func GenerateRootQC(block *flow.Block, votes []*model.Vote, participantData *ParticipantData, identities flow.IdentityList) (
+	*flow.QuorumCertificate, // the constructed QC
+	[]error, // return invalid votes error
+	error, // exception or could not construct qc
+) {
 	// create consensus committee's state
 	committee, err := committees.NewStaticCommittee(identities, flow.Identifier{}, participantData.Lookup, participantData.GroupKey)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	// STEP 1: create VoteProcessor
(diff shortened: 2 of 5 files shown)


cmd/bootstrap/cmd/qc.go
```diff
@@ -20,7 +20,13 @@ func constructRootQC(block *flow.Block, votes []*model.Vote, allNodes, internalN
 		log.Fatal().Err(err).Msg("failed to generate QC participant data")
 	}
 
-	qc, err := run.GenerateRootQC(block, votes, participantData, identities)
+	qc, invalidVotesErr, err := run.GenerateRootQC(block, votes, participantData, identities)
+	if len(invalidVotesErr) > 0 {
+		for _, err := range invalidVotesErr {
+			log.Warn().Err(err).Msg("invalid vote")
```
zhangchiqing, line 26: Will print like this even if a QC is built.

image

cmd/bootstrap/run/qc.go
```diff
@@ -45,11 +45,17 @@ func (pd *ParticipantData) Identities() flow.IdentityList {
...
 //
 // TODO: modularize QC construction code (and code to verify QC) to be instantiated without needing private keys.
-func GenerateRootQC(block *flow.Block, votes []*model.Vote, participantData *ParticipantData, identities flow.IdentityList) (*flow.QuorumCertificate, error) {
+// It returns (qc, nil, nil) if a QC can be constructed with enough votes, and there is no invalid votes
+// It returns (qc, invalidVotes, nil) if there are some invalid votes, but a QC can still be constructed
+// It returns (nil, invalidVotes, err) if no qc can be constructed with not enough votes or running any any exception
+func GenerateRootQC(block *flow.Block, votes []*model.Vote, participantData *ParticipantData, identities flow.IdentityList) (
+	*flow.QuorumCertificate, // the constructed QC
+	[]error, // return invalid votes error
+	error) { // exception or could not construct qc
```
jordanschalm, line 54:
```suggestion
	error,  // exception or could not construct qc
) {
```
nitpick: I find these much easier to read when there is a more visible distinction between the return variable line and the first line of the function.

cmd/bootstrap/run/qc_test.go
```diff
@@ -28,8 +28,32 @@ func TestGenerateRootQC(t *testing.T) {
...
+func TestGenerateRootQCWithSomeInvalidVotes(t *testing.T) {
+	participantData := createSignerData(t, 10)
+
+	block := unittest.BlockFixture()
+	block.Payload.Guarantees = nil
+	block.Payload.Seals = nil
+	block.Header.Height = 0
+	block.Header.ParentID = flow.ZeroID
+	block.Header.View = 3
+	block.Header.PayloadHash = block.Payload.Hash()
```
jordanschalm, line 45:
```suggestion
	block := unittest.GenesisFixture()
```

Could we do this instead? I think everything is the same except the view.

jordanschalm (approved):

cmd/bootstrap/cmd/qc.go
```diff
@@ -20,7 +20,13 @@ func constructRootQC(block *flow.Block, votes []*model.Vote, allNodes, internalN
...
+		}
+	}
```
sjonpaulbrown, line 28: Since we do not intend to fail when invalid votes are used, could we move this logic to `GenerateRootQC`? If move the logic down, we won't need to update the function signature and return multiple types of errors. Since we are logging them immediately, we could log them individually after evaluating each vote.

cmd/bootstrap/run/qc.go
```diff
@@ -59,34 +66,44 @@ func GenerateRootQC(block *flow.Block, votes []*model.Vote, participantData *Par
...
 				vote.SignerID,
 				err)
 		}
 	}
 
 	if createdQC == nil {
-		return nil, fmt.Errorf("QC is not created, total number of votes %v, expect to have 2/3 votes of %v participants",
+		return nil, invalidVotes, fmt.Errorf("QC is not created, total number of votes %v, expect to have 2/3 votes of %v participants",
 			len(votes), len(identities))
 	}
```
sjonpaulbrown, line 97: If we decide to log the invalid vote errors rather than returning `[]error`, we could shift this logic to separate vote processing function to add a unit test for it.

cmd/bootstrap/cmd/qc.go
zhangchiqing: I tried, we could log them inside GenerateRootQC, but the logger is not the same logger as being used by the bootstrap cmd. Therefore, in ansible script, it won't print that log if there is invalid vote, which might seem as if there is no invalid vote. 

To be safe, I decided to return invalid votes as errors, and use the same logger to print log and invalid votes, it will guarantee that the log for invalid votes will be printed along with other info logs.

sjonpaulbrown (approved):

zhangchiqing:
bors merge

zhangchiqing:
bors merge

zhangchiqing:
bors retry

zhangchiqing:
bors merge

zhangchiqing:
bors retry

PR #3841, opened 2023-01-20
Remove updatedAddresses from fvm state

pattyshack:
Use view's updated register ids as canonical source


pattyshack:
hmm... i've changed on mind on this.  it's not helpful

PR #3842, opened 2023-01-20
Remove Touch/Delete from view interface.

pattyshack:
This simplifies changing view implementations.  We can add them back in the future if the support is needed.

### DIFF
--- a/engine/execution/state/state_test.go
+++ b/engine/execution/state/state_test.go
@@ -178,7 +178,7 @@ func TestExecutionStateWithTrieStorage(t *testing.T) {
 
 		// update value and get resulting state commitment
 		view2 := es.NewView(sc2)
-		err = view2.Delete(registerID1, "")
+		err = view2.Set(registerID1, "", nil)
 		assert.NoError(t, err)
 
 		sc3, _, err := state.CommitDelta(l, view2.Delta(), sc2)
--- a/fvm/environment/accounts_test.go
+++ b/fvm/environment/accounts_test.go
@@ -144,29 +144,6 @@ func TestAccounts_GetPublicKeys(t *testing.T) {
 	})
 }
 
-// Some old account could be created without key count register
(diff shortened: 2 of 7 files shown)


fvm/environment/accounts_test.go
```diff
@@ -144,29 +144,6 @@ func TestAccounts_GetPublicKeys(t *testing.T) {
 	})
 }
 
-// Some old account could be created without key count register
```
pattyshack, line 147: @janezpodhostnik could you double check if this is still needed?  we should migrate old accounts / populate the public_key_count field instead of preserving this behaviour.

SaveTheRbtz (approved):

fvm/environment/accounts_test.go
janezpodhostnik: The key count register no longer exists it has been merged into the account status register https://github.com/onflow/flow-go/pull/2799. So this is no longer needed.

engine/execution/state/state_test.go
```diff
@@ -178,7 +178,7 @@ func TestExecutionStateWithTrieStorage(t *testing.T) {
 
 		// update value and get resulting state commitment
 		view2 := es.NewView(sc2)
-		err = view2.Delete(registerID1, "")
+		err = view2.Set(registerID1, "", []byte("orange"))
```
janezpodhostnik, line 181: How come this test wasn't changed to test `view2.Set(registerID1, "", nil)`?

janezpodhostnik (approved):

engine/execution/state/state_test.go
pattyshack: I wasn't sure if cadence is allowed to set the value to nil, so I picked an arbitrary non-nil value instead.  I'll change this to nil.

pattyshack:
bors merge

PR #3846, opened 2023-01-23
Use consistent method for creating temp dirs

jordanschalm:
This PR changes integration test utilities to use a consistent method and root location when creating temporary directories for use in tests. 

This is to mitigate an issue where integration tests would fail on macOS, due to a permissions issue when initializing bootstrap and data directories for Docker containers.
```
lchown /var/folders/bm/wbf3kfs91911lr5w6qyw126c0000gn/T/TestEpochJoinAndLeaveANTestEpochJoinAndLeaveAN4216456854/001/bootstrap/execution-state/00000000: operation not permitted
```

As far as I can tell, `t.TempDir()` created directories which caused `operation not permitted` errors, whereas `os.MkdirTemp("/tmp", ...)` did not. We aren't sure what causes the two methods to behave differently at this point, but using "the good one" consistently works to fix the test failure.

### DIFF
--- a/integration/testnet/network.go
+++ b/integration/testnet/network.go
@@ -578,9 +578,15 @@ func WithAdditionalFlag(flag string) func(config *NodeConfig) {
 	}
 }
 
-// integrationBootstrapDir creates a temporary directory at /tmp/flow-integration-bootstrap
-func integrationBootstrapDir() (string, error) {
-	return os.MkdirTemp(TmpRoot, integrationBootstrap)
+// tempDir creates a temporary directory at /tmp/flow-integration-bootstrap
+func tempDir(t *testing.T) string {
+	dir, err := os.MkdirTemp(TmpRoot, integrationBootstrap)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		err := os.RemoveAll(dir)
+		require.NoError(t, err)
+	})
+	return dir
 }
 
 func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig, chainID flow.ChainID) *FlowNetwork {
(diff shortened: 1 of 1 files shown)


integration/testnet/network.go
```diff
@@ -609,10 +615,13 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig, chainID flow.Ch
 	})
 
 	// create a temporary directory to store all bootstrapping files
-	bootstrapDir, err := integrationBootstrapDir()
-	require.Nil(t, err)
+	bootstrapDir := tempDir(t)
 
 	t.Logf("BootstrapDir: %s \n", bootstrapDir)
+	t.Cleanup(func() {
```
m4ksio, line 621: Do we need this since we add clean up in  https://github.com/onflow/flow-go/pull/3846/files#diff-7540be57ca1fb2714cfce630ad0eba7bfc97362d8e30b79ac2551dc68792e75bR585 ?

m4ksio (approved):

integration/testnet/network.go
```diff
@@ -609,10 +615,13 @@ func PrepareFlowNetwork(t *testing.T, networkConf NetworkConfig, chainID flow.Ch
...
+		err := os.RemoveAll(bootstrapDir)
+		require.NoError(t, err)
+	})
```
jordanschalm, line 624:
```suggestion
```

integration/testnet/network.go
jordanschalm: 👍 removed

kc1116 (approved):

peterargue (approved):
thanks for the fix @jordanschalm!

jordanschalm:
bors merge

PR #3847, opened 2023-01-23
remove MarshalJSON/UnmarshalJSON from delta

pattyshack:


SaveTheRbtz (approved):
be careful with that since it might break the code that tries to feed register entries into delta trough `UnmarshalJSON` -- if there is any.

engine/execution/state/delta/delta.go
```diff
@@ -95,30 +92,3 @@ func (d Delta) RegisterIDs() []flow.RegisterID {
...
-
-func (d Delta) MarshalJSON() ([]byte, error) {
-	m := make(flow.RegisterEntries, len(d.Data))
-	for key, value := range d.Data {
-		m = append(m, flow.RegisterEntry{Key: key, Value: value})
-	}
-	return json.Marshal(m)
-}
-
-func (d *Delta) UnmarshalJSON(data []byte) error {
```
janezpodhostnik, line 107: was this never unmarshalled? Interesting.

janezpodhostnik (approved):

pattyshack:
bors merge

pattyshack:
yeah, it's weird.  it looks like we only use marshal for the cmd line exporter, and never unmarshal.  The exporter's marshalling code is pretty weird too.  the fields are individually marshal / repacked.  I don't know why we don't just marshal the snapshot object

PR #3853, opened 2023-01-24
Remove UpdatedAddresses from fvm State

pattyshack:
The implementation is buggy; it doesn't handle DropDelta correctly. Luckily, we only use it for storage check and the bug is never triggered.

### DIFF
--- a/fvm/transactionStorageLimiter.go
+++ b/fvm/transactionStorageLimiter.go
@@ -1,16 +1,27 @@
 package fvm
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/onflow/cadence"
 	"github.com/onflow/cadence/runtime/common"
+	"golang.org/x/exp/slices"
 
 	"github.com/onflow/flow-go/fvm/environment"
 	"github.com/onflow/flow-go/fvm/errors"
+	"github.com/onflow/flow-go/fvm/state"
 	"github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/flow-go/module/trace"
 )
+
+func addressFromRegisterId(id flow.RegisterID) (flow.Address, bool) {
+	if len(id.Owner) != flow.AddressLength {
+		return flow.EmptyAddress, false
+	}
+
+	return flow.BytesToAddress([]byte(id.Owner)), true
+}
 
 type TransactionStorageLimiter struct{}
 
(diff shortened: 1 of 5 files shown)


fvm/transactionStorageLimiter.go
```diff
@@ -33,35 +36,63 @@ func (d TransactionStorageLimiter) CheckStorageLimits(
...
 	if err != nil {
 		return fmt.Errorf("storage limit check failed: %w", err)
 	}
 	return nil
 }
 
-func (d TransactionStorageLimiter) checkStorageLimits(
-	env environment.Environment,
-	addresses []flow.Address,
+func (limiter TransactionStorageLimiter) getStorageCheckAddresses(
```
zhangchiqing, line 46:
```suggestion
// getStorageCheckAddresses returns a list of addresses to be checked whether storage limit is exceeded.
// the returned address should be the accounts whose registers were updated by the transaction
func (limiter TransactionStorageLimiter) getStorageCheckAddresses(
```

fvm/transactionStorageLimiter.go
```diff
@@ -33,35 +36,63 @@ func (d TransactionStorageLimiter) CheckStorageLimits(
...
+	txnState *state.TransactionState,
 	payer flow.Address,
 	maxTxFees uint64,
-) error {
-	// in case the payer is not already part of the check, include it here.
-	// If the maxTxFees is zero, it doesn't matter if the payer is included or not.
+) []flow.Address {
+	updatedIds := txnState.UpdatedRegisterIDs()
+
+	dedup := make(map[flow.Address]struct{}, len(updatedIds)+1)
```
zhangchiqing, line 53:
```suggestion
	// multiple updated registers  might be from the same address, so it's necessary to deduplicate the addresses
	dedup := make(map[flow.Address]struct{}, len(updatedIds)+1)
```

zhangchiqing (approved):
Looks good. some minor suggestions. 

Also I wonder if we should add a test case for the drop delta case

pattyshack:
i'm planning to change the apis quite a bit.  i'll add tests as part of that

fvm/transactionStorageLimiter.go
```diff
@@ -33,35 +44,70 @@ func (d TransactionStorageLimiter) CheckStorageLimits(
...
 	maxTxFees uint64,
-) error {
-	// in case the payer is not already part of the check, include it here.
-	// If the maxTxFees is zero, it doesn't matter if the payer is included or not.
+) []flow.Address {
+	updatedIds := txnState.UpdatedRegisterIDs()
+
+	// Multiple updated registers might be from the same address.  We want to
+	// duplicated the addresses to reduce check overhead.
+	dedup := make(map[flow.Address]struct{}, len(updatedIds)+1)
```
janezpodhostnik, line 66: The value of `len(updatedIds)` is always larger or equal to the accounts updated, but might be a lot larger, because you generally touch more registers per address.

In this instance it might be better to just create the map (and array) without a capacity. Not sure though...

janezpodhostnik (approved):

fvm/transactionStorageLimiter.go
pattyshack: it's ok to over allocate since both the map and slice have short life span, and the memory will be reclaimed almost immediately.  map resizing would cause more gc overhead

pattyshack:
bors merge

PR #3860, opened 2023-01-25
[Execution] Temporary workaround for LN networking issue

peterargue:
Add a cli flag to enable a temporary workaround for a networking issue between a small number of ENs and LNs. When enabled, the EN will only request collections from nodes with a `*.onflow.org` hostname.

### DIFF
--- a/engine/execution/ingestion/engine.go
+++ b/engine/execution/ingestion/engine.go
@@ -1214,18 +1214,16 @@ func (e *Engine) fetchCollection(
 	}
 
 	// This is included to temporarily work around an issue observed on a small number of ENs.
-	// Using only the dapper run LNs seems to aleviate the issue. This will be removed once a
+	// Using only the onflow LNs seems to aleviate the issue. This will be removed once a
 	// proper fix is in place.
-	if os.Getenv("DAPPER_LN_ONLY_OVERRIDE") != "" {
-		var onlyDapperRegex = regexp.MustCompile(`.*onflow.org:3569$`)
+	if os.Getenv("ONFLOW_LN_ONLY_OVERRIDE") != "" {
+		var onlyOnflowRegex = regexp.MustCompile(`.*onflow.org:3569$`)
 
-		// onlyDapper(Identity("verification-049.mainnet20.nodes.onflow.org:3569")) => true
-		// onlyDapper(Identity("verification-049.hello.org:3569")) => false
-		onlyDapper := func(identity *flow.Identity) bool {
-			return onlyDapperRegex.MatchString(identity.Address)
-		}
-
-		filters = append(filters, onlyDapper)
+		// func(Identity("verification-049.mainnet20.nodes.onflow.org:3569")) => true
+		// func(Identity("verification-049.hello.org:3569")) => false
+		filters = append(filters, func(identity *flow.Identity) bool {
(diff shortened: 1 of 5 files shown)


engine/execution/ingestion/engine.go
```diff
@@ -1372,9 +1374,27 @@ func (e *Engine) fetchCollection(blockID flow.Identifier, height uint64, guarant
...
+
+	filters := []flow.IdentityFilter{
+		filter.HasNodeID(guarantors...),
+	}
+
+	// This is included to temporarily work around an issue observed on a small number of ENs.
+	// Using only the onflow LNs seems to aleviate the issue. This will be removed once a
+	// proper fix is in place.
+	if os.Getenv("ONFLOW_LN_ONLY_OVERRIDE") != "" {
+		var onlyOnflowRegex = regexp.MustCompile(`.*onflow.org:3569$`)
```
peterargue, line 1386: I think this is too permissive
```suggestion
		var onlyOnflowRegex = regexp.MustCompile(`.*\.onflow\.org:3569$`)
```

zhangchiqing, line 1386:
```suggestion
		var onlyOnflowRegex = regexp.MustCompile(os.Getenv("ONFLOW_LN_ONLY_OVERRIDE"))
```

are we supposed to use from env var?

engine/execution/ingestion/engine.go
```diff
@@ -1372,9 +1374,27 @@ func (e *Engine) fetchCollection(blockID flow.Identifier, height uint64, guarant
...
 	}
+
+	filters := []flow.IdentityFilter{
+		filter.HasNodeID(guarantors...),
+	}
+
+	// This is included to temporarily work around an issue observed on a small number of ENs.
+	// Using only the onflow LNs seems to aleviate the issue. This will be removed once a
+	// proper fix is in place.
+	if os.Getenv("ONFLOW_LN_ONLY_OVERRIDE") != "" {
```
zhangchiqing, line 1385: do we want to use flag instead of env var?

engine/execution/ingestion/engine.go
peterargue: no, I was thinking it would just be set to `ONFLOW_LN_ONLY_OVERRIDE=1` or something like that.

engine/execution/ingestion/engine.go
peterargue: the reason I used an envvar was so operators wouldn't accidentally discover and use this flag. Also, the cli flags are somewhat of an API for operators, so we should avoid adding and removing flags unnecessarily. Undocumented envvars make no promises

zhangchiqing: We have prefix the flag with `temp_`, and add comments to suggest this is a temporary flag.

I think it's easier to manage with flags than evnvar, as EN prints all the flags that we have configured on startup. So that if we notice any behavior we will be able to check what was the actual flags being used. 

With `envvar`, we would also need to print the env var as well

SaveTheRbtz (approved):

zhangchiqing (approved):

pattyshack (approved):
rubberstamp

peterargue:
bors merge

PR #3871, opened 2023-01-26
[FVM] Add restriction that program loading cannot write to registers

janezpodhostnik:
Just to be more safe, before setting the program in the cache, check the no writes have been performed during program loading.


turbolent (approved):

pattyshack (approved):

janezpodhostnik:
bors merge

Issue #3782, opened 2023-01-06
Make metering interaction less agressive


janezpodhostnik:
## Some Context

1. Interaction metering is done in the FVM to ensure that scripts and transactions don't read or write excessively to storage.
  interaction is measured in bytes:
  
    - every byte read from storage. Reading the same register multiple times doesn't count multiple times as it has already been cached the first time, and does not need to be fetched from storage the next time.
    - every byte written to storage. When writing to the same register multiple times during a transaction only the last write counts towards interactions, since only that is going to be persisted to storage.
    - current interaction limit is 20Mb

2. Direct contract function invocation is when the FMV directly invokes a smart contract function. These are predefined and the relevant ones are: 
- get balance, called when calling `.balance` in cadence
- get available balance, called when calling `.availableBalance` in cadence
- create account, called when calling `AuthAccount(...)` in cadence

## Problem Definition

Interaction metering is aggressive when it comes to direct contract function invocation.

When running a script (or tx) all register reads go through the scripts `Meter` is tallying up the reads. if register `a` is read twice the `Meter` is responsible to keep track and ignore the second read. 

Once a direct contract function invocation is called another (child)   `Meter` is created with the same logic but is independent. Once the direct contract function invocation is done the child meter is merged into the parent. 

This merge is the problematic bit. Currently the interaction used is simply added to the parents interaction used. This does not take into account that if the child `Meter` counted a read for register `b` that the parent `Meter` already counted the total interaction used of the child and parent actually has an overlap and cannot be summed up.

## Proposed Solution

When merging meter interaction limits for a register from meter B to meter A
- `RA[k]`  is the **R**ead of register with **k**ey in meter **A**
- `WB[k]`  is the **W**rite of register with **k**ey in meter **B**
The following rules apply:

1. `RB[k] != nil && WB[k] == nil`:
    1. `RA[k] ==nil && WA[k] == nil` -> do: `RA[k] = RB[k]`
    2. `RA[k] != nil && WA[k] == nil` -> `RA[k]` must be equal to `RB[k]` as B is reading the same register as A did. Nothing to do.
    3. `RA[k] == nil && WA[k] != nil`-> when register k is read in B the value is taken from the changed value that A metered as a write so no storage read happened in B, `RB[k]` must be equal to `WA[k]`. Nothing to do.
[...]

PR #3874, opened 2023-01-27
[FVM] Change interaction metering

janezpodhostnik:
closes: https://github.com/onflow/flow-go/issues/3782

Interaction metering was very aggressive.

When merging sub-meters (as from the importing cadence contracts) the storage reads were simply added together. Which meant that if the parent meter already counted a read for a register that the child meter also counted, the same register will be counted as read twice.

In reality interaction metering is supposed to only meter first read from a register, and last write to a register.

### DIFF
--- a/fvm/meter/interaction_meter.go
+++ b/fvm/meter/interaction_meter.go
@@ -3,6 +3,8 @@ package meter
 import (
 	"math"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/onflow/flow-go/fvm/errors"
 	"github.com/onflow/flow-go/model/flow"
 )
@@ -27,18 +29,24 @@ func (params MeterParameters) WithStorageInteractionLimit(
 	return newParams
 }
 
+// InteractionMeter is a meter that tracks storage interaction
+// Only the first read of a given key is counted
+// Only the last write of a given key is counted
 type InteractionMeter struct {
 	params InteractionMeterParameters
 
-	storageUpdateSizeMap     map[flow.RegisterID]uint64
+	reads  map[flow.RegisterID]uint64
+	writes map[flow.RegisterID]uint64
+
 	totalStorageBytesRead    uint64
 	totalStorageBytesWritten uint64
 }
 
 func NewInteractionMeter(params InteractionMeterParameters) InteractionMeter {
 	return InteractionMeter{
-		params:               params,
-		storageUpdateSizeMap: make(map[flow.RegisterID]uint64),
+		params: params,
+		reads:  make(map[flow.RegisterID]uint64),
+		writes: make(map[flow.RegisterID]uint64),
 	}
 }
 
@@ -51,32 +59,65 @@ func (m *InteractionMeter) MeterStorageRead(
 ) error {
 
 	// all reads are on a View which only read from storage at the first read of a given key
-	if _, ok := m.storageUpdateSizeMap[storageKey]; !ok {
+	if _, ok := m.reads[storageKey]; !ok {
 		readByteSize := getStorageKeyValueSize(storageKey, value)
 		m.totalStorageBytesRead += readByteSize
-		m.storageUpdateSizeMap[storageKey] = readByteSize
+		m.reads[storageKey] = readByteSize
 	}
 
 	return m.checkStorageInteractionLimit(enforceLimit)
 }
 
 // MeterStorageWrite captures storage written bytes count and returns an error
-// if it goes beyond the total interaction limit and limit is enforced
+// if it goes beyond the total interaction limit and limit is enforced.
+// If a key is written multiple times, only the last write is counted.
+// If a key is written before it has been read, next time it will be read it will be from the view,
+// not from storage, so count it as read 0.
 func (m *InteractionMeter) MeterStorageWrite(
 	storageKey flow.RegisterID,
 	value flow.RegisterValue,
 	enforceLimit bool,
 ) error {
-	// all writes are on a View which only writes the latest updated value to storage at commit
-	if old, ok := m.storageUpdateSizeMap[storageKey]; ok {
-		m.totalStorageBytesWritten -= old
+	updateSize := getStorageKeyValueSize(storageKey, value)
+	m.replaceWrite(storageKey, updateSize)
+
+	if _, ok := m.reads[storageKey]; !ok {
+		// write without read, count as read 0 because next time you read it the written value
+		// will be returned from cache, so no interaction with storage will happen.
+		m.reads[storageKey] = 0
 	}
 
-	updateSize := getStorageKeyValueSize(storageKey, value)
-	m.totalStorageBytesWritten += updateSize
-	m.storageUpdateSizeMap[storageKey] = updateSize
-
 	return m.checkStorageInteractionLimit(enforceLimit)
+}
+
+// replaceWrite replaces the write size of a given key with the new size, because
+// only the last write of a given key is counted towards the total interaction limit.
+// These are the only write usages of `m.totalStorageBytesWritten` and `m.writes`,
+// which means that `m.totalStorageBytesWritten` can never become negative.
+// oldSize is always <= m.totalStorageBytesWritten.
+func (m *InteractionMeter) replaceWrite(
+	k flow.RegisterID,
+	newSize uint64,
+) {
+	totalBefore := m.totalStorageBytesWritten
+
+	// remove old write
+	oldSize := m.writes[k]
+	m.totalStorageBytesWritten -= oldSize
+
+	// sanity check
+	// this should never happen, but if it does, it should be fatal
+	if m.totalStorageBytesWritten > totalBefore {
+		log.Fatal().
+			Str("component", "interaction_meter").
+			Uint64("total", totalBefore).
+			Uint64("subtract", oldSize).
+			Msg("totalStorageBytesWritten would have become negative")
+	}
+
+	// add new write
+	m.writes[k] = newSize
+	m.totalStorageBytesWritten += newSize
 }
 
 func (m *InteractionMeter) checkStorageInteractionLimit(enforceLimit bool) error {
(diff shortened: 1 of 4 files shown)


fvm/meter/interaction_meter.go
```diff
@@ -51,32 +57,43 @@ func NewInteractionMeter(params InteractionMeterParameters) *InteractionMeter {
 func (m *InteractionMeter) MeterStorageRead(
 	storageKey StorageInteractionKey,
 	value flow.RegisterValue,
-	enforceLimit bool) error {
+	enforceLimit bool,
+) error {
 
 	// all reads are on a View which only read from storage at the first read of a given key
-	if _, ok := m.storageUpdateSizeMap[storageKey]; !ok {
+	if _, ok := m.reads[storageKey]; !ok {
```
pattyshack, line 64: if you unconditionally set a value (without reading it first), then get it, you will be now count that against both total bytes read and total byte written.  this is more aggressive than the original implementation, which only counts against bytes written in this scenario.

seems like what you really want is to ensure the key is in neither reads / writes sets here, and when you merge, "the sum of the union of child's reads set with parent's writes set" should be discounted (child read entries should not be included in the parent's reads set if the entry is in the parent's write set)

fvm/meter/interaction_meter.go
```diff
@@ -51,32 +57,43 @@ func NewInteractionMeter(params InteractionMeterParameters) *InteractionMeter {
...
+	if old, ok := m.writes[storageKey]; ok {
 		m.totalStorageBytesWritten -= old
 	}
 
 	updateSize := getStorageKeyValueSize(storageKey, value)
 	m.totalStorageBytesWritten += updateSize
-	m.storageUpdateSizeMap[storageKey] = updateSize
+	m.writes[storageKey] = updateSize
+
+	if _, ok := m.reads[storageKey]; !ok {
```
pattyshack, line 92: o, i missed this part.  let me double check if my comment still make sense

pattyshack (approved):

fvm/meter/interaction_meter.go
```diff
@@ -51,30 +57,39 @@ func (m *InteractionMeter) MeterStorageRead(
...
 	}
 
 	return m.checkStorageInteractionLimit(enforceLimit)
 }
 
 // MeterStorageWrite captures storage written bytes count and returns an error
-// if it goes beyond the total interaction limit and limit is enforced
+// if it goes beyond the total interaction limit and limit is enforced.
+// If a key is written multiple times, only the last write is counted.
+// If a key is written before it has been read, next time it will be read it will be from the view,
```
zhangchiqing, line 72: This behavior sounds like the metering module is making assumptions of the actual behavior of reading storage, such as the second time reading from storage will actually read from the view instead. 

However, can the metering module guarantee that the actual behavior matches with the assumption?

Wouldn't it be simplier of we don't need the metering to understand the behavior, but wraps the storage, so that when it's actually being read from storage, then it will be metered?

janezpodhostnik: That's is a very good point. Unfortunately I don't yet fully know how that would work.

The problem is with the programs cache. It stores a detached state that is attached when you get a program from the cache (to replay register reads).  Because of this, when reading a value from the state, you cannot know if that is going to be fetched from the parent state or from storage.

This is a bit complicated, so let me try to illustrate it better with an example.
Lets say that during loading of program *MyContract* you need to read register *R*. After this contract is first used in a transaction it will be in the cache, together with a *State* (a delta / a recording) of all registers touched during loading of *MyContract* including the read of register *R*.

When this contract will be used from the cache in future transaction there are 2 possibilities for register *R*:
- *R* was not touched (get-ed) before loading/applying *MyContract* from the programs cache, in which case *R* should be counted as being retrieved from storage
- *R* was touched (get-ed) before loading/applying *MyContract* from the programs cache, in which case *R* should be counted as being retrieved from the parent state

This logic is easier to do in meter.Merge then applying in State.Merge and then fixing the metered values from there.

fvm/meter/interaction_meter.go
```diff
@@ -117,14 +132,31 @@ func GetStorageKeyValueSizeForTesting(
...
+func (m *InteractionMeter) GetStorageRWSizeMapForTesting() (
+	reads MeteredStorageInteractionMap,
+	writes MeteredStorageInteractionMap,
+) {
+	return m.reads, m.writes
 }
 
+// Merge merges the child interaction meter into the parent interaction meter
+// Prioritise parent reads because they happened first
+// Prioritise child writes because they happened last
```
zhangchiqing, line 144:
```suggestion
// Prioritise parent reads because they happened first, and child reads will 
//    read from the view (cache) even if they are updated
// Prioritise child writes because they happened last, and parent writes will be overwritten by child writes.
```

fvm/meter/interaction_meter.go
```diff
@@ -117,14 +132,31 @@ func GetStorageKeyValueSizeForTesting(
...
+			continue
+		}
+
+		m.reads[key] = value
+		m.totalStorageBytesRead += value
+	}
+
+	for key, value := range child.writes {
+		parentWriteBytes := m.writes[key]
+		m.totalStorageBytesWritten -= parentWriteBytes
```
zhangchiqing, line 158: ⚠️  Note, this might be negative. If the write is to reduces the data size to be stored or even completely delete the registers, then totalStorageBytesWritten might become negative, will it cause underflow?

fvm/meter/interaction_meter.go
```diff
@@ -117,14 +132,31 @@ func GetStorageKeyValueSizeForTesting(
...
+		m.writes[key] = value
+		m.totalStorageBytesWritten += value
```
zhangchiqing, line 160: we do `-=` before `+=`, so the first `-=` would cause underflow. 

can we change `totalStorageBytesWritten` into `int64`?

fvm/meter/interaction_meter.go
janezpodhostnik: That cannot happen. I extracted the relevant code into a method and added comments to better illustrate why. Please let me know if it clearer now.

fvm/meter/interaction_meter.go
```diff
@@ -51,34 +57,55 @@ func (m *InteractionMeter) MeterStorageRead(
...
+// These are the only write usages of `m.totalStorageBytesWritten` and `m.writes`,
+// which means that `m.totalStorageBytesWritten` can never become negative.
+// oldSize is always <= m.totalStorageBytesWritten.
+func (m *InteractionMeter) replaceWrite(
+	k flow.RegisterID,
+	newSize uint64,
+) {
+	// remove old write
+	oldSize := m.writes[k]
+	m.totalStorageBytesWritten -= oldSize
```
zhangchiqing, line 102: can we add a sanity check that after this operation, `m.totalStorageBytesWritten` is not greater than before.

janezpodhostnik: I added a log instead of an error, as the error would be difficult to propagate out (the `Merge` function intentionally doesn't return an error).

I also set the `totalStorageBytesWritten` to 0 in that case. If this does happen (I still don't see how), worst that will happen is that the interaction is measured incorrectly.

fvm/meter/interaction_meter.go
```diff
@@ -51,34 +59,68 @@ func (m *InteractionMeter) MeterStorageRead(
...
+
+	// sanity check
+	// this should never happen, but if it does, we want to know about it
+	if m.totalStorageBytesWritten > totalBefore {
+		log.Error().
+			Str("component", "interaction_meter").
+			Uint64("total", totalBefore).
+			Uint64("subtract", oldSize).
+			Msg("totalStorageBytesWritten would have become negative")
+		m.totalStorageBytesWritten = 0
```
zhangchiqing, line 116: Is this a fatal error? if so, we should just crash the node, rather than log and continue. This type of error is easy to be missed and ignored. 
```suggestion
		log.Fatal().
			Str("component", "interaction_meter").
			Uint64("total", totalBefore).
			Uint64("subtract", oldSize).
			Msg("totalStorageBytesWritten would have become negative")
```

janezpodhostnik: I would not treat it as fatal, as it only influences how quickly transactions hit the interaction limit, and nothing else. On the other hand, this should not (is not able to) happen so it could be fatal.

I'm a bit torn about it, but I decided to go with error. What do you think?

fvm/meter/interaction_meter.go
```diff
@@ -117,14 +159,28 @@ func GetStorageKeyValueSizeForTesting(
...
+// Merge merges the child interaction meter into the parent interaction meter
+// Prioritise parent reads because they happened first
+// Prioritise child writes because they happened last
 func (m *InteractionMeter) Merge(child InteractionMeter) {
-	for key, value := range child.storageUpdateSizeMap {
-		m.storageUpdateSizeMap[key] = value
+	for key, value := range child.reads {
+		_, parentRead := m.reads[key]
+		if parentRead {
+			continue
```
zhangchiqing, line 176:
```suggestion
			// avoid metering the same read more than once, because a second read
			// is from the cache
			continue
```

fvm/meter/interaction_meter.go
```diff
@@ -51,34 +59,68 @@ func (m *InteractionMeter) MeterStorageRead(
...
-	}
-
 	updateSize := getStorageKeyValueSize(storageKey, value)
-	m.totalStorageBytesWritten += updateSize
-	m.storageUpdateSizeMap[storageKey] = updateSize
+	m.replaceWrite(storageKey, updateSize)
+
+	if _, ok := m.reads[storageKey]; !ok {
+		// write without read, count as read 0 because next time you read it the written value will be returned
+		// so no interaction with storage will happen.
```
zhangchiqing, line 86:
```suggestion
		// write without read, count as read 0 because next time you read it the written value 
		// from cache will be returned, so no interaction with storage will happen.
```

fvm/meter/interaction_meter.go
zhangchiqing: I still think this should be a `Fatal` error. As the reason running into here is unknown, in the worst case, some node ran into this error, and some didn't, then different nodes would produce different transaction results, which create a result fork. That would take much bigger effort to fix than simply crashing the node here, where all nodes are stopped at the same height with a consistent state.

zhangchiqing (approved):

janezpodhostnik:
bors merge

PR #3701, opened 2022-12-10
[Networking] Bump libp2p v0.23.3 -> v0.24.2

SaveTheRbtz:
Updated dependencies:
`https://github.com/libp2p/go-libp2p-tls` -> `github.com/libp2p/go-libp2p/p2p/security/tls`

This is our last dependency on the deprecated `github.com/libp2p/go-libp2p-core`:
```
$ go mod why github.com/libp2p/go-libp2p-core
# github.com/libp2p/go-libp2p-core
(main module does not need package github.com/libp2p/go-libp2p-core)
```

Full release notes: https://github.com/libp2p/go-libp2p/releases/tag/v0.24.0

While here, also removed the `replace mellium.im/sasl => github.com/mellium/sasl v0.2.1`. Module is not used:
```
$ go mod why mellium.im/sasl
# mellium.im/sasl
(main module does not need package mellium.im/sasl)
```

PR #3875, opened 2023-01-27
[Networking] Refactors test scenarios to comply with libp2p upgrade

yhassanzadeh13:
This PR refactors some test scenarios to comply with the libp2p upgrade to v0.24 https://github.com/onflow/flow-go/pull/3701. 

With the upgrade, we discovered that when node `A` connection gates node `B`, node `B` does not face an error dialing `A` or creating a stream to it. Nevertheless, no connection would be established in practice between nodes `A` and `B`. As our connection gater tests relied heavily on error checking, they would keep failing. With this PR we change the testing scenarios to look into more robust shreds of evidence. In particular, our updated tests will check that: 
- When a connection is established, it allocates resources on the remote peer (i.e., receiver) side, e.g., memory, stream, service, protocol, etc. 
- When a connection is rejected by the remote peer (i.e., connection gated), **it does not take any resources at the remote side**. So there is no evident concern of resource leakage.
- Even though the local peer does not receive an error, when its connection is rejected by the remote peer, no connections and streams are established in practice.


kc1116 (approved):

SaveTheRbtz:
What is the libp2p maintainers view on the issue?  It sound like node B should receive at least some error when trying to connect to a gated node?

network/internal/p2pfixtures/fixtures.go
```diff
@@ -334,10 +342,16 @@ func EnsureNoStreamCreation(t *testing.T, ctx context.Context, from []p2p.LibP2P
...
-			// stream creation should fail
-			_, err := this.CreateStream(ctx, other.Host().ID())
-			require.Error(t, err)
-			require.True(t, flownet.IsPeerUnreachableError(err))
+			// we intentionally do not check the error here, with libp2p v0.24 connection gating at the "InterceptSecured" level
+			// does not cause the nodes to complain about the connection being rejected at the dialer side.
+			// Hence, we instead check for any trace of the connection being established in the receiver side.
+			otherId := other.Host().ID()
+			thisId := this.Host().ID()
+			_, err := this.CreateStream(ctx, otherId)
```
SaveTheRbtz, line 350:
```suggestion
			_, _ := this.CreateStream(ctx, otherId)
```

SaveTheRbtz (approved):

network/internal/p2pfixtures/fixtures.go
yhassanzadeh13: We still need this for external error checking when test needs: 

https://github.com/onflow/flow-go/blob/f20b8c85872bd07097fefde1fadb07bd5ef3ca07/network/internal/p2pfixtures/fixtures.go#L358

PR #3880, opened 2023-01-29
[Provider] Temporarily disable weight check

peterargue:
The staking contract was upgraded on devnet recently and resulted in all ANs having their weight reset to 0. This caused collection nodes to reject requests for collections, breaking the access api.

This PR updates the Provider engine to disable the requirement that the requester has weight.

This is a temporary workaround until the staking contract is fixed and the nodes' protocol state is refreshed

### DIFF
--- a/engine/common/provider/engine.go
+++ b/engine/common/provider/engine.go
@@ -78,7 +78,8 @@ func New(
 	// make sure we don't respond to request sent by self or unauthorized nodes
 	selector = filter.And(
 		selector,
-		filter.HasWeight(true),
+		// TODO: disabling this on TESTNET only to work around an issue in the staking contract where all ANs were reset to 0 weight
+		// filter.HasWeight(true),
 		filter.Not(filter.HasNodeID(me.NodeID())),
 	)
 


Kay-Zee (approved):

engine/common/provider/engine.go
```diff
@@ -78,7 +78,8 @@ func New(
 	// make sure we don't respond to request sent by self or unauthorized nodes
 	selector = filter.And(
 		selector,
-		filter.HasWeight(true),
+		// TODO: disabling this on TESTNET only to work around an issue in the staking contract where all ANs were reset to 0 weight
+		// filter.HasWeight(true),
```
jordanschalm, line 82: I would suggest to make this change permanent (remove `HasWeight` from this selector by default), and change the specific instances of `provider` which need it to enforce `weight>0`.
- In general, sometimes nodes with zero weight should be able to send/receive certain network messages. This is enforced on a per-message basis, it isn't a blanket policy either way
- The `provider` engine is generic, so it should be able to support either allowing or disallowing zero-weight requests
- The way the `IdentityFilter` functions work, including `HasWeight(true)` by default can't be adjusted with an input `selector` to allow 0-weight requests
- For this case in particular, I think we _should_ allow requests from 0-weight (but un-ejected) Access nodes. Access nodes begin populating their transaction index as soon as they come online. If they are unable to get collections until the first epoch transition after they join, they will have a lot of catch-up to do afterward -- there would be a substantial backlog of requests to collection nodes to get through.

peterargue: opened https://github.com/onflow/flow-go/pull/3923

PR #3884, opened 2023-01-31
Upgrade Cadence to latest version (v0.35.0)

janezpodhostnik:
implements cadence API changes from:

 https://github.com/onflow/cadence/pull/2222
 https://github.com/onflow/cadence/pull/2254
 
and more from https://github.com/onflow/flow-go/pull/3938


fvm/errors/errors.go
```diff
@@ -114,18 +113,6 @@ func HandleRuntimeError(err error) error {
 		return NewUnknownFailure(err)
 	}
 
-	// External errors are reported by the runtime but originate from the VM.
-	// External errors may be fatal or non-fatal, so additional handling by SplitErrorTypes
-	if externalErr, ok := errors.GetExternalError(err); ok {
```
janezpodhostnik, line 119: I think `GetExternalError` is no longer doing anything useful, and can be deleted (from cadence) @onflow/cadence

fvm/meter/memory_meter.go
```diff
@@ -139,17 +138,19 @@ var (
...
-		common.MemoryKindDictionaryEntry: 33,
+		common.MemoryKindProgram:           220,
+		common.MemoryKindIdentifier:        17,
+		common.MemoryKindArgument:          49,
+		common.MemoryKindBlock:             25,
+		common.MemoryKindFunctionBlock:     25,
+		common.MemoryKindParameter:         25,
+		common.MemoryKindParameterList:     59,
+		common.MemoryKindTypeParameter:     42, // TODO: add actual memory
+		common.MemoryKindTypeParameterList: 42, // TODO: add actual memory
```
janezpodhostnik, line 149: This needs to be set before merging.

fvm/fvm_signature_test.go
```diff
@@ -721,7 +722,8 @@ func TestBLSMultiSignature(t *testing.T) {
 						)
 
 						err := vm.Run(ctx, script, view)
-						assert.Error(t, err)
+						assert.NoError(t, err)
+						assert.Error(t, script.Err)
```
janezpodhostnik, line 726: Due to the new `Unwrap` some errors are now user errors instead of implementation errors. We need to check if that is correct behavior in this case.

janezpodhostnik:
An emulator upgrade is also needed together with this

fvm/meter/memory_meter.go
SupunS: can use the following values:
```go
common.MemoryKindTypeParameter:     9
common.MemoryKindTypeParameterList: 59,
```

fvm/errors/errors.go
turbolent: Not sure here, I'm not familiar with it

fvm/fvm_signature_test.go
turbolent: As far as I understand the test case, which works with BLS signatures in a Cadence program, working with invalid BLS signatures should be a user error. Could you please add an assertion for the kind of error that is returned?

fvm/errors/errors.go
SupunS: Cadence can still return `ExternalErrors` (e.g: atree related non-user errors), and they should be treated as fatal errors here I guess. So maybe we should keep this?

SupunS: To add more context: With https://github.com/onflow/cadence/pull/2286, `externalErr.Recovered` is always an `error`, so this will never return a `NewUnknownFailure(externalErr)`.

So the only difference this snippet makes is to whether wrap the error with `NewCadenceRuntimeError` or not. I'm not sure what difference it makes to the FVM on wrapping/not-wrapping. My gut feeling is that `ExternalError`s, like atree/fvm errors, should not be wrapped as a Cadence errors. But totally up to the FVM.

janezpodhostnik: What [get external error](https://github.com/onflow/cadence/blob/158f50017f319875180512aeb6d2650e0fa72b1b/runtime/errors/errors.go#L215) is doing is just manually unwrapping an external error:

```
// GetExternalError returns the ExternalError in the error chain, if any
func GetExternalError(err error) (ExternalError, bool) {
	switch err := err.(type) {
	case ExternalError:
		return err, true
	case xerrors.Wrapper:
		return GetExternalError(err.Unwrap())
	default:
		return ExternalError{}, false
	}
}
```

That is why I think its no longer needed.

fvm/fvm_signature_test.go
janezpodhostnik: I assumed that was the case, so I already did that.
```
assert.NoError(t, err) // < -- not implementation error
assert.Error(t, script.Err) // < -- but user error
```

It looks like this was previously reported incorrectly, and will now be ok.

fvm/errors/errors.go
SupunS: Had an offline chat with Janez, think it should be OK to remove

fvm/environment/programs.go
```diff
@@ -176,6 +178,45 @@ func (programs *Programs) get(
...
 
+// GetAndSetProgram gets the program from the cache,
+// or loads it (by calling load) if it is not in the cache.
+// When loading a program, this method will be re-entered
+// to load the dependencies of the program.
+//
+// TODO: this function currently just calls GetProgram and SetProgram in pair.
+// This method can be re-written in a far better way by removing the individual
+// GetProgram and SetProgram methods.
+func (programs *Programs) GetAndSetProgram(
```
pattyshack, line 189: is it too late to rename this to GetOrLoadProgram?

janezpodhostnik: cc: @turbolent @SupunS

janezpodhostnik: Get or load seems better to me

turbolent: Yeah, we can change it. Could you please open an issue or even PR in the Cadence repo?

turbolent:
onflow/cadence#2298 simplified lookeaheads by reducing the use of replaying tokens. 

That means the test which attempts to deploy a contract which is "too complex" needed to get updated. I fixed it in [d4da190](https://github.com/onflow/flow-go/pull/3884/commits/d4da190de7ce85db2c9893ac4ea4a7476de6bcd6), and also separated the tests into two.

turbolent (approved):

SupunS (approved):

turbolent:
bors merge

PR #3887, opened 2023-01-31
[Access] Fix panic converting execution data to protobuf messages

peterargue:
This PR fixes a panic in the state stream API when handling an execution data for a block with an empty service chunk. It also relaxes validation of transactions from execution data to handle service transactions.

Finally, it introduces a new method `execution_data.CalculateID()` which allows you to calculate the execution data id from a `BlockExecutionData`. This enables consumers of the state stream API to verify the data they receives matches the data included in the sealed `ExecutionResult`


koko1123 (approved):

lolpuddle (approved):

peterargue:
bors merge

PR #3898, opened 2023-02-03
[Networking] Enables logging of peer ids together with node id upon connection establishment

yhassanzadeh13:
This PR wires in an identity provider to the connection gater to ensure that an `info` level log is always printed displaying the remote `peer_id` together with its `flow.Identifier` upon dialing or rejecting a connection.


peterargue (approved):
thanks for adding this!

kc1116 (approved):

PR #3900, opened 2023-02-03
Ensure NewRegisterID initialize 8 byte owners

pattyshack:


model/flow/ledger.go
```diff
@@ -78,6 +78,13 @@ func CadenceRegisterID(owner []byte, key []byte) RegisterID {
 	}
 }
 
+func NewRegisterID(owner, key string) RegisterID {
+	return RegisterID{
+		Owner: addressToOwner(BytesToAddress([]byte(owner))),
```
zhangchiqing, line 83:
```suggestion
		// left-pad owner with zeros to ensure owner has 8 bytes.
		Owner: addressToOwner(BytesToAddress([]byte(owner))),
```

zhangchiqing (approved):

model/flow/ledger.go
pattyshack: i'll keep this as is for now.  this conversion won't be needed in the future (i'm planning to replace Owner's type from string to Address)

SaveTheRbtz (approved):

pattyshack:
bors merge

